PR #208, opened 2026-04-24
Ignore subagent hook events fired after a turn's Stop (#199)

### LINKED ISSUE #199: Session state sticks at "working" after Claude Code emits a session recap
## Summary

Claude Code added a **session recap** feature (shipped in **v2.1.108 on 2026-04-14**, controlled by the `awaySummaryEnabled` setting) that prints a one-line summary at the end of a turn, for example:

```
✻ Brewed for 35s

※ recap: Adding a GitHub Actions workflow to push Corveil images to commercial ECR on
  merge to main (ticket #324); PR #326 is open and ready for review. Next: merge it,
  then run the post-merge ECR verification. (disable recaps in /config)
```

After this prints, Crow's session indicator stays on **`working`** and never transitions back to **`done`**, even though the agent is idle and ready for input.

## Why this matters

The whole value of the status indicator is knowing when a session is done so I can pick up the next one. With the recap feature on (which is the default for recent Claude Code versions), every session I finish looks like it's still running. I end up force-checking each one manually.

## Root cause — not terminal parsing, but hook events

Crow does not grep terminal output for `✻` or `※ recap:` — it transitions state purely from Claude Code hook events via the `hook-event` RPC handler:

- `Sources/Crow/App/AppDelegate.swift:802-968` — hook-event dispatcher
- `Packages/CrowCore/Sources/CrowCore/Models/Enums.swift:33-38` — `ClaudeState { idle, working, waiting, done }`
- `AppDelegate.swift:872` — `PreToolUse` → `.working`
- `AppDelegate.swift:930-931` — `UserPromptSubmit` → `.working`
- `AppDelegate.swift:933-935` — `Stop` → `.done`

So the sequence for a normal turn is: `UserPromptSubmit` → `.working` → (tool calls) → `Stop` → `.done`.

The hypothesis: when the recap generates, Claude Code re-enters a generation cycle *after* the turn has already emitted `Stop`. That secondary cycle likely fires one of `UserPromptSubmit` / `PreToolUse` / `SubagentStart`, which flips us back to `.working` — but the closing event for that cycle is either missing, uses a different event name (e.g. `SubagentStop` not mapped to `.done`), or is swallowed, so we never transition back.

**This needs to be confirmed by capturing the actual hook-event stream around a recap.** Possible outcomes:

1. **Recap runs as a subagent**: fires `SubagentStart` → `.working`, then `SubagentStop` which is not currently mapped back to `.done` at line 933-935.
2. **Recap fires a `UserPromptSubmit`** (synthetic, internal prompt): flips to `.working`, then completes with a `Stop` that arrives but is missed because of ordering (e.g. two `Stop`s collapse, or state is clobbered by a later event).
3. **Recap doesn't emit any hook events at all** and the `.working` state we're seeing is left over from something else the session did right before the recap rendered.

Option 1 is the most likely fit — the recap is essentially a short summarization task, and Claude Code's recap docs describe it as running in the background.

## Reproduction

1. Use Claude Code ≥ v2.1.108 (recap defaults to on).
2. Start a Crow session on any ticket that takes more than a couple of turns to complete.
3. Let the agent finish naturally (it will print a `※ recap: ... (disable recaps in /config)` line at the end).
4. Observe that the Crow sidebar keeps showing the session as `working` indefinitely.

## Suggested fix

Investigate first, then implement:

1. **Instrument** `AppDelegate.swift` hook-event handler to log every inbound event name + source/session ID for a single reproduction. Capture what actually fires around a recap.
2. Based on what we see, either:
   - Add a mapping for `SubagentStop` (or whatever closing event the recap uses) → `.done` in the `Stop`/`StopFailure` block at `AppDelegate.swift:933-938`.
   - Or: tag events originating from the recap generator (if Claude Code identifies them via a `source`, `agentType`, or similar field on the hook payload) and ignore them for state-transition purposes so the state stays `.done` across the recap.

Do NOT try to detect the recap by pattern-matching terminal text for `※ recap:` — it sidesteps the actual hook-lifecycle bug and will rot the next time the recap format changes.

## Workaround for users in the meantime

Users can disable the feature in one of three ways:

- `settings.json`: `\"awaySummaryEnabled\": false`
- `/config` inside Claude Code → toggle \"Session recap\" off
- Env: `CLAUDE_CODE_ENABLE_AWAY_SUMMARY=0`

Worth calling out in the README or the \"Known issues\" note once we know the exact behavior — but the real fix belongs in the hook-event handler.

## References

- Claude Code changelog entry for v2.1.108 (2026-04-14): session recap feature.
- Claude Code interactive-mode docs: describe recap as background-generated and invokable via `/recap`.
- Claude Code settings docs: `awaySummaryEnabled`.

dhilgaertner:
## Summary

Fixes #199. Claude Code 2.1.108's `awaySummaryEnabled` "session recap" generates a `SubagentStop` hook event ~minutes after the user's turn has ended. Crow's old `TaskCreated, TaskCompleted, SubagentStop` arm unconditionally drove `claudeState` back to `.working`, trapping the sidebar dot until session end.

The fix tracks `lastTopLevelStopAt: Date?` on `SessionHookState`. When set, the dispatcher suppresses state elevation in the `SubagentStart` and `TaskCreated/TaskCompleted/SubagentStop` arms — those events are treated as background. The flag is cleared on `UserPromptSubmit` (next real turn), `SessionStart`, and `SessionEnd`, so legitimate mid-turn subagents (`/explore`, sub-tasks, etc.) are unaffected.

This generalizes beyond recap: any future "Claude wakes up after a turn ends" scenario (background telemetry, async cleanup hooks) will be correctly ignored.

## Evidence the fix works

Captured with the new `CROW_HOOK_DEBUG=1` flag during a real session that produced a recap:

```
16:22:32  UserPromptSubmit  state=done→working      ← turn begins
16:22:45  Stop              state=working→done      ← lastTopLevelStopAt set
16:23:45  Notification      (no state change)
16:25:48  SubagentStop      payload=[agent_id, agent_transcript_path, agent_type, …]
          ↑ NO state-change line — guard suppressed it
```

Pre-fix: that `SubagentStop` would drive `done → working` and never recover. Post-fix: `lastTopLevelStopAt != nil`, so `claudeState` stays at `.done`. A separate session that was actively mid-turn during the same capture window correctly kept its state at `.working` while running `TaskCreated`/`TaskCompleted` events, confirming the guard is scoped to post-`Stop`.

## Why hook events, not terminal text

Per the ticket: text-pattern matching `※ recap:` would rot the next time Claude Code changes the recap glyph or copy. The fix lives in the hook dispatcher where the actual lifecycle bug is.

## Also included

- `CROW_HOOK_DEBUG=1`-gated `[hook-event]` NSLog stream — logs every event arrival + every `ClaudeState` transition. Off by default. Kept in for the next time Claude Code's hook schema shifts (it changes often). Documented in `docs/troubleshooting.md`.
- Troubleshooting docs row covering the user-side workaround for older builds (`awaySummaryEnabled = false` / `/config` / `CLAUDE_CODE_ENABLE_AWAY_SUMMARY=0`).

## Test plan

- [x] All 31 existing tests pass (`make test`)
- [x] Reproduced original bug behavior would have triggered (`SubagentStop` arrives 3 min after user-turn `Stop`)
- [x] Confirmed sidebar stays at done after recap renders
- [x] Confirmed mid-turn subagents still elevate state to working (regression check on a separate session in the same capture window)
- [ ] Reviewer: try a session with `/explore` or another subagent-heavy slash command — sidebar should still show working during the run

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -353,6 +353,12 @@ public final class SessionHookState {
     public var lastToolActivity: ToolActivity?
     public var hookEvents: [HookEvent] = []
     public var analytics: SessionAnalytics?
+    /// Timestamp of the most recent top-level Stop / StopFailure for this session.
+    /// Used to suppress state elevation from background activity (e.g. the
+    /// `awaySummaryEnabled` recap subagent in Claude Code ≥ 2.1.108) that
+    /// fires after the user's turn has ended. Cleared on the next
+    /// UserPromptSubmit, which marks the start of a new real turn.
+    public var lastTopLevelStopAt: Date?
 
     public init() {}
 }
--- a/Sources/Crow/App/AppDelegate.swift
+++ b/Sources/Crow/App/AppDelegate.swift
@@ -468,6 +468,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         let capturedNotifManager = notificationManager
         let capturedService = sessionService
         let capturedTelemetryPort = sessionService.telemetryPort
+        let hookDebug = ProcessInfo.processInfo.environment["CROW_HOOK_DEBUG"] == "1"
 
         let router = CommandRouter(handlers: [
             "new-session": { @Sendable params in
@@ -809,6 +810,12 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                 }
                 let payload = params["payload"]?.objectValue ?? [:]
 
+                if hookDebug {
+                    let shortID = String(sessionIDStr.prefix(8))
+                    let keys = payload.keys.sorted().joined(separator: ",")
+                    NSLog("[hook-event] session=\(shortID) event=\(eventName) payload-keys=[\(keys)]")
+                }
+
                 // Build a human-readable summary from the event
                 let summary: String = {
                     switch eventName {
@@ -858,6 +865,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
                 return await MainActor.run {
                     let state = capturedAppState.hookState(for: sessionID)
+                    let stateBefore = state.claudeState
 
                     // Append to ring buffer (keep last 50 events per session)
                     state.hookEvents.append(event)
@@ -931,13 +939,18 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
                     case "UserPromptSubmit":
                         state.claudeState = .working
+                        // A new real turn has begun — clear the post-Stop guard so
+                        // legitimate subagents in this turn can elevate state again.
+                        state.lastTopLevelStopAt = nil
 
                     case "Stop":
                         state.claudeState = .done
                         state.lastToolActivity = nil
+                        state.lastTopLevelStopAt = Date()
 
                     case "StopFailure":
                         state.claudeState = .waiting
+                        state.lastTopLevelStopAt = Date()
 
                     case "SessionStart":
                         let source = payload["source"]?.stringValue ?? "startup"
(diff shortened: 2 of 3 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues

None.

### Security Review

**Strengths:**
- `CROW_HOOK_DEBUG` reads from the process environment once at startup and uses a simple `Bool` — no injection surface, no runtime overhead when disabled.
- Debug logging truncates session IDs to 8 chars (`shortID`) — safe, but even full UUIDs aren't sensitive here since they're local-only.
- `Date()` stored on `lastTopLevelStopAt` is never serialized or sent externally — pure in-memory state.

**Concerns:**
- None identified. No new external inputs, no new persistence paths, no new network calls.

### Code Quality

**Well done:**
- The `lastTopLevelStopAt` guard is the right abstraction level — it's event-schema-agnostic, so it naturally handles future background-event scenarios without needing pattern updates per event type.
- Clearing the guard on `UserPromptSubmit`, `SessionStart`, and `SessionEnd` covers all legitimate "turn is live" transitions. No gap I can find where a real mid-turn subagent would be incorrectly suppressed.
- Capturing `stateBefore` before the switch block and logging the transition only when the state actually changed is clean — no noise in debug output.
- The troubleshooting docs include the three workaround paths (settings.json, `/config`, env var) for users on older builds — good completeness.

**One observation (not blocking):**
- `lastTopLevelStopAt` stores a `Date` but the value is only ever checked for `nil` vs non-nil — a plain `Bool` like `hasStopped` would be slightly simpler. That said, keeping the timestamp costs nothing extra and could be useful for future diagnostics (e.g. "how long ago did the turn end?"), so this is fine as-is.

### Summary Table

| Priority | Issue |
|----------|-------|
| 🟢 | `lastTopLevelStopAt` could be a `Bool` instead of `Date?` — but `Date?` is arguably better for future debug use |

**Recommendation:** Approve. The fix is well-scoped, correctly guarded, and all 112 CrowCore tests pass. The approach is resilient to future Claude Code hook schema changes since it keys off the turn lifecycle (`Stop` → `UserPromptSubmit`) rather than matching specific event names or payloads.

PR #209, opened 2026-04-24
Auto-start review sessions for opted-in repos

### LINKED ISSUE #193: Auto-review when assigned as reviewer
## Summary

Automatically start a review session when the user is assigned as a reviewer on a PR, with per-project opt-in.

## Details

- Add a setting for a CSV list of projects to enable auto-review on (opt-in per project)
- When a review request is detected for an enabled project, automatically kick off the review process
- Should integrate with the existing notification/event system for review assignments

dhilgaertner:
## Summary
- Adds a per-workspace **Auto-Review Repos** CSV field (next to **Always Include Repos**) in the Workspace editor.
- When `IssueTracker` detects a new review request whose repo slug matches any workspace's `autoReviewRepos` (case-insensitive), `AppDelegate` automatically kicks off `SessionService.createReviewSession(prURL:)` — the same path as the manual **Start Review** button.
- Dedups via the existing `ReviewRequest.reviewSessionID` cross-reference, so app restarts don't create duplicate sessions. Normal review-request notifications still fire.

## Test plan
- [x] `make build` — clean build
- [x] `swift test --package-path Packages/CrowCore` — 114 tests pass, including new `workspaceAutoReviewReposRoundTrip` and `workspaceAutoReviewReposDefaultsEmptyWhenKeyMissing`
- [x] Open Settings → Workspaces → edit a workspace, set **Auto-Review Repos** to `org/repo`, save
- [x] Trigger a review request on that repo; within ~60s verify a review session auto-appears in the sidebar and is selected
- [x] Restart the app with the review still open — confirm no second session is created (dedup)
- [x] Clear the repo from the setting — confirm new review requests only notify, don't auto-create

Closes #193

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -61,6 +61,7 @@ public struct WorkspaceInfo: Identifiable, Codable, Sendable, Equatable {
     public var cli: String            // "gh" or "glab" — kept for config file compat
     public var host: String?          // GitLab host (e.g., "gitlab.example.com")
     public var alwaysInclude: [String] // repos to always list in prompt table
+    public var autoReviewRepos: [String] // repos where review requests auto-create a review session
 
     /// The CLI tool name derived from the current `provider` value.
     /// Unlike `cli` (which may be stale from an old config file), this is always correct.
@@ -74,7 +75,8 @@ public struct WorkspaceInfo: Identifiable, Codable, Sendable, Equatable {
         provider: String = "github",
         cli: String = "gh",
         host: String? = nil,
-        alwaysInclude: [String] = []
+        alwaysInclude: [String] = [],
+        autoReviewRepos: [String] = []
     ) {
         self.id = id
         self.name = name
@@ -82,6 +84,22 @@ public struct WorkspaceInfo: Identifiable, Codable, Sendable, Equatable {
         self.cli = cli
         self.host = host
         self.alwaysInclude = alwaysInclude
+        self.autoReviewRepos = autoReviewRepos
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        id = try container.decode(UUID.self, forKey: .id)
+        name = try container.decode(String.self, forKey: .name)
+        provider = try container.decode(String.self, forKey: .provider)
+        cli = try container.decode(String.self, forKey: .cli)
+        host = try container.decodeIfPresent(String.self, forKey: .host)
+        alwaysInclude = try container.decodeIfPresent([String].self, forKey: .alwaysInclude) ?? []
+        autoReviewRepos = try container.decodeIfPresent([String].self, forKey: .autoReviewRepos) ?? []
+    }
+
+    private enum CodingKeys: String, CodingKey {
+        case id, name, provider, cli, host, alwaysInclude, autoReviewRepos
     }
 
     /// Characters that are unsafe in directory names (workspace names become directory names).
--- a/Packages/CrowCore/Tests/CrowCoreTests/AppConfigTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AppConfigTests.swift
@@ -130,6 +130,25 @@ import Testing
     #expect(stale.derivedCLI == "glab")
 }
 
+@Test func workspaceAutoReviewReposRoundTrip() throws {
+    let config = AppConfig(workspaces: [
+        WorkspaceInfo(name: "Org", autoReviewRepos: ["org/repo1", "org/repo2"])
+    ])
+    let data = try JSONEncoder().encode(config)
+    let decoded = try JSONDecoder().decode(AppConfig.self, from: data)
+    #expect(decoded.workspaces[0].autoReviewRepos == ["org/repo1", "org/repo2"])
+}
+
+@Test func workspaceAutoReviewReposDefaultsEmptyWhenKeyMissing() throws {
+    // Legacy configs without the key should default to empty (feature off).
+    let json = """
+    {"workspaces": [{"id": "00000000-0000-0000-0000-000000000001", "name": "Org", "provider": "github", "cli": "gh"}]}
+    """.data(using: .utf8)!
+    let config = try JSONDecoder().decode(AppConfig.self, from: json)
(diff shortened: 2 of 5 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Case-insensitive repo matching (`lowercased()`) prevents bypass via casing differences.
- Dedup via `reviewSessionID` cross-reference prevents duplicate sessions on app restart.
- No user-controlled input reaches shell commands — `createReviewSession` uses the existing trusted `request.url` and the PR URL is parsed/validated before use.
- The `onNewReviewRequests` callback runs on the `@MainActor` (both `IssueTracker` and `AppDelegate` are `@MainActor`-isolated), so accessing `appConfig` is thread-safe.

**Concerns:**
- None identified.

### Code Quality

**Model layer (`AppConfig.swift`)** — Clean addition. Custom `init(from:)` with `decodeIfPresent` + `?? []` handles backward compatibility with existing config files gracefully. The explicit `CodingKeys` enum is the right approach since the default synthesized Codable won't use `decodeIfPresent`.

**Tests (`AppConfigTests.swift`)** — Good coverage: round-trip encoding/decoding and the legacy-config-missing-key case. The legacy test also validates that `alwaysInclude` defaults correctly, which is a nice bonus.

**UI (`WorkspaceFormView.swift`)** — Straightforward mirror of the existing `alwaysInclude` CSV field. Parsing logic (split → trim → filter empty) is consistent with the existing pattern. Frame bump from 300→400 height accommodates the new field.

**Auto-start logic (`AppDelegate.swift`)** — The `enabledRepos` list is recomputed inside the loop on each call rather than cached outside, but since `onNewReviewRequests` only fires with genuinely new requests (delta detection in `IssueTracker`), the batch size is always small so this is fine.

**Minor observation (non-blocking):** If two new review requests for the *same* auto-review repo arrive in a single polling cycle (e.g., two PRs opened simultaneously), both would pass the `reviewSessionID == nil` guard and spawn separate `createReviewSession` tasks. This is an extremely unlikely edge case and each session would still be valid — just noting it for awareness. Not worth adding complexity to handle.

### Summary Table
| Priority | Issue |
|----------|-------|
| 🟢 | Consider caching `enabledRepos` outside the loop if the workspace list grows large (not needed now) |
| 🟢 | Theoretical double-session if two PRs from same repo arrive in one poll cycle (extremely unlikely) |

**Recommendation:** Approve — clean, well-tested feature addition with correct backward compatibility handling and no security concerns.

dgershman (approved):
## Code & Security Review

### Critical Issues
None found.

### Security Review
**Strengths:**
- No secrets, credentials, or sensitive data introduced
- Case-insensitive repo matching prevents bypass via casing tricks (`lowercased()` on both sides)
- Dedup guard (`request.reviewSessionID == nil`) prevents re-creating sessions on app restart — well-thought-out
- `[weak self]` capture in the closure avoids retain cycles
- Custom `Codable` init uses `decodeIfPresent` with `?? []` default, so existing config files without the new key decode safely

**Concerns:**
- None significant. The `autoReviewRepos` field is only used in an equality check against the tracker-supplied `repo` slug; there is no path traversal, injection, or privilege-escalation vector.

### Code Quality

1. **`enabledRepos` rebuilt per request** (`AppDelegate.swift:235-237`): The `flatMap` + `lowercased()` runs inside the `for request in newRequests` loop. For a handful of requests this is negligible, but if you wanted to tighten it, hoisting it above the loop is a one-line change. Not worth blocking on.

2. **No dedup guard in `createReviewSession` itself**: The dedup relies on `request.reviewSessionID == nil` at the call site. If two review requests for the same PR arrive in the same poll cycle (unlikely but theoretically possible with re-assignments), two `Task { createReviewSession(...) }` calls could race. In practice the clone-and-checkout step serializes on the filesystem and the window is tiny, but a future hardening pass could add an in-flight set. Again, non-blocking for this PR scope.

3. **Form frame bump** (`WorkspaceFormView.swift:116`): `400×300 → 440×400` — reasonable for the new field. Looks fine.

4. **Tests are solid**: `workspaceAutoReviewReposRoundTrip` and `workspaceAutoReviewReposDefaultsEmptyWhenKeyMissing` cover the happy path and backward-compat decode. Good coverage.

### Summary Table
| Priority | Issue |
|----------|-------|
| 🟢 Green | `enabledRepos` could be hoisted above the loop |
| 🟢 Green | Concurrent same-PR race is theoretically possible but practically harmless |

**Recommendation:** Approve — clean, well-scoped feature addition. The model change is backward-compatible, the UI is straightforward, the auto-start logic correctly uses existing dedup, and test coverage is present. Ship it.

dgershman (approved):
## Code & Security Review

### Critical Issues
None found.

### Security Review
**Strengths:**
- New field defaults to empty (`[]`), so the feature is opt-in only — no behavior change for existing users
- Custom `init(from:)` uses `decodeIfPresent` with `?? []` fallback, ensuring backward compatibility with legacy config files that lack the key
- Case-insensitive repo matching (`lowercased()`) prevents bypass via casing tricks
- Dedup via `request.reviewSessionID == nil` guard prevents duplicate review sessions across app restarts
- No new user input flows to shell commands — repo slugs are only compared as strings

**Concerns:**
- None significant. The `autoReviewRepos` values are only used for in-memory string comparison, never interpolated into commands or queries.

### Code Quality

**AppConfig.swift** — Clean addition. The custom `CodingKeys` enum and `init(from:)` are necessary to support the `decodeIfPresent` fallback pattern. Consistent with how `alwaysInclude` is handled.

**AppConfigTests.swift** — Good coverage: round-trip encoding test and a legacy-config-missing-key test. The missing-key test also validates that `alwaysInclude` defaults correctly in the same scenario, which is a nice touch.

**WorkspaceFormView.swift** — The CSV parsing logic for `autoReviewReposText` mirrors the existing `alwaysIncludeText` pattern exactly. Frame bump from 400×300 → 440×400 is reasonable for the added field.

**AppDelegate.swift** — The `[weak self]` capture is correct. One minor observation: `enabledRepos` is recomputed inside the `for` loop for every request, but in practice the request batch is small and the workspace list is tiny, so this is negligible.

### Summary Table
| Priority | Issue |
|----------|-------|
| Green | `enabledRepos` could be hoisted above the `for` loop for minor efficiency, but not worth a change request |

**Recommendation:** Approve — well-scoped feature, backward-compatible, no security concerns, good test coverage.

PR #214, opened 2026-04-24
Auto-respond to PR status changes (#201)

### LINKED ISSUE #201: Auto-detect and respond to PR status changes (changes requested, failed pipeline)
## Feature Request

Crow should automatically detect PR status changes and take action without user intervention.

## Desired Behavior

- **Changes Requested**: When a PR receives a "changes requested" review, Crow should automatically read the review comments and begin addressing the feedback.
- **Failed Pipeline**: When a CI/CD pipeline fails on a PR, Crow should automatically read the failure logs and attempt to fix the issue.

## Use Case

Currently the user has to manually notice these status changes, switch to the relevant session, and instruct Claude to fix the problem. This is tedious when managing multiple PRs across workspaces.

Automating this would make Crow a true "set it and forget it" workflow — open a PR, and Crow handles the feedback loop until it's merged.

## Possible Approach

- Poll or subscribe to PR status via GitHub/GitLab API
- On status change, activate the relevant session and send appropriate instructions
- Could leverage existing `crow send` infrastructure to inject prompts into the session's terminal

dhilgaertner:
## Summary
- Detects PR transitions into `changesRequested` / `checksFailing` by comparing each polling cycle's `PRStatus` against the previous one — piggybacks on the existing 60s GitHub GraphQL query, no new API calls or scopes.
- When the user opts in (Settings → Notifications → Auto-respond), Crow types a short prompt into the session's managed Claude Code terminal asking Claude to fetch the review/logs (`gh`/`glab`) and address them. Crow doesn't fetch review bodies or CI logs itself — that keeps the implementation simple and avoids fragile log parsing.
- Fires a macOS notification on every transition (gated by the existing notification settings) so the user is informed even with auto-respond off.
- Dedupe key is `(sessionID, kind, headSha)` so re-runs on the same commit don't re-fire; auto-rearms when the rule clears (approved → changesRequested again, or new commit also fails).
- Both opt-in toggles default **off** — typing into a terminal unprompted is intrusive.
- First poll never fires (matches the existing `previousReviewRequestIDs` first-fetch gate) so existing PR state isn't replayed at startup.

## Implementation
- New `PRStatusTransition` value type + pure `PRStatus.transitions(from:to:...)` helper in CrowCore — keeps the comparison logic unit-testable.
- Added `headSha` to `PRStatus` (populated from a new `headRefOid` field on `ViewerPR` / GraphQL).
- Two new `NotificationEvent` cases (`changesRequested`, `checksFailing`) — they automatically inherit `NotificationSettings`'s per-event sound/system-notification toggles.
- New `AutoRespondSettings` struct on `AppConfig` (separate from `NotificationSettings` because auto-respond is an action, not a notification).
- New `AutoRespondCoordinator` in the main target — finds the session's managed terminal via `Terminal.isManaged` and sends via `TerminalManager.shared.send(id:text:)`, the same path `crow send` uses.
- `IssueTracker.applyPRStatuses` rewritten to track `previousPRStatus[sessionID]` and emit transitions through a new `onPRStatusTransitions` closure (mirrors `onNewReviewRequests`).
- `NotificationManager.notifyPRTransition` reuses the same gate logic as `notifyReviewRequest`.

Closes #201

## Test plan
- [x] `make build` succeeds
- [x] `make test` — all 127 CrowCore tests + 31 root CrowTests + every other package suite pass
- [x] New `PRStatus transitions` test suite (14 tests) covers: first-observation never fires, single-fire on transition, no-op on repeat poll, no-fire when rule clears, simultaneous transitions, dedupe key shape (with vs without SHA)
- [x] Existing `ViewerPR` test helpers updated for new `headRefOid` field
- [ ] Manual smoke (recommend running once after merge): enable `Respond to failed CI checks`, push a CI-breaking commit on a session-linked PR, verify notification fires within ~60s and prompt appears in the managed terminal; re-run CI on same SHA → no second prompt; push new failing SHA → second prompt fires; disable toggle → notification only


dgershman (approved):
## Code & Security Review

### Critical Issues
None found.

### Security Review

**Strengths:**
- Auto-respond toggles default **off** — good security posture for a feature that types into a terminal unsolicited.
- Dedupe key design (`sessionID|kind` / `sessionID|kind|headSha`) prevents replay attacks from re-polling the same state.
- The coordinator checks for both a managed terminal AND an initialized surface before dispatching — prevents sending to stale/orphaned terminals.
- No new API scopes or tokens — piggybacks on the existing `gh`/`glab` GraphQL payload and delegates review/CI log fetching to Claude, keeping the Crow app's privilege boundary unchanged.
- The `headRefOid` field is sourced from GitHub's GraphQL schema, not user input, so no injection risk from SHA values.
- Prompt text is constructed with string interpolation of PR numbers/URLs (already validated upstream via `ProviderManager.parseTicketURLComponents`), not raw user input.

**Concerns:**
- None significant. The prompts injected into the terminal are deterministic and constructed from trusted state (session metadata, PR metadata from GitHub API).

### Code Quality

**Well done:**
- Clean separation: `PRStatusTransition` is a pure value type in CrowCore, `PRStatus.transitions(from:to:...)` is a pure function — both are unit-testable without any UI/AppState dependency.
- 14-test suite (`PRStatusTransitionTests`) thoroughly covers the transition matrix: first-observation gate, directional transitions, no-ops, simultaneous transitions, and dedupe key semantics.
- `AutoRespondSettings` is separate from `NotificationSettings` — correct modeling since auto-respond is an action, not a notification.
- Forward-compatible `Codable` with `decodeIfPresent` + defaults on `AppConfig`, `AutoRespondSettings`, and `PRStatus`.
- Re-arm logic in `applyPRStatuses` (lines 1326–1334 of IssueTracker.swift) correctly clears emitted keys when the triggering condition clears, so future re-entries fire as expected.
- The `settingsProvider` closure pattern on `AutoRespondCoordinator` means Settings UI toggles take effect on the next transition without explicit wiring — simple and correct.
- `NotificationSettingsView` cleanly adds the auto-respond section with appropriate explanatory caption text.
- Prompt construction in `AutoRespondPrompts.build` handles both GitHub and GitLab providers with appropriate CLI hints.

- `AutoRespondPrompts.build` uses `transition.prNumber.map(String.init)` for the inline comments hint but falls back to `"<number>"` as a placeholder when `prNumber` is nil (line 82). In practice `prNumber` is always populated from `ViewerPR.number`, so this is a safe defensive path rather than a real concern.
- The `failedCheckNames.prefix(5)` truncation in the prompt (line 95) is a nice touch to avoid overwhelming Claude with a massive list.

### Summary Table

| Priority | Issue |
|----------|-------|
| Green | Consider adding an integration test that exercises the full `IssueTracker.applyPRStatuses → onPRStatusTransitions → AutoRespondCoordinator.handle` pipeline end-to-end (currently each piece is tested in isolation, which is fine) |
| Green | The `AutoRespondCoordinator` could log when a transition is skipped due to the toggle being off (currently it silently `continue`s) — helpful for debugging "why didn't auto-respond fire?" |

**Recommendation:** Approve — this is a well-structured, defensively coded feature with good test coverage, appropriate opt-in defaults, and no security concerns. The pure/stateful split keeps the logic testable and the implementation stays within Crow's existing privilege boundary by delegating review/CI log fetching to Claude.

dgershman (approved):
## Code & Security Review

### Critical Issues
None found.

### Security Review
**Strengths:**
- Auto-respond is **off by default** — typing into a terminal unprompted is correctly treated as an intrusive action, requiring explicit opt-in per transition type.
- Crow delegates all review/CI log fetching to Claude via `gh`/`glab` rather than fetching and parsing them itself — avoids new API scopes and keeps the attack surface small.
- The prompt text is constructed from data Crow already has in memory (PR URL, PR number, check names), not from user-controlled free-text fields like review bodies. This limits injection risk in the terminal prompt.
- The `settingsProvider` closure reads settings lazily, so a toggle-off takes effect on the very next poll cycle — no stale-permission window.
- `@MainActor` on `AutoRespondCoordinator` serializes all dispatch calls, avoiding TOCTOU races between the settings check and the terminal send.

**Concerns:**
- **(Low)** `AutoRespondPrompts.build` interpolates `transition.prURL` directly into the prompt string. Since `prURL` comes from the GitHub GraphQL API (not user input), this is safe in practice — but if the app ever ingests PR URLs from other sources (e.g. user-pasted links), a malicious URL containing backticks or shell metacharacters could be interpreted by the terminal. The current trust boundary is fine; just worth noting.

### Code Quality

**Well done:**
- Clean separation of concerns: pure `PRStatus.transitions()` for logic, `IssueTracker` for stateful dedupe, `AutoRespondCoordinator` for dispatch, `NotificationManager` for user-facing alerts. Each is independently testable.
- The dedupe key design is thoughtful — `changesRequested` keys on `(session, kind)` to be SHA-agnostic, while `checksFailing` includes SHA so a new failing commit re-fires. The test suite covers both.
- The re-arm logic in `applyPRStatuses` correctly clears emitted keys when the triggering condition clears, preventing stale dedupe entries from accumulating.
- 14 new transition tests with good coverage of edge cases: first-observation gate, no-op on repeat, re-arm, simultaneous transitions, dedupe key shape.
- Backward-compatible `Codable` with `decodeIfPresent` for both `AutoRespondSettings` and the new `headSha` on `PRStatus`.

**Minor observations:**
- `IssueTracker.swift:1413-1416` — the re-arm for `checksFailing` guards on `old.headSha != nil`, but the dedupe key for `headSha == nil` is `"…|checksFailing|"` (empty string). If checks transition to failing with `headSha == nil`, that key will be emitted but never re-armed. In practice `headSha` is always populated from `headRefOid` in the GraphQL response, so this is cosmetic — but adding an `else` branch to also remove the empty-SHA key would make the logic fully symmetric.
- `AutoRespondPrompts.build` line 82: the GitHub inline-comments hint includes a literal `{owner}/{repo}` placeholder. Claude will figure it out, but interpolating the actual owner/repo from `transition.prURL` would make the prompt more actionable. Very minor.

### Summary Table
| Priority | Issue |
|----------|-------|
| 🟢 Green | Re-arm doesn't cover `headSha == nil` edge case (cosmetic — SHA always present in practice) |
| 🟢 Green | GitHub inline-comments `gh api` hint uses literal `{owner}/{repo}` placeholder |

**Recommendation:** Approve — this is a clean, well-tested feature with solid architecture, good defaults (off by default), and no blocking issues. The two green items are cosmetic and can be addressed in a follow-up if desired.

dgershman (approved):
## Code & Security Review

### Critical Issues
None found.

### Security Review
**Strengths:**
- Both auto-respond toggles default **off** — good security posture for a feature that types into a terminal unprompted.
- Dedupe key `(sessionID, kind, headSha)` prevents replay attacks from repeat polling cycles. Re-arm logic is sound: keys are removed when the triggering condition clears.
- The `AutoRespondCoordinator` checks for both a managed terminal and an initialized surface before sending, avoiding blind writes.
- No new API scopes or secrets — prompts delegate `gh`/`glab` fetching to Claude, keeping Crow's permission surface unchanged.
- Prompt text is built from trusted internal state (session metadata, PR URLs from GitHub GraphQL), not from user/reviewer-controlled content — no injection vector into the terminal.

**Concerns:**
- None identified. The prompt content in `AutoRespondPrompts.build` is constructed from PR numbers and URLs already held in trusted app state.

### Code Quality

**Well done:**
- Clean separation: `PRStatusTransition` is a pure value type in CrowCore, testable without AppState/UI. The stateful dedupe lives in `IssueTracker`, and the coordinator is a thin dispatch layer.
- `PRStatus.transitions(from:to:...)` returning `[]` on first observation (`old == nil`) mirrors the existing `previousReviewRequestIDs` first-fetch gate — consistent behavior, no startup flood.
- `settingsProvider` closure pattern in `AutoRespondCoordinator` avoids stale-settings bugs without explicit re-wiring when the user toggles preferences.
- 14 new unit tests with good edge coverage: first-observation gate, both transition directions, simultaneous transitions, dedupe key shape with/without SHA.
- Existing test fixtures updated for the new `headRefOid` field — no broken tests left behind.

**Minor observations:**
- `AutoRespondCoordinator.dispatch` looks up the session's `provider` with a default of `.github` (`?? .github`). This is reasonable for the current user base but worth noting: if a GitLab-only user has a session with a nil provider, the prompt will suggest `gh` commands. Low risk since nil-provider sessions are rare and the prompt is advisory.
- The `NotificationSettingsView` now takes both `settings: Binding<NotificationSettings>` and `autoRespond: Binding<AutoRespondSettings>` — the API grew but it's justified since auto-respond is semantically distinct from notification preferences.

### Summary Table
| Priority | Issue |
|----------|-------|
| 🟢 | GitLab fallback defaults to `.github` provider in `AutoRespondCoordinator.dispatch` — low risk, consider a comment |
| 🟢 | `headRefOid` added to GraphQL query — verify GitHub token scopes haven't changed (they haven't, `headRefOid` is on `PullRequest` type, no new scope needed) |

**Recommendation:** Approve — clean implementation, good test coverage, no security concerns. The feature is well-scoped with sensible defaults (off) and proper dedupe. The pure-transition logic in CrowCore and the coordinator's lazy settings pattern are particularly well designed.

dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review

**Strengths:**
- Both auto-respond flags default **off** — correct posture for a feature that types into a live terminal unprompted.
- Prompt text is injected via `TerminalManager.shared.send(id:text:)`, not evaluated by a shell directly, which substantially limits injection risk.
- No new GitHub/GitLab API scopes required — the feature piggybacks on the existing 60s GraphQL poll.
- `headSha` used for CI dedupe so re-runs on the same commit don't re-fire.

**Concerns (informational only):**
- `failedCheckNames` in `AutoRespondPrompts.build` are sourced from the GitHub API and embedded verbatim into the injected prompt text (capped at 5). A check run whose name contains a newline could inject an extra line into the terminal input. In practice GitHub validates check names, but worth a `replacingOccurrences(of: "\n", with: " ")` guard if this surfaces.

### Code Quality

- **Duplicated dedupeKey strings in re-arm logic** (`IssueTracker.swift:805,809-810`): the re-arm code hand-constructs the same key strings that `PRStatusTransition.dedupeKey` computes (`"\(session.id.uuidString)|changesRequested"`, etc.). If the key format in `PRStatusTransition.dedupeKey` ever changes, the re-arm paths won't update automatically. Consider a static helper or a dedicated `rearmKey(for kind:sessionID:sha:)` factory to keep both sides in sync.
- **Stale entries in `previousPRStatus` / `emittedTransitionKeys`**: sessions that are archived or deleted will leave entries in these dictionaries indefinitely. Not a correctness issue today, but a long-lived session churn could cause unbounded growth. Consider pruning on session deletion.
- Everything else is clean: the `@MainActor` coordinator, the `settingsProvider` closure pattern, the first-poll-never-fires gate, and the pure-value `PRStatus.transitions(from:to:)` helper are all well-designed. Test coverage is thorough (14 focused unit tests covering all transition edges including dedupe key shape).

### Summary Table

| Priority | Issue |
|----------|-------|
| Yellow | Duplicated dedupeKey string construction in re-arm logic — maintenance coupling risk |
| Green | Guard `failedCheckNames` against embedded newlines in prompt text |
| Green | Prune `previousPRStatus`/`emittedTransitionKeys` on session deletion |

**Recommendation:** Approve — no blocking issues. The feature is well-scoped, defaults are conservative, tests are solid, and the architecture is clean. The yellow item above is a low-risk maintenance concern, not a correctness bug.

dgershman (approved):
## Code & Security Review

### Critical Issues
None found.

### Security Review

**Strengths:**
- Auto-respond toggles default **off** — typing into a terminal unprompted is intrusive, and this PR correctly treats it as opt-in.
- The `settingsProvider` closure pattern reads settings lazily, so toggling auto-respond off takes effect on the very next transition without requiring a restart or explicit propagation.
- No new API scopes introduced — the auto-respond prompts delegate all external data fetching (`gh pr view`, `gh run view`) to Claude Code inside the terminal, so Crow itself doesn't need additional GitHub permissions.
- Dedupe key design is well-considered: `changesRequested` keys on `(session, kind)` while `checksFailing` keys on `(session, kind, headSha)`, correctly reflecting that a new failing commit should re-fire but repeated "changes requested" polls should not.
- Forward-compatible `Codable` init with `decodeIfPresent` + defaults on both `AutoRespondSettings` and `PRStatus.headSha` — existing config files won't break.

**Concerns:**
- None significant. The `AutoRespondCoordinator` only sends text through the existing `TerminalManager.shared.send` path (same as `crow send`), so there's no new attack surface.

### Code Quality

- **`PRStatusTransition` as a pure value type** — clean separation of the transition-detection logic (pure, in CrowCore) from the stateful dedupe (in `IssueTracker`) and the side-effectful dispatch (in `AutoRespondCoordinator`). Each layer is independently testable.
- **`AutoRespondPrompts.build` using `\r` instead of `\n`** — the commit message and inline comment (`AutoRespondCoordinator.swift:72-81`) explain the rationale clearly: `\r` is delivered as a PTY byte that submits regardless of focus, while `\n` triggers the synthetic-Return key-event path which is focus-dependent. Good catch on an edge case that would've been subtle to debug.
- **Re-arm logic in `IssueTracker.applyPRStatuses`** (`IssueTracker.swift:1417-1425`) — correctly clears the emitted dedupe key when the triggering condition clears (e.g., `changesRequested → approved`), so a subsequent regression fires again. Mirrors the existing `previousReviewRequestIDs` pattern.
- **Test coverage** is solid: 14 new tests covering first-observation gate, single-fire, no-op on repeat, re-arm, simultaneous transitions, and dedupe key shape.
- **`NotificationManager.notifyPRTransition`** (`NotificationManager.swift:151`) intentionally skips the foreground/visible suppression that `handleEvent` uses — the comment explains why (the user typically wants to know about review feedback even when looking at the session). Good design decision.
- Minor: `NotificationSettingsView` now takes an `autoRespond` binding and the `SettingsView` threads it through cleanly. The UI is grouped under a separate "Auto-respond" section with clear help text.

### Summary Table

| Priority | Issue |
|----------|-------|
| Green | Consider: the `AutoRespondCoordinator` silently skips when there's no managed terminal surface — this is correct behavior but the only feedback is an `NSLog`. If a user enables auto-respond and wonders why nothing happened, they'd need to check logs. A future enhancement could surface a one-time hint in the notification. |
| Green | Consider: `AutoRespondPrompts` hardcodes the PR number format string for inline comments (`gh api repos/{owner}/{repo}/pulls/<number>/comments`). If `prNumber` is nil, it falls back to `<number>` as a literal placeholder. This is fine since the prompt is directed at Claude which can resolve the actual number, but it's worth noting. |

**Recommendation:** Approve — clean implementation with good separation of concerns, solid test coverage, sensible defaults (both toggles off), and no security concerns. The `\r` vs `\n` terminal submission fix shows good attention to edge cases. Ship it.

dgershman (approved):
## Code & Security Review

### Critical Issues
None found.

### Security Review
**Strengths:**
- Auto-respond toggles default **off** — the most intrusive action (typing into a terminal unprompted) requires explicit opt-in.
- Dedupe key `(sessionID, kind, headSha)` prevents runaway prompt injection on repeated poll cycles. The re-arm logic (clearing emitted keys when the triggering condition clears) is correct and well-tested.
- No new API scopes or network calls — piggybacks on the existing 60s GraphQL query. The `headRefOid` field addition is read-only from the GitHub API.
- `AutoRespondCoordinator` reads settings lazily via a closure (`settingsProvider`), so toggling the setting off takes effect on the next transition without a restart.
- The coordinator guards on both `isManaged` (won't type into a user shell tab) and `existingSurface` (won't attempt to send to an uninitialized PTY).
- Prompt text is deterministic (no user-controlled interpolation beyond `prNumber`/`prURL` which are integers and GitHub URLs from the GraphQL response) — no injection risk.

**Concerns:**
- None identified. The prompt payloads are hardcoded templates with safe interpolations (PR number, URL from trusted GraphQL source).

### Code Quality

**Well done:**
- Clean separation: `PRStatusTransition` is a pure value type in CrowCore, `PRStatus.transitions(from:to:...)` is a pure function — both fully unit-testable without AppState/UI dependencies.
- The `\r` termination strategy for auto-respond prompts is well-documented and correctly avoids the focus-dependent `\n` → synthetic-Return path in Ghostty. The comment block in `AutoRespondPrompts` explains the rationale clearly.
- 14 new transition tests cover: first-observation gate, each direction of each transition, simultaneous transitions, dedupe key semantics. Good edge-case coverage.
- `AutoRespondSettings` uses the same `decodeIfPresent` pattern as the rest of `AppConfig` for forward compatibility with existing config files.
- The `IssueTracker.applyPRStatuses` re-arm logic correctly clears emitted transition keys when the triggering condition clears (e.g., `changesRequested` → `approved`), allowing future re-entries to fire.
- UI: auto-respond section is logically grouped under Notifications with a clear explanatory caption.

- `AutoRespondCoordinator.dispatch` silently skips when there's no managed terminal or no surface. The `NSLog` messages are good for debugging, but there's no user-visible indication that auto-respond was skipped. This seems intentional (the notification fires regardless), just noting it.
- The `failedCheckNames` list is capped at 5 in the prompt but the `PRStatusTransition` struct carries the full list. Fine — the prompt is the only consumer and it applies its own truncation.

### Summary Table
| Priority | Issue |
|----------|-------|
| 🟢 | All toggles default off — safe-by-default |
| 🟢 | Pure transition logic with thorough test coverage (14 new tests) |
| 🟢 | No new API calls or scopes required |
| 🟢 | Correct dedupe + re-arm prevents duplicate fires and allows recovery |
| 🟢 | Build succeeds, all 129 CrowCore + 31 CrowTests pass |

**Recommendation:** Approve — clean, well-tested feature with appropriate safety defaults. The implementation is well-scoped: Crow detects transitions and delegates the actual review/CI investigation to Claude Code via the terminal, keeping the codebase simple. No security concerns.

PR #218, opened 2026-04-29
Recover from failed Ghostty surface creation (#217)

### LINKED ISSUE #217: Auto-review session stuck on "Waiting for terminal" when Ghostty createSurface() fails
## Summary

Auto-review created a review session overnight but the terminal never appeared — the session hung indefinitely on the "Waiting for terminal" screen. The underlying cause appears to be a failed Ghostty surface creation, after which `TerminalManager` happily returns the existing (broken) view on subsequent lookups instead of failing the session or retrying.

## Repro

1. Run latest crow on a Mac mini, leave it overnight with auto-review enabled and a watched repo that receives a "request changes" / review-requested event.
2. Auto-review fires (in this case for `radiusmethod/tanzanite#155`).
3. In the morning the session exists in the sidebar but the terminal pane is stuck on "Waiting for terminal".

## Logs (2026-04-29, ~08:24 local)

```
[SessionService] Cloning radiusmethod/tanzanite into /Users/dustinhilgaertner/Projects/crow-reviews/tanzanite-pr-155
[TerminalManager] trackReadiness for 06CF4874-E79D-4D01-B76C-700995FB6518
[TerminalManager] preInitialize(06CF4874-...) — creating surface in offscreen window
[GhosttyApp] Unhandled action: tag=ghostty_action_tag_e(rawValue: 41)
[Ghostty] createSurface() FAILED — surface is nil
[SessionService] Created review session 'review-tanzanite-155' for https://github.com/radiusmethod/tanzanite/pull/155
[TerminalManager] surface(for: 06CF4874-...) — returning EXISTING view
[GhosttyApp] Unhandled action: tag=ghostty_action_tag_e(rawValue: 41)
[Ghostty] createSurface() FAILED — surface is nil
```

After that point the only repeating log entries are `[IssueTracker] refresh:` lines — there are no further terminal-related events, which means there's no retry, no surfacing of the failure, and no path out of the stuck state without restarting the app.

## Observations / hypotheses

- `Ghostty.createSurface()` returned nil, but the terminal id `06CF4874-...` was still registered with `TerminalManager`. The next `surface(for:)` call returned the existing broken entry rather than treating nil-surface as "needs reinit".
- The unhandled `ghostty_action_tag_e(rawValue: 41)` from `GhosttyApp` happens immediately before each surface failure — could be related; worth confirming what action 41 corresponds to in the current Ghostty SDK build.
- The session was created (`Created review session 'review-tanzanite-155'`) *after* the first surface failure, so session creation doesn't gate on terminal readiness. That's fine in general, but combined with the lack of retry, it means the user-visible state is "session exists, terminal forever pending".

## Expected behavior

At minimum one of:

1. `createSurface()` failure triggers a bounded retry on the same terminal id, with backoff.
2. After N failures the session surfaces an error state in the UI ("Terminal failed to launch — retry / open logs") instead of "Waiting for terminal" forever.
3. `TerminalManager.surface(for:)` should not return an entry whose surface is nil — treat that as a miss and re-`preInitialize`.

Auto-review specifically should also probably mark the session as failed (or auto-retry the whole review) if the terminal can't come up within some deadline, so overnight runs don't silently no-op.

## Environment

- Latest crow (running overnight on a Mac mini)
- Auto-review enabled
- Triggering event: review requested on `radiusmethod/tanzanite#155`

dhilgaertner:
## Summary

Closes #217

Auto-review sessions could hang on "Waiting for terminal" forever when
`ghostty_surface_new` returned nil — the broken `GhosttySurfaceView`
stayed cached in `TerminalManager`, no retry was attempted, and the UI
had no failure path. This implements all three remediations called out
in the ticket:

- **Bounded retry** inside `GhosttySurfaceView.createSurface()`: 4
  attempts at 0.5s / 1.0s / 2.0s / 4.0s before giving up. On exhaustion
  it fires a new `onSurfaceCreationFailed` callback.
- **`TerminalManager.surface(for:)` rejects broken cached views**: if
  the cached entry's `hasSurface` is false, it's destroyed and replaced
  instead of returned.
- **Failure surfaces in the UI**: a new `TerminalReadiness.failed`
  state (also `SurfaceState.failed`) flows from
  `surfaceDidFail` → `wireTerminalReadiness` → `appState.terminalReadiness`,
  and `ReadinessAwareTerminal` renders an error overlay with a **Retry**
  button. Retry routes through a new `appState.onRetryTerminal`
  callback to `SessionService.retryTerminal(terminalID:)`, which resets
  readiness, re-arms auto-launch, and asks `TerminalManager` to destroy
  the broken view and re-`preInitialize`.

The session-list status dot also gets a red "failed" variant so the
problem is visible from the sidebar.

Out of scope: a session-level "failed" `SessionStatus`. The
terminal-level state is enough to recover; adding a new session status
would ripple through CLI, persistence, and validation.

## Test plan

- [ ] \`make build\` succeeds.
- [ ] Force the failure path (e.g., temporarily set \`self.surface = nil\`
      after \`ghostty_surface_new\` in \`GhosttySurfaceView.createSurface\`)
      and confirm the retry log fires 4 times with backoff, then
      \`[Ghostty] createSurface() exhausted retries\` is logged.
- [ ] After retries exhaust, the UI shows the error overlay with a
      Retry button instead of an indefinite "Waiting for terminal" spinner.
- [ ] Clicking Retry (with the injection removed) creates a fresh
      \`GhosttySurfaceView\`, reaches \`.shellReady\`, and auto-launches Claude.
- [ ] Navigating away from a tab with a broken cached surface and back
      logs \`cached view has nil surface, discarding\` and rebuilds the view.
- [ ] Happy-path regression: ordinary work-session and review-session
      preinit still reach \`.shellReady\` within ~5s and auto-launch Claude.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -181,6 +181,9 @@ public final class AppState {
     /// Called to launch Claude in a terminal that just became ready.
     public var onLaunchClaude: ((UUID) -> Void)?  // receives terminal ID
 
+    /// Called when the user clicks "Retry" on a failed terminal surface.
+    public var onRetryTerminal: ((UUID) -> Void)?  // receives terminal ID
+
     /// Called to add a new plain-shell terminal tab to a session.
     public var onAddTerminal: ((UUID) -> Void)?  // receives session ID
 
--- a/Packages/CrowCore/Sources/CrowCore/Models/Enums.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/Enums.swift
@@ -39,6 +39,7 @@ public enum ClaudeState: String, Codable, Sendable {
 
 /// Terminal surface lifecycle state.
 public enum TerminalReadiness: String, Codable, Sendable, Comparable {
+    case failed           // createSurface() exhausted retries; UI shows error overlay with Retry
     case uninitialized    // GhosttySurfaceView exists but createSurface() not called
     case surfaceCreated   // ghostty_surface_t exists, shell process spawning
     case shellReady       // Shell prompt detected (probe file appeared)
@@ -46,6 +47,7 @@ public enum TerminalReadiness: String, Codable, Sendable, Comparable {
 
     private var sortOrder: Int {
         switch self {
+        case .failed: -1
         case .uninitialized: 0
         case .surfaceCreated: 1
         case .shellReady: 2
--- a/Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift
@@ -8,11 +8,19 @@ public final class GhosttySurfaceView: NSView {
     private var markedTextStorage = NSMutableAttributedString()
     private var trackingArea: NSTrackingArea?
 
+    /// Backoff schedule for retrying createSurface() when ghostty_surface_new returns nil.
+    /// 4 retries totalling ~7.5s before declaring permanent failure.
+    private static let retryDelays: [TimeInterval] = [0.5, 1.0, 2.0, 4.0]
+    private var createAttempts: Int = 0
+
     /// Whether the Ghostty surface has been created (needs window attachment first).
     public var hasSurface: Bool { surface != nil }
 
     /// Called after createSurface() succeeds.
     public var onSurfaceCreated: (() -> Void)?
+
+    /// Called after createSurface() exhausts its retry budget without producing a surface.
+    public var onSurfaceCreationFailed: (() -> Void)?
 
     /// The working directory for the shell spawned in this surface.
     public var workingDirectory: String?
@@ -82,6 +90,7 @@ public final class GhosttySurfaceView: NSView {
         updateTrackingAreaInternal()
 
         if surface != nil {
+            createAttempts = 0
             NSLog("[Ghostty] createSurface() succeeded, hasCallback=\(onSurfaceCreated != nil)")
             // Flush any text that arrived before the surface was ready
             if !pendingText.isEmpty {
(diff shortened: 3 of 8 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None found.

### Security Review
**Strengths:**
- All new state and callbacks are `@MainActor`-isolated, consistent with the existing concurrency model.
- No new external inputs or command injection surfaces introduced — retry logic operates purely on internal state.
- `TerminalManager.retry()` properly destroys the old surface before re-creating, preventing resource leaks.
- The `onSurfaceCreationFailed` callback is cleared in `destroy()`, avoiding retain-cycle or use-after-free issues.

**Concerns:**
- None. The changes are internal to the app's process and don't introduce new trust boundaries.

### Code Quality

**Well-structured changes:**
- Clean separation of concerns: `GhosttySurfaceView` owns retry mechanics, `TerminalManager` bridges to `SessionService`, and `SessionService` owns the recovery/re-arm logic.
- The `TerminalReadiness.failed` enum case is inserted with `sortOrder: -1`, correctly ordering it below all other states. The `Comparable` conformance means the existing `< .shellReady` guard in the loading overlay naturally covers the new case without modification.
- `retryDelays` as a static constant with exponential backoff (0.5s → 4.0s, ~7.5s total) is a reasonable budget — long enough for transient Ghostty init issues, short enough to not leave users hanging.
- The `surface(for:)` cache-miss path correctly discards broken views (`hasSurface == false`), preventing callers from getting permanently stuck with a nil-surface view.
- `DispatchQueue.main.asyncAfter` retry with `[weak self]` guard + `self.surface == nil` check is correct — prevents retrying after success or deallocation.

- `createAttempts` is reset to 0 both on success (line 93) and after exhausting retries (line 118). This means a view that fails and is then `destroy()`'d + re-created will correctly restart from attempt 0. Good.
- The `wireTerminalReadiness` callback guard `guard let currentState = self.appState.terminalReadiness[terminalID]` would skip the `.failed` state update if the terminal isn't tracked. However, the `.failed` case inside the switch is only reachable after the guard passes, and all managed terminals are tracked — so this is correct.

### Summary Table
| Priority | Issue |
|----------|-------|
| 🟢 | Retry backoff schedule is reasonable but not configurable — fine for now |
| 🟢 | The fixed 5s shell-readiness delay (existing code, not new) is still a TODO — not introduced by this PR |

**Recommendation:** Approve — well-scoped, clean implementation with correct resource management and no security concerns.

PR #222, opened 2026-04-29
Auto-label PRs with crow:auto on open

### LINKED ISSUE #221: Auto-assign reviewers and label on PR open
## Description

When a PR is opened in `radiusmethod/crow`, automatically:

1. **Assign reviewers:** `@dhilgaertner` and `@dgershman`
2. **Apply label:** `crow:auto`

## Implementation

Add a GitHub Actions workflow (e.g., `.github/workflows/pr-auto-assign.yml`) that triggers on `pull_request` opened events and:
- Requests review from both users
- Adds the `crow:auto` label

The `crow:auto` label should be created if it doesn't already exist.

dgershman:
## Summary

- Adds a GitHub Actions workflow that automatically applies the `crow:auto` label to every newly opened PR
- The label is created if it doesn't already exist
- Reviewer auto-assignment is already handled by the existing `.github/CODEOWNERS` file

Closes #221

### DIFF
--- a/.github/workflows/pr-auto-label.yml
+++ b/.github/workflows/pr-auto-label.yml
@@ -1,0 +1,24 @@
+name: Auto-label PRs
+
+on:
+  pull_request:
+    types: [opened]
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Ensure crow:auto label exists
+        run: gh label create "crow:auto" --repo "$GITHUB_REPOSITORY" --color "1D76DB" --force
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Apply label
+        run: gh pr edit "$PR" --repo "$GITHUB_REPOSITORY" --add-label "crow:auto"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PR: ${{ github.event.pull_request.number }}


dhilgaertner (approved):
## Code & Security Review

Small, focused workflow addition (24 lines, single file). Reviewed against GitHub Actions security best practices and the PR's stated intent.

### Security Review

**Strengths:**
- Uses `pull_request` (not `pull_request_target`) — avoids the well-known footgun where fork PRs get write-scoped tokens.
- `permissions:` block is least-privilege: `contents: read`, `pull-requests: write`. No unnecessary scopes.
- PR number is passed via env (`PR: ${{ github.event.pull_request.number }}`) and quoted as `"$PR"` — not interpolated directly into the shell, so no script-injection risk. (`pull_request.number` is an integer anyway, but defense-in-depth is the right pattern.)
- Uses the default `secrets.GITHUB_TOKEN` rather than a long-lived PAT.
- `$GITHUB_REPOSITORY` is a trusted built-in env var.

**Concerns:**
- None blocking.

### Code Quality

- Label creation is idempotent via `--force` — safe to re-run.
- Workflow scopes to `types: [opened]` only, which matches the stated intent (apply once on open). If someone removes the label manually, it won't be re-added — likely intentional, just worth being aware of.

### Notes / Minor (non-blocking)

- `gh label create --force` updates color/description if the label already exists. If anyone manually customizes the `crow:auto` label color/description in the GitHub UI, every subsequent PR open will reset it to `1D76DB`. Probably fine since this workflow is the source of truth for that label, but worth knowing.
- For PRs opened from forks, GitHub restricts `GITHUB_TOKEN` to read-only on the `pull_request` event regardless of the `permissions:` block, so the `Apply label` step would fail there. Not a concern unless this repo accepts fork PRs.
- Branch name and linked issue (#221) reference "auto-assign-pr-reviewers," but the PR body correctly clarifies reviewer assignment is handled by `.github/CODEOWNERS` — verified that file already lists `@dhilgaertner @dgershman`. No action needed; just flagging the naming mismatch in case it confuses future readers.

### Summary Table
| Priority | Issue |
|----------|-------|
| 🟢 | `--force` resets manual label customizations on every PR open |
| 🟢 | Won't apply label to fork PRs (GITHUB_TOKEN read-only on fork `pull_request` events) |
| 🟢 | Branch/issue naming mentions reviewer auto-assignment but only labeling is added (clarified in PR body) |

**Recommendation:** Approve. Clean, narrowly scoped, follows GitHub Actions security best practices.

PR #226, opened 2026-04-30
Per-section select all and icon-only cancel button

dgershman:
## Summary
- Move select-all from a global toggle to per-section buttons in each section header (Active, Reviews, In Review, Completed)
- Replace the text "Cancel" button with an `xmark` icon in the bulk action bar
- Fix `build-ghostty.sh` to select the correct `libghostty_zcu.o` containing API symbols (not the SIMD-only variant)
- Fix `build-ghostty.sh` to copy `module.modulemap` to the library identifier level for SPM 6.3+ compatibility

## Test plan
- [ ] Enter selection mode and verify each section header shows a select-all icon
- [ ] Tap section select-all, confirm all sessions in that section are selected and icon changes to filled checkmark
- [ ] Tap again to deselect all in section
- [ ] Verify cancel button is an xmark icon, not text
- [ ] Run `make ghostty && make` to verify build fixes

### DIFF
--- a/Packages/CrowUI/Sources/CrowUI/SessionListView.swift
+++ b/Packages/CrowUI/Sources/CrowUI/SessionListView.swift
@@ -213,6 +213,7 @@ public struct SessionListView: View {
             }
             .buttonStyle(.plain)
             .help("Cancel selection")
+            .accessibilityLabel("Cancel selection")
 
             if !selectedSessionIDs.isEmpty {
                 Button {
@@ -231,6 +232,7 @@ public struct SessionListView: View {
                     .clipShape(RoundedRectangle(cornerRadius: 6))
                 }
                 .buttonStyle(.plain)
+                .accessibilityLabel("Delete \(selectedSessionIDs.count) selected sessions")
             }
         }
         .padding(.horizontal, 12)
@@ -334,6 +336,7 @@ struct SectionDivider: View {
                 }
                 .buttonStyle(.plain)
                 .help(allSelected ? "Deselect all \(title.lowercased())" : "Select all \(title.lowercased())")
+                .accessibilityLabel(allSelected ? "Deselect all \(title.lowercased())" : "Select all \(title.lowercased())")
             }
         }
         .padding(.top, 10)
--- a/Packages/CrowUI/Sources/CrowUI/SessionListView.swift
+++ b/Packages/CrowUI/Sources/CrowUI/SessionListView.swift
@@ -18,7 +18,7 @@ public struct SessionListView: View {
         VStack(spacing: 0) {
             sessionList
 
-            if isSelectionMode && !selectedSessionIDs.isEmpty {
+            if isSelectionMode {
                 bulkActionBar
             }
         }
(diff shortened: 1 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

Small, focused PR — two unrelated but well-scoped changes (UI tweak + Ghostty build fix). Reviewed both files end to end. Logic is correct and the build-ghostty.sh fix follows the same `nm | grep -q` pattern already used elsewhere in the script for `ext.o`, `wuffs-v0.4.o`, and `vt.o`, so it's consistent.

### Security Review
**Strengths:**
- No external/untrusted input touched. SwiftUI changes are pure view-state; `selectedSessionIDs` mutations stay in-process.
- Shell script changes only read from a local Zig cache. The new `find -exec sh -c '... nm "$1" ... grep -q "T _ghostty_app_new" ...' _ {} \;` correctly quotes `$1` (no word-splitting risk on cache paths).

**Concerns:** None.

### Code Quality

**Accessibility regression on the new icon-only buttons** (yellow). The file's existing convention is to pair `.help` with `.accessibilityLabel` on every icon-only `Button` (see `Packages/CrowUI/Sources/CrowUI/SessionListView.swift:37-38` for the speaker button and `:193-194` for the select-toggle). The three new buttons added in this PR have `.help` but no `.accessibilityLabel`:
- `:215` — Cancel xmark button. VoiceOver will announce a generic "button" instead of "Cancel selection".
- `:336` — Per-section select-all/deselect-all. Same issue.
- `:217-233` — Bulk delete now reads as just the count ("(3)") since the descriptive `Text("Delete (\(count))")` was replaced with `Text("(\(count))")`. Without an `.accessibilityLabel`, the trash icon's purpose is invisible to VoiceOver. Suggest `.accessibilityLabel("Delete \(selectedSessionIDs.count) selected sessions")`.

**Minor observations** (green):
- `SectionDivider.init` provides a default `Binding<Set<UUID>> = .constant([])` for `selectedSessionIDs`, but every call site now passes a real binding. The default is harmless backwards-compat scaffolding; you could drop it to keep the API minimal, but it's fine either way.
- `Set(filteredSessions(...).map(\.id))` is rebuilt on every parent body recomputation. Fine at the expected scale of dozens of sessions; flagging only so future changes are aware.
- The empty-section / search-filtered behavior is correct: `allSelected` guards on `!sectionIDs.isEmpty`, and `sectionIDs` reflects only the visible/filtered rows, so "select all" while searching only adds the visible items — good UX.

**Bulk action bar visibility change is intentional and correct.** Switching the outer guard from `isSelectionMode && !selectedSessionIDs.isEmpty` to just `isSelectionMode` (with the inner Delete button gated by `!selectedSessionIDs.isEmpty`) is the right move — the bar now shows the live count + cancel even before anything is selected, instead of popping in/out as the count crosses zero.

**build-ghostty.sh:**
- `find ... -exec ... grep -q "T _ghostty_app_new" ... | head -1` is robust. With `set -euo pipefail`, the `head -1` short-circuit can SIGPIPE upstream `find`, but since this is wrapped in command substitution the exit status is discarded — fine.
- `module.modulemap` copy at `:79-81` is idempotent (`! -f "$XCFW_DIR/module.modulemap"` guard) and only fires when a Headers-level modulemap exists. Good.

### Summary Table
| Priority | Issue |
|----------|-------|
| 🟡 Yellow | Add `.accessibilityLabel` to the three new icon-only buttons (cancel, section select-all, bulk-delete trash) to match the file's existing pattern |
| 🟢 Green | Consider dropping the now-unused default `Binding` parameter on `SectionDivider.init` |

**Recommendation:** Approve. No blocking or security issues; the accessibility nits are easy follow-ups and not worth holding the PR.

PR #234, opened 2026-05-02
Silently skip hook-event when Crow app is not running

### LINKED ISSUE #227: Stop hook fails with connection refused when Crow app is not running
## Summary

The stop hook fails with a socket connection error when the Crow app is not running, causing noisy errors on Claude Code session exit.

## Error

```
Stop hook error: Failed with non-blocking status code: Error: Socket connection failed: Connection refused
```

Full hook output:
```
/Users/danny/Projects/devroot/rm/crow-225-select-all-delete-icon/.build/debug/crow hook-event --session 5D09D025-4C37-4F7A-BF7D-395C620E5D9C --event Stop
```

## Expected Behavior

The stop hook should gracefully handle the case where the Crow app is not running — either silently exit 0 or log a non-error warning. Hooks should not fail loudly when the socket is unavailable.

## Reproduction

1. Quit the Crow app
2. Exit a Claude Code session that has a Crow stop hook configured
3. Observe the connection refused error

dhilgaertner:
## Summary

- `crow hook-event` now silently exits 0 when the Crow app isn't running, instead of failing with "Socket connection failed: Connection refused".
- Catch is scoped to `SocketError.connectionFailed` inside `HookEventCmd` only — other CLI commands (`new-session`, `set-status`, etc.) still fail loudly when the socket is unavailable, since those are user-initiated.
- Other socket errors (timeout, write/read failures) and JSON-RPC errors continue to propagate so genuine misbehavior stays visible.

Closes #227

## Test plan

- [x] `make build` succeeds.
- [x] `swift test --package-path Packages/CrowCLI` — 35 tests pass, including new `forwardHookEventSilentWhenAppNotRunning`.
- [x] Repro: `echo '{}' | CROW_SOCKET=/tmp/missing.sock .build/debug/crow hook-event --session $(uuidgen) --event Stop` → no output, exit 0.
- [x] Sanity: `CROW_SOCKET=/tmp/missing.sock .build/debug/crow list-sessions` still prints "Socket connection failed: …" and exits 1.
- [x] Happy path with live app: `echo '{...}' | .build/debug/crow hook-event --session <real-uuid> --event Stop` → exit 0, event reaches the app.

### DIFF
--- a/Packages/CrowCLI/Sources/CrowCLILib/Commands/HookEventCommand.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Commands/HookEventCommand.swift
@@ -24,10 +24,23 @@ public struct HookEventCmd: ParsableCommand {
 
     public func run() throws {
         let payload = parseHookPayload(from: FileHandle.standardInput.readDataToEndOfFile())
+        try forwardHookEvent(sessionID: session, eventName: event, payload: payload)
+    }
+}
 
+/// Forward a parsed hook event over the Unix socket.
+///
+/// Silently no-ops when the Crow app is not running (socket connection
+/// refused or socket file absent). Hooks are fire-and-forget — a missing
+/// listener is an expected state, not an error, so we must not exit
+/// non-zero or write to stderr (it pollutes Claude Code's hook output).
+/// Other socket errors (timeout, write/read failures) and JSON-RPC
+/// errors still propagate so genuine misbehavior is visible.
+func forwardHookEvent(sessionID: String, eventName: String, payload: [String: JSONValue]) throws {
+    do {
         let result = try rpc("hook-event", params: [
-            "session_id": .string(session),
-            "event_name": .string(event),
+            "session_id": .string(sessionID),
+            "event_name": .string(eventName),
             "payload": .object(payload),
         ])
 
@@ -35,6 +48,8 @@ public struct HookEventCmd: ParsableCommand {
         if result["error"] != nil {
             printJSON(result)
         }
+    } catch SocketError.connectionFailed {
+        return
     }
 }
 
--- a/Packages/CrowCLI/Tests/CrowCLITests/HookEventTests.swift
+++ b/Packages/CrowCLI/Tests/CrowCLITests/HookEventTests.swift
@@ -48,3 +48,20 @@ import CrowIPC
     let payload = parseHookPayload(from: data)
     #expect(payload.isEmpty)
 }
+
+// MARK: - Connection Refused Handling
+
+/// Regression test for #227: when the Crow app is not running, the hook-event
+/// command must silently no-op instead of exiting non-zero. Otherwise Claude
+/// Code surfaces "Stop hook error: …" on every session exit.
+@Test func forwardHookEventSilentWhenAppNotRunning() throws {
+    let nonExistent = NSTemporaryDirectory() + "crow-test-\(UUID().uuidString).sock"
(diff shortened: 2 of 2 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Error suppression is scoped strictly to `SocketError.connectionFailed` within `forwardHookEvent` only — other CLI commands retain their fail-loud behavior for user-initiated operations.
- No new stdin/stdout paths that could leak data; hook payload parsing already existed and the security posture is unchanged.
- The `CROW_SOCKET` env-var override is existing behavior (legacy support) — not introduced here.

**Concerns:**
- None identified. The fire-and-forget semantics are appropriate for hook events where the app being absent is a normal condition, not a security boundary.

### Code Quality
- Clean extraction: the `forwardHookEvent` function has a single responsibility and is independently testable.
- The `catch SocketError.connectionFailed` pattern correctly matches all associated errno values (both `ECONNREFUSED` and `ENOENT`), covering both "app not running" and "socket file absent" scenarios.
- Documentation accurately describes the contract (which errors propagate vs. which are swallowed).
- Test uses `setenv`/`unsetenv` which modifies process-global state — acceptable here given test speed, but worth noting for future test growth. A `defer { unsetenv(...) }` correctly restores state.
- All 6 HookEvent tests pass; build is clean with no warnings in the changed files.

### Summary Table
| Priority | Issue |
|----------|-------|
| Green | Consider: if concurrent Swift Testing adoption grows, the `setenv`-based test could benefit from process isolation, but not blocking today |

**Recommendation:** Approve — well-scoped fix for #227 with appropriate error handling granularity and good test coverage.

dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Error handling is scoped narrowly to `SocketError.connectionFailed` inside `forwardHookEvent` only — other CLI commands continue to fail loudly when the app is unavailable, preserving the correct behavior for user-initiated operations.
- No new attack surface: no new inputs, no new IPC methods, no privilege changes.
- The `CROW_SOCKET` env var override in the test is properly cleaned up via `defer { unsetenv(...) }`.

**Concerns:**
- None identified. The change correctly distinguishes between expected (app not running → connection failed) and unexpected (timeouts, write/read failures, JSON-RPC errors) socket conditions.

### Code Quality
- **Clean extraction:** The RPC logic was pulled out of `HookEventCmd.run()` into a standalone `forwardHookEvent()` function, making it independently testable — good design.
- **Catch pattern correctness:** `catch SocketError.connectionFailed` matches all `connectionFailed(Int32)` variants regardless of errno, which is the right behavior — both ECONNREFUSED (app not listening) and ENOENT (socket file absent) should be silently swallowed.
- **Test coverage:** `forwardHookEventSilentWhenAppNotRunning` is a clean regression test — points at a non-existent socket via `CROW_SOCKET`, verifies the function completes without throwing.
- **Doc comment on `forwardHookEvent`** clearly explains the rationale and the error taxonomy (which errors propagate vs. which are swallowed).

### Summary Table
| Priority | Issue |
|----------|-------|
| ✅ Green | Consider whether `SocketError.createFailed` should also be caught — it would fire if the OS can't allocate a socket fd, which is unlikely but also not an app-connectivity issue. Current behavior (letting it propagate) is defensible since it signals a system-level problem. |

**Recommendation:** Approve — this is a well-scoped, well-tested fix for a real usability issue (#227). The error handling taxonomy is correct and the blast radius is minimal.

PR #233, opened 2026-05-01
Fix glab fetch failures from non-repo cwd and nested-group slug truncation

dhilgaertner:
## Summary

- `fetchGitLabIssues` was running `glab issue list -a @me --output-format json` with `cwd: NSHomeDirectory()`. `glab issue list` shells out to `git` even when no repo is involved and aborts with *"fatal: not a git repository"* when the cwd is not a git working tree. Swap to `glab api "issues?scope=assigned_to_me&state=opened&per_page=100"` — same JSON fields the existing parser already reads (`iid`, `title`, `web_url`, `state`, `labels`, `references.full`), no git-repo dependency. Same pattern Crow already uses in `fetchGitLabMRsForReconcile`.
- `resolveRepoInfo` truncated nested-group GitLab slugs because its regex `[:/]([^/:]+/[^/:]+)$` only captured the last two path segments. `big-bang/product/packages/elasticsearch-kibana` collapsed to `packages/elasticsearch-kibana`, then the reconcile path hit a 404 against a non-existent project. Replaced with a new `IssueTracker.extractSlug(fromRemote:)` helper (parallel to `extractHost`) that returns the full path after the host for both SSH and HTTPS remotes and strips a trailing `.git`.
- Added a `IssueTrackerRemoteSlugTests` Swift Testing suite mirroring the existing `IssueTrackerRemoteHostTests`: GitHub SSH/HTTPS, two-level GitLab, nested-group GitLab (HTTPS and SSH — the regression for #232), `.git` stripping, and unrecognized inputs.

Closes #232

## Test plan

- [x] `swift test` — all 38 tests in 5 suites pass, including the new `IssueTracker remote slug extraction` suite (7 tests).
- [ ] Manual: with no GitLab repo cloned at `~`, confirm Crow's GitLab issue tracker now returns assigned issues instead of an empty list.
- [ ] Manual: register a worktree whose origin points at a nested GitLab group (e.g. `big-bang/product/packages/elasticsearch-kibana`) and confirm the merged-MR reconcile flow finds the project (no 404).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #242, opened 2026-05-04
Route stale-PR follow-up by provider so merged GitLab MRs poll correctly

### LINKED ISSUE #241: Merged PRs not turning purple on periodic GitHub polling (only on app open); IssueTracker error for 'big-bang/product'
## Symptom

Merged PRs no longer transition to the purple ("merged") state during the periodic polling of GitHub issues. They only update on application open.

## Possibly related log error

This error appears in the logs and may or may not be related:

```
[IssueTracker] Stale-PR follow-up failed (exit 1): gh: Could not resolve to a Repository with the name 'big-bang/product'.
```

`big-bang/product` is a GitLab repo in the PlatformOne workspace (host `repo1.dso.mil`), so invoking `gh` for it would always fail. If the periodic stale-PR follow-up bails on the first error per cycle, that could explain why none of the other merged PRs get re-checked between app launches.

## Asks

1. Diagnose why the periodic poll no longer flips merged PRs to purple (regression vs. open-time path).
2. Diagnose the `gh` invocation against a GitLab repo — should be routed through `glab` (with `GITLAB_HOST=repo1.dso.mil`) instead.
3. Confirm whether the two are the same root cause or independent.

dhilgaertner:
## Summary

- `IssueTracker.fetchStalePRStates` hardcoded `gh api graphql` for every stale PR URL. One GitLab MR in the batch (e.g. `big-bang/product` on `repo1.dso.mil`) failed the `gh` call, returned `nil`, and wiped the whole cycle's merged-state update — so merged PRs (GitHub or GitLab) only turned purple on app launch.
- Split URLs by provider: GitHub PRs → existing batched aliased GraphQL via `gh`; GitLab MRs → one `glab api projects/{slug}/merge_requests/{iid}` per MR with `GITLAB_HOST` set per host (mirroring the pattern in `fetchGitLabMRsForReconcile`).
- Replaced the `[ViewerPR]?` return with a `StalePRFetchResult { prs, complete }` struct so a single-provider failure no longer masks the other provider's PRs — partial success contributes its PRs to `applyPRStatuses` while `prDataComplete=false` still blocks auto-completion on degraded data.
- Added a robust GitLab MR URL parser (`parseGitLabMRURL`) that handles nested groups, plus a shared `normalizeGitLabPRState` helper used by both stale-fetch and reconcile.

Closes #241

## Test plan

- [x] `make build` — clean
- [x] `swift test --filter IssueTrackerStalePRTests` — 17 new tests pass (URL parse, state normalization, JSON parse, mixed-provider dedup)
- [x] `swift test` — all IssueTracker suites green; one pre-existing flaky `tmuxBackendDefaultIsOff` race unrelated to this change
- [ ] Manual: with both a merged GitHub PR and a merged GitLab MR linked to active sessions, wait one poll cycle (≤60s) without restarting the app — both badges flip to purple and no `[IssueTracker] Stale-PR follow-up failed (exit 1)` log lines appear

### DIFF
--- a/Sources/Crow/App/IssueTracker.swift
+++ b/Sources/Crow/App/IssueTracker.swift
@@ -274,16 +274,17 @@ final class IssueTracker {
             // for every viewer (which routinely returned 100 PRs / ~86 KB).
             let openPRURLs = Set(ghResult.viewerPRs.map(\.url))
             let staleCandidateURLs = collectStalePRURLs(excluding: openPRURLs)
-            // nil here means the follow-up fetch errored (rate limit, exit != 0,
-            // parse failure). We thread that through to auto-complete so
-            // "PR missing from payload" doesn't get treated as "PR is closed"
-            // on a degraded response. An empty list with no candidate URLs is
-            // a real empty success.
-            let staleFetchResult: [ViewerPR]? = staleCandidateURLs.isEmpty
-                ? []
+            // `complete == false` means at least one provider's follow-up errored
+            // (rate limit, exit != 0, parse failure). We thread that through to
+            // auto-complete so "PR missing from payload" doesn't get treated as
+            // "PR is closed" on a degraded response. Partial-success is allowed:
+            // PRs from the working provider still flow through so merged badges
+            // can flip even if the other provider failed.
+            let staleFetch = staleCandidateURLs.isEmpty
+                ? StalePRFetchResult(prs: [], complete: true)
                 : await fetchStalePRStates(urls: staleCandidateURLs)
-            let stalePRs = staleFetchResult ?? []
-            let prDataComplete = staleFetchResult != nil
+            let stalePRs = staleFetch.prs
+            let prDataComplete = staleFetch.complete
             let allKnownPRs = Self.dedupedByURL(ghResult.viewerPRs + stalePRs)
 
             applyPRStatuses(viewerPRs: allKnownPRs)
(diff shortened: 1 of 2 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None found.

### Security Review
**Strengths:**
- URL-encoded slug passed to the GitLab API endpoint (`addingPercentEncoding(withAllowedCharacters: .alphanumerics)`) prevents path traversal via malicious slug strings.
- Error output is truncated (`.prefix(200)`) in log messages, preventing large stderr from flooding logs.
- `GITLAB_HOST` is set per-host in the process environment, mirroring the existing `fetchGitLabMRsForReconcile` pattern — no credential leak across hosts.
- `NSHomeDirectory()` as the cwd avoids the non-repo-cwd failure that #233 fixed.

**Concerns:**
- None. The attack surface is limited to locally-stored session URLs that originate from the user's own linked PRs.

### Code Quality

**Well done:**
- The `StalePRFetchResult` struct cleanly replaces the `[ViewerPR]?` optional with explicit partial-success semantics — this is a meaningful improvement in clarity over nil-means-failure.
- Extracting `normalizeGitLabPRState` and reusing it in `fetchGitLabMRsForReconcile` eliminates the inline switch duplication.
- `parseGitLabMRURL` handles nested groups correctly (the two-segment heuristic in `parseTicketURLComponents` would truncate), and the test suite explicitly guards this regression.
- The host guard at line 699 (`host != "github.com"`) is a sensible safety net — unknown hosts that don't match the `/-/merge_requests/` pattern won't get blindly routed to `gh`.
- Test coverage is comprehensive: URL parsing (top-level, nested, rejection cases), state normalization, JSON parsing (merged/opened/closed/malformed/legacy WIP flag), and dedup integration.

- The `for (host, parsed) in gitlabParsedByHost` loop at line 721 runs sequentially. For the stated assumption ("stale set is usually tiny") this is fine, but if it ever grows, these could be parallelized with a TaskGroup since they're independent per-host.
- `parseGitLabStaleMRResponse` returns `nil` (marking the cycle incomplete) on unparseable JSON, but this could also happen if the MR was deleted (404 returning an error body). The error-vs-parse-failure distinction is already handled by the try/catch in the caller, so this is fine in practice.

### Summary Table
| Priority | Issue |
|----------|-------|
| Green | Sequential per-host GitLab fetches are fine for current scale; consider TaskGroup if stale sets grow |
| Green | 404 vs. malformed JSON both result in `ok = false` — correct behavior, just worth noting for future debugging |

**Recommendation:** Approve. The change is well-structured, correctly isolates provider failures so one bad GitLab host can't suppress GitHub PR updates, and the test suite thoroughly covers the new parsing/normalization logic. The architecture matches existing patterns in the codebase (reconcile path) and the PR description accurately reflects what was done.

PR #245, opened 2026-05-06
Allow ignoring PRs with specific labels from review

### LINKED ISSUE #244: Allow ignoring PRs with specific labels from review
## Summary

Add the ability to configure a list of labels that should be excluded from PR review. PRs with any of these labels would be automatically skipped by Crow's review process.

## Motivation

Automated PRs (e.g., Renovate dependency updates) currently get queued for review alongside human-authored PRs. These PRs are typically auto-merged and don't need the same level of review, creating unnecessary noise.

## Proposed Behavior

- Add a configuration option (e.g., `ignoreLabels`) that accepts a list of GitHub/GitLab label names
- When Crow fetches PRs for review, skip any PR that has one or more of the configured ignore labels
- Common use case: ignoring labels like `dependencies`, `renovate`, `automated`, etc.

## Example Configuration

```
ignoreLabels:
  - dependencies
  - renovate
  - automated
```

dgershman:
Closes #244

## Summary

- Adds `ignoreReviewLabels` configuration option (comma-separated label names) to filter PRs from the review board
- Fetches PR labels via the GitHub GraphQL query (`labels(first: 20) { nodes { name } }`)
- Applies case-insensitive label matching alongside the existing repo-exclude filter
- PRs with any matching label are hidden from the review board, badge counts, and notifications
- Settings UI in Automation → Reviews mirrors the existing "Excluded Repos" pattern

## Test plan

- [x] CrowCore package builds cleanly
- [x] All 143 unit tests pass (including 2 new config round-trip tests)
- [ ] Manual: configure `ignoreReviewLabels: ["dependencies"]` and verify matching PRs are filtered from the review board
- [ ] Manual: verify existing PRs without matching labels still appear

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -129,10 +129,20 @@ public final class AppState {
 
     public var excludeReviewRepos: [String] = []
     public var excludeTicketRepos: [String] = []
+    public var ignoreReviewLabels: [String] = []
 
     public var filteredReviewRequests: [ReviewRequest] {
-        guard !excludeReviewRepos.isEmpty else { return reviewRequests }
-        return reviewRequests.filter { !repoMatchesExcludePatterns($0.repo, patterns: excludeReviewRepos) }
+        var result = reviewRequests
+        if !excludeReviewRepos.isEmpty {
+            result = result.filter { !repoMatchesExcludePatterns($0.repo, patterns: excludeReviewRepos) }
+        }
+        if !ignoreReviewLabels.isEmpty {
+            let lowerLabels = Set(ignoreReviewLabels.map { $0.lowercased() })
+            result = result.filter { request in
+                !request.labels.contains(where: { lowerLabels.contains($0.lowercased()) })
+            }
+        }
+        return result
     }
 
     /// IDs of review requests the user has already seen (for badge count).
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -176,6 +176,7 @@ public struct ConfigDefaults: Codable, Sendable, Equatable {
     public var excludeDirs: [String]
     public var excludeReviewRepos: [String]
     public var excludeTicketRepos: [String]
+    public var ignoreReviewLabels: [String]
 
     /// Characters that are invalid in git ref names (see `git check-ref-format`).
     private static let invalidBranchChars = CharacterSet(charactersIn: " ~^:?*[\\")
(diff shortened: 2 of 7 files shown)


dhilgaertner (approved):
## Code & Security Review

### Security Review
**Strengths:**
- No new attack surface. The GraphQL change just adds `labels(first: 20) { nodes { name } }` to an existing hardcoded query — values flow through `gh api graphql -F` parameters, not string interpolation, so no injection vector.
- Label parsing in `IssueTracker.swift:1106-1107` is type-safe (`compactMap { $0["name"] as? String }`) with a sensible empty-array fallback.
- All filtering is local; the user-configured label list is never sent to a remote system.

**Concerns:**
- None.

### Code Quality
**Strengths:**
- Backwards-compatible config decoding — missing `ignoreReviewLabels` key falls back to `[]`, covered by `ignoreReviewLabelsDefaultsEmptyWhenKeyMissing` test (`AppConfigTests.swift:261-265`).
- Filter pattern mirrors the existing `excludeReviewRepos` flow exactly: applied in `IssueTracker.swift:307-313` before storing into `appState.reviewRequests`, and again in `AppState.filteredReviewRequests` for defense in depth. Consistency is good.
- Notification dedup correctness preserved: `previousReviewRequestIDs = allCurrentIDs` is set from the *unfiltered* ID set, so a PR transitioning between hidden/visible (label added or removed) doesn't spuriously fire as a "new review request" toast (`IssueTracker.swift:302,316`).
- Case-insensitive label match aligns with the repo-pattern matching convention.
- 17 lines of new tests including round-trip and missing-key cases.

**Considerations (non-blocking):**
- `labels(first: 20)` in `IssueTracker.swift:572` — a PR with more than 20 labels would silently truncate beyond the cap; if a configured ignore-label is the 21st on a noisy PR it won't be filtered. Almost certainly fine in practice; just worth being aware of.
- Labels are now fetched on every review-PR refresh even when `ignoreReviewLabels` is empty — payload-size cost is trivial, but notable.
- The settings `TextField` is initialized from `defaults.wrappedValue.ignoreReviewLabels` once at init time (`AutomationSettingsView.swift:31`); external mutations to defaults won't update the visible text. This matches the pre-existing pattern for `excludeReviewReposText`, so not a regression.
- `ReviewRequest.labels` is only populated on the GitHub path (`IssueTracker.swift:1109-1122`); there's no GitLab MR equivalent, but GitLab review requests aren't currently surfaced through `reviewRequests` anyway, so nothing breaks.

### Build & Tests
- `swift build` against `Packages/CrowCore` succeeds clean.
- `swift test` passes all 143 tests in 5 suites (including the 2 new ones).
- The umbrella `swift build` from repo root fails on a missing `Frameworks/GhosttyKit.xcframework` binary artifact — unrelated to this PR's changes.

### Summary Table
| Priority | Issue |
|----------|-------|
| 🟢 | Consider surfacing why a PR is hidden in the review board UI (e.g., a "filtered by label" affordance), so users don't wonder where their PRs went. |
| 🟢 | Consider widening `labels(first: 20)` or paginating if any tracked repos use heavy label taxonomies. |

**Recommendation:** Approve. Tightly scoped, follows existing patterns, backwards-compatible, well tested. No security or correctness concerns.

PR #249, opened 2026-05-07
Add Crow-Session attribution trailer to commits

### LINKED ISSUE #246: design: add crow attribution to Claude Code commits
## Context

When Claude Code makes a commit, it appends a co-author trailer:

```
```

This happens because Claude Code's built-in instructions tell it to add that trailer. The result is genuinely useful: a reader of `git log` / GitHub PR pages / `git blame` can see at a glance that the commit was AI-assisted, and tooling can grep for it.

When the Claude Code session was launched by **crow**, that fact is currently invisible in the commit history. A reader can see "Claude helped write this" but not "crow orchestrated the session." There's an opportunity to make crow's involvement visible the same way Claude's is — both as a small bit of vanity ("crow is what got me to this commit") and, more usefully, as a traceability hook back to crow session metadata.

## Goal

When a crow-orchestrated session commits, append a second trailer alongside Claude's. Shape options to discuss:

```
Crow-Session: 9B753450-3A6A-4DED-BFE3-A681480B8FA3
```

or vanity-only:

```
Orchestrated-By: Crow <https://github.com/radiusmethod/crow>
```

or a tagline plus the existing trailer:

```
🤖 Generated with Claude Code, orchestrated by Crow
```

### 1. Trailer format

Probably structured (`Crow-Session: <uuid>`) wins over free-text. Git has native trailer support (`git interpret-trailers`, `--trailer` flag, `%(trailers)` log format) — structured trailers are greppable, parseable, and play well with existing tooling. Free-text taglines are throwaway.

### 2. What to include in the trailer value

Candidates:

- Just \"this commit came from crow\" — pure vanity, no value beyond the existing `Co-Authored-By`
- **Session UUID** — lets future archeologists match a commit to \"what task was Claude working on at the time\" via `crow get-session <uuid>`
- Workspace name
- Crow version

Recommendation: **session UUID**. It's the most useful — a stable handle back to session metadata. Privacy concern is essentially zero (UUIDs are locally-meaningful only). Workspace and version can be looked up via the session if needed.

### 3. How to inject the trailer

Three real approaches, ordered by complexity:

**A. Bootstrap-prompt instruction.** crow already writes a per-session prompt to `.claude/prompts/crow-prompt-{session_name}.md`. Add one line: \"When committing, include a `Crow-Session: {session_id}` trailer.\" Cheapest. Relies on Claude following instructions — the same mechanism Claude already uses for its own trailer, so the trust model is identical.

**B. Worktree-local CLAUDE.md snippet.** crow could write a small `CLAUDE.md` (or merge into the existing one) inside the worktree with the trailer instruction. Survives prompt compaction; gets re-loaded each turn. Belt-and-suspenders alongside A.

**C. Git `commit-msg` hook.** Install a hook in the worktree that appends the trailer to every commit. Most deterministic — works for human commits too, not just Claude. But touches `.git/hooks` (a shared resource between worktrees), can collide with user hooks, doesn't work for users with `core.hooksPath` set, and is generally surprising tooling behavior.

**D. Claude Code settings.json PostToolUse hook.** Intercept `git commit` and rewrite the message. Most invasive; probably not worth it.

**Recommendation: A + B together.** A is cheap and gets 90% of the way. B is the safety net for long-running sessions where the bootstrap prompt has scrolled out of context. C is interesting but hook ergonomics don't justify it for a single trailer line.

### 4. Opt-in vs default

Default-on (matches Claude's own trailer) with a `crow.attributionTrailers: false` config flag for users who'd rather not. People with strict commit-message conventions or open-source repos with explicit \"no AI attribution\" rules will want to disable it.

### 5. Composability with Claude's own trailer

Make sure the trailers stack cleanly:

```
fix(ui): center the modal

Crow-Session: 9B753450-3A6A-4DED-BFE3-A681480B8FA3
```

Order doesn't matter for git's trailer parser. Both appear; neither replaces the other.

## Concrete proposal (Approach A + B)

1. In `setup.sh`, when writing the bootstrap prompt, append a section:

   ```
   ## Commit attribution

   When you make commits in this workspace, include a `Crow-Session: {session_id}` trailer in the commit message footer alongside your existing `Co-Authored-By: Claude` trailer:

       Crow-Session: 9B753450-3A6A-4DED-BFE3-A681480B8FA3
   ```

2. Also write/merge into a worktree-local `.claude/CLAUDE.md` with the same instruction, so it survives prompt compaction.

3. Add a `crow.attributionTrailers` boolean to `.claude/config.json` (default `true`) so users can opt out without editing code.

4. (Optional, future) A `crow log` subcommand that takes a commit SHA, parses the `Crow-Session:` trailer, and prints the matching session metadata. Nice add but not required for this ticket.

## Acceptance criteria

- [ ] Bootstrap prompt template includes the commit-attribution instruction with `{session_id}` interpolated.
- [ ] A `CLAUDE.md` snippet inside each new worktree carries the same instruction so it persists past prompt compaction.
- [ ] Commits made in a crow-orchestrated session contain both `Co-Authored-By: Claude` and `Crow-Session: {uuid}` trailers.
- [ ] Config flag `crow.attributionTrailers: false` disables the instruction (no trailer added).
- [ ] Commits made outside a crow session in the same repo are unaffected.

## Out of scope

- Wrapping or intercepting `git commit` itself. We rely on Claude following instructions, same as its own trailer.
- Post-commit verification / enforcement. If Claude omits the trailer occasionally, no big deal.
- A `crow log` archeology command that searches `git log` for `Crow-Session:` and links back to session metadata. Genuinely useful future feature but separate ticket.
- Backfilling trailers on past commits.

## Effort

~30 min: prompt-template change + CLAUDE.md snippet + config flag + smoke test that the trailer actually lands in a real commit.

dhilgaertner:
## Summary
- `setup.sh` writes a per-worktree `.claude/settings.local.json` that overrides Claude Code's native `attribution.commit` to include a `Crow-Session: <session-uuid>` trailer alongside the standard `Co-Authored-By: Claude` line. The session UUID is a stable handle back to session metadata via `crow get-session <uuid>`.
- New top-level `attributionTrailers` flag on `AppConfig` (default `true`) toggles the behavior. Exposed in Settings → Automation → Attribution.
- `setup.sh` also appends `.claude/settings.local.json` to the per-worktree git exclude file so the override never gets committed accidentally, even when the repo's tracked `.gitignore` doesn't already cover it.

The implementation switched from the originally-proposed CLAUDE.md / prompt-injection approach to Claude Code's native `attribution.commit` setting after research — it's deterministic (no relying on the LLM to remember an instruction every commit) and survives prompt compaction by virtue of being a config file Claude Code re-reads.

The trailer body chosen:

```
🤖 Generated with Claude Code, orchestrated by Crow

Crow-Session: <UUID>
```

## Test plan
- [x] `make app` builds cleanly
- [x] `swift test --package-path Packages/CrowCore` (145 tests, including 2 new ones)
- [x] `swift test --package-path Packages/CrowPersistence` (24 tests)
- [x] Smoke tested `write_settings_local` against a scratch git worktree:
  - default-on (no config) writes the file with the real UUID interpolated
  - `"attributionTrailers": false` skips the file
  - whitespace-tolerant config parsing (`"attributionTrailers" :  false`)
  - per-worktree exclude entry added; `git status` reports the file as untracked-and-excluded
  - generated JSON parses cleanly via `python3 -c 'json.load(...)'` and decodes to the expected three-line trailer body
- [ ] Manual end-to-end check on a real workspace: launch a crow session, ask Claude to commit, verify `git log -1 --format='%B'` shows both `Co-Authored-By: Claude` and `Crow-Session: <real-uuid>` lines
- [ ] Settings UI smoke test: Settings → Automation → Attribution toggle persists to `{devRoot}/.claude/config.json`

Closes #246

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -17,6 +17,10 @@ public struct AppConfig: Codable, Sendable, Equatable {
     /// Opt into the tmux backend (#198). Surfaced in Settings → Experimental.
     /// Read once at app launch; takes effect on next relaunch.
     public var experimentalTmuxBackend: Bool
+    /// When true, `setup.sh` writes a per-worktree `.claude/settings.local.json`
+    /// that overrides Claude Code's `attribution.commit` to include the crow
+    /// session UUID alongside the standard `Co-Authored-By: Claude` trailer.
+    public var attributionTrailers: Bool
 
     public init(
         workspaces: [WorkspaceInfo] = [],
@@ -27,7 +31,8 @@ public struct AppConfig: Codable, Sendable, Equatable {
         managerAutoPermissionMode: Bool = true,
         telemetry: TelemetryConfig = TelemetryConfig(),
         autoRespond: AutoRespondSettings = AutoRespondSettings(),
-        experimentalTmuxBackend: Bool = false
+        experimentalTmuxBackend: Bool = false,
+        attributionTrailers: Bool = true
     ) {
         self.workspaces = workspaces
         self.defaults = defaults
@@ -38,6 +43,7 @@ public struct AppConfig: Codable, Sendable, Equatable {
         self.telemetry = telemetry
         self.autoRespond = autoRespond
         self.experimentalTmuxBackend = experimentalTmuxBackend
+        self.attributionTrailers = attributionTrailers
     }
 
     public init(from decoder: Decoder) throws {
@@ -51,10 +57,11 @@ public struct AppConfig: Codable, Sendable, Equatable {
         telemetry = try container.decodeIfPresent(TelemetryConfig.self, forKey: .telemetry) ?? TelemetryConfig()
         autoRespond = try container.decodeIfPresent(AutoRespondSettings.self, forKey: .autoRespond) ?? AutoRespondSettings()
         experimentalTmuxBackend = try container.decodeIfPresent(Bool.self, forKey: .experimentalTmuxBackend) ?? false
+        attributionTrailers = try container.decodeIfPresent(Bool.self, forKey: .attributionTrailers) ?? true
     }
 
     private enum CodingKeys: String, CodingKey {
-        case workspaces, defaults, notifications, sidebar, remoteControlEnabled, managerAutoPermissionMode, telemetry, autoRespond, experimentalTmuxBackend
+        case workspaces, defaults, notifications, sidebar, remoteControlEnabled, managerAutoPermissionMode, telemetry, autoRespond, experimentalTmuxBackend, attributionTrailers
     }
 }
 
--- a/Packages/CrowCore/Tests/CrowCoreTests/AppConfigTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AppConfigTests.swift
@@ -45,6 +45,7 @@ import Testing
     #expect(config.remoteControlEnabled == false)
     #expect(config.managerAutoPermissionMode == true)
     #expect(config.experimentalTmuxBackend == false)
+    #expect(config.attributionTrailers == true)
 }
 
 @Test func appConfigRemoteControlRoundTrip() throws {
@@ -244,6 +245,28 @@ import Testing
     let data2 = try JSONEncoder().encode(config)
     let decoded2 = try JSONDecoder().decode(AppConfig.self, from: data2)
     #expect(decoded2.experimentalTmuxBackend == false)
+}
+
+@Test func appConfigAttributionTrailersRoundTrip() throws {
+    var config = AppConfig()
+    config.attributionTrailers = false
+
+    let data = try JSONEncoder().encode(config)
+    let decoded = try JSONDecoder().decode(AppConfig.self, from: data)
+    #expect(decoded.attributionTrailers == false)
+
+    config.attributionTrailers = true
+    let data2 = try JSONEncoder().encode(config)
+    let decoded2 = try JSONDecoder().decode(AppConfig.self, from: data2)
+    #expect(decoded2.attributionTrailers == true)
+}
+
+@Test func appConfigAttributionTrailersDefaultsTrueWhenKeyMissing() throws {
+    // Legacy configs without the key opt in by default — matches the behavior
+    // users see when they install the feature without touching settings.
+    let json = #"{"workspaces": [], "remoteControlEnabled": false}"#.data(using: .utf8)!
+    let config = try JSONDecoder().decode(AppConfig.self, from: json)
+    #expect(config.attributionTrailers == true)
 }
 
 @Test func ignoreReviewLabelsRoundTrip() throws {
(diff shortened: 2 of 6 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues

None blocking.

### Security Review

**Strengths:**
- The `settings.local.json` is excluded from version control via the per-worktree git exclude file, preventing accidental commits of session metadata.
- The `is_attribution_trailers_enabled` function defaults to **on** (return 0) when the config is missing or malformed, matching `AppConfig`'s `decodeIfPresent` default — good parity.
- The `write_settings_local` function gracefully handles missing `SESSION_ID` and disabled config without failing the overall setup.

**Concerns:**
- `setup.sh:349-354` — `SESSION_ID` is interpolated directly into a heredoc without sanitization. When supplied via `--session-id` from the CLI, a malicious value could inject arbitrary keys into the JSON. In practice, the value comes from `crow new-session` (a trusted binary), and the `--session-id` path is only used internally by the LLM orchestrator, so the risk is low. Still, a UUID format check (e.g., `[[ "$SESSION_ID" =~ ^[a-f0-9-]{36}$ ]]`) before interpolation would be a cheap hardening measure.
- `setup.sh:83-84` — The grep-based config parsing (`is_attribution_trailers_enabled`) searches for a literal string pattern rather than parsing JSON. A contrived config with `"attributionTrailers": false` inside a string value (e.g., a `"note"` field) would produce a false negative. Using `jq` would be more robust, but the current approach is consistent with the existing `is_remote_control_enabled` pattern and unlikely to cause issues with normal Crow-managed configs.

### Code Quality

- **AppConfig.swift:** Clean addition following the established pattern — `decodeIfPresent` with a default, `CodingKeys` updated, init parameter added. Consistent with `experimentalTmuxBackend` which was the last addition.
- **Tests:** Two new tests (`appConfigAttributionTrailersRoundTrip`, `appConfigAttributionTrailersDefaultsTrueWhenKeyMissing`) cover the round-trip and legacy-config-upgrade paths. The empty-JSON decode test also asserts the new default. All 145 CrowCore tests pass.
- **UI wiring:** `AutomationSettingsView` correctly takes an `attributionTrailers` binding and saves on toggle. `SettingsView` passes it through. The help text ("Applies to new worktrees only") is accurate and sets user expectations correctly.
- **SKILL.md:** Documentation is well-placed and includes opt-out instructions.
- **setup.sh:** The `write_settings_local` function is well-structured — it runs for all worktrees (not just `--primary`), handles edge cases gracefully, and the git-exclude logic is idempotent (checks `grep -qxF` before appending).

### Summary Table

| Priority | Issue |
|----------|-------|
| 🟡 Yellow | Consider adding UUID format validation for `SESSION_ID` before JSON interpolation in `write_settings_local` (`setup.sh:338-339`) |
| 🟢 Green | Consider using `jq` for config flag parsing instead of grep (lower priority — consistent with existing `is_remote_control_enabled` pattern) |

**Recommendation:** Approve — clean, well-tested feature with good UX (default-on, easy opt-out, clear help text). The yellow item is a hardening suggestion for a future pass, not a blocker.

PR #251, opened 2026-05-08
Run session deletion off the main actor (closes #250)

### LINKED ISSUE #250: Session deletion blocks UI with macOS beach-ball spinner — make it async
## Context

When deleting sessions in the Crow app — either single sessions via the per-row delete action, or multiple sessions via the multi-select batch delete flow — the macOS beach-ball spinner appears and the entire UI blocks until the operation completes.

This makes the app feel hung even though the work is progressing. For batch deletes of several sessions (each touching state, persistence, possibly worktree cleanup), the freeze can last several seconds.

## Expected behavior

Delete should be asynchronous:
- UI remains responsive during the operation
- A non-blocking progress indicator (e.g., inline spinner on the affected row(s), or a status pill in the sidebar) shows work in progress
- The user can continue interacting with other parts of the app — switching sessions, scrolling the list, etc.
- Failures surface via toast/inline error rather than blocking modal

## Likely root cause

The deletion path is probably running on the main thread (or `await`ing on `MainActor` for the full duration). Per `CLAUDE.md`, all RPC handlers serialize state mutations via `await MainActor.run { ... }` — that's correct for the state mutation itself, but if the delete handler also performs the worktree teardown / disk persistence inline on the main actor, the UI can't repaint while it runs.

Investigation should look at:
- The delete-session RPC handler — what work happens on `MainActor.run` vs. what could move to a background task
- JSONStore persistence on delete — NSLock-serialized disk writes are fine, but if they happen synchronously on the main thread the UI stalls
- Any worktree filesystem cleanup triggered by delete (rm -rf of worktree directories is the most likely culprit for multi-second freezes)
- Multi-select batch delete: is it iterating sequentially with main-thread awaits, or fanning out work concurrently?

## Acceptance criteria

- [ ] Single-session delete does not produce a beach-ball spinner; UI remains interactive
- [ ] Batch delete of N sessions (test with N=5, N=10) does not block the UI
- [ ] Per-session delete progress is visible without blocking
- [ ] Errors surface non-modally
- [ ] No regression in correctness — sessions, worktrees, terminals, and persisted state all clean up properly
- [ ] No race conditions: deleting a session while it's still mid-cleanup either queues or rejects safely

dhilgaertner:
## Summary
- `SessionService.deleteSession` now snapshots its work on the main actor, runs `git worktree remove`/`branch -D`/`worktree prune` and the residual `FileManager.removeItem` inside `Task.detached(priority: .utility)`, then finalizes state back on the main actor — so the UI repaints during the slow part instead of beach-balling.
- Adds `AppState.isDeletingSession` and `AppState.sessionDeletionError`. While a delete is in flight the row dims and shows a `ProgressView`; on success the row goes away; on a fatal cleanup failure the session stays in place with a red `exclamationmark.triangle.fill` (with hover-text describing the cause) for ~6 s so the user can retry. Per-row and detail-view delete actions, plus the "Mark as In Review" / "Mark as Completed" context-menu items, are disabled while a delete is in flight.
- `BulkDeleteSessionsAlert` now fans each delete out as its own `Task` (joined via `await task.value`) instead of iterating serially, so multi-select delete of N sessions finishes in roughly the time of a single delete.
- The CLI `delete-session` RPC inherits the same path — no separate code path.

## Test plan
- [ ] `make build` is clean
- [ ] `swift test` (root) and `swift test` in `Packages/CrowUI` both pass (existing 63 + 22 tests)
- [ ] Single delete via row context menu: row dims + spinner appears; sidebar/main view stay interactive (scroll list, switch sessions); row disappears once cleanup completes
- [ ] Multi-select delete of 5 and 10 sessions: all selected rows show spinners simultaneously; UI stays responsive; rows disappear in roughly one delete's time, not N×
- [ ] Force a failure (`rm -rf` a worktree directory before pressing delete so `git worktree remove` fails): row stays, red triangle appears with hover-text, NSLog matches; retrying after the auto-clear succeeds
- [ ] CLI: `crow delete-session --session <uuid>` still completes without UI freeze
- [ ] After each scenario, `git worktree list` and `~/Library/Application Support/Crow/store.json` show no leftover state for the deleted IDs

Closes #250

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -254,6 +254,15 @@ public final class AppState {
     /// Must be cleaned up when a session is deleted (see `SessionService.deleteSession`).
     public var isMarkingInReview: [UUID: Bool] = [:]
 
+    /// Sessions whose async deletion (worktree teardown, branch removal, persistence)
+    /// is currently in progress. Set on the main actor at the start of
+    /// `SessionService.deleteSession` and cleared when the session is fully removed.
+    public var isDeletingSession: [UUID: Bool] = [:]
+
+    /// Most recent delete-cleanup error for a session, surfaced inline on the row
+    /// so failures aren't silent. Auto-cleared after a short delay or on retry.
+    public var sessionDeletionError: [UUID: String] = [:]
+
     /// Called to open a session's primary worktree in VS Code.
     public var onOpenInVSCode: ((UUID) -> Void)?
 
--- a/Packages/CrowUI/Sources/CrowUI/DeleteSessionAlert.swift
+++ b/Packages/CrowUI/Sources/CrowUI/DeleteSessionAlert.swift
@@ -134,7 +134,8 @@ enum DeleteSessionMessageBuilder {
 // MARK: - Bulk Delete Sessions Alert
 
 /// View modifier that attaches a bulk delete-sessions confirmation alert.
-/// Iterates `selectedIDs` serially through `appState.onDeleteSession`.
+/// Fans out `selectedIDs` concurrently through `appState.onDeleteSession` so
+/// the UI doesn't wait N×single-delete to repaint.
 struct BulkDeleteSessionsAlert: ViewModifier {
     @Binding var isPresented: Bool
     let selectedIDs: Set<UUID>
@@ -147,14 +148,22 @@ struct BulkDeleteSessionsAlert: ViewModifier {
             Button(buttonLabel, role: .destructive) {
                 let snapshot = sortedSnapshot
                 Task {
+                    // Fan each delete out as its own Task so they run concurrently.
+                    // SessionService.deleteSession yields the main actor while its
+                    // disk/git cleanup runs on a detached task, so all selected
+                    // sessions clean up in parallel rather than serially.
+                    var tasks: [Task<Void, Never>] = []
                     for id in snapshot {
-                        do {
-                            try await appState.onDeleteSession?(id)
-                        } catch {
-                            NSLog("Failed to delete session \(id): \(error)")
-                        }
+                        tasks.append(Task {
+                            do {
+                                try await appState.onDeleteSession?(id)
+                            } catch {
+                                NSLog("Failed to delete session \(id): \(error)")
+                            }
+                        })
                     }
-                    await MainActor.run { onCompletion() }
+                    for t in tasks { await t.value }
+                    onCompletion()
                 }
             }
         } message: {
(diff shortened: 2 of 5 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None found.

### Security Review
**Strengths:**
- `WorktreeCleanupItem` is `Sendable` — clean data transfer across actor boundaries with no risk of data races on `AppState`.
- Protected branch guard (`SessionWorktree.isProtectedBranch`) is preserved on the background path, preventing accidental deletion of `main`/`master`.
- Manager session (`00000000-...`) delete guard is preserved.
- `ShellEnvironment.shared.env` is used for `Process.environment`, keeping the environment consistent and avoiding leaking user `PATH` variations.

**Concerns:**
- None material. The `runShellSync` helper constructs `Process` arguments from trusted internal state (repo paths, branch names) rather than user input, so command injection is not a risk here.

### Code Quality

**Well done:**
- The snapshot-then-detach pattern (`WorktreeCleanupItem` value type → `Task.detached`) is the correct way to avoid holding `@MainActor`-isolated references across an `await` boundary. Clean separation of concerns.
- The re-entrancy guard (`guard appState.isDeletingSession[id] != true`) prevents double-delete races.
- `performDiskCleanup` is `nonisolated static` — no accidental captures of `self` or `appState`. Correct.
- Error auto-clear with the 6-second `Task.sleep` is a nice UX pattern, and the `[weak appState]` capture avoids a retain cycle.
- Bulk delete fan-out (`Task` per session, `await t.value` to join) is a straightforward improvement over serial iteration.

1. **`runShellSync` reads pipes after `waitUntilExit`** (`SessionService.swift:651-652`): calling `readDataToEndOfFile()` *after* the process exits is fine for the short outputs expected here (git status messages), but for very large stdout/stderr this can technically deadlock if the pipe buffer fills before the process exits. Not a practical concern for `git worktree remove` / `git branch -D` / `git worktree prune` — mentioning only for awareness.

2. **Bulk delete error visibility** (`DeleteSessionAlert.swift:157-163`): when individual deletes in a bulk operation fail, errors are `NSLog`'d but the per-session error indicators will show on the rows (via `sessionDeletionError`). The `onCompletion()` callback (which clears the selection) fires after all tasks join — if some sessions failed and remained, they'll be deselected but show the red triangle. This is reasonable behavior.

3. **`onCompletion()` removed `MainActor.run` wrapper** (`DeleteSessionAlert.swift:165`): the outer `Task { }` inherits the main actor from the `@MainActor`-isolated `body` context, so after the `for t in tasks { await t.value }` loop resumes, we're back on the main actor. This is correct.

### Summary Table
| Priority | Issue |
|----------|-------|
| 🟢 | `readDataToEndOfFile` after `waitUntilExit` — fine for these short git commands |
| 🟢 | Consider whether bulk-delete selection clearing after partial failure is the desired UX |

**Recommendation:** Approve — the async-delete architecture is sound, the concurrency boundaries are correct, the UI feedback (spinner, error indicator, disabled state) is thorough, and there are no security concerns. Ship it.

PR #253, opened 2026-05-12
Stop forging tmux .shellReady on first-prompt timeout

### LINKED ISSUE #252: Auto-review pastes Claude launch command before tmux shell is ready
## Summary

When **auto-review is enabled** and the **tmux backend** is in use, the review tab can open with the literal `claude --rc --name 'review-X-N' "$(cat …prompt.md)"` rendered in a blank pane and no Claude TUI ever appears.

Reported by a teammate; reproduces most reliably when:
- it's the user's **first** tmux session of the app lifetime (cold tmux server),
- the Crow.app is **backgrounded** when the auto-review fires (App Nap throttles the tmux server, the wrapper, the user's shell, and Crow's own polling Task),
- the user has a **heavy `.zshrc`** (oh-my-zsh, asdf, nvm, starship, etc.).

## Root cause

`TmuxBackend.startReadinessWatch` waits for the bundled `crow-shell-wrapper.sh`'s `precmd` to touch a sentinel file. The watch has a 30 s budget. **On timeout it forges `.shellReady` anyway** — see `Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:356-370`. The accompanying comment justifies this for `claude --continue` (small, idempotent), but the same code path now runs for review sessions whose payload is a multi-KB `"$(cat .crow-review-prompt.md)"` shell substitution that only works against a live `zle`.

When the timeout fires:
1. `SessionService.wireTerminalReadiness` sees the (fake) `.shellReady` and calls `launchClaude`.
2. `launchClaude` builds the review-prompt command for review sessions (`Sources/Crow/App/SessionService.swift:358-369`).
3. `TmuxBackend.sendText` pastes it via `load-buffer` + `paste-buffer` into a pane where the shell may still be starting (or where `zle` doesn't yet exist). The bytes pile up in the PTY input queue or get partially consumed by the wrapper's own subshells, and the user sees the dead "blank window with command" state.

## Expected behavior

- The readiness state machine should not lie. `.shellReady` should mean "sentinel actually fired" — not "we gave up waiting."
- A timed-out terminal should surface a recoverable affordance to the user.
- Returning the app to the foreground should automatically retry stuck readiness watches without requiring user action (since App Nap is the dominant cause).

## Proposed fix (implemented in the linked PR)

1. New `TerminalReadiness.timedOut` case, ordered between `.surfaceCreated` and `.shellReady` so all existing `< .shellReady` gates keep working.
2. `TmuxBackend.startReadinessWatch` now emits `.timedOut` on timeout instead of `.shellReady`. New `retryReadinessWatch(id:, timeoutBudget:)` (default 120 s) clears the stale sentinel and re-arms.
3. `SessionService` handles `.timedOut` without calling `launchClaude`. New `retryReadiness(terminalID:)` and `reArmStuckReadinessWatches()` helpers.
4. `AppDelegate` observes `NSApplication.didBecomeActiveNotification` and re-arms stuck watches automatically when the app returns to the foreground.
5. UI renders a "Terminal didn't become ready in time — Retry" overlay for `.timedOut` in `SessionDetailView`, and a pulsing yellow indicator in `SessionListView`.

## Verification

1. Set `autoReviewRepos` to a repo with at least one open PR awaiting review.
2. Fully quit and relaunch Crow, then immediately `cmd-h` to background it. Let the machine sit idle 60+ s so the poll fires while App Nap is engaged.
3. Re-foreground: the review session should either show Claude already running (happy path) or the new Retry overlay (degraded path). It must never show the bare `claude …` paste with no Claude UI.
4. Click Retry on the degraded case; confirm the watch re-arms and Claude launches.
5. Console logs: `[CrowTelemetry tmux:first_prompt_timeout …]` should still appear, but no `tmux readiness: … state=shellReady` should follow it on the same terminal id.

dhilgaertner:
Closes #252.

## Summary
- TmuxBackend used to fire `.shellReady` after a 30 s sentinel timeout to dodge a stuck spinner. That was safe for `claude --continue` but pastes auto-review's multi-KB `"$(cat …prompt.md)"` into a possibly-unborn `zle`, producing the blank-pane / orphaned-command state our teammate hit.
- Add a `.timedOut` state, route timeouts through it without calling `launchClaude`, and expose Retry (manual button + automatic on `didBecomeActive` since App Nap is the dominant cause).
- All 9 changed files; no behavior change on the happy path.

## Test plan
- [x] `make build`
- [x] CrowCore tests (146), incl. new `timedOutGatesAutoPaste`
- [x] CrowTerminal tests (21), incl. new `retryReadinessEmitsTimedOutWhenSentinelMissing` — backend log confirms `tmux:first_prompt_timeout terminal=… budget_ms=150` fires
- [x] CrowUI tests (22)
- [x] Root tests (63)
- [ ] Manual: enable auto-review for a repo with a pending PR, fully quit and relaunch Crow, immediately `cmd-h`, let the poll fire while backgrounded, re-foreground and confirm either Claude is running or the new Retry overlay appears (never the bare-command paste)
- [ ] Manual: confirm Retry on the degraded case re-arms the watch and Claude launches
- [ ] Manual happy-path regression: non-review tmux session with a normal shell still hits `.shellReady` and runs `claude --continue` instantly

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -218,6 +218,10 @@ public final class AppState {
     /// Called when the user clicks "Retry" on a failed terminal surface.
     public var onRetryTerminal: ((UUID) -> Void)?  // receives terminal ID
 
+    /// Called when the user clicks "Retry" on a terminal whose tmux readiness
+    /// watch timed out before the shell signaled it was interactive.
+    public var onRetryReadiness: ((UUID) -> Void)?  // receives terminal ID
+
     /// Called to add a new plain-shell terminal tab to a session.
     public var onAddTerminal: ((UUID) -> Void)?  // receives session ID
 
--- a/Packages/CrowCore/Sources/CrowCore/Models/Enums.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/Enums.swift
@@ -42,6 +42,7 @@ public enum TerminalReadiness: String, Codable, Sendable, Comparable {
     case failed           // createSurface() exhausted retries; UI shows error overlay with Retry
     case uninitialized    // GhosttySurfaceView exists but createSurface() not called
     case surfaceCreated   // ghostty_surface_t exists, shell process spawning
+    case timedOut         // Sentinel never appeared within the readiness budget; UI shows Retry
     case shellReady       // Shell prompt detected (probe file appeared)
     case claudeLaunched   // claude --continue has been sent
 
@@ -50,8 +51,9 @@ public enum TerminalReadiness: String, Codable, Sendable, Comparable {
         case .failed: -1
         case .uninitialized: 0
         case .surfaceCreated: 1
-        case .shellReady: 2
-        case .claudeLaunched: 3
+        case .timedOut: 2
+        case .shellReady: 3
+        case .claudeLaunched: 4
         }
     }
 
--- a/Packages/CrowCore/Tests/CrowCoreTests/TerminalReadinessTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/TerminalReadinessTests.swift
@@ -9,7 +9,8 @@ struct TerminalReadinessTests {
 
     @Test func statesAreOrdered() {
         #expect(TerminalReadiness.uninitialized < .surfaceCreated)
-        #expect(TerminalReadiness.surfaceCreated < .shellReady)
+        #expect(TerminalReadiness.surfaceCreated < .timedOut)
+        #expect(TerminalReadiness.timedOut < .shellReady)
         #expect(TerminalReadiness.shellReady < .claudeLaunched)
     }
 
@@ -17,12 +18,22 @@ struct TerminalReadinessTests {
         #expect(TerminalReadiness.uninitialized < .claudeLaunched)
         #expect(TerminalReadiness.uninitialized < .shellReady)
         #expect(TerminalReadiness.surfaceCreated < .claudeLaunched)
+        #expect(TerminalReadiness.timedOut < .claudeLaunched)
+    }
+
+    @Test func timedOutGatesAutoPaste() {
+        // The launchClaude guard is `== .shellReady`, so .timedOut must NOT
+        // satisfy `< .shellReady ? false : true` equality checks. Verify the
+        // ordering keeps .timedOut strictly below .shellReady.
+        #expect(TerminalReadiness.timedOut < .shellReady)
+        #expect(TerminalReadiness.timedOut != .shellReady)
     }
 
     @Test func equalStatesAreNotLessThan() {
         #expect(!(TerminalReadiness.uninitialized < .uninitialized))
         #expect(!(TerminalReadiness.shellReady < .shellReady))
         #expect(!(TerminalReadiness.claudeLaunched < .claudeLaunched))
+        #expect(!(TerminalReadiness.timedOut < .timedOut))
     }
 
     // MARK: - Equality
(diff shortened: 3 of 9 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None found.

### Security Review
**Strengths:**
- Sentinel file paths use `TMPDIR` (line 332 of `TmuxBackend.swift`) — no hardcoded `/tmp` path.
- The `retryReadinessWatch` method removes the stale sentinel file before re-arming (`try? FileManager.default.removeItem`), preventing a stale sentinel from immediately satisfying the next watch.
- No secrets, credentials, or sensitive data in logging — telemetry lines only log terminal UUIDs and timing.

**Concerns:**
- None identified. The changes don't introduce new IPC, network, or user-input surfaces.

### Code Quality

**State machine design is solid:**
- The new `.timedOut` state is correctly ordered between `.surfaceCreated` (2) and `.shellReady` (3) in the `Comparable` implementation, so existing `< .shellReady` guards throughout the codebase correctly exclude it without modification.
- The `launchClaude` guard (`== .shellReady`, `SessionService.swift:333`) is the right check — equality, not `>=` — so `.timedOut` can never accidentally trigger auto-paste. The new `timedOutGatesAutoPaste` test explicitly validates this invariant.
- `retryReadiness` correctly leaves the terminal in `autoLaunchTerminals` (the set is only drained on successful `launchClaude`), so a successful retry still flows through to Claude launch.

**Concurrency:**
- Both `retryReadiness` and `reArmStuckReadinessWatches` execute on `@MainActor` via their call sites (button action → `AppState.onRetryReadiness`, and `didBecomeActiveNotification` → `Task { @MainActor }`). No off-actor mutation of `appState`.
- Multiple concurrent watches for the same terminal are safe: each `startReadinessWatch` creates a fresh `SentinelWaiter` + `Task`, and the `onReadinessChanged` callback guards (`currentState < .shellReady`) prevent double-launch.

**Minor nit (non-blocking):**
- `retryReadiness` allows re-entry when `current < .shellReady` (i.e., `.surfaceCreated`), which means a rapid double-click on "Retry" while a watch is already running would spin up a second parallel watcher. This is harmless (both resolve independently and the callback is idempotent), but you could tighten the guard to `current == .timedOut` if you want to prevent the extra work. The `reArmStuckReadinessWatches` path already checks `state == .timedOut` exclusively.

**UI:**
- The `.timedOut` overlay in `SessionDetailView` uses a distinct icon (`clock.badge.exclamationmark`) and yellow coloring — clearly differentiated from the `.failed` overlay (orange triangle). Good UX.
- The `SessionListView` dot indicator for `.timedOut` uses a pulsing ring effect to distinguish it from the plain yellow `.surfaceCreated` dot. Accessible label is descriptive.
- `fixedSize(horizontal: false, vertical: true)` on the explanation text prevents truncation — correct for variable-width parent.

**Tests:**
- New `timedOutGatesAutoPaste` test in `TerminalReadinessTests` validates the core safety invariant.
- New `retryReadinessEmitsTimedOutWhenSentinelMissing` test in `TmuxBackendTests` exercises the real retry path with a 150ms budget and confirms `.timedOut` is emitted (not `.shellReady`).
- Existing tests updated to include `.timedOut` in ordering, equality, raw-value, and codable round-trip checks.

### Summary Table
| Priority | Issue |
|----------|-------|
| Green | Double-retry creates parallel watchers (harmless, could tighten guard) |

**Recommendation:** Approve — the fix correctly addresses the root cause (blind paste into unready shell), the state machine is well-designed, concurrency is sound, tests cover the critical invariant, and the UI provides both manual and automatic recovery paths.

PR #258, opened 2026-05-12
tmux readiness diagnostics: wrapper breadcrumbs + on-timeout capture

### LINKED ISSUE #256: tmux readiness: shell never reaches precmd for some users — add diagnostics + harden wrapper hook
## Context

Follow-up to #252 / PR #253. With #253 merged, a 30-second sentinel miss now surfaces as `.timedOut` with a Retry affordance (no more blind paste). That fixed the *symptom*. A teammate still hits the underlying problem on the latest build: **the sentinel never fires, no matter how long you wait** — Retry doesn't recover because the shell is never reaching `_crow_precmd` in the first place.

Reporter's observation: "It appears to me as if the tmux session is never coming up because the shell is never coming up. No matter how long you wait it's not just a timing issue."

The current logs in this area are insufficient to tell us *why*. We need stage-by-stage diagnostics through the wrapper + Swift watch so the next person reporting this can hand us a useful trace.

## Why the existing logs aren't enough

`TmuxBackend.startReadinessWatch` (`Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:337-372`) emits exactly two telemetry lines per terminal:

- `tmux:first_prompt_ms=<n>` on success
- `tmux:first_prompt_timeout terminal=<id> budget_ms=30000` on timeout

That's it. On timeout we don't know:

- Did `crow-shell-wrapper.sh` even start?
- Did it get past sourcing the user's `~/.zshrc`?
- Did `add-zsh-hook precmd _crow_precmd` succeed?
- Did `exec "$SHELL" -i` succeed?
- Did `_crow_precmd` fire but fail to write `$CROW_SENTINEL` (the `: > "$CROW_SENTINEL" 2>/dev/null || true` swallows that)?
- Is the pane process the shell, the wrapper still, or dead?

Every one of those is a plausible failure mode and we currently can't tell them apart.

### 1. Wrapper-side stage breadcrumbs

`Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-shell-wrapper.sh`:

- Add a `CROW_WRAPPER_LOG` env var (set by the Swift side to e.g. `$TMPDIR/crow-wrapper-<terminal-uuid>.log`).
- Emit one timestamped line per stage to that file:
  - `start pid=$$ shell=$SHELL zdotdir_user=$ZDOTDIR tmux=${TMUX:+yes}`
  - `zdotdir_temp_created path=$ZTMP` (or whatever bash equivalent)
  - `user_rc_sourced rc=<path> rc_size=<bytes>` (or `_skipped reason=<reason>`)
  - `hook_installed mechanism=add-zsh-hook|precmd_functions|PROMPT_COMMAND`
  - `pre_exec shell=$SHELL`
  - From inside `_crow_precmd`: `precmd_fired sentinel=$CROW_SENTINEL`
  - And: `sentinel_written` **or** `sentinel_write_failed errno=$?` (stop swallowing the failure)
- Optional `CROW_WRAPPER_DEBUG=1` toggle that adds `set -x` redirected to the same log.
- If `mktemp -d` fails, log it and `exit 65` instead of `exec`'ing a broken environment.

### 2. Swift-side diagnostics on timeout

`TmuxBackend.swift`:

- Pass `CROW_WRAPPER_LOG` into the terminal env (alongside `CROW_SENTINEL`).
- On `.timedOut`, capture and log:
  - **`tmux capture-pane -t <pane> -p -S -200`** — last 200 lines of the pane's visible output (most useful single signal — usually shows whether we're stuck at the shell prompt, mid-`.zshrc`, or a Python error from oh-my-zsh).
  - **`tmux display-message -p '#{pane_pid} #{pane_current_command}'`** — what's actually running in the pane right now.
  - **`ps -o pid,ppid,stat,etime,command -p <pane_pid>`** (and one level of descendants if the pane PID is still the wrapper) — is the wrapper still alive? is `exec` having an off day?
  - **Wrapper log contents** (`cat $CROW_WRAPPER_LOG`) — the breadcrumb trail from step 1.
  - **Sentinel state**: does `$CROW_SENTINEL` exist? Is its parent writable? `stat` it.
- Add a periodic mid-wait log every 10s: `tmux:first_prompt_progress terminal=<id> elapsed_ms=<n> sentinel_exists=<bool>` so we know whether the watch is healthy.

### 3. UI affordance for capturing all this

In the `.timedOut` Retry overlay (added in #253), add a "Copy diagnostics" button that bundles everything from §2 above + `tmux -V` + `$PATH` + `$SHELL` + `dscl . -read /Users/$USER UserShell` into one clipboard payload. Saves the next reporter from screenshot-archaeology.

### 4. Speculative robustness fixes (do these while we're in here)

The hypothesis is "people have different shell setups and they're tripping the wrapper." Two concrete shapes that could break us today:

- **`.zshrc` wipes `precmd_functions`.** Our `add-zsh-hook precmd _crow_precmd` runs at the end of our ZDOTDIR's `.zshrc`, *after* sourcing the user's. If the user's `.zshrc` ends by setting `precmd_functions=(starship_precmd)` (rare but exists in some dotfiles), we never get installed. **Fix**: append, don't replace — re-install with `precmd_functions+=(_crow_precmd)` after the user config sources, **and** also keep the `add-zsh-hook` call as a primary path.
- **`$SHELL` is set to something exotic** (chsh'd to a brew-installed shell at a weird path, or a wrapper script). The `case "$SHELL" in *zsh) … *bash) … *)` falls through to the best-effort branch that doesn't install a hook. **Fix**: cross-reference with `dscl . -read /Users/$USER UserShell` and log when `$SHELL` doesn't match the directory-services value (don't override — just surface).

These are guesses informed by the wrapper's current shape. The diagnostics in §1-3 are what tell us if these guesses are even on the right track.

## Acceptance criteria

- [ ] A teammate hitting the bug can paste a one-clipboard diagnostic bundle into a comment, and the bundle includes wrapper stage logs + pane capture + sentinel state.
- [ ] `tmux:first_prompt_timeout` log line is accompanied by a follow-up telemetry block with the captured pane + ps tree + wrapper log.
- [ ] Wrapper exits non-zero (and logs) if any of its critical setup steps fail (mktemp, source, hook install) rather than silently `exec`'ing a broken shell.
- [ ] Speculative hardening fixes from §4 are in place with a one-line comment each citing this issue.
- [ ] Unit test asserts the wrapper writes the expected breadcrumbs given `CROW_WRAPPER_LOG` set + a fake shell.
- [ ] Manual test: reproduce the teammate's "shell never starts" repro (heavy `.zshrc` + cold app + background) and confirm the captured bundle pinpoints the failing stage.

## Out of scope

- Switching away from tmux as the backend (#198 already debated this).
- Eliminating the wrapper entirely (e.g., reading prompts directly via a tmux hook). Worth considering, but a different ticket — this one is about making the current architecture diagnosable.
- Fixing whatever specific user-setup bug the bundle ultimately surfaces. Once we have the diagnostics, the actual root-cause fix is a follow-up ticket.

## Related

- #252 / PR #253 — surface `.timedOut` instead of forging `.shellReady` (just merged)
- #198 — tmux backend behind feature flag
- Wrapper source: `Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-shell-wrapper.sh`
- Swift watch: `Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:337-372`

dhilgaertner:
Closes #256.

## Summary
- **Wrapper-side stage breadcrumbs** in `crow-shell-wrapper.sh`: new `CROW_WRAPPER_LOG` env var; one timestamped line per stage (`start`, `zdotdir_temp_created`, `user_rc_sourced`/`_skipped`, `hook_installed mechanism=…`, `pre_exec`, `precmd_fired`, `sentinel_written`/`sentinel_write_failed status=$?`). Stops swallowing the sentinel-write failure. Hard `exit 73` on `mktemp` failure instead of `exec`-ing into a broken env. Opt-in `CROW_WRAPPER_DEBUG=1` enables `set -x` to the same log.
- **Swift-side diagnostics on `.timedOut`**: `TmuxBackend.captureDiagnostics(id:)` builds a multi-section bundle — `tmux -V`/`PATH`/`SHELL`/`dscl . -read /Users/$USER UserShell` header, `display-message` for pane PID + current command, `capture-pane -p -S -200`, `ps` tree (pane PID + descendants via `pgrep -P`), sentinel state + writability, full wrapper log. Each section size-capped to keep the clipboard payload sane. NSLogged on timeout under `[CrowTelemetry tmux:first_prompt_diagnostics …]`.
- **Periodic progress logging**: a sibling task fires `tmux:first_prompt_progress terminal=… elapsed_ms=… sentinel_exists=…` every 10s so operators tailing logs can see whether the watch is healthy.
- **UI "Copy diagnostics" button** in the `.timedOut` overlay added by #253 — bundles the same diagnostics onto the clipboard with a 1.5s "Copied!" confirmation.
- **Speculative hardening** (one-line comments cite #256): `precmd_functions+=(_crow_precmd)` belt-and-braces after `add-zsh-hook` for the deferred-plugin-reassignment edge case; `$SHELL` vs `dscl` cross-check logged at the `start` stage when they disagree.
- **Tests**: new `CrowShellWrapperTests.swift` — setup-phase breadcrumbs assertion (always runs, uses a fake `*-zsh` shell), runtime breadcrumbs (gated on `/bin/zsh`), sentinel-write-failure logging.

## Test plan
- [x] `swift build` at root — green
- [x] `swift test --package-path Packages/CrowTerminal` — 24/24 (incl. 3 new wrapper tests)
- [x] `swift test --package-path Packages/CrowCore` — 150/150
- [x] `swift test --package-path Packages/CrowUI` — 22/22
- [x] `swift test` at root — 68/68
- [x] Inspect captured bundle inline (visible in `retryReadinessEmitsTimedOutWhenSentinelMissing` test stderr) — wrapper log + pane capture + ps tree + sentinel state all rendered.
- [ ] Manual: heavy `.zshrc` (`sleep 5`) + cold app + immediate `cmd-h` → `.timedOut` overlay → click "Copy diagnostics" → verify pasted bundle contains breadcrumbs pinpointing the failing stage.
- [ ] Manual happy-path regression: normal session reaches `.shellReady`; wrapper log shows full `start → … → sentinel_written` chain.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -222,6 +222,12 @@ public final class AppState {
     /// watch timed out before the shell signaled it was interactive.
     public var onRetryReadiness: ((UUID) -> Void)?  // receives terminal ID
 
+    /// Called when the user clicks "Copy diagnostics" on a terminal whose
+    /// tmux readiness watch timed out. The handler captures a multi-section
+    /// bundle (wrapper log, pane capture, ps tree, sentinel state) and
+    /// places it on the clipboard (issue #256).
+    public var onCopyDiagnostics: ((UUID) -> Void)?  // receives terminal ID
+
     /// Called to add a new plain-shell terminal tab to a session.
     public var onAddTerminal: ((UUID) -> Void)?  // receives session ID
 
--- a/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-shell-wrapper.sh
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-shell-wrapper.sh
@@ -4,7 +4,9 @@
 # Bundled with Crow.app, copied to a temp path at terminal create time. The
 # Crow app sets CROW_SENTINEL in the env (per-terminal) before exec'ing this
 # script. Each prompt firing touches that file, which the Swift readiness
-# detector polls.
+# detector polls. Optional CROW_WRAPPER_LOG points at a per-terminal log file
+# the wrapper appends stage breadcrumbs to (issue #256) — Swift surfaces it
+# on .timedOut so a teammate can copy a one-clipboard diagnostic bundle.
 #
 # This wrapper is intentionally minimal:
 #   1. Source the user's normal shell startup (zsh: .zshrc; bash: .bashrc) so
@@ -27,7 +29,37 @@ if [ -z "${CROW_SENTINEL:-}" ]; then
 fi
 export CROW_SENTINEL
 
+# CROW_WRAPPER_LOG is optional. Default to /dev/null so the helper is always
+# safe to call without an unset-var guard. Issue #256.
+CROW_WRAPPER_LOG="${CROW_WRAPPER_LOG:-/dev/null}"
+export CROW_WRAPPER_LOG
+
+crow_log() {
+  printf '%s %s\n' "$(date -u +%Y-%m-%dT%H:%M:%SZ)" "$*" >> "$CROW_WRAPPER_LOG" 2>/dev/null || true
+}
+
+# Opt-in verbose tracing for hard-to-repro setup bugs. Routes shell xtrace to
+# the same log file (issue #256).
+if [ "${CROW_WRAPPER_DEBUG:-}" = "1" ] && [ "$CROW_WRAPPER_LOG" != "/dev/null" ]; then
+  exec 2>>"$CROW_WRAPPER_LOG"
+  set -x
+fi
+
 if [ -z "${SHELL:-}" ]; then SHELL=/bin/zsh; fi
+
+crow_log "start pid=$$ shell=$SHELL tmux=${TMUX:+yes} sentinel=$CROW_SENTINEL log=$CROW_WRAPPER_LOG"
+
+# Cross-check $SHELL against directory services. Mismatch can mean the user
+# chsh'd to a brew-installed shell but the parent process inherited an old
+# $SHELL — surfacing this in the log helps diagnose the case in issue #256 §4.
+# We do NOT override $SHELL; just record the discrepancy.
+if command -v dscl >/dev/null 2>&1 && [ -n "${USER:-}" ]; then
+  _crow_dscl_shell=$(dscl . -read "/Users/$USER" UserShell 2>/dev/null | awk '{print $2}')
+  if [ -n "$_crow_dscl_shell" ] && [ "$_crow_dscl_shell" != "$SHELL" ]; then
+    crow_log "shell_mismatch env_shell=$SHELL dscl_shell=$_crow_dscl_shell"
+  fi
+  unset _crow_dscl_shell
+fi
 
 case "$SHELL" in
   *zsh)
@@ -35,16 +67,32 @@ case "$SHELL" in
     # at a temp dir whose .zshrc sources the user's real config, then appends
     # our hook via add-zsh-hook (which composes with any existing precmd).
     ZTMP="$(mktemp -d -t crowzdotdir)"
+    if [ -z "$ZTMP" ] || [ ! -d "$ZTMP" ]; then
+      crow_log "mktemp_failed branch=zsh status=$?"
+      echo "crow-shell-wrapper: mktemp -d failed for zsh ZDOTDIR" >&2
+      exit 73
+    fi
+    crow_log "zdotdir_temp_created path=$ZTMP"
     cat > "$ZTMP/.zshrc" <<'ZRC'
+# Helper redefined inside the embedded rc — it runs in the new shell, so it
+# needs its own crow_log. Same file the wrapper appends to (issue #256).
+crow_log() { printf '%s %s\n' "$(date -u +%Y-%m-%dT%H:%M:%SZ)" "$*" >> "${CROW_WRAPPER_LOG:-/dev/null}" 2>/dev/null || true; }
+
 # Source the user's real config from their original ZDOTDIR (or $HOME).
 if [ -n "${CROW_USER_ZDOTDIR:-}" ]; then
   ZDOTDIR="$CROW_USER_ZDOTDIR"
 else
   ZDOTDIR="$HOME"
 fi
-[ -f "$ZDOTDIR/.zshrc" ] && source "$ZDOTDIR/.zshrc"
+if [ -f "$ZDOTDIR/.zshrc" ]; then
+  source "$ZDOTDIR/.zshrc"
+  crow_log "user_rc_sourced rc=$ZDOTDIR/.zshrc"
+else
+  crow_log "user_rc_skipped reason=missing rc=$ZDOTDIR/.zshrc"
+fi
 
 _crow_precmd() {
+  crow_log "precmd_fired"
   if [ -n "${TMUX:-}" ]; then
     printf '\033Ptmux;\033\033]133;A\007\033\\'
     printf '\033Ptmux;\033\033]9;crow-ready\007\033\\'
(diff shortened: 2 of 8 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None found.

### Security Review

**Strengths:**
- `CROW_SENTINEL` is required at startup — early exit (64) on missing env prevents the wrapper from running in an undefined state
- `mktemp` failures are caught with a hard `exit 73` instead of silently falling through to a broken environment
- Per-section output capping (`sectionCap = 8_192`) in `captureDiagnostics` prevents unbounded memory/clipboard usage
- `runShortCommand` has a 2s timeout so any single subprocess hanging can't wedge the main actor
- Wrapper log path uses `$TMPDIR` (not hardcoded `/tmp`), matching the sandbox expectations documented in CLAUDE.md
- The `crow_log()` function swallows its own write failures (`2>/dev/null || true`) so logging itself never breaks the wrapper
- `CROW_WRAPPER_DEBUG` xtrace is opt-in only, preventing accidental log exposure of env vars in normal runs

**Concerns:**
- Minor: `captureDiagnostics` calls `dscl . -read /Users/$USER UserShell` and includes `PATH`, `SHELL`, and `USER` in the diagnostic bundle that goes to clipboard and NSLog. This is appropriate for developer diagnostics but the `PATH` may contain non-standard paths that reveal workspace structure. Acceptable for a debugging tool intended for developer use — just noting it.

### Code Quality

**Shell wrapper (`crow-shell-wrapper.sh`):**
- Clean stage breadcrumb design — each transition logs a unique keyword (`start`, `zdotdir_temp_created`, `user_rc_sourced`, `hook_installed`, `precmd_fired`, `sentinel_written`), making grep-based diagnosis trivial
- `crow_log() { :; }` self-disarm after first prompt is clever — prevents log bloat on every subsequent prompt while keeping the diagnostic chain intact for the critical first-prompt path
- Belt-and-braces `precmd_functions+=(_crow_precmd)` after `add-zsh-hook` is a pragmatic defense against deferred plugin reassignment — and `add-zsh-hook` deduplicates, so no double-fire risk
- The `$SHELL` vs `dscl` cross-check at startup is a nice diagnostic aid for chsh mismatches
- The bash path log line says `zdotdir_temp_created` (line 137) — misleading since bash uses a temp rcfile, not a ZDOTDIR. Minor naming nit.

**Swift diagnostics (`TmuxBackend.captureDiagnostics`):**
- Well-structured multi-section bundle with graceful fallback (`<failed>`, `<no pane pid available>`) for each section that can independently fail
- `truncated(_:max:)` uses `String.count` (Character count, not byte count) for the cap — fine for clipboard/log sizing, just noting the semantic difference from the `sectionCap` name
- The progress beacon task (`progressTask`) in `startReadinessWatch` correctly cancels on waiter resolution — no leak
- Good use of `[weak self]` in both the progress task and the readiness watch task to avoid retain cycles

**Tests (`CrowShellWrapperTests.swift`):**
- Three well-stratified test cases: deterministic setup-phase (fake shell), real-zsh runtime (gated on `/bin/zsh` availability), and sentinel failure path
- The runtime test handles the non-PTY limitation gracefully — asserts on what's guaranteed (setup breadcrumbs) and soft-checks what depends on interactive prompting
- Sentinel failure test correctly uses a truly impossible path (`/this/path/cannot/exist/`) rather than a permission-based approach that might be flaky on CI
- `DispatchWorkItem` timeout on the subprocess is a proper belt-and-braces for the interactive zsh case

**UI (`SessionDetailView.swift` / `ReadinessAwareTerminal`):**
- "Copy diagnostics" button with 1.5s "Copied!" feedback is clean UX — uses `DispatchQueue.main.asyncAfter` which is fine for a one-shot UI animation
- The `@State private var diagnosticsCopied` correctly resets even if the user rapidly double-clicks (the latest dispatch wins, resetting to false)

**Wiring (`AppDelegate` / `SessionService`):**
- `onCopyDiagnostics` closure wiring follows the same pattern as existing `onRetryReadiness` — consistent
- `copyDiagnostics` correctly logs the byte count of the bundle for post-mortem size analysis

### Summary Table
| Priority | Issue |
|----------|-------|
| Green | Bash log line "zdotdir_temp_created" at line 137 of wrapper is misleading for the bash code path (uses a temp rcfile, not a ZDOTDIR) |
| Green | Consider noting in the diagnostic bundle header that PATH contents are included (aids awareness when pasting into public issue trackers) |

**Recommendation:** Approve — this is a well-executed diagnostics feature. The shell wrapper breadcrumbs are cleanly designed, the Swift-side capture is robust with per-section fault isolation and size capping, the UI affordance is minimal and effective, and the test coverage is pragmatic. The two green items are minor polish that can be addressed in a follow-up.

---

PR #260, opened 2026-05-13
Scrub host-terminal env from tmux server

### LINKED ISSUE #259: tmux readiness: scrub host-terminal env contamination from tmux server
Follow-up to #256 / PR #258. The diagnostics from #258 captured a real failing case and the reporter's lsof + stack-sample investigation pinned the root cause: **the tmux server inherits host-terminal env vars (cmux, Ghostty, etc.) from CrowApp, and those leak into every shell window — triggering an oh-my-zsh `$(…)` substitution that deadlocks**.

This ticket implements the env-scrub fix.

## Root cause (confirmed, not speculation)

On a user with cmux installed:
- CrowApp was launched from cmux, inheriting `CMUX_*`, `TERM_PROGRAM=ghostty`, `TERMINFO=/Applications/cmux.app/…/terminfo`, etc.
- Crow's tmux server inherits CrowApp's env, then propagates it to every tmux window's child shell.
- In the shell: our injected `.zshrc` runs → sources user's `.zshrc` → sources oh-my-zsh, which sees `CMUX_LOAD_GHOSTTY_ZSH_INTEGRATION=1` and triggers a `$(…)` command substitution path. The child of that substitution blocks; the parent zsh deadlocks reading from the pipe. Sentinel never written. `.timedOut` fires.

`lsof` on the deadlocked child confirms the file chain (injected zshrc → user zshrc → oh-my-zsh.sh) loaded correctly. Stack sample ends in zsh's `getoutput` (`$(…)` handler) inside oh-my-zsh init. The same wrapper + tmux conf runs cleanly in a `tmux -f crow-tmux.conf` server *not* launched by CrowApp, ruling out the wrapper and conf.

See https://github.com/radiusmethod/crow/issues/259#issuecomment-4436418954 for the full investigation transcript.

## Fix

Scrub host-terminal env contamination from the tmux server's global environment at server-creation time. Single call site: `TmuxBackend.swift::ensureRunningServer`, immediately after `ctrl.newSessionDetached(…)` succeeds.

Approach: run `tmux set-environment -gu <var>` for each contaminating var. The `-g` makes it global; `-u` unsets it. Affects new windows from that point on; existing window envs are unchanged (we're about to create the first window, so this is fine).

Contaminating var families to scrub:
- `CMUX_*` — cmux Claude tooling layer (the proximate cause in this report)
- `GHOSTTY_*` and `TERM_PROGRAM=ghostty` — Ghostty integration vars sometimes embed scripts that re-exec
- `ITERM_*`, `ITERM_PROFILE`, `ITERM_SESSION_ID` — iTerm shell integration
- `KITTY_*` — kitty terminal integration
- `WEZTERM_*` — WezTerm integration
- `ALACRITTY_*` — Alacritty integration
- `TERM_PROGRAM`, `TERM_PROGRAM_VERSION` — generic host-terminal markers
- `TERMINFO` and `TERMINFO_DIRS` — but only when they point outside the system. If the path is `/usr/share/terminfo` or under `$XDG_DATA_HOME`, leave alone. Simplest safe heuristic: scrub unconditionally if set — tmux will fall back to the system default which is what we want inside the cockpit.

We can enumerate by prefix in code: read `ProcessInfo.processInfo.environment.keys`, filter by prefix matches against the above list, build a single batch of `set-environment -gu` calls (one tmux invocation per var; cheap, server-local). Alternatively send them via `tmux source-file -` with newline-separated commands — one round-trip — but the per-var path is simpler and the cost is negligible at server-start time.

Also: log the scrub to telemetry so future diagnostics show what was removed.

```
[CrowTelemetry tmux:env_scrubbed vars=CMUX_SHELL_INTEGRATION,CMUX_SHELL_INTEGRATION_DIR,…]
```

## Out of scope

- **Per-window env scrub**: the existing `newWindow(env:)` already only passes through what we explicitly set (`CROW_SENTINEL`, `CROW_WRAPPER_LOG`, `PWD`). The leak is at the server level, not the window level. No change needed.
- **Rosetta investigation**: the reporter observed that every shell under Crow's tmux runs under Rosetta (`/usr/libexec/rosetta/runtime`) even though tmux and CrowApp are arm64. Likely caused by arch inheritance from cmux (which is probably x86_64). Worth a separate ticket if the env-scrub fix doesn't fully resolve the hang, but env-scrub alone is expected to be sufficient since the same arch setup runs cleanly in a non-Crow tmux server.
- **Wrapper changes**: the diagnostics did their job; no wrapper modifications needed.

## Acceptance

- [ ] `TmuxBackend.ensureRunningServer` scrubs the listed env-var prefixes from the tmux server's global env via `set-environment -gu` after `newSessionDetached`.
- [ ] One NSLog telemetry line records which vars were scrubbed (operator visibility).
- [ ] Unit test in `TmuxBackendTests` that registers a terminal in a backend started with a synthetic contaminating env (`CMUX_FOO=1`) set in the parent process; verify `tmux show-environment -g` afterwards does not contain `CMUX_FOO`. (Use `setenv` in the test harness, restore in `defer`.)
- [ ] Manual verification with reporter: re-run the failing repro on a build with the fix; sentinel fires, `.shellReady` reached within budget.

## Related

- #256, #258 — diagnostics that surfaced this
- #252, #253 — `.timedOut` plumbing

dhilgaertner:
Closes #259

## Summary
- Crow's tmux server inherits `CMUX_*`, `GHOSTTY_*`, `TERM_PROGRAM`, `TERMINFO` and similar host-terminal markers from CrowApp's process env. These vars propagate to every child tmux window. In oh-my-zsh, `CMUX_LOAD_GHOSTTY_ZSH_INTEGRATION` triggers a `$(...)` command substitution whose child blocks; the parent zsh deadlocks on the pipe and the readiness sentinel never fires (timeout from #253).
- Fix: in `TmuxBackend.ensureRunningServer`, immediately after `newSessionDetached`, intersect `ProcessInfo.processInfo.environment.keys` with a known set of host-terminal var families and call `tmux set-environment -g -u <name>` for each match. The session anchor (`tail -f /dev/null`) is unaffected and no user-visible window has been created yet, so the global scrub is safe.
- Adds two thin wrappers (`unsetGlobalEnv`, `showGlobalEnv`) on `TmuxController` so the backend doesn't poke at raw arg arrays and tests have a typed entry point. One telemetry NSLog records the scrubbed list.

Var families scrubbed (prefix match): `CMUX_`, `GHOSTTY_`, `ITERM_`, `KITTY_`, `WEZTERM_`, `ALACRITTY_`
Exact keys: `TERM_PROGRAM`, `TERM_PROGRAM_VERSION`, `TERMINFO`, `TERMINFO_DIRS`

## Test plan
- [x] `swift test --package-path Packages/CrowTerminal` — new `ensureRunningServerScrubsContaminatingEnvVars` plants `CMUX_FOO=1` via `setenv`, starts the backend, and asserts the unscrubbed assignment doesn't appear in `tmux show-environment -g`.
- [ ] Manual: launch Crow from a Ghostty/cmux terminal that exports `CMUX_*`. Confirm in `Console.app` that `[CrowTelemetry tmux:env_scrubbed vars=…]` lists the `CMUX_*` family and the shell prompt appears within the readiness budget.
- [ ] Reporter (@danny) re-runs the original repro on a build with this fix.

## Out of scope
- Per-window env scrub — `newWindow(env:)` already only passes through `CROW_SENTINEL`, `CROW_WRAPPER_LOG`, `PWD`. Leak is at the server level.
- Rosetta investigation — separate ticket if env scrub doesn't fully resolve the deadlock.

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -348,7 +348,65 @@ public final class TmuxBackend {
             command: "/usr/bin/tail -f /dev/null"
         )
         controller = ctrl
+        scrubInheritedHostTerminalEnv(ctrl: ctrl)
         return ctrl
+    }
+
+    /// Host-terminal env-var families that must not propagate from the
+    /// CrowApp process into the tmux server's global env. When inherited,
+    /// these trigger oh-my-zsh integration code paths (notably
+    /// `CMUX_LOAD_GHOSTTY_ZSH_INTEGRATION`) whose `$(...)` subshell
+    /// deadlocks the shell that backs each tmux window — the readiness
+    /// sentinel never fires and the watch reports `.timedOut`. See #259.
+    private static let hostTerminalEnvPrefixes: [String] = [
+        "CMUX_",
+        "GHOSTTY_",
+        "ITERM_",
+        "KITTY_",
+        "WEZTERM_",
+        "ALACRITTY_",
+    ]
+
+    /// Single-key host-terminal markers. `TERMINFO`/`TERMINFO_DIRS` are
+    /// scrubbed unconditionally if present — tmux falls back to the
+    /// system terminfo, which is what the cockpit shells expect.
+    private static let hostTerminalEnvExact: Set<String> = [
+        "TERM_PROGRAM",
+        "TERM_PROGRAM_VERSION",
+        "TERMINFO",
+        "TERMINFO_DIRS",
+    ]
+
+    /// Remove host-terminal contamination from the freshly-started tmux
+    /// server's global env. Runs once per server launch from
+    /// `ensureRunningServer` immediately after `newSessionDetached`. The
+    /// session anchor (`tail -f /dev/null`) doesn't care about env, and
+    /// no user-visible window has been created yet, so scrubbing here is
+    /// safe and affects every subsequent `new-window`.
+    private func scrubInheritedHostTerminalEnv(ctrl: TmuxController) {
+        let env = ProcessInfo.processInfo.environment
+        var toScrub: [String] = []
+        for key in env.keys {
+            if Self.hostTerminalEnvExact.contains(key) {
+                toScrub.append(key)
+                continue
+            }
+            if Self.hostTerminalEnvPrefixes.contains(where: { key.hasPrefix($0) }) {
+                toScrub.append(key)
+            }
+        }
+        guard !toScrub.isEmpty else { return }
+        toScrub.sort()
+        for name in toScrub {
+            do {
+                try ctrl.unsetGlobalEnv(name)
+            } catch {
+                // Best-effort: a single var we can't unset shouldn't wedge
+                // server startup. Log and keep going.
+                NSLog("[CrowTelemetry tmux:env_scrub_failed var=\(name) error=\"\(error)\"]")
+            }
+        }
+        NSLog("[CrowTelemetry tmux:env_scrubbed vars=\(toScrub.joined(separator: ","))]")
     }
 
     private func sentinelPath(for id: UUID) -> String {
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxController.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxController.swift
@@ -104,6 +104,19 @@ public struct TmuxController: Sendable {
         return out.split(separator: "\n").compactMap { Int($0.trimmingCharacters(in: .whitespaces)) }
     }
 
+    /// Mark `name` as removed in the server's global environment so new
+    /// windows do not inherit it from the env tmux was launched with
+    /// (issue #259).
+    public func unsetGlobalEnv(_ name: String) throws {
+        try run(["set-environment", "-g", "-u", name])
+    }
+
+    /// `tmux show-environment -g`. Returns the global env one entry per
+    /// line (`NAME=value` for set vars, `-NAME` for vars marked removed).
+    public func showGlobalEnv() throws -> String {
+        try run(["show-environment", "-g"])
+    }
+
     // MARK: - Windows
 
     public func newWindow(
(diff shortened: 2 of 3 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- The scrub list is sourced from `ProcessInfo.processInfo.environment.keys` — real process env keys, not user input — so there is no injection vector through `unsetGlobalEnv`.
- The `run()` path already passes args as an array (not shell-interpolated), and `set-environment -g -u` is a safe tmux subcommand.
- Best-effort error handling per-var (`catch` + log + continue) prevents a single failure from wedging server startup.

**Concerns:**
- None identified. The scrub surface is narrow and well-contained.

### Code Quality
- **Placement is correct**: the scrub runs immediately after `newSessionDetached` and before any user-visible window is created, so no race with window creation.
- **`controller = ctrl` before `scrubInheritedHostTerminalEnv`**: `controller` is set first at line 350, then scrub at 351. If `scrub` threw (it can't — errors are caught per-var), the controller would still be cached. This is the right order — a partial scrub is better than an uncached controller that causes `ensureRunningServer` to spin up a second tmux server.
- **Prefix list is well-chosen**: `CMUX_`, `GHOSTTY_`, `ITERM_`, `KITTY_`, `WEZTERM_`, `ALACRITTY_` covers the major terminal emulators that inject integration vars. Exact-match set for `TERM_PROGRAM`, `TERM_PROGRAM_VERSION`, `TERMINFO`, `TERMINFO_DIRS` is also correct — these are the standard markers.
- **Test (`ensureRunningServerScrubsContaminatingEnvVars`)**: Plants `CMUX_FOO` via `setenv`, starts the backend, then queries `show-environment -g` to confirm the assignment is gone. Clean use of `defer` for cleanup. The assertion `!env.contains("CMUX_FOO=1")` correctly allows either omission or `-CMUX_FOO` (tmux's "marked removed" form).
- **`TmuxController` wrappers**: `unsetGlobalEnv` and `showGlobalEnv` are thin typed wrappers matching the existing controller style. `showGlobalEnv` is used only in tests, which is fine — it keeps the raw tmux arg assembly out of test code.
- Telemetry log line (`tmux:env_scrubbed vars=...`) follows the existing `[CrowTelemetry ...]` convention and is grep-friendly.

### Summary Table
| Priority | Issue |
|----------|-------|
| Green | Could consider adding `TABBY_`, `HYPER_`, or `TERMINAL_` prefix families for completeness, but these are uncommon and the current list covers the known trigger (`CMUX_LOAD_GHOSTTY_ZSH_INTEGRATION`). Not blocking. |

**Recommendation:** Approve — clean, well-scoped fix for a real deadlock. Placement is safe, error handling is defensive, test coverage is solid, and the change is additive-only (0 deletions).

---

PR #263, opened 2026-05-13
fix(tmux): enable extended-keys so Shift+Enter inserts newline in Claude Code

### LINKED ISSUE #262: fix(tmux): enable extended-keys so Shift+Enter inserts newline in Claude Code
## Bug

In Crow sessions backed by tmux, **Shift+Enter submits instead of inserting a newline** in the Claude Code prompt. This makes multi-line messages painful (you have to type `\` then Enter, or Ctrl+J).

The behavior is inconsistent across sessions: older sessions like `ontology-prototype-1-build-prototype` accept Shift+Enter; newer ones like `codename-spotlight-145-chat-ui-research` (created today via `/crow-workspace`) do not. The likely explanation is that the older session predates the tmux-backend rollout (#198 / #229) and is still running in a directly-attached Ghostty pane, while newer sessions go through Crow's embedded tmux server and pick up the bundled `crow-tmux.conf`.

## Root cause

Per the [Claude Code terminal-config docs](https://code.claude.com/docs/en/terminal-config#configure-tmux), tmux needs three settings for the inner app to see Shift+Enter as a distinct key:

```bash
set -g allow-passthrough on
set -s extended-keys on
set -as terminal-features 'xterm*:extkeys'
```

`Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf` has only the first one. The other two are missing:

- **`extended-keys on`** — tells the tmux server to *accept* the CSI-u / xterm extended-key sequences from the outer terminal (Ghostty).
- **`terminal-features '...:extkeys'`** — tells the tmux server to *send* the same extended-key sequences to the inner pty (Claude Code).

There's a second, subtler issue: even if we add the `xterm*:extkeys` line verbatim from the docs, **it won't match** Crow's tmux because `crow-tmux.conf:20` sets `default-terminal screen-256color`. The `xterm*` glob doesn't match `screen-256color`. We need a pattern that does — either match the actual default-terminal value or use a wider glob.

## Fix

Add to `crow-tmux.conf` (no removals, additive only):

```diff
 # Required for OSC 133 / OSC 9 / cwd-reporting sequences to reach the
 # attached Ghostty surface. Without this, raw OSC bytes from the shell
 # wrapper are consumed by tmux's emulator. Must be set at server start.
 set -gs allow-passthrough on

+# Allow tmux to receive xterm extended-key sequences from the outer
+# terminal (Ghostty) and forward them to the inner pty. Required for
+# Shift+Enter, Ctrl+Enter, etc. to reach Claude Code as distinct keys
+# instead of bare Enter. Must be set at server start.
+# Ref: https://code.claude.com/docs/en/terminal-config#configure-tmux
+set -gs extended-keys on
+set -gas terminal-features 'screen*:extkeys'
+set -gas terminal-features 'xterm*:extkeys'
+
 # 256-color profile that Ghostty handles cleanly. xterm-256color works
 # but causes occasional underline-color glitches at libghostty's edge.
 set -gs default-terminal screen-256color
```

Two `terminal-features` lines because `default-terminal` is currently `screen-256color` — the `xterm*` pattern is kept for future-proofing if we ever switch back. Append (`-as`) rather than overwrite so we don't clobber other features. `-gs` so it sticks at the server level.

### Why not just change `default-terminal` to `xterm-256color`

The existing comment notes \"xterm-256color works but causes occasional underline-color glitches at libghostty's edge.\" Leaving `screen-256color` and broadening the pattern is the lower-risk change.

## Applying without a restart

`crow-tmux.conf` is loaded via `tmux -f <path>` at server start, so existing tmux servers won't pick up the change. Two options after merging:

1. **Restart Crow** — kills the tmux server, next launch picks up the new conf.
2. **Live reload via Swift** — `TmuxBackend.ensureRunningServer` could `set-environment`-style apply these three options to a running server too, so users hitting an upgrade don't need to restart. Cheap to do alongside #259's env-scrub work, but not required for this fix.

Suggest option 1 for the initial PR (smaller blast radius) and consider option 2 as a tiny follow-up if it's annoying in practice.

## Verification

- [ ] Open a new Crow session (post-fix build), launch Claude Code, press Shift+Enter in the prompt — newline inserted, not submit.
- [ ] Confirm the OSC passthrough still works (sentinel-based readiness from #258 should be unaffected — `allow-passthrough` is unchanged).
- [ ] Test with both fresh and resumed sessions to confirm the fix applies on server cold-start.
- [ ] Optional regression: confirm Ctrl+J and `\\` + Enter (the fallbacks documented in the same Claude Code docs) still work as alternatives.

## Out of scope

- Live-reloading the tmux conf on existing servers (small follow-up if needed).
- A user-level escape hatch at `~/.config/crow/tmux.conf` mentioned in the bundled conf's header comment — separate ticket.
- Switching `default-terminal` away from `screen-256color` — unrelated and noted as a known tradeoff in the existing comment.

## Refs

- Claude Code docs (the source of the recommended settings): https://code.claude.com/docs/en/terminal-config#configure-tmux
- Bundled conf: `Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf`
- Tmux backend rollout: #198 / #229
- Related diagnostics work: #256, #258, #259

dhilgaertner:
## Summary

- Adds `set -gs extended-keys on` to Crow's bundled tmux conf so the inner app sees CSI-u / modifyOtherKeys sequences and can distinguish **Shift+Enter** from plain **Enter**.
- Appends `set -gas terminal-features ',screen*:extkeys,xterm*:extkeys'`. Crow's `default-terminal` is `screen-256color`, so the Claude Code docs' literal `xterm*:extkeys` glob does **not** match on its own — `screen*` is the load-bearing entry; `xterm*` is kept for future-proofing.
- `-gas` appends to the server-scope list so tmux's built-in `terminal-features` merges are preserved; the leading comma keeps the append safe when the existing value is empty.
- Purely additive — no existing settings change, OSC passthrough / sentinel readiness from #258 are untouched.

Config is read via `tmux -f` at server start, so a Crow restart is required to pick up the change. Live-reload onto an already-running tmux server, a user-level `~/.config/crow/tmux.conf` escape hatch, and changing `default-terminal` away from `screen-256color` are explicitly out of scope.

## Test plan

- [ ] Quit any running Crow + `tmux kill-server` (or just relaunch Crow) so a fresh tmux server picks up the new conf.
- [ ] New Crow session → launch Claude Code → **Shift+Enter** inserts a newline (does not submit).
- [ ] Plain **Enter** still submits.
- [ ] **Ctrl+J** and **`\` + Enter** newline fallbacks still work (regression check).
- [ ] Sentinel-based shell readiness from #258 still fires; OSC passthrough unaffected.
- [ ] Verify on both a **resumed** session and a **server cold-start** session.

Closes #262

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf
@@ -39,3 +39,16 @@ set -gs history-limit 5000
 # No escape-key delay — interactive editors (vi-mode shells, etc.) feel
 # sluggish with tmux's default 500ms.
 set -gs escape-time 0
+
+# Emit CSI-u / modifyOtherKeys sequences so the inner app can distinguish
+# Shift+Enter from Enter (and other modified keys). Required by Claude
+# Code to insert a newline on Shift+Enter instead of submitting.
+set -gs extended-keys on
+
+# Advertise the `extkeys` capability for the terminal types we actually
+# run under. `default-terminal` above is `screen-256color`, so the docs'
+# literal `xterm*:extkeys` glob does NOT match on its own — `screen*` is
+# the load-bearing entry here. `-gas` appends to the server-level list so
+# tmux's built-in terminal-features merges are preserved; the leading
+# comma keeps the append safe when the existing value is empty.
+set -gas terminal-features ',screen*:extkeys,xterm*:extkeys'


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Purely additive config changes — no existing settings altered
- Static directives with no dynamic/user-supplied values — zero injection surface
- `allow-passthrough on` (pre-existing, not part of this PR) is the only security-sensitive tmux option in this file; this PR does not widen that surface

**Concerns:**
- None

### Code Quality
- **Correct terminal-type targeting**: `screen*:extkeys` correctly matches the `default-terminal screen-256color` set on line 20. The PR description and inline comments both call out that the Claude Code docs' `xterm*:extkeys` alone would silently fail — good attention to detail.
- **Safe append semantics**: `-gas` (global, append, server) preserves tmux's built-in `terminal-features` merges rather than overwriting. The leading comma ensures safe concatenation against an empty existing value.
- **Consistent flag style**: All directives use `-gs` (server-scope), matching the rest of the file.
- **Comments explain "why" not just "what"**: Each block documents the motivation and non-obvious interactions (e.g., why `screen*` is load-bearing, why `-gas` instead of `-s`).
- **Scope discipline**: PR description explicitly lists what's out of scope (live-reload, user-level override, changing `default-terminal`) — keeps the change minimal and reviewable.

### Summary Table
| Priority | Issue |
|----------|-------|
| Green | `xterm*:extkeys` is kept for future-proofing; could be dropped since `default-terminal` is `screen-256color`, but the cost is zero and it's documented |

**Recommendation:** Approve — clean, well-documented, single-purpose change with no security or correctness concerns.

---

PR #267, opened 2026-05-14
fix(tmux): submit quick action prompts via Enter keycode after paste

### LINKED ISSUE #264: fix(tmux): quick action prompts not submitted — trailing \n inside bracketed paste
## Bug

When clicking a quick action button (Merge PR, Address Changes, Fix Checks, Fix Conflicts), the prompt text is pasted into Claude Code's input but never submitted. Only affects the tmux backend.

## Root Cause

`TmuxBackend.sendText()` uses `paste-buffer` to send text. When Claude Code has bracketed paste mode enabled, tmux wraps pasted content in `\e[200~`...`\e[201~` escape sequences. The trailing `\n` inside the bracket is treated as literal text content, not as Enter. The Ghostty path works because it sends a synthetic Return keycode (36) outside any paste bracket.

## Fix

Strip the trailing `\n` from the text before pasting via `paste-buffer`, then send a separate `Enter` key via `tmux send-keys Enter` after the paste completes.

### Files to modify

1. **`Packages/CrowTerminal/Sources/CrowTerminal/TmuxController.swift`**
   - Add `sendKeys(target:keys:)` method that runs `tmux send-keys <key> -t <target>`

2. **`Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift`** (~line 244, `sendText`)
   - Strip trailing `\n` from text before loading into paste buffer
   - After `pasteBuffer`, call `ctrl.sendKeys(target:keys:["Enter"])` if the original text ended with `\n`

dhilgaertner:
## Summary

Quick action prompts (Merge PR, Address Changes, Fix Checks, Fix Conflicts) end with a trailing \`\n\` so they auto-submit in Claude Code's input. The Ghostty backend honored that contract by splitting on \`\n\` and emitting a synthetic Return keycode (36) — those events fire outside any bracketed-paste bracket, so the TUI saw them as real Enter keys.

The tmux backend shipped the entire string through \`tmux load-buffer\` + \`tmux paste-buffer\`. Claude Code's TUI enables bracketed-paste mode, so tmux wrapped the buffer in \`\e[200~ … \e[201~\` and the trailing \`\n\` was delivered as literal text — typed as a newline character in the input field, never as Enter. Prompt visible, never submitted.

## Changes

- \`TmuxController.sendKeys(target:keys:)\` — thin wrapper over \`tmux send-keys -t <target> <keys...>\` reusing the existing \`run\` watchdog/timeout path.
- \`TmuxBackend.sendText\` — strip the single trailing \`\n\` from the paste payload, then send a separate \`Enter\` via \`sendKeys\` *after* \`paste-buffer\` completes (outside the bracket). Internal newlines are preserved as-is; an empty-after-strip payload skips the paste entirely (some tmux builds reject empty \`load-buffer\` input).

Ghostty path is untouched.

## Test plan

- [x] \`make build\` — green
- [x] \`swift test\` for the CrowTerminal package — all 24 tests pass, including \`TmuxBackendTests.sendTextRoundTripsThroughBuffer\`
- [ ] Manual: with \`experimentalTmuxBackend: true\`, click **Fix Conflicts** / **Merge PR** on a PR session and confirm the prompt is submitted automatically
- [ ] Manual regression: same flow with \`experimentalTmuxBackend: false\` (Ghostty) — behavior unchanged

Closes #264

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -241,16 +241,33 @@ public final class TmuxBackend {
     /// Send text to `id`'s window via the buffer-paste path. Works for
     /// arbitrary-size payloads (Phase 3 §3 finding: send-keys -l fails
     /// on >10KB; load-buffer + paste-buffer scales to 50KB+ in 133ms).
+    ///
+    /// Quirk: Claude Code's TUI enables bracketed-paste mode, which wraps
+    /// `paste-buffer` output in `\e[200~…\e[201~`. A trailing `\n` inside the
+    /// bracket is treated as literal text, not as Enter — so prompts that
+    /// rely on `\n` to submit (quick actions: Merge PR, Fix Conflicts, …) get
+    /// pasted but never submitted (#264). Strip the trailing newline before
+    /// pasting and deliver a separate `Enter` via `send-keys` afterwards,
+    /// mirroring what `GhosttySurfaceView.writeText` does with keycode 36.
     public func sendText(id: UUID, text: String) throws {
         guard let windowIndex = bindings[id] else {
             throw TmuxBackendError.unknownTerminal(id)
         }
         do {
             let ctrl = try ensureRunningServer()
-            let bufferName = "crow-\(id.uuidString)"
-            try ctrl.loadBufferFromStdin(name: bufferName, data: Data(text.utf8))
-            defer { ctrl.deleteBuffer(name: bufferName) }
-            try ctrl.pasteBuffer(name: bufferName, target: "\(ctrl.sessionName):\(windowIndex)")
+            let target = "\(ctrl.sessionName):\(windowIndex)"
+            let endsWithNewline = text.hasSuffix("\n")
+            let payload = endsWithNewline ? String(text.dropLast()) : text
+
+            if !payload.isEmpty {
+                let bufferName = "crow-\(id.uuidString)"
+                try ctrl.loadBufferFromStdin(name: bufferName, data: Data(payload.utf8))
+                defer { ctrl.deleteBuffer(name: bufferName) }
+                try ctrl.pasteBuffer(name: bufferName, target: target)
+            }
+            if endsWithNewline {
+                try ctrl.sendKeys(target: target, keys: ["Enter"])
+            }
         } catch {
             reportIfTimeout(error)
             throw error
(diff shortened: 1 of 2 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No user-controlled data reaches shell execution — `sendKeys` passes `["Enter"]` as a literal tmux key name, not raw shell input.
- Buffer names are derived from UUID, no injection vector via `tmux load-buffer` / `paste-buffer`.
- The existing `run()` watchdog/timeout path is correctly reused for the new `sendKeys` method, so a hung tmux server can't block the main actor.

**Concerns:**
- None identified. The change stays within the existing trust boundary (Crow → tmux IPC over a private socket).

### Code Quality

**Well done:**
- The doc comment on `sendText` (TmuxBackend.swift:244-251) clearly explains the bracketed-paste root cause and references the Ghostty parity (`keycode 36`). Excellent for future maintainers.
- Empty-payload guard (`if !payload.isEmpty`) correctly avoids sending an empty `load-buffer`, which some tmux builds reject — good defensive coding.
- `sendKeys` in TmuxController is a thin, general-purpose wrapper (`send-keys -t <target> <keys...>`) that will be reusable for future key-event needs (e.g., `C-c`, `Escape`).
- The `defer { ctrl.deleteBuffer(...) }` stays inside the `if !payload.isEmpty` block, so the buffer is only cleaned up when it was actually created.

**Observations (non-blocking):**
- `sendText` only strips a **single** trailing `\n` (`text.dropLast()`). If a caller sends `"hello\n\n"`, only one `\n` is stripped and the other is pasted as literal text inside the bracket. This matches the Ghostty backend's behavior (which splits on every `\n` and sends a keycode-36 for each), so multi-newline strings will behave differently between backends. Not a concern for the quick-action use case (always exactly one trailing `\n`), but worth noting if `sendText` is later used for multi-line submission. Current callers at `registerTerminal` (line 183: `command + "\n"`) and `crow send` all append a single `\n`, so this is fine today.
- The existing test `sendTextRoundTripsThroughBuffer` (TmuxBackendTests.swift:103) sends a payload **without** a trailing newline, so it exercises the paste-only path but not the new `Enter`-after-paste path. A test case like `sendText(id: id, text: payload + "\n")` would cover the new branch. Low risk given the simplicity, but would strengthen regression coverage.

### Summary Table
| Priority | Issue |
|----------|-------|
| Green | Consider adding a test case for the trailing-newline + Enter path |
| Green | Document multi-newline behavior difference vs. Ghostty if `sendText` scope widens |

**Recommendation:** Approve — clean, focused fix that correctly addresses the bracketed-paste issue (#264). The approach mirrors what the Ghostty backend already does, the code is well-commented, and there are no security concerns.

PR #271, opened 2026-05-15
feat: per-workspace custom instructions appended to initial prompt

### LINKED ISSUE #265: feat: per-workspace custom instructions appended to initial prompt
## Feature

Add a `customInstructions` text field to each workspace configuration. When a session is set up via `/crow-workspace`, these instructions get appended to the initial `/plan` prompt that Claude Code receives.

**Example use case:** "Always run `npm test` before committing" or "Use the existing auth middleware in `src/middleware/auth.ts` as a pattern."

### Model — `WorkspaceInfo` in `AppConfig.swift`
- Add `customInstructions: String?` field
- Backward-compatible decoding (defaults to `nil`)

### UI — `WorkspaceFormView.swift`
- Add a `TextEditor` for multi-line custom instructions
- Pass through to `WorkspaceInfo` on save

### Prompt generation — `ClaudeLauncher.swift`
- Add `customInstructions: String?` parameter to `generatePrompt()`
- If non-nil/non-empty, append a `## Custom Instructions` section after the standard instructions

### Skill — `SKILL.md`
- Document `customInstructions` in the config format section
- Update the prompt template to show where custom instructions appear

### Data flow
The `/crow-workspace` skill reads `config.json` → finds the workspace → the Manager's Claude Code reads `customInstructions` from the workspace config → writes it into the prompt file → `setup.sh` passes it to Claude Code.

### Tests
- Add `ClaudeLauncher` test for `generatePrompt` with `customInstructions`

dgershman:
## Summary

- Adds `customInstructions: String?` field to `WorkspaceInfo` with backward-compatible decoding (defaults to `nil`)
- Adds `TextEditor` in workspace settings UI for editing multi-line custom instructions
- Appends `## Custom Instructions` section to generated prompts when the field is non-empty (both in `ClaudeLauncher.generatePrompt()` and the SKILL.md prompt template)

Closes #265

## Test plan

- [x] `swift test --package-path Packages/CrowCore` — 162 tests pass (including 2 new: round-trip + backward compat)
- [x] `swift test --package-path Packages/CrowClaude` — 13 tests pass (including 3 new: with/without/empty instructions)
- [ ] Manual: open Settings → edit workspace → verify TextEditor appears, saves, and persists
- [ ] Manual: `/crow-workspace` with custom instructions set → verify prompt file contains `## Custom Instructions`

### DIFF
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeLauncher.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeLauncher.swift
@@ -10,7 +10,8 @@ public actor ClaudeLauncher {
         session: Session,
         worktrees: [SessionWorktree],
         ticketURL: String?,
-        provider: Provider?
+        provider: Provider?,
+        customInstructions: String? = nil
     ) -> String {
         var lines: [String] = []
         lines.append("/plan")
@@ -64,6 +65,14 @@ public actor ClaudeLauncher {
                 ticketNumber: session.ticketNumber,
                 hasTicket: ticketURL != nil
             )
+        }
+
+        if let instructions = customInstructions,
+           !instructions.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
+            lines.append("")
+            lines.append("## Custom Instructions")
+            lines.append("")
+            lines.append(instructions)
         }
 
         return lines.joined(separator: "\n")
--- a/Packages/CrowClaude/Tests/CrowClaudeTests/ClaudeLauncherTests.swift
+++ b/Packages/CrowClaude/Tests/CrowClaudeTests/ClaudeLauncherTests.swift
@@ -192,6 +192,54 @@ import Testing
     #expect(command.contains("'\\''"))
 }
 
+// MARK: - Custom instructions
+
+@Test func generatePromptWithCustomInstructions() async {
+    let launcher = ClaudeLauncher()
+    let session = Session(name: "test-session")
+
+    let prompt = await launcher.generatePrompt(
+        session: session,
+        worktrees: [],
+        ticketURL: nil,
+        provider: nil,
+        customInstructions: "Always run npm test before committing"
+    )
+
+    #expect(prompt.contains("## Custom Instructions"))
+    #expect(prompt.contains("Always run npm test before committing"))
+}
+
+@Test func generatePromptWithoutCustomInstructions() async {
+    let launcher = ClaudeLauncher()
+    let session = Session(name: "test-session")
+
+    let prompt = await launcher.generatePrompt(
+        session: session,
+        worktrees: [],
+        ticketURL: nil,
+        provider: nil,
+        customInstructions: nil
+    )
+
+    #expect(!prompt.contains("## Custom Instructions"))
+}
+
+@Test func generatePromptWithEmptyCustomInstructions() async {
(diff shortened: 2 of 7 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review

**Strengths:**
- `WorkspaceInfo.init(from:)` uses `decodeIfPresent` for `customInstructions`, so malformed/legacy configs default to `nil` rather than throwing — verified by `workspaceCustomInstructionsDefaultsNilWhenKeyMissing`.
- The trimming guard in `ClaudeLauncher.generatePrompt` (`Packages/CrowClaude/Sources/CrowClaude/ClaudeLauncher.swift:70-76`) prevents whitespace-only values from emitting an empty `## Custom Instructions` header.
- The string is rendered as Markdown into a prompt file — no shell/command execution path, so no command injection vector.
- The UI trims on save (`WorkspaceFormView.swift:110-123`), keeping persisted state clean.

**Concerns (low severity):**
- **Self-controlled prompt injection.** `customInstructions` is inserted verbatim into the generated prompt. A user (or anyone who can edit `~/Library/Application Support/crow/config.json`) could insert content like `## Instructions\n1. ...` to override or append fake instructions. This is a self-inflicted risk on a local config file, but worth noting if workspace configs are ever shared/imported from untrusted sources. No mitigation needed today.
- **No length cap on the TextEditor.** `customInstructionsText` can be arbitrarily long, and the full text is appended to every session prompt in the workspace. A 100KB paste here will bloat every future prompt silently. Consider a soft character counter or a reasonable cap (e.g., 4–8KB) before merging if you want defense-in-depth, but it's not blocking.

### Code Quality

- **`generatePrompt` is currently test-only in production.** `Sources/Crow/App/SessionService.swift:343` (`launchClaude`) doesn't call `ClaudeLauncher.generatePrompt`; the real prompt flow is the LLM interpreting `skills/crow-workspace/SKILL.md`. The Swift change keeps the two implementations consistent, which is good — just flagging that the new Swift code path isn't exercised at runtime today. The SKILL.md template update is the load-bearing change.
- **SKILL.md template substitution is LLM-driven** (`skills/crow-workspace/SKILL.md:345-348`). The placeholder `{workspace customInstructions text — include verbatim}` plus the prose instruction below it relies on the model correctly omitting the section when the field is empty/null. Consistent with how the rest of the skill template works; pattern is acceptable.
- Frame height bump 400 → 520 in `WorkspaceFormView.swift:131` accommodates the new editor — looks reasonable.
- `Equatable` and `Codable` synthesis update cleanly with the new field.

### Tests
- 162 CrowCore tests pass (including 2 new round-trip + backward-compat).
- 13 CrowClaude tests pass (including 3 new: with/without/empty whitespace instructions).
- Coverage of the three meaningful states (set / nil / whitespace) is appropriate.

### Summary Table
| Priority | Issue |
|----------|-------|
| Yellow | Consider a soft length cap or character counter on the `TextEditor` — no current upper bound on `customInstructions` length |
| Green | `generatePrompt` is test-only in production; the runtime path is via the SKILL.md template. Worth flagging in case the parallel implementation drifts in the future |
| Green | Document that `customInstructions` is verbatim-injected into the prompt so users understand the security/UX implications |

**Recommendation:** Approve. Implementation is clean, backward-compatible, well-tested, and the only concerns are minor hardening suggestions rather than blockers.

---

PR #284, opened 2026-05-18
fix: harden terminal teardown races on tab close (#282)

### LINKED ISSUE #282: Adding/switching tabs periodically crashes Crow with no console output
## Description
Adding tabs and switching between them for additional terminals periodically crashes Crow. No console messages are emitted indicating the cause of the crash.

## Repro
1. Open a Crow session with at least one terminal tab
2. Add additional tabs (terminals) and switch between them repeatedly
3. After some number of add/switch cycles, Crow crashes

## Actual
- Crow crashes silently
- No console output / log messages indicating the cause

## Expected
- Tabs can be added and switched between without crashing
- Any crash should leave a diagnostic trail (crash log, console output, or stderr)

## Notes / Next steps
- Need to capture a crash log from `~/Library/Logs/DiagnosticReports/` next time it reproduces
- Consider running Crow from terminal to capture stderr
- Check for unbalanced retain/release on terminal/tab teardown, or AppKit notifications firing on a deallocated observer

dgershman:
Closes #282

## Summary

Closes two teardown races that match the bug report's "unbalanced retain/release on terminal/tab teardown" hypothesis.

- **`SessionService.closeTerminal` / `closeGlobalTerminal`** previously called `TerminalRouter.destroy(terminal)` *before* mutating `appState.terminals`. Result: `ghostty_surface_free` (or `tmux kill-window`) ran while the `GhosttySurfaceView` was still in the SwiftUI view hierarchy. Any Metal redraw, mouse-tracking, or input callback in that window operated on a freed surface pointer. Fix: mutate `appState` first so SwiftUI's render pass detaches the view; dispatch the backing destroy to the next main-actor tick.
- **`TmuxBackend.startReadinessWatch`** spawns a 10s progress beacon and a 30s sentinel waiter as `Task { … }` blocks. `destroyTerminal(id:)` removed the binding but never cancelled the Tasks. After the user closed a tab, the waiter could still fire `onReadinessChanged` ~29s later for a stale id. Fix: track the Task handles in `readinessTasks: [UUID: [Task<Void, Never>]]`, cancel them in `destroyTerminal`, and gate the post-wait callback on `bindings[id] != nil` so a Task that wins the cancel race bails.

The existing `CrashReporter` (`Sources/Crow/App/CrashReporter.swift`) already installs POSIX signal handlers and redirects stderr to `~/.local/share/crow/crash-logs/`, so if the bug still reproduces after this change, the next iteration has a stack trace to work from.

## Test plan

- [ ] `swift build --arch arm64` succeeds (verified locally)
- [ ] Add 10+ shell tabs to a session via the "+" button, switch between them rapidly, then close them in arbitrary order — no crash
- [ ] Same flow on the Global Terminals row
- [ ] Close a tmux-backed tab within 30s of opening it; confirm no late `onReadinessChanged` fires (look for stale `[CrowTelemetry tmux:first_prompt_ms=…]` lines in the log)
- [ ] If a crash still reproduces, confirm a stack trace lands in `~/.local/share/crow/crash-logs/`

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/SentinelWaiter.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/SentinelWaiter.swift
@@ -38,10 +38,18 @@ public struct SentinelWaiter: Sendable {
         let nanosPerInterval = UInt64(pollInterval * 1_000_000_000)
         let fm = FileManager.default
         while Date() < deadline {
+            // Cancellation: TmuxBackend.destroyTerminal cancels the waiter
+            // Task when the user closes a tab. Without this check, `try?`
+            // below would swallow `CancellationError` and the loop would
+            // tighten into a `fileExists` spin until the deadline (#282
+            // review). Re-check before the sleep too in case cancellation
+            // arrives mid-iteration.
+            if Task.isCancelled { return nil }
             if fm.fileExists(atPath: sentinelPath) {
                 return Date().timeIntervalSince(start)
             }
             try? await Task.sleep(nanoseconds: nanosPerInterval)
+            if Task.isCancelled { return nil }
         }
         return nil
     }
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -112,6 +112,11 @@ public final class TmuxBackend {
         sharedSurface = nil
         controller = nil
         bindings.removeAll()
+        // Cancel any in-flight readiness watches so they don't keep polling
+        // (sentinel files are about to be unlinked) after server teardown.
+        // Mirrors the `destroyTerminal` cleanup (#282).
+        for tasks in readinessTasks.values { tasks.forEach { $0.cancel() } }
+        readinessTasks.removeAll()
         for path in sentinels.values {
             try? FileManager.default.removeItem(atPath: path)
         }
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -58,6 +58,11 @@ public final class TmuxBackend {
     /// on destroy. Issue #256.
     private var wrapperLogs: [UUID: String] = [:]
 
+    /// UUID → in-flight readiness watch Tasks (the 10s progress beacon and
+    /// the waiter). `destroyTerminal` cancels these so they don't fire
+    /// `onReadinessChanged` for a tab the user just closed. Issue #282.
+    private var readinessTasks: [UUID: [Task<Void, Never>]] = [:]
+
     /// Offscreen window the shared surface lives in until SwiftUI re-parents
     /// it into the visible UI. Same trick the Ghostty path uses for
     /// background surface creation.
@@ -294,6 +299,10 @@ public final class TmuxBackend {
             controller?.killWindow(index: windowIndex)
         }
         bindings.removeValue(forKey: id)
+        // Cancel in-flight readiness watch Tasks so a 30s waiter doesn't
+        // fire `onReadinessChanged` against a stale id long after the tab
+        // is gone (issue #282).
+        readinessTasks.removeValue(forKey: id)?.forEach { $0.cancel() }
         if let sentinelPath = sentinels.removeValue(forKey: id) {
             try? FileManager.default.removeItem(atPath: sentinelPath)
         }
@@ -417,7 +426,7 @@ public final class TmuxBackend {
                 NSLog("[CrowTelemetry tmux:first_prompt_progress terminal=\(id) elapsed_ms=\(elapsed) sentinel_exists=\(exists)]")
             }
         }
-        Task { [weak self] in
+        let waiterTask = Task { [weak self] in
             // 30s default budget (was 5s). On app restart with many managed
             // terminals hydrating concurrently, shell startup is CPU-contended
             // and the wrapper's first precmd may not fire within 5s. Callers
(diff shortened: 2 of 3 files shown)


dhilgaertner (commented):
## Code & Security Review

Both teardown races are well-diagnosed, and the fixes match the symptoms in #282. Below are a few observations — none are blocking, but the first one is worth a follow-up because this PR is the first thing that actually exercises the cancellation path it relies on.

### Concerns

**`SentinelWaiter.waitForPrompt` doesn't propagate cancellation — new path can hot-loop**
`Packages/CrowTerminal/Sources/CrowTerminal/SentinelWaiter.swift:31-47`

```swift
while Date() < deadline {
    if fm.fileExists(atPath: sentinelPath) { ... }
    try? await Task.sleep(nanoseconds: nanosPerInterval)
}
```

After `destroyTerminal` cancels `waiterTask`, every subsequent `Task.sleep` inside `waitForPrompt` throws `CancellationError` immediately. `try?` swallows it, the `Date() < deadline` predicate stays true, and the loop tightens into a `fileExists` spin until the deadline (30s default, 120s on retry). The sentinel file is also already unlinked at that point (`destroyTerminal` removes it), so every iteration is a wasted `stat`. The `bindings[id] \!= nil` guard correctly suppresses the post-await readiness emission, so no user-visible misbehavior — just CPU/syscall burn for up to the timeout window after each tab close.

Suggested fix in the waiter:

```swift
while Date() < deadline {
    if Task.isCancelled { return nil }
    if fm.fileExists(atPath: sentinelPath) { ... }
    try? await Task.sleep(nanoseconds: nanosPerInterval)
}
```

Or replace `try?` with `try` and let the `CancellationError` propagate. Either way fixes it.

**`shutdown()` doesn't cancel/clear `readinessTasks`**
`Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:106-123`

`shutdown()` clears `bindings`, `sentinels`, `wrapperLogs`, but not the new `readinessTasks`. The `bindings[id] \!= nil` guard saves correctness, but the in-flight tasks keep running (and the CPU-spin issue above compounds for every in-flight watch at shutdown). Adding the cancel+clear there mirrors the pattern in `destroyTerminal`:

```swift
for tasks in readinessTasks.values { tasks.forEach { $0.cancel() } }
readinessTasks.removeAll()
```

**`DispatchQueue.main.async` ordering vs SwiftUI render pass**
`Sources/Crow/App/SessionService.swift:1009-1016, 1065-1073`

The comment claims the dispatched block runs *after* SwiftUI has detached the view from the hierarchy. In practice that's usually true (state mutation → publisher fires → SwiftUI commits at the next CATransaction → main-queue blocks drain) but it's not strictly guaranteed by either Combine or DispatchQueue. If the bug reproduces, the next escalation is `DispatchQueue.main.async { DispatchQueue.main.async { TerminalRouter.destroy(terminal) } }` (two-tick defer past the SwiftUI commit), or moving the destroy off the close path entirely (deferred sweep on the next runloop quiesce). Worth a comment noting this is the conservative first attempt.

### Code Quality

- **No tests for the new cancellation path.** The existing `retryReadinessEmitsTimedOutWhenSentinelMissing` test covers the timeout case. Adding a sibling test that registers with `trackReadiness: true`, calls `destroyTerminal` before the watch resolves, sleeps past the budget, and asserts no `.shellReady`/`.timedOut` fired would lock this in. Not a blocker — the path is small and the guard is obvious — but it would catch a future regression.
- **Captured `[weak self]` is duplicated** on `waiterTask` (outer Task + inner `MainActor.run`). Functionally fine — the inner closure shadows the outer — but slightly noisy.
- **`readinessTasks[id, default: []].append(...)`** for the retry path is correct but accumulates references to resolved Tasks until `destroyTerminal`. Bounded (one retry budget per terminal × tasks-per-watch), not a leak in practice.

### Security Review

No new attack surface. Changes are lifecycle-only: no new I/O sources, no new external input, no auth/crypto changes. Sentinel and wrapper-log paths still go through `TMPDIR` and are per-UUID, so no path-traversal regressions.

### Build / Test

Couldn't run `swift build` locally — `Frameworks/GhosttyKit.xcframework` is missing its binary artifact in this checkout, which appears to be a workstation setup issue rather than a PR defect. PR description states the build was verified.

### Summary Table

| Priority | Issue |
|----------|-------|
| Yellow   | `SentinelWaiter.waitForPrompt` hot-loops after cancellation — first PR to actually trigger this |
| Yellow   | `shutdown()` doesn't cancel/clear `readinessTasks` |
| Green    | `DispatchQueue.main.async` is the conservative first attempt at the render-pass-vs-destroy ordering; may need a two-tick defer if the crash returns |
| Green    | No test covers the cancel-before-resolve path |

**Recommendation:** Comment. The two diagnoses are sound and the fixes are the right shape — the readiness-Task cancellation + `bindings[id] \!= nil` guard, and the state-first / deferred-destroy reorder. The waiter cancellation issue is a follow-up worth filing rather than a merge blocker, since the guard prevents user-visible misbehavior. Suggest filing a follow-up for the `SentinelWaiter` fix + `shutdown()` cleanup before this lands so the cancellation path stops being a latent CPU sink.

---

dhilgaertner (approved):
## Approval

Approving — the two fixes are the right shape for the diagnosed races:

- Mutate `appState.terminals` first, then defer `TerminalRouter.destroy(terminal)` to the next main-actor tick so SwiftUI's render pass detaches the view before the underlying surface is freed.
- Track readiness-watch `Task` handles in `readinessTasks`, cancel them in `destroyTerminal`, and gate the post-wait callback on `bindings[id] != nil` so a Task that wins the cancel race bails.

See [my prior comment](https://github.com/radiusmethod/crow/pull/284#pullrequestreview-) on this PR for two follow-up items that are worth filing but don't block merge:

1. `SentinelWaiter.waitForPrompt` doesn't propagate `Task.isCancelled`, so the new cancel path turns into a bounded `fileExists` hot-loop. The `bindings[id] != nil` guard prevents user-visible misbehavior; the CPU burn is the only cost.
2. `TmuxBackend.shutdown()` should also cancel/clear `readinessTasks` to mirror the per-id cleanup in `destroyTerminal`.

Security review: clean — lifecycle-only, no new I/O or auth surface.

---

PR #286, opened 2026-05-18
ui: readable label-tag pills in dark mode (WCAG AA)

### LINKED ISSUE #285: session screen: ticket label tags have insufficient contrast in dark mode
## Problem

The label/tag pills shown at the top of a session screen (under the session name + ticket title) are rendered with the raw GitHub label color as the pill background. In dark mode this produces near-illegible chips — the dark-yellow `medium` pill and the dark-green `multi-tenant` pill both have very low contrast against the near-black session header background, and the pill *text* (also a saturated near-color) compounds the problem.

## Repro

1. Open any session whose ticket has GitHub labels with saturated dark hues (e.g. `medium` → `#FBCA04`, `multi-tenant` → `#0E8A16`).
2. Switch the app to dark mode (or use auto with the system in dark).
3. The pills appear immediately under the session title — text is barely distinguishable from the pill fill, and the pills barely lift off the page background.

## Example

Session `corveil-624-handler-property-tests`. Two labels:
- `medium` — color `#FBCA04` (yellow). Renders as a muddy dark-yellow chip; text effectively invisible.
- `multi-tenant` — color `#0E8A16` (green). Renders as a dark-green chip on near-black; readable only if you lean in.

## What good looks like

Pick one of these patterns (GitHub itself uses #1 on `github.com` in dark mode, which is why the contrast issue is invisible until you mirror the colors into your own dark UI):

1. **Hue-as-accent, not as fill.** Use the label color for the border / text / a 10–15% alpha background tint; keep the chip background close to a neutral surface color (e.g. `#21262d` on dark, `#eaeef2` on light). This is what github.com does in dark mode and what makes their labels readable regardless of hue.
2. **Contrast-aware fill.** Keep the label color as the pill fill but compute the foreground text color to meet WCAG AA (≥4.5:1) against that fill — and similarly nudge the fill toward the surface if its contrast against the page background is below ~3:1.

(1) is the lower-risk fix and matches the upstream rendering users already expect from GitHub itself.

## Acceptance

- [ ] Every label pill rendered on the session screen meets WCAG AA contrast (≥4.5:1) between pill text and pill fill, AND ≥3:1 between pill fill and the surrounding header background — in both light and dark modes.
- [ ] Sample labels to verify: `medium` (#FBCA04), `multi-tenant` (#0E8A16), `high` (#D93F0B), and at least one near-white / near-black label color.
- [ ] No regressions in light mode — the current light-mode rendering is fine.

## Notes

- This is purely a presentation bug; the data (label name + color from `gh issue view --json labels`) is correct.
- Same pills appear in the sidebar session list (smaller); fix should cover both surfaces if they share a component.

(Reported from the manager session while reviewing the corveil-624 setup.)

dhilgaertner:
## Summary
- `LabelPillsView` (used by the session header at `SessionDetailView.swift:68` and the sidebar session row at `SessionListView.swift:577`) now renders chips with a neutral fill (`#21262D` dark / `#EAEEF2` light), and uses the GitHub label hue for the border and text after HSL saturation+lightness clamping so every hue clears WCAG AA against the fill.
- New helper `CorveilTheme.accentColor(for:in:)` plus minimal RGB↔HSL conversion. Clamps saturation to `[0.40, 0.85]` and lightness to `≥ 0.72` (dark) / `≤ 0.30` (light). The previous `contrastingTextColor` path was unreliable because it computed contrast against the opaque label color, not the 15%-alpha fill that was actually rendered.
- `muted` mode (used by ticket-board "done" cards) now reads as `textMuted` text + border on the neutral fill — the hue is dropped intentionally to deprioritize the chip visually.
- Adds a `#Preview` rendering five sample labels (`medium #FBCA04`, `multi-tenant #0E8A16`, `high #D93F0B`, near-white `#F5F5F5`, near-black `#111111`) in both `.dark` and `.light` color schemes so the fix is visually verifiable on every change.

## Test plan
- [x] `make build` clean
- [x] `swift test --package-path Packages/CrowCore` — 165 tests pass
- [ ] Open the `#Preview` in Xcode; verify all five labels are legible in both panels (including muted row).
- [ ] Run the app, open a session linked to a GitHub issue with `medium` / `multi-tenant` / `high` labels — chips legible on the dark session header.
- [ ] Confirm sidebar session row chips legible on both active (`bgCard`) and inactive (`bgSurface`) rows.

Closes #285

### DIFF
--- a/Packages/CrowUI/Sources/CrowUI/CorveilTheme.swift
+++ b/Packages/CrowUI/Sources/CrowUI/CorveilTheme.swift
@@ -65,6 +65,62 @@ extension CorveilTheme {
         let luminance = 0.2126 * linearize(r) + 0.7152 * linearize(g) + 0.0722 * linearize(b)
         return luminance > 0.179 ? Color(hex: 0x1A1D20) : .white
     }
+
+    /// Hue-preserving variant of `hexString` suitable for chip text + border
+    /// against the neutral label-pill surface in the given color scheme.
+    /// Saturation is clamped to a friendly band; lightness is clamped so any
+    /// hue clears WCAG AA against the neutral fill.
+    public static func accentColor(for hexString: String, in scheme: ColorScheme) -> Color {
+        let hex = hexString.trimmingCharacters(in: CharacterSet(charactersIn: "#"))
+        guard hex.count == 6, let value = UInt(hex, radix: 16) else {
+            return scheme == .dark ? .white : Color(hex: 0x1A1D20)
+        }
+        let r = Double((value >> 16) & 0xFF) / 255
+        let g = Double((value >> 8) & 0xFF) / 255
+        let b = Double(value & 0xFF) / 255
+
+        var (h, s, l) = rgbToHSL(r: r, g: g, b: b)
+        s = min(max(s, 0.40), 0.85)
+        l = scheme == .dark ? max(l, 0.72) : min(l, 0.30)
+
+        let (nr, ng, nb) = hslToRGB(h: h, s: s, l: l)
+        return Color(red: nr, green: ng, blue: nb)
+    }
+
+    private static func rgbToHSL(r: Double, g: Double, b: Double) -> (Double, Double, Double) {
+        let maxC = max(r, g, b)
+        let minC = min(r, g, b)
+        let l = (maxC + minC) / 2
+        guard maxC != minC else { return (0, 0, l) }
+        let d = maxC - minC
+        let s = l > 0.5 ? d / (2 - maxC - minC) : d / (maxC + minC)
+        let h: Double
+        switch maxC {
+        case r: h = (g - b) / d + (g < b ? 6 : 0)
+        case g: h = (b - r) / d + 2
+        default: h = (r - g) / d + 4
+        }
+        return (h / 6, s, l)
+    }
+
+    private static func hslToRGB(h: Double, s: Double, l: Double) -> (Double, Double, Double) {
+        guard s != 0 else { return (l, l, l) }
+        let q = l < 0.5 ? l * (1 + s) : l + s - l * s
+        let p = 2 * l - q
+        return (hue2rgb(p, q, h + 1.0 / 3),
+                hue2rgb(p, q, h),
+                hue2rgb(p, q, h - 1.0 / 3))
+    }
+
+    private static func hue2rgb(_ p: Double, _ q: Double, _ t0: Double) -> Double {
+        var t = t0
+        if t < 0 { t += 1 }
+        if t > 1 { t -= 1 }
+        if t < 1.0 / 6 { return p + (q - p) * 6 * t }
+        if t < 1.0 / 2 { return q }
+        if t < 2.0 / 3 { return p + (q - p) * (2.0 / 3 - t) * 6 }
+        return p
+    }
 }
 
 // MARK: - Status Color Extensions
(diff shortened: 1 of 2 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Pure UI/color math — no I/O, no network, no untrusted input parsing beyond a 6-char hex string with `guard let` validation (CorveilTheme.swift:75).
- Existing `hexString` validation pattern preserved; invalid input gracefully returns a sensible per-scheme fallback rather than crashing.

**Concerns:**
- None.

### Code Quality
**Strengths:**
- HSL conversion (CorveilTheme.swift:90-123) matches the standard formulation. Spot-checked against known palette colors (`#0E8A16`, `#0366D6`, near-white, near-black) and outputs land in expected ranges.
- The previous approach (computing contrast against an opaque label color while rendering against a 15%-alpha fill) was actually unsound — the new design fixes the underlying premise rather than tuning numbers.
- `LabelPillsView` now reads `colorScheme` from the environment so it responds correctly to light/dark switches, including the preview side-by-side.
- DEBUG-gated `#Preview` (CorveilTheme.swift:309-339) covering both schemes and the `muted` variant is a meaningful regression safety net for a visual-only change.
- Surgical diff — single file, no API breakage at the `LabelPillsView` call sites in `ReviewBoardView`, `SessionDetailView`, `SessionListView`, `TicketBoardView`.

**Suggestions (non-blocking):**
- `CorveilTheme.contrastingTextColor(for:)` (CorveilTheme.swift:53) is no longer called anywhere in-repo. Since it's `public`, leaving it for now is fine, but worth a follow-up to deprecate/remove if there are no external consumers.
- No unit tests were added for `rgbToHSL` / `hslToRGB` / `accentColor`. The math is pure and easily testable; adding cases for the canonical GitHub palette (especially the borderline hues — see below) would lock the WCAG guarantees in place against future refactors. CrowUITests already imports `@testable import CrowUI`, so the surface is reachable.
- The PR description states "every hue clears WCAG AA against the fill." For real GitHub label colors this holds in practice, but pure saturated blue (~h=240°) at the clamped HSL `L=0.72, S=0.85` against the dark fill `#21262D` lands at roughly **4.3:1**, just below the 4.5:1 normal-text AA threshold (10pt is below the "large text" cutoff, so 4.5 applies). Real palettes are off-axis enough that this isn't hit by anything I see in `samples`, but the claim isn't strictly universal. Possible nudges if you want strict universality: raise dark-mode lightness clamp to ~`0.78`, or bias the clamp by hue (boost blues slightly more since the blue channel contributes only ~7% of luminance). Light mode (`L≤0.30` against `#EAEEF2`) is comfortably above AA for all hues I checked.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | — |
| Yellow | Saturated-blue edge case lands ~4.3:1 in dark mode; PR's "every hue clears AA" claim is approximate. |
| Yellow | No unit tests for new HSL/accent math despite the WCAG-correctness focus. |
| Green | `contrastingTextColor` is now dead code in-repo; consider deprecating in a follow-up. |
| Green | Consider documenting the chosen clamp constants (`0.40/0.85`, `0.72/0.30`) inline so future tweaks know which knobs affect contrast vs. saturation. |

**Recommendation:** Approve. The fix replaces a fundamentally flawed contrast computation with a hue-preserving, scheme-aware accent that demonstrably works for the realistic GitHub palette and is visually verifiable via the bundled preview. The edge case noted above is narrow enough to land as-is and address in a small follow-up if desired.

---

PR #291, opened 2026-05-20
Fix re-review not triggering reviewer Crow after /refine

### LINKED ISSUE #290: Re-review doesn't trigger reviewer's Crow after /refine + re-request
## Problem

After the PR author runs `/refine` and re-requests a review, the reviewer's Crow instance does **not** auto-launch a new review session. The PR sits in the review queue but no kickoff fires.

## Root cause

Two compounding gates in the auto-review pipeline:

1. **`reviewSessionID == nil` guard is permanent.** `Sources/Crow/App/AppDelegate.swift:465` skips kickoff whenever the PR has a linked review session:
   ```swift
   guard request.reviewSessionID == nil, !autoReviewedIDs.contains(request.id), ... else { continue }
   ```
   `reviewSessionID` is set every refresh in `Sources/Crow/App/IssueTracker.swift:299-305` whenever an `.active` review session has a PR link matching the request URL. Once round 1 creates a session, the gate stays closed.

2. **Review sessions are only auto-completed on merge/close.** `autoCompleteFinishedReviews` / `decideReviewCompletions` in `IssueTracker.swift:1887-1914` only complete review sessions when the PR transitions to `MERGED` or `CLOSED`. No signal for "the reviewer's Claude finished reviewing", so the session lingers in `.active` indefinitely — keeping gate #1 closed forever.

Secondary (notification path): `previousReviewRequestIDs` in `IssueTracker.swift:319` only detects PRs that *re-enter* the `review-requested:@me` search. If the reviewer only commented (no formal review), the PR never left the set, so the notification also misses.

### A. Auto-complete review session when the reviewer submits a review

- Extend the consolidated GraphQL refresh in `IssueTracker.swift` (around line 599+) to include `reviews { author { login }, submittedAt }` on each review-search PR.
- Propagate the viewer's latest review timestamp into `ReviewRequest`.
- In `IssueTracker`, when a viewer-authored review with `submittedAt > session.createdAt` (or `> lastKickoffAt`) appears for an active review session, call `appState.onCompleteSession?(session.id)`.
- Result: completed sessions drop out of `appState.reviewSessions` (filtered in `Packages/CrowCore/Sources/CrowCore/Models/AppState.swift:318-319`), `reviewSessionID` goes nil on the next refresh, and the existing kickoff at `AppDelegate.swift:464-474` fires naturally for round 2.

### B. Head-SHA change as a fallback re-kick signal

- Add `headRefOid` to the review-search GraphQL payload; propagate through `ReviewRequest`.
- Add a persisted `lastReviewedHeadSha: String?` to `Session` (`Packages/CrowCore/Sources/CrowCore/Models/Session.swift`) with a decode default for back-compat with existing `state.json` files.
- Write the SHA in `SessionService` when a review session is created/kicked.
- Relax the kickoff guard in `AppDelegate.swift:465`: allow re-kick when `request.headRefOid != session.lastReviewedHeadSha`, even if `reviewSessionID` is non-nil.
- Covers: force-push without an explicit re-request, and the race window where round-2 commits land before A's review-submitted signal is observed.

## Critical files

- `Sources/Crow/App/AppDelegate.swift` — kickoff guard (444-474)
- `Sources/Crow/App/IssueTracker.swift` — GraphQL refresh, cross-ref, auto-complete (280-330, 1887-1914, GraphQL builder ~599+)
- `Sources/Crow/App/SessionService.swift` — persist `lastReviewedHeadSha` on review-session creation/kickoff
- `Packages/CrowCore/Sources/CrowCore/Models/Session.swift` — new field
- `ReviewRequest` model — carry `headRefOid` + viewer's latest review timestamp

## Verification

1. Unit-test the new "viewer submitted review after session start → complete" decision (mirror `decideReviewCompletions`).
2. E2E happy path:
   - Open PR → Crow auto-launches review session → submit review.
   - Confirm session moves to `.completed` on next 60s tick.
   - Author pushes fix + re-requests review.
   - Confirm a new review session kicks off.
3. Force-push path (B only):
   - Round 1, no formal review submitted; author force-pushes.
   - Confirm kickoff re-fires due to `headRefOid` mismatch.
4. Back-compat: load existing `~/.local/share/crow/state.json` with old Session records; confirm decode + first refresh populates `lastReviewedHeadSha`.

## Alternatives considered

- **Resume existing terminal via `crow send`** — preserves Claude context across rounds, can be layered on later as a UX refinement.
- **Search-set re-entry detection only** — misses comment-only review workflows; too weak standalone.

dgershman:
## Summary

Closes #290.

After `/refine` + re-request, the reviewer's Crow did **not** auto-launch a new review session. Two compounding gates both keyed on terminal PR state:

- `AppDelegate.enqueueReviewKickoff` skipped any request whose `reviewSessionID != nil`, and
- `decideReviewCompletions` only completed sessions on `MERGED` / `CLOSED`.

Result: the round-1 session lingered `.active` forever while the PR was still open, holding the kickoff gate closed.

Two signals now unblock the kickoff (both implemented — A is primary, B is a fallback that also handles force-push):

### A. Viewer-submitted review completes the session

- GraphQL `reviewPRs` block now selects `reviews(last:20) { author { login } submittedAt state }` + top-level `viewer { login }`.
- `parseReviewRequests` surfaces `viewerLastReviewedAt` (latest `APPROVED` / `CHANGES_REQUESTED` / `DISMISSED` timestamp authored by the viewer) on `ReviewRequest`.
- `decideReviewCompletions` completes any active review session whose viewer-verdict timestamp is later than `session.createdAt`.
- Next refresh sees `reviewSessionID == nil`, and the existing kickoff fires for round 2 naturally.

### B. Head-SHA advance — fallback re-kick

- `Session` persists `lastReviewedHeadSha` (set from `gh pr view --json …headRefOid` at session creation, decoded with a nil default for back-compat with existing `state.json`).
- The kickoff guard also fires when `request.headRefOid != session.lastReviewedHeadSha`. Catches force-push without an explicit re-request, and the race window where round-2 commits land before signal A is observed.
- For the B path, the stale session is auto-completed inline before the new kickoff so `reviewSessions` doesn't double up on the same PR.

### Idempotency

The in-flight auto-review dedupe set is now keyed on `(request.id, headRefOid)` so a stale entry from a completed-and-re-requested PR doesn't block a fresh kickoff. Previously it was keyed on `request.id` alone, which would have re-blocked round 2 even with the gate fix.

## Test plan

- [x] Four new `decideReviewCompletions` cases:
  - Viewer-submitted review after session start → completes.
  - Viewer-submitted review before session start → ignored (round-1 leftover).
  - No viewer verdict (only other-user or COMMENTED) → ignored.
  - Viewer-submitted review under partial fetch → still completes (rule 1 doesn't gate on `prDataComplete`).
- [x] `Session` back-compat decode test: legacy `state.json` without `lastReviewedHeadSha` decodes to nil.
- [x] All existing CrowCore + root tests pass (`swift test` + per-package).
- [ ] Manual E2E happy path: open PR → review session auto-launches → submit review → session flips to `.completed` on next 60s tick → author runs `/refine` + re-requests → new session auto-launches.
- [ ] Manual E2E force-push path: with an active review session, force-push without re-request → next refresh detects `headRefOid` mismatch and re-kicks.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Models/ReviewRequest.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/ReviewRequest.swift
@@ -15,6 +15,8 @@ public struct ReviewRequest: Identifiable, Codable, Sendable {
     public var labels: [LabelInfo]
     public var provider: Provider
     public var reviewSessionID: UUID?  // set if a review session already exists
+    public var headRefOid: String?     // PR head commit SHA — used to detect new pushes
+    public var viewerLastReviewedAt: Date?  // viewer's most recent APPROVED/CHANGES_REQUESTED/DISMISSED review timestamp
 
     public init(
         id: String,
@@ -29,7 +31,9 @@ public struct ReviewRequest: Identifiable, Codable, Sendable {
         requestedAt: Date? = nil,
         labels: [LabelInfo] = [],
         provider: Provider = .github,
-        reviewSessionID: UUID? = nil
+        reviewSessionID: UUID? = nil,
+        headRefOid: String? = nil,
+        viewerLastReviewedAt: Date? = nil
     ) {
         self.id = id
         self.prNumber = prNumber
@@ -44,5 +48,7 @@ public struct ReviewRequest: Identifiable, Codable, Sendable {
         self.labels = labels
         self.provider = provider
         self.reviewSessionID = reviewSessionID
+        self.headRefOid = headRefOid
+        self.viewerLastReviewedAt = viewerLastReviewedAt
     }
 }
--- a/Packages/CrowCore/Sources/CrowCore/Models/Session.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/Session.swift
@@ -16,6 +16,13 @@ public struct Session: Identifiable, Codable, Sendable {
     // prompt dispatched. Gates the launchClaude prompt-vs-`--continue`
     // branch so completed reviews don't restart on app relaunch.
     public var reviewPromptDispatched: Bool
+    // Head SHA of the PR at the time the review session was created or
+    // last re-launched. Used by the kickoff guard as a fallback re-kick
+    // signal when a PR's head advances without an explicit re-request
+    // (e.g. force-push) or before the viewer-submitted-review signal is
+    // observed. Nil for non-review sessions and for legacy persisted
+    // sessions predating this field (CROW-290).
+    public var lastReviewedHeadSha: String?
 
     public init(
         id: UUID = UUID(),
@@ -28,7 +35,8 @@ public struct Session: Identifiable, Codable, Sendable {
         provider: Provider? = nil,
         createdAt: Date = Date(),
         updatedAt: Date = Date(),
-        reviewPromptDispatched: Bool = false
+        reviewPromptDispatched: Bool = false,
+        lastReviewedHeadSha: String? = nil
     ) {
         self.id = id
         self.name = name
@@ -41,6 +49,7 @@ public struct Session: Identifiable, Codable, Sendable {
         self.createdAt = createdAt
         self.updatedAt = updatedAt
         self.reviewPromptDispatched = reviewPromptDispatched
+        self.lastReviewedHeadSha = lastReviewedHeadSha
     }
 
     // Backward-compatible decoding: default `kind` to `.work` when missing
@@ -60,5 +69,6 @@ public struct Session: Identifiable, Codable, Sendable {
         createdAt = try container.decode(Date.self, forKey: .createdAt)
         updatedAt = try container.decode(Date.self, forKey: .updatedAt)
         reviewPromptDispatched = try container.decodeIfPresent(Bool.self, forKey: .reviewPromptDispatched) ?? true
+        lastReviewedHeadSha = try container.decodeIfPresent(String.self, forKey: .lastReviewedHeadSha)
     }
 }
--- a/Packages/CrowCore/Tests/CrowCoreTests/SessionModelTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/SessionModelTests.swift
@@ -48,6 +48,41 @@ import Testing
     #expect(decoded.provider == nil)
 }
 
+@Test func sessionBackwardCompatDecodingLastReviewedHeadSha() throws {
+    // Persisted state.json predating CROW-290 has no `lastReviewedHeadSha`.
+    // Decode must succeed and default the field to nil.
+    let id = UUID()
+    let date = Date()
+    let formatter = ISO8601DateFormatter()
+    formatter.formatOptions = [.withInternetDateTime]
+    let dateStr = formatter.string(from: date)
+    let json: [String: Any] = [
+        "id": id.uuidString,
+        "name": "legacy",
+        "status": "active",
+        "kind": "review",
+        "createdAt": dateStr,
+        "updatedAt": dateStr,
+    ]
+    let data = try JSONSerialization.data(withJSONObject: json)
+    let decoder = JSONDecoder()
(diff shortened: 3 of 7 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None — the fix is sound, the tests cover the meaningful cases, and the CrowCore unit tests pass locally (167/167).

### Security Review
**Strengths:**
- No secrets or credentials touched. New data flowing in is GraphQL response payloads (`headRefOid`, `submittedAt`, `state`, `author.login`, `viewer.login`) — all already trusted-from-GitHub.
- `viewerLastReviewedAt` whitelists verdict states (`APPROVED`/`CHANGES_REQUESTED`/`DISMISSED`) before promoting a review to "satisfied" — a `COMMENTED` review can't spuriously close a session.
- `headRefOid` is treated as opaque equality (no shell interpolation, no path construction), so the new field doesn't open an injection surface.
- Back-compat decode of `Session.lastReviewedHeadSha` is `decodeIfPresent` with a `nil` default — legacy `state.json` stays loadable, no crash-on-launch risk after upgrade.

**Concerns:**
- None.

### Code Quality

**Strengths:**
- Decision logic in `IssueTracker.decideReviewCompletions` cleanly separates Rule 1 (viewer-verdict signal, only needs `ReviewRequest`) from Rules 2/3 (terminal-PR-state, gated on `prDataComplete`). Tests assert exactly this distinction (`viewerReviewCompletesEvenUnderPartialFetch`).
- The `(request.id, headRefOid)` fingerprint keying for in-flight dedupe is the right shape — the previous `id`-only set was the precise reason round 2 was blocked.
- Strict `>` comparison on `reviewedAt > session.createdAt` correctly handles the round-1-verdict-before-round-2-session case (verified by `reviewSubmittedBeforeSessionStartIgnored`).
- B-fallback explicitly completes the stale session inline before re-kicking, so `reviewSessions` never double-binds the same PR.

### Summary Table
| Priority | Issue |
|----------|-------|
| Yellow | Legacy sessions (pre-CROW-290 `state.json` with `lastReviewedHeadSha == nil`) hit the B-fallback as soon as the upgraded app refreshes: `nil != request.headRefOid` is `true`, so `shaAdvanced` fires and the in-progress review session is force-completed + re-kicked. `AppDelegate.swift:483-486`. Affects users mid-review at upgrade time. Easy fix: skip `shaAdvanced` when `linkedSession?.lastReviewedHeadSha == nil`, and backfill the SHA on first observation. |
| Yellow | `reviews(last: 20)` in the GraphQL query (`IssueTracker.swift:583`) caps the viewer-verdict scan at the 20 most recent reviews. On chatty PRs (lots of `COMMENTED` reviews from multiple reviewers) the viewer's actual `APPROVED`/`CHANGES_REQUESTED` could fall off the window, leaving rule 1 silently inert. Rule B (SHA advance) would still re-kick on next push, so this isn't a hang — just a missed signal A. Consider scanning only `states: [APPROVED, CHANGES_REQUESTED, DISMISSED]` (GraphQL supports it) to make the 20-cap meaningful. |
| Green | `autoReviewedFingerprints` grows monotonically across the app's lifetime. Each (PR, SHA) tuple is small but never evicted. Not a real leak; mention only because it's trivial to bound (drop entries when a PR leaves `reviewRequests`). |
| Green | `viewerLastReviewedAt > session.createdAt` mixes a GitHub-authored timestamp against a local `Date()` — clock skew of seconds is fine, but a heavily-skewed local clock could miss a verdict on the first tick after a re-request. Self-heals on next tick. |

**Recommendation:** Approve. The two yellow items are worth addressing in a follow-up, but neither blocks landing the fix: the bug being fixed (round-2 never auto-launches) is materially worse than either edge case, and the back-compat path self-corrects after one verdict cycle.

---

PR #294, opened 2026-05-20
fix: async per-terminal rehydration + review-kickoff yielding (#293)

### LINKED ISSUE #266: App crashes when reviews kick off; only ObjC exceptions are logged today
## Symptom

CrowApp has crashed/disappeared multiple times recently on `dhilgaertner`'s machine when one or more PR reviews kick off (manual `Start Review` button and the auto-review path on `IssueTracker` refresh). No diagnostic report is left behind, so the user just sees the app vanish.

## Local investigation findings (2026-05-14)

`~/Library/Logs/DiagnosticReports/` is empty — no `.ips` file was produced for any of the recent CrowApp exits. This confirms the existing crash-logging gap that `AppDelegate.installUncaughtExceptionHandler` (Sources/Crow/App/AppDelegate.swift:47) was added to mitigate for issue #240. The current handler only covers `NSException`; Swift runtime traps, SwiftUI internal aborts, and POSIX signals still bypass it.

The unified log shows two CrowApp exits today (12:39 and 16:17), both preceded by repeated SwiftUI faults:

```
CrowApp: (SwiftUICore) [com.apple.SwiftUI:Invalid Configuration]
  Currently, only presenting a single sheet is supported.
  The next sheet will be presented when the currently presented sheet gets dismissed.

CrowApp: (SwiftUICore) [com.apple.SwiftUI:Invalid Configuration]
  NSHostingView is being laid out reentrantly while rendering its SwiftUI content.
  This is not supported and the current layout pass will be skipped.
```

`launchd` then reports `[CrowApp]: shutting down` — but `launchd` logs the same line for clean quits and for many crash-induced terminations, so absence of an `.ips` does not rule out an abort/trap.

## Why \"review kickoff\" likely correlates

`AppDelegate.swift:382-393` fires reviews via fire-and-forget `Task { await sessionService?.createReviewSession(prURL: ...) }`. The batch and auto-review paths spawn one task per PR with no serialization:

```swift
appState.onBatchStartReview = { [weak self] prURLs in
    for url in prURLs {
        Task { await self.sessionService?.createReviewSession(prURL: url) }
    }
}
```

And `onReviewRequestsRefreshed` (AppDelegate.swift:407-421) does the same for every auto-review-enabled repo on each `IssueTracker` poll tick.

`createReviewSession` (SessionService.swift:1066) runs several `gh`/`git` shell commands, then on `MainActor` appends to `appState.sessions`/`worktrees`/`terminals`, calls `prepareTerminal` (spinning up a Ghostty surface or tmux window), and finally sets `appState.selectedSessionID = session.id`. With N tasks racing to mutate the published state and switch the selected session, SwiftUI ends up doing nested layouts that can trip the \"reentrant layout\" and \"single sheet\" assertions seen in the log. Under the right timing, the AttributeGraph can abort the process.

## Proposed work

1. **Reproduce.** Either trigger `onBatchStartReview` with 2-3 PRs in quick succession, or watch an auto-review repo with multiple stacked review requests, and confirm we can produce the reentrant-layout fault on demand.
2. **Serialize review session creation.** Funnel `createReviewSession` calls through an actor-owned queue (or at minimum, await each `appState` mutation block in series) so the UI sees one session-add at a time. Stagger `selectedSessionID` updates — possibly only select the *first* of a batch instead of letting each call clobber it.
3. **Audit `.sheet` usage near review flow** for any view that may be presented twice when two sessions appear in rapid succession.
4. **Broaden crash logging** so the next occurrence leaves evidence:
   - Install POSIX signal handlers (SIGABRT, SIGSEGV, SIGBUS, SIGILL, SIGFPE, SIGTRAP) that write `~/.local/share/crow/crash-logs/crow-<timestamp>.log` with `Thread.callStackSymbols`, recent `NSLog` tail, and current session/terminal counts before re-raising.
   - Capture `Swift.setErrorHandler`-style precondition failures where feasible (or at minimum, redirect stderr to a rotated log file at app launch so `Fatal error:` lines land somewhere we can read after the fact).
   - Consider integrating PLCrashReporter or KSCrash for symbolicated reports if we want parity with macOS's native `.ips` output. Lower-effort alternative: ensure unsymbolicated `Thread.callStackReturnAddresses` are written so we can symbolicate offline.
   - Surface the crash log path in the About window and on next launch (\"Crow crashed last time — view report\").

## Acceptance criteria

- A repro that produces the reentrant-layout fault or a crash log when reviews are kicked off in parallel.
- Concurrent review starts no longer trigger \"NSHostingView is being laid out reentrantly\" / \"single sheet\" faults in the unified log.
- A signal-level crash leaves a readable artifact in `~/.local/share/crow/crash-logs/` (or wherever we choose) on the next reproduction.

## References

- `Sources/Crow/App/AppDelegate.swift:47` — existing `NSException` handler (issue #240)
- `Sources/Crow/App/AppDelegate.swift:382-421` — review kickoff entry points
- `Sources/Crow/App/SessionService.swift:1066` — `createReviewSession`
- Related closed crash: #240 (terminal-switching crash; ObjC exception path)

### LINKED ISSUE #293: App hangs at launch (and during review bursts) — main actor blocked by serial terminal rehydration
## Symptom

CrowApp shows the macOS "Application Not Responding" spinner:

1. **At launch**, when there are many persisted sessions/terminals (observed with 11 sessions, each with several managed terminals). The window paints late and is unresponsive for ~20–30s.
2. **At runtime**, during bursts of review-session creation (e.g. several `/crow-review-pr` kickoffs in close succession, as can happen after #291's re-review trigger landed). The hang clears on its own once the queued work drains — not a true deadlock.

The crow CLI socket remains responsive during the hang (`crow list-sessions` returns), confirming the UI/main thread is the bottleneck and the socket server (on the global concurrent queue) is fine.

## Root cause

`SessionService.hydrateState()` (`Sources/Crow/App/SessionService.swift:128-150`) iterates every persisted session and every terminal and calls `rehydrateTerminalSurface` synchronously on the main actor:

- For `.ghostty` rows, `TerminalManager.shared.preInitialize` creates a Metal/AppKit surface (main-thread work).
- For `.tmux` rows, `TmuxBackend.shared.registerTerminal` shells out to `tmux new-window` per terminal.

Both run inline on the main actor inside a tight `for` loop, so the main thread is held until the entire fan-out completes. With N sessions × M terminals, launch latency scales linearly with no chance for the run loop to service AppKit / SwiftUI / window painting until it's done.

Log evidence from a recent hang:

```
23:31:47.506 [Crow] Session state hydrated (11 sessions)
23:31:47.507 [Crow] Main app launch complete — creating window
23:32:05.403 [CrowTelemetry tmux:first_prompt_progress … elapsed_ms=18514]  (×10, all at once)
23:33:13.421 [CrowTelemetry tmux:first_prompt_ms=24856 …]                     (terminals come ready sequentially)
…
23:33:20.285 last terminal ready
```

`sample` could not capture stacks (hardened binary), but the timing + code path match exactly.

## Proposed fix

In `SessionService.hydrateState()`, after `wireTerminalReadiness()` runs (so the readiness callback is wired before any `createSurface` events fire), dispatch the per-terminal `rehydrateTerminalSurface` calls as separate `Task { @MainActor in … }` units instead of a tight synchronous loop. That lets the run loop service window/UI updates between rehydrations.

The readiness state machine already tolerates terminals coming up async — that's how new terminals are created at runtime.

Secondary: also consider rate-limiting `enqueueReviewKickoff` so a refresh that fires multiple kickoffs spreads them across run-loop turns rather than back-to-back on the main actor.

## Repro

1. Build up ~10+ sessions with managed terminals.
2. Quit Crow, relaunch.
3. Observe spinner / "Not Responding" for 20–30s while terminals rehydrate.

## Acceptance

- Launching with 11+ sessions shows the window within ~1s and remains responsive while terminals start in the background.
- Creating 3+ review sessions in quick succession does not stall the UI.

dgershman:
Closes #293

## Summary
- Dispatch each per-terminal rehydration in `SessionService.hydrateState` as its own `Task { @MainActor in }` so the main actor gets to service AppKit/SwiftUI between calls. Previously a profile with many persisted `.tmux` rows pinned the run loop for seconds at launch (each `TmuxBackend.registerTerminal` spawns a subprocess).
- Replace the batched `rehydrationOverwrites` accumulator with a per-terminal `applyRehydrationResult` helper that locates the row by ID and persists fallback backend changes individually through `store.mutate`.
- Keep `wireTerminalReadiness()` and the `terminalReadiness` / `autoLaunchTerminals` pre-seeding synchronous and before any task dispatch — readiness callbacks fired from inside `preInitialize` still need their slots ready.
- Add `await Task.yield()` between kickoffs in `AppDelegate.enqueueReviewKickoff` so a burst of pending PRs spreads across run-loop turns. The first iteration runs immediately so single-PR kickoffs have no added latency; the serial-tail discipline that fixed #266 stays intact.

## Test plan
- [ ] Launch the app on a profile with many persisted sessions/terminals — UI should respond within the first run-loop turn (no multi-second hang); restored terminals still come up and auto-launch `claude --continue`.
- [ ] With `CROW_TMUX_BACKEND` unset on a profile that has persisted `.tmux` rows, confirm fallback to `.ghostty` rendering and the persisted row is left alone (existing semantics preserved).
- [ ] Trigger an auto-review refresh with ≥3 pending PRs (or batch Start Review on ≥3 PRs) — sidebar populates one row at a time and the rest of the UI stays responsive while the queue drains.
- [ ] Single-PR "Start Review" click still creates the session promptly.

### DIFF
--- a/Sources/Crow/App/AppDelegate.swift
+++ b/Sources/Crow/App/AppDelegate.swift
@@ -95,7 +95,12 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         let previous = reviewKickoffTail
         reviewKickoffTail = Task { @MainActor in
             await previous?.value
-            for url in urls {
+            // Yield between kickoffs so a burst of pending PRs spreads
+            // across run-loop turns and SwiftUI/AppKit can render between
+            // each `createReviewSession` (#293). The first iteration runs
+            // immediately so the single-PR case has no added latency.
+            for (i, url) in urls.enumerated() {
+                if i > 0 { await Task.yield() }
                 _ = await service.createReviewSession(prURL: url, selectAfterCreate: false)
             }
         }
--- a/Sources/Crow/App/SessionService.swift
+++ b/Sources/Crow/App/SessionService.swift
@@ -124,41 +124,52 @@ final class SessionService {
         // launches; spec §12). If a .tmux row's re-registration fails (e.g.
         // tmux uninstalled), silently fall back to .ghostty so the user
         // can still use the app.
-        var rehydrationOverwrites: [SessionTerminal] = []
+        //
+        // Each per-terminal rehydration is dispatched as its own @MainActor
+        // task so the run loop can service AppKit/SwiftUI between them.
+        // Previously this loop ran synchronously and pinned the main actor
+        // for seconds on profiles with many persisted .tmux rows (each
+        // call to TmuxBackend.registerTerminal spawns a subprocess) — #293.
         for session in appState.sessions {
-            guard var terminals = appState.terminals[session.id] else { continue }
+            guard let terminals = appState.terminals[session.id] else { continue }
             let isManagerSession = session.id == AppState.managerSessionID
-            for i in terminals.indices {
-                let trackReadiness = !isManagerSession && terminals[i].isManaged
-                let updated = rehydrateTerminalSurface(terminals[i], trackReadiness: trackReadiness)
-                if updated.backend != terminals[i].backend || updated.tmuxBinding != terminals[i].tmuxBinding {
-                    rehydrationOverwrites.append(updated)
+            let sid = session.id
+            for original in terminals {
+                let trackReadiness = !isManagerSession && original.isManaged
+                Task { @MainActor in
+                    let updated = self.rehydrateTerminalSurface(original, trackReadiness: trackReadiness)
+                    self.applyRehydrationResult(sessionID: sid, original: original, updated: updated)
                 }
-                terminals[i] = updated
             }
-            appState.terminals[session.id] = terminals
         }
 
-        if var globals = appState.terminals[AppState.globalTerminalSessionID] {
-            for i in globals.indices {
-                let updated = rehydrateTerminalSurface(globals[i], trackReadiness: false)
-                if updated.backend != globals[i].backend || updated.tmuxBinding != globals[i].tmuxBinding {
-                    rehydrationOverwrites.append(updated)
+        if let globals = appState.terminals[AppState.globalTerminalSessionID] {
+            let sid = AppState.globalTerminalSessionID
+            for original in globals {
+                Task { @MainActor in
+                    let updated = self.rehydrateTerminalSurface(original, trackReadiness: false)
+                    self.applyRehydrationResult(sessionID: sid, original: original, updated: updated)
                 }
-                globals[i] = updated
             }
-            appState.terminals[AppState.globalTerminalSessionID] = globals
         }
+    }
 
-        // Persist any backend / windowIndex changes from re-hydration back
-        // to the store so a subsequent crash doesn't try to re-register
-        // terminals that are already on the fallback backend.
-        if !rehydrationOverwrites.isEmpty {
+    /// Commit the result of a per-terminal rehydration task back to
+    /// `appState.terminals`, locating the row by ID since the array may
+    /// have shifted while the task awaited. If the backend or tmuxBinding
+    /// changed (e.g. .tmux → .ghostty fallback), persist the updated row
+    /// so a subsequent crash doesn't re-attempt the failed backend.
+    @MainActor
+    private func applyRehydrationResult(sessionID: UUID, original: SessionTerminal, updated: SessionTerminal) {
+        if var terminals = appState.terminals[sessionID],
+           let idx = terminals.firstIndex(where: { $0.id == updated.id }) {
+            terminals[idx] = updated
+            appState.terminals[sessionID] = terminals
+        }
+        if updated.backend != original.backend || updated.tmuxBinding != original.tmuxBinding {
             store.mutate { data in
-                for updated in rehydrationOverwrites {
-                    if let i = data.terminals.firstIndex(where: { $0.id == updated.id }) {
-                        data.terminals[i] = updated
-                    }
+                if let i = data.terminals.firstIndex(where: { $0.id == updated.id }) {
+                    data.terminals[i] = updated
                 }
             }
         }


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Purely a concurrency / scheduling change — no new IO, no new shell invocations, no new untrusted-data handling. `TmuxBackend.registerTerminal` is called identically to before; only the dispatch envelope changed.
- No `self` retention surprises in the new `Task { @MainActor in … }` closures: `SessionService` is owned by `AppDelegate` for the app's lifetime, and `Task.yield()` in `enqueueReviewKickoff` doesn't introduce reentrancy because `reviewKickoffTail` is still a single serial chain.

**Concerns:** None identified.

### Code Quality

**Correctness — looks good:**
- Pre-seeding `terminalReadiness` and `autoLaunchTerminals` before the async dispatch (SessionService.swift:101-105) is preserved, so readiness callbacks fired synchronously from inside `TmuxBackend.startReadinessWatch` → `MainActor.run` (TmuxBackend.swift:444) still find their slot. `wireTerminalReadiness()` is also still called before any task is dispatched.
- `applyRehydrationResult` correctly relocates the row by `id` instead of trusting an index (SessionService.swift:165), which is the right call now that the array can theoretically shift between dispatch and execution. All tasks run on the same `@MainActor`, so even though dispatch order isn't a Swift Concurrency guarantee in general, the read-modify-write of `appState.terminals[sid]` is safe — only one task body executes at a time.
- Routing is unaffected by the brief window where `appState.terminals` holds a stale `tmuxBinding`: `TerminalRouter` keys off `terminal.id` and dispatches to `TmuxBackend.shared`, which maintains its own `bindings[id] = windowIndex` map updated synchronously inside `registerTerminal` (TmuxBackend.swift:184). The persisted `tmuxBinding` is for next-launch recovery only.
- `onReadinessChanged` is only fired asynchronously through `SentinelWaiter` → `MainActor.run` (TmuxBackend.swift:444-453), never synchronously inside `registerTerminal`. So by the time `launchClaude` runs, `applyRehydrationResult` has already committed the row.
- The `await Task.yield()` between review kickoffs (AppDelegate.swift:103) is a clean way to keep the serial-tail discipline from #266 while spreading work across run-loop turns. Skipping the yield on the first iteration is the right call for the single-PR latency story.

**Minor — non-blocking:**
- The previous code batched all rehydration overwrites into a single `store.mutate` at the tail. The new code calls `store.mutate` per terminal whose backend/binding changed (SessionService.swift:170). Because `windowIndex` is re-issued by tmux on every launch, this is almost always at least one mutate per `.tmux` row on every launch (each does its own NSLock acquire + atomic file write). On a profile with many `.tmux` rows this trades a single batched disk write for N writes — not a correctness issue, but it does erode some of the responsiveness gain you're chasing. Worth considering a coalescing layer (e.g. debounce store sync, or queue updates and drain on idle) if profiling shows it.
- The unstructured `Task { @MainActor in … }` instances aren't tracked anywhere, so there's no way to `await` "all rehydrate complete" if a future caller needs that signal (e.g. a test, or a follow-up that wants to start telemetry only after rehydrate). Consider gathering them into a `[Task]` if/when that need arises — not needed for this PR.
- No test coverage added for the new async path — the test plan is entirely manual. Given the difficulty of unit-testing `@MainActor` task scheduling against AppKit/SwiftUI, that's reasonable, but worth a follow-up: even a tiny test verifying `applyRehydrationResult` correctly relocates a row whose array position has shifted would document the contract.

### Summary Table
| Priority | Issue |
|----------|-------|
| Green | Consider batching per-terminal `store.mutate` calls during hydrate to avoid N disk writes per launch on tmux-heavy profiles. |
| Green | Optional: track the dispatched rehydrate Tasks so future code can `await` rehydrate completion. |
| Green | Optional: add a small regression test for `applyRehydrationResult` ID-based row lookup. |

**Recommendation:** Approve. Focused, well-reasoned fix to a real responsiveness problem. The async-dispatch design preserves all the existing invariants (readiness pre-seeding, serial review kickoff, fallback semantics), and the by-ID write-back in `applyRehydrationResult` correctly handles the ordering uncertainty the change introduces. The only quibble is N small store writes replacing one batched write, which is non-blocking.

---

PR #297, opened 2026-05-20
crow-workspace: pre-fetch & embed ticket content (closes #295)

### LINKED ISSUE #295: /crow-workspace: `gh issue view` returns empty output in launched Claude Code session
## Summary

After `/crow-workspace` finishes setting up a session and Claude Code launches in the new worktree terminal, the first thing the agent does is run `gh issue view <url> --comments` to read the ticket. In recent sessions this command **returns no output and exits 0**, so Claude has no ticket context to work from.

`gh api repos/{owner}/{repo}/issues/{num}` works fine in the same shell, so it appears specific to `gh issue view` (likely a pager / TTY detection / sandbox interaction issue), not `gh` auth or network.

## Repro

1. Run `/crow-workspace https://github.com/radiusmethod/corveil/issues/723` from the manager.
2. Wait for Claude Code to launch in the new worktree terminal.
3. Observe the agent's first command: `gh issue view 723 --repo radiusmethod/corveil --comments`.
4. Output is empty; exit code is 0. Retrying with `2>&1`, `| cat`, `> /tmp/file`, and inside the worktree directory all yield zero bytes.

## Observed

```
⏺ Bash(gh issue view 723 --repo radiusmethod/corveil --comments)
  ⎿  (No output)

⏺ Bash(gh issue view 723 --comments 2>&1; echo "EXIT: $?")
  ⎿  EXIT: 0

⏺ Bash(gh issue view 723 --comments > /tmp/issue.txt 2>&1; wc -l /tmp/issue.txt; cat /tmp/issue.txt)
  ⎿         0 /tmp/issue.txt

⏺ Bash(gh api repos/radiusmethod/corveil/issues/723 2>&1 | head -100)
  ⎿  {
       "url": "https://api.github.com/repos/radiusmethod/corveil/issues/723",
       ...
```

## Expected

`gh issue view` prints the issue title, body, and comments to stdout, the same as it does when run from a normal terminal.

## Environment

- `gh version 2.91.0 (2026-04-22)`
- Worktree launched via `setup.sh` → `crow new-terminal` + `crow send` to start Claude Code
- macOS, zsh

## Suggested next steps

- Check whether `gh` is detecting a non-TTY and trying to invoke a pager that silently swallows output in the crow-managed terminal.
- Consider having `setup.sh` (or the first-prompt template) prefer `gh api` or `GH_PAGER=cat gh issue view ...` for the ticket fetch, since `gh api` is reliable in this environment.
- Reproduce by launching a bare terminal via `crow new-terminal` and running `gh issue view <url>` directly to isolate whether it is the terminal backend or Claude Code's Bash tool.

dgershman:
## Summary

- Add a **Step 0** to the `/crow-workspace` skill: the manager pre-fetches the ticket title, body, and comments (and the existing PR, if any) before writing the launch prompt.
- The fetched content is embedded verbatim into the `## Ticket` (and `## Existing Pull Request`) sections of the first-prompt template, so the launched Claude reads everything from message zero.
- Keep the `gh issue view` / `gh pr view` commands documented in the prompt as a fallback for re-fetching fresh data later — including a `gh api repos/{owner}/{repo}/issues/{n}` fallback for the issue #295 case where `gh issue view` returns empty.

## Why

Per #295, the launched Claude Code session is started with `--permission-mode plan`, and the prompt instructed it to call `gh issue view ... --comments` with `dangerouslyDisableSandbox: true` on the very first turn. Sandbox-disabled Bash calls always require human approval in plan mode, so the call sits waiting on a permission prompt — producing the "empty output, exit 0" symptom in a probabilistic, race-prone way.

Pre-fetching once in the manager (which already has full ticket-fetching capability) avoids the startup approval entirely and gives the launched Claude richer context from the start. The fallback commands stay in the prompt so re-fetching fresh data remains a one-liner.

## Test plan

- [ ] Run `/crow-workspace https://github.com/radiusmethod/crow/issues/295` from a manager session.
- [ ] Verify `{devRoot}/.claude/prompts/crow-prompt-{session_name}.md` contains the ticket title, body, and (rendered) comments under `## Ticket`.
- [ ] Verify the launched Claude Code session does NOT issue a `gh issue view` Bash call at startup and does not sit on a permission prompt.
- [ ] Run `/crow-workspace` against an issue that already has an open PR. Verify the `## Existing Pull Request` section is similarly embedded with PR title/body/comments.
- [ ] Sanity-check fallback path: temporarily simulate a pre-fetch failure and confirm the prompt falls back to the documented `_(Ticket pre-fetch failed — run the gh command below to retrieve.)_` note + commands.

### DIFF
--- a/skills/crow-workspace/SKILL.md
+++ b/skills/crow-workspace/SKILL.md
@@ -207,9 +207,38 @@ After the LLM resolves names (slug, branch, worktree path, session name), detect
 
 > **IMPORTANT:** All `crow`, `gh`, `glab`, and `git` commands require `dangerouslyDisableSandbox: true`. The `setup.sh` call itself must use `dangerouslyDisableSandbox: true` since it runs these commands internally.
 
+#### Step 0: Pre-fetch ticket (and PR) content
+
+Before writing the prompt file, fetch the ticket title, body, and comments so they can be embedded directly into the `## Ticket` section. This avoids the launched Claude Code session sitting on a `dangerouslyDisableSandbox` permission prompt at startup when it tries to run `gh issue view` itself (issue #295).
+
+Use `dangerouslyDisableSandbox: true` for all of these fetches.
+
+**GitHub:**
+```bash
+gh issue view {ticket_url} --comments
+# Fallback if the above returns empty output (see issue #295):
+gh api repos/{owner}/{repo}/issues/{number}
+gh api repos/{owner}/{repo}/issues/{number}/comments
+```
+
+**GitLab (non-default host):**
+```bash
+GITLAB_HOST={host} glab issue view {number} --repo {org/repo} --comments
+```
+
+**If an existing PR was detected for this ticket**, also fetch the PR view so it can be embedded:
+```bash
+gh pr view {pr_url} --comments
+# Fallback:
+gh api repos/{owner}/{repo}/pulls/{pr_number}
+gh api repos/{owner}/{repo}/issues/{pr_number}/comments
+```
+
+Render the fetched content into the prompt template per the formatting rules in **First Prompt Template** below.
+
 #### Step 1: Write the prompt file
 
-The LLM writes the prompt content (see template below) to a file:
+The LLM writes the prompt content (see template below — with the pre-fetched ticket/PR content embedded) to a file:
 
 ```bash
 mkdir -p {devRoot}/.claude/prompts
@@ -322,14 +351,27 @@ Extract dynamically from first non-heading line of CLAUDE.md or README.md. If no
 
 ## Ticket
 
-IMPORTANT: All gh/glab commands MUST use dangerouslyDisableSandbox: true. They will fail with TLS certificate errors otherwise. Do NOT attempt sandboxed first.
+**{ticket_title}** — {ticket_url}
+
+{ticket_body_verbatim}
+
+### Comments
+
+{rendered_comments_or_no_comments_marker}
+
+---
+
+If you need fresher ticket data later, re-fetch with (all gh/glab commands MUST use dangerouslyDisableSandbox: true — they will fail with TLS certificate errors otherwise, and will prompt for approval):
 
 ```bash
-gh issue view https://github.com/org/repo/issues/123 --comments
+gh issue view {ticket_url} --comments
+# Fallback if the above returns empty output (see issue #295):
+gh api repos/{owner}/{repo}/issues/{number}
+gh api repos/{owner}/{repo}/issues/{number}/comments
 ```
 
 ## Instructions
-1. Study the ticket thoroughly — use dangerouslyDisableSandbox: true for ALL gh/glab commands
+1. Study the ticket above — it has been pre-fetched and embedded. Only re-run gh/glab if you need fresher data; those calls use dangerouslyDisableSandbox: true and will prompt for approval.
 2. Create an implementation plan
 3. Implement the plan
 4. Commit the changes with a descriptive message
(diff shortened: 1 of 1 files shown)


dhilgaertner (approved):
## Code & Security Review

Scope: documentation-only change to `skills/crow-workspace/SKILL.md` (+85/-15). No executable code, no tests affected. The change moves ticket/PR fetching from the launched Claude's first turn into a manager-side **Step 0**, then embeds the fetched content verbatim into the launch prompt.

### Critical Issues
None blocking.

### Security Review

**Strengths:**
- Net attack surface is unchanged. The same issue/PR content (which can contain attacker-controlled text from any GitHub user who can comment) was previously read by the launched Claude via a tool result; now it arrives in the initial prompt. From a prompt-injection perspective this is a relocation, not a new vector.
- Pre-fetching in the manager avoids a `dangerouslyDisableSandbox: true` call on the launched session's very first turn, which has UX value even if the underlying #295 symptom is a TTY/pager interaction rather than a permission-mode race.
- The fallback path (`_(Ticket pre-fetch failed — run the gh command below to retrieve.)_`) is documented, so a failed pre-fetch degrades gracefully to prior behavior.

**Concerns (informational, not blocking):**
- Issue and PR bodies/comments are now embedded into the launch prompt with explicit instructions to render them **verbatim — preserve markdown, code fences, line breaks. Do not summarize or reformat** (SKILL.md:397-400). This is a pre-existing prompt-injection surface, but worth flagging since the change makes it more explicit and persistent (it lives in the initial prompt rather than a tool result). Consider a future note suggesting the manager fence embedded user-controlled content (e.g., a clearly marked block) so downstream Claude can more reliably treat it as data rather than instructions.

### Code Quality

**Yellow — should consider:**
- **GitLab MR pre-fetch is documented for issues but not MRs.** Step 0 (SKILL.md:229-235) shows `gh pr view` + `gh api .../pulls/{n}` for the existing-PR branch but the GitLab block (SKILL.md:224-227) only covers `glab issue view`. For symmetry with the GitHub variant — and given the existing-MR flow is documented elsewhere in the skill — Step 0 should also document the `glab mr view {number} --repo {org/repo} --comments` pre-fetch.
- **HEREDOC delimiter collision risk.** Step 1 still shows `cat > ... << 'PROMPT'` (SKILL.md:243-248) with the pre-fetched issue/PR text embedded inside. If a ticket body or comment contains a line that is exactly `PROMPT`, the heredoc terminates early. Single-quoting the delimiter prevents *variable* expansion but does not protect against delimiter collision in content. Suggest either using a less collision-prone delimiter (e.g. `__CROW_PROMPT_END__`) or noting that the LLM should write via the Write tool when the content is large/user-supplied.

**Green — consider:**
- **Fallback message text is ticket-specific.** The fallback note (SKILL.md:402) reads `_(Ticket pre-fetch failed — ...)_`. The same fallback is implicitly reused for the PR variant; consider adding a parallel `_(PR pre-fetch failed — ...)_` line for the existing-PR section so the embedded text is accurate.
- **Diagnosis in PR description vs. #295 symptoms.** Issue #295 reports `gh issue view` returning empty output + exit 0 (a probable TTY/pager interaction), not blocking on a permission prompt. The PR's framing attributes it to plan-mode + `dangerouslyDisableSandbox: true`. The fix sidesteps both, so this isn't blocking — but if the underlying #295 symptom persists for downstream `gh` calls in the launched session, the prompt's documented fallbacks (`gh issue view ...`) would also fail. The `gh api` fallbacks are correctly the more reliable path.
- Step renumbering (existing-PR Instructions 6→5 steps, MyGitLab additional step 7→6) is consistent.

### Summary Table
| Priority | Issue |
|----------|-------|
| Yellow | Step 0 missing `glab mr view` pre-fetch documentation for existing-MR variant |
| Yellow | HEREDOC `'PROMPT'` delimiter can collide with user-controlled content; recommend a more unique sentinel or Write-tool path |
| Green | Embedded user content is verbatim — consider future hardening with explicit data-vs-instruction fencing |
| Green | Fallback message text is ticket-only; PR variant could use a parallel line |
| Green | PR diagnosis vs. #295 reported symptom differ; fix is still correct |

**Recommendation:** Approve. This is a pragmatic, documentation-only improvement that gives the launched Claude richer ticket context from message zero and avoids a startup approval prompt. The flagged items are minor and can be addressed in a follow-up.

---

PR #229, opened 2026-05-01
Add tmux backend behind CROW_TMUX_BACKEND feature flag (#198)

dhilgaertner:
## Summary

Implements the recommendation from the [#198 spike](https://github.com/radiusmethod/crow/issues/198): a single embedded Ghostty surface fronting a per-app tmux server, where each Crow tab is a tmux window. Off by default; flipping `CROW_TMUX_BACKEND=1` in the environment opts the user in for testing.

The wins (when the flag is on):

- **Reliable shell-readiness** via a sentinel-file polled by `SentinelWaiter` — replaces the 5s sleep at `TerminalManager.swift:113-126`. Spike measured 172ms (zsh) / 231ms (bash); through tmux, sub-50ms.
- **Single libghostty surface for the whole app** — N sessions = 1 Metal context, not N.
- **Switching tabs is just `tmux select-window`** — no second surface created.
- **Bounded 50KB+ input routing** for `crow send` via `tmux load-buffer + paste-buffer`. `send-keys -l` would fail with `command too long`.
- **Free session persistence design** in place — v1 still kills the server on app exit (spec §12), but `SessionTerminal.tmuxBinding` rows survive restart and re-bind cleanly.

## Ticket map (spec §14)

Seven discrete pieces, each its own commit:

| # | Commit | Result |
|---|---|---|
| 1 | `63595d1` Add TerminalBackend discriminator | Backwards-compat persisted-row migration to `{ ghostty, tmux }` + `TmuxBinding`. |
| 2 | `8312622` Bundle wrapper.sh + tmux.conf | Real on-disk artifacts, syntax-highlightable. |
|   | `18253f8` Add SentinelWaiter + TmuxController | Async-friendly readiness; ARG_MAX-safe paste. |
|   | `71e4806` Add TmuxBackend orchestrator | Owns server + cockpit surface + UUID→window map. |
|   | `5dae7a0` Wire CROW_TMUX_BACKEND flag | App-launch detect, RPC dispatch, surface dispatch in `TerminalSurfaceView`. |
| 3 | (folded into #2) | `TmuxBackend.sendText` already uses `load-buffer + paste-buffer`. |
| 4 | `e35a984` First-run onboarding alert | Native `NSAlert` when tmux ≥ 3.3 is missing — copies `brew install tmux`, opens install guide, or continues with Ghostty. |
| 5 | `7f04986` 2s timeout watchdog | `TmuxController.run` SIGTERMs hung tmux, throws `.timedOut` instead of letting the UI freeze. |
[...]

PR #302, opened 2026-05-21
feat: make tmux the default terminal backend (closes #301)

### LINKED ISSUE #301: Make tmux the default terminal backend; drop "Experimental" framing
## Summary

The tmux terminal backend (#198 → shipped behind a flag in #229) has been the experimental opt-in for several releases. Multiple readiness/escape-hatch bugs found post-ship have been fixed (#252, #256, #259, #262, #264) and the backend now appears stable enough to be the default. Drop the "experimental" framing and flip the default so new and existing users get tmux without having to toggle anything.

This is a **default-flip + UI relabel**, not a removal of the legacy code path. The non-tmux path stays in the tree as a one-release escape hatch in case of regressions, then gets deleted in a follow-up.

## Today

- `AppConfig.experimentalTmuxBackend: Bool = false` — persisted in `{devRoot}/.claude/config.json`.
- `FeatureFlags.tmuxBackend` is OR-merged from the `CROW_TMUX_BACKEND` env var and the config toggle (`Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift:19`, `Sources/Crow/App/FeatureFlags.swift:18`).
- Surfaced in **Settings → Experimental** via `Packages/CrowUI/Sources/CrowUI/ExperimentalSettingsView.swift` ("Use tmux for managed terminals", "Experimental — see #198 for context").
- Two gated code paths consume the flag: `AppDelegate.swift:278`, `AppDelegate.swift:1030`, `SessionService.swift:1372`.

## Goal

After this ticket:

1. **Default is tmux.** New installs get tmux. Existing users with no explicit setting get tmux on next launch.
2. **No more "Experimental" framing.** The toggle either disappears entirely or is reframed as a "Use legacy per-terminal Ghostty backend" escape-hatch under a regular Terminal/Advanced section — not under Experimental.
3. **Single source of truth field.** Rename `experimentalTmuxBackend` to something neutral (`legacyTerminalBackend` if we keep an opt-out, or remove it outright if we don't). Decode-with-default keeps old configs forward-compatible.
4. **Env-var direction inverts.** `CROW_TMUX_BACKEND=1` is no longer interesting (it's the default). Add `CROW_LEGACY_TERMINAL_BACKEND=1` (or `CROW_TMUX_BACKEND=0` parsed as explicit-off) so we can disable on the fly during the transition.
5. **Legacy code path stays for one release.** Don't delete the non-tmux branches in `AppDelegate` / `SessionService` yet — gate them off by default but keep them reachable via the escape hatch. Open a follow-up ticket to remove them after a release of soak.

## Acceptance criteria

1. Fresh install with no `config.json` → tmux backend on first launch (verifiable via `[Crow] tmux backend enabled` log line at startup).
2. Existing user with `experimentalTmuxBackend: false` (or unset) → tmux on next launch. Existing user with `experimentalTmuxBackend: true` → unchanged.
3. `ExperimentalSettingsView` is either deleted or no longer contains the tmux toggle. If a "use legacy backend" opt-out toggle is kept, it lives under a non-Experimental section and uses neutral copy.
4. `CROW_TMUX_BACKEND=0` (or the new `CROW_LEGACY_TERMINAL_BACKEND=1`) reverts to the legacy path for that launch.
5. README + CHANGELOG updated. Note in CHANGELOG calls out the default flip and the escape-hatch env var.
6. Tests in `FeatureFlagsTests.swift` and `AppConfigTests.swift` updated to reflect the new default and decode-with-default behavior. Round-trip test ensures existing configs without the new key still load.

## Out of scope

- **Deleting the legacy Ghostty per-terminal code path.** That's a follow-up; this ticket only flips the default and re-frames the UI.
- **Bundling tmux** in the `.app`. We still rely on `brew install tmux` (>= 3.3) on the user's PATH; the spike (#198 §C) called this out and the brew-prompt UX hasn't been re-litigated.
- **Migration of in-flight legacy session state.** Open sessions on the legacy backend at launch should continue on legacy until the app restarts — the flag is decided once at startup per `FeatureFlags` contract.

## Open questions

- **Do we keep the opt-out at all?** Argument for: lets us tell users "flip this back" if a regression appears. Argument against: we already have the env var; a SwiftUI toggle is more UI surface than a one-release escape hatch deserves. Recommend env-var only, no UI toggle — keeps Settings clean.
- **Naming for the escape-hatch env var.** `CROW_LEGACY_TERMINAL_BACKEND=1` reads clearly; `CROW_TMUX_BACKEND=0` reuses the existing var name. Recommend the latter (smaller blast radius — anyone already setting it to `1` keeps working, setting to `0` newly opts out).
- **Where to surface the tmux requirement note** now that "Experimental" copy is gone. README is enough; consider a one-line first-run check that nags if tmux is missing or < 3.3 (separate ticket if we want this).

dgershman:
## Summary
- Flips the tmux terminal backend from opt-in (#229) to default. New and existing users get tmux on next launch without toggling anything.
- Drops the "Experimental" framing entirely: Settings → Experimental tab is removed (it was tmux-only), `ExperimentalSettingsView` is deleted, and `AppConfig.experimentalTmuxBackend` is gone.
- Inverts the env-var direction: `CROW_TMUX_BACKEND=0` / `false` / `no` / `off` is the escape hatch for the launch. Anything else (or unset) keeps the default-on tmux. Legacy Ghostty per-terminal code paths stay reachable for one release of soak; their deletion is tracked separately.

## Why no UI opt-out
Per the open question in the ticket, a SwiftUI toggle is more surface than a one-release escape hatch deserves — the env var is enough. Reusing `CROW_TMUX_BACKEND` rather than introducing a new var name keeps the blast radius tiny: anyone already setting it to `1` keeps working; setting it to `0` newly opts out.

## Backwards compatibility
- Existing `config.json` containing `experimentalTmuxBackend: true` or `false` continues to decode — unknown keys are ignored. The key is silently dropped on next save (`appConfigDecodesLegacyExperimentalTmuxBackendKey` test pins this).
- Existing users with the toggle previously **off** now get tmux on next launch (acceptance criterion #2). Users with it **on** are unchanged.

## Watch items called out
- The first-launch experience for users without tmux installed now hits `showTmuxNotFoundOnboardingSheet()`, which falls back to Ghostty for that launch with a `brew install tmux` prompt. README + runtime-deps table now flag tmux ≥ 3.3 as a runtime dep.
- Manager terminal is still pinned to Ghostty; that migration is its own follow-up and unaffected here.

## Test plan
- [x] `cd Packages/CrowCore && swift test` — 169 tests pass, including the new `appConfigDecodesLegacyExperimentalTmuxBackendKey` forward-compat decode test.
- [ ] CI: full `make build` + `swift test` (local environment is missing a working GhosttyKit xcframework — code touches no Ghostty interface so CI is the canonical check).
- [ ] Fresh install with no `config.json` → confirm `[Crow] tmux backend configured: …` log line at startup.
- [ ] Existing user with `experimentalTmuxBackend: false` in config.json → confirm tmux still enabled on next launch.
- [ ] `CROW_TMUX_BACKEND=0 open -a Crow` → confirm legacy Ghostty backend selected (no tmux configured log line).
- [ ] Settings UI shows only General / Automation / Workspaces / Notifications — no Experimental tab.

Closes #301.


dhilgaertner (approved):
## Code & Security Review

Solid, well-scoped flip from opt-in to default-on, with a clean escape hatch and intact backward-compat decode. No critical or blocking issues — the diff is mostly deletions, which is the kind of PR I like reviewing.

### Security Review
**Strengths:**
- Surface area shrinks: `ExperimentalSettingsView.swift` (39 lines of UI), the persisted `experimentalTmuxBackend` config key, and the `tmuxBackendConfigOverride` `nonisolated(unsafe)` static are all gone. One less mutable global to reason about.
- Env-var parsing in `Sources/Crow/App/FeatureFlags.swift:21-27` is a pure switch over lowercased input — no injection, no shell-out, no surprises. Whitespace-not-trimmed behavior is intentional and pinned by `treatsEverythingElseAsDefault`.
- Forward-compat decode is exercised end-to-end (`Packages/CrowCore/Tests/CrowCoreTests/AppConfigTests.swift:296-307`): old configs with `experimentalTmuxBackend:true` still load, and re-encoding silently drops the key. That's the cleanest possible migration story.

**Concerns:**
- None.

### Code Quality
**Strengths:**
- `FeatureFlags.tmuxBackend` is now a one-liner computed property; the OR-merge plumbing and the `nonisolated(unsafe)` override are deleted instead of left dangling. Good cleanup discipline.
- `Sources/Crow/App/AppDelegate.swift:255-260` comment + the updated `showTmuxNotFoundOnboardingSheet` copy + the `TmuxBackendError.notConfigured` description (`Packages/CrowTerminal/.../TmuxBackend.swift:714`) all stay coherent with the new policy. No half-updated user-facing strings.
- `CHANGELOG.md`, `README.md`, `docs/architecture.md`, `docs/configuration.md`, `docs/troubleshooting.md` are all updated consistently. The env-var direction inversion (`=0` to opt out vs. the old `=1` to opt in) is explained in every place a user might look.
- PR description explicitly calls out acceptance criterion #2 (users who had toggled it OFF now get tmux ON) as intended — that's the right behavior to ship loudly rather than try to migrate.

**Yellow (should-fix, non-blocking):**
- `docs/terminal-runtime-research.md:757-758` is stale: \"Default backend remains Ghostty\" and \"Opt in via `CROW_TMUX_BACKEND=1` or Settings → Experimental\" are both wrong post-#301. The implementation-status section is the one place a curious reader will land for backend context, so it's worth a one-paragraph update mirroring `architecture.md`.
- `Tests/CrowTests/FeatureFlagsTests.swift` dropped the `.serialized` annotation from the suite, which was correct given `tmuxBackendConfigOverride` is gone — but `tmuxBackendOffWhenEnvExplicitlyOff` (lines 52-61) now calls `setenv` / `unsetenv` on the shared process environment. If `tmuxBackendDefaultIsOn` runs concurrently and reads `FeatureFlags.tmuxBackend` after the env check but before the assertion, it could observe a transient `CROW_TMUX_BACKEND=0`. The failure mode is flaky-fail rather than silent-skip. Re-adding `.serialized` (or moving the setenv test into its own single-test suite) would close the window. Probably won't bite in practice — both tests are fast — but the precedent is worth preserving.

**Green (consider):**
- The `parse` → `isExplicitlyOff` test helper still mirrors production logic by hand. If a future change touches `envExplicitlyOff` (e.g., accepting `disabled` or trimming whitespace), the helper has to be updated in lockstep or the test pins the wrong policy. The comment acknowledges this, so it's fine, but a tiny `@testable import` of the actual function would be more durable. Not worth doing now.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red      | — |
| Yellow   | `docs/terminal-runtime-research.md:757-758` still describes the old opt-in policy |
| Yellow   | `FeatureFlagsTests` lost `.serialized`; `setenv` test can race with the unset-env assertion |
| Green    | Test helper `isExplicitlyOff` duplicates production parse logic |

**Recommendation:** Approve. Code change is clean, well-tested, and the decode forward-compat is exactly the right shape for a config-key removal. The two yellow items are worth addressing — the stale research-doc paragraph in particular — but neither is blocking.

PR #308, opened 2026-05-22
fix: delete review-session clone on disk (closes #305)

### LINKED ISSUE #305: Review session delete leaves crow-reviews/<clone> directory on disk
## TL;DR

Deleting a review session via the UI removes the in-memory + persisted session state but **leaves the `{devRoot}/crow-reviews/<cloneDirName>` directory on disk**. Over time, every reviewed PR accumulates orphaned clones in `~/Dev/crow-reviews/` (or wherever `devRoot` is set). Each clone is a full git checkout, so disk usage grows steadily.

## Repro

1. Trigger a review session (e.g., via `/review` on a PR).
2. Observe `{devRoot}/crow-reviews/<clone-name>` directory exists on disk.
3. Delete the review session via the Crow UI (right-click → Delete, or whatever the affordance is).
4. **Bug:** the session row disappears from the UI, but the directory at `{devRoot}/crow-reviews/<clone-name>` is still on disk.

## Code references

The bug is in the disk-cleanup branch for review sessions. The hook is already there but apparently doesn't reach the clone path.

| Surface | File:line | Notes |
|---|---|---|
| Review-clone creation | `Sources/Crow/App/SessionService.swift:1128-1135` | `reviewsDir = devRoot + "/crow-reviews"` + `clonePath = reviewsDir + "/" + cloneDirName` |
| Session delete entrypoint | `Sources/Crow/App/SessionService.swift:540` (`deleteSession(id:)`) | Already gates on `isReview` at L547 |
| Cleanup dispatch | `Sources/Crow/App/SessionService.swift:562-564` (`performDiskCleanup(items: items, isReview: isReview)`) | The `isReview` flag is passed in but the `items` list is the worktree set — review clones aren't necessarily registered as worktrees |

Likely root cause: review sessions store their on-disk clone path in something other than `appState.worktrees(for: sessionID)`, so the `items` passed to `performDiskCleanup` is empty (or doesn't include the clone path). The `isReview` flag is set correctly but the cleanup function has nothing to delete.

**Verify before fixing:** read `performDiskCleanup` and check what it does when `items` is empty and `isReview: true`. If it's a no-op, that's the bug.

### Option A (recommended) — derive the clone path from session metadata at delete time

In `deleteSession`, when `session.kind == .review`, look up the clone path from the session's persisted state (likely a property on the `Session` struct or a separate review-specific store) and pass it explicitly to `performDiskCleanup`.

```swift
let reviewClonePath: String? = isReview ? appState.reviewClonePath(for: id) : nil  // helper to add
// ... pass into performDiskCleanup, which removes it via FileManager.removeItem(atPath:)
```

### Option B — register the review clone as a "worktree" in `appState.worktrees`

So that `appState.worktrees(for: id)` returns it like a normal worktree and the existing cleanup path picks it up.

**Recommended: Option A.** Review clones aren't real worktrees (they're full clones, not `git worktree add` artifacts) and conflating them in the worktree store would muddy invariants elsewhere. Keep them as a separate concept; add an explicit cleanup hook.

## Tests

```swift
@MainActor
func testDeleteReviewSession_RemovesCloneDirectory() async throws {
    // Set up a temp devRoot. Create a review session that clones to
    // devRoot/crow-reviews/<name>. Assert directory exists.
    // Call deleteSession. Assert directory no longer exists.
}

@MainActor
func testDeleteReviewSession_IdempotentAfterPartialFailure() async throws {
    // Set up a review session with the clone directory already removed
    // (simulating a prior failed cleanup mid-way). Call deleteSession.
    // Assert: no error; session is removed from state.
}

@MainActor
func testDeleteRegularSession_DoesNotTouchReviewsDir() async throws {
    // Confirm the fix doesn't accidentally reach into crow-reviews/ for
    // non-review sessions.
}
```

## Hard constraints

- **Don't delete the parent `crow-reviews/` directory.** Only the per-session `<cloneDirName>` subdirectory.
- **Idempotent:** if the directory is already gone (because the user deleted it manually), the cleanup must not error.
- **Don't widen worktree semantics.** Review clones are distinct from `git worktree add` artifacts; don't conflate the two stores.
- **Preserve the existing retry-on-failure UX.** Today `deleteSession` already surfaces `sessionDeletionError[id]` on cleanup failure (`SessionService.swift:566-577`). The fix must continue to honor that — if `removeItem` fails (e.g., file permissions), surface the error and leave the session row in place for retry.
- **No regression on non-review sessions.** Their worktree cleanup path stays unchanged.

## Acceptance

- [ ] Deleting a review session removes both the in-memory state AND the `{devRoot}/crow-reviews/<cloneDirName>` directory.
- [ ] Three regression tests added (removes clone, idempotent after partial failure, doesn't touch reviews dir for non-review sessions).
- [ ] Manual smoke: review a PR, observe `~/Dev/crow-reviews/<name>` exists, delete the review session, observe directory gone.
- [ ] No regression on regular session delete behavior.
- [ ] Cleanup failure (e.g., permission error on `removeItem`) surfaces in `sessionDeletionError[id]` and allows retry, same as today.

## Out of scope

- A bulk "purge orphaned crow-reviews/* clones" admin tool (file separately if needed — this ticket just fixes the going-forward case).
- Changing how review-session clones are created (the bug is in delete, not create).
- Cleaning up the `crow-reviews/` parent directory if empty.
- Cross-session shared clones / deduping review clones across PRs.

## Cross-links

- `Sources/Crow/App/SessionService.swift:540` (deleteSession)
- `Sources/Crow/App/SessionService.swift:1128` (review clone create)
- `Sources/Crow/App/Scaffolder.swift:31-33` (`crow-reviews` directory initialization)

## Backfill note

Existing orphaned directories in `crow-reviews/` from previously-deleted review sessions will remain on disk after this fix lands. A one-shot manual cleanup or follow-up purge-tool ticket may be warranted depending on volume. Document in PR description: "this fix is forward-looking; existing orphan clones need manual rm."

dhilgaertner:
## Summary

- Deleting a review session left `{devRoot}/crow-reviews/<cloneDirName>` orphaned on disk. Over time these full clones accumulate and grow disk usage.
- Root cause was different from the ticket's hypothesis: review clones **are** registered as `SessionWorktree`s, but `performDiskCleanup` checked `item.isMainCheckout` before the `isReview` branch. Review-session worktrees always have `repoPath == worktreePath == clonePath`, so `SessionWorktree.isMainRepoCheckout` returns `true` and the loop hit `continue` before the clone-removal code ran.
- Fix is a localized swap of the two early-exit branches in `performDiskCleanup` so `if isReview` runs first. Review clones are standalone `git clone` checkouts (not `git worktree add` artifacts) and don't need the main-checkout protection.
- Non-review session cleanup is unchanged.
- Existing retry-on-failure UX (`sessionDeletionError[id]`) is preserved.

## Test plan

- [x] `swift test --no-parallel` — 107 tests pass (full suite is green serially; the `FeatureFlags` suite has a pre-existing parallel env-var ordering flake unrelated to this change).
- [x] `swift build` — clean.
- [x] New tests in `Tests/CrowTests/SessionServiceDiskCleanupTests.swift`:
  - `deletesReviewClone` — confirms the clone dir is removed.
  - `reviewCleanupIsIdempotent` — no error when the dir is already gone.
  - `nonReviewMainCheckoutIsSkipped` — regression guard for the main-checkout protection on regular sessions.
  - `reviewCleanupDoesNotRemoveParent` — enforces "only the per-session subdir is removed, never `crow-reviews/`".
- [ ] Manual smoke: trigger a review on a real PR, confirm `~/Dev/crow-reviews/<name>` exists, delete the review session from the UI, confirm the directory is gone and the row disappears.
- [ ] Regression smoke: delete a regular (non-review) work session; confirm worktree-remove behavior is unchanged.

## Notes

- **Forward-looking only.** Existing orphaned directories under `crow-reviews/` from previously-deleted review sessions need a one-shot manual `rm`. A bulk purge tool is out of scope per the ticket.

Closes #305

### DIFF
--- a/Sources/Crow/App/SessionService.swift
+++ b/Sources/Crow/App/SessionService.swift
@@ -630,11 +630,9 @@ final class SessionService {
         var firstFatalError: String? = nil
 
         for item in items {
-            if item.isMainCheckout {
-                NSLog("Skipping worktree cleanup for main checkout: \(item.worktreePath) (branch: \(item.branch))")
-                continue
-            }
-
+            // Review clones are standalone `git clone` checkouts (not `git worktree add`
+            // artifacts) and always have repoPath == worktreePath, which would otherwise
+            // trip the main-checkout guard below and leave the clone orphaned on disk.
             if isReview {
                 guard FileManager.default.fileExists(atPath: item.worktreePath) else { continue }
                 do {
@@ -645,6 +643,11 @@ final class SessionService {
                     NSLog("[SessionService] \(msg) (\(item.worktreePath))")
                     if firstFatalError == nil { firstFatalError = msg }
                 }
+                continue
+            }
+
+            if item.isMainCheckout {
+                NSLog("Skipping worktree cleanup for main checkout: \(item.worktreePath) (branch: \(item.branch))")
                 continue
             }
 
--- a/Tests/CrowTests/SessionServiceDiskCleanupTests.swift
+++ b/Tests/CrowTests/SessionServiceDiskCleanupTests.swift
@@ -1,0 +1,91 @@
+import Foundation
+import Testing
+@testable import Crow
+
+/// Covers the disk-cleanup branch hit when a review session is deleted.
+/// Regression guard for #305: review clones were being skipped because
+/// `isMainCheckout` was checked before the `isReview` branch.
+@Suite("SessionService.performDiskCleanup")
+struct SessionServiceDiskCleanupTests {
+
+    private static func makeTempDir(name: String = "crow-cleanup-test") -> URL {
+        let url = URL(fileURLWithPath: NSTemporaryDirectory())
+            .appendingPathComponent("\(name)-\(UUID().uuidString)")
+        try? FileManager.default.createDirectory(at: url, withIntermediateDirectories: true)
+        return url
+    }
+
+    @Test func deletesReviewClone() {
+        let clone = Self.makeTempDir(name: "review-clone")
+        defer { try? FileManager.default.removeItem(at: clone) }
+
+        // Matches how `deleteSession` builds the item for review sessions:
+        // repoPath == worktreePath == clonePath, isMainCheckout: true (because
+        // SessionWorktree.isMainRepoCheckout returns true when the paths match).
+        let item = SessionService.WorktreeCleanupItem(
+            repoPath: clone.path,
+            worktreePath: clone.path,
+            branch: "feature/some-pr",
+            isMainCheckout: true
+        )
+
+        let error = SessionService.performDiskCleanup(items: [item], isReview: true)
+
+        #expect(error == nil)
+        #expect(!FileManager.default.fileExists(atPath: clone.path))
+    }
+
+    @Test func reviewCleanupIsIdempotent() {
+        let clone = Self.makeTempDir(name: "review-clone-gone")
+        try? FileManager.default.removeItem(at: clone)
+        #expect(!FileManager.default.fileExists(atPath: clone.path))
+
+        let item = SessionService.WorktreeCleanupItem(
+            repoPath: clone.path,
+            worktreePath: clone.path,
+            branch: "feature/some-pr",
+            isMainCheckout: true
+        )
+
+        let error = SessionService.performDiskCleanup(items: [item], isReview: true)
+
+        #expect(error == nil)
+    }
+
+    @Test func nonReviewMainCheckoutIsSkipped() {
+        let checkout = Self.makeTempDir(name: "main-checkout")
+        defer { try? FileManager.default.removeItem(at: checkout) }
+
+        let item = SessionService.WorktreeCleanupItem(
+            repoPath: checkout.path,
+            worktreePath: checkout.path,
+            branch: "main",
+            isMainCheckout: true
+        )
+
+        let error = SessionService.performDiskCleanup(items: [item], isReview: false)
+
+        #expect(error == nil)
+        #expect(FileManager.default.fileExists(atPath: checkout.path))
+    }
+
+    @Test func reviewCleanupDoesNotRemoveParent() {
+        let parent = Self.makeTempDir(name: "reviews-parent")
+        defer { try? FileManager.default.removeItem(at: parent) }
+        let clone = parent.appendingPathComponent("repo-pr-123")
+        try? FileManager.default.createDirectory(at: clone, withIntermediateDirectories: true)
+
+        let item = SessionService.WorktreeCleanupItem(
+            repoPath: clone.path,
+            worktreePath: clone.path,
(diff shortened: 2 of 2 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Root Cause — Verified
The PR's revised root-cause analysis is accurate. I traced it end to end:
- Review sessions register their worktree with `repoPath == worktreePath == clonePath` (`SessionService.swift:1200-1201`).
- `SessionWorktree.isMainRepoCheckout` returns `true` whenever `worktreePath == repoPath` (`Packages/CrowCore/.../Worktree.swift:55-60`).
- The old ordering checked `item.isMainCheckout` *before* the `isReview` branch, so every review clone hit `continue` and was never removed — the orphaned-directory bug in #305.

The fix simply swaps the two early-exit branches so `isReview` runs first. This is the correct, minimal change.

### Security Review
**Strengths:**
- The removal target is always the per-PR clone dir `{devRoot}/crow-reviews/{repo}-pr-{n}` (`SessionService.swift:1147-1148`), never the user's working repo, so the direct `FileManager.removeItem` is bounded and safe.
- `reviewCleanupDoesNotRemoveParent` explicitly guards against ever deleting the shared `crow-reviews/` parent.
- Existence guard makes cleanup idempotent; failures propagate as `firstFatalError`, preserving the retry UX (`sessionDeletionError[id]`).

**Concerns:** None. No new input, no shell interpolation, no credential handling introduced.

### Code Quality
- Behavior for non-review sessions is provably unchanged: when `isReview == false` the new first block is skipped and control falls through to the unchanged `isMainCheckout` guard and `git worktree remove` path.
- Branch deletion is correctly skipped for review clones — a standalone clone has no parent repo to prune branches from; removing the directory removes everything.
- The relocated comment accurately documents *why* the order matters.
- Tests are well-targeted: happy path, idempotency, the non-review regression guard, and the parent-dir safety check.

### Notes
- Verified `performDiskCleanup` has a single caller (`deleteSession`), so the reordering has no other blast radius.
- Could not build/run `Tests/CrowTests` locally — the `GhosttyKit.xcframework` binary artifact is absent from the checkout (environment limitation, not a PR defect). `CrowCore`'s 169 tests pass, including the `isMainRepoCheckout` cases the fix relies on.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | None |
| Yellow | None |
| Green | `isReview` is applied to *all* of a session's worktree items; harmless today since review sessions only ever have the single clone, but it's a latent assumption worth a one-line comment if review sessions ever gain extra worktrees. |

**Recommendation:** Approve. Correct root-cause fix, minimal blast radius, good regression coverage, and the forward-only scope is clearly documented.

PR #319, opened 2026-05-22
fix: skip auto-respond automation on review sessions

### LINKED ISSUE #311: Auto-fix automation runs on review sessions and makes commits on the branch being reviewed
## Description

When a review session is started — either auto-kicked-off or launched manually — and the review surfaces fixes that need to be made, the **auto-fix / auto-respond automation** (the one that responds to changes-requested / failed checks by committing fixes) is firing on the review session itself.

The result is that Crow ends up making commits on the branch it's supposed to be reviewing, on behalf of the reviewer. That's wrong: when you are reviewing someone else's work, you should be leaving comments, not pushing code.

## Expected

- Review sessions must never trigger the auto-fix-on-failed-review or auto-fix-on-failed-checks automation.
- Those automations should remain enabled for normal feature/work sessions, where the session owns the branch and is expected to push fixes.
- The gating should be based on session type (review vs. work), not on the user disabling automation per-session.

## Likely places to look

- `autoRespond.respondToChangesRequested` and `autoRespond.respondToFailedChecks` in config.json — these are global today and don't appear to distinguish session kind.
- Session creation path for review sessions (`/review-pr`, `/ultrareview`, scheduled auto-review hooks) — should mark the session so auto-respond skips it.
- Auto-respond trigger logic — should early-return when the active session is a review session.

## Acceptance criteria

- A review session that encounters changes-requested or failed checks does **not** commit or push to the branch under review.
- Normal (non-review) sessions continue to auto-respond as configured.
- Existing review sessions are correctly classified after the fix (or the fix is forward-only and documented).

dhilgaertner:
## Summary

- Gate `AutoRespondCoordinator.handle(_:)` on `Session.kind` so review sessions never trigger the auto-fix prompts.
- Review sessions exist to leave comments on someone else's PR; auto-respond would have Crow committing and pushing on the reviewer's behalf to the branch under review.
- Work sessions are untouched — they continue to auto-respond per the existing `respondToChangesRequested` / `respondToFailedChecks` toggles.
- The macOS notification still fires for review sessions, so the user can act manually if they choose.

## Why this is small

The existing `SessionKind` enum (`.work` / `.review`) and the persisted `Session.kind` field already carry the policy. `SessionService.createReviewSession()` already stamps `kind: .review`. No schema change, no migration — existing review sessions on disk are already classified correctly.

## Out of scope

- Manual quick-action buttons (`dispatchManual`) remain enabled on all sessions: a button click is the user's explicit consent.
- Per-session config overrides for auto-respond.

Closes #311

## Test plan

- [x] `make build` — clean
- [x] `swift test --filter AutoRespondCoordinatorReviewSessionTests` — 4/4 pass
- [x] Full `swift test` — 117/117 pass
- [ ] Manual smoke (optional, live PR): start `/review-pr <url>` with both auto-respond toggles enabled, trigger a "changes requested" review on that PR, confirm no prompt is injected into the review session's Claude Code terminal and no commits land on the branch under review. Repeat against a normal work session and confirm auto-respond still fires.

### DIFF
--- a/Sources/Crow/App/AutoRespondCoordinator.swift
+++ b/Sources/Crow/App/AutoRespondCoordinator.swift
@@ -29,6 +29,11 @@ final class AutoRespondCoordinator {
     func handle(_ transitions: [PRStatusTransition]) {
         let cfg = settingsProvider()
         for t in transitions {
+            if shouldSkipReviewSession(t) {
+                NSLog("[AutoRespond] Skipping %@ for session %@: review session",
+                      t.kind.rawValue, t.sessionID.uuidString)
+                continue
+            }
             switch t.kind {
             case .changesRequested where cfg.respondToChangesRequested:
                 dispatch(t)
@@ -38,6 +43,14 @@ final class AutoRespondCoordinator {
                 continue
             }
         }
+    }
+
+    /// Review sessions exist to review someone else's PR. Auto-respond would
+    /// have Crow committing on behalf of the reviewer to the branch under
+    /// review — never correct. This gate is policy, independent of the
+    /// `AutoRespondSettings` toggles, and applies even when both toggles are on.
+    func shouldSkipReviewSession(_ transition: PRStatusTransition) -> Bool {
+        appState.sessions.first(where: { $0.id == transition.sessionID })?.kind == .review
     }
 
     private func dispatch(_ transition: PRStatusTransition) {
--- a/Tests/CrowTests/AutoRespondCoordinatorReviewSessionTests.swift
+++ b/Tests/CrowTests/AutoRespondCoordinatorReviewSessionTests.swift
@@ -1,0 +1,72 @@
+import Foundation
+import Testing
+import CrowCore
+@testable import Crow
+
+/// Verifies the policy gate added for #311: review sessions must never
+/// trigger auto-respond automation, regardless of the global
+/// `respondToChangesRequested` / `respondToFailedChecks` toggles. Work
+/// sessions continue to flow through unaffected.
+@Suite("AutoRespondCoordinator review-session gate")
+@MainActor
+struct AutoRespondCoordinatorReviewSessionTests {
+
+    private func makeCoordinator(sessions: [Session]) -> AutoRespondCoordinator {
+        let state = AppState()
+        state.sessions = sessions
+        return AutoRespondCoordinator(
+            appState: state,
+            settingsProvider: { AutoRespondSettings(respondToChangesRequested: true, respondToFailedChecks: true) }
+        )
+    }
+
+    @Test func skipsReviewSessionOnChangesRequested() {
+        let review = Session(name: "review-crow-123", kind: .review)
+        let coord = makeCoordinator(sessions: [review])
+        let t = PRStatusTransition(
+            kind: .changesRequested,
+            sessionID: review.id,
+            prURL: "https://github.com/radiusmethod/crow/pull/123",
+            prNumber: 123
+        )
+        #expect(coord.shouldSkipReviewSession(t))
+    }
+
+    @Test func skipsReviewSessionOnChecksFailing() {
+        let review = Session(name: "review-crow-456", kind: .review)
+        let coord = makeCoordinator(sessions: [review])
+        let t = PRStatusTransition(
+            kind: .checksFailing,
+            sessionID: review.id,
+            prURL: "https://github.com/radiusmethod/crow/pull/456",
+            prNumber: 456
+        )
+        #expect(coord.shouldSkipReviewSession(t))
+    }
+
+    @Test func doesNotSkipWorkSession() {
+        let work = Session(name: "feature-crow-789", kind: .work)
+        let coord = makeCoordinator(sessions: [work])
+        let t = PRStatusTransition(
+            kind: .changesRequested,
+            sessionID: work.id,
+            prURL: "https://github.com/radiusmethod/crow/pull/789",
+            prNumber: 789
+        )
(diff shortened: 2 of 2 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
This PR is itself a defensive hardening: it prevents Crow from committing/pushing to a branch under review on the reviewer's behalf when a review session's tracked PR transitions to *changes requested* or *checks failing*. That is the correct trust boundary — a reviewer's automation should never mutate someone else's PR branch.

**Strengths:**
- The gate (`shouldSkipReviewSession`) is placed at the single auto-respond entry point (`handle(_:)` → `AutoRespondCoordinator`, called only from `AppDelegate.swift:517`), *before* the kind switch. `dispatch(t)` is unreachable except through this gate, so every current and future transition kind is covered.
- The gate is policy, independent of the `respondToChangesRequested` / `respondToFailedChecks` toggles — review sessions are skipped even with both toggles on. Verified by tests.
- Notification still fires upstream (`AppDelegate.swift:513-516`) before `handle()`, so the reviewer is still alerted and can act manually. Matches the PR description.
- Unknown/stale session IDs (race with delete) correctly fall through to the normal path instead of being treated as reviews (`?.kind == .review` is `false` on `nil`). Explicitly tested.
- No new inputs, scopes, secrets, or external calls. Logging emits only the transition kind and session UUID — nothing sensitive.

**Concerns:**
- None blocking.

### Code Quality
- Classification source is sound: review sessions are stamped `kind: .review` at creation (`SessionService.swift:1191`), and `Session` decodes a missing `kind` as `.work` for backward compatibility — so existing on-disk sessions classify correctly with no migration.
- `dispatchManual` is intentionally left ungated, which is a reasonable design call: a button click is explicit user consent. Documented in the PR's *Out of scope* section.
- Minor (non-blocking): `shouldSkipReviewSession` and `dispatch` each do their own `appState.sessions.first(where:)` lookup for the same session. Transitions are rare and lists small, so the O(n) duplication is immaterial — not worth changing.
- Minor (non-blocking): the tests exercise the `shouldSkipReviewSession` predicate directly rather than asserting end-to-end that `handle()` performs no terminal send. The predicate is the whole decision and is simple, so this is adequate coverage; an end-to-end `handle()` test would need a terminal-send spy.

### Build / Test Verification
Could not run `swift build` / `swift test` in this review clone: it fails on a missing `Frameworks/GhosttyKit.xcframework` binary artifact, which is a submodule-built dependency (`make ghostty`) absent from the clone — an environment limitation, not a defect in this PR. The PR test plan documents `make build` clean and full `swift test` 117/117 passing, including the 4 new gate tests.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | None |
| Yellow | None |
| Green | Consider an end-to-end `handle()` test asserting no send for review sessions; dedupe the two session lookups |

**Recommendation:** Approve. The fix is small, correctly placed at the chokepoint, independent of user toggles, backward-compatible with no migration, and well-tested. The only items are optional polish.

---

dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- This change is itself a security/correctness improvement. Without the gate, an enabled `respondToChangesRequested` / `respondToFailedChecks` toggle would let Crow inject prompts that commit and push to the branch *under review* — i.e., act on someone else's PR on the reviewer's behalf. Gating on `Session.kind == .review` closes that.
- The gate is enforced as policy independent of the user toggles (`shouldSkipReviewSession` is checked before the toggle `switch`), so it cannot be re-enabled by misconfiguration.
- No secrets, injection, or auth surface introduced. Logging (`NSLog`) emits only the transition kind and session UUID — nothing sensitive.

**Concerns:**
- None.

### Code Quality
- **Placement is correct.** The gate runs once per transition before the `switch`, so it covers both `.changesRequested` and `.checksFailing` and any future transition kinds automatically (`AutoRespondCoordinator.swift:32`).
- **Stale/unknown session handled safely.** `appState.sessions.first(where:)?.kind == .review` returns `false` for a missing session, so a delete/poll race falls through to the normal path rather than wrongly suppressing a work session. This edge case is explicitly tested (`unknownSessionIsNotTreatedAsReview`).
- **Notification path preserved.** `notifyPRTransition` fires in `AppDelegate.swift:514` before `handle(_:)` is invoked, so review sessions still surface the macOS notification for manual action — matching the PR description.
- **Out-of-scope paths intact.** `dispatchManual` (quick-action buttons) is intentionally left enabled — a click is explicit consent. Reasonable.
- Minor (consider, non-blocking): the new tests assert the `shouldSkipReviewSession` predicate directly rather than driving the full `handle(_:)` skip path. Since `handle`'s side effect requires a live managed terminal, testing the predicate is a fair proxy and the predicate is the whole policy — acceptable as-is.

### Note on verification
- Could not run `swift build` / `swift test` in this review clone: it fails with `local binary target 'GhosttyKit' ... does not contain a binary artifact`. This is an environment limitation (missing xcframework binary) and is unrelated to this PR, which only touches `AutoRespondCoordinator.swift` and a new test file. Author reports `make build` clean and `swift test` 117/117 passing.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | None |
| Yellow | None |
| Green | Consider exercising the full `handle(_:)` skip path in a future test if a managed-terminal test fixture becomes available |

**Recommendation:** Approve. Small, well-scoped, correctly placed gate with good comments and unit coverage of the key cases (both transition kinds, work session, stale session). It fixes a real correctness/security issue (#311) without touching the work-session path or schema.

---

Issue #299, opened 2026-05-21
Auto-merge PRs when crow:merge label is set, but only on Crow-authored PRs


dgershman:
## Goal

Add a new repo label `crow:merge` that opts a PR into auto-merge **once the GitHub "Merge PR" button is available** (i.e. all required reviews approved and all required checks passing). Crow watches the PR and merges it as soon as it becomes eligible.

**Critical guard:** only act on PRs that **Crow itself authored**. Crow-reviewed-only PRs (where Crow ran a review skill against someone else's branch but didn't write the commits) must be ignored — auto-merging someone else's work on Crow's signal would be a footgun.

## Trigger

A PR is eligible for Crow auto-merge when **all** of these hold:

1. PR has the `crow:merge` label.
2. PR was authored by Crow — detected via the `Crow-Session: <uuid>` commit trailer that `setup.sh` writes into every Crow-orchestrated commit (`.claude/skills/crow-workspace/setup.sh:352`; `attributionTrailers` defaults to `true`). If **any** commit on the PR's branch carries a `Crow-Session:` trailer, treat the PR as Crow-authored. If none do, ignore the label entirely.
3. PR is mergeable in GitHub's eyes — `mergeable == true`, `mergeStateStatus == CLEAN` (or `HAS_HOOKS`), required reviews satisfied, required checks all passing.

When all three hold → merge.

## Implementation sketch

There's already a sticky-plan-comment / status-watcher pattern in Crow for the workspace skill's plan post-back. The auto-merge watcher should slot into the same poll loop.

For each open PR Crow knows about (its session has a `pr_url` link), on each poll:

```text
if pr.labels contains "crow:merge"
  and any(commit.message contains "Crow-Session: " for commit in pr.commits)
  and pr.mergeable and pr.mergeStateStatus in (CLEAN, HAS_HOOKS)
  and pr.reviewDecision == APPROVED (when reviews required)
then
  gh pr merge {pr_url} --auto --{method}   # let GH wait if hooks still pending
  log + emit notification "Auto-merging PR #N (crow:merge)"
fi
```

Notes:

- Prefer GitHub's native **auto-merge** (`gh pr merge --auto`) over a busy-poll-then-merge. Native auto-merge queues the merge server-side and fires once checks settle; we don't need to babysit the poll. Crow's role is just to *enable* it once eligibility is met, then walk away.
- One-shot per PR: once auto-merge is enabled on a PR, don't re-enable on subsequent polls. Track in the session metadata.
[...]

PR #320, opened 2026-05-22
feat: opt-in toggle for crow:auto label automation

### LINKED ISSUE #312: Add opt-in setting for crow-auto label automation (default off)
## Description

We have a `crow-auto` feature: when an issue or PR is tagged with the `crow-auto` label, the Manager session auto-kicks off a workspace via `/crow-workspace`. Right now this behavior is implicit — there's no UI control to turn it on or off.

Add a setting in **Settings → Automation** that controls whether the Manager listens for the `crow-auto` label and auto-launches workspaces. The setting should default to **off**, so the feature is opt-in.

## UX

- **Location:** Settings → Automation tab, alongside the existing auto-respond toggles (`respondToChangesRequested`, `respondToFailedChecks`).
- **Label:** "Auto-launch workspaces for `crow-auto` labeled issues" (or similar).
- **Default:** off.
- **Help text:** brief explanation that when enabled, the Manager will detect new issues/PRs tagged `crow-auto` and run `/crow-workspace` automatically. Note that this requires the Manager to be running.

## Config

- Add a new field to `config.json` under an appropriate key (e.g. `autoRespond.respondToCrowAutoLabel` or a dedicated `crowAuto.enabled`). Default value is `false`.
- The Manager's polling / event-handling code should early-return when the flag is false.
- Existing users (who currently have the feature implicitly on, if any) should see no behavior change *unless* they explicitly opt in — i.e., the migration is "default off for everyone, including current installs."

## Acceptance criteria

- A new toggle exists in Settings → Automation, off by default.
- Flipping it on enables the `crow-auto` label automation; flipping it off disables it.
- The setting persists to `config.json` and survives app restart.
- No `crow-auto` workspace is launched while the toggle is off, regardless of how the label is applied (manually, via webhook, by another bot).
- Help text in the UI makes the behavior unambiguous.

dhilgaertner:
Closes #312.

## Summary

- New top-level `autoCreateWatcherEnabled: Bool` on `AppConfig`, default `false` — covers all `crow:auto`-label automation.
- New "Auto-launch workspaces" section in **Settings → Automation** (placed above "Auto-merge") with the toggle and help text.
- `IssueTracker.detectAutoCreateCandidates` early-returns when the toggle is off, so no `/crow-workspace` dispatch happens regardless of how the label was applied (manual, webhook, bot). The label is intentionally left in place while disabled so a later opt-in still picks up previously-labeled issues.
- Mirrors the `autoMergeWatcherEnabled` (CROW-299) pattern: closure provider on the tracker, wired from `AppDelegate`, so flipping the toggle takes effect on the next poll without re-init.
- Legacy `config.json` files decode with the field defaulting to `false`, satisfying the "default off for everyone, including current installs" requirement.

## Test plan

- [x] `make build` succeeds.
- [x] `swift test --package-path Packages/CrowCore` passes (171 tests, including 3 new ones: round-trip, default-off-on-missing-key, extended empty-JSON check).
- [x] `swift test --filter IssueTrackerAutoMergeTests` still green (no regression to the sibling watcher).
- [ ] Manual smoke:
  - Settings → Automation shows the new "Auto-launch workspaces" section above "Auto-merge"; toggle is off.
  - With toggle **off**: applying `crow:auto` to an assigned issue does not launch a workspace and the label remains. Verified by waiting through one 60s poll cycle.
  - Toggle on: next poll launches the workspace and strips the label (pre-change behavior).
  - Toggle state persists across app restart (written to `~/.claude/config.json`).

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -23,6 +23,13 @@ public struct AppConfig: Codable, Sendable, Equatable {
     /// authored by Crow (Crow-Session trailer matching a known session).
     /// Opt-in: defaults to false (CROW-299).
     public var autoMergeWatcherEnabled: Bool
+    /// When true, the IssueTracker dispatches `/crow-workspace` to the
+    /// Manager terminal for assigned open issues labeled `crow:auto`.
+    /// Opt-in: defaults to false (CROW-312). The label is still stripped
+    /// after a successful dispatch so the trigger remains one-shot per
+    /// issue. While disabled, the label is left alone so a later opt-in
+    /// can still pick up previously-labeled issues.
+    public var autoCreateWatcherEnabled: Bool
     public var cleanup: CleanupConfig
 
     public init(
@@ -36,6 +43,7 @@ public struct AppConfig: Codable, Sendable, Equatable {
         autoRespond: AutoRespondSettings = AutoRespondSettings(),
         attributionTrailers: Bool = true,
         autoMergeWatcherEnabled: Bool = false,
+        autoCreateWatcherEnabled: Bool = false,
         cleanup: CleanupConfig = CleanupConfig()
     ) {
         self.workspaces = workspaces
@@ -48,6 +56,7 @@ public struct AppConfig: Codable, Sendable, Equatable {
         self.autoRespond = autoRespond
         self.attributionTrailers = attributionTrailers
         self.autoMergeWatcherEnabled = autoMergeWatcherEnabled
+        self.autoCreateWatcherEnabled = autoCreateWatcherEnabled
         self.cleanup = cleanup
     }
 
@@ -63,11 +72,12 @@ public struct AppConfig: Codable, Sendable, Equatable {
         autoRespond = try container.decodeIfPresent(AutoRespondSettings.self, forKey: .autoRespond) ?? AutoRespondSettings()
         attributionTrailers = try container.decodeIfPresent(Bool.self, forKey: .attributionTrailers) ?? true
         autoMergeWatcherEnabled = try container.decodeIfPresent(Bool.self, forKey: .autoMergeWatcherEnabled) ?? false
+        autoCreateWatcherEnabled = try container.decodeIfPresent(Bool.self, forKey: .autoCreateWatcherEnabled) ?? false
         cleanup = try container.decodeIfPresent(CleanupConfig.self, forKey: .cleanup) ?? CleanupConfig()
     }
 
     private enum CodingKeys: String, CodingKey {
-        case workspaces, defaults, notifications, sidebar, remoteControlEnabled, managerAutoPermissionMode, telemetry, autoRespond, attributionTrailers, autoMergeWatcherEnabled, cleanup
+        case workspaces, defaults, notifications, sidebar, remoteControlEnabled, managerAutoPermissionMode, telemetry, autoRespond, attributionTrailers, autoMergeWatcherEnabled, autoCreateWatcherEnabled, cleanup
     }
 }
 
--- a/Packages/CrowCore/Tests/CrowCoreTests/AppConfigTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AppConfigTests.swift
@@ -46,6 +46,7 @@ import Testing
     #expect(config.managerAutoPermissionMode == true)
     #expect(config.attributionTrailers == true)
     #expect(config.autoMergeWatcherEnabled == false)
+    #expect(config.autoCreateWatcherEnabled == false)
     #expect(config.cleanup.enabled == false)
     #expect(config.cleanup.retentionHours == 24)
 }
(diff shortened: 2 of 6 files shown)


dgershman (approved):
## Code & Security Review

Focused, well-scoped change that gates the existing `crow:auto` workspace automation behind a new opt-in toggle, defaulting **off** for all installs (CROW-312). It faithfully mirrors the `autoMergeWatcherEnabled` (CROW-299) pattern.

### Critical Issues
None.

### Security Review
**Strengths:**
- Safe-by-default: `autoCreateWatcherEnabled` decodes to `false` for legacy `config.json` files missing the key (`AppConfig.swift:75`), so no install silently gains the ability to auto-spawn workspaces on label.
- The guard sits at the single dispatch entry point — `detectAutoCreateCandidates` early-returns before any `onAutoCreateRequest` fire or label mutation (`IssueTracker.swift:419`), so it cannot be bypassed regardless of how the label was applied (manual, webhook, bot).
- Label stripping (`removeAutoCreateLabel`) passes `issue.url` / `issue.number` as discrete `gh`/`glab` argv entries, not shell-interpolated strings — no command-injection surface.
- Closure-provider wiring reads a fresh snapshot each poll, so disabling takes effect on the next cycle without leaving a stale enabled state.

**Concerns:**
- None.

### Code Quality
- Wiring is complete and consistent across every layer: model, decode, `CodingKeys`, `AppDelegate` provider, `SettingsView` binding, and the `AutomationSettingsView` toggle.
- The "leave the label in place while disabled" behavior is the right call and is clearly documented in both the field doc and the function doc — a later opt-in still picks up previously-labeled issues.
- Test coverage matches the sibling watcher: round-trip + default-off-on-missing-key, plus the empty-JSON default assertion. Verified locally: `swift test --package-path Packages/CrowCore --filter AppConfig` → 34/34 pass, including the two new cases.

### Minor (non-blocking)
- `detectAutoCreateCandidates(issues:config:)` no longer needs its `config` parameter — candidate filtering already happens at the call site (`IssueTracker.swift:306-310`) and the new guard uses the provider closure. This is pre-existing dead-parameter, not introduced here; safe to drop on a future pass.
- While disabled, stale-label cleanup for issues already picked up elsewhere is also skipped (same early-return). This is intentional and documented, just worth being aware of.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | None |
| Yellow | None |
| Green | Drop the now-unused `config` param on `detectAutoCreateCandidates` (pre-existing) |

**Recommendation:** Approve. The change is minimal, defaults safely off, is backward-compatible with existing config files, gates at the correct chokepoint, and is covered by tests that pass locally. No blocking or security issues.

dgershman (changes requested):
## Code & Security Review

Clean, well-scoped change that faithfully mirrors the `autoMergeWatcherEnabled` (CROW-299) pattern. The code itself is approve-quality; the one blocking item is documentation drift on the exact feature this PR changes.

### Critical Issues
None in the code.

### Security Review
**Strengths:**
- Pure feature-gating change. No new inputs, network calls, credentials, or shell construction. The new toggle only *prevents* an existing dispatch path (`onAutoCreateRequest` + label strip) from firing — strictly a privilege-reduction, fail-closed change (default `false`).
- The gate is fail-safe: provider defaults to `{ false }` on the tracker, the decode fallback defaults to `false`, and `AppDelegate` wiring uses `?? false`. A legacy/missing config key keeps automation off, which matches the stated "default off for everyone, including current installs" requirement.

**Concerns:**
- None.

### Code Quality
- `Sources/Crow/App/IssueTracker.swift:419` — guard placed at the top of `detectAutoCreateCandidates`, before the in-flight purge. Correct: when disabled nothing is added to `autoCreateInFlight`, and the purge self-heals on re-enable (it subtracts active sessions before dispatching). No leak.
- Default value is consistent across all four sites (init param, `decodeIfPresent` fallback, provider default, AppDelegate `?? false`). `CodingKeys` updated. 
- Tests cover the two behaviors that matter (round-trip + default-off-when-key-missing). `swift test --package-path Packages/CrowCore` → 171 passed locally.
- `AutomationSettingsView` has exactly one caller (`SettingsView.swift:42`), correctly updated. No orphaned previews.
- (Pre-existing, non-blocking) `detectAutoCreateCandidates(issues:config:)` still takes a `config` parameter it doesn't reference — the gate reads from the provider, not `config`. Not introduced by this PR, but worth a cleanup pass since you're in here.

### Blocking: documentation is now stale
`docs/automation.md` declares itself "the single source of truth for what each automation does, how to turn it on or off, and where the toggles live." After this PR it actively misdescribes the feature you changed:

- **Lifecycle, line 9:** "If it carries the `crow:auto` label, Crow auto-creates a workspace for it (#211)." — now conditional on a toggle that is **off by default**.
- **Section #211 (lines 65-67):** describes auto-create as unconditional on the next polling cycle, with no mention of the new opt-in. As written, a reader on a fresh install would expect labeling an issue to launch a workspace; it now won't.
- **Settings → Automation section (lines 53-59):** the sibling **Auto-merge** toggle (#299) got its own subsection here when it shipped. This PR adds a parallel "Auto-launch workspaces" toggle to the same tab but doesn't document it. Please add a matching subsection (and ideally an `AppConfig.autoCreateWatcherEnabled` backing reference, as the auto-merge entry does).

This is a one-paragraph fix, but it's the difference between the doc being right and wrong on a default-behavior change, so I'm flagging it as a must-fix rather than a nit — consistent with the precedent #299 set.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | `docs/automation.md` still describes `crow:auto` auto-create (#211) as unconditional; it's now opt-in/off-by-default. Update the lifecycle bullet, the #211 section, and add an "Auto-launch workspaces" toggle subsection mirroring #299. |
| Yellow | None |
| Green | Drop the unused `config` parameter from `detectAutoCreateCandidates` while in this file (pre-existing). |

**Recommendation:** Request Changes — code and tests are solid and ready to merge as-is, but the PR changes a documented default and leaves `docs/automation.md` (the self-declared single source of truth) incorrect. Update the docs and this is a clean approve.

---

PR #323, opened 2026-05-22
feat: scheduled jobs — timed prompt sets scoped to a repo (closes #317)

### LINKED ISSUE #317: Crow Jobs: scheduled sets of timed prompts (new Settings section, scoped to a repo)
## Summary
Add a concept of **Jobs**: named sets of timed prompts that kick off automatically on a schedule.
Jobs live in a new "Jobs" section under Settings and are scoped to a repo.

## Background
There is no scheduling/cron concept in Crow today. Prompts are delivered to a session's terminal via
the `send` RPC → `TerminalManager.shared.send(id:text:)` (`Sources/Crow/App/AppDelegate.swift`), and
sessions/terminals are created programmatically (new-session / new-terminal RPCs). Settings is a
`TabView` in `Packages/CrowUI/.../SettingsView.swift`, backed by `AppConfig`
(`CrowCore/Models/AppConfig.swift`) and persisted by `ConfigStore`. Repos are scoped through the
`workspaces` array (`WorkspaceInfo`, `alwaysInclude`).

## Ask
- New **Jobs** section under Settings (new tab/section in `SettingsView`), with a form to define jobs
  (mirror `WorkspaceFormView`).
- A **Job** = one or more prompts plus a schedule (interval and/or cron-like time). Scoped to a repo
  (and its workspace).
- New `JobConfig` model in `AppConfig`, persisted via `ConfigStore` to `{devRoot}/.claude/config.json`.
- A scheduler (Timer / `DispatchSourceTimer`) that, when a job fires, spins up a session + terminal
  in the scoped repo's worktree and sends the job's prompt(s).
- Surface job runs in the UI (last run / next run); allow enable/disable per job.

## Acceptance
- A user can define a repo-scoped job of one or more prompts on a schedule from Settings → Jobs.
- The job fires on schedule, launches Claude Code in the scoped repo, and sends the prompt(s).
- Jobs persist across app restarts and can be enabled/disabled.

dgershman:
Closes #317

## Summary
Adds **Crow Jobs**: named, repo-scoped sets of prompts that fire automatically on a schedule. Jobs live in a new **Settings → Jobs** tab, persist in `{devRoot}/.claude/config.json`, and can be enabled/disabled with last-run / next-run shown in the UI.

When a job fires, the scheduler creates a **fresh worktree + session + Claude Code terminal** in the scoped repo (off `origin/main`) and sends the job's prompts in order.

## What's included
- **Model** (`CrowCore`): `JobConfig` + `JobSchedule` (interval **or** daily-at-`HH:MM` on selected weekdays), forward-compatible `Codable`; added `jobs` to `AppConfig`; new `SessionKind.job`.
- **Scheduler** (`JobScheduler`): a `Timer`-driven evaluator modeled on `IssueTracker`. Fires due jobs, persists `lastRunAt` (so restarts don't replay), and delivers follow-up prompts once the terminal reports `.claudeLaunched`. Wired into `AppDelegate` (start in `launchMainApp`, stop on terminate).
- **Session** (`SessionService.runJob`): mirrors `createReviewSession` — creates a real git worktree via `GitManager.createWorktree`, writes the first prompt to `.crow-job-prompt.md`, builds a `.job` session + managed terminal, and arms auto-launch. `launchClaude` generalized to dispatch the initial prompt file for `.job` (and `.review`) kinds.
- **UI**: Settings **Jobs** tab + `JobFormView` (workspace/repo pickers, editable prompt list, interval/daily schedule editor, enable toggle), mirroring `WorkspaceFormView`.

## Design decisions
- **Schedule**: interval or daily-at-time (no raw cron parser — covers the common cases with a validatable form).
- **Worktree**: fresh worktree + session per run, matching the normal session pattern; runs accumulate as ordinary sessions you can review and clean up (existing `CleanupConfig` applies).
- Multi-prompt delivery is best-effort timed sends (first prompt launches Claude; the rest follow spaced by a fixed gap).

## Acceptance criteria
- [x] Define a repo-scoped job of one or more prompts on a schedule from Settings → Jobs.
- [x] The job fires on schedule, launches Claude Code in the scoped repo, and sends the prompt(s).
- [x] Jobs persist across app restarts and can be enabled/disabled.

## Testing
- New `JobConfigTests`: `JobConfig`/`JobSchedule` round-trip (interval + daily), `nextRunDate` for interval and weekday skipping, forward-compat decode (old config → `jobs: []`), name validation.
- `swift build` (arm64) succeeds; full test suites pass: **CrowCore 177**, **CrowUI 31**, root **113**.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -31,6 +31,9 @@ public struct AppConfig: Codable, Sendable, Equatable {
     /// can still pick up previously-labeled issues.
     public var autoCreateWatcherEnabled: Bool
     public var cleanup: CleanupConfig
+    /// Scheduled jobs: named sets of prompts that fire automatically on a
+    /// schedule, scoped to a repo. Driven by `JobScheduler` (CROW-317).
+    public var jobs: [JobConfig]
 
     public init(
         workspaces: [WorkspaceInfo] = [],
@@ -44,7 +47,8 @@ public struct AppConfig: Codable, Sendable, Equatable {
         attributionTrailers: Bool = true,
         autoMergeWatcherEnabled: Bool = false,
         autoCreateWatcherEnabled: Bool = false,
-        cleanup: CleanupConfig = CleanupConfig()
+        cleanup: CleanupConfig = CleanupConfig(),
+        jobs: [JobConfig] = []
     ) {
         self.workspaces = workspaces
         self.defaults = defaults
@@ -58,6 +62,7 @@ public struct AppConfig: Codable, Sendable, Equatable {
         self.autoMergeWatcherEnabled = autoMergeWatcherEnabled
         self.autoCreateWatcherEnabled = autoCreateWatcherEnabled
         self.cleanup = cleanup
+        self.jobs = jobs
     }
 
     public init(from decoder: Decoder) throws {
@@ -74,10 +79,11 @@ public struct AppConfig: Codable, Sendable, Equatable {
         autoMergeWatcherEnabled = try container.decodeIfPresent(Bool.self, forKey: .autoMergeWatcherEnabled) ?? false
         autoCreateWatcherEnabled = try container.decodeIfPresent(Bool.self, forKey: .autoCreateWatcherEnabled) ?? false
         cleanup = try container.decodeIfPresent(CleanupConfig.self, forKey: .cleanup) ?? CleanupConfig()
+        jobs = try container.decodeIfPresent([JobConfig].self, forKey: .jobs) ?? []
     }
 
     private enum CodingKeys: String, CodingKey {
-        case workspaces, defaults, notifications, sidebar, remoteControlEnabled, managerAutoPermissionMode, telemetry, autoRespond, attributionTrailers, autoMergeWatcherEnabled, autoCreateWatcherEnabled, cleanup
+        case workspaces, defaults, notifications, sidebar, remoteControlEnabled, managerAutoPermissionMode, telemetry, autoRespond, attributionTrailers, autoMergeWatcherEnabled, autoCreateWatcherEnabled, cleanup, jobs
     }
 }
 
--- a/Packages/CrowCore/Sources/CrowCore/Models/Enums.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/Enums.swift
@@ -4,6 +4,7 @@ import Foundation
 public enum SessionKind: String, Codable, Sendable {
     case work    // Normal development session (default)
     case review  // PR review session
+    case job     // Session spun up by a scheduled job (CROW-317)
 }
 
 /// Status of a development session.
--- a/Packages/CrowCore/Sources/CrowCore/Models/JobConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/JobConfig.swift
@@ -1,0 +1,164 @@
+import Foundation
+
+/// A scheduled job: one or more prompts that fire automatically on a schedule,
+/// scoped to a repo within a workspace.
+///
+/// When a job fires, the scheduler creates a fresh worktree + session + Claude
+/// Code terminal in `{devRoot}/{workspace}/{repo}` and sends `prompts` in order.
+/// Persisted as part of `AppConfig` in `{devRoot}/.claude/config.json`.
+///
+/// Like `WorkspaceInfo`, decoding is forward-compatible: missing keys fall back
+/// to defaults so older config files keep working.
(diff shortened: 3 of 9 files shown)


dhilgaertner (approved):
## Code & Security Review

Clean, well-scoped implementation. Mirrors existing patterns (`IssueTracker`, `createReviewSession`), persists `lastRunAt` outside the scheduler to avoid a second source of truth, and ships meaningful tests for the model and scheduling logic.

### Security Review

**Strengths:**
- `GitManager.createWorktree` uses `Process` with an argv array — no shell injection in worktree creation, even though the branch name embeds a slug.
- `slugify` (`Sources/Crow/App/SessionService.swift:1232-1244`) strips everything but alphanumerics before going into the branch/worktree path, so the job name itself can't smuggle in path or shell metacharacters.
- Name validation is case-insensitive and forward-compatible decoding falls back to safe defaults (`AppConfig`, `JobConfig`) — old configs keep working.
- `JobScheduler.inFlight` guards against double-firing within a tick before `lastRunAt` persists.
- `onJobRan` is called *before* `deliverRemainingPrompts`, so a multi-prompt failure can't cause a replay on the next tick.

**Concerns (non-blocking):**
- **Pre-existing, slightly widened attack surface:** `launchClaude` still interpolates the prompt path into a shell command unquoted (`Sources/Crow/App/SessionService.swift:422` — `\"$(cat \(promptPath))\"`). For review sessions the worktree path is fully Crow-generated; for jobs the path now embeds the user-typed `job.repo` (`{devRoot}/{workspace}/{repo}-job-{slug}-{stamp}`). Exploitation requires a repo folder whose literal name contains shell metacharacters (`;`, `$(...)`, backticks), so this is local-user-shoots-self territory — but consider either validating `job.repo` against the same alphanumeric/dash set used by `slugify`, or using `shellQuote(promptPath)` here. Either is a small change with real defense-in-depth value now that jobs let a user enter the repo as free-form text when the workspace `alwaysInclude` list is empty (`JobFormView.swift:489`).

### Code Quality

- `firstPrompt.write(toFile: promptPath, atomically: true, encoding: .utf8)` is `try?` (`SessionService.swift:1294`). If the write silently fails (disk full, perms), `cat` returns empty and Claude launches with no prompt while `lastRunAt` still gets stamped — the run is effectively lost. Consider logging the error and/or returning `nil` from `runJob` so the scheduler doesn't record the run. Matches existing review-session behavior, but worth a note.
- `JobScheduler.deliverRemainingPrompts` escapes the awaiting `Task`: `inFlight.remove` runs via `defer` as soon as `runJob` returns, but the `DispatchQueue.main.asyncAfter` chain for follow-up prompts keeps going. This is intentional (the guard only covers worktree creation), but worth a one-line comment noting that.
- `runStamp()` is `yyyyMMdd-HHmmss` with second resolution. With `inFlight` and unique job names this should never collide in practice, but if two distinct jobs ever ended up with the same slug, simultaneous ticks could produce identical worktree paths. Slug uniqueness isn't explicitly enforced (only name case-insensitive uniqueness is). Adding milliseconds, or a short random suffix, would harden this for ~zero cost.
- `JobSchedule.dailyAt` decode (`JobConfig.swift:194`) accepts any `[Int]` for `weekdays`; out-of-range values (e.g. `[0]`, `[9]`) make `nextRunDate` return `nil` after the 7-attempt bound, silently disabling the schedule. The UI can't produce these, but a hand-edited config can. Clamping to `1...7` on decode (or in `nextRunDate`) would fail loudly instead of going silent.
- `SettingsView.scheduleSummary` for `dailyAt` with an unknown weekday int would just drop it from the summary (`compactMap`), which is fine but compounds the "silent disable" behavior above.
- `JobConfigTests` is thorough for the model. The scheduler itself isn't directly tested — a small unit test that drives `tick()` with a mock `jobsProvider`/`onJobRan` would be valuable given the role this plays, but the existing surface (interval next-run, weekday skip, round-trip, forward-compat) is the right priority.

### Summary Table

| Priority | Issue |
|----------|-------|
| Yellow | `repo` free-form text → unquoted `cat` shell substitution. Either validate `repo` or shell-quote the path. (`SessionService.swift:422`) |
| Yellow | `try?` swallows prompt-file write errors; `lastRunAt` still stamped → silent run loss. (`SessionService.swift:1294`) |
| Green | Clamp `weekdays` to `1...7` in `JobSchedule` decode. (`JobConfig.swift:194`) |
| Green | Add ms / random suffix to `runStamp()` for collision safety. (`SessionService.swift:1249`) |
| Green | Consider a small unit test for `JobScheduler.tick()` (mocked providers). |
| Green | One-liner comment in `JobScheduler.tick()` clarifying that `inFlight` only protects worktree creation, not the multi-prompt tail. |

**Recommendation:** Approve. The implementation is solid, the tests cover what they should, and nothing here is blocking. The yellow items are worth a follow-up but don't need to hold the merge.

---

PR #324, opened 2026-05-23
Run Manager session on the tmux backend

### LINKED ISSUE #314: Manager session should run on the tmux backend like work sessions
## Summary
The Manager session should use the same terminal backend (tmux) as work sessions, so it behaves
consistently with everything else — same persistence, reattachment, and lifecycle.

## Background
The Manager session is created in `ensureManagerSession(devRoot:)`
(`Sources/Crow/App/SessionService.swift`), keyed off `AppState.managerSessionID`
(`00000000-0000-0000-0000-000000000000`). Work sessions can run on the tmux backend
(`experimentalTmuxBackend`), but the Manager terminal is spun up through its own path and does not
share that backend. This split means the Manager doesn't get the persistence/reattach benefits the
tmux backend provides, and its behavior diverges from every other session.

## Ask
- Route the Manager terminal through the same tmux-backed terminal creation path as work sessions.
- Reuse the existing backend selection rather than special-casing the Manager.
- Verify reattachment after app restart works for the Manager the same as for work sessions.

## Acceptance
- Manager terminal is created on the tmux backend when `experimentalTmuxBackend` is enabled.
- No separate/legacy terminal path remains specific to the Manager.
- Closing and reopening the app reattaches the Manager session like any work session.

dgershman:
Closes #314

## Summary
Routes the Manager session's terminal through the same backend-selection path as work sessions, removing the Manager-specific force-pin to the Ghostty backend. The Manager now honors the tmux backend flag like every other session — so it gets the same persistence/reattach behavior and renders through the shared tmux cockpit surface.

## Changes
- **`prepareTerminal`** (`SessionService.swift`) no longer excludes the Manager from tmux. For the Manager, `TmuxBackend.registerTerminal` pastes the stored `claude` command into the new tmux window directly (`trackReadiness: false`, matching the Manager's command-launches-claude model — no readiness/`launchClaude` flow).
- **`ensureManagerSession`** creates the Manager terminal via `prepareTerminal` and persists the returned backend-tagged row, dropping the direct `TerminalManager.preInitialize` call.
- **`new-terminal` RPC handler** (`AppDelegate.swift`) drops its Manager-specific `useTmux`/`trackReadiness` exclusions, so no separate terminal path remains specific to the Manager.
- **`onWorkOnIssue` / `onBatchWorkOnIssues`** send `/crow-workspace` commands via `TerminalRouter` (backend-aware) so they reach the Manager whether it's on Ghostty or tmux.

## Behavior notes
- Backend choice stays **sticky per persisted terminal row**, consistent with how work-session backends already behave: existing `.ghostty` Manager rows are not migrated; only a newly-created Manager terminal (fresh install / cleared row) gets tmux.
- Reattach on app restart reuses the existing backend-aware rehydrate path — a persisted `.tmux` Manager row re-registers a fresh window and re-pastes its `claude` command, exactly like a work session.
- The `hydrateState` Manager command-rebuild (keeping `--rc` / `--permission-mode auto` in sync) is retained — it's command-string maintenance, not a separate backend path.

## Acceptance criteria
- [x] Manager terminal is created on the tmux backend when the flag is enabled
- [x] No separate/legacy terminal path remains specific to the Manager
- [x] Reattach on restart works for the Manager the same as for work sessions (shared rehydrate path)

## Testing
- `swift build` — clean (full app links).
- `swift test` (root) — 113 tests / 12 suites pass, including `IssueTracker completion decisions` (`managerSessionIsProtected`) and `FeatureFlags tmux backend default`.
- `swift test` (CrowTerminal) — 26 tests / 5 suites pass, including `TmuxBackend integration`.

### DIFF
--- a/Sources/Crow/App/AppDelegate.swift
+++ b/Sources/Crow/App/AppDelegate.swift
@@ -406,11 +406,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         appState.onWorkOnIssue = { [weak self] issueURL in
             guard let self, let managerTerminals = self.appState.terminals[AppState.managerSessionID],
                   let managerTerminal = managerTerminals.first else { return }
-            // Type the /crow-workspace command into the Manager terminal
-            TerminalManager.shared.send(
-                id: managerTerminal.id,
-                text: "/crow-workspace \(issueURL)\n"
-            )
+            // Type the /crow-workspace command into the Manager terminal.
+            // Route through TerminalRouter so it reaches the Manager regardless
+            // of backend (Ghostty or tmux, #314).
+            TerminalRouter.send(managerTerminal, text: "/crow-workspace \(issueURL)\n")
             // Switch to Manager tab
             self.appState.selectedSessionID = AppState.managerSessionID
         }
@@ -420,10 +419,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             guard let self, let managerTerminals = self.appState.terminals[AppState.managerSessionID],
                   let managerTerminal = managerTerminals.first else { return }
             let urls = issueURLs.joined(separator: " ")
-            TerminalManager.shared.send(
-                id: managerTerminal.id,
-                text: "/crow-batch-workspace \(urls)\n"
-            )
+            // Route through TerminalRouter so it reaches the Manager regardless
+            // of backend (Ghostty or tmux, #314).
+            TerminalRouter.send(managerTerminal, text: "/crow-batch-workspace \(urls)\n")
             self.appState.selectedSessionID = AppState.managerSessionID
         }
 
@@ -1020,13 +1018,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                             && !cmd.contains("--rc")
                             && !cmd.contains("--remote-control")
                     }
-                    let trackReadiness = isManaged && sessionID != AppState.managerSessionID
-                    // Decide backend at terminal-create time. Manager terminal
-                    // stays on Ghostty for now (special case; migrating it is
-                    // its own follow-up). Everything else honors the flag.
+                    let trackReadiness = isManaged
+                    // Decide backend at terminal-create time. Every session,
+                    // including the Manager (#314), honors the flag.
                     let useTmux = FeatureFlags.tmuxBackend
                         && !TmuxBackend.shared.tmuxBinary.isEmpty
-                        && sessionID != AppState.managerSessionID
                     var terminal = SessionTerminal(
                         sessionID: sessionID,
                         name: terminalName,
--- a/Sources/Crow/App/SessionService.swift
+++ b/Sources/Crow/App/SessionService.swift
@@ -498,12 +498,19 @@ final class SessionService {
                 sessionName: "Manager",
                 autoPermissionMode: autoMode
             )
-            let terminal = SessionTerminal(
+            let rawTerminal = SessionTerminal(
                 sessionID: managerID,
                 name: "Manager",
                 cwd: devRoot,
                 command: managerCommand
             )
+
+            // Route through the same backend-selection path as work sessions
+            // (#314). On tmux this registers a window and pastes the `claude`
+            // command into it; on Ghostty it pre-initializes the offscreen
+            // surface. `trackReadiness: false` matches the Manager's
+            // command-launches-claude model — no readiness/launchClaude flow.
+            let terminal = prepareTerminal(rawTerminal, trackReadiness: false)
             appState.terminals[managerID] = [terminal]
 
             store.mutate { data in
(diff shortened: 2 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

Routes the Manager terminal through the same `prepareTerminal` → backend-dispatch flow as work sessions. Clean removal of the Manager-specific force-pin to Ghostty in three places (`prepareTerminal`, `ensureManagerSession`, and the `new-terminal` RPC), plus a `TerminalRouter.send` switch in the two Manager send-text closures.

### Critical Issues
None.

### Security Review
**Strengths:**
- No new trust boundary crossings. The `issueURL` / `issueURLs` strings flow through `TerminalRouter.send`, which for tmux uses the buffer-paste path (load-buffer + paste-buffer with bracketed paste). That avoids shell-quoting hazards on tmux just as Ghostty's `writeText` did.
- `prepareTerminal` reuses the same `FeatureFlags.tmuxBackend && \!tmuxBinary.isEmpty` guard the work-session path uses; no privilege or capability change for the Manager beyond what the tmux backend already does.
- No secrets or paths are logged differently than before; persistence (`store.mutate { … }`) and `appState.terminals` write the row returned by `prepareTerminal`, so the new `.tmux` backend tag is recorded correctly.

**Concerns:**
- None blocking.

### Code Quality

**Strengths**

- `Sources/Crow/App/SessionService.swift:501-525` — the `rawTerminal` → `prepareTerminal(_, trackReadiness: false)` → persist sequence captures the `backend` / `tmuxBinding` correctly before `store.mutate`. The `trackReadiness: false` matches the Manager's command-launches-claude semantics: the persisted `claude …` command is the *initial command* pasted by `TmuxBackend.registerTerminal` (line 192-194), not a follow-up after `.shellReady`, so we deliberately bypass the sentinel watch + `launchClaude` flow. Defense-in-depth: even if `onReadinessChanged` fired, `launchClaude` is gated on `autoLaunchTerminals`, which the Manager is never added to (`SessionService.swift:101-106`).
- `Sources/Crow/App/SessionService.swift:1378-1402` — single dispatch site for backend choice. Tmux failure falls back to Ghostty's `preInitialize` with the same `command`, preserving prior behavior on first launch when tmux is unavailable.
- `Sources/Crow/App/AppDelegate.swift:1018-1033` — the `new-terminal` RPC now just builds a SessionTerminal with the resolved backend and lets the same tmux register / Ghostty preInitialize branches handle it. The structural symmetry with `prepareTerminal` is reassuring (you'd want to dedupe these eventually, but that's outside this PR).
- `Sources/Crow/App/AppDelegate.swift:406-426` — switching to `TerminalRouter.send` for the two Manager send closures is correct: the router dispatches on the `terminal.backend` discriminator. Existing `.ghostty` Manager rows keep working, and a fresh `.tmux` Manager row routes through `TmuxBackend.sendText`.
- Rehydrate path consistency: `SessionService.swift:138` already uses `trackReadiness: \!isManagerSession && original.isManaged`, so a persisted `.tmux` Manager row re-registers a tmux window with `trackReadiness: false` and re-pastes the (rebuilt at lines 80-92) `claude` command. No change needed there — the existing path already handles the new state correctly.
- Backend stickiness on existing installs: the comment at PR body matches reality — `hydrateState` doesn't migrate `.ghostty` Manager rows to `.tmux`, and `ensureManagerSession`'s create branch only runs when `terminals(for: managerID).isEmpty`. So no surprise backend flip for users mid-flight.

**Minor (non-blocking)**

- No test added that asserts the new behavior (e.g., "with `FeatureFlags.tmuxBackend == true`, `ensureManagerSession` produces a terminal with `backend == .tmux`"). The PR relies on existing tests staying green and on the manual `swift test` claim in the PR description. Adding a targeted unit test would have caught any future regression that re-pins the Manager to Ghostty.
- With `trackReadiness: false` on the tmux path, a stuck Manager wrapper (mktemp failure, broken `.zshrc`, etc.) won't surface the `.timedOut` Retry overlay that work sessions get. This matches the prior Ghostty Manager behavior (no readiness watch either), so it's not a regression — but it's worth being aware that the Manager will now also fail silently on tmux startup hiccups. If the Manager ever needs Retry, that's a follow-up.
- `prepareTerminal` and the `new-terminal` RPC handler still duplicate the `FeatureFlags.tmuxBackend && \!TmuxBackend.shared.tmuxBinary.isEmpty` + register-then-fallback logic. Not introduced by this PR, but with the Manager now on the same path, the case for consolidating both into a single helper has grown stronger.

### Summary Table

| Priority | Issue |
|----------|-------|
| 🟢 Green | Consider adding a unit test pinning Manager → tmux backend selection with the flag on |
| 🟢 Green | Future cleanup: dedupe `prepareTerminal` vs. the `new-terminal` RPC's tmux register-and-fallback block |
| 🟢 Green | Be aware: Manager loses no readiness/Retry affordance because it never had one — `trackReadiness: false` matches the existing model |

**Recommendation:** Approve. Clean removal of Manager-specific special-casing, behavior matches the PR description, no security or correctness regressions identified. The new code path is consistent with how work sessions already pick a backend and the rehydrate path already handled the Manager-on-tmux case via `\!isManagerSession && original.isManaged`, so the lifecycle invariants line up end-to-end.

---

PR #325, opened 2026-05-23
Restart Manager session after its process exits

### LINKED ISSUE #315: Allow restarting the Manager session if its process is killed
## Summary
If the Manager session's `claude` process dies (crash, manual kill, OOM), there is currently no way
to bring it back short of relaunching the app. We should be able to restart it in place.

## Background
The Manager terminal is launched once in `ensureManagerSession(devRoot:)`
(`Sources/Crow/App/SessionService.swift`) and is treated as a singleton. There is no detection of a
dead Manager process and no respawn logic — the terminal tab remains but input silently fails.

## Ask
- Detect when the Manager's process has exited.
- Provide a "Restart Manager" action (and/or automatic restart) that relaunches Claude Code in the
  devRoot using the existing launch args (`ClaudeLaunchArgs`, remote-control + auto-permission mode).
- Preserve the Manager's identity (`AppState.managerSessionID`) across the restart.

## Acceptance
- A killed Manager process is detected and surfaced in the UI.
- The user can restart the Manager without restarting the whole app; the session keeps its identity.

dgershman:
Closes #315

## Problem
The Manager session runs a single `claude` process launched once in `SessionService.ensureManagerSession`. If that process died (crash, manual kill, OOM) there was **no detection and no recovery** — the terminal tab remained but input silently failed, and the only fix was relaunching the whole app.

## Approach
- **Detect** the exit via Ghostty's `GHOSTTY_ACTION_SHOW_CHILD_EXITED` action. The dead surface is mapped back to its terminal UUID (via `ghostty_surface_userdata`), and if it's the Manager terminal, `AppState.managerProcessExited` is set.
- **Surface** it in the UI: a "Manager process exited" banner over the Manager terminal with a **Restart Manager** button, plus a "Restart Manager" item in the app menu (reachable from any tab).
- **Restart** in place: `SessionService.restartManager(devRoot:)` tears down the dead Ghostty surface, drops the stale terminal row (memory + disk), and re-runs `ensureManagerSession` to recreate a fresh Manager terminal with the current `--rc` / auto-permission args. The Manager session row and `AppState.managerSessionID` are preserved.

Restart is **manual-only** (no auto-respawn) to avoid crash-loops when claude fails repeatedly on launch.

## Changes
- `GhosttyApp`: handle `SHOW_CHILD_EXITED`, fire `onChildExited(terminalID, exitCode)` on the main actor (mirrors the existing `wakeup_cb` pattern)
- `GhosttySurfaceView` / `TerminalManager`: tag each surface with its `terminalID`
- `AppState`: `managerProcessExited` flag + `onRestartManager` callback
- `SessionService`: `restartManager(devRoot:)`
- `AppDelegate`: wire detection + restart + "Restart Manager" menu item
- `SessionDetailView`: dead-Manager banner with a Restart button

## Testing
- `CrowCore` builds cleanly (`swift build --package-path Packages/CrowCore`).
- The Ghostty/UI targets could not be linked on this dev host (x86_64) because GhosttyKit ships an arm64-only slice — relying on CI (arm64) to fully type-check/link. The Ghostty change deliberately mirrors the proven `wakeup_cb` callback pattern.
- Manual verification plan: kill the Manager's claude process → banner appears → click **Restart Manager** (and the menu item) → fresh claude launches in the devRoot with the same args, banner clears, session identity unchanged, `crow send` works again; other sessions' terminals untouched.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -43,6 +43,11 @@ public final class AppState {
     /// launch; worker sessions and CLI-spawned terminals are unaffected.
     public var managerAutoPermissionMode: Bool = true
 
+    /// `true` when the Manager's `claude` process has exited (crash, kill, OOM)
+    /// and has not yet been restarted. Drives the "Manager process exited" banner
+    /// and enables the "Restart Manager" action. Reset when the Manager relaunches.
+    public var managerProcessExited: Bool = false
+
     /// Terminal IDs whose Claude Code was launched with `--rc` — drives the
     /// per-session indicator badge. Survives toggle changes so existing sessions
     /// keep showing the badge until they're restarted.
@@ -222,6 +227,10 @@ public final class AppState {
 
     /// Called when the user clicks "Retry" on a failed terminal surface.
     public var onRetryTerminal: ((UUID) -> Void)?  // receives terminal ID
+
+    /// Called to relaunch the Manager's `claude` process after it exited, while
+    /// preserving the Manager session identity. Wired to `SessionService.restartManager`.
+    public var onRestartManager: (() -> Void)?
 
     /// Called when the user clicks "Retry" on a terminal whose tmux readiness
     /// watch timed out before the shell signaled it was interactive.
--- a/Packages/CrowTerminal/Sources/CrowTerminal/GhosttyApp.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/GhosttyApp.swift
@@ -9,6 +9,12 @@ public final class GhosttyApp {
     private(set) var app: ghostty_app_t?
     private(set) var config: ghostty_config_t?
 
+    /// Fired when a surface's child process exits (Ghostty's
+    /// `GHOSTTY_ACTION_SHOW_CHILD_EXITED`). Delivers the terminal UUID and the
+    /// child's exit code. Used to detect a dead Manager `claude` process so the
+    /// UI can offer a restart. Always invoked on the main actor.
+    public var onChildExited: ((UUID, Int32) -> Void)?
+
     private init() {}
 
     public func initialize() {
@@ -98,6 +104,21 @@ public final class GhosttyApp {
         target: ghostty_target_s,
         action: ghostty_action_s
     ) -> Bool {
+        // The child process backing a surface exited (crash, kill, normal quit).
+        // Map the surface back to its terminal UUID via the userdata pointer we
+        // set in GhosttySurfaceView.createSurface, then notify on the main actor.
+        if action.tag == GHOSTTY_ACTION_SHOW_CHILD_EXITED {
+            if target.tag == GHOSTTY_TARGET_SURFACE,
+               let userdata = ghostty_surface_userdata(target.target.surface) {
+                let exitCode = Int32(bitPattern: action.action.child_exited.exit_code)
+                DispatchQueue.main.async {
+                    let view = Unmanaged<GhosttySurfaceView>.fromOpaque(userdata).takeUnretainedValue()
+                    guard let terminalID = view.terminalID else { return }
+                    GhosttyApp.shared.onChildExited?(terminalID, exitCode)
+                }
+            }
+            return true
+        }
         if !silencedActionTags.contains(UInt32(action.tag.rawValue)) {
             NSLog("[GhosttyApp] Unhandled action: tag=\(action.tag)")
         }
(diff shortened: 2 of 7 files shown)


dhilgaertner (approved):
## Code & Security Review

### Summary
Tight, well-scoped fix for #315. The detection path mirrors the existing `wakeup_cb` callback pattern, the restart preserves session identity, and the manual-only policy (no auto-respawn) correctly avoids crash-loops. No security concerns: no new IPC surface, no path/command injection, no secret handling. Build of the `CrowCore` package passes; the Ghostty/UI targets aren't linkable on x86_64 dev hosts (GhosttyKit is arm64-only) — matches the author's note about relying on CI.

### Security Review
**Strengths:**
- No new external attack surface; restart is in-process, gated on a UI affordance the user must trigger.
- `restartManager` only acts on `AppState.managerSessionID` (a fixed UUID constant); no caller-provided ID can be used to teardown arbitrary sessions through this path.
- Manager `claude` command is rebuilt from `ClaudeLaunchArgs.argsSuffix(...)` + `findClaudeBinary()` (no user-controlled string interpolation).
- `passUnretained` of `GhosttySurfaceView` into Ghostty userdata is safe in practice because `TerminalManager.surfaces` retains the view until `destroy(id:)` calls `ghostty_surface_free` first — Ghostty stops emitting events after the surface is freed.

**Concerns:**
- None blocking.

### Code Quality

**Yellow — worth addressing:**
- `Sources/Crow/App/AppDelegate.swift:716-718` — the **Restart Manager** menu item has no `validateMenuItem` and is always enabled. Clicking it while the Manager is healthy will tear down the running `claude` and start a new one. That may be intentional (force-restart), but worth either (a) wiring `validateMenuItem` so it's only enabled when `appState.managerProcessExited == true`, or (b) renaming to "Force Restart Manager" to set expectations.
- `Sources/Crow/App/AppDelegate.swift:346-353` — the `onChildExited` closure captures `terminalID` but discards the exit code (`_`). Since `handleAction` already extracts it via `Int32(bitPattern:)`, propagating it into the log line (`NSLog("[Crow] Manager process exited terminal=%@ code=%d", ...)`) would make post-mortem triage of "did claude crash or did the user kill it?" much easier later, with zero added complexity.

**Green — consider:**
- `Sources/Crow/App/SessionService.swift:537-557` — `restartManager` is unit-testable without Ghostty (just mock/stub `TerminalRouter.destroy`). Given the PR notes the Ghostty/UI targets can't be exercised locally, a focused CrowCore-level test that asserts (a) terminal row removed from memory + store, (b) `managerProcessExited` cleared, (c) session row preserved, (d) a fresh terminal recreated, would be a cheap safety net for the next refactor.
- `Packages/CrowUI/Sources/CrowUI/SessionDetailView.swift:318-351` — the banner sits in a `ZStack` over the dead `TerminalSurfaceView`. Background opacity is 0.95 so it visually covers, but the underlying view still receives hit-test/keyboard events. Adding `.allowsHitTesting(false)` to the terminal layer when `managerProcessExited == true` would prevent the user from clicking into a dead terminal that silently swallows input.
- `Sources/Crow/App/AppDelegate.swift:381-383` — `onRestartManager` captures the launch-time `devRoot` by value. If the user changes devRoot via Settings, restart will still use the old path. Consistent with surrounding closures in this file, so probably out of scope, but flagging in case future cleanup wants to centralize this.
- `Packages/CrowTerminal/Sources/CrowTerminal/GhosttyApp.swift:110-121` — `DispatchQueue.main.async { ... GhosttyApp.shared.onChildExited?(...) }` accesses `@MainActor`-isolated state from inside a `@Sendable` closure. The existing `wakeup_cb` does the same thing, so this isn't novel — and it'll keep working in Swift 5 mode — but if/when this package flips to strict concurrency, both call sites will need `Task { @MainActor in ... }` instead.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | (none) |
| Yellow | Menu item always enabled — can nuke a healthy Manager unintentionally |
| Yellow | Exit code is captured but dropped — log it for diagnostics |
| Green | Add a `SessionService.restartManager` unit test in CrowCore (compensates for CI-only Ghostty path) |
| Green | `.allowsHitTesting(false)` on the dead terminal layer behind the banner |
| Green | Captured `devRoot` in `onRestartManager` closure becomes stale if user changes devRoot in Settings |
| Green | Heads-up: `DispatchQueue.main.async` → `@MainActor` access pattern will need `Task { @MainActor in ... }` under strict concurrency |

**Recommendation:** **Approve.** The fix is correct, scoped, and follows established patterns. The yellow items are polish, not correctness bugs — feel free to address them in this PR or a follow-up.

---

PR #326, opened 2026-05-23
Support multiple Manager sessions

### LINKED ISSUE #316: Support spawning multiple Manager sessions
## Summary
Allow more than one Manager session to run at a time (e.g. one per workspace, or several parallel
orchestrators).

## Background
The Manager is currently a hard singleton: a single hardcoded UUID
(`AppState.managerSessionID = 00000000-0000-0000-0000-000000000000`), inserted once by
`ensureManagerSession(...)`, auto-selected on launch, and protected from deletion in the RPC handler
(`Sources/Crow/App/AppDelegate.swift`). This prevents running multiple orchestration contexts.

## Ask
- Move from the hardcoded-UUID singleton to a Manager *kind* of session (e.g. a `SessionKind`
  value / flag in `CrowCore/Models/Session.swift`) so multiple can coexist.
- Let the user create additional Manager sessions, each with its own terminal running Claude Code in
  the devRoot (or a chosen workspace root).
- Keep one "primary" Manager for back-compat with the well-known UUID where other tooling expects it.

## Acceptance
- More than one Manager session can exist and run concurrently.
- Each Manager has an independent terminal/process; deleting a non-primary Manager is allowed.
- Existing single-Manager behavior still works as the default.

dgershman:
Closes #316

## Summary
Moves the Manager from a hardcoded-UUID singleton to a `SessionKind.manager` *kind*, so multiple Manager sessions can run concurrently — each with its own Claude Code terminal running in the devRoot.

The well-known UUID (`00000000-0000-0000-0000-000000000000`) is retained as the **primary** Manager for back-compat: it's the only one protected from deletion, and `/crow-workspace` ("Work on this issue") routing stays pinned to it.

## Changes
- **Model:** add `SessionKind.manager` + `Session.isManager`. The legacy primary Manager (persisted as `.work`) is migrated to `.manager` on launch in `ensureManagerSession`.
- **Generalized special-casing:** every `id == managerSessionID` check became `isManager` across `SessionService` (hydration, readiness, status, Ghostty pin), `AppDelegate` (new-terminal), and `IssueTracker` (PR/issue tracking + auto-cleanup).
- **Creation:** new `SessionService.createManagerSession(name:cwd:)` + a reusable terminal helper. CLI gains `crow new-session --kind manager`; the `new-session` RPC accepts an optional `kind`.
- **UI:** sidebar "+" button spawns a Manager (auto-named "Manager N"); additional managers render as selectable rows with a Delete context menu; `SessionDetailView` special-cases switch to `isManager`.
- **Cleanup:** all managers excluded from auto-cleanup and PR/issue reconciliation; the remote-control set is cleaned up on session deletion (pre-existing latent leak).

## Acceptance
- ✅ More than one Manager can exist and run concurrently.
- ✅ Each Manager has an independent terminal/process; deleting a non-primary Manager is allowed (primary stays protected).
- ✅ Existing single-Manager behavior is the default.

## Testing
- `swift build --arch arm64` — full app builds.
- Tests pass: CrowCore (175), CrowCLI (38), CrowUI (31), app target (114). Added coverage for manager Codable/`isManager`, AppState lookups, CLI `--kind` parsing, and a non-primary-manager cleanup-exclusion case.

### DIFF
--- a/Packages/CrowCLI/Sources/CrowCLILib/Commands/SessionCommands.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Commands/SessionCommands.swift
@@ -6,15 +6,26 @@ import Foundation
 
 /// Create a new development session.
 ///
-/// Returns the new session's UUID and name as JSON.
+/// Returns the new session's UUID and name as JSON. Pass `--kind manager` to
+/// create a Manager (orchestration) session with its own Claude Code terminal
+/// running in the devRoot.
 public struct NewSession: ParsableCommand {
     public static let configuration = CommandConfiguration(commandName: "new-session", abstract: "Create a new session")
     @Option(name: .long, help: "Session name") var name: String
+    @Option(name: .long, help: "Session kind: work (default), review, or manager") var kind: String?
 
     public init() {}
 
+    public func validate() throws {
+        if let kind, !["work", "review", "manager"].contains(kind) {
+            throw ValidationError("kind must be one of: work, review, manager")
+        }
+    }
+
     public func run() throws {
-        let result = try rpc("new-session", params: ["name": .string(name)])
+        var params: [String: JSONValue] = ["name": .string(name)]
+        if let kind { params["kind"] = .string(kind) }
+        let result = try rpc("new-session", params: params)
         printJSON(result)
     }
 }
--- a/Packages/CrowCLI/Sources/CrowCLILib/Commands/SessionCommands.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Commands/SessionCommands.swift
@@ -8,17 +8,19 @@ import Foundation
 ///
 /// Returns the new session's UUID and name as JSON. Pass `--kind manager` to
 /// create a Manager (orchestration) session with its own Claude Code terminal
-/// running in the devRoot.
+/// running in the devRoot. The supplied name is used verbatim (the caller is
+/// responsible for uniqueness — sessions are identified by UUID, not name).
+/// Review and job sessions are created through their own setup flows, not here.
 public struct NewSession: ParsableCommand {
     public static let configuration = CommandConfiguration(commandName: "new-session", abstract: "Create a new session")
     @Option(name: .long, help: "Session name") var name: String
-    @Option(name: .long, help: "Session kind: work (default), review, or manager") var kind: String?
+    @Option(name: .long, help: "Session kind: work (default) or manager") var kind: String?
 
     public init() {}
 
     public func validate() throws {
-        if let kind, !["work", "review", "manager"].contains(kind) {
-            throw ValidationError("kind must be one of: work, review, manager")
+        if let kind, !["work", "manager"].contains(kind) {
+            throw ValidationError("kind must be one of: work, manager")
         }
     }
 
--- a/Packages/CrowCLI/Tests/CrowCLITests/CommandParsingTests.swift
+++ b/Packages/CrowCLI/Tests/CrowCLITests/CommandParsingTests.swift
@@ -20,6 +20,26 @@ private let validUUID = "a1b2c3d4-e5f6-7890-abcd-ef1234567890"
     }
 }
 
+@Test func newSessionParsesManagerKind() throws {
+    let cmd = try NewSession.parse(["--name", "Manager 2", "--kind", "manager"])
+    #expect(cmd.name == "Manager 2")
+    #expect(cmd.kind == "manager")
+    try cmd.validate()
+}
+
+@Test func newSessionDefaultsToNoKind() throws {
+    let cmd = try NewSession.parse(["--name", "feature"])
+    #expect(cmd.kind == nil)
+    try cmd.validate()
+}
+
+@Test func newSessionRejectsInvalidKind() {
+    // ArgumentParser runs validate() during parse, so an invalid kind throws here.
+    #expect(throws: (any Error).self) {
+        _ = try NewSession.parse(["--name", "x", "--kind", "bogus"])
+    }
+}
+
 @Test func setStatusParsesArgs() throws {
     let cmd = try SetStatus.parse(["--session", validUUID, "active"])
     #expect(cmd.session == validUUID)
--- a/Packages/CrowCLI/Tests/CrowCLITests/CommandParsingTests.swift
+++ b/Packages/CrowCLI/Tests/CrowCLITests/CommandParsingTests.swift
@@ -40,6 +40,16 @@ private let validUUID = "a1b2c3d4-e5f6-7890-abcd-ef1234567890"
     }
 }
 
+@Test func newSessionRejectsReviewAndJobKinds() {
+    // Review and job sessions need dedicated setup flows, so new-session only
+    // accepts work and manager.
+    for kind in ["review", "job"] {
+        #expect(throws: (any Error).self) {
+            _ = try NewSession.parse(["--name", "x", "--kind", kind])
+        }
+    }
+}
+
 @Test func setStatusParsesArgs() throws {
     let cmd = try SetStatus.parse(["--session", validUUID, "active"])
     #expect(cmd.session == validUUID)
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -82,8 +82,22 @@ public final class AppState {
     /// Fixed UUID for the global terminals page.
     nonisolated public static let globalTerminalSessionID = UUID(uuidString: "00000000-0000-0000-0000-000000000004")!
 
+    /// The primary (back-compat) Manager session identified by the well-known UUID.
     public var managerSession: Session? {
         sessions.first { $0.id == Self.managerSessionID }
+    }
+
+    /// All Manager-kind sessions (primary + any additional ones).
+    public var managerSessions: [Session] {
+        sessions.filter { $0.isManager }
+    }
+
+    /// Whether the session with the given id is a Manager. Falls back to the
+    /// well-known primary UUID when the session isn't loaded yet (e.g. during
+    /// terminal creation before the session row is present).
+    public func isManagerSession(_ id: UUID) -> Bool {
+        if let session = sessions.first(where: { $0.id == id }) { return session.isManager }
+        return id == Self.managerSessionID
     }
 
     // MARK: - Issue Tracking
@@ -210,6 +224,9 @@ public final class AppState {
         _sessionState.removeValue(forKey: sessionID)
     }
 
+    /// Called when the user clicks the sidebar "+" to spawn a new Manager session.
+    public var onCreateManager: (() -> Void)?
+
     /// Called when user clicks "Work on" for an assigned issue.
     public var onWorkOnIssue: ((String) -> Void)?  // receives issue URL
 
@@ -317,11 +334,11 @@ public final class AppState {
     }
 
     public var inReviewSessions: [Session] {
-        sessions.filter { $0.status == .inReview && $0.id != Self.managerSessionID }
+        sessions.filter { $0.status == .inReview && !$0.isManager }
     }
 
     public var completedSessions: [Session] {
-        sessions.filter { $0.status == .completed || $0.status == .archived }
+        sessions.filter { ($0.status == .completed || $0.status == .archived) && !$0.isManager }
     }
 
     public var reviewSessions: [Session] {
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -330,7 +330,7 @@ public final class AppState {
     }
 
     public var activeSessions: [Session] {
-        sessions.filter { $0.status == .active && $0.id != Self.managerSessionID && $0.kind == .work }
+        sessions.filter { $0.status == .active && $0.kind == .work }
     }
 
     public var inReviewSessions: [Session] {
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -92,9 +92,11 @@ public final class AppState {
         sessions.filter { $0.isManager }
     }
 
-    /// Whether the session with the given id is a Manager. Falls back to the
-    /// well-known primary UUID when the session isn't loaded yet (e.g. during
-    /// terminal creation before the session row is present).
+    /// Whether the session with the given id is a Manager. When the session row
+    /// isn't loaded yet (e.g. during terminal creation before it's inserted) the
+    /// fallback only recognizes the well-known *primary* UUID — a not-yet-loaded
+    /// non-primary manager id returns `false`, so callers must not rely on this
+    /// for non-primary managers pre-load.
     public func isManagerSession(_ id: UUID) -> Bool {
         if let session = sessions.first(where: { $0.id == id }) { return session.isManager }
         return id == Self.managerSessionID
--- a/Packages/CrowCore/Sources/CrowCore/Models/Enums.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/Enums.swift
@@ -1,10 +1,12 @@
 import Foundation
 
-/// Whether a session is a normal work session or a PR review session.
+/// What kind of session this is. Multiple `.manager` sessions can coexist;
+/// the one with `AppState.managerSessionID` is the back-compat "primary".
 public enum SessionKind: String, Codable, Sendable {
     case work    // Normal development session (default)
     case review  // PR review session
     case job     // Session spun up by a scheduled job (CROW-317)
+    case manager // Orchestration session running Claude Code in the devRoot
 }
 
 /// Status of a development session.
--- a/Packages/CrowCore/Sources/CrowCore/Models/Session.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/Session.swift
@@ -28,6 +28,10 @@ public struct Session: Identifiable, Codable, Sendable {
     // run; the auto-merge watcher skips this session on subsequent polls.
     public var autoMergeEnabledAt: Date?
 
+    /// Whether this session is a Manager (orchestration) session. Managers run
+    /// Claude Code in the devRoot and are excluded from PR/issue tracking.
+    public var isManager: Bool { kind == .manager }
+
     public init(
         id: UUID = UUID(),
         name: String,
--- a/Packages/CrowCore/Tests/CrowCoreTests/AppStateLookupTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AppStateLookupTests.swift
@@ -105,6 +105,44 @@ import Testing
     #expect(appState.labels(forSession: session).isEmpty)
 }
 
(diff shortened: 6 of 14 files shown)


dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues

**Legacy primary-Manager migration is ordered after the hydration that mangles its terminal.** In `AppDelegate.swift:319` (`service.hydrateState()`) runs before `AppDelegate.swift:341` (`service.ensureManagerSession(...)`). For a user with an existing install, the primary Manager session is persisted with `kind: .work` (since the legacy `ensureManagerSession` created `Session(id:, name:, status:)` without a kind, defaulting to `.work` — and `Session.encode` writes the field). Hydration enters the `!session.isManager` branch at `SessionService.swift:38`, which:

  1. Marks the first claude terminal as `isManaged = true` via the heuristic backfill (`SessionService.swift:42-50`)
  2. **Clears the terminal's claude command to `nil`** (`SessionService.swift:55-65`)
  3. Inserts the terminal into `autoLaunchTerminals` (`SessionService.swift:101-105`)

Only *after* that does `ensureManagerSession` flip the session's `kind` to `.manager` (`SessionService.swift:479-485`). It does not undo the terminal-state mutations because `terminals(for: managerID).isEmpty` is false. The terminal will now boot as a plain shell, and the auto-launch path will eventually type `claude --continue` into it. `launchClaude` (`SessionService.swift:380-424`) does **not** include `--auto-permission-mode`, so legacy users silently lose their configured `managerAutoPermissionMode` until they delete the terminal manually. The terminal also no longer launches claude as its shell process, which is a separate behavior change.

Two reasonable fixes:
- Migrate the persisted `kind` for the well-known managerSessionID inside `hydrateState` (before the per-session loop), or
- Treat `session.id == AppState.managerSessionID` as `isManager` inside the hydration branch decisions (mirroring the fallback already in `AppState.isManagerSession`).

A regression test that hydrates a `data.sessions = [Session(id: managerSessionID, name: "Manager", kind: .work)]` plus its claude terminal would prevent reintroduction.

### Security Review

**Strengths:**
- New `--kind` RPC param is validated against the `SessionKind` enum (`AppDelegate.swift:850-854`) and rejected with `invalidParams` before reaching state mutation.
- CLI `validate()` rejects unknown kinds before the RPC call (`SessionCommands.swift:17-22`).
- Manager creation continues to run inside the existing `MainActor.run` serialization, so concurrent `crow new-session --kind manager` calls are safe.
- Primary Manager remains protected from deletion via both the CLI RPC handler (`AppDelegate.swift:950`) and the in-process `deleteSession` guard (`SessionService.swift:590`).
- `BulkDeleteSessionsAlert` selection only populates from `SessionRow` checkboxes and `SectionDivider` "select all" buttons; the `ManagerSessionRow` rows don't bind to `selectedSessionIDs`, so the primary Manager can't sneak in via bulk delete either.

**Concerns:**
- None blocking. The Manager terminal command is built from configuration values, not user input, so no injection surface is introduced by the new `createManagerTerminal` helper.

### Code Quality

- **Auto-naming can collide.** `AppDelegate.swift:409` computes `"Manager \(self.appState.managerSessions.count + 1)"`. Deleting `Manager 2` while `Manager 3` exists, then clicking "+", produces another `Manager 3`. Names aren't identifiers so this is cosmetic, but a `max(suffix) + 1` strategy (or a uniqueness pass) avoids the surprise.
- **Redundant filter clause.** `AppState.swift:324` keeps both `$0.id != Self.managerSessionID && $0.kind == .work` for `activeSessions`. After this PR, `kind == .work` already excludes all manager-kind sessions; the explicit ID check is dead weight and inconsistent with `inReviewSessions`/`completedSessions` which were updated to `!isManager`. Either drop it or switch the trio to a uniform predicate.
- **Stale doc comment.** `SessionService.swift:583` still reads "The manager session cannot be deleted." — only the primary one is protected now; non-primary managers are deletable by design.
- **Missing test for the migration path.** The `.work → .manager` upgrade branch in `ensureManagerSession` is not exercised by any test in this PR. Given the ordering bug above, this is the gap that lets the regression slip through.

### Summary Table

| Priority | Issue |
|----------|-------|
| Red | Legacy primary-Manager hydration runs before kind-migration, clearing its claude command and rerouting it through the work-session auto-launch path (drops `--auto-permission-mode`) |
| Yellow | `"Manager N"` auto-naming can collide after a delete-in-the-middle |
| Yellow | Redundant `id != managerSessionID && kind == .work` clause in `activeSessions` (inconsistent with the other filters now using `!isManager`) |
| Yellow | `deleteSession` doc says "manager session cannot be deleted" — only the primary one is protected now |
| Green | Add a hydration test for a legacy `kind: .work` primary Manager so the migration path is locked down |

**Recommendation:** Request Changes — the legacy migration regression will hit every existing user on upgrade and silently change their Manager terminal's claude args. Resolving the hydration/migration ordering (plus a test covering it) is enough to clear the block; the yellow items are nice-to-haves.

---

dgershman:
Thanks for the thorough review — addressed all items (rebased onto latest `main`, which now routes the Manager through the tmux backend per #314, so the multi-manager code follows that path too rather than re-pinning to Ghostty).

**🔴 Critical — legacy migration ordering:** Fixed. Added a pure, testable `SessionService.migrateLegacyManagerKind(_:)` and call it at the **top of `hydrateState`, before the per-session loop**, so a legacy `.work` primary Manager becomes `.manager` before the work-session branch can clear its claude command / reroute it through auto-launch. The change is persisted so the upgrade is one-shot. `ensureManagerSession` now reuses the same helper as defense-in-depth. Covered by new `ManagerMigrationTests` (legacy→manager, no-op when already manager, no-op when absent).

**🟡 Auto-naming collision:** Now picks the lowest unused `Manager N` suffix, so deleting `Manager 2` while `Manager 3` exists won't produce a duplicate.

**🟡 Redundant filter clause:** Dropped `id != managerSessionID` from `activeSessions`; `kind == .work` already excludes managers, matching `inReviewSessions`/`completedSessions`.

**🟡 Stale doc comment:** `deleteSession` doc now says only the *primary* Manager is protected; additional managers are deletable.

Also added the `.manager` case to the `SessionKind` switch in `initialPromptFileName` (surfaced by the #317 rebase; managers resume with `--continue`, no initial prompt).

All tests pass: CrowCore (185), CrowCLI (38), CrowUI (31), app target (121).

dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues
None.

### Security Review

**Strengths:**
- Session name validated at the RPC boundary via `AppDelegate.isValidSessionName` (length + control-character check) before reaching `createManagerSession`.
- `ClaudeLaunchArgs.argsSuffix` POSIX-single-quote-escapes `sessionName` via `shellQuote`, so even a malicious name can't break out of `--name '…'`.
- `cwd` for managers is pinned to `devRoot` on both the UI path (`AppDelegate.onCreateManager`) and the RPC path; user input cannot redirect Manager's working directory.
- Managers are excluded from PR/issue reconciliation, auto-merge, and auto-cleanup, so a malicious or accidental `--kind manager` cannot poison the GitHub-state machinery.
- Pre-existing latent leak fixed: `remoteControlActiveTerminals` entries are now cleaned up on session deletion (`SessionService.swift:638-640`).

**Concerns:**
- None — no new auth, secrets, or injection surface introduced.

### Code Quality

**Strengths:**
- The migration from \"hardcoded UUID singleton\" to a `SessionKind.manager` *kind* is mechanical, consistent, and well-justified. `isManager` is used in every place that previously did `id == managerSessionID` for kind-based filtering (cleanup, PR tracking, hydration, status transitions), while the well-known UUID is intentionally retained for the back-compat \"primary\" (deletion guard, `onWorkOnIssue` routing).
- `migrateLegacyManagerKind` is pure / `nonisolated` and reused on both the in-memory and persisted copies, with defense-in-depth in both `hydrateState` and `ensureManagerSession`. Unit tests in `ManagerMigrationTests` lock down all three branches (migrate, no-op-when-manager, no-op-when-missing).
- Good test coverage on the new behavior: `AppStateLookupTests` covers `managerSessions` / `isManagerSession` / activeSessions exclusion, `SessionModelTests` covers Codable round-trip + `isManager`, `IssueTrackerCompletionTests` adds the crucial \"non-primary manager excluded from cleanup\" case, and CLI parsing has happy/sad/default coverage.
- `onCreateManager` picks the lowest unused `Manager N` so delete-then-create doesn't collide — nice touch (`AppDelegate.swift:428-431`).
- Hydration's manager branch now rebuilds the claude command from `session.name` rather than the hardcoded `\"Manager\"` literal, so additional managers get correct `--name` labels in the RC panel after restart.

**Concerns:**

1. **`--kind review` (and `--kind job`) at the RPC layer produces a malformed session — `Yellow`.**
   `AppDelegate.swift:907` validates `kind` against `SessionKind(rawValue:) != nil`, so it accepts any case of the enum, including `.job` — but the error string lists only \"work, review, or manager\". The else-branch then does `let session = Session(name: name, kind: kind); capturedAppState.sessions.append(session)` — a *bare* session with no worktree, no terminal, no hooks, and (for review) no `.crow-review-prompt.md` setup that the rest of the app expects. Review and job sessions are normally created through dedicated setup paths (`createReviewSession`, the scheduler), so exposing them via `new-session` creates broken state.
   Suggested fix: restrict the RPC explicitly to `{.work, .manager}`, or route review/job through their proper setup helpers. At minimum, align the validation set with the error message:
   ```swift
   let allowed: Set<String> = [\"work\", \"manager\"]
   guard kindStr == nil || allowed.contains(kindStr!) else { … }
   ```
   The CLI's own `validate()` in `SessionCommands.swift` has the same gap (accepts `\"review\"`).

2. **`isManagerSession(_:)` fallback is broader than its doc-comment — `Green`.**
   `AppState.swift:96-99` falls back to `id == managerSessionID` when the session isn't loaded. The doc says this covers \"terminal creation before the session row is present.\" That's true for the primary, but a non-primary manager id passed in pre-load would return `false`, silently routing through the work-session code path. There's only one current caller (`updateSessionStatus`), so the risk is low — worth a one-line note that the fallback is primary-only, so future callers don't rely on it for non-primary managers.

3. **No regression test for the rebuild-on-hydrate `sessionName: session.name` change — `Green`.**
   The diff at `SessionService.swift:103` is exactly the kind of change that's silent if it regresses (the RC `--name` label is cosmetic). `ManagerMigrationTests` covers the kind upgrade but not the per-session-name rebuild. A small test that drives `hydrateState` with two `.manager` sessions and asserts each terminal's stored `command` contains the right `--name '…'` would lock this in.

4. **`createManagerSession` doesn't dedupe names from the CLI path — `Green`.**
   `onCreateManager` (UI path) computes the lowest-free `Manager N`, but the RPC path takes the user-supplied name verbatim. Two `crow new-session --kind manager --name \"Manager\"` calls will produce two sessions with identical display names, and the UI uses name as its sidebar label. Not a correctness bug (UUIDs are unique), but the UX is confusing. Consider either (a) appending a numeric suffix on collision, or (b) documenting that the caller is responsible for uniqueness.

5. **Style nit — `Green`.**
   `createManagerTerminal` is called in two places (`ensureManagerSession` and `createManagerSession`) and is correct, but the doc-comment on `createManagerSession` re-states what `createManagerTerminal` already says. Minor; ignore if you prefer.

### Testing

- PR description states `swift build --arch arm64` plus CrowCore (175), CrowCLI (38), CrowUI (31), and app (114) tests all pass. I could not reproduce locally — the review clone is missing `Frameworks/GhosttyKit.xcframework`, which is expected for a CI-only artifact.
- The added tests cover the high-risk surfaces: Codable round-trip, kind-based filtering in `AppState`, the legacy-manager migration's three branches, non-primary-manager cleanup exclusion, and CLI `--kind` parsing/validation. Coverage is appropriate.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | None |
| Yellow | `--kind review` / `--kind job` exposed via RPC and CLI produces malformed sessions (no worktree, no setup). Restrict to `{work, manager}` or route through proper setup. |
| Green | Document `isManagerSession` fallback as primary-only. |
| Green | Add a hydration test asserting per-manager `--name` rebuild uses `session.name`. |
| Green | Decide whether CLI-supplied manager names should auto-dedupe or stay verbatim; either way, document. |
| Green | Doc-comment overlap between `createManagerSession` and `createManagerTerminal`. |

**Recommendation:** Request Changes — only the `Yellow` is a real bug (the validation surface is wider than what's actually wired up). The rest are minor. Tightening the RPC/CLI to `{work, manager}` is a one-line fix and would close the gap cleanly.

---

dgershman:
Thanks — addressed the second round.

**🟡 `--kind review`/`--kind job` exposed via RPC/CLI (the real bug):** Fixed. `new-session` now accepts only `{work, manager}` at both the RPC handler (`AppDelegate`) and the CLI `validate()`, with the error message aligned to the allowed set. Review/job sessions stay on their dedicated setup paths. Added a CLI test asserting `review` and `job` are rejected.

**🟢 Per-manager `--name` rebuild test + single source:** Consolidated the hydrate manager-command rebuild to reuse the shared `managerCommand(sessionName:)` helper (so the `session.name → --name` flow has one code path), made it `internal`, and added `managerCommandUsesSessionNameForRemoteControlLabel` asserting distinct names yield distinct `--name '…'` flags (plus `--rc`/`--permission-mode auto`). A full `hydrateState` test isn't feasible headlessly — it drives `TerminalManager`/Ghostty, and the review clone lacks `GhosttyKit.xcframework` (as you noted) — so I locked it in at the shared helper instead.

**🟢 `isManagerSession` fallback:** Doc comment now states the pre-load fallback is primary-only; a not-yet-loaded non-primary manager id returns `false`, so future callers shouldn't rely on it for non-primary managers.

**🟢 CLI name dedupe:** Chose verbatim names (UUIDs are the identity); documented in the command help/doc that the caller owns uniqueness. The UI `+` path still auto-picks the lowest free `Manager N`.

**🟢 Doc overlap:** Trimmed `createManagerSession`'s comment so it no longer restates `createManagerTerminal`.

Rebased onto the latest branch tip (Crow's auto-rebase merged `main` in). Build clean; tests pass — app target (134), CrowCLI (39), CrowCore (185).

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review

**Strengths:**
- `new-session` RPC validates both `name` (control chars, length via `AppDelegate.isValidSessionName`) and `kind` (allowlist of `work`/`manager`), so a CLI client can't sneak `.review`/`.job` through the bare path that would skip their dedicated setup (`AppDelegate.swift:917-927`).
- Manager `cwd` is the resolved `devRoot`, not caller-controlled — no path traversal surface introduced by `createManagerSession(name:cwd:)`.
- `managerCommand()` builds the claude shell command via `ClaudeLaunchArgs.argsSuffix`, which is the same path already in use; no new shell-injection vector.
- `store.mutate` writes are NSLock-serialized, and all `appState` mutations go through `MainActor.run` — the migration is safe to run concurrently with hydration.

**Concerns:**
- None material to security.

### Code Quality

**Strengths:**
- Nice central abstraction: `SessionKind.manager` + `Session.isManager`, then every `id == managerSessionID` site that needs to recognize *all* managers became `isManager` — the change reads as a clean generalization rather than a sprawl of special cases.
- The legacy `.work`-persisted primary Manager migration (`migrateLegacyManagerKind`) is well-engineered: `nonisolated static`, pure, idempotent, runs against both the in-memory and persisted copies, and the ordering hazard (must run *before* the per-session hydration loop or `--auto-permission-mode` gets dropped) is captured in a load-bearing comment at `SessionService.swift:39-46` plus locked in by `ManagerMigrationTests`.
- `managerCommand(sessionName:)` consolidates command construction so the per-session `--name 'Manager N'` label has one source — `managerCommandUsesSessionNameForRemoteControlLabel` locks the distinct-label behavior in.
- Pre-existing latent leak fixed in `deleteSession`: `remoteControlActiveTerminals` is now cleaned up alongside `autoLaunchTerminals` (`SessionService.swift:687-694`).
- Auto-naming for new managers picks the lowest unused `Manager N` (`AppDelegate.swift:424-433`), so a delete-then-add cycle doesn't produce duplicate names.
- Test coverage is thorough across all the right layers: model Codable + `isManager`, AppState `managerSessions`/`isManagerSession` (including the documented fallback), `activeSessions`/`completedSessions` exclusion, CLI `--kind` parsing (positive + invalid + rejected `review`/`job`), migration, manager-command name flow, and an `IssueTracker` cleanup-exclusion case for non-primary managers.

**Concerns (yellow):**
- `IssueTracker.swift:2067` — `applyAutoRebase` still iterates with `where session.id != AppState.managerSessionID` instead of `!session.isManager`. Every other manager-exclusion site in this file was updated to `!session.isManager` in this PR (`collectStalePRURLs`, `buildReconcileCandidates`, `sessionsWithPRs`, the auto-merge loop, the reconcile filter, and `sessionsEligibleForCleanup`). The auto-rebase loop short-circuits on the `prLink` guard for managers (which never get `.pr` links), so this is **not a functional bug** — but it's a defense-in-depth + consistency miss, likely from the merge with main that brought in #332. Worth tightening for the same reason every other site was tightened.

**Nits (green, optional):**
- CLI `crow new-session --kind manager --name "Manager 2"` accepts the exact same name a future UI "+" click would have picked, so a CLI-then-UI flow can produce a `Manager 3` while `Manager 2` is also a CLI-named manager. Documented in `NewSession`'s docstring ("caller is responsible for uniqueness"), so this is acceptable; flagging only because the UI's collision-aware allocator could be reused by the RPC for parity if you want.
- `isManagerSession(id)`'s fallback (recognize the well-known UUID only) is documented at the function and is fine for its only call site (`updateSessionStatus`), but worth keeping an eye on if future callers come in — they'd get a false negative for a non-primary manager whose row hasn't loaded yet.
- `ManagerSessionRow` uses `person.2.fill` while the primary manager row is text-only. Minor visual inconsistency, not a defect.

### Verification
- `swift test` (CrowCore filter `SessionModel|AppStateLookup|Manager`): **31/31 passed**, including the new `managerSessionsIncludesPrimaryAndAdditional`, `isManagerSessionRecognizesAllManagers`, `managerSessionsExcludedFromActiveAndCompleted`, `sessionManagerKindRoundTrip`, `sessionWorkKindIsNotManager`, `sessionKindRawValues`.
- `swift test` (CrowCLI filter `newSession`): **4/4 passed** (parses manager kind, defaults to nil, rejects bogus, rejects review/job).
- `swift build` (CrowCore): clean.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | — |
| Yellow | `IssueTracker.applyAutoRebase` at `IssueTracker.swift:2067` still gates on `id != managerSessionID` instead of `!isManager`; functionally safe (managers have no `.pr` links) but inconsistent with the rest of the file post-PR. |
| Green | CLI-vs-UI naming collisions are permitted by design — consider reusing the UI's collision-aware allocator in the RPC if you want parity. |
| Green | `isManagerSession` fallback semantics — fine today, worth flagging if new callers appear. |
| Green | `ManagerSessionRow` icon vs. text-only primary row — minor visual nit. |

**Recommendation:** **Approve.** Clean abstraction, careful migration with the right ordering invariant captured in both a comment and a test, good test coverage across model/CLI/service/tracker, and a pre-existing leak fixed as a bonus. The only follow-up worth addressing is the `applyAutoRebase` consistency miss, which is non-blocking.

---

PR #328, opened 2026-05-23
docs: document install / launch path + add make install targets

### LINKED ISSUE #307: Docs: no guidance on how to launch / install Crow for day-to-day use
# Docs: no guidance on how to launch / install Crow for day-to-day use

## Summary

The README's Quick Start ends with `.build/debug/CrowApp` and `docs/getting-started.md`
follows the same pattern. There's no documented path for "I built it, now how do I
actually launch it from anywhere?" — no `make install` target, no recommended
symlink/alias setup, no `.app` bundle install step.

For a new user this leaves a real gap: after `make build` succeeds, the only
documented way to launch Crow is to `cd` back into the repo and run
`.build/debug/CrowApp` every time. Same for the `crow` CLI — it lives at
`.build/debug/crow` with no hint about putting it on `PATH`, even though the
Manager terminal and `/crow-workspace` skill assume `crow` is invokable as a
bare command.

## What's missing

1. **No `make install` target.** `make help` lists `build`, `setup`, `ghostty`,
   `app`, `release`, `sign`, `test`, `check`, `clean`, `clean-all` — no
   `install`.
2. **No README/docs section on "putting it on PATH"** — neither `README.md`
   nor `docs/getting-started.md` mention symlinking, aliasing, or copying the
   `.app` bundle to `/Applications`.
3. **The `crow` CLI is assumed to be on PATH elsewhere in the project**
   (e.g. `CLAUDE.md`, `/crow-workspace` skill, Manager terminal commands all
   call bare `crow ...`), but nothing tells the user how to make that true
   after a fresh build.

## Suggested fix

Add a short "Install / Launch" section to the Quick Start (or
`docs/getting-started.md`) covering at minimum one of:

- A `make install` target that symlinks (or copies) `crow` and `CrowApp`
  into `~/.local/bin` or `/usr/local/bin`.
- Documented manual steps, e.g.:

  ```bash
  ln -s "$PWD/.build/debug/crow"    ~/.local/bin/crow
  ln -s "$PWD/.build/debug/CrowApp" ~/.local/bin/CrowApp
  ```

  …with a note that `~/.local/bin` needs to be on `PATH`.

- Or an alias snippet for `~/.zshrc`:

  ```bash
  alias crow="$HOME/path/to/crow/.build/debug/crow"
  ```

Whichever option, the doc should also call out:

- Debug vs release build paths (`.build/debug/` vs `.build/release/`) and
  which one to point the symlink/alias at.
- That the symlink/alias needs to be re-pointed (or rebuilt) after pulling
  changes.
- Whether there's a supported way to produce a real `.app` bundle to drop in
  `/Applications` (the Makefile has `release` and `sign` targets but no
  documented "drag this to Applications" flow).

## Why it matters

The rest of the project — `CLAUDE.md`, the `/crow-workspace` skill, Manager
terminal commands — all assume `crow` is on `PATH`. Without docs bridging
that assumption, a first-time user who follows the Quick Start exactly ends
up with binaries they have to invoke by full path, which silently breaks the
documented workflows the moment they start a session.

dgershman:
Closes #307

## Problem

After `make build`, the only documented way to launch Crow was by full binary path (`.build/debug/CrowApp`, `.build/debug/crow`). But `CLAUDE.md`, the `/crow-workspace` skill, and the Manager terminal all invoke bare `crow ...` — so a first-time user who follows the Quick Start exactly ends up with binaries they can't invoke the way the rest of the project assumes, silently breaking the documented workflows the moment they start a session.

## Changes

**`Makefile`**
- `install` — symlinks `crow` + `CrowApp` into `~/.local/bin`. Overridable: `BINDIR=` (e.g. `/usr/local/bin`) and `CONFIG=` (`debug`/`release`). Guards against missing binaries (errors with a build-first hint) and prints a PATH reminder only when the target dir isn't on `PATH`.
- `install-app` — copies `Crow.app` into `/Applications` (run `make release` first).
- `uninstall` — removes the symlinks.
- New entries in `make help`.

Symlinks (not copies) are deliberate: `swift build` overwrites `.build/debug/` binaries in place, so the symlinks self-heal on rebuild — you only re-run `make install` when switching debug↔release or after `make clean`.

**`README.md`** — adds an "Install (put `crow` on your PATH)" subsection to Quick Start.

**`docs/getting-started.md`** — adds a "6. Install (Optional)" section (PATH setup, `BINDIR`/`CONFIG` overrides, rebuild/re-point behavior, `/Applications` GUI install) plus `install`/`install-app`/`uninstall` rows in the Makefile Targets table.

## Verification

- `make help` lists the three new targets.
- `make install` with no binaries → errors with build-first hint (exit 1).
- `make install-app` with no `Crow.app` → errors (exit 1).
- `make install` (happy path, stub binaries) → creates absolute-path symlinks; PATH NOTE prints only when `BINDIR` isn't on `PATH`.
- `make uninstall` → removes the symlinks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,15 @@
-.PHONY: build setup ghostty app release sign clean clean-all check test help
+.PHONY: build setup ghostty app release sign install install-app uninstall clean clean-all check test help
 
 FRAMEWORKS_DIR := Frameworks
 XCFW := $(FRAMEWORKS_DIR)/GhosttyKit.xcframework
 SUBMODULE_MARKER := vendor/ghostty/build.zig
+
+# Install destination and build config (override on the command line, e.g.
+# `make install BINDIR=/usr/local/bin` or `make install CONFIG=release`).
+PREFIX    ?= $(HOME)/.local
+BINDIR    ?= $(PREFIX)/bin
+CONFIG    ?= debug
+BUILD_OUT := .build/$(CONFIG)
 
 # Default target
 build: setup ghostty app
@@ -19,6 +26,9 @@ help:
 	@echo "  app        Swift build only (debug)"
 	@echo "  release    Release build + .app bundle"
 	@echo "  sign       Sign, create DMG, and notarize (requires DEVELOPER_ID_APPLICATION)"
+	@echo "  install    Symlink crow + CrowApp into ~/.local/bin (override BINDIR=, CONFIG=release)"
+	@echo "  install-app Copy Crow.app into /Applications (run 'make release' first)"
+	@echo "  uninstall  Remove installed crow + CrowApp symlinks"
 	@echo "  clean      Remove .build/ (keeps ghostty framework)"
 	@echo "  clean-all  Remove .build/ and Frameworks/ (full rebuild)"
 	@echo ""
@@ -57,6 +67,28 @@ release: $(XCFW)
 
 sign: release
 	bash scripts/sign-and-notarize.sh
+
+# --- Install ---
+
+install:
+	@test -x "$(CURDIR)/$(BUILD_OUT)/crow" && test -x "$(CURDIR)/$(BUILD_OUT)/CrowApp" || \
+		{ echo "ERROR: binaries not found in $(BUILD_OUT)/. Run 'make build' (debug) or 'make release' (then 'make install CONFIG=release') first."; exit 1; }
+	@mkdir -p "$(BINDIR)"
+	@ln -sf "$(CURDIR)/$(BUILD_OUT)/crow" "$(BINDIR)/crow"
+	@ln -sf "$(CURDIR)/$(BUILD_OUT)/CrowApp" "$(BINDIR)/CrowApp"
+	@echo "Symlinked crow + CrowApp into $(BINDIR) (from $(BUILD_OUT)/)"
+	@case ":$$PATH:" in *":$(BINDIR):"*) ;; \
+		*) echo "NOTE: $(BINDIR) is not on PATH. Add to your shell rc: export PATH=\"$(BINDIR):\$$PATH\"";; esac
+
+install-app:
+	@test -d "$(CURDIR)/Crow.app" || { echo "ERROR: Crow.app not found. Run 'make release' first."; exit 1; }
+	rm -rf "/Applications/Crow.app"
+	cp -R "$(CURDIR)/Crow.app" "/Applications/Crow.app"
+	@echo "Installed Crow.app to /Applications"
+
+uninstall:
+	@rm -f "$(BINDIR)/crow" "$(BINDIR)/CrowApp"
+	@echo "Removed crow + CrowApp symlinks from $(BINDIR)"
 
 # --- Test ---
 
--- a/README.md
+++ b/README.md
@@ -56,6 +56,26 @@ On first launch, a setup wizard guides you through choosing your development roo
 
 > **Note:** The required GitHub scope is the **write** `project` scope — `read:project` is insufficient because Crow updates ticket status via the `updateProjectV2ItemFieldValue` GraphQL mutation. See [docs/getting-started.md](docs/getting-started.md#3-github-authentication) for details.
 
+### Install (put `crow` on your PATH)
+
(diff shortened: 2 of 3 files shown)


dhilgaertner (approved):
## Code & Security Review

Docs + Makefile-only PR. No code paths changed, no new runtime behavior beyond what `make install` does locally. Verified all five scenarios listed in the PR description on this checkout.

### Critical Issues
None.

### Security Review

**Strengths:**
- No new runtime code, no new attack surface.
- Install targets are scoped to user-controlled `BINDIR` (defaults to `$HOME/.local/bin`); no sudo, no system paths required by default.
- `mkdir -p`, `ln -sf`, and `cp -R` invocations are properly double-quoted; `BUILD_OUT := .build/$(CONFIG)` keeps the build path inside the repo even with an unusual `CONFIG` value, and a non-existent binary trips the `test -x` guard before any FS mutation.
- Symlinks (not copies) prevent stale binaries — a real ergonomic + correctness win over the alternative.

**Concerns:**
- `install-app` runs `rm -rf "/Applications/Crow.app"` before `cp -R`. Destructive but standard for `.app` replacement; no user-owned data lives inside a bundle. Worth a one-line caveat in `docs/getting-started.md` ("Replaces any existing `/Applications/Crow.app`.") so it isn't a surprise.

### Code Quality

**Verified behavior (`Makefile:73-91`):**
- `make install` with no binaries → `Makefile:74-75` errors with build-first hint, exits 1. ✓
- `make install-app` with no `Crow.app` → `Makefile:84` errors, exits 1. ✓
- Happy-path install with stub binaries → absolute symlinks created in `BINDIR`. ✓
- `PATH` NOTE at `Makefile:80-81` correctly suppressed when `BINDIR` is already on `PATH`. ✓
- `make uninstall` removes both symlinks. ✓

**Minor suggestions (non-blocking):**

- `Makefile:77-78` — `ln -sf` silently clobbers any pre-existing non-symlink at `$BINDIR/crow` (verified: a stub shell script named `crow` was replaced by the symlink with no warning). Realistically unlikely given the project-specific name, but a safer pattern would be:
  ```
  @if [ -e "$(BINDIR)/crow" ] && [ ! -L "$(BINDIR)/crow" ]; then \
      echo "ERROR: $(BINDIR)/crow exists and is not a symlink; refusing to overwrite. Remove it manually first."; exit 1; \
  fi
  ```
  Same idea for `CrowApp`. Up to you — current behavior matches what most users expect from `make install`.

- `Makefile:89-91` — `uninstall` uses `rm -f` and will unconditionally delete whatever lives at those paths, including a non-symlink. Symmetric to the `install` concern. Could guard with `test -L`.

- `BINDIR` is not remembered across invocations. A user who runs `make install BINDIR=/usr/local/bin` and later runs a bare `make uninstall` ends up with orphaned symlinks at `/usr/local/bin` while `uninstall` reports "Removed..." against the default `~/.local/bin`. A note in the `## 6. Install (Optional)` section ("Pass the same `BINDIR` to `make uninstall` that you used for `make install`.") would prevent this footgun.

- `Makefile:30` — the `install-app` help line (`install-app Copy Crow.app ...`) loses the 2-space description column the rest of the targets use because the target name is 10 chars instead of ≤8. Purely cosmetic.

- README and `docs/getting-started.md` are accurate, internally consistent, and the worked example flows (`make install`, `make install CONFIG=release`, `make uninstall`) all line up with the implementation. The "Rebuilds and re-pointing" subsection is especially well-pitched for first-time users who will otherwise wonder whether they need to reinstall after `make build`.

### Summary Table
| Priority | Issue |
|----------|-------|
| Green | `install` silently clobbers a non-symlink at `$BINDIR/crow` (`Makefile:77-78`) |
| Green | `uninstall` removes whatever is at those paths via `rm -f`, even if not a symlink (`Makefile:90`) |
| Green | `BINDIR` mismatch between `install` and `uninstall` can orphan symlinks — doc note suggested |
| Green | `install-app Copy ...` help line breaks column alignment (`Makefile:30`) |
| Green | One-line caveat that `install-app` replaces an existing `/Applications/Crow.app` |

**Recommendation:** **Approve.** The PR delivers exactly what #307 asked for: a documented, idempotent install path so the workflows that assume bare `crow` on `PATH` actually work for a first-time user. Implementation is small, well-guarded (existence checks before any FS mutation, PATH reminder, clear error messages), and the docs match the behavior. The items above are all Green/nice-to-have; none should block merge.

---

PR #331, opened 2026-05-23
feat: jobs — provider-sourced repo picker scoped to a workspace (closes #327)

### LINKED ISSUE #327: Crow Jobs: repo field should be free-form; drop workspace (infer from repo name)
Follow-up to the Jobs feature (#317, PR #323).

## Current behavior
In the Settings → Jobs form, a job is scoped by **two** fields:
- a **Workspace** picker, and
- a **Repo** dropdown populated from the selected workspace's `alwaysInclude` repos.

## Desired behavior
- **Repo** should be a plain **free-form text field** (just the repo's folder name, e.g. `api`).
- **Workspace** should be **removed entirely** — it's implied by the repo name.

## Why
The workspace is redundant: repos live on disk at `{devRoot}/{workspace}/{repo}`, so the repo name already determines its location. Requiring a workspace pick and constraining repos to `alwaysInclude` is unnecessary friction.

## Mechanism
`GitManager.discoverRepos()` already enumerates every real checkout under the dev root as `RepoInfo(name, path, workspace, …)`. When a job fires, resolve the free-form repo name to a concrete checkout by matching `RepoInfo.name`, then create the worktree as a sibling of that path (unchanged naming).

## Acceptance criteria
- [ ] The Jobs form has a single free-form repo text field and no workspace field.
- [ ] `JobConfig` no longer stores `workspace`; existing persisted jobs still decode (the stale key is ignored, `repo` preserved).
- [ ] When a job fires, the repo is resolved by name under the dev root and the worktree/session is created in that repo's checkout.

## Known limitation
If the same repo folder name exists in multiple workspaces, the first match wins. Acceptable for v1.

dgershman:
Closes #327

## Summary
The Jobs form scopes a job to a **workspace + a repo chosen from that workspace's provider** (GitHub/GitLab). The repo list is built by expanding the workspace's **"Always Include Repos"** specs — `owner/*` org globs and explicit `owner/repo` slugs — against the provider. Repos not yet checked out are **cloned on demand** when the job fires.

> Note: the branch's first commit (`df1b2c8`) implemented a free-form repo field and dropped the workspace; that direction was **superseded** by maintainer feedback. The net diff is the workspace + provider-sourced picker described here.

## Changes
- **`ProviderManager`** (net-new): `listRepos(owner:provider:host:)` shells out (`gh repo list` / `glab api`); `reposForSpecs(_:provider:host:)` expands `alwaysInclude` specs (globs + explicit slugs) into a de-duplicated, sorted slug list; `classifySpec` is a `nonisolated static` helper with unit tests.
- **`AppState`**: new `onListWorkspaceRepos` hook (async, returns slugs).
- **`AppDelegate`**: wires the hook to a **reused** `ProviderManager`, caches results per `(workspace, specs)` with a 5-min TTL, and logs unknown-provider fallback to GitHub.
- **`JobConfig`**: `workspace` field restored; `repo` now holds an `owner/repo` slug. Decoding stays forward/back-compatible — legacy `{workspace, repo}` jobs **and** the transient free-form `repo`-only jobs both decode.
- **`JobFormView`**: workspace `Picker` + a provider-loaded repo `Picker` (spinner while loading; stale loads cancelled via a generation token; generic empty-state hint).
- **`SessionService.runJob`**: resolves `{devRoot}/{workspace}/{repoFolder}` (pure `jobRepoFolder` / `jobWorktreeLayout` helpers) and **clones on demand** (gh/glab, `GITLAB_HOST` via an `env:` dict on `shell`) when the repo isn't checked out. Blank-workspace (legacy) jobs fall back to a sorted discover-by-name.
- **`SettingsView`** (workspace/repo row + add-a-workspace-first guard) and **`WorkspaceFormView`** help text updated for the new spec format.

## Testing
arm64 `swift build` succeeds. Suites pass: **CrowCore 180**, **CrowProvider 33** (incl. `classifySpec`/`reposForSpecs` tests), **CrowUI 31**, root **122** (incl. new `JobRepoResolutionTests`).

## Known limitation
Duplicate repo folder names across workspaces resolve to the first match (now sorted, so deterministic).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Backfill of merged PRs since the 0.1.0 release, grouped by theme.
 - #214 — Optional auto-respond toggles: when enabled, Crow types an instruction into the session's Claude Code terminal in response to changes-requested reviews and failed CI checks. Off by default.
 - #222 — PRs opened from a Crow session are auto-labeled `crow:auto`.
 - #228 — Settings split into discrete tabs; every automation toggle lives under Settings → Automation. New `docs/automation.md` covers the full lifecycle.
+- #327 — Scheduled Jobs scope to a workspace + repo, where the repo picker is populated from the workspace's provider by expanding its "Always Include Repos" specs (`owner/*` globs or `owner/repo` slugs). Repos not yet checked out are cloned on demand when the job fires.
 
 ### Review Board & Sessions
 
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -135,6 +135,11 @@ public final class AppState {
     /// Called to promote selected patterns to the global settings.
     public var onPromoteToGlobal: ((Set<String>) -> Void)?
 
+    /// Lists the repos available to a workspace, as `owner/repo` slugs, by
+    /// expanding its `alwaysInclude` specs against the provider. Wired in
+    /// AppDelegate; used by the Jobs form's repo picker.
+    public var onListWorkspaceRepos: ((WorkspaceInfo) async -> [String])?
+
     /// Called when the user clicks the gear icon in the sidebar toolbar.
     /// AppDelegate wires this to its `showSettings()` method.
     public var onShowSettings: (() -> Void)?
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -202,6 +202,11 @@ public struct WorkspaceInfo: Identifiable, Codable, Sendable, Equatable {
         if name.unicodeScalars.contains(where: { unsafeCharacters.contains($0) }) {
             return "Name cannot contain /, :, or null characters"
         }
+        // The name becomes a path component under devRoot; "." / ".." would
+        // resolve outside the intended directory.
+        if name == "." || name == ".." {
+            return "Name cannot be “.” or “..”"
+        }
         let lowercased = name.lowercased()
         if existingNames.contains(where: { $0.lowercased() == lowercased }) {
             return "A workspace with this name already exists"
--- a/Packages/CrowCore/Sources/CrowCore/Models/JobConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/JobConfig.swift
@@ -1,20 +1,24 @@
 import Foundation
 
 /// A scheduled job: one or more prompts that fire automatically on a schedule,
-/// scoped to a repo.
+/// scoped to a repo within a workspace.
 ///
-/// When a job fires, the scheduler resolves `repo` to a checkout under the dev
-/// root (by folder name), creates a fresh worktree + session + Claude Code
-/// terminal there, and sends `prompts` in order. Persisted as part of
-/// `AppConfig` in `{devRoot}/.claude/config.json`.
+/// When a job fires, the scheduler resolves the repo under
+/// `{devRoot}/{workspace}/` — cloning it on demand if it isn't checked out yet —
+/// then creates a fresh worktree + session + Claude Code terminal there and
+/// sends `prompts` in order. Persisted as part of `AppConfig` in
+/// `{devRoot}/.claude/config.json`.
 ///
 /// Like `WorkspaceInfo`, decoding is forward-compatible: missing keys fall back
-/// to defaults so older config files keep working (a legacy `workspace` key is
-/// simply ignored).
+/// to defaults so older config files keep working.
 public struct JobConfig: Identifiable, Codable, Sendable, Equatable {
     public let id: UUID
     public var name: String
-    /// Repo folder name. Resolved to a checkout under the dev root at run time.
+    /// Workspace name (matches `WorkspaceInfo.name`); the repo's parent folder
+    /// under the dev root and the source of its provider/host.
+    public var workspace: String
+    /// The repo as an `owner/repo` slug (chosen from the workspace's provider).
+    /// Its last path component is the local folder name under the workspace.
     public var repo: String
     /// Ordered prompts. The first launches Claude; the rest are sent after launch.
     public var prompts: [String]
@@ -27,6 +31,7 @@ public struct JobConfig: Identifiable, Codable, Sendable, Equatable {
     public init(
         id: UUID = UUID(),
         name: String,
+        workspace: String,
         repo: String,
         prompts: [String],
         schedule: JobSchedule,
@@ -36,6 +41,7 @@ public struct JobConfig: Identifiable, Codable, Sendable, Equatable {
     ) {
         self.id = id
         self.name = name
+        self.workspace = workspace
         self.repo = repo
         self.prompts = prompts
         self.schedule = schedule
@@ -48,6 +54,7 @@ public struct JobConfig: Identifiable, Codable, Sendable, Equatable {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         id = try container.decode(UUID.self, forKey: .id)
         name = try container.decode(String.self, forKey: .name)
+        workspace = try container.decodeIfPresent(String.self, forKey: .workspace) ?? ""
         repo = try container.decodeIfPresent(String.self, forKey: .repo) ?? ""
         prompts = try container.decodeIfPresent([String].self, forKey: .prompts) ?? []
         schedule = try container.decodeIfPresent(JobSchedule.self, forKey: .schedule) ?? .interval(seconds: 86400)
@@ -57,7 +64,7 @@ public struct JobConfig: Identifiable, Codable, Sendable, Equatable {
     }
 
     private enum CodingKeys: String, CodingKey {
-        case id, name, repo, prompts, schedule, enabled, lastRunAt, createdAt
+        case id, name, workspace, repo, prompts, schedule, enabled, lastRunAt, createdAt
     }
 
     /// Validate a job name, returning an error message or `nil` if valid.
--- a/Packages/CrowCore/Sources/CrowCore/Models/JobConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/JobConfig.swift
@@ -1,20 +1,20 @@
 import Foundation
 
 /// A scheduled job: one or more prompts that fire automatically on a schedule,
-/// scoped to a repo within a workspace.
+/// scoped to a repo.
 ///
-/// When a job fires, the scheduler creates a fresh worktree + session + Claude
-/// Code terminal in `{devRoot}/{workspace}/{repo}` and sends `prompts` in order.
-/// Persisted as part of `AppConfig` in `{devRoot}/.claude/config.json`.
+/// When a job fires, the scheduler resolves `repo` to a checkout under the dev
+/// root (by folder name), creates a fresh worktree + session + Claude Code
+/// terminal there, and sends `prompts` in order. Persisted as part of
+/// `AppConfig` in `{devRoot}/.claude/config.json`.
 ///
 /// Like `WorkspaceInfo`, decoding is forward-compatible: missing keys fall back
-/// to defaults so older config files keep working.
+/// to defaults so older config files keep working (a legacy `workspace` key is
+/// simply ignored).
 public struct JobConfig: Identifiable, Codable, Sendable, Equatable {
     public let id: UUID
     public var name: String
-    /// Workspace name (matches `WorkspaceInfo.name`).
-    public var workspace: String
-    /// Repo folder name within the workspace (the checkout at `{devRoot}/{workspace}/{repo}`).
+    /// Repo folder name. Resolved to a checkout under the dev root at run time.
     public var repo: String
     /// Ordered prompts. The first launches Claude; the rest are sent after launch.
     public var prompts: [String]
@@ -27,7 +27,6 @@ public struct JobConfig: Identifiable, Codable, Sendable, Equatable {
     public init(
         id: UUID = UUID(),
         name: String,
-        workspace: String,
         repo: String,
         prompts: [String],
         schedule: JobSchedule,
@@ -37,7 +36,6 @@ public struct JobConfig: Identifiable, Codable, Sendable, Equatable {
     ) {
         self.id = id
         self.name = name
-        self.workspace = workspace
         self.repo = repo
         self.prompts = prompts
         self.schedule = schedule
@@ -50,7 +48,6 @@ public struct JobConfig: Identifiable, Codable, Sendable, Equatable {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         id = try container.decode(UUID.self, forKey: .id)
         name = try container.decode(String.self, forKey: .name)
-        workspace = try container.decodeIfPresent(String.self, forKey: .workspace) ?? ""
         repo = try container.decodeIfPresent(String.self, forKey: .repo) ?? ""
         prompts = try container.decodeIfPresent([String].self, forKey: .prompts) ?? []
         schedule = try container.decodeIfPresent(JobSchedule.self, forKey: .schedule) ?? .interval(seconds: 86400)
@@ -60,7 +57,7 @@ public struct JobConfig: Identifiable, Codable, Sendable, Equatable {
     }
 
     private enum CodingKeys: String, CodingKey {
-        case id, name, workspace, repo, prompts, schedule, enabled, lastRunAt, createdAt
+        case id, name, repo, prompts, schedule, enabled, lastRunAt, createdAt
     }
 
     /// Validate a job name, returning an error message or `nil` if valid.
--- a/Packages/CrowCore/Tests/CrowCoreTests/AppConfigTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AppConfigTests.swift
@@ -252,6 +252,12 @@ import Testing
     #expect(WorkspaceInfo.validateName("My/Org", existingNames: []) != nil)
     #expect(WorkspaceInfo.validateName("My:Org", existingNames: []) != nil)
 
+    // Path-traversal names (would escape devRoot)
+    #expect(WorkspaceInfo.validateName(".", existingNames: []) != nil)
+    #expect(WorkspaceInfo.validateName("..", existingNames: []) != nil)
+    // A name merely containing a dot is still fine.
+    #expect(WorkspaceInfo.validateName("my.org", existingNames: []) == nil)
+
     // Valid with existing names that don't conflict
     #expect(WorkspaceInfo.validateName("NewOrg", existingNames: ["OtherOrg"]) == nil)
 }
--- a/Packages/CrowCore/Tests/CrowCoreTests/JobConfigTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/JobConfigTests.swift
@@ -6,7 +6,8 @@ import Testing
     let config = AppConfig(jobs: [
         JobConfig(
             name: "Nightly Audit",
-            repo: "api",
+            workspace: "RadiusMethod",
+            repo: "radiusmethod/api",
             prompts: ["Run the audit", "Summarize findings"],
             schedule: .interval(seconds: 3600),
             enabled: true
@@ -21,7 +22,8 @@ import Testing
     #expect(decoded.jobs.count == 1)
     let job = decoded.jobs[0]
     #expect(job.name == "Nightly Audit")
-    #expect(job.repo == "api")
+    #expect(job.workspace == "RadiusMethod")
+    #expect(job.repo == "radiusmethod/api")
     #expect(job.prompts == ["Run the audit", "Summarize findings"])
     #expect(job.schedule == .interval(seconds: 3600))
     #expect(job.enabled == true)
@@ -30,7 +32,8 @@ import Testing
 @Test func jobConfigRoundTripDailyAt() throws {
     let job = JobConfig(
         name: "Standup",
-        repo: "web",
+        workspace: "Acme",
+        repo: "acme/web",
         prompts: ["Generate the standup report"],
         schedule: .dailyAt(hour: 9, minute: 30, weekdays: [2, 3, 4, 5, 6]),
         lastRunAt: Date(timeIntervalSince1970: 1_700_000_000)
@@ -39,21 +42,25 @@ import Testing
     let data = try JSONEncoder().encode(job)
     let decoded = try JSONDecoder().decode(JobConfig.self, from: data)
 
+    #expect(decoded.workspace == "Acme")
+    #expect(decoded.repo == "acme/web")
     #expect(decoded.schedule == .dailyAt(hour: 9, minute: 30, weekdays: [2, 3, 4, 5, 6]))
     #expect(decoded.lastRunAt == Date(timeIntervalSince1970: 1_700_000_000))
 }
 
-/// A job persisted before the workspace field was dropped must still decode:
-/// the stale `workspace` key is ignored and `repo` is preserved.
-@Test func jobConfigForwardCompatLegacyWorkspaceKey() throws {
+/// A job persisted before the workspace field returned (the brief free-form
+/// `repo`-only era) must still decode: the missing `workspace` defaults to ""
+/// and `repo` is preserved so it resolves by folder name at run time.
+@Test func jobConfigBackCompatMissingWorkspaceKey() throws {
     let json = """
-    {"id":"\(UUID().uuidString)","name":"Legacy","workspace":"Acme","repo":"api",
+    {"id":"\(UUID().uuidString)","name":"FreeForm","repo":"api",
      "prompts":["go"],"schedule":{"type":"interval","seconds":3600},
      "enabled":true,"createdAt":0}
     """
     let decoded = try JSONDecoder().decode(JobConfig.self, from: Data(json.utf8))
+    #expect(decoded.workspace == "")
     #expect(decoded.repo == "api")
-    #expect(decoded.name == "Legacy")
+    #expect(decoded.name == "FreeForm")
     #expect(decoded.prompts == ["go"])
 }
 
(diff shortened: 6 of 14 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- `job.repo` flows into git via `Process.arguments`, never a shell — no command-injection surface.
- Path-traversal is bounded: lookup matches against `RepoInfo.name`, which comes from `FileManager.contentsOfDirectory` (single directory components, no `/` or `..`). An attacker-supplied `repo: "../etc"` simply fails to match.
- The worktree path is composed from internally-derived `repoInfo.path` + an internal slug/timestamp, so a malicious `job.repo` cannot escape the discovered repo's parent dir.
- Forward-compat decode is implemented correctly: the legacy `workspace` key drops out of `CodingKeys` and JSONDecoder silently ignores it, with a dedicated test (`jobConfigForwardCompatLegacyWorkspaceKey`) pinning the behavior.

**Concerns:**
- None security-blocking.

### Code Quality
- `Sources/Crow/App/SessionService.swift:1318` — `discoverRepos()` order is filesystem-dependent (`FileManager.contentsOfDirectory(atPath:)` makes no ordering guarantee). The PR documents "first match wins," but for two repos of the same name this can shift between runs and across machines. Worth either sorting the result before `.first(where:)` or noting explicitly that the binding is non-deterministic.
- **Silent rebind on upgrade**: legacy jobs persisted with `{workspace: "Acme", repo: "api"}` decode to `repo: "api"` and drop the workspace context. If the user has another workspace whose repo is also `api`, the job may quietly run against a different checkout than originally intended. The decoder is correct; the migration semantics are just worth flagging — a one-time log/warn when a job's resolved repo path differs from any prior recorded location would help, but not required.
- `Packages/CrowUI/Sources/CrowUI/JobFormView.swift:121` — the form no longer validates `repo` against any known list; an invalid name only surfaces at run time as an `NSLog`. Minor UX regression from the previous picker, but acceptable given the free-form intent.
- `discoverRepos()` is called on every job firing. Cheap for typical dev roots, but scales with `workspaces × repos` per tick. Caching is overkill for now; flagging for future awareness.
- New test JSON uses `createdAt: 0` (decodes to 2001-01-01 under default `dateDecodingStrategy`). Harmless — just not asserted.

### Summary Table
| Priority | Issue |
|----------|-------|
| Yellow | `first(where:)` over `discoverRepos()` is filesystem-order-dependent; consider sorting for deterministic resolution of duplicate names. |
| Yellow | Legacy `{workspace, repo}` jobs may silently rebind to a different `repo` after upgrade if the name is duplicated across workspaces. |
| Green | `JobFormView` lost compile-time/UI validation of repo names — failures now deferred to run time. |
| Green | `discoverRepos()` runs per job tick; consider caching if dev root grows large. |
| Green | Test fixture uses `createdAt: 0`; consider asserting decoded `createdAt` or using an explicit date. |

**Recommendation:** Approve. The change is well-scoped, the model migration is correct and tested, and there are no security or correctness issues. The "first match wins" caveat is acknowledged in the PR description; the only nudge worth considering is sorting the discovered repos so duplicate-name resolution is deterministic.

dgershman:
## Approach changed — re-review requested

Heads up: this pushes a **different design** than the one previously approved. Per maintainer direction, the Jobs repo field is no longer free-form text — it's now a **dropdown sourced from the workspace's provider**, and the **workspace picker is back**.

### What changed since the approved commit
- **`JobConfig`** re-adds `workspace`; `repo` now stores an `owner/repo` slug. Decoding stays forward/back-compatible (legacy `{workspace, repo}` jobs and the brief free-form `repo`-only jobs both decode).
- **`ProviderManager`** (net-new): `reposForSpecs` expands a workspace's **`alwaysInclude`** entries — `owner/*` globs (via `gh repo list` / `glab api`) and explicit `owner/repo` slugs — into the repo list. `classifySpec` is a `nonisolated static` helper with unit tests.
- **`JobFormView`**: workspace `Picker` + a provider-loaded repo `Picker` (spinner while loading, empty-state hint), wired through `AppState.onListWorkspaceRepos`.
- **`SessionService.runJob`**: resolves the repo at `{devRoot}/{workspace}/{repoFolder}` and **clones on demand** (gh/glab, `GITLAB_HOST` for self-hosted) when it isn't checked out. Empty-workspace jobs fall back to discover-by-name.
- **`WorkspaceFormView`**: "Always Include Repos" help text now documents the `owner/*` / `owner/repo` spec format.

### Testing
arm64 `swift build` succeeds. Suites pass: **CrowCore 180**, **CrowProvider 33** (incl. new spec tests), **CrowUI 31**, root **117**.

### Note
- Provider listing for GitLab uses `glab api groups/<owner>/projects` (best-effort); GitHub via `gh repo list` is the primary path.

dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues

**1. PR title and body describe the opposite of the code being merged.** This branch contains two commits:

- `df1b2c8` — *"jobs — free-form repo field, drop workspace (closes #327)"*
- `e5a40a6` — *"jobs — provider-sourced repo picker scoped to a workspace (#327). Supersedes the free-form repo direction."*

The current PR title (`feat: jobs — free-form repo field, drop workspace (closes #327)`) and body describe the first, *superseded* commit. The actual net diff (302 / 44 across 11 files) implements the second — a workspace-scoped, provider-sourced repo picker with on-demand cloning. The CHANGELOG entry added in this PR (line 24) correctly describes the *new* direction, which contradicts the PR description.

This matters because squash-merging will turn the misleading title into a permanent commit message in `git log`. **Please rewrite the PR title and description to reflect the actual final code** (workspace + provider-sourced repo picker, `alwaysInclude` glob expansion, clone-on-demand, the new `onListWorkspaceRepos` AppState hook, `ProviderManager.classifySpec`/`reposForSpecs`/`listRepos`, and the back-compat decode for legacy free-form jobs).

### Security Review

**Strengths:**
- All shell-outs go through `Process` with explicit `arguments` arrays — no shell interpretation, no command injection from `job.repo`, `workspace.host`, or `job.workspace`.
- `JobConfig` decoding is forward- and back-compatible; the new back-compat decode test (`jobConfigBackCompatMissingWorkspaceKey`) covers the transient free-form-only schema.
- `ProviderManager.listRepos` swallows provider CLI failures with an `NSLog` instead of throwing — degrades to an empty picker rather than leaking error details to the UI.

**Concerns:**
- None that rise to a security blocker.

### Code Quality

**Yellow — should fix:**

1. **`Sources/Crow/App/SessionService.swift:1422` uses `/usr/bin/env`'s positional `KEY=VAL` argv parsing for the GitLab clone:**
   ```swift
   _ = try await shell(\"GITLAB_HOST=\\(host)\", \"glab\", \"repo\", \"clone\", job.repo, destination)
   ```
   This works but is fragile — it depends on the `shell` helper's `executableURL` staying as `/usr/bin/env`, and it puts an env var into the args vector. `ProviderManager.shell` (lines 210-218) already takes an explicit `env:` dictionary and uses `ShellEnvironment.shared.merging(env)`. Extend the `SessionService.shell` helper the same way and pass `env: [\"GITLAB_HOST\": host]` instead. Same fix removes the now-unneeded branch on `host.isEmpty`.

2. **`Sources/Crow/App/AppDelegate.swift:600-606` constructs a fresh `ProviderManager()` and shells out on every job-form open.** No caching. For a workspace with `owner/*` and a few hundred repos, opening the form repeatedly thrashes `gh repo list --limit 1000`. Cache per-workspace results (per session, or with a short TTL).

3. **`Sources/Crow/App/AppDelegate.swift:603` silently maps any unknown provider string to `.github`:** `Provider(rawValue: ws.provider) ?? .github`. A workspace with a typo or a future provider value would silently behave as GitHub. At minimum log the fallback; better, surface it.

4. **No tests for `cloneJobRepo` or the workspace+slug resolution branch in `runJob` (`SessionService.swift:1306-1403`).** The clone-on-demand flow is the new headline behavior in the PR and is currently uncovered. A minimum-viable test: a job whose `repo` is bare (no `/`) hits the back-compat `discoverRepos` path; a job whose `repo` is `owner/repo` resolves to `{devRoot}/{workspace}/{repoFolder}`. `cloneJobRepo` itself can at least be unit-tested for the early-return on a non-slug `repo`.

5. **`Packages/CrowUI/Sources/CrowUI/JobFormView.swift:182` uses the workspace *name* as the suggested provider owner** in the empty-state hint (`\\(workspace.lowercased())/*`). Workspace names and GitHub/GitLab org names are unrelated — \"Personal\" or \"Work\" as a workspace name would produce a nonsense suggestion. Use a generic `owner/*` placeholder or omit the example.

6. **`Packages/CrowUI/Sources/CrowUI/JobFormView.swift:143-151` doesn't cancel in-flight `loadRepos()` on rapid workspace switches.** Last-write-wins is OK in practice, but a stale slow response can clobber a fast subsequent one. A generation token or a `Task` ref that gets cancelled in the `onChange(of: workspace)` handler would fix it.

**Green — consider:**

7. **`ProviderManager.reposForSpecs` (`Packages/CrowProvider/Sources/CrowProvider/ProviderManager.swift:123-138`) runs `listRepos` sequentially across specs.** Multiple `owner/*` globs in one workspace serialize. A `withTaskGroup` would parallelize without much complexity.

8. **`ProviderManager.listRepos` silently returns `[]` on failure** (line 165). The current empty-state hint suggests the user set their `alwaysInclude`, but the cause might be `gh auth`, network, or rate limit. Returning a `Result`/error from `onListWorkspaceRepos` would let the form distinguish \"no repos configured\" from \"lookup failed\".

### Build & Tests

`swift build` couldn't run locally (`GhosttyKit.xcframework` is missing a binary artifact — local environment issue, not a PR issue). Ran package tests directly:

- `Packages/CrowCore swift test`: **180 tests passed** (includes the new `jobConfigBackCompatMissingWorkspaceKey`).
- `Packages/CrowProvider swift test`: **33 tests passed** (includes 8 new `classifySpec`/`reposForSpecs` tests).

### Summary Table

| Priority | Issue |
|----------|-------|
| Red | PR title/body describe the superseded `df1b2c8` direction; actual code is the opposite (`e5a40a6`). Rewrite before squash-merge. |
| Yellow | `shell(\"GITLAB_HOST=…\", …)` argv-env hack — refactor `SessionService.shell` to take an `env:` dict like `ProviderManager.shell`. |
| Yellow | `ProviderManager()` re-created per `onListWorkspaceRepos` call; no caching of `gh repo list` results. |
| Yellow | Silent provider-string fallback to `.github` in `AppDelegate`. |
| Yellow | No tests for `runJob`'s workspace+slug resolution or `cloneJobRepo`. |
| Yellow | Empty-state hint suggests workspace name as `owner` — misleading. |
| Yellow | `loadRepos()` not cancelled on rapid workspace switches. |
| Green  | Serial spec expansion in `reposForSpecs`. |
| Green  | `listRepos` swallows errors silently — no way to show \"auth required\". |

**Recommendation:** Request Changes — the PR description mismatch is a must-fix because squash-merge would produce a permanent, misleading commit message. The yellow items are pre-merge polish; #1 (the GitLab env hack) and #4 (no test coverage on the new clone-on-demand path) are the most important.

---

dgershman:
Thanks for the thorough review. Addressed in `614fee4`, plus the PR title/body rewrite:

**Red**
- ✅ **PR title/body** rewritten to describe the actual final code (workspace + provider-sourced picker, `alwaysInclude` glob expansion, clone-on-demand, `onListWorkspaceRepos`, `ProviderManager.classifySpec`/`reposForSpecs`/`listRepos`, back-compat decode), with a note that `df1b2c8` was superseded.

**Yellow**
1. ✅ `SessionService.shell` now takes an `env:` dict (mirrors `ProviderManager.shell`); GitLab clone passes `env: ["GITLAB_HOST": host]` — argv-env hack and the `host.isEmpty` branch are gone.
2. ✅ `AppDelegate` reuses a single `ProviderManager` and caches expanded results per `(workspace, specs)` with a 5-min TTL (the spec string is in the key, so editing `alwaysInclude` invalidates).
3. ✅ Unknown provider strings now log before falling back to GitHub.
4. ✅ Extracted pure `jobRepoFolder` / `jobWorktreeLayout` helpers and added `JobRepoResolutionTests` (slug → folder, nested GitLab group, bare-name passthrough, `{devRoot}/{workspace}/{repoFolder}` composition). `cloneJobRepo`'s non-slug early return is now covered indirectly via the bare-name folder case.
5. ✅ Empty-state hint uses a generic `owner/* or owner/repo` example (no longer the unrelated workspace name) and mentions checking gh/glab auth.
6. ✅ `loadRepos()` uses a generation token; a stale slow response is dropped if the workspace changed.

Bonus: sorted the back-compat `discoverRepos` fallback so a duplicated bare name binds deterministically (the first review's nudge), and asserted the decoded `createdAt` in the back-compat test.

**Green — deferred, with reasoning**
7. **Parallelize `reposForSpecs` with `withTaskGroup`** — held off: `ProviderManager.shell` runs `process.waitUntilExit()` synchronously on the actor, so a task group wouldn't actually parallelize without also making the shell non-blocking (a larger change). Typical workspaces have 0–1 globs, so the serial cost is negligible. Happy to do the shell rework in a follow-up if you'd like.
8. **`Result`-typed `listRepos` to distinguish "no repos" from "lookup failed"** — deferred; for now the empty-state hint nudges toward both causes (specs + auth). A typed error channel through `onListWorkspaceRepos` would be the clean fix and is a good follow-up.

Re-tested: arm64 `swift build` ✓; CrowCore 180, CrowProvider 33, CrowUI 31, root 122.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None — no blocking bugs found. The diff is well-structured, the new resolution helpers are pure and tested, decoding stays forward/backward compatible, and the second commit (`614fee4`) cleans up the round-1 review feedback.

### Security Review
**Strengths:**
- All shell-outs use `Process.arguments` (not `/bin/sh -c`), so user-entered specs/repo slugs can't be injected as commands — verified for both `gh repo clone`/`list` and `glab api`/`repo clone`.
- `GITLAB_HOST` is passed via an env dict rather than spliced into a command string.
- `WorkspaceInfo.validateName` blocks `/`, `:`, and `\0` in workspace names, so the workspace-name segment of `{devRoot}/{workspace}/{repo}` can't introduce a path separator.
- `JobConfig` decoding remains tolerant: the brief free-form `repo`-only era and legacy `{workspace, repo}` jobs both still decode (`JobConfigTests.jobConfigBackCompatMissingWorkspaceKey`).

**Concerns (low severity):**
- `WorkspaceInfo.validateName` does not reject `.` or `..`. A user with a `..` workspace would resolve `{devRoot}/../{repoFolder}` and escape devRoot. Self-inflicted, but tightening `unsafeCharacters` (or rejecting names that resolve to `.`/`..`) would close the foot-gun cheaply.

### Code Quality

**Yellow — would fix before the next release:**
- `AppDelegate.swift:620` — the workspace-repo cache key is `"\(ws.name)\u{1}\(ws.alwaysInclude.joined(separator: \",\"))\"`. It omits `provider` and `host`, so if a user switches a workspace's provider (or GitLab host) without touching its specs, the picker keeps returning the old provider's slugs for up to 5 min. Adding `ws.provider` and `ws.host ?? \"\"` to the key (and/or invalidating on workspace edit) would fix it.
- `ProviderManager.swift:158-161` — for a nested-group glob (`group/sub/*`), `classifySpec` produces `.glob(owner: \"group/sub\")` and `listRepos` then shells out with `groups/group/sub/projects`. GitLab's group-by-path endpoint expects the group path URL-encoded (`group%2Fsub`); the un-encoded form 404s and `listRepos` silently swallows it back to `[]`. Either URL-encode the owner segment for GitLab (`owner.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed)` and replace `/` → `%2F`) or reject nested-group globs in `classifySpec` so users get a clearer signal than \"no repos found.\"

**Green — consider:**
- `SettingsView.jobScope` renders `\"workspace/repo\"` where `repo` is already `owner/repo`, producing strings like `\"RadiusMethod/radiusmethod/api\"`. Just showing the slug (or `workspace · slug`) would read cleaner.
- `cloneJobRepo` loads the full config from disk on every clone (`ConfigStore.loadConfig(devRoot:)`). Fine in practice (jobs fire on a 30s tick, not per second) but plumbing the workspace through `runJob` would avoid the round-trip and keep behavior consistent if the user is mid-edit.
- The 5-min cache TTL is reasonable, but explicit invalidation when `appConfig.workspaces` mutates (added/removed/edited in Settings) would make the staleness window deterministic instead of time-based.

### Summary Table
| Priority | Issue |
|----------|-------|
| Yellow | Cache key omits provider/host — stale picker after provider switch (AppDelegate.swift:620) |
| Yellow | Nested-group GitLab globs silently 404 (un-encoded slash in `groups/<path>/projects`) |
| Yellow | Workspace name validation allows `.`/`..` — could escape devRoot |
| Green | `jobScope` display is redundant when repo is already `owner/repo` |
| Green | `cloneJobRepo` re-reads config from disk per run |
| Green | Cache invalidation is purely TTL-based |

**Recommendation:** Approve. The PR delivers the workspace + provider-sourced repo picker described, tests the new resolution helpers (`classifySpec`, `reposForSpecs`, `jobRepoFolder`, `jobWorktreeLayout`), and preserves config compatibility across all three iterations of this feature. The yellow items are real but low-impact and good candidates for a follow-up.

---

dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues
None — no blocking bugs or security issues.

### Security Review
**Strengths:**
- `gh` / `glab` invocations pass arguments as separate `Process` argv elements (`SessionService.swift:760-768`, `ProviderManager.swift:140-159`) — no shell interpolation, so user-controlled workspace specs and `owner/repo` slugs can't introduce command injection.
- `GITLAB_HOST` is plumbed per-call via the new `env:` parameter rather than mutated globally, keeping host scoping tight.
- `cloneJobRepo` uses an absolute destination computed via `NSString.appendingPathComponent`, so the clone path is unaffected by the inherited CWD.

**Concerns:**
- None worth blocking on.

### Code Quality

**Yellow — should fix before merge**

- **`cloneJobRepo` doesn't ensure the workspace parent directory exists** (`Sources/Crow/App/SessionService.swift:1431-1454`). If the user creates a workspace in Settings and immediately schedules a job before any checkout lives under `{devRoot}/{workspace}/`, the parent directory won't exist on disk. `gh repo clone owner/repo /abs/missing-parent/api` then fails because git can't create leading directories, the catch block logs and falls through, and `runJob` returns `nil` — so the headline clone-on-demand feature silently no-ops on first run for a brand-new workspace. Add `try? FileManager.default.createDirectory(atPath: (destination as NSString).deletingLastPathComponent, withIntermediateDirectories: true)` just before the `shell(...)` call.

**Green — nits / follow-ups**

- **Cache key omits `provider` and `host`** (`Sources/Crow/App/AppDelegate.swift:621-628`). Key is `"{ws.name}{specs}"`. If a workspace flips GitHub↔GitLab or changes `GITLAB_HOST` without the name/specs changing, the 5-min cache returns stale repos. Include `ws.provider` and `ws.host ?? ""` in the key.
- **No de-dup of in-flight loads.** Two concurrent form opens for the same uncached workspace both shell out to `gh repo list`. Harmless extra call; mention only if you ever expand the provider call surface.
- **`runJob` paths beyond pure math are uncovered.** `JobRepoResolutionTests` correctly tests the extracted `jobRepoFolder` / `jobWorktreeLayout` helpers, but clone-on-demand and the legacy bare-name discover-by-folder branch in `runJob` aren't reachable from tests without DI. Acceptable for now since the risky logic was pulled into the pure helpers.
- **Cosmetic:** the empty-state hint string in `JobFormView.swift:191-193` is long enough that the curly quotes (`"Always Include Repos"`) read slightly oddly inline; consider straight quotes or italics for the field name. Non-blocking.

### Summary Table

| Priority | Issue |
|----------|-------|
| Red | — |
| Yellow | `cloneJobRepo` doesn't `mkdir -p` the workspace parent before invoking `gh`/`glab` clone, so the first job in a freshly-created workspace silently fails. |
| Green | `workspaceRepoCache` key omits provider/host (5-min stale window on workspace edits); no in-flight de-dup of provider calls; clone-on-demand and legacy-fallback paths in `runJob` lack integration coverage. |

**Recommendation:** Request Changes — only because the missing `mkdir -p` regresses the headline "clone on demand" promise the first time a user wires a job into a brand-new workspace. The fix is one line, the surrounding architecture (actor-isolated `ProviderManager`, MainActor-confined cache, generation-token-protected UI loads, forward/back-compatible `JobConfig` decoding with tests for both transitions) is well thought out, and the test coverage of the extracted helpers is good. Once the directory-creation tweak lands this is ready to ship.

---

dgershman:
Addressed in `88bbad4`:

**Yellow (blocking)**
- ✅ **`cloneJobRepo` mkdir -p** — creates the workspace parent (`destination.deletingLastPathComponent`, `withIntermediateDirectories: true`) before the `gh`/`glab` clone, so the first job in a freshly-created workspace clones instead of silently no-opping.

**Also picked up from the prior review's Yellow items**
- ✅ **Cache key omits provider/host** — `workspaceRepoCache` key now includes `ws.provider` and `ws.host ?? ""`, so a provider/host flip without a spec change no longer returns stale slugs within the TTL.
- ✅ **Nested-group GitLab globs 404** — `listRepos` now URL-encodes the group path (`group/sub` → `group%2Fsub`) via `encodeGitLabGroupPath` before `groups/<path>/projects`. New unit tests cover it.
- ✅ **Workspace name `.`/`..`** — `WorkspaceInfo.validateName` now rejects `.` and `..` (a name merely containing a dot is still valid). New test assertions added.

**Green**
- ✅ **`jobScope` redundancy** — now shows the `owner/repo` slug as-is (no more `RadiusMethod/radiusmethod/api`); legacy bare-name jobs keep the workspace prefix.
- ✅ **Cosmetic empty-state hint** — reworded to drop the inline curly-quoted field name.
- ↩️ **`cloneJobRepo` re-reads config per run** — deferred. It fires on the 30s job tick, not per-second; plumbing `WorkspaceInfo` through `runJob` is a reasonable follow-up but out of scope for this fix.
- ↩️ **Explicit cache invalidation on workspace edits** — deferred. Including provider/host in the key already covers the common edits; the only remaining staleness is adding a brand-new spec to an existing workspace within the 5-min window. Good follow-up.
- ↩️ **In-flight load de-dup** — deferred (you flagged it as harmless); a duplicate `gh repo list` on two concurrent first-opens is cheap.
- ↩️ **`runJob` clone/legacy branches lack integration coverage** — the risky logic is in the tested pure helpers (`jobRepoFolder`/`jobWorktreeLayout`); exercising the full `runJob` would need DI on `ProviderManager`/filesystem, deferred.

Re-tested: arm64 `swift build` ✓; CrowCore 186, CrowProvider 35 (+2), CrowUI 31, root 135.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review

**Strengths:**
- All shell-outs (`gh repo clone/list`, `glab repo clone`, `glab api`) pass arguments as separate `Process` argv elements — no shell interpolation, so `job.repo`, `workspace.host`, and `job.workspace` cannot introduce command injection.
- `GITLAB_HOST` is plumbed via the new `env:` dict on `SessionService.shell` (mirroring `ProviderManager.shell`), so the argv-env hack flagged in round 2 is gone.
- `WorkspaceInfo.validateName` now rejects `.` and `..` in addition to `/`, `:`, and `\0`, closing the path-traversal foot-gun called out last round (`AppConfig.swift:208-211`).
- `ProviderManager.encodeGitLabGroupPath` percent-encodes `/` → `%2F` so nested-group globs like `group/sub/*` hit a valid GitLab `groups/group%2Fsub/projects` URL instead of silently 404ing into an empty picker (`ProviderManager.swift:259-262`).
- `JobConfig` decoding stays forward/back-compatible across all three iterations of this feature (legacy `{workspace, repo}`, transient free-form `repo`-only, and the final workspace + slug shape).

**Concerns:**
- None.

### Code Quality

**Round-3 fixes are all in place and correctly scoped:**
- `cloneJobRepo` now `mkdir -p`s the workspace parent (`destination.deletingLastPathComponent`, `withIntermediateDirectories: true`) before invoking `gh`/`glab` clone, so the headline clone-on-demand feature works on first run for a brand-new workspace (`SessionService.swift:1391-1394`).
- `workspaceRepoCache` key now includes `ws.provider` and `ws.host ?? ""`, so flipping a workspace's provider or GitLab host doesn't return stale wrong-provider slugs within the 5-min TTL (`AppDelegate.swift:653-655`).
- `SettingsView.jobScope` renders the `owner/repo` slug as-is (no more `RadiusMethod/radiusmethod/api`) while still prefixing the workspace for legacy bare-name jobs (`SettingsView.swift:402-405`).
- New unit tests pin `encodeGitLabGroupPath` and the `.`/`..` rejection.

**Deferred items (acknowledged in the commit body), all reasonable:**
- `cloneJobRepo` re-reads the full config from disk per run — fine on a 30s job tick; plumbing `WorkspaceInfo` through `runJob` is a good follow-up.
- Explicit cache invalidation on workspace edits — current TTL + provider/host-in-key covers the common cases; the only remaining staleness is adding a brand-new spec within the 5-min window.
- In-flight load de-dup on concurrent first-opens — duplicate `gh repo list` is cheap.
- Integration coverage for the `runJob` clone-on-demand and legacy bare-name branches — the risky path math is covered by the extracted pure helpers (`jobRepoFolder`, `jobWorktreeLayout`); the rest needs DI on `ProviderManager`/filesystem.

### Build & Tests
Couldn't `swift build` the root locally (missing `GhosttyKit.xcframework` binary — local env issue, identical to last review). Package tests run cleanly:

- `Packages/CrowCore swift test` — **186 tests passed** (incl. `.`/`..` validation, `jobConfigBackCompatMissingWorkspaceKey`).
- `Packages/CrowProvider swift test` — **35 tests passed** (incl. `encodeGitLabGroupPathEncodesSeparators`, `encodeGitLabGroupPathLeavesSimpleOwnerUnchanged`, the 7 `classifySpec` cases, and `reposForSpecsKeepsExplicitSlugsSortedAndDeduped`).

### Summary Table

| Priority | Issue |
|----------|-------|
| Red | — |
| Yellow | — |
| Green | Deferred items above remain good follow-ups (config plumbing into `runJob`, explicit cache invalidation, in-flight load de-dup, integration coverage of the clone-on-demand branch). |

**Recommendation:** Approve. Round 3 cleanly addresses every Yellow flagged in the previous round — the `mkdir -p` fix restores the clone-on-demand promise on first use, the cache key correctly invalidates on provider/host flips, GitLab nested-group globs now resolve, and the path-traversal foot-gun on workspace names is closed. The remaining items are explicit, justified follow-ups.

---

PR #332, opened 2026-05-23
feat: auto-rebase watcher for behind/conflicting PR branches (closes #318)

### LINKED ISSUE #318: Auto-rebase + conflict resolution (an option, not label-gated like auto-merge)
## Summary
Add an auto-rebase automation that keeps PR branches up to date and resolves conflicts, modeled on
auto-merge — but driven by a plain option/toggle rather than requiring a special label.

## Background
Auto-merge ("on green") is gated by `autoMergeWatcherEnabled` and the `crow:merge` label, with PR
state polled in `Sources/Crow/App/IssueTracker.swift` (60s timer, consolidated GraphQL) and tracked
via `CrowCore/Models/PRStatus.swift` (`mergeable`, `mergeStateStatus`). Git operations live in
`Packages/CrowGit/.../GitManager.swift` (worktree create/remove only — no rebase today). This mirrors
the existing `autoRespond` watchers (`respondToChangesRequested`, `respondToFailedChecks`).

## Ask
- When a watched PR's branch is behind base or in conflict (`mergeStateStatus` BEHIND/`mergeable ==
  .conflicting`), automatically rebase onto the base branch and push (`--force-with-lease`).
- For non-trivial conflicts, delegate resolution to a Claude session in the worktree rather than
  failing outright.
- Gate the whole behavior on a simple option (global toggle, and ideally per-repo) — **no special
  label required**, unlike `crow:merge`.
- Add `git rebase` support to `GitManager` and reuse the `IssueTracker` polling loop for triggering.

## Acceptance
- With the option enabled, a behind/conflicting PR is rebased onto its base and the branch updated,
  with no label needed.
- Conflicts that can't be resolved mechanically are handed to a Claude session; failures are surfaced.
- With the option disabled, behavior is unchanged.

dgershman:
Closes #318

## What

Adds an opt-in **auto-rebase** automation, modeled on auto-merge but driven by a plain toggle — **no label required**. When a watched Crow-authored PR falls behind its base or develops conflicts, Crow rebases the session's worktree onto the base and force-pushes with `--force-with-lease`. Rebases that hit conflicts are delegated to the session's Claude terminal instead of failing.

## How it works

- **Config** (`AppConfig.autoRebaseWatcherEnabled`, off by default) — surfaced as a new **Auto-rebase** toggle under Settings → Automation.
- **`GitManager.rebaseOntoBase`** (new) returns a `RebaseOutcome`:
  - Refuses to touch a **dirty worktree** (`git status --porcelain`) so it never clobbers a Claude session mid-edit.
  - Verifies HEAD is on the expected branch before rebasing.
  - **Aborts on conflict** to restore a clean tree, then reports `.conflicts` (conflict detected via unmerged paths, not fragile stderr matching).
  - Force-pushes with `--force-with-lease`; disables commit signing during the rebase so an unattended rewrite can't hang on a pinentry / SSH-agent prompt.
- **`IssueTracker`** reuses the existing 60s poll loop, the Crow-authorship gate (`prHasCrowAuthoredCommit`), and per-head loop-safety. `shouldAttemptAutoRebase` accepts OPEN/non-draft PRs that are `BEHIND` or `CONFLICTING` — independent of label or review state.
- **Precedence:** when auto-merge is also enabled and the PR is a `crow:merge` BEHIND merge candidate, auto-rebase yields to auto-merge's existing `gh pr update-branch` so the two watchers don't fight. Auto-rebase still owns every CONFLICTING PR and every non-labeled BEHIND PR.
- **Conflict delegation** reuses the existing `fixConflicts` quick action (rebase + resolve + force-push prompt) injected into the session's managed terminal; a notification fires regardless.

## Acceptance criteria

- ✅ With the option enabled, a behind/conflicting PR is rebased onto base and the branch updated, no label needed.
- ✅ Conflicts that can't be resolved mechanically are handed to a Claude session; failures are surfaced (log + notification).
- ✅ With the option disabled, behavior is unchanged (toggle defaults off; watcher inert).

## Tests

- `IssueTrackerAutoRebaseTests` — candidacy predicate (accepts BEHIND/CONFLICTING with or without labels and regardless of review state; rejects clean/draft/closed/merged).
- `RebaseTests` (CrowGit) — integration tests against throwaway repos: clean rebase → pushed, conflict → aborted + clean tree, dirty worktree → skipped, branch mismatch → failed.
- Full suites green: CrowGit 17, CrowCore 169, CrowTests 123.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -30,6 +30,12 @@ public struct AppConfig: Codable, Sendable, Equatable {
     /// issue. While disabled, the label is left alone so a later opt-in
     /// can still pick up previously-labeled issues.
     public var autoCreateWatcherEnabled: Bool
+    /// When true, the IssueTracker rebases watched Crow-authored PR branches
+    /// onto their base and force-pushes (`--force-with-lease`) whenever they
+    /// fall BEHIND base or become CONFLICTING — no label required (unlike
+    /// `autoMergeWatcherEnabled`). Rebases that hit conflicts are handed to
+    /// the session's Claude terminal. Opt-in: defaults to false (CROW-318).
+    public var autoRebaseWatcherEnabled: Bool
     public var cleanup: CleanupConfig
     /// Scheduled jobs: named sets of prompts that fire automatically on a
     /// schedule, scoped to a repo. Driven by `JobScheduler` (CROW-317).
@@ -47,6 +53,7 @@ public struct AppConfig: Codable, Sendable, Equatable {
         attributionTrailers: Bool = true,
         autoMergeWatcherEnabled: Bool = false,
         autoCreateWatcherEnabled: Bool = false,
+        autoRebaseWatcherEnabled: Bool = false,
         cleanup: CleanupConfig = CleanupConfig(),
         jobs: [JobConfig] = []
     ) {
@@ -61,6 +68,7 @@ public struct AppConfig: Codable, Sendable, Equatable {
         self.attributionTrailers = attributionTrailers
         self.autoMergeWatcherEnabled = autoMergeWatcherEnabled
         self.autoCreateWatcherEnabled = autoCreateWatcherEnabled
+        self.autoRebaseWatcherEnabled = autoRebaseWatcherEnabled
         self.cleanup = cleanup
         self.jobs = jobs
     }
@@ -78,12 +86,13 @@ public struct AppConfig: Codable, Sendable, Equatable {
         attributionTrailers = try container.decodeIfPresent(Bool.self, forKey: .attributionTrailers) ?? true
         autoMergeWatcherEnabled = try container.decodeIfPresent(Bool.self, forKey: .autoMergeWatcherEnabled) ?? false
         autoCreateWatcherEnabled = try container.decodeIfPresent(Bool.self, forKey: .autoCreateWatcherEnabled) ?? false
+        autoRebaseWatcherEnabled = try container.decodeIfPresent(Bool.self, forKey: .autoRebaseWatcherEnabled) ?? false
         cleanup = try container.decodeIfPresent(CleanupConfig.self, forKey: .cleanup) ?? CleanupConfig()
         jobs = try container.decodeIfPresent([JobConfig].self, forKey: .jobs) ?? []
     }
 
     private enum CodingKeys: String, CodingKey {
-        case workspaces, defaults, notifications, sidebar, remoteControlEnabled, managerAutoPermissionMode, telemetry, autoRespond, attributionTrailers, autoMergeWatcherEnabled, autoCreateWatcherEnabled, cleanup, jobs
+        case workspaces, defaults, notifications, sidebar, remoteControlEnabled, managerAutoPermissionMode, telemetry, autoRespond, attributionTrailers, autoMergeWatcherEnabled, autoCreateWatcherEnabled, autoRebaseWatcherEnabled, cleanup, jobs
     }
 }
 
--- a/Packages/CrowGit/Sources/CrowGit/GitManager.swift
+++ b/Packages/CrowGit/Sources/CrowGit/GitManager.swift
@@ -3,14 +3,20 @@ import CrowCore
 
 /// Discovers git repos across workspaces and manages worktrees.
 public actor GitManager {
-    private let config: WorkspaceConfig
+    /// Workspace config is only needed by `discoverRepos`. Operations that
+    /// act on an explicit path (worktree create/remove, `rebaseOntoBase`)
+    /// don't use it, so it's optional and the no-arg `init()` lets callers
+    /// that only rebase (e.g. IssueTracker) construct one without plumbing
+    /// `WorkspaceConfig` through.
+    private let config: WorkspaceConfig?
 
-    public init(config: WorkspaceConfig) {
+    public init(config: WorkspaceConfig? = nil) {
         self.config = config
     }
 
     /// Discover all git repositories under devRoot.
     public func discoverRepos() async throws -> [RepoInfo] {
+        guard let config else { return [] }
         let devRoot = config.devRoot
         let fm = FileManager.default
         guard let workspaceDirs = try? fm.contentsOfDirectory(atPath: devRoot) else {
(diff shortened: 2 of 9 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Solid, well-scoped feature with thoughtful safety guards (`--force-with-lease`, dirty-tree refusal, branch verification, abort-on-conflict, gpgsign-disabled, per-head dedup, precedence with auto-merge). Tests cover both the candidacy predicate and the rebase mechanism against throwaway repos. A few concrete issues below — none catastrophic, but #1 is a real silent-failure mode.

### Yellow — Should Fix

**1. `.failed` outcomes are never retried until the head commit changes** — `Sources/Crow/App/IssueTracker.swift:2092-2094`

`attemptRebase` only re-arms `autoRebaseAttempted` on `.dirtyWorktree`. For transient failures — `git fetch` flaking, `--force-with-lease` rejected by a concurrent push, base ref temporarily unreachable — the per-head key stays set, so the watcher won't retry until `pr.headRefOid` changes. For a `BEHIND` PR with no further commits, that's effectively never. The watcher silently stops working for that PR.

Suggest also removing the key on `.failed`, optionally with a small retry cap (e.g., max 3 attempts per head-key) to avoid hot-looping on a genuinely-broken config.

```swift
case .failed(let msg):
    autoRebaseAttempted.remove(headKey)  // allow next poll to retry
    NSLog("[Crow] Auto-rebase failed on %@: %@", pr.url as NSString, msg as NSString)
```

**2. Worktree existence check happens AFTER the `gh api` Crow-authorship call** — `IssueTracker.swift:2056-2069`

For a session that has been completed/archived (worktree removed) but still has an open `.pr` link, every poll for every head state will spend one `gh api .../pulls/N/commits` call before discovering there's no worktree to rebase into. Flip the order: cheap local `worktrees(for:)`/`isMainRepoCheckout`/branch check first, then `prHasCrowAuthoredCommit`. Also, `applyAutoRebase` (line 2022) iterates all non-manager sessions including completed/archived — unlike auto-merge there's no label gate to limit candidates, so this matters more here.

**3. Force-push silently rewrites *committed-but-unpushed* local commits** — `GitManager.swift:120-124`

The dirty-tree guard catches uncommitted changes, but a Claude session that has *committed* (e.g., mid-`/clear` workflow) and not yet pushed has a clean `git status --porcelain`. `rebaseOntoBase` will rewrite those local commits onto base and force-push them, publishing work the author may have intended to amend or squash first. Not data loss, but surprising.

A `git rev-list --count origin/<branch>..HEAD` (or `--left-right --count`) check, returning `.dirtyWorktree` (or a new `.unpushedCommits`) when the local branch is ahead of the remote, would close this. Lower priority than #1, but worth considering.

### Green — Nits / Recommendations

- **NSLog ternary type mismatch** — `IssueTracker.swift:2078-2080`. `pr.mergeable == "CONFLICTING" ? "CONFLICTING" : "BEHIND" as NSString` — operator precedence binds the cast to the false branch only, so the ternary has mixed `String`/`NSString` branches. Compiles via `CVarArg`, but the intent is clearly `(... ? "CONFLICTING" : "BEHIND") as NSString`.
- **Unused `prURL` parameter** — `NotificationManager.swift:167, 184`. Neither `notifyAutoRebasePushed` nor `notifyAutoRebaseConflicts` use `prURL` in the body — it's only there to display `number`. Drop or include it.
- **Conflict notification fires regardless of terminal availability** — `AppDelegate.swift:538-546`. Good call: `dispatchManual` silently no-ops without a managed terminal, and the notification ensures the user still sees the conflict.
- **Test coverage gap** — Candidacy predicate and the git mechanism are well covered, but the `applyAutoRebase` orchestration (precedence with auto-merge, in-flight dedup, per-head re-arm on dirty) isn't exercised. Not blocking, just noting.

### Security Review

**Strengths**
- `--force-with-lease` (against last-fetched remote head) instead of `--force` — protects against concurrent pushes.
- `commit.gpgsign=false` scoped to the single rebase invocation, with a clear rationale (unattended op can't hang on pinentry).
- All git args passed as `[String]` arrays — no shell injection surface.
- `worktreePath`, `branch`, `baseBranch` come from internal session/PR records, not user input.
- Crow-authorship trailer (`Crow-Session: <uuid>` matching a known session) gates every rebase — same trust model as auto-merge.
- Opt-in toggle, defaults off.

**Concerns**
- None that block.

### Summary Table

| Priority | Issue |
|----------|-------|
| Yellow   | `.failed` rebases never retry until head changes — silent watcher stall (IssueTracker.swift:2092) |
| Yellow   | `gh api` Crow-authorship call runs before cheap worktree-existence check (IssueTracker.swift:2056) |
| Yellow   | Force-push publishes committed-but-unpushed local commits (GitManager.swift:120) |
| Green    | NSLog ternary `as NSString` binds only to false branch (IssueTracker.swift:2080) |
| Green    | Unused `prURL` param on rebase notifications (NotificationManager.swift:167,184) |
| Green    | No orchestration tests for `applyAutoRebase` (precedence, dedup, re-arm) |

**Recommendation:** Request Changes. #1 is a real failure mode that silently disables the watcher on the next transient `git fetch`/push failure — please address before merge. #2 and #3 are reasonable to address in the same revision. Greens are optional.

---

dgershman:
Thanks for the thorough review — all three Yellows and both code Greens are addressed in 689518c.

**Yellow #1 — `.failed` never retried (silent stall).** `attemptRebase` now re-arms the per-head key on `.failed` with a bounded cap (`maxAutoRebaseFailureRetries = 3`) tracked in `autoRebaseFailureCounts`, keyed per head state. Transient fetch/lease/base failures retry on the next poll; a genuinely broken config gives up after 3 attempts (logged) instead of hot-looping. The cap decision is a pure `shouldRetryFailedRebase(failureCount:)` helper with unit tests. The count resets on any non-failure outcome.

**Yellow #2 — `gh api` before worktree check.** Reordered: the local `worktrees(for:)` / `isMainRepoCheckout` / branch-match guards now run *before* `prHasCrowAuthoredCommit`, so a completed/archived session with a dangling `.pr` link and no worktree no longer spends an API call per poll.

**Yellow #3 — force-push of committed-but-unpushed commits.** `rebaseOntoBase` now fetches `origin/<branch>` alongside the base and requires `HEAD == origin/<branch>` before rewriting. If they differ — local ahead (unpushed work) *or* behind/diverged (stale local that would revert remote) — it returns the new `.outOfSyncWithRemote` outcome and touches nothing. The caller treats it as transient (re-arm + retry). Fetching the branch up front also makes the `--force-with-lease` baseline accurate. New `unpushedLocalCommitsAreSkipped` test covers it.

**Greens.** Fixed the NSLog ternary by computing `priorState` before the call (no more `as NSString` binding only to the false branch). Dropped the unused `prURL` param from `notifyAutoRebasePushed`/`notifyAutoRebaseConflicts` and their callers.

On the orchestration-test gap: I extracted the retry policy to a pure helper so it's now covered, but kept `applyAutoRebase`'s in-flight/precedence/dedup logic untested to match the existing auto-merge pattern (its `applyAutoMerge` is likewise only covered via the pure predicates). Happy to add a harness-based orchestration test if you'd like it in this PR.

Suites green: CrowGit 18, CrowCore 169, CrowTests 125.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review

**Strengths:**
- **No shell injection**: `rebaseOntoBase` builds git invocations as argv arrays via the `run([...])` helper, never `sh -c`. The new code path doesn't introduce string interpolation into a shell.
- **`--force-with-lease` is correctly baselined**: the explicit `git fetch origin <base> <branch>` immediately before the rebase sets the remote-tracking ref that lease compares against, so a concurrent push from the Claude session loses the lease rather than getting silently overwritten (`GitManager.swift:140`).
- **Crow-authorship gate**: `prHasCrowAuthoredCommit` is enforced before any rewrite — Crow will not force-push branches that don't carry a `Crow-Session` trailer matching a known local session.
- **Defense-in-depth against protected branches**: `!primary.isMainRepoCheckout` in `attemptRebase` (IssueTracker.swift:2080-2086) bails out for `main`/`master`/`develop`/`trunk` etc. via `Worktree.protectedBranchNames`, so the watcher can't rewrite a default branch even if data is malformed.
- **In-sync precondition**: refusing to rewrite when `HEAD != origin/<branch>` is the right call — it closes off both "publishes committed-but-unpushed work" and "reverts remote commits" footguns. Good catch in the fix commit.
- **`commit.gpgsign=false` only on the rebase**: keeps unattended rewrites from hanging on pinentry/SSH-agent but doesn't affect new commits the user creates. Push doesn't sign, so this is scoped correctly.
- **Conflict detection via unmerged paths**: `git diff --name-only --diff-filter=U` instead of stderr text matching survives across git versions.

**Concerns:** None material. Worktree paths are validated against `devRoot` at write time (`AppDelegate.isPathWithinDevRoot`); this PR consumes already-validated paths.

### Code Quality

**Strengths:**
- Pure, `nonisolated` helpers (`shouldAttemptAutoRebase`, `shouldRetryFailedRebase`) are easy to unit-test and the test file exercises every branch.
- `RebaseOutcome` as a discriminated enum (instead of throwing) forces the caller to handle each case explicitly — `attemptRebase`'s switch is exhaustive.
- Precedence with auto-merge (`IssueTracker.swift:2045-2048`) is correctly carved out: auto-rebase yields to `gh pr update-branch` only for `crow:merge` BEHIND merge candidates; it still owns CONFLICTING and non-labeled BEHIND PRs.
- Cheap local worktree/branch checks done before the `gh api` Crow-authorship call (addressed in the fix commit) — no API call burned when there's no usable worktree.
- Bounded `.failed` retry policy (`maxAutoRebaseFailureRetries = 3`) prevents stalling on transient flakes without hot-looping on a broken config.
- Integration tests in `RebaseTests.swift` against throwaway repos cover clean / conflict / dirty / out-of-sync / branch-mismatch — exactly the right surface to lock down.
- Verified locally: CrowGit (18) and CrowCore (169) test suites pass.

- `autoRebaseAttempted` and `autoRebaseFailureCounts` grow unboundedly across the app lifetime — entries for stale head commits/gave-up failures never get purged. For a long-running app that's a small leak; could opportunistically prune to the set of currently-open PR URLs on each refresh, but not worth holding the PR.
- `prURL` is now unused in the `onAutoRebasePushed` / `onAutoRebaseConflicts` callback signatures (the fix commit drops it on the consuming side via `_`). Tightening the closure type to drop the parameter entirely would prevent future drift, but it's stylistic.
- `applyAutoRebase` has no first-fetch suppression — on app launch, a still-BEHIND PR will rebase immediately. This matches auto-merge's behavior and is fine given the opt-in default-off + Crow-authorship gate; one line of comment would make the intent explicit.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | None |
| Yellow | None |
| Green | Unbounded growth of `autoRebaseAttempted` / `autoRebaseFailureCounts` (cosmetic memory) |
| Green | Drop unused `prURL` from callback signatures, or document why it's kept |
| Green | Note first-fetch behavior in `applyAutoRebase` comment |

**Recommendation:** **Approve.** The implementation is conservative and correct — every dangerous operation is gated by an explicit precondition (dirty tree, branch match, in-sync remote, Crow authorship, protected branch), conflicts are handled cleanly by abort + delegate, and the prior review's three yellow items were all addressed in the fix commit. The remaining observations are cosmetic.

---

PR #333, opened 2026-05-23
fix: keep UI responsive when concurrent PR reviews arrive (#304)

### LINKED ISSUE #304: UI freezes when multiple PR reviews arrive concurrently
## Summary

When several inbound PR reviews land at roughly the same time, the Crow UI locks up and stops responding to input. Recovery requires waiting it out or restarting the app.

## Repro

1. Have multiple sessions in \`inReview\` state with open PRs.
2. Reviewers approve / leave comments on those PRs in close succession (or a webhook burst arrives — e.g. CI finishing on several PRs at once).
3. Observe the Crow window: clicks/keystrokes stop registering, session list does not scroll, terminal does not echo.

## Expected

UI stays responsive; review/state updates apply incrementally without blocking the main thread.

## Likely area

Probably review-ingest / status-refresh work running on the main actor — concurrent inbound updates serialize behind \`@MainActor\` AppState mutations and starve the UI. Worth checking:

- Whether GitHub review-event handlers do JSON parsing or network calls inside \`MainActor.run { ... }\`.
- Whether each inbound event triggers a full session-list re-render rather than a diff.
- Whether the JSONStore lock (\`NSLock\`) is held across long writes when many sessions update at once.

## Asks

- Capture a sample / spindump from a reproducer to confirm where the main thread is stuck.
- Move parsing and network IO off the main actor; only the final state mutation should hop back.
- Coalesce rapid review-state updates so N concurrent events don't trigger N full re-renders.

dgershman:
Closes #304

## Problem

When several inbound PR reviews / status changes landed at roughly the same time, the Crow UI froze and stopped responding to input. The cause is `IssueTracker.refresh()` — it runs on `@MainActor` and performed CPU- and IO-heavy work on the main thread.

## Fixes

1. **Parsing off the main actor.** The consolidated GitHub-response parse helpers (`parseConsolidatedResponse` + the issue/PR/review/rate-limit parsers) are now `nonisolated` and run via `Task.detached`; `ConsolidatedGitHubResponse` is `Sendable`, so only the result hops back to the main actor. Parsing ~50 PRs + ~50 review requests + ~150 issues (with nested label/check/review traversal) no longer blocks the UI thread.

2. **Batched store writes.** `applySessionPRLinks` and `applyReconciledPRLinks` accumulated new links and wrote them one-per-session inside the loop — each call was a full-store JSON encode + atomic disk write. They now accumulate and persist with a single `JSONStore.mutate`, so a burst of N newly-linked review PRs is **1** write instead of N.

3. **JSONStore lock scope.** `mutate()` now applies the transform and snapshots under the data lock, then writes to disk **outside** it. Writes are serialized by a separate write lock keyed on a monotonic sequence, so disk order matches mutation order and stale snapshots are coalesced (never dropping data). Readers and other mutators no longer block for the full encode + atomic-write duration.

These keep writes synchronous and ordered, preserving the read-after-write semantics the fresh-per-call `JSONStore()` pattern depends on.

## Testing

- `CrowPersistence`: all 26 tests pass, including two new ones — cross-field concurrent consistency and last-write-wins coalescing — plus the existing concurrency guard.
- Root `Crow` package builds and links cleanly (Swift 6, arm64), and all 117 root tests across 13 suites pass (including the 5 `IssueTracker*` suites).

### DIFF
--- a/Packages/CrowPersistence/Sources/CrowPersistence/JSONStore.swift
+++ b/Packages/CrowPersistence/Sources/CrowPersistence/JSONStore.swift
@@ -37,6 +37,15 @@ public final class JSONStore: Sendable {
     private let fileURL: URL
     private let lock = NSLock()
     private nonisolated(unsafe) var _data: StoreData
+    /// Monotonic mutation counter, bumped under `lock` for every `mutate`.
+    /// Each snapshot carries the sequence it was taken at so the write path
+    /// can drop stale snapshots (see `mutate`).
+    private nonisolated(unsafe) var writeSeq: UInt64 = 0
+    /// Serializes disk writes independently of `lock`, so the in-memory data
+    /// lock is never held across the (potentially slow) encode + atomic write.
+    private let writeLock = NSLock()
+    /// Highest sequence already persisted, guarded by `writeLock`.
+    private nonisolated(unsafe) var lastWrittenSeq: UInt64 = 0
 
     public var data: StoreData {
         lock.lock()
@@ -70,10 +79,28 @@ public final class JSONStore: Sendable {
     }
 
     public func mutate(_ transform: (inout StoreData) -> Void) {
+        // Apply the mutation and snapshot under `lock`, then release it before
+        // touching disk. Holding `lock` across the encode + atomic write blocks
+        // every reader (`data`) and other mutators for the full duration of the
+        // I/O — exactly the contention that froze the UI when many sessions
+        // updated at once (#304).
         lock.lock()
-        defer { lock.unlock() }
         transform(&_data)
-        Self.save(_data, to: fileURL)
+        writeSeq &+= 1
+        let mySeq = writeSeq
+        let snapshot = _data
+        lock.unlock()
+
+        // Serialize writes on a separate lock so disk order matches mutation
+        // order. Each save carries its sequence; a snapshot that is already
+        // stale (a newer one has been written) is dropped. Because the
+        // highest-sequence snapshot reflects every mutation up to that point,
+        // coalescing redundant writes can never drop data.
+        writeLock.lock()
+        defer { writeLock.unlock() }
+        guard mySeq > lastWrittenSeq else { return }
+        lastWrittenSeq = mySeq
+        Self.save(snapshot, to: fileURL)
     }
 
     private static func save(_ data: StoreData, to url: URL) {
--- a/Packages/CrowPersistence/Tests/CrowPersistenceTests/JSONStoreTests.swift
+++ b/Packages/CrowPersistence/Tests/CrowPersistenceTests/JSONStoreTests.swift
@@ -59,6 +59,63 @@ import Testing
     #expect(reloaded.data.sessions.count == iterations)
 }
 
+// Guards the lock-scope + write-coalescing change (#304): the in-memory data
+// lock is released before the disk write, but writes must still land in
+// mutation order and the newest snapshot must win. A stale snapshot winning
+// would leave the on-disk state internally inconsistent across fields.
+@Test func concurrentMutatesAcrossFieldsAreConsistent() throws {
+    let dir = FileManager.default.temporaryDirectory.appendingPathComponent(UUID().uuidString)
+    defer { try? FileManager.default.removeItem(at: dir) }
+
+    let store = JSONStore(directory: dir)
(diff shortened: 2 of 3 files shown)


dhilgaertner (approved):
## Code & Security Review

### Summary
Three coordinated changes to keep the UI responsive when many GitHub PR reviews/status changes land at once:
1. JSON parsing moved off `@MainActor` via `Task.detached` (parsers marked `nonisolated`, `ConsolidatedGitHubResponse: Sendable`).
2. `applySessionPRLinks` / `applyReconciledPRLinks` accumulate links and persist with a single `JSONStore.mutate` (was N writes per burst).
3. `JSONStore.mutate` releases the data lock before disk I/O, with a sequence-number coalescing scheme on a separate `writeLock` to keep on-disk order matching mutation order.

### Correctness

**`JSONStore.mutate` coalescing** — verified sound:
- `writeSeq` is bumped under `lock`; snapshot is taken under `lock`. Therefore the snapshot at sequence N reflects every mutation up to and including N.
- Coalescing skip (`mySeq > lastWrittenSeq` guard) only drops a snapshot when a strictly higher-sequence snapshot has already been persisted — and that higher-sequence snapshot necessarily includes the caller's mutation. No data loss is possible.
- The post-condition that callers rely on ("after `mutate()` returns, the mutation is on disk, possibly subsumed by a later snapshot") is preserved — important because `applySessionPRLinks`/`applyReconciledPRLinks`/`AppDelegate` all use the fresh-per-call `JSONStore()` pattern.
- `writeSeq &+= 1` wrapping is fine in practice (2^64 mutations).

**Parsers off the main actor** — verified safe:
- `parseConsolidatedResponse`, `parseIssueNodes`, `parseViewerPRs`, `parseReviewRequests`, `parseRateLimit` only touch their arguments and a locally-allocated `ISO8601DateFormatter` (thread-safe per Apple docs since macOS 10.12).
- Return values are all `Sendable` value types.
- `Task.detached { self.parseConsolidatedResponse(output) }` captures `self` (a `@MainActor final class`, implicitly Sendable) and calls a `nonisolated` method — no actor hop required for the work.

**Batched link writes** — verified correct:
- In-memory `appState.links[...]` is still updated per-link inside the loop (so subsequent loop iterations see the new link via `existing.contains`), then `newLinks` is persisted once after the loop. Same observable behavior, one disk write instead of N.

### Security Review
**Strengths:**
- File permissions (0o600, owner-only) still applied after every write.
- Atomic file write semantics preserved — no partial file ever visible to a fresh `JSONStore()`.
- No new untrusted input handling or shell-injection surface.
- `nonisolated` parsers still treat all GraphQL fields as optional and skip malformed nodes — no new crash surface from off-actor parsing.

**Concerns:** None.

### Code Quality
- New behavior is well-tested: `concurrentMutatesAcrossFieldsAreConsistent` guards against torn cross-field snapshots, and `lastMutationInBurstIsPersisted` guards against coalescing dropping the latest. Together with the existing `concurrentMutatesAreConsistent`, these pin down the contract.
- Inline comments are excellent — each non-obvious block points back to #304 and explains the *why*.
- Local verification: `swift test` in `Packages/CrowPersistence` passes all 26 tests (including the 2 new ones); root build couldn't be verified locally due to an unrelated missing `GhosttyKit.xcframework` binary artifact in this checkout, but the PR description reports the root suite (117 tests / 13 suites) passes.

### Summary Table
| Priority | Issue |
|----------|-------|
| Green | The class-level docstring on `JSONStore` (`Packages/CrowPersistence/Sources/CrowPersistence/JSONStore.swift:24-29`) still says "Uses `NSLock` to serialize access to the in-memory `StoreData` and disk writes" — slightly out of date now that data access and disk writes use *different* locks. The per-property doc comments cover it, so this is a minor consistency nit. |
| Green | `mutate()` is still synchronous on the calling thread for the disk write itself, so a single MainActor call still pays for one encode + atomic write. The PR intentionally preserves this to keep the sync API contract; combined with the per-burst batching the actual stall drops from N writes to 1, which matches the stated goal. Worth a follow-up if a single write ever becomes the dominant cost. |

**Recommendation:** Approve. The locking design is correct, the off-actor parsing is sound (parsers are genuinely pure), the batching is a clean win, and the new tests pin down the subtle invariants. Minor doc-only nits are non-blocking.

---

PR #336, opened 2026-05-23
fixes for tmux refactor

### LINKED ISSUE #338: Fixes for TMUX refactor

dgershman:


dhilgaertner (approved):
## Code & Security Review

### Summary

Two bundled changes:

1. **tmux cockpit race fix** (`TmuxBackend.swift`, `TerminalSurfaceView.swift`)
2. **Removal of the standalone-terminals feature** (replaced by multi-Manager sessions from #326)

Both are well-reasoned, the comments explaining *why* in the diff are unusually good, and the cleanup of the removed feature is thorough.

### Security Review

**Strengths:**
- No new external input, file I/O, or network calls.
- No credential / secret handling changes.
- `notConfigured` check still gates execution before any `Process` is spawned (`TmuxBackend.swift:371-377`).
- Sandbox / process-spawn surface is unchanged; the new code paths are pure control-flow inside the existing tmux subprocess wrapper.

**Concerns:**
- None.

### Code Quality

**Race fix in `ensureRunningServer` (`TmuxBackend.swift:387-419`)**
The double-check pattern is correct:
- First `hasSession()` (line 398) avoids the most common race (lost the creation race before the subprocess even started).
- The catch block then catches the *narrower* race (lost it after the subprocess began) and adopts the existing session.
- If `hasSession()` is still false in the catch, the original error correctly propagates — so genuine misconfigurations (bad conf path, tmux gone, etc.) are not swallowed.

`TmuxController` is a struct (`TmuxController.swift:19`), so multiple racing callers each constructing their own `ctrl` and assigning it to `controller` is benign — they all describe the same `crow-cockpit` session at the same socket.

**SwiftUI deferral (`TerminalSurfaceView.swift`)**
The comment at lines 33–42 captures a real and easy-to-miss live-lock (run-loop-pumping `Process.waitUntilExit()` re-entering the initial `window.contentView = hostingView` layout). Moving everything — surface lookup, re-parent, `makeActive`, first-responder — onto the next main-queue turn is the right shape of fix.

One minor behavioral shift worth flagging: the previous code used `Task { @MainActor [weak surface] in ... }` with explicit guards (`surface.superview === nsView`, `surface.window === window`) so stale closures from rapid tab switches wouldn't redo work against a different container. The new `DispatchQueue.main.async` block has no such "still relevant?" guard — each rapid `updateNSView` enqueues another full sync. The blocks are idempotent (the `superview !==` check no-ops the re-parent) so this is not a correctness bug, but on a flurry of tab switches you'll issue N redundant `tmux select-window` subprocesses. Probably fine in practice; mentioning in case it shows up later.

**Removal of standalone terminals**
- `AppState.globalTerminalSessionID` is correctly kept as a legacy constant only for purge identification — well-commented (`AppState.swift:82-85`).
- The purge in `SessionService.swift:130-138` is gated on `data.terminals.contains(...)` so we don't write to disk on every launch — nice.
- The purge runs *before* the rehydrate loop, so we don't briefly register tmux windows for rows we're about to drop.
- All wires removed cleanly: `onAddGlobalTerminal` / `onCloseGlobalTerminal` from `AppState` and `AppDelegate`, the early-return in the `terminals` RPC handler (`AppDelegate.swift`), the protected-ID list in `IssueTracker.swift`, the matching test entry, and the dedicated `GlobalTerminalView.swift` file. `grep -r GlobalTerminal` / `onAddGlobalTerminal` / `onCloseGlobalTerminal` returns no stragglers.
- `ReviewTerminalsSidebarRow` → `ReviewsAllowListRow` rename and the `ManagerAllowListRow` → `ManagerSidebarRow` reshuffle are mechanical and consistent with the navigation change.

**Tests**
- The new race-handling branch (catch + adopt) has no direct unit test. The path is hard to exercise without injecting a fake `TmuxController`, so this is understandable — but if a future refactor regresses the catch into a `throw`, the build will still pass. Worth a follow-up if the controller ever grows a test seam.

### Summary Table
| Priority | Issue |
|----------|-------|
| 🟢 | New `ensureRunningServer` catch-and-adopt branch has no automated test (hard to seam today). |
| 🟢 | `DispatchQueue.main.async` in `syncSurface` no longer guards "still the active container"; rapid tab switches may enqueue redundant `select-window` calls. Idempotent — informational. |
| 🟢 | Two semi-independent changes ship in one PR (race fix + feature removal). Stylistic. |

**Recommendation:** Approve. The race-condition fix is correct and the diagnosis (#293 hydration × #326 multi-Manager interaction) is convincing. The standalone-terminals removal is thorough with no leftover references. No security or correctness concerns.

---

dhilgaertner (approved):
## Code & Security Review

Reviewed both commits on the branch:
- `ce53442` — fixes for tmux refactor (live-lock + race condition)
- `013b07a` — remove standalone terminals feature

### Critical Issues
None — the changes address concrete production bugs (window-never-shows live-lock; duplicate-session race with multi-Manager hydration) and the rationale is captured well in the code comments.

### Security Review
**Strengths:**
- No new attack surface. The race-fix path in `ensureRunningServer` only reads `hasSession()` and either adopts or re-throws — no privilege change, no new shell invocation.
- `shellQuote` use around `tmuxBinary`/`socketPath`/`sessionName` for the attach command is unchanged.
- Sentinel / wrapper-log temp file handling unchanged (still uses `$TMPDIR`).

**Concerns:**
- The startup purge in `SessionService.swift:134-138` silently deletes any persisted `globalTerminalSessionID` terminal rows on first launch with this build. This is one-way data loss for users with long-running standalone terminals from prior versions. Acceptable for the product stage, but worth calling out in release notes.

### Code Quality

**Tmux refactor fixes** (`TerminalSurfaceView.swift`, `TmuxBackend.swift`)

- `TerminalSurfaceView.syncSurface` deferring all surface work to the next run-loop turn is the right call — the comment at L33–42 nails the root cause (run-loop pumping inside SwiftUI layout). Good fix.
- `ensureRunningServer` race fix is correct: the pre-check at `TmuxBackend.swift:398` short-circuits the common reentrant case, and the catch-block fallback at L416 covers the residual TOCTOU window where another caller wins the `new-session` race. Both branches only adopt when `hasSession()` is true, so non-race failures (missing config, bad binary) still propagate.
- **Yellow — `existingCockpitSurface` is now dead code.** The public getter at `TmuxBackend.swift:354` was the only thing the old `updateNSView` consulted; after this PR nothing references it (`grep` confirms zero callers). Suggest deleting along with its doc comment to keep the API surface honest.
- **Yellow — `makeActive` is now fired from both `makeNSView` and `updateNSView`.** The comment removed from the old `makeNSView` (`// makeActive is fired from updateNSView — issuing it here too can double-fire tmux select-window…`) was load-bearing: `syncSurface` runs from both lifecycle methods and unconditionally calls `try? TmuxBackend.shared.makeActive(id:)`. On `.id()` recreation SwiftUI will invoke make then update back-to-back, so we will now shell out to `tmux select-window` twice for the same activation. Not visually broken (idempotent), but it doubles the subprocess count on every tab switch and undoes the prior optimization. Consider coalescing by either gating `makeActive` on a change-tracker, or only calling it from `updateNSView`.
- The `DispatchQueue.main.async` closure captures `container` strongly. If SwiftUI tears the container down before the next runloop tick, the closure still runs and re-parents the shared surface into a doomed view — `container.window` will be `nil`, so `makeFirstResponder` is skipped, but the surface ends up momentarily attached to an orphan. In practice the next live `updateNSView` repairs this, so minor.
- Making `attach` and `cockpitSurface` `static` is a sensible mechanical refactor — neither touched instance state.
- No tests cover the new `ensureRunningServer` race path. The fix is hard to drive deterministically without simulating concurrent reentrancy, but a regression test that calls `ensureRunningServer()` twice on a backend with `controller == nil` after `crow-cockpit` already exists at the socket would lock in the adoption behavior. Optional, not blocking.

**Standalone terminals removal** (`AppState.swift`, `MainContentView.swift`, `ReviewBoardView.swift`, `SessionListView.swift`, `AppDelegate.swift`, `IssueTracker.swift`, `SessionService.swift`, `IssueTrackerCompletionTests.swift`, `GlobalTerminalView.swift`)

- Clean, complete removal — `grep` finds no leftover references to `GlobalTerminalView` / `onAddGlobalTerminal` / `onCloseGlobalTerminal`.
- Keeping `AppState.globalTerminalSessionID` purely for purge-on-load is the right tradeoff — comment at `AppState.swift:80-82` explains why. Same rationale should arguably go on the field doc in `IssueTracker.swift` (the protected-IDs list dropped it without comment).
- **Yellow — purged terminals do not get a `TmuxBackend.shared.destroyTerminal(id:)` call.** `SessionService.swift:134-138` removes the rows from disk but does not destroy the corresponding tmux windows. In practice the cockpit session is per-app-launch and stale windows would be reaped by `TmuxOrphanReaper`, so the windows don't truly leak — but if the rows are pruned post-`ensureRunningServer`, the cleanup is one line away and worth doing for symmetry with the rest of the lifecycle.
- The `ReviewBoardView` → `ReviewsAllowListRow` and `SessionListView` → `ManagerSidebarRow` renames are accurate to the new layout. Allow List is now reachable via the Reviews row, Manager row keeps the "+" affordance.
- `IssueTrackerCompletionTests` correctly drops the global-terminal entry from `protectedIDs`; no other test assertions reference it.
- No security-sensitive code paths touched in the removal.

### Summary Table
| Priority | Issue |
|----------|-------|
| 🟢 Green | Consider removing dead `existingCockpitSurface` getter (`TmuxBackend.swift:354`) |
| 🟡 Yellow | `makeActive` now double-fires on `.id()` recreation — the warning comment removed from old `makeNSView` was load-bearing |
| 🟡 Yellow | Startup purge in `SessionService.swift:134-138` skips `TmuxBackend.destroyTerminal` for each pruned row |
| 🟢 Green | Optional regression test for the `ensureRunningServer` adoption-on-existing-session path |
| 🟢 Green | One-way data loss on global-terminal purge — flag in release notes |

**Recommendation:** Approve. The two bug fixes are correct and well-documented; the feature removal is complete and consistent. The yellow items are minor follow-ups (a double-`select-window` per tab switch and a stale dead-code accessor) that are worth a quick cleanup pass but don't block merge.

---

dhilgaertner (approved):
## Code & Security Review

Reviewed commit `c386475` — the follow-up addressing both load-bearing nits from the prior approvals.

### Security Review
**Strengths:**
- No new external input, file I/O, or network calls. The new `CockpitSessionStarter` protocol is internal and only abstracts two existing ops.
- No change to process-spawn surface or arg quoting. `newSessionDetached` still routes through `TmuxController.run` with the same `-S socket` + `-f conf` plumbing.
- New unit tests use a pure in-memory fake; no tmux server, no temp files, no shell invocations.

**Concerns:**
- None.

### Code Quality

**`makeActive` dedup (`TmuxBackend.swift:255-270`, `:131`, `:331-333`)**
The fix is correctly shaped:
- The early-return is in `TmuxBackend.makeActive` (not `TerminalSurfaceView`) so the dedupe is enforced regardless of caller — both `makeNSView` and `updateNSView` benefit.
- Keyed by UUID, not window index — comment at L62-67 correctly calls out that tmux can reuse a freed window index, which would otherwise let a new terminal silently inherit the previous selection.
- `activeTerminalID` is recorded **after** `selectWindow` succeeds (L266), so a transient `tmux select-window` failure can't suppress the next attempt.
- Cleared on `destroyTerminal` (L332) when the active terminal is the one being destroyed, and on `shutdown` (L131). Both clearing paths look complete.
- `@MainActor`-isolated, so the read/write of `activeTerminalID` is serialized with all other state.

One residual edge: if tmux's active window changes out-of-band (user types `:next` inside the cockpit, scripted `select-window`, etc.), the cached `activeTerminalID` will desync from tmux. The cockpit isn't user-driven today (it's the hidden shared surface), and the next user-initiated tab switch supplies a different UUID which naturally heals the state, so this is benign and not worth coding around.

**`ensureCockpitSession` testability (`TmuxBackend.swift:421-455`, `TmuxController.swift:253-257`)**
- Clean extraction: `nonisolated static` is justified — the function reads/writes no instance state, only its `CockpitSessionStarter` arg.
- Protocol matches `TmuxController.newSessionDetached(configPath:env:command:)` exactly; the `extension TmuxController: CockpitSessionStarter {}` works without a shim because Swift's default-argument matching covers the call site in production (`env: [:]` passed explicitly there).
- The original three-line race-window contract is preserved verbatim: pre-check, attempt create, catch + recheck, re-throw only when the session genuinely doesn't exist.
- `controller = ctrl` is set on the **caller** side after `ensureCockpitSession` returns successfully (L417), so a thrown error never poisons the cache — same behavior as before.

**`EnsureCockpitSessionTests`**
- Covers all four branches: already-live (no create call), cold-create (one create call), lost-race adopt (create throws + adopt), genuine-failure re-throw. The lost-race test is the regression guard the prior review asked for — if the `catch` regresses to a bare `throw`, this test fails.
- `FakeStarter`'s sticky-last-value queue is reasonable for a 2-call max sequence. Drains correctly: `[false, true]` → `false` then `true`; `[false, false]` → `false` then `false`.
- Runs unconditionally (no tmux dependency), so the regression guard survives in environments without tmux installed.

**`makeActiveTracksActiveTerminal` integration test (`TmuxBackendTests.swift:68-97`)**
- Exercises the four real transitions: cold activate, dedup re-activate, switch to other terminal, destroy clears marker. Gated by `discoveredTmuxBinary != nil` on the suite, so it stays runnable.

### Carried-forward items (non-blocking, from prior reviews)
- 🟢 `existingCockpitSurface` getter (`TmuxBackend.swift:382`) is still dead — no callers after the `TerminalSurfaceView` refactor. Worth a one-line delete in a cleanup pass.
- 🟢 Startup purge in `SessionService.swift:130-138` removes persisted standalone-terminal rows but doesn't call `TmuxBackend.destroyTerminal` for each. Reaped by `TmuxOrphanReaper` in practice; symmetry nit only.
- 🟢 The standalone-terminal purge is one-way data loss on first launch with this build — release-notes item.
- ⚠️ PR is currently in `CONFLICTING` mergeable state — rebase needed before merge (separate from review verdict).

### Summary Table
| Priority | Issue |
|----------|-------|
| 🟢 | `existingCockpitSurface` getter still has no callers — delete in cleanup. |
| 🟢 | Purge path skips `destroyTerminal` for each pruned row (orphan reaper covers it). |
| 🟢 | Standalone-terminal purge is silent one-way data loss — flag in release notes. |

**Recommendation:** Approve. Commit `c386475` correctly addresses both the double-fire `select-window` regression and the missing-regression-test concern from the prior reviews. The dedupe is keyed and cleared correctly, the protocol-seam extraction preserves the original race semantics exactly, and the new tests would catch a regression of the catch-and-adopt branch back to a bare throw. Outstanding items are minor follow-ups.

---

dhilgaertner (approved):
## Code & Security Review

Reviewed at the merged head (`3ae6627` / equivalent of `c386475`). PR is already merged — this review is informational, but the verdict is APPROVE.

### Security Review

**Strengths:**
- New `CockpitSessionStarter` protocol is `internal`, no new external attack surface.
- `ensureCockpitSession` only reads `hasSession()` and re-throws on genuine failure — no privilege change, no new process spawn beyond the existing `new-session` call.
- `nonisolated static` extraction touches no instance state, so race-window semantics are preserved verbatim from the inline implementation.
- Sentinel / wrapper-log temp file handling (`$TMPDIR`) unchanged.
- New unit tests use a pure in-memory `FakeStarter`; no tmux server, no temp files, no shell invocations.

**Concerns:** None.

### Code Quality

**`makeActive` dedup (`TmuxBackend.swift:251-273`, `:131`, `:331-333`)**
- Dedup lives in the backend, not the view — both `makeNSView` and `updateNSView` benefit regardless of caller.
- Keyed by UUID, not window index — the L62-67 comment correctly justifies this; tmux can reuse a freed window index.
- `activeTerminalID` is recorded **after** `selectWindow` succeeds (L267), so a transient failure can't suppress the next attempt.
- Cleared on `destroyTerminal` when the active terminal is destroyed (L333) and on `shutdown` (L131). Both clearing paths look complete.
- `@MainActor`-isolated read/write, so no concurrent access concern.

One residual edge: out-of-band tmux activity (`:next` inside the cockpit, scripted `select-window`) desyncs the cached marker. The cockpit isn't user-driven and the next UI-initiated switch supplies a different UUID that naturally heals the state — benign, not worth coding around.

**`ensureCockpitSession` extraction (`TmuxBackend.swift:420-456`, `TmuxController.swift:253-256`)**
- Clean extraction: `nonisolated static` is justified (reads/writes no instance state).
- Protocol matches `TmuxController.newSessionDetached(configPath:env:command:)` exactly; conformance is a one-line extension with no shim.
- Original three-step race contract preserved: pre-check, attempt create, catch + recheck, re-throw only when the session genuinely doesn't exist.
- `controller = ctrl` is set on the **caller** side after `ensureCockpitSession` returns (L416), so a thrown error never poisons the cache.

**`syncSurface` deferral (`TerminalSurfaceView.swift:30-79`)**
- Moving surface lookup / re-parent / `makeActive` / first-responder onto the next main-queue turn is correct; the comment at L33-42 captures the live-lock root cause (run-loop pumping inside SwiftUI layout).
- Re-parent is gated on `surface.superview !== container`, so constraint churn is avoided on the common no-op path.
- After the dedup, repeat enqueues are cheap: `makeActive` early-returns and `makeFirstResponder` is idempotent.

**Tests**
- `EnsureCockpitSessionTests` covers the four real branches (already-live, cold-create, lost-race adopt, genuine-failure rethrow). The lost-race test is the regression guard the prior reviews asked for — if `catch` regresses to a bare `throw`, this test fails.
- `FakeStarter`'s sticky-last-value queue is the right shape for a 2-call max sequence. Drains correctly: `[false, true]` → `false` then `true`; `[false, false]` → `false` then `false`.
- `makeActiveTracksActiveTerminal` exercises cold activate, dedup re-activate, switch, and destroy-clears-marker. Gated by `discoveredTmuxBinary != nil` so it stays runnable on environments without tmux.

### Carried-forward items (non-blocking, from prior reviews)
- 🟢 `existingCockpitSurface` getter (`TmuxBackend.swift:382`) is still dead — `grep` confirms zero callers after the `TerminalSurfaceView` refactor. Worth a one-line delete in a cleanup pass.
- 🟢 Startup purge in `SessionService.swift:130-138` removes persisted standalone-terminal rows but doesn't call `TmuxBackend.destroyTerminal` for each. Reaped by `TmuxOrphanReaper` in practice; symmetry nit only.
- 🟢 The standalone-terminal purge is one-way data loss on first launch with this build — release-notes item.

### Summary Table
| Priority | Issue |
|----------|-------|
| 🟢 | `existingCockpitSurface` getter still has no callers — delete in cleanup. |
| 🟢 | Purge path skips `destroyTerminal` per pruned row (orphan reaper covers it). |
| 🟢 | Standalone-terminal purge is silent one-way data loss — flag in release notes. |

**Recommendation:** Approve. The `makeActive` dedup is keyed/cleared correctly and lives at the right layer. The `ensureCockpitSession` extraction preserves the original race semantics exactly and is now backed by a regression test against the catch-and-adopt branch. No security or correctness concerns; the remaining items are minor follow-ups.

---

PR #341, opened 2026-05-23
feat: jobs — manual "Run now", scrolling prompt editor, wider "Every" labels

### LINKED ISSUE #340: Jobs: manual run (even disabled), multiline scrolling prompt, widen 'Every' labels
## Summary

Three improvements to the Jobs feature (CROW-317):

### 1. Manually run jobs, even when disabled / unscheduled
Add a "Run now" action so a job can be fired on demand regardless of its `enabled`
flag or schedule. `SessionService.runJob` already ignores enabled/schedule, so the
work is exposing a manual trigger.

- New `AppState.onRunJob` callback, wired in `AppDelegate` to a new
  `JobScheduler.runNow(_:)`.
- Refactor `JobScheduler.tick()` to share a `fire()` helper (keeps the in-flight
  double-launch guard) used by both scheduled and manual runs.
- ▶ "Run now" button on every row in the Jobs settings tab.
- Manual runs record `lastRunAt` (consistent with scheduled runs).

### 2. Prompt box → multiline field that scrolls
Replace the capped `TextField(axis: .vertical).lineLimit(1...6)` in `JobFormView`
with a fixed-height `TextEditor` that scrolls internally, with a bordered look and a
placeholder overlay.

### 3. Widen the "Every" elements in the Schedule section
- Drop the redundant "Run" label so the segmented "Every / Daily at" toggle uses the
  full row width (so "Every" isn't clipped).
- Widen the interval number field from 70 → 90pt.

## Files
- `Packages/CrowCore/Sources/CrowCore/AppState.swift` — `onRunJob` callback
- `Sources/Crow/App/JobScheduler.swift` — extract `fire()`, add `runNow()`
- `Sources/Crow/App/AppDelegate.swift` — wire `appState.onRunJob`
- `Packages/CrowUI/Sources/CrowUI/SettingsView.swift` — "Run now" button
- `Packages/CrowUI/Sources/CrowUI/JobFormView.swift` — TextEditor prompts + "Every" widths

## Verification
- `swift build` / `swift test` (JobConfigTests) green.
- Run a **disabled** job via ▶ → worktree+session+Claude spins up, prompt fires, "Last run" updates.
- Run an **enabled** job via ▶ → fires immediately; in-flight guard prevents duplicate launch.
- Long multi-paragraph prompt scrolls within the editor; placeholder only shows when empty; round-trips on save/reopen.
- "Every / Daily at" toggle shows "Every" fully; interval field fits placeholder + 2–3 digits.

dgershman:
Closes #340

Three improvements to the Jobs settings feature (CROW-317).

## 1. Run now (even when disabled / unscheduled)
A ▶ button on every job row fires the job on demand, regardless of its `enabled` flag or schedule. `SessionService.runJob` already ignores those, so this exposes a manual trigger that reuses the existing launch path.

- `JobScheduler.tick()`'s fire-and-deliver logic is extracted into a shared `fire()` helper that owns the in-flight double-launch guard; a new `runNow(_:)` reuses it.
- Manual runs record `lastRunAt`, consistent with scheduled runs.
- Wired through a new `AppState.onRunJob` callback in `AppDelegate`.

## 2. Multiline scrolling prompt editor
Replaced the capped `TextField(axis: .vertical).lineLimit(1...6)` with a fixed-height `TextEditor` that scrolls internally, with a bordered look and a placeholder overlay shown only when empty.

## 3. Wider "Every" labels in the Schedule section
Dropped the redundant "Run" label so the segmented **Every / Daily at** toggle uses the full row width (no clipping), and widened the interval number field 70 → 90pt.

## Files
- `Packages/CrowCore/Sources/CrowCore/AppState.swift` — `onRunJob` callback
- `Sources/Crow/App/JobScheduler.swift` — extract `fire()`, add `runNow(_:)`
- `Sources/Crow/App/AppDelegate.swift` — wire `appState.onRunJob`
- `Packages/CrowUI/Sources/CrowUI/SettingsView.swift` — "Run now" ▶ button
- `Packages/CrowUI/Sources/CrowUI/JobFormView.swift` — `TextEditor` prompts + widened "Every"

## Verification
- `swift build --arch arm64` → build complete (arm64 required: this host's shell runs under Rosetta, and the GhosttyKit xcframework only ships an arm64 slice).
- `swift test` in `Packages/CrowCore` → 186 tests pass, including `JobConfigTests`.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -332,6 +332,9 @@ public final class AppState {
     /// Called when the sound mute toggle is changed.
     public var onSoundMutedChanged: ((Bool) -> Void)?
 
+    /// Fire a job immediately, ignoring its enabled flag and schedule (job ID).
+    public var onRunJob: ((UUID) -> Void)?
+
     // MARK: - Computed Properties
 
     public var selectedSession: Session? {
--- a/Packages/CrowUI/Sources/CrowUI/JobFormView.swift
+++ b/Packages/CrowUI/Sources/CrowUI/JobFormView.swift
@@ -198,9 +198,21 @@ public struct JobFormView: View {
             Section {
                 ForEach(prompts.indices, id: \.self) { idx in
                     HStack(alignment: .top) {
-                        TextField("Prompt \(idx + 1)", text: $prompts[idx], axis: .vertical)
-                            .textFieldStyle(.roundedBorder)
-                            .lineLimit(1...6)
+                        TextEditor(text: $prompts[idx])
+                            .font(.body)
+                            .scrollContentBackground(.hidden)
+                            .frame(height: 90)
+                            .padding(4)
+                            .overlay(RoundedRectangle(cornerRadius: 6).stroke(Color.secondary.opacity(0.3)))
+                            .overlay(alignment: .topLeading) {
+                                if prompts[idx].isEmpty {
+                                    Text("Prompt \(idx + 1)")
+                                        .foregroundStyle(.secondary)
+                                        .padding(.horizontal, 9)
+                                        .padding(.vertical, 12)
+                                        .allowsHitTesting(false)
+                                }
+                            }
                         Button(role: .destructive) {
                             prompts.remove(at: idx)
                             if prompts.isEmpty { prompts = [""] }
@@ -229,17 +241,18 @@ public struct JobFormView: View {
             }
 
             Section("Schedule") {
-                Picker("Run", selection: $scheduleMode) {
+                Picker("", selection: $scheduleMode) {
                     Text("Every").tag(ScheduleMode.interval)
                     Text("Daily at").tag(ScheduleMode.daily)
                 }
                 .pickerStyle(.segmented)
+                .labelsHidden()
 
                 if scheduleMode == .interval {
                     HStack {
                         TextField("Every", value: $intervalValue, format: .number)
                             .textFieldStyle(.roundedBorder)
-                            .frame(width: 70)
+                            .frame(width: 90)
                         Picker("", selection: $intervalUnit) {
                             ForEach(IntervalUnit.allCases) { Text($0.label).tag($0) }
                         }
--- a/Packages/CrowUI/Sources/CrowUI/SettingsView.swift
+++ b/Packages/CrowUI/Sources/CrowUI/SettingsView.swift
@@ -352,6 +352,14 @@ public struct SettingsView: View {
                             Spacer()
 
                             Button {
+                                appState.onRunJob?(job.id)
+                            } label: {
+                                Image(systemName: "play.fill")
+                            }
+                            .buttonStyle(.borderless)
+                            .accessibilityLabel("Run \(job.name) now")
+
+                            Button {
                                 editingJob = job
                             } label: {
                                 Image(systemName: "pencil")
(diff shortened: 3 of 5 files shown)


dhilgaertner (approved):
## Code & Security Review

Small, focused change (5 files, +59/-14). The `fire()` extraction is the right shape: it makes `tick()` and `runNow()` share one chokepoint for the in-flight guard, the Task spawn, and `onJobRan` persistence. Wiring through an `AppState.onRunJob` callback (rather than handing the scheduler to the view) is consistent with the existing pattern (`onMarkInReview`, `onSoundMutedChanged`, etc.). No security concerns — the play button is a simple ID dispatch; prompts continue to flow through the existing `TerminalRouter` path.

`JobConfigTests` pass locally (9/9). Full `swift build` couldn't run in this sandbox (GhosttyKit xcframework is arm64-only and this shell is under Rosetta — same constraint the PR description calls out), but the changes are mechanical enough that the passing core tests are reasonable signal.

### Security Review
**Strengths:**
- No new external input surface — the Run-now button just dispatches the job's existing ID through the existing `SessionService.runJob` path.
- Concurrency-safe: both `tick()` and `runNow()` are `@MainActor`, and the `inFlight` set is mutated only on the main actor; the guard inside `fire()` correctly serializes a manual click against an in-progress scheduled launch.
- `onRunJob` is `Optional` and the call site uses `?.` — no crash if the callback isn't wired (e.g., in a unit-test harness).

**Concerns:**
- None.

### Code Quality

**Yellow — worth a follow-up but not blocking:**

- `Packages/CrowUI/Sources/CrowUI/JobFormView.swift:201` — accessibility regression. The previous `TextField(\"Prompt \\(idx + 1)\", ...)` exposed the label to VoiceOver; the new `TextEditor` only shows a visual placeholder overlay and has no `.accessibilityLabel`. The Remove button next to it (line 224) does have one, which makes the missing label on the editor more conspicuous. Suggest: `.accessibilityLabel(\"Prompt \\(idx + 1)\")` on the `TextEditor`.
- `Sources/Crow/App/JobScheduler.swift:67` — the `guard !inFlight.contains(job.id) else { continue }` check in `tick()` is now redundant with the same guard inside `fire()` (line 87). Harmless, but the comment on the `inFlight` property still talks about the scheduled-run rationale; consider either removing the outer guard or expanding the comment to cover manual runs.
- `Packages/CrowUI/Sources/CrowUI/SettingsView.swift:354` — Run-now button has no visual indication while a launch is in flight (worktree creation can take a few seconds). A second click during that window is silently dropped by the `inFlight` guard. Not a correctness bug, but a brief disabled/progress state would prevent users from wondering whether the first click registered. Could also disable the button when `appState.onRunJob == nil` to mirror the wiring.

**Green — design notes, not asks:**

- `JobScheduler.runNow` calls `onJobRan` on success, which updates `lastRunAt` and therefore shifts the next scheduled fire. The PR description flags this as intentional (\"consistent with scheduled runs\"), so this is a documented choice — just worth keeping in mind for users who think of \"Run now\" as additive to the schedule. The opposite choice (don't bump `lastRunAt` on manual runs) is also defensible; current behavior is fine.
- `TextEditor` placeholder padding `(9, 12)` is empirical to match the editor's text inset; if SwiftUI ever changes that inset the placeholder will drift. Minor.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | — |
| Yellow | Missing `accessibilityLabel` on new prompt `TextEditor` |
| Yellow | Redundant `inFlight` guard in `tick()` after `fire()` extraction |
| Yellow | No in-flight visual state on the Run-now button |
| Green | Manual run updates `lastRunAt` (delays next scheduled fire — intentional, documented) |
| Green | Placeholder overlay padding is empirically matched to `TextEditor` inset |

**Recommendation:** Approve. The change is well-scoped, the refactor cleanly removes the duplicated launch path, and there are no correctness or security concerns. The yellow items are worth a follow-up but shouldn't block merge.

---

PR #342, opened 2026-05-23
feat: "Changes" on Manager sidebar row + user-selectable summary repos

### LINKED ISSUE #339: Move 'Changes Summary' onto the Manager sidebar line (before Manager), rename to 'Changes'
## Summary

In the Crow sidebar, the changes-summary entry currently sits on its **own full-width row** (`SummaryBoardSidebarRow`, labeled "Changes Summary"), directly above the **Manager row** (`ManagerSidebarRow`: the `Manager` button + the `+` "New Manager" button).

Move the changes-summary button **up onto the same line as Manager, positioned before it**, and shorten the label from "Changes Summary" to just **"Changes"**.

Target row layout: `[ Changes ] [ Manager ] [ + ]`

## Implementation notes

All edits in `Packages/CrowUI/Sources/CrowUI/SessionListView.swift`.

1. **Collapse the two rows (lines ~80–90):** Remove the separate `SummaryBoardSidebarRow` row and the conditional `ManagerSidebarRow` row; replace with a single **unconditional** `ManagerSidebarRow`. (Manager row was gated on `appState.managerSession != nil`, which is nil only briefly before the primary manager loads — gate the Manager/`+` buttons *inside* the view so **Changes stays always visible**.)

2. **Add the "Changes" button into `ManagerSidebarRow` (struct ~389–452):** Make it the first element of the existing `HStack(spacing: 6)`, reusing the private `sidebarButton(title:isActive:action:)` helper so it matches the Manager button styling.
   - `isActive: appState.selectedSessionID == AppState.summaryBoardSessionID`
   - action: `appState.selectedSessionID = AppState.summaryBoardSessionID`
   - Preserve the loading spinner as a trailing overlay gated on `appState.isLoadingSummary`.
   - Gate the existing Manager + `+` buttons on `appState.managerSession != nil`.
   - Both `Changes` and `Manager` use `sidebarButton` (`maxWidth: .infinity`) → they split row width evenly, with the fixed 28pt `+` on the right.

3. **Remove the now-unused `SummaryBoardSidebarRow` struct (~456–496)** plus its `// MARK:` comment. The `chart.bar.doc.horizontal` icon and "Changes Summary" string go away; the new inline button is text-only ("Changes") to match Manager.

## Verification

1. `swift build` (or build the Crow scheme).
2. Sidebar shows a single row `[ Changes ] [ Manager ] [ + ]`, Changes left of Manager.
3. Click **Changes** → changes-summary board opens and **Generate** still populates.
4. Click **Manager** → manager selects; `+` still creates a new Manager.
5. Loading spinner shows on the Changes button while a summary generates.

## Note (informational — no change)

The change summary is **not** scoped to a session/worktree. `GitManager.summarizeCommits` → `discoverRepos()` scans **every git repo under the dev root** (all workspace subfolders, skipping `excludeDirs`), running `git log --all --no-merges --since=… --numstat` per repo. `--all` counts **all branches, including worktree feature branches**.

dgershman:
Closes #339

## Part A — sidebar relabel/move (ticket #339)
Collapses the standalone "Changes Summary" sidebar row onto the Manager line, before Manager, renamed to **"Changes"**. Row layout is now `[ Changes ] [ Manager ] [ + ]`. "Changes" is always visible; the Manager and `+` buttons gate on the primary Manager having loaded. The now-unused `SummaryBoardSidebarRow` is removed.

## Part B — user-selectable summary repos
Previously the changes summary silently scanned **every** git repo under the dev root, with no way to control which. This adds a **persisted setting** — a checklist of discovered repos under **Settings → General → Changes Summary** — that scopes the summary. Per design, an **empty selection summarizes nothing** (explicit opt-in).

- `ConfigDefaults.summaryRepos: [String]` — selections stored as `"workspace/name"` keys in `config.json`.
- `GitManager.summarizeCommits` gains `includeRepos: Set<String>?` (nil = all repos, preserving prior behavior for callers that don't pass it; empty set = nothing).
- `AppState.onListSummaryRepos` lists discovered repos for the picker, wired in `AppDelegate` to `GitManager.discoverRepos`.
- Both `onGenerateSummary` (in-app board) and the `get-summary` RPC (`crow summary` CLI) pass the configured set, so both surfaces honor the same scope.
- New `SummaryReposPicker` checklist view; `SummaryBoardView` help/empty-state copy updated to point at the setting.

## Verification
- `swift build --arch arm64` for CrowCore, CrowGit, CrowUI, and the root Crow app — all compile/link cleanly.
- Existing `CommitSummaryTests` are unaffected (new param defaults to `nil`).

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -261,6 +261,10 @@ public final class AppState {
     /// per-repo commit digest for the given window (git date strings).
     public var onGenerateSummary: ((_ since: String, _ until: String?) async -> [RepoCommitSummary])?
 
+    /// Lists discovered repos (as "workspace/name" keys) for the
+    /// Changes-summary scope picker in Settings.
+    public var onListSummaryRepos: (() async -> [String])?
+
     /// Called to launch Claude in a terminal that just became ready.
     public var onLaunchClaude: ((UUID) -> Void)?  // receives terminal ID
 
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -224,6 +224,9 @@ public struct ConfigDefaults: Codable, Sendable, Equatable {
     public var excludeReviewRepos: [String]
     public var excludeTicketRepos: [String]
     public var ignoreReviewLabels: [String]
+    /// Repos (as "workspace/name" keys) the Changes board summarizes. Empty
+    /// means nothing is summarized — the user opts in explicitly.
+    public var summaryRepos: [String]
 
     /// Characters that are invalid in git ref names (see `git check-ref-format`).
     private static let invalidBranchChars = CharacterSet(charactersIn: " ~^:?*[\\")
@@ -250,7 +253,8 @@ public struct ConfigDefaults: Codable, Sendable, Equatable {
         excludeDirs: [String] = ["node_modules", ".git", "vendor", "dist", "build", "target"],
         excludeReviewRepos: [String] = [],
         excludeTicketRepos: [String] = [],
-        ignoreReviewLabels: [String] = []
+        ignoreReviewLabels: [String] = [],
+        summaryRepos: [String] = []
     ) {
         self.provider = provider
         self.cli = cli
@@ -259,6 +263,7 @@ public struct ConfigDefaults: Codable, Sendable, Equatable {
         self.excludeReviewRepos = excludeReviewRepos
         self.excludeTicketRepos = excludeTicketRepos
         self.ignoreReviewLabels = ignoreReviewLabels
+        self.summaryRepos = summaryRepos
     }
 
     public init(from decoder: Decoder) throws {
@@ -270,10 +275,11 @@ public struct ConfigDefaults: Codable, Sendable, Equatable {
         excludeReviewRepos = try container.decodeIfPresent([String].self, forKey: .excludeReviewRepos) ?? []
         excludeTicketRepos = try container.decodeIfPresent([String].self, forKey: .excludeTicketRepos) ?? []
         ignoreReviewLabels = try container.decodeIfPresent([String].self, forKey: .ignoreReviewLabels) ?? []
+        summaryRepos = try container.decodeIfPresent([String].self, forKey: .summaryRepos) ?? []
     }
 
     private enum CodingKeys: String, CodingKey {
-        case provider, cli, branchPrefix, excludeDirs, excludeReviewRepos, excludeTicketRepos, ignoreReviewLabels
+        case provider, cli, branchPrefix, excludeDirs, excludeReviewRepos, excludeTicketRepos, ignoreReviewLabels, summaryRepos
     }
 }
 
--- a/Packages/CrowGit/Sources/CrowGit/GitManager.swift
+++ b/Packages/CrowGit/Sources/CrowGit/GitManager.swift
@@ -231,11 +231,14 @@ public actor GitManager {
         }
     }
 
-    /// Collect commits across every discovered repo within a time window and
-    /// group them by repo. Deterministic digest — no LLM.
+    /// Collect commits across discovered repos within a time window and group
+    /// them by repo. Deterministic digest — no LLM.
     ///
     /// - `since`/`until` are passed straight to git, which parses flexible date
     ///   strings ("1 week ago", "2026-05-01"); `until` defaults to now when nil.
+    /// - `includeRepos`, when non-nil, scopes the scan to repos whose
+    ///   "workspace/name" key is in the set (an empty set yields nothing); `nil`
+    ///   includes every discovered repo.
     /// - `--all` picks up commits on every branch (worktree feature branches
     ///   share the main checkout's object store), `--no-merges` keeps diffstats
     ///   from double-counting merge commits.
@@ -245,8 +248,13 @@ public actor GitManager {
     ///
     /// Runs sequentially: `run` is actor-isolated and blocks on
     /// `waitUntilExit`, so a task group would serialize on the actor anyway.
-    public func summarizeCommits(since: String, until: String?) async throws -> [RepoCommitSummary] {
-        let repos = try await discoverRepos()
+    public func summarizeCommits(since: String, until: String?, includeRepos: Set<String>? = nil) async throws -> [RepoCommitSummary] {
+        var repos = try await discoverRepos()
+        // When an include-set is provided, scope to repos whose "workspace/name"
+        // key is selected; an empty set yields nothing. `nil` keeps all repos.
+        if let includeRepos {
+            repos = repos.filter { includeRepos.contains("\($0.workspace)/\($0.name)") }
+        }
         var summaries: [RepoCommitSummary] = []
         for repo in repos {
             var args = [
--- a/Packages/CrowUI/Sources/CrowUI/SessionListView.swift
+++ b/Packages/CrowUI/Sources/CrowUI/SessionListView.swift
@@ -77,17 +77,10 @@ public struct SessionListView: View {
                 .listRowSeparator(.hidden)
                 .listRowBackground(Color.clear)
 
-            // Changes Summary row
-            SummaryBoardSidebarRow(appState: appState)
+            // Changes + Manager + New Manager (+) row
+            ManagerSidebarRow(appState: appState)
                 .listRowSeparator(.hidden)
                 .listRowBackground(Color.clear)
-
-            // Manager + New Manager (+) row
-            if appState.managerSession != nil {
-                ManagerSidebarRow(appState: appState)
-                    .listRowSeparator(.hidden)
-                    .listRowBackground(Color.clear)
-            }
 
             // Additional (non-primary) Manager sessions, each deletable.
             let extraManagers = appState.managerSessions.filter { $0.id != AppState.managerSessionID }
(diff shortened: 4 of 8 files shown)


dhilgaertner (dismissed):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- New `summaryRepos` field decodes with `decodeIfPresent` → backwards compatible; no config-version bump or migration risk.
- `includeRepos: Set<String>?` keeps the prior signature working (`nil = all repos`), so existing tests/callers retain their behavior; matching unit-level intent is preserved.
- No new external command construction or user-input → shell paths; the include set is matched purely against the discovered `workspace/name` keys (set membership), so no injection surface is added.
- Empty-selection-means-nothing is an explicit opt-in, which is the safe default for a feature that walks the disk and reports commit metadata.

**Concerns:**
- None observed.

### Code Quality

**Yellow — should fix before merge**

- **`crow summary` (`get-summary` RPC) ignores Settings changes until app restart.** `AppDelegate.swift:965` snapshots `let capturedSummaryRepos = Set(appConfig?.defaults.summaryRepos ?? [])` at socket-server start, and the handler at `AppDelegate.swift:1559` uses that snapshot forever. `saveSettings` (`AppDelegate.swift:902`) updates `self.appConfig` and persists to disk, but never refreshes any captured router state. Net effect: after the user toggles repos in *Settings → Changes Summary*, the in-app board picks up the change immediately (because `onGenerateSummary` at `AppDelegate.swift:375` reads `self.appConfig?.defaults.summaryRepos` at call time), but `crow summary` continues to return commits scoped to whatever was selected at launch. This directly contradicts the PR description's claim that "both surfaces honor the same scope." Same staleness applies to `capturedExcludeDirs` for `summaryRepos`-related calls — but `excludeDirs` is rarely changed at runtime, while `summaryRepos` is the *entire point* of the new picker.
  - Fix: read `appConfig` lazily inside the handler (e.g. capture `self` weakly and read `self?.appConfig?.defaults.summaryRepos` / `excludeDirs` per-call), or introduce a `summaryReposProvider: () -> Set<String>` similar to `tracker.autoMergeWatcherEnabledProvider` / `scheduler.jobsProvider`, which is already the pattern used elsewhere in this file for live config reads.

**Green — consider**

- `SummaryReposPicker.binding(for:)` at `SummaryReposPicker.swift:39` does `selected.contains(repo)` and `selected.removeAll { $0 == repo }` per-toggle — O(n) per row, O(n²) for the whole picker render. Fine for tens of repos; if dev roots in the wild trend large (hundreds of repos), a `Set<String>`-backed view model would render more cleanly. No action required at current scale.
- `SummaryReposPicker.load()` is gated by `didLoad`, so a repo added to the dev root after Settings is opened won't appear until the sheet is dismissed and reopened. Acceptable; worth a one-line comment if you want to leave a breadcrumb.
- `appState.onListSummaryRepos` and `appState.onGenerateSummary` each build a fresh `GitManager` with an empty `workspaces: [:]` map. Functionally fine (the actor is cheap and `discoverRepos` only needs `devRoot` + `excludeDirs`), but the two call sites duplicate the construction. Trivial — only worth factoring if a third caller appears.
- `ManagerSidebarRow` always shows the "Changes" button but gates "Manager" + "+" on `appState.managerSession != nil`. Per `ensureManagerSession(devRoot:)` the primary Manager is always seeded at launch, so this gate is effectively defensive — fine, just worth a quick mental check that the loading window between hydrate and `ensureManagerSession` is short enough that the user never sees a "Changes"-only row in practice. (It is.)
- The new `SummaryReposPicker` has no test coverage. The picker logic is thin enough that a snapshot/test isn't load-bearing, but a small unit test for the on/off binding behavior would catch a future regression where `onSave` stops firing.

### Summary Table
| Priority | Issue |
|----------|-------|
| 🟡 Yellow | `get-summary` RPC handler uses a stale snapshot of `summaryRepos` from server-start; CLI scope drifts from in-app scope after any Settings edit until app restart. (`AppDelegate.swift:965`, `:1559`) |
| 🟢 Green | `SummaryReposPicker` uses an Array for membership checks (O(n²) render with many repos). |
| 🟢 Green | Picker only loads repos once per sheet open — new repos added to the dev root mid-session aren't discovered until reopen. |
| 🟢 Green | No unit test for the new picker's toggle/save behavior. |

**Recommendation:** Request Changes. The CLI/in-app scope divergence is a small but user-visible correctness issue that contradicts the PR's stated design. Once the handler reads `summaryRepos` lazily (one-line change at `AppDelegate.swift:965`/`:1559`), the rest of the PR looks solid — the sidebar reshuffle is clean, the new `includeRepos` parameter is backwards compatible, and the empty-set-means-nothing opt-in is the right safety default.

---

dgershman:
Thanks for the review! Addressed in 335544f:

**🟡 Yellow — fixed.** The `get-summary` RPC handler no longer snapshots `summaryRepos`/`excludeDirs` at server start. Both are now read per-call via `@Sendable` provider closures that hop to the main actor (`MainActor.run { self?.appConfig?.defaults... }`), mirroring the existing `jobsProvider`/`autoMergeWatcherEnabledProvider` pattern. `crow summary` now tracks Settings edits live, matching the in-app board — no restart needed. `devRoot` stays snapshotted since it's fixed for the socket server's lifetime.

**🟢 Green — picker load-once:** added a one-line breadcrumb comment on `SummaryReposPicker.load()`.

On the remaining 🟢 items, all of which you flagged as optional/no-action — leaving as-is to keep the diff focused:
- **Array vs `Set` membership in the picker:** O(n²) render is negligible at realistic repo counts; a `Set`-backed view model would be the move only if dev roots grow into the hundreds.
- **Duplicate `GitManager` construction:** still just the two call sites; will factor out a helper if a third appears, per your note.
- **No unit test for the picker:** the toggle/save logic lives inside a SwiftUI `View`'s `@State`, so a load-bearing test would need extracting a view model — not worth it for logic this thin. Happy to add one if you'd prefer.

Re-requesting review.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review

**Strengths:**
- `summaryRepos` is stored as plain `workspace/name` keys; the actual `git log` runs against `repo.path` returned by `discoverRepos()`, not against any user-supplied path, so the picker can't be used to escape the dev root.
- Directory names cannot contain `/` on macOS, so the `"\(workspace)/\(name)"` key is unambiguous (no key-collision/forgery concern).
- No secrets, network calls, or shell composition were added.
- The follow-up commit (`335544f`) tightened the `get-summary` RPC to read `summaryRepos`/`excludeDirs` live via main-actor hops instead of snapshotting at server start — so toggling in Settings can't drift between the in-app board and the CLI.

**Concerns:**
- None blocking.

### Code Quality

- `GitManager.summarizeCommits` (Packages/CrowGit/Sources/CrowGit/GitManager.swift:251): `includeRepos: nil` = all, `includeRepos: []` = none — the contrast is documented in the doc comment and the call-site comment, and behavior is consistent with the new "explicit opt-in" model.
- `AppDelegate.onGenerateSummary` and `onListSummaryRepos` (Sources/Crow/App/AppDelegate.swift:372–395) both build a `GitManager` with `workspaces: [:]` — fine because the picker only needs `workspace/name` and `summarizeCommits` only uses `repo.path`. The provider/cli/host fields on `RepoInfo` are unused on these paths.
- `SessionListView.ManagerSidebarRow` (Packages/CrowUI/Sources/CrowUI/SessionListView.swift:383) — "Changes" is always visible; Manager + `+` correctly gate on `appState.managerSession != nil`. Matches the PR description.
- `SummaryReposPicker.load()` (Packages/CrowUI/Sources/CrowUI/SummaryReposPicker.swift:55) uses a `didLoad` guard so a repo added to the dev root while Settings is open won't appear until the panel is reopened. Documented in the comment — acceptable for a Settings panel.
- `ConfigDefaults` (Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift:229–278) follows the established forward-compat pattern (`decodeIfPresent ?? []`) for `summaryRepos`.
- Existing `CommitSummaryTests` still pass; new `includeRepos` parameter is source-compatible via its default.

### Behavior Change to Flag (not blocking)

Existing users upgrading from a version where the Changes board scanned every discovered repo will land on an empty `summaryRepos` and see no commits until they opt-in repos under Settings → General → Changes Summary. The empty-state copy in `SummaryBoardView` directs them there, so discoverability is fine, but it's a silent regression in behavior for anyone who relied on the prior "everything" default. Worth a note in release/changelog if you do those.

### Suggestions (Green — consider)

1. **Add coverage for `includeRepos`** in `CommitSummaryTests` — the three branches (`nil`, empty set, single repo) are trivial to verify against the existing harness and would lock in the documented `nil` vs `[]` contract.
2. **Picker write amplification**: each checkbox toggle in `SummaryReposPicker` triggers `save()`, which writes the entire `config.json`. Not a correctness issue, just N file writes if a user toggles many at once. A short debounce would smooth this, but it's an unnecessary optimization at current scale.
3. **`SummaryBoardView` empty state** is the same whether (a) no repos were selected or (b) repos were selected but the window had zero commits. Differentiating ("No repos selected for summary." vs. "No commits in this window.") would make the opt-in step clearer to first-time users.

### Summary Table

| Priority | Issue |
|----------|-------|
| Red | — |
| Yellow | Behavior change for upgrades (empty `summaryRepos` ⇒ nothing summarized) — intentional, mention in release notes |
| Green | Add `includeRepos` test coverage; debounce picker saves; differentiate empty-state copy |

**Recommendation:** Approve. The feature is well-scoped, the follow-up commit closes the staleness gap on `get-summary`, security posture is unchanged, and existing tests pass. The opt-in behavior change is deliberate and surfaced in the UI.

---

dhilgaertner (approved):
## Code & Security Review

Nicely scoped change with backward-compatible Codable migration, sensible nil-vs-empty semantics for `includeRepos`, and the live-read providers in the get-summary RPC handler keep `crow summary` and the in-app board on the same wavelength. CrowCore builds cleanly; CrowGit tests pass (CrowUI couldn't be built locally — missing `GhosttyKit.xcframework` binary, unrelated to this PR).

### Critical Issues
None — no blocking issues found.

### Security Review
**Strengths:**
- No new user input crosses a shell or DB boundary; `summaryRepos` is consumed only as a `Set<String>.contains(...)` filter against `discoverRepos()` output, so user-supplied strings never reach `git`.
- Backward-compatible `decodeIfPresent` for the new `summaryRepos` field — old `config.json` files load without throwing.
- Provider closures use `[weak self]` and hop to `MainActor.run` for live reads, avoiding races with Settings edits while keeping `appConfig` access actor-isolated.

**Concerns:**
- None identified.

### Code Quality

**Behavior change worth a release note (yellow):** existing users had the Changes board summarizing every repo under devRoot; after upgrade, `summaryRepos` decodes to `[]`, which the wired callers (`onGenerateSummary`, `get-summary` RPC) pass as an empty `Set` — meaning the board/CLI silently goes empty until the user opts in via Settings. The PR description calls this out as intentional, but for upgrading users it will look like the feature broke. Consider a one-time onboarding/migration hint, or note this in CHANGELOG/release notes so users aren't confused.

**Test coverage gap (yellow):** `CommitSummaryTests` covers `summarizeCommits` but no test exercises the new `includeRepos` filter. The integration harness already builds a multi-repo dev root in spirit — adding a case with two repos where `includeRepos` selects one (plus `nil` keeps all, empty `Set()` yields none) would lock the contract that callers depend on. `Packages/CrowGit/Tests/CrowGitTests/CommitSummaryTests.swift:94`.

**Picker reload is one-shot per view lifetime (green):** `SummaryReposPicker.didLoad` gates `load()` so adding a workspace/repo while Settings is open won't reflect until it's reopened. Documented in the code comment, but a `.refreshable`-style affordance (or invalidating `didLoad` on workspace changes) would be a small UX win. `Packages/CrowUI/Sources/CrowUI/SummaryReposPicker.swift:53`.

**Stale selections aren't pruned (green):** If a user deletes/renames a workspace or repo, its `"workspace/name"` key stays in `summaryRepos` forever (just filtered out by `discoverRepos`). Harmless, but the picker could optionally show orphaned selections with a "(missing — remove)" hint so they don't accumulate silently.

**Spinner overlap in the "Changes" button (green):** the loading `ProgressView` is now an `.overlay(alignment: .trailing)` on a centered-text button, so on narrow sidebars it can sit on top of the "Changes" label. The previous full-width row used `Spacer()` and reflowed cleanly. Switching to a leading-icon HStack inside the button, or shrinking the active region, would avoid the overlap. `Packages/CrowUI/Sources/CrowUI/SessionListView.swift:394`.

**Missing a11y on the "Changes" button (green):** the shared `sidebarButton` helper on `ManagerSidebarRow` doesn't attach `.help` / `.accessibilityLabel`, so neither "Changes" nor "Manager" gets a hover tooltip or VO label — whereas the sibling `+` button does. The original `SummaryBoardSidebarRow` also didn't, so this is not a regression, but worth a one-line fix while you're here. `Packages/CrowUI/Sources/CrowUI/SessionListView.swift:441`.

**Per-call `GitManager` construction (green / consider):** `onGenerateSummary`, `onListSummaryRepos`, and the `get-summary` RPC each instantiate a fresh `GitManager` actor on every call. Cheap, but if you ever want to cache the discovered-repos result for the picker session, a single instance would make that cleaner.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | — |
| Yellow | Upgrade behavior change: existing users see an empty Changes board until they opt in — call this out in release notes |
| Yellow | No unit test for the new `includeRepos` filter in `summarizeCommits` |
| Green | `SummaryReposPicker` doesn't refresh after initial load (documented) |
| Green | Stale `summaryRepos` entries linger forever |
| Green | Spinner overlay can overlap centered "Changes" text on narrow sidebars |
| Green | `sidebarButton` lacks `.help` / `.accessibilityLabel` (pre-existing, but worth fixing) |

**Recommendation:** Approve — no blocking issues. The yellow items are worth addressing in a follow-up (especially the test for `includeRepos` and a release note for the opt-in behavior change), but they shouldn't gate merge.

---

PR #351, opened 2026-05-23
fix: Changes-summary board — dedupe PR clones, fetch remote main, scrollbar, dropdown (#348)

### LINKED ISSUE #348: Changes summary: dedupe PR-review clones, fetch latest main, repo dropdown, scrollbar, rename Summarize
Follow-up to #345 / #346. Five rough edges on the Changes-summary board surfaced in use.

### 1. Only show matching repos — exclude `crow-pr-*` review clones
Now that scope keys on the `org/repo` slug derived from each repo's `origin` remote (#346), PR-review clones collide with the real checkout. The PR-review feature clones into `<devRoot>/crow-reviews/<repo>-pr-<number>` as **full clones** (`SessionService.swift:1087`), so `GitManager.discoverRepos` picks them up and they resolve to the *same* `org/repo` slug as the canonical repo — producing duplicate rows (e.g. `crow` and `crow-pr-346` both as `radiusmethod/crow`).

- Dedupe by slug, keeping the canonical checkout (e.g. the one whose directory basename equals the repo name), **or** exclude the `crow-reviews` workspace dir / `*-pr-<n>` clones from the summary scan.
- Acceptance: each configured `org/repo` appears at most once; transient PR-review clones never appear.

### 2. Pull the latest `main` before summarizing
`GitManager.summarizeCommits` runs `git log --all --since=…` against whatever is checked out locally (`GitManager.swift`). It never fetches, so commits merged to `main` on the remote but not pulled locally are missed, and `--all` mixes in local feature branches.

- Before scanning each scoped repo, `git fetch origin <defaultBranch>` and base the digest on `origin/<defaultBranch>` (the latest `main`) instead of `--all` local branches.
- Resolve the default branch per repo (e.g. `git remote show origin` / `symbolic-ref refs/remotes/origin/HEAD`).
- Acceptance: the digest reflects what landed on the remote default branch in the window, regardless of local state. (Consider: fetch cost across many repos — fetch only scoped repos, shallow/`--no-tags`.)

### 3. Rename the "LLM Summarize" button to "Summarize"
Just the label in `SummaryBoardView` (the `sparkles` button). The underlying `ClaudeSummarizer` behavior is unchanged.

### 4. Results window has no scrollbar
The per-repo results list in `SummaryBoardView.resultsList` doesn't show a scrollbar / doesn't scroll when commits overflow the pane. Make the list scroll with a visible scroller when content exceeds the viewport.

### 5. Dropdown showing the in-scope repo list
Add a dropdown/disclosure on the board that lists the repos from the configured comma-separated scope (`config.defaults.summaryRepos`), so it's clear which repos' commits are being shown. Reuse the existing `summaryRepos` config.

---

### Acceptance summary
- [ ] Each configured `org/repo` shows once; `crow-pr-*` review clones are excluded.
- [ ] Digest is based on the freshly-fetched remote default branch (`origin/main`).
- [ ] Button reads "Summarize".
- [ ] Results list scrolls with a visible scrollbar.
- [ ] A dropdown lists the in-scope repos from the CSV.

### Key files
- `Packages/CrowGit/Sources/CrowGit/GitManager.swift` — `discoverRepos`, `summarizeCommits`, `remoteInfo`
- `Packages/CrowUI/Sources/CrowUI/SummaryBoardView.swift` — button label, `resultsList` scroll, repo dropdown
- `Sources/Crow/App/SessionService.swift:1087` — `crow-reviews/<repo>-pr-<n>` clone naming (for the exclusion rule)
- `Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift` — `summaryRepos`

dgershman:
Closes #348

Follow-up to #345 / #346 — fixes five rough edges on the Changes-summary board.

## Changes

1. **Dedupe PR-review clones** (`GitManager.discoverRepos` / `summarizeCommits`) — the `crow-reviews` workspace dir (full clones at `<devRoot>/crow-reviews/<repo>-pr-<n>`) is skipped during discovery, and the scan dedupes by `org/repo` slug. Each configured repo now appears at most once; transient `*-pr-<n>` clones never show.
2. **Fetch the remote default branch** (`GitManager.summarizeCommits`) — before scanning each in-scope repo we `git fetch --no-tags origin <defaultBranch>` (resolved from `origin/HEAD`) and base the digest on `origin/<defaultBranch>` instead of `--all` local branches. Falls back to `--all` when the default branch can't be resolved, and the fetch is best-effort so an offline machine still produces a (stale) digest.
3. **Rename button** — the `sparkles` button label "LLM Summarize" now reads "Summarize" (`ClaudeSummarizer` behavior unchanged).
4. **Visible scrollbar** — the per-repo results `List` gets `.scrollIndicators(.visible)` so it scrolls with a persistent scroller on overflow.
5. **In-scope repo dropdown** — a header `Menu` lists the configured scope (`config.defaults.summaryRepos`, synced into `AppState.summaryRepoScope`), making it clear which repos' commits are shown.

## Verification
- `swift build --arch arm64` — app builds.
- `swift test` (CrowGit package) — 30/30 pass, including a new `discoverReposSkipsCrowReviewsClones` test and the existing `summarizeCommits` suite (exercises the `--all` fallback for remote-less local repos).

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -173,6 +173,10 @@ public final class AppState {
     public var lastSummary: [RepoCommitSummary] = []
     public var isLoadingSummary: Bool = false
 
+    /// The configured Changes-summary scope (from config.defaults.summaryRepos),
+    /// surfaced in the board's repo dropdown. Synced from AppConfig.
+    public var summaryRepoScope: [String] = []
+
     /// LLM narrative of the current digest (item 4). Transient.
     public var llmNarrative: String = ""
     public var isSummarizingLLM: Bool = false
--- a/Packages/CrowGit/Sources/CrowGit/GitManager.swift
+++ b/Packages/CrowGit/Sources/CrowGit/GitManager.swift
@@ -29,6 +29,10 @@ public actor GitManager {
             var isDir: ObjCBool = false
             guard fm.fileExists(atPath: wsPath, isDirectory: &isDir), isDir.boolValue else { continue }
             guard !config.defaults.excludeDirs.contains(wsDir) else { continue }
+            // crow-reviews holds transient PR-review clones (full clones that
+            // resolve to the same org/repo slug as the canonical checkout); never
+            // discover them or they'd duplicate the real repo in the summary scan.
+            guard wsDir != "crow-reviews" else { continue }
 
             let wsConfig = config.workspaces[wsDir]
             let provider = wsConfig?.provider ?? config.defaults.provider
@@ -242,9 +246,14 @@ public actor GitManager {
     ///   discovered repo. Bare names are ambiguous across orgs, so the scope key
     ///   is the full slug; a repo with no parseable remote can't match a scoped
     ///   set and is dropped.
-    /// - `--all` picks up commits on every branch (worktree feature branches
-    ///   share the main checkout's object store), `--no-merges` keeps diffstats
-    ///   from double-counting merge commits.
+    /// - Before scanning each repo we `git fetch --no-tags origin <defaultBranch>`
+    ///   and base the digest on `origin/<defaultBranch>` (the latest remote
+    ///   default branch) so it reflects what landed on the remote regardless of
+    ///   local state. The default branch is resolved from `origin/HEAD`; if that
+    ///   can't be resolved we fall back to `--all` (commits on every branch).
+    ///   `--no-merges` keeps diffstats from double-counting merge commits.
+    /// - Repos sharing an `org/repo` slug are deduped (first seen wins) so each
+    ///   configured repo appears at most once.
     /// - Repos with zero commits in the window are omitted; a repo whose
     ///   `git log` errors (empty/detached) is skipped rather than failing the
     ///   whole scan.
@@ -257,6 +266,7 @@ public actor GitManager {
         // since git hosts treat org/repo case-insensitively. nil = include all.
         let wantedSlugs = includeRepos.map { Set($0.map { $0.lowercased() }) }
         var summaries: [RepoCommitSummary] = []
+        var seenSlugs = Set<String>()
         for repo in repos {
             // Derive org/repo + the hosted-commit-page prefix from the remote.
             // When scoping, skip repos that don't match (or have no remote).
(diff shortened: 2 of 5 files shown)


dhilgaertner (approved):
## Code & Security Review

Tight, well-scoped follow-up to #346 — all five rough edges are addressed, the diff is small (73/-6 across 5 files), and `swift test` in CrowGit passes 30/30.

### Security Review

**Strengths:**
- No new attack surface. `git fetch --no-tags origin <defaultBranch>` uses a branch name resolved from `git symbolic-ref refs/remotes/origin/HEAD` (parsed off the local refs DB) — not from user input — so the new git arg isn't reachable from untrusted strings.
- The `--no-tags` flag is a sensible hardening: it skips fetching tag refs that could otherwise be force-pushed by a remote.
- The fetch is gated on `defaultBranch` being non-nil, so a parse failure can't synthesize a partial argv.
- `try?` swallowing the fetch keeps an offline machine functional (best-effort, falls through to the local `origin/<default>` ref).

**Concerns:**
- None that block merge.

### Code Quality

**Yellow — worth considering, not blocking:**

1. **`"crow-reviews"` is now a magic string in three call sites** (`GitManager.swift:35`, `Scaffolder.swift:32`, `SessionService.swift:1086`, and the test). If the directory name ever changes, the discovery skip will silently drift out of sync. A shared `CrowCore` constant (e.g. `ReviewsWorkspace.dirName`) would centralize it. (`Packages/CrowGit/Sources/CrowGit/GitManager.swift:35`)

2. **Dedupe winner is non-deterministic.** `discoverRepos` iterates `FileManager.contentsOfDirectory`, whose order isn't guaranteed; "first seen wins" in `seenSlugs` therefore picks an arbitrary clone when the same `org/repo` exists in two workspaces. In practice this is fine because both clones produce the same `origin/<default>` log, but the comment ("first seen wins") slightly overstates the determinism. Consider either sorting `repos` for stable ordering or noting the FS-order caveat. (`Packages/CrowGit/Sources/CrowGit/GitManager.swift:269-280`)

3. **Per-repo `git fetch` runs sequentially on the actor.** Each in-scope repo now does a synchronous network round-trip. For users with many scoped repos on a slow link, this lengthens digest generation noticeably and there's no UI signal that fetching is happening (the existing `isLoadingSummary` covers it, but per-repo progress would be friendlier). Not a regression — just a UX consideration as scope grows. (`Packages/CrowGit/Sources/CrowGit/GitManager.swift:283-286`)

4. **Menu footer is informational text, not actionable.** `"Edit in Settings → General → Changes Summary"` reads like a link but isn't a `Button`. Users may try to click it. A no-op `Text` in a menu is acceptable, but wiring it to `appState.onShowSettings?()` (already plumbed through AppState) would close the loop. (`Packages/CrowUI/Sources/CrowUI/SummaryBoardView.swift:85-86`)

5. **Scope counter shows `0 repos` with a filter icon when unconfigured.** The dropdown explains it ("No repos configured"), but the collapsed label is a bit cryptic on first encounter. Consider a different label when the count is zero. (`Packages/CrowUI/Sources/CrowUI/SummaryBoardView.swift:90`)

**Green — observations:**
- `appState.summaryRepoScope` is correctly synced at all three entry points: initial load (`AppDelegate.swift:320`), config reload (`AppDelegate.swift:920`), and the live `saveSettings` path triggered from the SettingsView `.onChange`. No stale-state risk.
- `@Observable` AppState ensures the dropdown updates automatically when settings are saved.
- New test `discoverReposSkipsCrowReviewsClones` is correctly placed alongside the existing `discoverReposSkipsExcludedDirs` test and asserts the right behavior.
- The fallback path (no `origin/HEAD` → `--all`) is already covered by the existing `summarizeCommits` integration tests (the test harness builds a repo with no remote, then expects `--all` to pick up `sidebranch`).

### Summary Table

| Priority | Issue |
|----------|-------|
| Red | None |
| Yellow | Magic string "crow-reviews" duplicated across 3 files; dedupe ordering depends on FS order; per-repo fetch is sequential blocking |
| Green | Menu footer could be actionable; "0 repos" label is slightly cryptic |

**Recommendation:** **Approve.** The fixes are correct, well-documented, and tested. The yellow items are polish — none warrant blocking the merge.

---

PR #353, opened 2026-05-23
Persist tmux backend server across Crow restarts (#330)

### LINKED ISSUE #330: Persist tmux backend server across Crow app restarts (sessions survive quit/relaunch, not reboot)
## Summary

When Crow quits and relaunches, the embedded tmux server is torn down and a fresh one is
started, so all terminal sessions are lost. It would be much better if the tmux backend
server **outlived the app process**: quitting Crow and reopening it should re-attach to the
still-running sessions instead of starting from scratch.

Surviving a full **reboot** is explicitly out of scope (see "Non-goals") — the goal is
persistence across an app quit/relaunch cycle only.

## Current behavior (why sessions die on restart)

Three mechanisms in the current tmux backend each independently prevent cross-restart
persistence:

1. **The server socket is keyed on the CrowApp PID.**
   `Sources/Crow/App/AppDelegate.swift` (~L282–285) builds the socket as
   `$TMPDIR/crow-tmux-<pid>.sock` from `ProcessInfo.processInfo.processIdentifier`. A
   relaunched app has a new PID, hence a new socket path, hence a brand-new server — there is
   no path back to a prior server even if it were still alive.

2. **Clean quit kills the server.**
   On ⌘Q, `TmuxBackend.shutdown()` runs `tmux kill-server` (see the header note in
   `Packages/CrowTerminal/Sources/CrowTerminal/TmuxOrphanReaper.swift` and the "v1 of the
   rollout kills the…" comment in `AppDelegate.swift`). `TmuxController.killServer()`
   (`TmuxController.swift` ~L76) is the underlying `kill-server`.

3. **The orphan reaper kills survivors on next launch.**
   `TmuxOrphanReaper.reap(...)` runs at launch before the new server is configured
   (`AppDelegate.swift` ~L271). It enumerates `$TMPDIR/crow-tmux-*.sock` and, for any socket
   whose encoded PID is not a live `CrowApp`, runs `kill-server` and unlinks the socket
   (`TmuxOrphanReaper.swift` L30–57). So a server that happened to survive an ungraceful exit
   is intentionally reaped on the next start.

This design was a deliberate "one server per app instance, clean up leaks" choice (the reaper
references PR #229). This issue proposes shifting to a persistent shared server.

## Proposed behavior

- Quitting Crow leaves the tmux server (and all its sessions/windows) running.
- Relaunching Crow re-attaches to that server and rebinds the existing sessions back into the
  Crow UI, rather than spawning a fresh server.
- A reboot still clears everything (acceptable — see Non-goals).

## Suggested approach (not prescriptive)

1. **Use a stable socket path** instead of a PID-keyed one — e.g. `$TMPDIR/crow-tmux.sock`
   (keeps the reboot-clears-it property the user expects) or a fixed path under
   `~/.local/share/crow/`. This is the load-bearing change: it lets a relaunched app find and
   attach to the prior server. (`AppDelegate.swift` ~L282–285.)
2. **Stop killing the server on clean quit.** `TmuxBackend.shutdown()` should detach the
   client / leave the server running rather than `kill-server`.
3. **Rework launch-time reconciliation.** With a stable socket, `TmuxOrphanReaper`'s
   PID-keyed heuristic no longer applies. On launch, if a healthy server already exists at the
   stable socket, *adopt* it (re-attach to its sessions) instead of reaping. Retain a
   best-effort cleanup path for genuinely stale/abandoned state (e.g. guarded by a single
   advisory lock so concurrent Crow instances don't fight, and/or an explicit
   "reset terminal backend" affordance).
4. **Restore the session ↔ tmux-session/window mapping on launch.** Either persist the mapping
   or rediscover live tmux sessions and rebind them to Crow sessions in the UI. Relevant code:
   `TmuxBackend` binding logic (`socketPath:sessionName` bindings, `cockpitSessionName`),
   `TmuxController.newSession`/`newWindow`, and `TmuxDiscovery` (binary discovery only today).
5. **Concurrency / multi-instance.** A shared persistent server makes the "two Crow instances"
   case real. Define ownership (single-writer lock, or first-launcher-owns) so a second
   instance doesn't reap or duplicate the shared server.

## Acceptance criteria

1. Quit Crow with ≥1 active terminal session, relaunch → the same sessions are present and
   attached, with scrollback intact (within the configured `history-limit`).
2. The tmux server is no longer `kill-server`'d on clean quit.
3. The launch-time reaper no longer destroys a healthy server that belongs to the just-quit
   instance.
4. A reboot results in a clean slate with no errors (sessions gone, as expected).
5. Two concurrent Crow instances do not corrupt or reap each other's backend.
6. Tests updated to cover attach-to-existing-server and the new reconciliation path.

## Non-goals

- **Surviving reboots.** The socket lives in `$TMPDIR` (`/var/folders/.../T/`), cleared by
  macOS on reboot, and the tmux process dies on reboot regardless. Persisting across reboots
  (e.g. tmux-resurrect-style state capture) is explicitly out of scope here.

## Relevant files

- `Sources/Crow/App/AppDelegate.swift` — socket construction (~L282–285) and reaper invocation
  (~L271).
- `Packages/CrowTerminal/Sources/CrowTerminal/TmuxOrphanReaper.swift` — launch-time reaping.
- `Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift` — server configure/start,
  `shutdown()`, session bindings, `cockpitSessionName`.
- `Packages/CrowTerminal/Sources/CrowTerminal/TmuxController.swift` — `kill-server`,
  `new-session`, raw `tmux` invocation.
- `Sources/Crow/App/TmuxDiscovery.swift` — tmux binary discovery (would be the natural home
  for "discover existing sessions" if rediscovery is chosen over persistence).

dgershman:
Closes #330

## What & why

When Crow quit and relaunched, the embedded tmux server was torn down and a fresh one started, so every terminal session (scrollback + running Claude process) was lost. Three mechanisms each prevented persistence: a PID-keyed socket, `tmux kill-server` on clean quit, and the launch-time orphan reaper.

This makes the tmux server **outlive the app process**: quitting leaves it running, and relaunching re-attaches and rebinds the existing sessions into the UI. A reboot still clears everything (the socket lives in `$TMPDIR`, wiped by macOS on reboot) — surviving a reboot is an explicit non-goal.

## Changes

- **Stable socket** — `$TMPDIR/crow-tmux.sock` (no PID) so a relaunched process can find the prior server (`AppDelegate.swift`).
- **`TmuxBackend.shutdown(killServer:)`** — a clean quit (`applicationWillTerminate`) detaches and leaves the server *and its sentinel files* alive; the crash-watchdog's "Restart tmux server" still tears it down via the default.
- **Adopt on relaunch** — `SessionService.rehydrateTerminalSurface` now calls `adoptTerminal` for a persisted binding instead of always creating a fresh window. It adopts with `trackReadiness: false` and marks the terminal `.claudeLaunched` / drops it from `autoLaunchTerminals`, so the readiness→`launchClaude` path can't paste a **second** `claude --continue` into a pane where Claude is already running. Falls back to `registerTerminal` when the binding is stale (post-reboot, closed window, or a legacy PID socket).
- **Single instance per user** — an `flock` advisory lock (`$TMPDIR/crow-instance.lock`) acquired at launch; a second Crow launch shows an "already running" alert and exits, making the sole instance the unambiguous owner of the shared persistent server (acceptance criterion 5).
- **Reaper** — doc-only: its `crow-tmux-<pid>.sock` regex never matches the stable `crow-tmux.sock`, so a healthy server is never reaped; it now only cleans up legacy per-PID orphans.

## Tests

New `TmuxBackend` integration tests: attach-to-existing-server after a non-killing shutdown, sentinel kept-on-detach / removed-on-kill, kill-server tears the server down, and adopt-with-`trackReadiness:false` does not fire `.shellReady`. All CrowTerminal + 135 root tests pass (`arch -arm64 swift test`).

## Known edge case

If a persisted terminal's Claude process exited *while Crow was quit*, adopt re-binds a window whose child is now a dead shell (adopt validates the window index is live, not what's running in it). Acceptable for v1 — the user can restart claude in-pane.

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -115,33 +115,52 @@ public final class TmuxBackend {
 
     // MARK: - Lifecycle
 
-    /// Whether the cockpit session has been started this app launch.
+    /// Whether the cockpit session is live. Note this may be true on a fresh
+    /// app launch (before this process has created anything) when a prior Crow
+    /// quit left the server running at the stable socket — see #330.
     public var isRunning: Bool { controller?.hasSession() ?? false }
 
-    /// Tear down the tmux server (used by the crash-watchdog in PROD #5,
-    /// and by app quit). Resets internal state.
-    public func shutdown() {
+    /// Detach this Crow process from the tmux backend, resetting in-memory
+    /// state. Used by app quit and by the crash-watchdog (PROD #5).
+    ///
+    /// `killServer` controls whether the underlying tmux server is torn down:
+    ///   - `false` (clean app quit, #330): leave the server — and all its
+    ///     sessions/windows — running so the next launch can re-attach via
+    ///     `adoptTerminal`. The sentinel and wrapper-log files are *kept* on
+    ///     disk for the same reason: `adoptTerminal` re-fires `.shellReady`
+    ///     off the surviving sentinel.
+    ///   - `true` (default — crash-watchdog "Restart tmux server"): run
+    ///     `kill-server` and unlink the per-terminal scratch files.
+    public func shutdown(killServer: Bool = true) {
         if controller != nil {
-            NSLog("[CrowTelemetry tmux:server_shutdown bindings=\(bindings.count)]")
+            NSLog("[CrowTelemetry tmux:\(killServer ? "server_killed" : "server_detach") bindings=\(bindings.count)]")
         }
-        controller?.killServer()
+        if killServer {
+            controller?.killServer()
+        }
         sharedSurface?.destroy()
         sharedSurface = nil
         controller = nil
         bindings.removeAll()
         activeTerminalID = nil
         // Cancel any in-flight readiness watches so they don't keep polling
-        // (sentinel files are about to be unlinked) after server teardown.
-        // Mirrors the `destroyTerminal` cleanup (#282).
+        // after we let go of the backend. Mirrors the `destroyTerminal`
+        // cleanup (#282).
         for tasks in readinessTasks.values { tasks.forEach { $0.cancel() } }
         readinessTasks.removeAll()
-        for path in sentinels.values {
-            try? FileManager.default.removeItem(atPath: path)
+        // Only unlink the per-terminal scratch files when we're actually
+        // killing the server. On a clean quit that leaves the server running
+        // they must survive so the next launch's `adoptTerminal` can detect
+        // the already-ready shell from the existing sentinel (#330).
+        if killServer {
+            for path in sentinels.values {
+                try? FileManager.default.removeItem(atPath: path)
+            }
+            for path in wrapperLogs.values {
+                try? FileManager.default.removeItem(atPath: path)
+            }
         }
         sentinels.removeAll()
-        for path in wrapperLogs.values {
-            try? FileManager.default.removeItem(atPath: path)
-        }
         wrapperLogs.removeAll()
     }
 
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -104,8 +104,11 @@ public final class TmuxBackend {
     /// Must be called before any other method.
     public private(set) var tmuxBinary: String = ""
 
-    /// Persistent socket path — Crow uses one explicit socket per app
-    /// instance so it never collides with a user's own tmux.
+    /// Persistent socket path. Crow uses one explicit, per-user socket
+    /// (`$TMPDIR/crow-tmux.sock`) so it never collides with a user's own tmux.
+    /// Since #330 it is stable across app instances: the server outlives a
+    /// clean quit and a relaunch re-attaches to it (single-instance guard in
+    /// AppDelegate guarantees only one owner).
     public private(set) var socketPath: String = ""
 
     public func configure(tmuxBinary: String, socketPath: String) {
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxOrphanReaper.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxOrphanReaper.swift
@@ -1,25 +1,29 @@
 import Foundation
 
-/// Reaps orphaned tmux servers left behind by previous Crow instances that
-/// exited ungracefully (Force Quit, crash, SIGKILL — anything that bypasses
-/// `applicationWillTerminate` and therefore the shutdown fix from PR #229).
-/// Called once at app launch, BEFORE the new instance configures its own
-/// tmux server.
+/// Reaps orphaned, *legacy PID-keyed* tmux sockets left behind by pre-#330
+/// Crow builds. Called once at app launch, BEFORE the new instance configures
+/// its tmux server.
 ///
-/// Each Crow instance creates a socket at `$TMPDIR/crow-tmux-<pid>.sock`.
-/// On a clean ⌘Q, `TmuxBackend.shutdown()` runs and reaps the server. On
-/// an ungraceful exit the server lives on as an orphan, accumulating ~10-20
-/// MB of RSS per stuck instance and a stale socket file each. Without this
-/// reaper, dev iteration that involves Force Quit (or `pkill`) leaks one
-/// tmux server per cycle.
+/// Since #330 the running instance uses a single stable socket
+/// (`$TMPDIR/crow-tmux.sock`) that deliberately *outlives* the app process so
+/// a relaunch can re-attach to the still-running sessions. That stable socket
+/// is intentionally NEVER reaped — the regex below only matches the old
+/// `$TMPDIR/crow-tmux-<pid>.sock` naming, so `crow-tmux.sock` (no PID) falls
+/// through untouched, and a healthy persistent server is preserved.
+///
+/// What's left for this reaper is one-time cleanup of orphans from the old
+/// per-PID design: a `crow-tmux-<pid>.sock` whose owning CrowApp is gone leaks
+/// ~10-20 MB of RSS plus a stale socket file. (These self-heal on the app side
+/// too — a persisted binding pointing at a dead PID socket mismatches the
+/// stable socket and falls back to a fresh window — but reaping reclaims the
+/// leaked server.)
 ///
 /// The reaper is idempotent and best-effort. It enumerates
-/// `$TMPDIR/crow-tmux-*.sock`, extracts the PID encoded in each filename,
+/// `$TMPDIR/crow-tmux-<pid>.sock`, extracts the PID encoded in each filename,
 /// and:
 ///   - Skips the current process's own socket (we're about to bind it).
 ///   - Skips sockets whose PID is still bound to a live `CrowApp` process
-///     (the rare "two Crow instances running concurrently" case — must not
-///     reap a peer's server out from under it).
+///     (defensive — must not reap a peer's server out from under it).
 ///   - Otherwise: runs `tmux -S <socket> kill-server` (no-op if already
 ///     dead), then unlinks the socket file.
 public enum TmuxOrphanReaper {
--- a/Packages/CrowTerminal/Tests/CrowTerminalTests/TmuxBackendTests.swift
+++ b/Packages/CrowTerminal/Tests/CrowTerminalTests/TmuxBackendTests.swift
@@ -12,12 +12,41 @@ import Testing
 struct TmuxBackendTests {
 
     private func makeBackend() -> TmuxBackend {
-        let id = UUID().uuidString.prefix(8).lowercased()
-        let socket = (NSTemporaryDirectory() as NSString)
-            .appendingPathComponent("crow-test-backend-\(id).sock")
+        makeBackend(socket: sharedSocketPath())
+    }
+
+    /// Configure a backend on a caller-chosen socket. The persistence tests
+    /// (#330) need two distinct backend instances pointed at the *same* socket
+    /// to model a quit/relaunch where the server outlived the app.
+    private func makeBackend(socket: String) -> TmuxBackend {
         let backend = TmuxBackend()
         backend.configure(tmuxBinary: discoveredTmuxBinary!, socketPath: socket)
         return backend
+    }
+
+    private func sharedSocketPath() -> String {
+        let id = UUID().uuidString.prefix(8).lowercased()
+        return (NSTemporaryDirectory() as NSString)
+            .appendingPathComponent("crow-test-backend-\(id).sock")
+    }
+
+    /// Mirrors `TmuxBackend.sentinelPath(for:)` (which is private). Kept in sync
+    /// by the sentinel-survival test — if the format drifts, that test's
+    /// killServer assertion fails loudly.
+    private func sentinelPath(for id: UUID) -> String {
+        let dir = ProcessInfo.processInfo.environment["TMPDIR"] ?? "/tmp/"
+        return (dir as NSString)
+            .appendingPathComponent("crow-ready-\(id.uuidString).sentinel")
+    }
+
+    /// Best-effort kill of a server left running by a `killServer: false`
+    /// shutdown in a test, so we don't leak tmux servers across the suite.
+    private func killLeftoverServer(socket: String) {
+        TmuxController(
+            tmuxBinary: discoveredTmuxBinary!,
+            socketPath: socket,
+            sessionName: TmuxBackend.cockpitSessionName
+        ).killServer()
     }
 
     @Test func registerTerminalCreatesWindowAndBinding() throws {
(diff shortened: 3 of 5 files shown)


dhilgaertner (dismissed):
## Code & Security Review

Nicely-scoped change with thoughtful primary/fallback paths and a strong test suite. One real defensive concern worth fixing before merge (lock fd inheritance) plus a stale comment.

### Security Review

**Strengths:**
- Lock-file path is `$TMPDIR/...` (per-user `/var/folders/...` on macOS) — not a world-writable location, so other users on the host can't squat the lock or race-create the file. `open(..., 0o600)` also caps perms correctly.
- `flock(LOCK_EX | LOCK_NB)` is released by the kernel on process death, so a crash doesn't strand the next launch (subject to the inheritance concern below).
- The orphan reaper's regex (`^crow-tmux-(\d+)\.sock$`) deliberately excludes the new stable `crow-tmux.sock`, eliminating the obvious self-reap bug.
- `processIsCrowApp` keeps the legacy-reaper PID-reuse safety check; nothing tightened that we shouldn't have.
- No new attack surface (no new network listeners, no new IPC paths, no new file writes outside `$TMPDIR`).

**Concerns:**
- None blocking.

### Code Quality

**Yellow — should fix:**

1. **`AppDelegate.swift:97` — open the lock fd with `O_CLOEXEC`.** As written, the lock fd is inheritable by every subprocess Crow spawns. The whole point of #330 is to make the tmux daemon outlive Crow, and `Process` (NSTask) on macOS uses `posix_spawn` with default fd inheritance — so the tmux daemon process inherits this fd. POSIX `flock` semantics: the lock is released only when **all** descriptors referring to the open-file-description are closed, *across all processes*. If a tmux daemon (or any other long-lived child) inherits the lock fd and never closes it, Crow will fail to acquire the lock on relaunch even though no Crow process is alive — exactly the failure mode this PR is supposed to enable recovery from. In practice tmux 3.3+ does `closefrom()` after its daemon fork, which saves us — but relying on a subprocess's internal cleanup behavior for the correctness of a primary feature is fragile, and the fix is a one-token change:
   ```swift
   let fd = open(lockPath, O_CREAT | O_RDWR | O_CLOEXEC, 0o600)
   ```
   With `O_CLOEXEC`, the lock fd is closed automatically on `execve` in every child, regardless of what the child does. This is the standard pattern for advisory lock fds held for process lifetime.

2. **`SessionService.swift:144-147` — stale comment.** The block comment above the rehydrate loop still says:
   > "Re-hydrate every persisted terminal by re-registering it with a freshly-started tmux server (v1 doesn't keep the server alive across launches; spec §12)."
   
   That's exactly the assumption this PR overturns. Worth updating to reflect the new adopt-first / register-as-fallback flow, since this is the canonical comment a future reader will find when they grep for the hydration entry point.

**Green — consider:**

3. **`TmuxBackend.swift:107-109`** — `socketPath`'s docstring says "Crow uses one explicit socket per app instance so it never collides with a user's own tmux." Post-#330 it's effectively per-*user* (and persists across app instances). Small wording tweak.

4. **`AppDelegate.swift:1693-1696`** — releasing the lock at the very end of `applicationWillTerminate` (after `GhosttyApp.shared.shutdown()`) means if Ghostty shutdown is slow, the next launch waits longer than necessary before it can grab the lock. Since the comment says "flock would release on exit anyway; explicit is clearer," consider releasing it earlier (right after `socketServer?.stop()`) to minimize the "Crow is already running" window during a fast quit→relaunch.

5. **`TmuxBackendTests.swift:313-340`** — `sentinelKeptOnDetachRemovedOnKill` reuses the same `id` across the two halves and depends on `registerTerminal`'s implicit sentinel-wipe to clean up between them. It works, but a fresh UUID per half would make the intent clearer and remove the cross-half coupling.

6. **`presentAlreadyRunningAlert`** doesn't surface the lock path. If the user hits this in the (rare) case where the lock is actually stale, they'll have no idea what to remove. Including the lock-file path in the informative text — "If you see this and no Crow is running, remove `$TMPDIR/crow-instance.lock`" — would save a support round-trip. Nit.

7. **`acquireSingleInstanceLock` "fail-open" branch** — if `open()` itself fails, two Crow instances are silently allowed (the very thing the lock prevents). Defensible (better than blocking launch entirely), but worth a more visible signal than NSLog if it ever fires.

### Summary Table

| Priority | Issue |
|----------|-------|
| Red | — |
| Yellow | Lock fd opened without `O_CLOEXEC` — risk of inheritance by long-lived subprocesses (notably the persistent tmux daemon) holding the flock and blocking relaunch (`AppDelegate.swift:97`). |
| Yellow | Stale comment in `SessionService.hydrateState` still says v1 doesn't persist the tmux server (`SessionService.swift:144-147`). |
| Green | `socketPath` docstring out of date (per-app → per-user). |
| Green | Release `instanceLockFD` earlier in `applicationWillTerminate`. |
| Green | Test reuses UUID across halves; alert doesn't reveal lock path; fail-open errno path silent. |

**Recommendation:** Request Changes — the `O_CLOEXEC` fix is the only meaningful behavior concern, and it's a one-character defensive change in a code path whose correctness is load-bearing for the entire feature. The stale comment is doc-only but worth fixing in the same pass since it sits directly above the new logic.

---

dgershman:
Thanks for the thorough review — addressed everything in 11dadbf:

**Yellow**
- **`O_CLOEXEC` on the lock fd** ✅ — `open(lockPath, O_CREAT | O_RDWR | O_CLOEXEC, 0o600)`. Agreed this is load-bearing: the whole feature relies on the tmux daemon outliving Crow, and an inherited+held flock would block re-acquisition on relaunch. Now closed-on-exec unconditionally rather than relying on tmux's `closefrom()`.
- **Stale `hydrateState` comment** ✅ — rewritten to describe the adopt-first / register-as-fallback flow.

**Green**
- **`socketPath` docstring** ✅ — updated (per-app → stable per-user, persists across instances).
- **Release lock earlier** ✅ — moved right after `socketServer?.stop()` to minimize the spurious "already running" window on a fast quit→relaunch.
- **Test UUID coupling** ✅ — `sentinelKeptOnDetachRemovedOnKill` now uses a fresh UUID + socket per half.
- **Alert lock path** ✅ — the "already running" alert now prints the lock-file path with a "remove it if no Crow is running" hint.
- **Fail-open errno path** ✅ — now logged as `[CrowTelemetry instance_lock:open_failed …]` so it's captured rather than a plain NSLog.

Full app build + CrowTerminal/root test suites green (the one `retryReadinessEmitsTimedOutWhenSentinelMissing` failure is a pre-existing timing flake under parallel load — passes in isolation, untouched by this PR).

dhilgaertner (approved):
## Code & Security Review

Solid, carefully-scoped PR. Three independent persistence blockers are unwound (PID socket → stable socket; `kill-server` on quit → `kill-server` only on watchdog; reaper regex → unchanged, so it deliberately ignores the new path), and the relaunch path is wired through `adoptTerminal(trackReadiness: false)` so the readiness pipeline can't re-fire `claude --continue` into a live pane. The single-instance lock closes the obvious "two owners of one persistent server" hole. Tests cover the four interesting state transitions (survive-detach, kill teardown, sentinel kept-vs-removed, adopt-without-readiness). No critical issues.

### Security Review

**Strengths:**
- `open(O_CLOEXEC, 0o600)` — the lock fd is correctly NOT inherited across `posix_spawn` into the tmux daemon (this is exactly the foot-gun where the comment at `AppDelegate.swift:97-103` is doing real work). Mode `0600` keeps the lock file owner-only.
- `flock(LOCK_EX | LOCK_NB)` releases automatically on process death — a crashed Crow doesn't strand the lock.
- Lock path lives in `$TMPDIR`, which is per-user on macOS — no cross-user/cross-sandbox spoofing.
- No new shell/subprocess interpolation surface; binding adoption uses tmux's own session/window addressing.

**Concerns:**
- **Lock file is unlink-safe** (`AppDelegate.swift:104`). `flock` is held on the open file description, so a `rm $TMPDIR/crow-instance.lock` (which the \"already running\" alert literally instructs the user to do) lets a second concurrent Crow `open()` a fresh inode and acquire its own lock — both instances would then think they own the persistent tmux server. Mitigations to consider in a follow-up: write the holder's PID into the file and have the loser verify the PID is alive before showing the \"stale lock\" copy, or only suggest the manual rm when `processIsCrowApp(pid)` returns false.

### Code Quality

- `TmuxBackend.shutdown(killServer:)` — clean split, the docstring at lines 126-136 makes the contract obvious. The sentinel preservation pairs correctly with `adoptTerminal`'s `fileExists(atPath: sentinelPath)` check (`TmuxBackend.swift:264`).
- `SessionService.rehydrateTerminalSurface` belt-and-suspenders (`autoLaunchTerminals.remove` + `terminalReadiness = .claudeLaunched`, lines 217-220) is the right defense-in-depth against the second-`claude --continue` paste. Worth keeping the comment around since the invariant is non-obvious.
- **`TmuxOrphanReaper` comment is slightly stale** (`TmuxOrphanReaper.swift:25-26`): with the single-instance lock in place there's no legitimate \"two Crow instances running concurrently\" case anymore — the `processIsCrowApp` check is now only defending against the early-exit path of a lock-denied second instance (which never configures a backend). Worth updating the rationale so future readers don't think the reaper is part of the concurrency story.
- **Manager command rebuild silently dropped on adopt**. `SessionService.hydrateState` rebuilds the Manager `command` from current `remoteControlEnabled` / `managerAutoPermissionMode` (lines 100-113), but `rehydrateTerminalSurface` returns the terminal unchanged when adoption succeeds (line 221) — the running Claude keeps the OLD flags from before the user toggled settings. Documented edge case worth one line of comment at the adopt branch, and probably a follow-up to either flush an in-pane restart prompt or detect the divergence.
- **Known dead-shell case** (acknowledged in PR description): `terminalReadiness` is force-set to `.claudeLaunched` even if the adopted window's child shell died while Crow was quit. Users will see a \"running\" indicator on a dead pane. Acceptable for v1; a small follow-up could `display_message #{pane_dead}` during adopt to surface a banner.
- **No telemetry for lock-denied case** in `acquireSingleInstanceLock` — there's a log on `open_failed` (the fail-open path) but nothing on the `flock` denial. A `[CrowTelemetry instance_lock:already_held]` would help diagnose user reports of \"I see the alert but I don't think Crow is running.\"
- `lockPath` is recomputed in both `acquireSingleInstanceLock` and `presentAlreadyRunningAlert`. Trivial — fine as-is.
- Early `close(instanceLockFD)` in `applicationWillTerminate` (lines 1702-1705) is the right call for fast quit→relaunch and the comment explains why.

### Tests
- `serverSurvivesDetachAndIsAdoptedOnRelaunch`, `killServerShutdownTearsDownServer`, `sentinelKeptOnDetachRemovedOnKill`, `adoptWithoutReadinessDoesNotEmitShellReady` — good four-corner coverage of the new behaviors. Each test cleans up after itself (`killLeftoverServer` helper + per-test sockets) so no server leaks across the suite. Couldn't run them locally (`Frameworks/GhosttyKit.xcframework` is not in the worktree); trusting the PR's claim that all CrowTerminal + 135 root tests pass.

### Summary Table
| Priority | Issue |
|----------|-------|
| Yellow | Lock file is unlink-able — `rm` of `crow-instance.lock` lets two Crow instances acquire it (alert tells the user to do exactly this). Consider PID-in-file + liveness verification. |
| Yellow | Manager-terminal command rebuild silently lost when adopt succeeds — users who change `remoteControlEnabled` / `managerAutoPermissionMode` between launches will keep running the old flags. Worth a comment + follow-up. |
| Green | `TmuxOrphanReaper` rationale is stale post-lock — \"two concurrent Crow instances\" case no longer applies; update to reflect the new design. |
| Green | Add `[CrowTelemetry instance_lock:already_held]` log on `flock` denial. |
| Green | Adopted pane with dead child shell still reports `.claudeLaunched` — acknowledged in PR; small UX follow-up via `pane_dead` check. |

**Recommendation:** Approve. All concerns are non-blocking and can be addressed in follow-ups. The core persistence design (stable socket + detach-on-quit + adopt-on-relaunch + single-instance lock + readiness suppression on adopt) is correct, well-tested, and matches the PR description's claims.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review

**Strengths:**
- `acquireSingleInstanceLock` opens the lock file with `O_CLOEXEC` (mode `0600`), which correctly prevents fd inheritance into spawned subprocesses (notably the tmux daemon) — the fixup commit's stated motivation is right: without this, a subprocess could keep the open-file-description alive after Crow exits, blocking relaunch from re-acquiring the `flock`. (`AppDelegate.swift:104`)
- Lock lives under `$TMPDIR` (user-private on macOS GUI launches) and uses `LOCK_EX | LOCK_NB`, so a second launch never blocks — it fails fast and shows an alert.
- No new input boundaries are introduced; all new paths are local IPC / filesystem operations under user-owned scopes.
- The reaper regex (`^crow-tmux-(\d+)\.sock$`) deliberately does not match the new stable socket name (`crow-tmux.sock`), so a healthy persistent server cannot be reaped. The peer-PID check still defends against killing a concurrent CrowApp's socket.
- `shutdown(killServer:)` correctly cancels in-flight readiness tasks regardless of the kill flag, preventing leaked `MainActor` callbacks against stale ids.

**Concerns:**
- `/tmp` fallback when `$TMPDIR` is unset would put the lock at a host-shared path; two users on the same host could collide on the lock. Practically inert because launchd-launched GUI apps always have `$TMPDIR`, but worth a follow-up if anyone ever runs Crow.app from a non-GUI context. (`AppDelegate.swift:95`)

### Code Quality

**Correctness:**
- Adoption flow in `SessionService.rehydrateTerminalSurface` is well-thought-out: `trackReadiness: false` to `adoptTerminal`, plus dropping the terminal from `autoLaunchTerminals` and forcing `terminalReadiness = .claudeLaunched`, makes it structurally impossible for `launchClaude` to fire a second `claude --continue` against a pane where Claude is already running. (`SessionService.swift:205-225`)
- The fallback to `registerTerminal` on `adoptTerminal` failure (post-reboot / closed window / legacy PID socket) is the right safety net and matches the comment in the test `killServerShutdownTearsDownServer`.
- `ensureCockpitSession`'s adopt-on-race logic continues to behave correctly because `hasSession()` returns true for the persisted server on first call after relaunch — no behavior change there.

**Tests:**
- New `TmuxBackend` tests cover the four important transitions (attach-after-detach, kill-on-quit, sentinel survive-vs-removed, no-readiness-on-adopt). The fixup commit's per-half UUID/socket isolation in `sentinelKeptOnDetachRemovedOnKill` is the right cleanup — previously sharing state between halves would have been brittle.
- Could not run the suite locally (worktree is missing the `GhosttyKit.xcframework` binary), so I'm relying on the PR description's claim that all 135 tests pass.

**Minor nits (non-blocking):**
- `TmuxBackend.isRunning` doc (`TmuxBackend.swift:121-123`) says it "may be true on a fresh app launch before this process has created anything." But `isRunning` is `controller?.hasSession() ?? false`, and `controller` is nil until `ensureRunningServer` runs — so on a truly fresh launch it returns `false`. The comment reads as if it could be true without any backend interaction; suggest tightening to "…true after the first `ensureRunningServer()` call on a launch where a prior Crow quit left the server running."
- The "sentinels must survive so `adoptTerminal` can re-fire `.shellReady` off the surviving sentinel" rationale in `TmuxBackend.swift:131-134` and `TmuxBackend.swift:154-157` is currently defensive: every adopt path in this PR passes `trackReadiness: false`, so the on-disk sentinel is never consulted post-relaunch. The behavior is still correct (and worth preserving for future readiness-tracking adopts), but the comment could note that it's a forward-compatibility hedge rather than a load-bearing invariant — otherwise a future reader may try to "simplify" by removing the survival logic.
- Settings-toggle ergonomics: if the user changes `remoteControlEnabled` or `managerAutoPermissionMode` between launches, `hydrateState` rebuilds the manager's `command` to reflect the new prefs (`SessionService.swift:100`), but the *adopted* tmux window still hosts the old `claude` process. The rebuilt command only takes effect after a manual restart. Acceptable for v1 alongside the documented "Claude died while Crow was quit" edge case, but could be worth a brief doc note.
- Early `instanceLockFD` release in `applicationWillTerminate` (`AppDelegate.swift:1703-1710`) opens a small window where a relaunched instance can attach to the still-running tmux server while the old instance is still in `GhosttyApp.shared.shutdown()`. Across-process this is benign (libghostty + tmux both tolerate multiple clients), and the comment justifies the trade-off well; just calling it out.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | None |
| Yellow | `/tmp` fallback could allow cross-user lock collision in unusual configs (probably inert under launchd) |
| Green | `isRunning` doc comment is slightly misleading |
| Green | Sentinel-preservation comment overstates current necessity (it's defensive for future use) |
| Green | RC / auto-permission setting changes don't propagate to adopted manager terminal until restart |
| Green | Early lock-release vs Ghostty teardown opens a brief multi-instance overlap window (benign across processes) |

**Recommendation:** Approve. The design is sound, the security-relevant `O_CLOEXEC` detail is handled correctly, the tests cover the meaningful transitions, and the fallback paths recover gracefully from every failure mode I could enumerate. The remaining items are doc / UX polish, not correctness or security blockers.

---

PR #350, opened 2026-05-23
Rename non-primary Manager sessions from the UI

dgershman:
Closes #347

## Summary

Non-primary Manager sessions in the sidebar (the `extraManagers` rows rendered by `ManagerSessionRow`) previously had **no way to be renamed in-app** — they were auto-named "Manager 2", "Manager 3", … and the only way to rename them was the `crow rename-session` CLI. This adds a **Rename** affordance to those rows.

The rename backend already existed and was unrestricted (the `rename-session` RPC handler + CLI command rename any session by ID); this change is purely UI + callback plumbing.

## Changes

Mirrors the existing **terminal-tab rename** plumbing end-to-end:

- **`AppState.onRenameSession`** — new callback `((UUID, String) -> Void)?`, next to `onRenameTerminal`.
- **`SessionService.renameSession`** — validates via `Validation.isValidSessionName`, updates `appState.sessions`, and persists via `store.mutate`.
- **`AppDelegate`** — wires the callback alongside `onRenameTerminal`.
- **`SessionListView` / `ManagerSessionRow`** — adds a **"Rename"** button to the `extraManagers` context menu (above the destructive "Delete") and an inline `TextField` edit in the row, reusing the `TerminalTabBar` pattern. A lifted `editingManagerID` binding (mirroring the existing `sessionToDelete` state) bridges the parent context menu to the child row's edit state; commit on Enter/focus-loss, discard empty/whitespace names.

The primary "Manager" row (`ManagerSidebarRow`) is intentionally unchanged.

## Verification

- `make ghostty` + `swift build --arch arm64` → **Build complete.**
- Right-click a "Manager N" row → **Rename** → field becomes editable; Enter commits the new name.
- Empty/whitespace names are rejected (row keeps prior name).
- New name persists via the store (survives relaunch).
- Primary "Manager" row shows no rename affordance.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #362, opened 2026-05-23
Sync Remote Control name on session rename (#354)

### LINKED ISSUE #354: Renaming a session should sync the Remote Control name via /rename <name>
## Background

#347 (PR #350) added the ability to rename non-primary Manager sessions from the sidebar. However, the rename only updates Crow's own session name + persisted store — it does **not** update the name shown in claude.ai's **Remote Control** panel.

When a session is launched with remote control enabled, Claude Code is started with `--rc --name <sessionName>` (see `ClaudeLaunchArgs.argsSuffix`, `Packages/CrowCore/Sources/CrowCore/ClaudeLaunchArgs.swift`). That `--name` label is fixed **at launch time**, so after an in-app rename the Remote Control panel still shows the old name and the two get out of sync.

## Request

When a session is renamed (the `onRenameSession` path), Crow should also execute the Claude Code slash command `/rename <name>` in the session's running (managed) Claude Code terminal, so the Remote Control panel label updates to match the new session name.

## Implementation sketch

- Hook into the rename path: `SessionService.renameSession` / `AppState.onRenameSession` (added in #350).
- For the session's managed Claude Code terminal, send `/rename <name>\n` to the terminal (same mechanism as `crow send` → write text + Enter).
- Only do this when the session/terminal was launched with remote control active (`isRemoteControlActive(sessionID:)`); skip otherwise so we don't inject a slash command into a plain shell.
- Guard against no managed terminal / not-yet-running Claude.

## Acceptance criteria

- Renaming a remote-control session in the sidebar updates the name in claude.ai's Remote Control panel to match.
- Renaming a non-remote-control session does nothing extra (no stray `/rename` typed into a terminal).

dgershman:
Closes #354

## Problem

When a session is launched with remote control, Crow starts Claude Code with `--rc --name <sessionName>`. That `--name` label is fixed **at launch time**, so the in-app rename added in #350 (PR #350) only updated Crow's own session name + store — claude.ai's **Remote Control** panel kept showing the old name, and the two drifted out of sync.

## Change

`SessionService.renameSession` now pushes the new name into the running Claude Code by sending the `/rename <name>` slash command to any of the session's terminals that were launched with `--rc`.

Key detail: the target terminals are gated on membership in `remoteControlActiveTerminals` (the set of terminals actually launched with `--rc`), **not** the `isManaged` flag. Manager terminals run Claude with `--rc` but are created without `isManaged`, so gating on `isManaged` would miss the non-primary Manager sessions this ticket targets. A `TerminalRouter.canSend` guard fails soft when a terminal surface isn't ready, and the already-enforced session-name validation (no control characters) means the trailing newline is the only Enter keypress sent.

A pure, `nonisolated` `remoteControlRenameTargets` helper isolates the selection logic so it's unit-testable without a live app.

## Scope

Sidebar rename path only (`onRenameSession` → `SessionService.renameSession`). The `crow rename-session` CLI RPC is intentionally unchanged.

## Acceptance criteria

- ✅ Renaming a remote-control session in the sidebar updates the claude.ai Remote Control panel label.
- ✅ Renaming a non-remote-control session does nothing extra (no stray `/rename` typed into a terminal).

## Tests

- New `RemoteControlRenameTargetTests` (4 cases): selects the `--rc` terminal; returns empty when none is RC-active; selects an RC-active terminal even when not flagged `isManaged`; filters to only RC-active terminals.
- Full `CrowTests` suite passes (139 tests).

### DIFF
--- a/Sources/Crow/App/SessionService.swift
+++ b/Sources/Crow/App/SessionService.swift
@@ -1041,7 +1041,36 @@ final class SessionService {
                 data.sessions[i].name = name
             }
         }
+        syncRemoteControlName(sessionID: sessionID, newName: name)
         return true
+    }
+
+    /// Terminals to push a Remote-Control `/rename` into for a session: those
+    /// launched with `--rc` (tracked in `remoteControlActiveTerminals`). That's
+    /// the Claude Code instance whose claude.ai panel label is fixed at launch.
+    /// Manager terminals qualify even though they aren't flagged `isManaged`.
+    /// Pure/`nonisolated` so it can be unit-tested without a live app.
+    nonisolated static func remoteControlRenameTargets(
+        terminals: [SessionTerminal],
+        rcActiveTerminals: Set<UUID>
+    ) -> [SessionTerminal] {
+        terminals.filter { rcActiveTerminals.contains($0.id) }
+    }
+
+    /// After an in-app rename, push the new name to the running Claude Code via
+    /// its `/rename` slash command so claude.ai's Remote Control panel label
+    /// (fixed at launch via `--name`) stays in sync. No-op when the session has
+    /// no `--rc` terminal, or when a terminal's surface isn't ready to receive.
+    /// The name is already validated (no control characters) by the caller, so
+    /// the trailing newline is the only Enter keypress sent.
+    private func syncRemoteControlName(sessionID: UUID, newName: String) {
+        let targets = Self.remoteControlRenameTargets(
+            terminals: appState.terminals(for: sessionID),
+            rcActiveTerminals: appState.remoteControlActiveTerminals
+        )
+        for terminal in targets where TerminalRouter.canSend(terminal) {
+            TerminalRouter.send(terminal, text: "/rename \(newName)\n")
+        }
     }
 
     // MARK: - Global Terminal Management
--- a/Tests/CrowTests/RemoteControlRenameTargetTests.swift
+++ b/Tests/CrowTests/RemoteControlRenameTargetTests.swift
@@ -1,0 +1,59 @@
+import Foundation
+import Testing
+import CrowCore
+@testable import Crow
+
+/// Locks down which terminals receive a `/rename` slash command when a session
+/// is renamed (#354). The selection is driven by `remoteControlActiveTerminals`
+/// — the set of terminals launched with `--rc` — and NOT by `isManaged`, since
+/// Manager terminals are RC-active without carrying the `isManaged` flag.
+@Suite("SessionService.remoteControlRenameTargets")
+struct RemoteControlRenameTargetTests {
+
+    private func terminal(_ id: UUID, isManaged: Bool = false) -> SessionTerminal {
+        SessionTerminal(id: id, sessionID: UUID(), cwd: "/tmp", isManaged: isManaged)
+    }
+
+    @Test
+    func returnsRemoteControlTerminal() {
+        let rc = UUID()
+        let terminals = [terminal(rc)]
+        let targets = SessionService.remoteControlRenameTargets(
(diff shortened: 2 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review

**Strengths:**
- Session name is validated upstream via `Validation.isValidSessionName` (Packages/CrowCore/Sources/CrowCore/Validation.swift:16) — control characters (including `\n`, `\r`) are rejected, so the trailing `\n` in `"/rename \(newName)\n"` is the only Enter keypress sent. No risk of multi-line injection into the Claude Code prompt.
- Delivery goes through `TerminalRouter.send` → `TmuxBackend.sendText`, which uses the tmux load-buffer + paste-buffer route. No shell interpolation, so quotes / backticks / `;` / `$` in a name are inert.
- `TerminalRouter.canSend` guard fail-soft pattern matches the existing usage in `AutoRespondCoordinator` — no panic, no exception bubbling.
- All access (`renameSession`, `appState.terminals(for:)`, `appState.remoteControlActiveTerminals`, `TerminalRouter.canSend`/`send`) is on `@MainActor`, so the snapshot used to pick targets is consistent with the rename mutation that just happened.

**Concerns:**
- None blocking. The slash command lands in whatever input buffer Claude Code currently has; if the user has half-typed a prompt at rename time, the `/rename foo` will be appended and submitted alongside that text. This is the same caveat that already exists for `JobScheduler` (Sources/Crow/App/JobScheduler.swift:145) and `AutoRespondCoordinator` (Sources/Crow/App/AutoRespondCoordinator.swift:73, :109) — not a regression, and arguably acceptable for a rename action initiated explicitly by the user.

### Code Quality

- Selection logic is cleanly extracted into a pure, `nonisolated static` helper (`remoteControlRenameTargets`) — exactly the right shape for a unit test, no app fixture required. (Sources/Crow/App/SessionService.swift:1053)
- Gating on `remoteControlActiveTerminals` rather than `isManaged` is the correct call — confirmed by walking the `--rc` injection sites in `AppDelegate` (1211) and `SessionService` (110, 361, 487). Manager terminals are RC-active but not `isManaged`, and the test at `RemoteControlRenameTargetTests.swift:38` locks that invariant in.
- 4 test cases cover the meaningful slices: single RC terminal, no RC terminals, RC + non-managed, mixed plain+RC. Good coverage for a 5-line filter.
- Doc comments correctly explain the *why* (RC label fixed at launch via `--name`) and the *non-obvious* choice (not gating on `isManaged`) — earns its keep.

### Observations (non-blocking)

- **CLI/UI parity.** `crow rename-session` (Sources/Crow/App/AppDelegate.swift:1005) mutates `appState.sessions` + store directly and skips `SessionService.renameSession`, so the RC sync only fires from the sidebar. The PR description calls this out as intentional. It's a coherent scope choice for this ticket, but it does mean a future automation path renaming via CLI will silently drift the RC label — worth a follow-up ticket if the CLI rename ever becomes a primary surface.
- **No test for the `canSend` gate or the actual send call.** Acceptable — `TerminalRouter` is a thin facade — but if the slash-command-send pattern grows beyond rename, a fake `TerminalRouter` would be worth introducing once.

### Summary Table

| Priority | Issue |
|----------|-------|
| Red | None |
| Yellow | None |
| Green | CLI `rename-session` RPC bypasses RC sync — intentional per PR, but worth a follow-up if CLI renames become common. |
| Green | If more slash-command sends land in `SessionService`, introduce a `TerminalRouter` test seam. |

**Recommendation:** Approve. Tight, well-scoped fix; the testable seam (`remoteControlRenameTargets`) is the right abstraction, validation already neutralizes the only payload risk (newline injection), and the implementation reuses the established `canSend` fail-soft + paste-buffer send pattern.

---

PR #368, opened 2026-05-23
Persist & restore hook state so status colors survive relaunch (#367)

### LINKED ISSUE #367: Session status colors are stale after relaunch — tmux-persist adoption skips the hook-state refresh that re-`continue` used to trigger
## Summary

After #330 / #353 (persist tmux backend across restarts), the session **status
colors** in the sidebar are stale/blank for *all* items immediately after a
quit→relaunch. They no longer reflect Claude's real state (working / waiting /
done / needs-attention) until the user next interacts with Claude in that
session. Before #330 the colors were correct right after relaunch.

The user's observation is exact: "there was different behavior before when it
was continuing things — we probably need to do some kind of refresh or other
behavior to catch the UI up."

## Root cause

The status dot + row tint are computed from **in-memory hook state**, not from
anything persisted:

- `SessionListView.statusIndicator` (`Packages/CrowUI/Sources/CrowUI/SessionListView.swift:675-761`)
  switches on `session.status`, `terminalReadiness`, and the hook-derived
  `claudeState` / `pendingNotification` / `lastToolActivity`.
- Those last three come from `AppState.hookState(for:)`
  (`Packages/CrowCore/Sources/CrowCore/AppState.swift:234-243`), backed by an
  in-memory `[UUID: SessionHookState]`. It is **not** persisted (StoreData only
  holds sessions/worktrees/links/terminals —
  `Packages/CrowPersistence/.../JSONStore.swift:6-9`) and lazily returns a blank
  `SessionHookState` (`claudeState = .idle`) on first access
  (`AppState.swift:521-522`).
- `hookState` is repopulated only by live hook events arriving over the crow
  socket (`crow hook-event --session <UUID> --event …`, routed in
  `AppDelegate` ~`:1417-1620`).

So on every launch, all `hookState` is empty. **Pre-#330**, relaunch
re-registered each terminal and re-ran `claude --continue`, which fired a
`SessionStart(source: resume)` hook → repopulated `hookState` → colors snapped
to reality. **Post-#330**, `SessionService.rehydrateTerminalSurface` *adopts*
the still-running window with `trackReadiness: false` and sets
`terminalReadiness = .claudeLaunched` directly (by design, to avoid pasting a
second `claude --continue` into a live pane). Since Claude is already running
and idle, **no new hook event fires on relaunch**, so `hookState` stays blank
and the dot sits on a static default (solid green) regardless of true state.

This is a regression introduced by #330's adoption path: the side effect that
used to refresh the UI (the re-`continue`) was intentionally removed, and
nothing replaced the refresh.

## Secondary gaps in the adopt branch (same root, worth fixing together)

`rehydrateTerminalSurface`'s adopt branch
(`Sources/Crow/App/SessionService.swift` ~`:205-220`) also skips two side
effects that the normal launch path (`launchClaude`) performs:

1. **Hook config is not re-written** — `launchClaude` calls
   `HookConfigGenerator.writeHookConfig(worktreePath, sessionID, crowPath)`
   (`SessionService.swift` ~`:329`). On adoption it's skipped. The sessionID is
   stable and the socket is well-known, so the *previously written* config is
   usually still valid — but if it was removed/regenerated, the adopted Claude's
   hooks won't route back at all (colors permanently frozen).
2. **`remoteControlActiveTerminals` is not re-seeded** — so the RemoteControl
   badge disappears after relaunch for terminals that were launched with `--rc`
   (`launchClaude` ~`:392` does this; adopt does not).

## Proposed direction (needs design — not prescriptive)

The core need is a **launch-time catch-up/refresh** for adopted terminals so the
UI reflects reality without re-`continue`-ing. Options, roughly in order of
preference:

- **Persist + restore hook state.** Add `SessionHookState` (or at least
  `claudeState` / `pendingNotification` / `lastToolActivity`) to the persisted
  StoreData and rehydrate it on launch so colors are correct immediately, then
  let live hook events keep it fresh. Caveat: state can be slightly stale if it
  changed while the app was quit (Claude is idle then, so low risk).
- **Proactively derive/refresh state for adopted terminals on launch** — e.g.
  re-seed a sensible `claudeState` (idle/done) and trigger a lightweight refresh,
  rather than leaving the lazily-created blank state.
- Regardless of the above, in the adopt branch: **re-write hook config** and
  **re-seed `remoteControlActiveTerminals`** (when `remoteControlEnabled`) so
  adopted terminals match freshly-launched ones.

## Acceptance criteria

1. Quit Crow with ≥1 active session in a non-idle state (working / done /
   needs-attention), relaunch → the sidebar status colors reflect the true state
   without the user having to interact with Claude first.
2. The RemoteControlBadge reappears after relaunch for terminals launched with
   `--rc`.
3. Hook events from an adopted Claude continue to route to the correct session
   (hook config valid for adopted terminals).
4. No second `claude --continue` is pasted into a live pane (the #330 fix must
   not regress).

## Relevant files

- `Sources/Crow/App/SessionService.swift` — `rehydrateTerminalSurface` adopt
  branch (~L205-220), `launchClaude` side effects (~L314-405),
  `wireTerminalReadiness` (~L254-271).
- `Packages/CrowCore/Sources/CrowCore/AppState.swift` — `hookState(for:)` and
  the in-memory `_sessionState` (L234-243), `SessionHookState` (L521+).
- `Packages/CrowUI/Sources/CrowUI/SessionListView.swift` — `statusIndicator`
  (L675-761), `rowBackgroundColor` (L820-827).
- `Packages/CrowPersistence/Sources/CrowPersistence/JSONStore.swift` — StoreData
  (L6-9) — where persisted hook state would live.
- `Sources/Crow/App/HookConfigGenerator.swift` — `writeHookConfig` / `generateHooks`.
- `Sources/Crow/App/AppDelegate.swift` — `hook-event` RPC handler (~L1417-1620).

## Context

Introduced by #330 (PR #353). The adoption path correctly avoids re-launching
Claude, but the old re-launch was doubling as the mechanism that refreshed the
in-memory, non-persisted hook state on every boot.

dgershman:
Closes #367

## Problem

Since #330/#353 made the tmux server outlive the app, relaunching Crow **adopts** the still-running Claude window instead of re-running `claude --continue`. That re-`continue` previously doubled as the refresh mechanism: it fired a `SessionStart(resume)` hook which repopulated the in-memory, **non-persisted** `SessionHookState`, so the sidebar status dot + row tint snapped to reality on every boot.

Post-#330, no hook event fires on adoption, so all hook state is blank on launch and the colors sit on a static default (solid green) until the user next interacts with Claude.

## Fix

- **Persist + restore the color-driving hook state.** New Codable `PersistedHookState` (`claudeState`, `pendingNotification`, `lastTopLevelStopAt`) added to `StoreData`, restored in `hydrateState` so colors are correct on first paint. `lastToolActivity` is deliberately excluded — it changes on every `PostToolUse`, only feeds badge text (not colors), and would be stale after relaunch anyway. `StoreData.hookStates` is **optional**, so existing `store.json` files still decode (backward compatible).
- **Keep the store fresh** two ways: incrementally in the `hook-event` handler (only when the color-relevant snapshot actually changes, so frequent `PostToolUse` events don't trigger writes — a handful per turn) and as a full snapshot in `persistState()` at quit. Entries are pruned on session delete.
- **Restore the two adopt-branch side effects** that `launchClaude` performs but adoption skipped, gated on `trackReadiness` (managed work terminals only): re-write the hook config so the adopted Claude's hooks still route to the correct session, and re-seed `remoteControlActiveTerminals` so the RemoteControl badge reappears.

No second `claude --continue` is pasted into a live pane — the #330 adoption guards (`autoLaunchTerminals.remove` + `.claudeLaunched`) are untouched.

## Acceptance criteria

1. ✅ Quit with a non-idle session (working / done / needs-attention), relaunch → sidebar colors reflect true state without interaction.
2. ✅ RemoteControlBadge reappears after relaunch for `--rc` terminals.
3. ✅ Hook events from an adopted Claude still route (hook config re-written on adopt).
4. ✅ No second `claude --continue` pasted (the #330 fix is preserved).

## Testing

- `make ghostty` + `swift build --arch arm64` — clean build.
- Full test target compiles & links cleanly; the test *runner* can't execute in this environment due to a Rosetta arch mismatch (`have 'arm64', need 'x86_64'`) unrelated to these changes.
- Manual acceptance (1–4 above) to be verified in-app.


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review

**Strengths:**
- No new untrusted input enters the persistence path — `PersistedHookState` carries only enum + optional struct + Date, all derived from hook events Crow already accepts and processes today.
- Backward-compatible decoding via the optional `hookStates` field avoids the corrupt-store backup path on existing installs (verified by reading the existing decode logic in `JSONStore.init`).
- File ACL remains 0o600 (unchanged); persisted notification text is no more sensitive than what already lives in the hook ring buffer.
- The adopt-path hook config rewrite reuses the existing `HookConfigGenerator.writeHookConfig` API with `terminal.sessionID` (not a derived value), so it cannot misroute hooks to a foreign session.

**Concerns:**
- None observed.

### Code Quality

**Strengths:**
- The fields chosen for the snapshot (`claudeState`, `pendingNotification`, `lastTopLevelStopAt`) are exactly the subset that drives sidebar color; deliberately excluding `lastToolActivity` keeps `PostToolUse` (the highest-frequency event) from triggering disk writes.
- The `snapshotAfter != snapshotBefore` gate at `AppDelegate.swift:1627` is correct — `PersistedHookState` is `Equatable`, the comparison runs entirely inside `MainActor.run` so no race, and `store.mutate` is internally lock-safe.
- Restore is correctly scoped to live sessions only (`SessionService.swift:67-71`), so stale snapshots for sessions deleted between runs can't be resurrected; `persistState()` on the next clean quit rewrites the dict from live state and prunes them from disk.
- Delete-path cleanup at `SessionService.swift:662` uses `dict[key] = nil` which on a `[String: PersistedHookState]` dictionary genuinely removes the key (verified with a quick repl) — not just nilling a value.
- Build passes for the `CrowCore` and `CrowPersistence` packages; existing `JSONStoreTests` (26 tests) still pass with the optional `hookStates` field added.
- The adopt-path side-effect restore is correctly gated on `trackReadiness`, mirroring the exact set `launchClaude` would have touched, so the Manager (whose RC is separately seeded in `hydrateState`) is not double-seeded.

- No tests cover the new code paths: `StoreData` round-trip with `hookStates`, the delete-cleanup removing the entry, or the snapshot-diff write gate. The PR notes the test runner can't execute locally due to a Rosetta arch mismatch unrelated to this change, but pure-model tests (e.g. `PersistedHookState` Codable round-trip, `StoreData` decoding without the field) would compile and run in the existing `CrowPersistenceTests` target.
- Subtle behavior change worth being aware of: if Claude actually finished while the app was quit, the persisted `.working` state will now show stale (orange) until the next live hook event, where pre-#330 the `claude --continue` re-fire would have reset to `.done`. This is strictly better than the current solid-green-everywhere bug, but it's a slightly different failure mode rather than a complete fix.
- `state.lastTopLevelStopAt = Date()` on every Stop/StopFailure means every Stop event triggers a store write (snapshot always differs on the Date). Expected/intentional, but worth noting that the “only on real change” gate doesn't coalesce repeated Stops.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | None |
| Yellow | None |
| Green | Add a small `PersistedHookState`/`StoreData` Codable round-trip test (and a delete-cleanup test) once the test runner is unblocked |
| Green | Consider whether stale `.working` after relaunch (Claude finished while quit) warrants an explicit fast probe; out of scope here |

**Recommendation:** Approve. The fix is minimal, correctly scoped, backward-compatible, concurrency-safe, and matches the documented acceptance criteria. The lack of new tests is the only soft point and is explained by the environment constraint.

PR #370, opened 2026-05-23
Default adopted terminals to .done so resumed sessions show green (#367)

### LINKED ISSUE #371: Adopted sessions with no persisted hook state stay grey after relaunch (should be green)
Follow-up to #367 / #368.

## Summary

After #368 persisted & restored hook state, sidebar **card colors** are correct on relaunch for sessions that had a `done`/`working` `claudeState` written to disk. But sessions whose last `Stop` hook fired **before** the persistence feature shipped — most visibly already-**completed** issues — have no entry in `store.json`'s `hookStates`. On relaunch they restore to the `.idle` default, so their cards come up grey/plain instead of the green "done" tint, until the user next interacts with Claude.

## Why they used to be green

Pre-#330, relaunch re-ran `claude --continue`, which fired `SessionStart(source: resume)` and set `claudeState = .done` for **every** resumed session → green cards right after relaunch. #330's adopt path replaced that re-run (correctly, to avoid double-pasting `--continue`) but dropped the refresh side effect, and #368's persistence only helps sessions that already have a state on disk.

## Repro

1. Have one or more **completed** sessions whose tmux window still survives (managed work terminals with a `tmuxBinding`).
2. Quit Crow, relaunch.
3. Expected: those cards show the green `bgDone` tint (as pre-#330).
4. Actual: cards are grey/plain — `claudeState` is the `.idle` default because no `done` was persisted for them.

## Proposed fix

In `SessionService.rehydrateTerminalSurface`'s adopt branch (gated on `trackReadiness` — managed work terminals only), emulate the skipped `SessionStart(resume)`: seed `claudeState = .done` when nothing meaningful was restored (state is still the `.idle` default). A persisted in-progress state (`working`/`waiting`) is restored earlier in `hydrateState` and is more specific, so it still wins for a session quit mid-turn. No second `claude --continue` is pasted.

## Acceptance criteria

1. Relaunch with completed/idle adopted sessions present → their cards come up green without interaction.
2. A session quit mid-`working` still shows the working indicator after relaunch (persisted state wins).
3. No second `claude --continue` is pasted into a live pane (the #330 fix must not regress).

## Relevant files

- `Sources/Crow/App/SessionService.swift` — `rehydrateTerminalSurface` adopt branch, `hydrateState` restore.
- `Packages/CrowUI/Sources/CrowUI/SessionListView.swift` — `rowBackgroundColor` (`bgDone` requires `claudeState == .done && terminalReadiness == .claudeLaunched`).

dgershman:
Closes #371

Follow-up to #368 (issue #367).

## Problem

After #368, status colors restore correctly for sessions that had a `done`/`working` hook state persisted to disk. But sessions whose last `Stop` hook fired **before** the persistence feature shipped — notably already-**completed** issues — have no entry on disk, so on relaunch they restore to the `.idle` default and their cards never turn green.

## Why they used to be green

Pre-#330, relaunch re-ran `claude --continue`, which fired `SessionStart(source: resume)` → set `claudeState = .done` for **every** resumed session → green cards across the board right after relaunch. The #330 adopt path replaced the re-run (correctly, to avoid double-pasting `--continue`) but dropped that refresh side effect.

## Fix

In the adopt branch (gated on `trackReadiness`, i.e. managed work terminals — same scope as the hook-config/RemoteControl re-seeds from #368), emulate the skipped `SessionStart(resume)`: seed `claudeState = .done` when nothing meaningful was restored (state is still the `.idle` default).

A persisted in-progress state (`working`/`waiting`) is restored earlier in `hydrateState` and is more specific, so it still wins for a session quit mid-turn. No second `claude --continue` is pasted — the #330 guards are untouched.

## Effect

Adopted sessions with no persisted state (completed/idle issues) come up green again like pre-#330, while genuinely-in-progress sessions keep their accurate working/waiting indicator.

## Testing

- `swift build --arch arm64` — clean build.
- Manual: relaunch with completed sessions present → their cards come up green without interaction; a session quit mid-`working` still shows working.

### DIFF
--- a/Sources/Crow/App/SessionService.swift
+++ b/Sources/Crow/App/SessionService.swift
@@ -256,6 +256,17 @@ final class SessionService {
                     if appState.remoteControlEnabled {
                         appState.remoteControlActiveTerminals.insert(terminal.id)
                     }
+                    // Emulate the SessionStart(source: resume) hook that pre-#330
+                    // relaunch fired (via re-running `claude --continue`), which set
+                    // claudeState to .done — so an adopted, idle-at-the-prompt Claude
+                    // shows the "done" green card like it used to (#367). A persisted
+                    // in-progress state (working/waiting) was already restored in
+                    // hydrateState and is more specific, so only fill in .done when
+                    // nothing meaningful was restored (still the .idle default).
+                    let hookState = appState.hookState(for: terminal.sessionID)
+                    if hookState.claudeState == .idle {
+                        hookState.claudeState = .done
+                    }
                 }
                 return terminal  // binding unchanged → no redundant persist
             } catch {


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new external inputs, no IPC surface, no privileged operations — purely an in-memory UI-state seed on the `@MainActor` adopt path.
- No secrets, credentials, or PII touched.

**Concerns:**
- None.

### Code Quality

**Correctness — ordering is sound.** In `SessionService.hydrateState()` the synchronous `restoreHookState(...)` loop (`SessionService.swift:66-72`) runs before the per-terminal `Task { @MainActor in ... }` rehydration is dispatched (`SessionService.swift:171-182`). Because we are already on `@MainActor`, those tasks cannot run until `hydrateState` returns, so any persisted `working`/`waiting` is already restored when the new block at `SessionService.swift:266-269` checks `claudeState == .idle`. The "more specific persisted state wins" claim in the comment holds.

**Scope gate is right.** The new block is inside `if trackReadiness { ... }`, and `trackReadiness` is `!isManagerSession && original.isManaged` (`SessionService.swift:176`) — exactly the same set as the sibling hook-config / RemoteControl re-seeds from #368, so the Manager (seeded separately in `hydrateState`) is left alone as intended.

**Persistence is intentionally implicit.** The `.done` seed is not immediately written to `data.hookStates`; it relies on the next hook event or `applicationWillTerminate` → `persistState()` to flush. That's fine: on the next launch the same derivation re-runs from the default `.idle`, so dropping the write avoids an extra mutate per adopt.

**Minor — overwriting a persisted `.idle`.** The comment says "fill in .done when nothing meaningful was restored (still the .idle default)," but `.idle` is also what `SessionEnd` persists (`AppDelegate.swift:1580`). After a clean `SessionEnd`, the persisted state is `.idle`, and this block will rewrite it to `.done` on relaunch. This matches the pre-#330 behavior (where `claude --continue` re-fired and `SessionStart(resume)` → `.done`), so the resulting UX is correct, but the comment slightly understates the rewrite. Not blocking.

**No automated test.** Manual testing notes in the PR cover both branches (idle → green; mid-`working` preserved). A unit test that calls `rehydrateTerminalSurface` against a faked tmux adoption would be nice but is out of scope for an 11-line fix; the surrounding paths don't have direct coverage either.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | — |
| Yellow | — |
| Green | Comment at `SessionService.swift:265` could note that persisted `.idle` (from `SessionEnd`) is also overwritten, matching the pre-#330 behavior |

**Recommendation:** Approve. Tight, well-scoped restoration of the pre-#330 green-on-relaunch behavior; correctly gated to managed work terminals; ordering ensures persisted in-progress state still wins.

---

PR #376, opened 2026-05-24
Preserve tmuxBinding across hydrate rebuild so Manager re-attaches (#374)

### LINKED ISSUE #374: Manager tmux window leaks on every relaunch — hydrate command-rebuild drops tmuxBinding
## Summary

Since #330 (persist tmux server / reattach across relaunch), the **Manager** terminal spawns a brand-new tmux window and a fresh `claude` process on every Crow relaunch instead of re-attaching to its existing window. The previous Manager window is orphaned in the `crow-cockpit` tmux session and never reaped. Over a dev session with ~16 relaunches this accumulated ~18 orphaned `Manager` / `Manager 2` windows.

Work-session ("Claude Code") terminals are unaffected — they adopt cleanly.

## Root cause

`SessionService.hydrateState()` rebuilds the Manager terminal row to refresh its `--rc` / `--name` / `--permission-mode` flags (`Sources/Crow/App/SessionService.swift:116-128`):

```swift
terminals[i] = SessionTerminal(
    id: terminals[i].id, sessionID: terminals[i].sessionID,
    name: terminals[i].name, cwd: terminals[i].cwd,
    command: rebuiltCommand, isManaged: terminals[i].isManaged,
    createdAt: terminals[i].createdAt
)   // tmuxBinding NOT passed -> defaults to nil
```

`SessionTerminal.init` defaults `tmuxBinding` to `nil` (`Packages/CrowCore/Sources/CrowCore/Models/Terminal.swift:57`), so reconstructing the row **drops the binding**. Then `rehydrateTerminalSurface` (`SessionService.swift:220`) guards on `if let binding = terminal.tmuxBinding` — now nil — so it skips `adoptTerminal` and falls through to `registerTerminal`, spawning a fresh window + fresh `claude` (not `--continue`). `applyRehydrationResult` persists the new binding, so the store always looks clean while the prior window leaks.

Why only the Manager: the rebuild is gated on `command.contains("claude")`, and the Manager is the only terminal whose `command` is persisted with the claude invocation (work terminals persist with `command: nil`). The same latent binding-drop exists in the two work-terminal rebuild branches (`SessionService.swift:83-92` and `96-106`); they just do not fire on current data.

## Evidence

Live `crow-cockpit` session after a day of dev relaunches — store references only windows 4/5/6/9, everything else is leaked:

```
1,2,5,7,10,11,13,14,15,16:  Manager
6,8,12,20,21,22,23,24:      Manager 2
4: Claude Code   9: Claude Code
```

## Fix

Mutate in place instead of reconstructing — `command` is a `var`, so this preserves `tmuxBinding` and all other fields:

```swift
terminals[i].command = rebuiltCommand
```

Convert all three reconstruction sites in `hydrateState` (the Manager rebuild + the two work-terminal branches) to in-place mutation to eliminate the whole binding-drop class.

## Acceptance

- [ ] Relaunching Crow re-attaches the Manager to its existing tmux window (same `claude` process resumes; no new window).
- [ ] No orphaned `Manager` windows accumulate in `crow-cockpit` across relaunches.
- [ ] Work-session terminals still adopt correctly (no regression).
- [ ] `swift build` + full test suite pass.

dgershman:
Closes #374

## Problem
Since #330 the tmux server outlives the app, so terminals should re-attach to their existing windows on relaunch. But the **Manager** terminal spawned a brand-new tmux window and a fresh `claude` process on every relaunch, orphaning the prior window in `crow-cockpit` (~18 leaked `Manager` windows after a day of dev relaunches).

## Root cause
`SessionService.hydrateState()` rebuilt terminal rows via the memberwise `SessionTerminal.init` to refresh `command`/`isManaged`. That initializer defaults `tmuxBinding` to `nil`, so every rebuild **dropped the binding**. `rehydrateTerminalSurface` then failed its `if let binding = terminal.tmuxBinding` guard and fell through from `adoptTerminal` to registering a fresh window + fresh `claude`.

Only the Manager fired on current data (its rebuild is gated on `command.contains("claude")`, and it's the only terminal that persists its claude invocation in `command`), but the two work-terminal rebuild branches carried the same latent bug.

## Fix
Mutate the rows in place instead of reconstructing them. `command` and `isManaged` are `var`s, so in-place assignment refreshes them while preserving `tmuxBinding` (and `backend`). All three reconstruction sites in `hydrateState` are converted.

## Tests
- `hydratePreservesManagerTmuxBinding` — a Manager hydrate rebuild keeps the binding while still rebuilding the command (and re-seeds the Remote Control active set).
- `hydratePreservesWorkTerminalTmuxBinding` — a managed work terminal keeps its binding while its claude command is cleared (locks down the latent branch).

`swift build` and the root test suite (145 tests) pass. The one CrowTerminal failure (`retryReadinessEmitsTimedOutWhenSentinelMissing`) is a pre-existing timing-flaky tmux integration test, unrelated to this diff (passes in isolation).

### DIFF
--- a/Sources/Crow/App/SessionService.swift
+++ b/Sources/Crow/App/SessionService.swift
@@ -83,12 +83,10 @@ final class SessionService {
                 if !hasManagedTerminal, let idx = terminals.firstIndex(where: {
                     $0.name == "Claude Code" || ($0.command?.contains("claude") ?? false)
                 }) {
-                    terminals[idx] = SessionTerminal(
-                        id: terminals[idx].id, sessionID: terminals[idx].sessionID,
-                        name: terminals[idx].name, cwd: terminals[idx].cwd,
-                        command: terminals[idx].command, isManaged: true,
-                        createdAt: terminals[idx].createdAt
-                    )
+                    // Mutate in place so tmuxBinding (and every other field) is
+                    // preserved — reconstructing via the memberwise init defaults
+                    // tmuxBinding to nil and breaks adopt-on-relaunch (#374).
+                    terminals[idx].isManaged = true
                 }
 
                 // For managed terminals, clear the claude command so they start as plain shells.
@@ -96,12 +94,8 @@ final class SessionService {
                 for i in terminals.indices {
                     if terminals[i].isManaged,
                        let cmd = terminals[i].command, cmd.contains("claude") {
-                        terminals[i] = SessionTerminal(
-                            id: terminals[i].id, sessionID: terminals[i].sessionID,
-                            name: terminals[i].name, cwd: terminals[i].cwd,
-                            command: nil, isManaged: true,
-                            createdAt: terminals[i].createdAt
-                        )
+                        // In-place so tmuxBinding survives the rebuild (#374).
+                        terminals[i].command = nil
                     }
                 }
             } else {
@@ -115,12 +109,11 @@ final class SessionService {
                 let rebuiltCommand = managerCommand(sessionName: session.name)
                 for i in terminals.indices {
                     if let cmd = terminals[i].command, cmd.contains("claude") {
-                        terminals[i] = SessionTerminal(
-                            id: terminals[i].id, sessionID: terminals[i].sessionID,
-                            name: terminals[i].name, cwd: terminals[i].cwd,
-                            command: rebuiltCommand, isManaged: terminals[i].isManaged,
-                            createdAt: terminals[i].createdAt
-                        )
+                        // Mutate in place rather than reconstructing the row:
+                        // the memberwise init drops tmuxBinding, which made the
+                        // Manager spawn a fresh window + claude every relaunch
+                        // instead of re-attaching to its live window (#374).
+                        terminals[i].command = rebuiltCommand
                         if appState.remoteControlEnabled {
                             appState.remoteControlActiveTerminals.insert(terminals[i].id)
                         }
--- a/Tests/CrowTests/ManagerMigrationTests.swift
+++ b/Tests/CrowTests/ManagerMigrationTests.swift
@@ -61,4 +61,74 @@ struct ManagerMigrationTests {
         #expect(cmd2.contains("--permission-mode auto"))
         #expect(cmd2.contains("--rc"))
     }
+
+    /// #374: hydrating the Manager rebuilds its claude `command` to refresh the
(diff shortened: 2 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new I/O, parsing, shell-out, or persistence schema changes — the diff is purely a value-mutation refactor in already-trusted code paths.
- `TmuxBinding` is `Equatable` and the test compares full structural equality, so a regression that altered any of `socketPath`/`sessionName`/`windowIndex` would also be caught.

**Concerns:** None.

### Code Quality
- Root cause is correctly identified: `SessionTerminal`'s memberwise init defaults `tmuxBinding`, `backend`, and `createdAt` to zero-values, so reconstructing a row to refresh one field silently dropped all of them. In-place mutation on `var` fields is the right minimal fix and preserves `backend` too (latent — currently single-cased, but the schema is designed to evolve, per `Models/Terminal.swift:11`).
- All three reconstruction sites in `hydrateState` are converted (`SessionService.swift:89`, `:98`, `:116`). The third is the one #374 actually fired on; the other two are latent and now locked down.
- Comments are tight and explain *why* with an issue reference (#374) rather than restating *what*. Good.
- Test coverage is thorough for what the PR claims:
  - `hydratePreservesManagerTmuxBinding` exercises branch 3, asserts both binding preservation *and* the rebuild side effect (command refresh + RC active-set re-seed) — so a future "fix" that just skipped the rebuild wouldn't pass.
  - `hydratePreservesWorkTerminalTmuxBinding` exercises branch 2 (command → nil for managed work terminals).
  - Branch 1 (the `isManaged = true` backward-compat migration at `:89`) has no dedicated test, but it's a single-field assignment on an already-hard-to-exercise legacy path; not a gap worth holding the PR on.
- Test `socketPath: \"/tmp/crow.sock\"` is just a Codable-struct field that's never opened — the project's `$TMPDIR` guidance doesn't apply here.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | — |
| Yellow | — |
| Green | Optional: a unit test for branch 1 (the `isManaged` backward-compat migration preserving binding) would close the last seam, but the fix there is trivial enough that this is purely defense-in-depth. |

**Recommendation:** Approve — narrow, correct, well-explained fix with a clear root-cause writeup and tests that lock the regression down. The PR also fixes two latent variants of the same bug while it's there, which is the right call.

PR #377, opened 2026-05-24
Add "Restart tmux Server" menu item to recycle the cockpit backend

### LINKED ISSUE #375: Add "Restart tmux Server" menu item to recycle the cockpit backend
## Summary

Add an app-menu item — **"Restart tmux Server"** — that tears down the persistent `crow-tmux` backend server and rebuilds every session's terminal window from scratch. This is a manual recovery affordance for when the cockpit session gets into a bad state: orphaned/leaked windows (see #374), a wedged or unresponsive tmux server, or stale windows after a tmux upgrade.

Today the only ways to recover are to quit Crow (which by design *leaves the server running* since #330) or to kill the socket from a shell. There's no in-app, supported way to recycle the server.

## Motivation

- Since #330 the tmux server deliberately outlives the app (`$TMPDIR/crow-tmux.sock`), so a normal quit/relaunch does **not** clear a corrupted cockpit session.
- Bugs like #374 (Manager window leak) can pile up orphaned `Manager`/`Manager 2` windows in `crow-cockpit`; there's no UI to clear them.
- `TmuxBackend.onUnresponsive` already exists (`AppDelegate.swift:360`) — when the server wedges, a "Restart tmux Server" action is the natural remediation to offer.

## Proposed change

Add a menu item next to the existing **"Restart Manager"** item in `setupMenu()` (`Sources/Crow/App/AppDelegate.swift:858-885`), wired to a new `@objc func restartTmuxServer()` selector following the same pattern as `restartManager` (`:871-873`).

Behavior on invoke:
1. **Confirm** via `NSAlert` — this kills all terminal processes (running `claude` instances included), so it must not be a single misclick. Mirror the existing confirmation style.
2. **Tear down** the server: `TmuxBackend.shared.shutdown(killServer: true)` (`Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:137`), which runs `kill-server` and unlinks the scratch/sentinel files.
3. **Reconfigure** the backend with the same binary + stable socket path used at launch (`AppDelegate.launchMainApp` `:356-359`), recreating the cockpit session via `ensureCockpitSession`.
4. **Rebuild every terminal surface**: re-register a fresh tmux window for each persisted terminal across all sessions, then re-launch its `claude` (Manager launches its command directly; work terminals re-run `claude --continue`). This is the same work `hydrateState` / `rehydrateTerminalSurface` do on launch — factor out a reusable `rebuildAllSurfaces()` rather than duplicating it. Since the server was just killed, every binding is dead, so the rebuild takes the `registerTerminal` path (not adopt) and persists the new `tmuxBinding`s.
5. Re-select the previously selected session/terminal so focus is preserved.

## Notes / scope
- Work-session terminals resume with `claude --continue`; the Manager relaunches via its stored command. In-flight, unsaved TUI state in any pane is lost — that's inherent to a server restart and is what the confirmation guards.
- Keep it in the app menu (no global keyboard shortcut) to avoid accidental fires.
- Out of scope: auto-restart on `onUnresponsive` (could be a follow-up that reuses this same path); per-session window restart (distinct from a full server recycle).

## Acceptance
- [ ] "Restart tmux Server" appears in the Crow app menu and prompts for confirmation before acting.
- [ ] Invoking it kills the running tmux server and recreates one fresh window per persisted terminal, with `claude` relaunched in each (Manager + all work sessions).
- [ ] The `crow-cockpit` session afterward contains exactly the live terminals — no orphaned/leaked windows.
- [ ] Persisted `tmuxBinding`s are updated to the new window indices; the UI renders all surfaces without a manual relaunch.
- [ ] `swift build` + full test suite pass.

Related: #330, #374.

dgershman:
Closes #375

## What

Adds an app-menu item **"Restart tmux Server"** (next to "Restart Manager") that, after a confirmation alert, tears down the persistent `crow-tmux` backend and rebuilds every session's terminal window from scratch.

## Why

Since #330 the tmux server deliberately outlives the app (`$TMPDIR/crow-tmux.sock`), so a normal quit/relaunch does **not** clear a corrupted cockpit session — orphaned/leaked windows (#374), a wedged server, or stale windows after a tmux upgrade. The only previous recoveries were quitting Crow (which leaves the bad server running) or killing the socket from a shell. This adds an in-app, supported way to recycle the server.

## How

- **AppState**: new `onRestartTmuxServer` callback.
- **AppDelegate**: menu item (menu-only — no key equivalent — to avoid accidental fires); `@objc restartTmuxServer()` with an `NSAlert` confirmation mirroring the existing unresponsive-server alert; callback wiring.
- **SessionService**: factored the `hydrateState` tail into a reusable `rebuildAllSurfaces(forceRegister:)`, shared by launch and the new `restartTmuxServer()`.
  - `restartTmuxServer()` calls `TmuxBackend.shutdown(killServer: true)` then `rebuildAllSurfaces(forceRegister: true)`.
  - With `forceRegister: true`, each terminal's now-dead `tmuxBinding` is dropped so rehydration always takes the `registerTerminal` path (which lazily recreates the controller + cockpit session via `ensureRunningServer`), and managed work terminals' readiness/auto-launch state is re-armed so `claude --continue` relaunches. The Manager relaunches via its stored command.
  - Selection is re-applied to force a SwiftUI re-render that re-attaches a fresh cockpit surface and preserves focus.
  - The launch path is behaviorally unchanged (`forceRegister: false`).

## Testing

- `make ghostty` + `swift build --arch arm64` — clean build.
- `swift test --arch arm64` — 143 tests / 16 suites pass; `CrowCore` package tests — 189 tests / 5 suites pass.
- The rebuild touches `TmuxBackend.shared` / libghostty, so end-to-end behavior (kill server → fresh window per terminal with Claude relaunched, no orphans, focus preserved) is verified interactively per the ticket's acceptance criteria.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -283,6 +283,11 @@ public final class AppState {
     /// preserving the Manager session identity. Wired to `SessionService.restartManager`.
     public var onRestartManager: (() -> Void)?
 
+    /// Called from the "Restart tmux Server" menu item (after confirmation) to
+    /// kill the tmux server and rebuild every terminal surface. Wired to
+    /// `SessionService.restartTmuxServer`.
+    public var onRestartTmuxServer: (() -> Void)?
+
     /// Called when the user clicks "Retry" on a terminal whose tmux readiness
     /// watch timed out before the shell signaled it was interactive.
     public var onRetryReadiness: ((UUID) -> Void)?  // receives terminal ID
--- a/Sources/Crow/App/AppDelegate.swift
+++ b/Sources/Crow/App/AppDelegate.swift
@@ -467,6 +467,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             service?.restartManager(devRoot: devRoot)
         }
 
+        appState.onRestartTmuxServer = { [weak service] in
+            service?.restartTmuxServer()
+        }
+
         appState.onRetryReadiness = { [weak service] terminalID in
             service?.retryReadiness(terminalID: terminalID)
         }
@@ -871,6 +875,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         let restartManagerItem = NSMenuItem(title: "Restart Manager", action: #selector(restartManager), keyEquivalent: "")
         restartManagerItem.target = self
         appMenu.addItem(restartManagerItem)
+        // No keyEquivalent — recycling the tmux server kills every pane's claude,
+        // so keep it menu-only to avoid accidental fires (#375).
+        let restartTmuxItem = NSMenuItem(title: "Restart tmux Server", action: #selector(restartTmuxServer), keyEquivalent: "")
+        restartTmuxItem.target = self
+        appMenu.addItem(restartTmuxItem)
         appMenu.addItem(NSMenuItem.separator())
         appMenu.addItem(withTitle: "Hide Crow", action: #selector(NSApplication.hide(_:)), keyEquivalent: "h")
         let hideOthersItem = NSMenuItem(title: "Hide Others", action: #selector(NSApplication.hideOtherApplications(_:)), keyEquivalent: "h")
@@ -887,6 +896,24 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
     @objc private func restartManager() {
         appState.onRestartManager?()
+    }
+
+    @objc private func restartTmuxServer() {
+        let alert = NSAlert()
+        alert.alertStyle = .warning
+        alert.messageText = "Restart tmux server?"
+        alert.informativeText = """
+            This kills the tmux server and every terminal it hosts — including \
+            all running Claude sessions — then rebuilds a fresh window for each \
+            and relaunches Claude. Use it to recover a stuck or corrupted cockpit.
+
+            Unsaved work in any terminal will be lost.
+            """
+        alert.addButton(withTitle: "Restart tmux Server")
+        alert.addButton(withTitle: "Cancel")
+        if alert.runModal() == .alertFirstButtonReturn {
+            appState.onRestartTmuxServer?()
+        }
     }
 
     @objc private func showAbout() {
--- a/Sources/Crow/App/SessionService.swift
+++ b/Sources/Crow/App/SessionService.swift
@@ -152,35 +152,65 @@ final class SessionService {
             }
         }
 
-        // Wire the readiness callback BEFORE re-registering terminals so the
-        // tmux sentinel's .shellReady event is never lost.
+        // Wire the readiness callback and re-hydrate every persisted terminal.
+        // Since #330 the tmux server outlives the app, so on launch
+        // rehydrateTerminalSurface adopts the persisted window when it's still
+        // live and only re-registers a fresh one as a fallback (post-reboot,
+        // closed window, or a legacy socket). `forceRegister: false` keeps that
+        // adopt-first behavior; the manual "Restart tmux Server" path passes
+        // true. If both adopt and register fail (e.g. tmux uninstalled) the row
+        // is left as-is and simply won't render this launch.
+        rebuildAllSurfaces(forceRegister: false)
+    }
+
+    /// Wire the tmux readiness callback, then re-register a tmux window and
+    /// relaunch claude for every persisted terminal across all sessions. Shared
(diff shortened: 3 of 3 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Destructive action is guarded by an `NSAlert` confirmation with a clear warning about lost work (`AppDelegate.swift:901-917`).
- Menu item has no key equivalent — avoids accidental fires from typos, as called out in the inline comment.
- No external input enters the new code path; it's strictly an internal teardown driven off `TmuxBackend.shared.shutdown(killServer: true)` which uses already-trusted state.
- No new credentials, secrets, network, or process-exec surfaces are added.

**Concerns:**
- None observed.

### Code Quality
**Strengths:**
- Clean refactor: the `hydrateState` tail is factored into `rebuildAllSurfaces(forceRegister:)` with `forceRegister: false` preserving launch-time adopt-first semantics. Behavior on the launch path is unchanged, which is the key invariant.
- `wireTerminalReadiness()` is invoked BEFORE re-registering windows so the sentinel's `.shellReady` is never lost (`SessionService.swift:184-185`). Comment is explicit about the ordering.
- `forceRegister: true` correctly drops `tmuxBinding` per terminal so `rehydrateTerminalSurface` skips the now-dead `adoptTerminal` path and goes straight to `registerTerminal` — matches the post-kill reality.
- Re-arming `terminalReadiness = .uninitialized` + `autoLaunchTerminals.insert` only for managed work terminals (`SessionService.swift:192-199`) is the right scope — Manager sessions launch claude directly as the window child command via `registerTerminal`, so they don't go through `launchClaude`.
- Selection re-assignment trick (`SessionService.swift:580-581`) is sound because `AppState` uses `@Observable`, which fires on every write. The inline comment explains the intent.
- `shutdown(killServer: true)` already destroys `sharedSurface`, clears `bindings`, cancels readiness tasks, and unlinks sentinel/wrapper files — the new caller doesn't need any of those manually.
- Telemetry breadcrumb (`[CrowTelemetry tmux:server_restart_by_user]`) is consistent with existing format.

**Nits (non-blocking):**
- **Stale Manager command edge case** (`SessionService.swift:565`): `restartTmuxServer` re-uses each Manager terminal's stored `command`, which was last rebuilt in `hydrateState` against `remoteControlEnabled` / `managerAutoPermissionMode` at hydrate time. If those toggles changed between hydrate and the manual restart, the Manager relaunches with the stale setting. Consistent with the existing adopt-on-launch behavior, so no regression — but worth knowing. Calling `managerCommand(sessionName:)` again in the Manager branch of the `forceRegister: true` re-arm loop would close it.
- **Destructive button is default** (`AppDelegate.swift:912-914`): "Restart tmux Server" is the first-added button, so it's the rightmost/default — Enter triggers the destructive action. Mirrors the existing unresponsive-server alert, but Apple HIG suggests destructive actions should not be the default. Minor UX nit.
- **No guard against double-invocation**: nothing prevents the user from selecting "Restart tmux Server" twice in rapid succession while the first rebuild's per-terminal `Task`s are still queued. The result is likely benign (second `shutdown` is a no-op on an already-cleared backend, then the queue gets re-armed), but a `tmuxRestartInFlight` flag would be cheap insurance.
- **No new tests**: the `forceRegister: true` branch (binding clear + readiness re-arm) is not unit-tested. This is consistent with the existing `restartManager` path, which also has no test, and the heavy tmux dependency makes it awkward. Not blocking, but a small SessionService test that asserts `forceRegister: true` clears `tmuxBinding` on the seed and re-inserts work-terminal IDs into `autoLaunchTerminals` would be cheap and useful.
- **Doc drift in `TmuxBackend.shutdown` header** (`TmuxBackend.swift:135`): comment says `killServer: true` is for the "crash-watchdog 'Restart tmux server'" path; with #375 it's also the menu-driven path. Worth a one-line update.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | None |
| Yellow | None |
| Green | Stale Manager command on RC/auto-perm toggle; destructive button is default; no double-invoke guard; no unit test for `forceRegister: true`; `TmuxBackend.shutdown` doc could mention menu path |

**Recommendation:** Approve. The refactor is clean, the launch path is behaviorally unchanged, the new path is well-reasoned with explicit comments for every non-obvious step, the destructive action is properly guarded, and CrowCore tests pass (189/5 suites). All call-outs above are polish/nits.

---

PR #379, opened 2026-05-24
Add a way to duplicate an existing job (#378)

### LINKED ISSUE #378: Jobs: add a way to duplicate an existing job
## Summary

Add a way to **duplicate an existing job** in the Jobs feature. Jobs (`JobConfig`) often differ only slightly — same repo/workspace, a tweaked prompt set or schedule — and today the only way to make a near-copy is to re-enter everything by hand in the Add Job form. A "Duplicate" affordance would clone an existing job as the starting point.

## Where

- Model: `Packages/CrowCore/Sources/CrowCore/Models/JobConfig.swift`
- Jobs list + add/edit sheets: `Packages/CrowUI/Sources/CrowUI/SettingsView.swift` (the `jobsTab`; jobs are persisted via `config.jobs.append(job)` + `save()`, e.g. the Add sheet at ~L88-96, and per-row Edit/Delete buttons in the jobs list).
- Form: `Packages/CrowUI/Sources/CrowUI/JobFormView.swift` (already supports both create and edit — "Add" vs "Save" keyed on whether an existing job id is passed).

## Proposed behavior

Add a **Duplicate** action to each job row in the Jobs tab (next to the existing Edit / Delete buttons — e.g. a menu item or a copy icon). Two reasonable implementations; prefer the first:

1. **Open the Add form prefilled from the source job** — present `JobFormView` in *create* mode (no existing id) pre-populated with the source job's `workspace`, `repo`, `prompts`, and `schedule`, with the name pre-filled as a unique suggestion (see below). The user reviews/tweaks and taps **Add**, which appends a brand-new `JobConfig`. This reuses the existing create flow and lets the user adjust before committing.
2. **Immediate clone** — append a copy straight away (no form), then let the user edit it.

## Cloning rules (important)

A duplicate must be a genuinely new job, not a shared reference:

- **New `id`** — generate a fresh `UUID` (don't reuse the source's).
- **Unique `name`** — `JobConfig.validateName` rejects a name that already exists (case-insensitive), so the copy can't keep the original name. Suggest `"{name} copy"`, and if that's taken, `"{name} copy 2"`, etc.
- **Reset runtime state** — `lastRunAt = nil` and a fresh `createdAt`, so the scheduler treats it as new and doesn't immediately think it's overdue.
- **Default the copy to `enabled = false`** (recommended) so a duplicated job doesn't silently start firing on its schedule before the user has reviewed it. (Open to keeping it enabled if that's preferred — but disabled-by-default is the safer surprise-free choice.)
- Copy `workspace`, `repo`, `prompts`, `schedule` verbatim.

## Acceptance criteria

- [ ] Each job row in the Jobs tab has a Duplicate action.
- [ ] Duplicating produces a new job with a new `id`, a unique name (`"… copy"`), copied repo/workspace/prompts/schedule, `lastRunAt` cleared, and a fresh `createdAt`.
- [ ] The duplicate is persisted to `config.jobs` and written to `{devRoot}/.claude/config.json` via the existing `save()` path.
- [ ] The duplicated job does not immediately fire (disabled by default, or at minimum its schedule is computed from its own fresh `createdAt`).
- [ ] Editing/deleting the duplicate does not affect the original, and vice versa.
- [ ] `swift build` + tests pass; add a small unit test for the clone helper (new id, unique name, reset run state).

dgershman:
Closes #378

## What

Adds a **Duplicate** action to each job row in the Jobs tab. Jobs often differ only slightly (same repo/workspace, a tweaked prompt set or schedule), but the only way to make a near-copy was to re-enter everything by hand. Duplicating clones an existing job as a starting point the user can review and tweak before saving.

## How

- **`JobConfig.duplicated(existingNames:)`** — clones a job with a fresh `id` and `createdAt`, cleared `lastRunAt`, and `enabled = false` so it can't fire before review. `workspace`/`repo`/`prompts`/`schedule` are copied verbatim. The name is made unique against existing names (`"… copy"`, `"… copy 2"`, …) via a new `uniqueCopyName` helper that matches `validateName`'s case-insensitive rule.
- **`JobFormView`** — gains an `isDuplicate` flag: it prefills its fields from the passed job but treats the result as a brand-new job (fresh id/createdAt, cleared lastRunAt, **"Add"** button), reusing the existing create flow.
- **`SettingsView`** — a duplicate icon button on each job row opens the prefilled form; saving appends a new `JobConfig` via the existing `config.jobs.append` + `save()` path.

## Acceptance criteria

- [x] Each job row has a Duplicate action
- [x] Duplicating produces a new job: new `id`, unique `"… copy"` name, copied repo/workspace/prompts/schedule, `lastRunAt` cleared, fresh `createdAt`
- [x] Persisted to `config.jobs` via the existing `save()` path
- [x] Duplicate does not immediately fire (disabled by default + fresh `createdAt`)
- [x] Editing/deleting the duplicate doesn't affect the original (value type + new `id`)
- [x] Unit test for the clone helper (new id, unique name, reset run state, `"copy N"` escalation)

## Testing

CrowCore test suite passes (10/10), including the new `jobConfigDuplicated`. The CrowUI changes mirror the existing Add/Edit sheet patterns; full app build verification is left to CI.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Models/JobConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/JobConfig.swift
@@ -79,6 +79,36 @@ public struct JobConfig: Identifiable, Codable, Sendable, Equatable {
         return nil
     }
 
+    /// A new job seeded from this one as the starting point for a duplicate:
+    /// fresh `id` + `createdAt`, cleared `lastRunAt`, and `enabled = false` so it
+    /// can't fire before the user reviews it. `workspace`, `repo`, `prompts`, and
+    /// `schedule` are copied verbatim. The name is made unique against
+    /// `existingNames` by appending " copy" (then " copy 2", " copy 3", …).
+    public func duplicated(existingNames: [String]) -> JobConfig {
+        JobConfig(
+            id: UUID(),
+            name: Self.uniqueCopyName(of: name, existingNames: existingNames),
+            workspace: workspace,
+            repo: repo,
+            prompts: prompts,
+            schedule: schedule,
+            enabled: false,
+            lastRunAt: nil,
+            createdAt: Date()
+        )
+    }
+
+    /// Builds a "<base> copy" name absent from `existingNames` (case-insensitive,
+    /// matching `validateName`), escalating to " copy 2", " copy 3", … on collision.
+    static func uniqueCopyName(of base: String, existingNames: [String]) -> String {
+        let taken = Set(existingNames.map { $0.lowercased() })
+        let first = "\(base) copy"
+        if !taken.contains(first.lowercased()) { return first }
+        var n = 2
+        while taken.contains("\(base) copy \(n)".lowercased()) { n += 1 }
+        return "\(base) copy \(n)"
+    }
+
     /// The next time this job should fire strictly after `reference`.
     ///
     /// The scheduler treats the job as due when
--- a/Packages/CrowCore/Tests/CrowCoreTests/JobConfigTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/JobConfigTests.swift
@@ -124,3 +124,32 @@ import Testing
     #expect(JobConfig.validateName("Audit", existingNames: ["audit"]) != nil) // case-insensitive dupe
     #expect(JobConfig.validateName("Audit", existingNames: ["Other"]) == nil)
 }
+
+@Test func jobConfigDuplicated() {
+    let src = JobConfig(
+        name: "Audit",
+        workspace: "ws",
+        repo: "o/r",
+        prompts: ["a", "b"],
+        schedule: .interval(seconds: 3600),
+        enabled: true,
+        lastRunAt: Date(),
+        createdAt: Date(timeIntervalSince1970: 0)
+    )
+    let copy = src.duplicated(existingNames: ["Audit"])
+
+    #expect(copy.id != src.id)              // fresh id
+    #expect(copy.name == "Audit copy")      // unique name
+    #expect(copy.enabled == false)          // disabled by default
+    #expect(copy.lastRunAt == nil)          // run state reset
+    #expect(copy.createdAt > src.createdAt) // fresh createdAt
+    #expect(copy.workspace == src.workspace) // copied verbatim
+    #expect(copy.repo == src.repo)
+    #expect(copy.prompts == src.prompts)
+    #expect(copy.schedule == src.schedule)
+
+    // escalates when " copy" is already taken (case-insensitive)
+    #expect(src.duplicated(existingNames: ["Audit", "audit copy"]).name == "Audit copy 2")
+    // the suggested name must itself pass the validator against the originals
+    #expect(JobConfig.validateName(copy.name, existingNames: ["Audit"]) == nil)
+}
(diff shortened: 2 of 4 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface: the duplicated `name` flows only into SwiftUI text and the local config file, both already user-controlled. No shell, SQL, or HTML interpolation.
- `uniqueCopyName` uses Swift string interpolation only — no format-string or injection risk.
- `enabled = false` on duplicate is a meaningful safety property: a clone with copied repo/workspace/prompts can't fire unattended before review.

**Concerns:**
- None.

### Code Quality
**Strengths:**
- `duplicated(existingNames:)` is a clean value-type helper — fresh `id`/`createdAt`, cleared `lastRunAt`, copied payload — exactly matching the acceptance criteria.
- `uniqueCopyName` uses the same case-insensitive comparison as `validateName` (`JobConfig.swift:75`), so the generated name is guaranteed to pass the validator. The test asserts this (`JobConfigTests.swift:154`).
- The `O(n)` set construction plus linear escalation is fine for realistic job counts.
- `JobFormView`'s `isDuplicate` flag is a minimally-invasive change: nulling `existingID`/`existingLastRunAt` and refreshing `existingCreatedAt` reuses the existing add-vs-edit branching (`JobFormView.swift:293`), so the **\"Add\"** button label and `buildJob()` semantics fall out for free.
- `SettingsView` duplicate sheet mirrors the existing `isAddingJob` sheet (`SettingsView.swift:90`) — same `existingNames: otherJobNames()`, same `config.jobs.append + save()` path — keeping the patterns consistent.
- `.sheet(item: $duplicatingJob)` correctly uses `JobConfig`'s `Identifiable` conformance; the prefilled state is captured at button press, so the sheet identity is stable.
- Accessibility label `\"Duplicate \(job.name)\"` is provided on the new button (`SettingsView.swift:407`).

- `uniqueCopyName` jumps from `\" copy\"` directly to `\" copy 2\"`, skipping `\" copy 1\"`. This matches the Finder convention and the docstring, just worth being explicit about — could be worth a one-liner test for the `\" copy 3\"` / `\" copy 4\"` escalation chain to lock in the contract.
- Duplicating an already-duplicated job produces `\"Audit copy copy\"`. Functional but verbose; if this becomes a pain point, you could optionally strip a trailing `\" copy( N)?\"` before re-appending. Not worth doing speculatively.
- If the source `name` has trailing whitespace, the suffix lands as `\"X  copy\"`. Same behavior as `validateName`, so consistent — flagging only because the form doesn't trim the name field until `buildJob()`.

### Test Coverage
- `jobConfigDuplicated` exercises the main path (fresh id, name uniqueness, disabled, lastRunAt cleared, fresh createdAt, verbatim payload copy, single-step escalation, validator round-trip). Good coverage of the acceptance criteria.
- CrowUI changes aren't unit-tested; the PR notes this is left to CI/manual verification, which is reasonable given the SwiftUI surface.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | None |
| Yellow | None |
| Green | Optional: add escalation test for `\" copy 3\"`+ chain; consider whether to strip an existing `\" copy\"` suffix when duplicating a duplicate |

**Recommendation:** Approve. The change is small, well-scoped, mirrors existing patterns, has no security implications, and the new helper is covered by a focused test that asserts the contract that matters (uniqueness against the same validator the form uses).

PR #392, opened 2026-05-26
Launch scheduled jobs in auto-permission mode

### LINKED ISSUE #390: Jobs feature: launch scheduled sessions in auto-permission mode (global toggle)
## Summary

Sessions spawned by the new **Jobs** feature do not start in auto-permission mode. They behave like a fresh interactive session — each `crow` / `gh` / `git` invocation in the prompt prompts for approval, which defeats the point of an unattended scheduled run.

We already have an equivalent control for the Manager tab (`managerAutoPermissionMode`, exposed in Settings → Automation), which passes `--permission-mode auto` so the Manager session can run `crow`, `gh`, and `git` without per-call approval. Jobs should pick up the same behavior.

## Proposed behavior

Add a **global** toggle in the Jobs tab settings (not per-job) — something like `jobsAutoPermissionMode: Bool` on `AppConfig`, mirrored by an `appState.jobsAutoPermissionMode`. When enabled, every scheduled job-launched session is started with `--permission-mode auto`, the same way `managerAutoPermissionMode` is wired through `SessionService` today.

A per-job override is not needed for v1 — a single switch covering all jobs is enough. We can revisit per-job control later if a user wants to exempt one specific job.

## Why global, not per-job

- Matches the spirit of `managerAutoPermissionMode` (single global switch).
- Jobs are by definition unattended; if the user has authorized auto mode for the Manager, they almost certainly want it for scheduled jobs too.
- Avoids cluttering `JobFormView` with a per-row toggle.

## Implementation pointers

The Manager equivalent already exists end-to-end and is the template to copy:

- Config flag: `AppConfig.managerAutoPermissionMode` — `Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift`
- Settings UI: `Packages/CrowUI/Sources/CrowUI/AutomationSettingsView.swift` (the "Manager auto-permission mode" row, around line 106 with the `--permission-mode auto` copy)
- App state mirror: `AppDelegate.swift` lines 386 & 996 — `appState.managerAutoPermissionMode = config.managerAutoPermissionMode`
- Launch wiring: `Sources/Crow/App/SessionService.swift:588` — `autoPermissionMode: appState.managerAutoPermissionMode`
- Job model: `Packages/CrowCore/Sources/CrowCore/Models/JobConfig.swift`
- Job UI: `Packages/CrowUI/Sources/CrowUI/JobFormView.swift`

The change is roughly:
1. Add `jobsAutoPermissionMode: Bool` (default `false`, or `true` if we want it opt-out) to `AppConfig`, with the same persistence + migration test coverage as `managerAutoPermissionMode`.
2. Surface it in the Jobs settings UI as a single global toggle with copy similar to the Manager one (note Claude Code 2.1.83+ + Max/Team/Enterprise/API requirement).
3. In the code path that launches a session for a scheduled job, pass `autoPermissionMode: appState.jobsAutoPermissionMode` into `SessionService` the same way the Manager does.

## Acceptance

- [ ] Global toggle in the Jobs tab settings: "Run scheduled jobs in auto-permission mode".
- [ ] When ON, every job-launched session starts with `--permission-mode auto` — no per-call approval for `crow` / `gh` / `git`.
- [ ] When OFF, behavior matches today.
- [ ] Setting persists to `config.json` as `jobsAutoPermissionMode` and is covered by `AppConfigTests` / `ConfigStoreTests` and a migration test (default value for existing configs).
- [ ] Settings copy notes the same Claude Code 2.1.83+ / plan requirement as the Manager toggle, and that the change takes effect on the next job run.

## Out of scope

- Per-job override (file separately if anyone asks for it).
- Changing the default value of `managerAutoPermissionMode`.

dhilgaertner:
## Summary

- Adds a global **Jobs auto-permission mode** toggle (default ON) under Settings → Jobs, mirroring `managerAutoPermissionMode` end-to-end.
- When ON, sessions spawned by the Jobs scheduler launch `claude` with `--permission-mode auto`, so unattended runs don't prompt for every `crow`/`gh`/`git` call.
- Scoped to `.job`-kind sessions only — review and other session kinds are unaffected by the toggle.

Closes #390

## Implementation

| File | Change |
|---|---|
| `AppConfig.swift` | Add `jobsAutoPermissionMode: Bool` (default `true`) — stored property, init param, decoder fallback, `CodingKeys`. |
| `AppState.swift` | Mirror property with doc comment. |
| `AppDelegate.swift` | Two sync points (launch + on-save) parallel to `managerAutoPermissionMode`. |
| `SessionService.swift` (`launchClaude`) | Resolve the session once; pass `autoPermissionMode = (session.kind == .job) && appState.jobsAutoPermissionMode` to `ClaudeLaunchArgs.argsSuffix`. |
| `SettingsView.swift` (`jobsTab`) | New "Auto-permission mode" section above the job list with copy mirroring the Manager toggle (`AutomationSettingsView`). |
| `AppConfigTests.swift` | Round-trip + missing-key default tests, mirroring the existing Manager tests. |

`ClaudeLaunchArgs.argsSuffix` already accepts `autoPermissionMode` — no helper change needed.

## Default value

`true` (opt-out), matching `managerAutoPermissionMode`. Rationale: jobs are unattended by definition; users who have Manager auto-mode on almost certainly want it for jobs too, and legacy configs without the key get the behavior automatically.

## Test plan

- [x] `make build` — clean.
- [x] `swift test --package-path Packages/CrowCore` — all 194 tests pass, including the two new ones (`appConfigJobsAutoPermissionModeRoundTrip`, `appConfigJobsAutoPermissionModeDefaultsTrueWhenKeyMissing`).
- [ ] Manual: Settings → Jobs shows the new toggle defaulted ON; toggling persists across app restarts.
- [ ] Manual: with toggle ON, running a job (▶) launches `claude --permission-mode auto …` in the job's terminal.
- [ ] Manual: with toggle OFF, the same job launches without `--permission-mode auto`.
- [ ] Manual: non-job sessions (Review, Manager) are unaffected by the toggle either way.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -43,6 +43,12 @@ public final class AppState {
     /// launch; worker sessions and CLI-spawned terminals are unaffected.
     public var managerAutoPermissionMode: Bool = true
 
+    /// Whether sessions launched by the Jobs scheduler start with
+    /// `--permission-mode auto` so the job's prompts can run `crow`, `gh`, and
+    /// `git` without per-call approval. Mirrors `AppConfig.jobsAutoPermissionMode`.
+    /// Applies only to `.job`-kind sessions; manager/review/CLI sessions are unaffected.
+    public var jobsAutoPermissionMode: Bool = true
+
     /// `true` when the Manager's `claude` process has exited (crash, kill, OOM)
     /// and has not yet been restarted. Drives the "Manager process exited" banner
     /// and enables the "Restart Manager" action. Reset when the Manager relaunches.
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -12,6 +12,11 @@ public struct AppConfig: Codable, Sendable, Equatable {
     public var sidebar: SidebarSettings
     public var remoteControlEnabled: Bool
     public var managerAutoPermissionMode: Bool
+    /// When true, sessions launched by the Jobs scheduler start with
+    /// `--permission-mode auto` so their prompts can run `crow`, `gh`, and
+    /// `git` without per-call approval. Defaults to true — jobs are
+    /// unattended by definition.
+    public var jobsAutoPermissionMode: Bool
     public var telemetry: TelemetryConfig
     public var autoRespond: AutoRespondSettings
     /// When true, `setup.sh` writes a per-worktree `.claude/settings.local.json`
@@ -48,6 +53,7 @@ public struct AppConfig: Codable, Sendable, Equatable {
         sidebar: SidebarSettings = SidebarSettings(),
         remoteControlEnabled: Bool = false,
         managerAutoPermissionMode: Bool = true,
+        jobsAutoPermissionMode: Bool = true,
         telemetry: TelemetryConfig = TelemetryConfig(),
         autoRespond: AutoRespondSettings = AutoRespondSettings(),
         attributionTrailers: Bool = true,
@@ -63,6 +69,7 @@ public struct AppConfig: Codable, Sendable, Equatable {
         self.sidebar = sidebar
         self.remoteControlEnabled = remoteControlEnabled
         self.managerAutoPermissionMode = managerAutoPermissionMode
+        self.jobsAutoPermissionMode = jobsAutoPermissionMode
         self.telemetry = telemetry
         self.autoRespond = autoRespond
         self.attributionTrailers = attributionTrailers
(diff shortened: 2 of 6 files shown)


dgershman (approved):
## Code & Security Review

Clean, tightly-scoped change that mirrors the existing `managerAutoPermissionMode` plumbing end-to-end. Verified the launch path, the config round-trip, and the scoping logic.

### Critical Issues
None.

### Security Review
**Strengths:**
- `--permission-mode auto` is appended as a static literal in `ClaudeLaunchArgs.argsSuffix` — no user input is interpolated, so there's no new injection surface.
- The session `--name` label is still passed through `shellQuote`, unchanged.
- Behavior is gated behind a documented, opt-out toggle and is **scoped to `.job`-kind sessions only** (`SessionService.swift:420`). Manager/Review/CLI sessions are unaffected. I confirmed `runJob` (`SessionService.swift:1458-1484`) creates a `.job` session and routes through `autoLaunchTerminals` → `launchClaude`, where the gate lives, so the flag lands exactly where intended.

**Concerns:**
- None blocking. `--permission-mode auto` lets unattended jobs run `crow`/`gh`/`git` without per-call approval — an intentional design choice consistent with the existing Manager auto-mode, and the settings copy explains the plan/provider requirements and how to opt out.

### Code Quality
- AppConfig changes are complete and consistent: stored property, init param, `decodeIfPresent(...) ?? true` fallback, and `CodingKeys` entry all present. Both `AppDelegate` sync points (launch + on-save) updated in parallel with the manager flag.
- New tests (round-trip + missing-key-defaults-true) pass — confirmed locally: all 38 `AppConfigTests` green, including `appConfigJobsAutoPermissionModeRoundTrip` and `appConfigJobsAutoPermissionModeDefaultsTrueWhenKeyMissing`.

Minor, non-blocking nits (consider, not required):
- `launchClaude` now resolves `session` at `SessionService.swift:414`, but the closure at `:454-455` still re-fetches the same session from `appState.sessions`. Could reuse the already-resolved `session` to dedupe the lookup.
- The Manager toggle lives in the shared `AutomationSettingsView` component, while this Jobs toggle is an inline `Section` in `jobsTab`. Functionally fine (different tab), just a slight divergence in approach.
- `ConfigStoreTests` has a persistence round-trip for `managerAutoPermissionMode` but none was added for `jobsAutoPermissionMode`. Coverage is transitive through the AppConfig Codable tests, so this is optional.

### Summary Table
| Priority | Issue |
|----------|-------|
| Red | None |
| Yellow | None |
| Green | Dedupe the duplicate session lookup in `launchClaude`; optionally use the shared `AutomationSettingsView` component and add a `ConfigStore` round-trip test for parity |

**Recommendation:** Approve. The change is correct, well-tested, consistent with the established Manager pattern, and introduces no security regression. The remaining items are cosmetic.

PR #197, opened 2026-04-22
Phases A + B + C: agent abstraction + OpenAI Codex MVP

dhilgaertner:
Closes #166, #167, and #168. Parent spec: #150. Lands the full agent-abstraction stack — protocols, per-session selection, and a working second agent (OpenAI Codex).

## Phase A — CodingAgent protocol (behavior-preserving) — closes #166

- New protocols in `CrowCore`: `CodingAgent`, `AgentKind`, `AgentRegistry`, `StateSignalSource`, `AgentStateTransition`, `HookConfigWriter`, plus an `AgentHookEvent` payload type so `CrowCore` stays free of `JSONValue`.
- `ClaudeCodeAgent`, `ClaudeHookConfigWriter` (moved from `Sources/Crow/App/HookConfigGenerator.swift`), and `ClaudeHookSignalSource` in `Packages/CrowClaude/`. Registered at app launch.
- `AppDelegate.hook-event` 110-line state-machine switch becomes ~20 lines of apply-transition code. `ClaudeState` → `AgentActivityState`, `TerminalReadiness.claudeLaunched` → `.agentLaunched`, `onLaunchClaude` → `onLaunchAgent`, `SessionHookState.claudeState` → `activityState`.

## Phase B — per-session selection plumbing — closes #167

- `Session.agentKind: AgentKind` (persisted, backward-compat decoder).
- `AppConfig.defaultAgentKind: AgentKind` (top-level, backward-compat decoder, mirrored to `AppState`).
- `CodingAgent` gains `displayName` and `iconSystemName` so UI enumerates registered agents without baking names into `CrowCore`.
- Picker in `CreateSessionView` and Settings → Defaults; disabled when only one agent registers; auto-enables when a second registers.
- Sidebar leading agent icon, read-only "Agent: …" row in detail header (hidden for Manager).
- `crow new-session` grows `--agent <kind>`; RPC accepts `agent_kind`.
- Manager session hardcoded to `.claudeCode`.

## Phase C — OpenAI Codex MVP — closes #168

Empirically verified against `codex` 0.123.0 (`brew install --cask codex`). Codex uses Claude's hook engine internally — payload schemas are byte-compatible.

**Protocol extensions:**
[...]

PR #403, opened 2026-06-01
Per-workspace AI gateway config (ANTHROPIC_BASE_URL / ANTHROPIC_CUSTOM_HEADERS)

### LINKED ISSUE #402: Workspace-level AI gateway config: per-workspace ANTHROPIC_BASE_URL / ANTHROPIC_CUSTOM_HEADERS for Claude Code launches
## Problem

To route Claude Code through our Corveil gateway, I currently have this in `~/.zshrc`:

```sh
# --- Corveil Commercial (corveil.io) ---
export ANTHROPIC_BASE_URL="https://corveil.io"
export ANTHROPIC_CUSTOM_HEADERS="x-citadel-api-key: Bearer sk-citadel-8809…"
```

That's **global** — every Claude Code session in every workspace, including the Manager and any per-ticket workspace, inherits the same gateway and the same API key. That's wrong as soon as we want to:
- Route different workspaces through different gateways (e.g. a non-Radius workspace shouldn't hit our Corveil gateway at all).
- Use a different API key per workspace (an external client's gateway with their key, for instance).
- Have any workspace at all that uses the *vanilla* Anthropic API (right now, every claude launch on this machine is gated through Corveil).

The fix is to move this off `~/.zshrc` and into the per-workspace config that `crow` already manages.

## Proposed shape

Extend the workspace block in `{devRoot}/.claude/config.json` with an optional `gateway` field:

```jsonc
{
  "workspaces": [
    {
      "id": "B2C2F9BC-…",
      "name": "RadiusMethod",
      "provider": "github",
      "cli": "gh",
      "gateway": {
        "baseURL": "https://corveil.io",
        "customHeaders": {
          "x-citadel-api-key": "Bearer sk-citadel-…"
          // alternative: { "value": "...", "secret_ref": "op://Spotlight Prod/Citadel/api_key" }
        }
      }
    },
    {
      "id": "F82E0C3A-…",
      "name": "Personal",
      "provider": "github",
      "cli": "gh"
      // no gateway block → use vanilla Anthropic API (env unset)
    }
  ]
}
```

When a workspace has a `gateway` block, every `claude` launch that happens "in that workspace" inherits `ANTHROPIC_BASE_URL` and `ANTHROPIC_CUSTOM_HEADERS` derived from it. When it doesn't, the env vars are explicitly **unset** before launch so a global `~/.zshrc` setting (or a sibling workspace's gateway) doesn't bleed in.

## Where the injection points live

1. **`crow-workspace` setup.sh launch path** — `skills/crow-workspace/setup.sh:560`. The `claude` invocation goes through `crow send`. Today it's:
   ```sh
   cd $WORKTREE_PATH && $claude_bin --permission-mode plan$rc_args "$(cat $prompt_path)"
   ```
   It needs to become something like:
   ```sh
   cd $WORKTREE_PATH && \
     ANTHROPIC_BASE_URL=$(posix_quote "$WS_BASE_URL") \
     ANTHROPIC_CUSTOM_HEADERS=$(posix_quote "$WS_CUSTOM_HEADERS") \
     $claude_bin --permission-mode plan$rc_args "$(cat $prompt_path)"
   ```
   …where `$WS_*` is resolved from the workspace's `gateway` block before the `crow send` call. **Empty/unset workspace → use `unset ANTHROPIC_BASE_URL ANTHROPIC_CUSTOM_HEADERS &&` so the inherited shell env doesn't leak through.**

2. **Manager session launch in the Crow app** — `Sources/Crow/App/SessionService.swift:588` (the same site that wires `--permission-mode auto` per `managerAutoPermissionMode`). The Manager session sits at `devRoot` and isn't bound to a single workspace, so this needs a design decision (see Open questions below).

3. **Job-launched sessions** — once crow#390 (jobs auto-permission mode) lands, jobs run a `claude` launch too. They should pick up the gateway of the job's `workspace` field. Same plumbing as #1 if the job ends up using `setup.sh` shape, or the Manager wiring if it's app-side.

## Secret handling (don't punt on this)

`ANTHROPIC_CUSTOM_HEADERS` contains a bearer token. Plaintext in `config.json` is wrong — that file is already tracked in the project's `.claude/` and is otherwise reviewable. Options:

- **`op://` reference** (matching the existing 1Password pattern that `setup.sh` and the codename-spotlight tfvars already use). The app resolves it at session-launch time via the `op` CLI, never persists the resolved value to disk.
- **macOS Keychain** entry keyed by workspace id. The app reads it via the Security framework at launch; never stored in config.json.
- **Worktree-local `.claude/settings.local.json`** (already ignored by per-worktree git exclude per `setup.sh`'s `attributionTrailers` plumbing). Less ideal — the secret still lands on disk in plaintext, just in a gitignored location.

Recommend **`op://` references** as the v1 default since the rest of the stack already speaks that, plus a **macOS Keychain fallback for non-1Password users**. Allow a plain string in `customHeaders.value` for convenience in dev, but warn loudly in the Settings UI that it's stored plaintext.

## Open design questions

1. **What does the Manager session use?** The Manager isn't scoped to one workspace — it sits at `devRoot` and switches between them. Options:
   - (a) **Manager always uses the vanilla Anthropic API** (no gateway). Simple. Then per-workspace gateway only applies to sessions launched *into* a workspace (Crow sessions + Job sessions).
   - (b) **Manager picks up the gateway of the *current* workspace** in the sidebar selection. Cute, but switching workspaces would require restarting the Manager process to re-read env vars — `claude` reads `ANTHROPIC_BASE_URL` once at startup.
   - (c) **Manager has its own global gateway** in a top-level `managerGateway` field on AppConfig. Cleanest, lets you e.g. run the Manager against Corveil but route per-task sessions through whatever the workspace says.
   
   Leaning toward (c) — Manager has its own gateway, per-workspace gateway applies to non-Manager sessions only.

2. **Re-launch story.** After `setup.sh` exits, the user often `Ctrl-C` and re-runs `claude` in the same terminal. The env vars set on the original launch line are gone. Either:
   - Write the env-var block to `.claude/settings.local.json` per-worktree and have Claude Code's `env` setting key pick it up (verify the current Claude Code settings schema supports this).
   - Drop an `.envrc` in the worktree (works if the user has direnv).
   - Have `crow new-terminal` plumb env into the spawned shell's env, not just the launched command, so re-running `claude` inherits.
   
   Probably the third option — the crow tmux backend already manages the shell, and setting env at the `tmux new-window` layer would be inherited by every command in that terminal.

3. **Per-workspace `provider` field already mirrors this pattern.** `provider: "github"` vs `provider: "gitlab"` is per-workspace; a per-workspace `gateway` is the same shape. Use the same Settings UI section ("Per-workspace") and reject configs with a `gateway.baseURL` but no `customHeaders` (or vice versa) at parse time.

## Acceptance

- [ ] `AppConfig.Workspace` gains an optional `gateway: { baseURL: String, customHeaders: [String:String] }` block in `Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift`, with persistence + migration tests in `ConfigStoreTests` (default-nil for existing configs).
- [ ] Settings → per-workspace UI exposes `Base URL` + `Custom Headers` rows, with the secret-storage strategy of choice (`op://` ref / keychain / plaintext-with-warning).
- [ ] `skills/crow-workspace/setup.sh` resolves the workspace's gateway, prepends the env-var assignments to the `claude` launch line, and **explicitly unsets** them when the workspace has no gateway block (so `~/.zshrc` exports don't leak through).
- [ ] Crow Manager session launch (`SessionService.swift:588`) picks up its own gateway via the chosen design from Open Question 1.
- [ ] Re-running `claude` manually in a workspace terminal inherits the same env vars (Open Question 2 design landed).
- [ ] Documentation: `docs/configuration.md` (or wherever the workspace block is documented today) gains the gateway section with the security note about secret storage.
- [ ] My `~/.zshrc` exports can be deleted and the same routing keeps working for the RadiusMethod workspace.

## Out of scope

- Adding a UI for switching gateways mid-session (env is read at `claude` startup; this isn't a "live" knob).
- A registry of "known gateways" (Corveil, etc.) with pre-filled headers — keep v1 generic.
- Gateway health-check / fallback to the vanilla API — that's the gateway's job, not the client's.

## Related

- The existing `customInstructions` per-workspace field (`config.json`) is the closest precedent for this shape.
- The existing `op://` references in `terraform/environments/spotlight-commercial/terraform.tfvars.tpl` are the secret-storage pattern to reuse.
- The Manager-launch wiring touched by crow#390 (jobs auto-permission mode) is the same code path that needs this.

dhilgaertner:
Closes #402

Routes Claude Code through a **per-workspace** AI gateway (`ANTHROPIC_BASE_URL` / `ANTHROPIC_CUSTOM_HEADERS`) instead of a global `~/.zshrc` export, so different workspaces can use different gateways/keys — or the vanilla Anthropic API — and a non-Radius workspace never inherits the Corveil gateway.

## Design decisions

- **Q1 — Manager gateway: (c) its own `managerGateway`.** The Manager sits at devRoot and isn't bound to one workspace; a top-level `managerGateway` mirrors the existing `managerAutoPermissionMode` pattern. Per-workspace `gateway` applies to non-Manager sessions only.
- **Q2 — Re-launch: `settings.local.json` `env` block (not tmux `-e`).** setup.sh already writes `.claude/settings.local.json` (attribution) and the app already merges into it; Claude Code reads its `env` key on **every** run, so this covers initial launch *and* manual `claude` re-runs with no tmux plumbing and **no secret in `ps`/argv**. A launch-line prefix (or `unset … &&` when no gateway) handles the initial-launch override and the no-leak guard. This supersedes the ticket's tmux `-e` option, which would put the token in process args and lose to `~/.zshrc` sourcing anyway.
- **Q3 — Validation: parse-time rejection.** `WorkspaceGateway` decode throws on a half-filled block (baseURL xor customHeaders); the Settings UI blocks save on the same condition so a rejectable block is never written.

## Secret handling

Default is **`op://` reference + plaintext-with-warning**; **Keychain deferred** behind a clean `GatewayResolver` seam (you're a 1Password shop, and `op://` already keeps nothing-secret-at-rest). To flip:
- **`op://`** (recommended) — header value like `op://Vault/Item/field`, resolved at launch via `op read` (app-side and in setup.sh). Nothing secret in `config.json`.
- **plaintext** — any non-`op://` value; stored in `config.json` (0600) with a red warning in Settings.
- A failed `op` lookup drops that header (redacted log) and keeps the baseURL, so the gateway returns a loud 401 rather than silently falling back to vanilla. Resolved secrets are never logged.

## What changed (commit per phase)

1. `WorkspaceGateway` model + `WorkspaceInfo.gateway` + `AppConfig.managerGateway`, validating decode, tests.
2. `GatewayResolver` — resolve `op://`, serialize `Name: Value` headers, tests.
3. setup.sh — resolve gateway via jq/`op read`, merge `env` into settings.local.json, launch-line prefix / `unset`; `setup_gateway_test.sh`.
4. App-side — `HookConfigGenerator.writeGatewayEnv`, `ClaudeLaunchArgs.gatewayEnvPrefix`, `launchClaude` (work/review/job) + Manager wiring.
5. Settings UI — per-workspace "AI Gateway" + "Manager AI Gateway" editors.
6. Docs — `docs/configuration.md` AI Gateway section.

## Tests

- `swift test` green: CrowCore (212), CrowPersistence (28), root app target (145).
- `bash skills/crow-workspace/setup_gateway_test.sh` — 15/15 (op:// resolution via a fake `op`, settings merge, both prefix forms).
- `make build` clean.

## Manual verification (for the reviewer / @dustinhilgaertner to run)

1. Set the RadiusMethod workspace's gateway (Base URL `https://corveil.io`, header `x-citadel-api-key: op://…`) in Settings → Workspaces.
2. Comment out the `ANTHROPIC_BASE_URL` / `ANTHROPIC_CUSTOM_HEADERS` exports in `~/.zshrc`.
3. Launch a RadiusMethod workspace session → confirm it reaches Corveil; launch a no-gateway workspace → confirm vanilla Anthropic.
4. **Precedence check (one open item):** with a dummy `ANTHROPIC_BASE_URL` still in `~/.zshrc` and a different one in the worktree's `settings.local.json` `env`, confirm which wins. The launch-line prefix makes the override correct regardless, but this confirms whether settings `env` alone suffices.

## Coordination

crow#390 (jobs auto-permission mode) is **already merged/closed** — this is independent, no rebase. Job sessions pick up their workspace's gateway through the same `launchClaude` path.

### DIFF
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
@@ -92,6 +92,66 @@ public struct ClaudeHookConfigWriter: HookConfigWriter {
         try data.write(to: URL(fileURLWithPath: settingsPath))
     }
 
+    // MARK: - Gateway env
+
+    /// Keys we manage inside the settings `env` block (CROW-402).
+    private static let gatewayEnvKeys = ["ANTHROPIC_BASE_URL", "ANTHROPIC_CUSTOM_HEADERS"]
+
+    /// Write (or clear) the AI-gateway env vars in a directory's
+    /// `.claude/settings.local.json` `env` block, merging with existing settings.
+    ///
+    /// Claude Code reads this `env` block on every launch, so this makes the
+    /// gateway survive manual `claude` re-runs in the terminal — not just the
+    /// initial launch (CROW-402). Pass a resolved gateway to set the vars, or
+    /// `nil` to remove them (so switching a workspace off its gateway clears the
+    /// stale values rather than leaving them behind).
+    ///
+    /// `dirPath` is the worktree path for work/job/review sessions, or the dev
+    /// root for the Manager session.
+    public static func writeGatewayEnv(dirPath: String, resolved: GatewayResolver.Resolved?) {
+        let claudeDir = (dirPath as NSString).appendingPathComponent(".claude")
+        let settingsPath = (claudeDir as NSString).appendingPathComponent("settings.local.json")
+
+        // Read existing settings if present.
+        var settings: [String: Any] = [:]
+        if let data = FileManager.default.contents(atPath: settingsPath),
+           let parsed = try? JSONSerialization.jsonObject(with: data) as? [String: Any] {
+            settings = parsed
+        }
+
+        var env = settings["env"] as? [String: Any] ?? [:]
+        if let resolved {
+            env["ANTHROPIC_BASE_URL"] = resolved.baseURL
+            env["ANTHROPIC_CUSTOM_HEADERS"] = resolved.customHeaders
+        } else {
+            for key in gatewayEnvKeys { env.removeValue(forKey: key) }
+        }
+
+        if env.isEmpty {
+            settings.removeValue(forKey: "env")
+        } else {
+            settings["env"] = env
+        }
+
+        // Nothing to write and no file to clean up.
+        if settings.isEmpty && !FileManager.default.fileExists(atPath: settingsPath) {
+            return
+        }
+
+        do {
+            try FileManager.default.createDirectory(atPath: claudeDir, withIntermediateDirectories: true)
+            let data = try JSONSerialization.data(withJSONObject: settings, options: [.prettyPrinted, .sortedKeys])
+            try data.write(to: URL(fileURLWithPath: settingsPath))
+            // The env block can carry a resolved bearer token, so restrict the
+            // file to owner-only — matching ConfigStore's 0600 on config.json.
+            try? FileManager.default.setAttributes(
+                [.posixPermissions: 0o600], ofItemAtPath: settingsPath)
+        } catch {
+            NSLog("[ClaudeHookConfigWriter] Failed to write gateway env to %@: %@",
+                  settingsPath, error.localizedDescription)
+        }
+    }
+
     /// Remove our hook entries from a worktree's settings.local.json, preserving user settings.
     public func removeHookConfig(worktreePath: String) {
         let settingsPath = (worktreePath as NSString)
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeLaunchArgs.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeLaunchArgs.swift
@@ -1,4 +1,5 @@
 import Foundation
+import CrowCore
 
 /// Helpers for building the argument string appended to a `claude` shell invocation.
 ///
@@ -39,5 +40,37 @@ public enum ClaudeLaunchArgs {
             }
         }
         return s
+    }
+
+    /// Shell prefix that applies (or clears) the AI-gateway env vars on the
+    /// `claude` launch line (CROW-402). Placed immediately before the `claude`
+    /// binary path so it overrides any value exported by the user's `~/.zshrc`
+    /// for this invocation. Re-runs are covered separately by the
+    /// `settings.local.json` `env` block, so this is the initial-launch override
+    /// and the load-bearing no-leak guard.
+    ///
+    /// Uses `export … &&` (not bare `VAR=val` command-prefix assignments) so it
+    /// composes correctly in front of the OTEL `export … &&` prefix that
+    /// `ClaudeCodeAgent.autoLaunchCommand` bakes into the launch string — a bare
+    /// `VAR=val` prefix would bind only to that following `export` builtin, not to
+    /// the eventual `claude` process.
+    ///
+    /// - `resolved` present → `export ANTHROPIC_BASE_URL='…' ANTHROPIC_CUSTOM_HEADERS='…' && `
+    /// - `resolved` nil → `unset ANTHROPIC_BASE_URL ANTHROPIC_CUSTOM_HEADERS && `
+    ///   so a no-gateway workspace doesn't inherit a sibling's or `~/.zshrc`'s gateway.
+    /// - multi-header → the header value has an embedded newline and can't go on
+    ///   the line (a pasted newline would submit the command early), so it's
+    ///   carried solely by `settings.local.json`. We still `unset ANTHROPIC_CUSTOM_HEADERS`
+    ///   before exporting `ANTHROPIC_BASE_URL`, so the gateway's baseURL is never
+    ///   paired with a stale `~/.zshrc`-inherited header value.
+    public static func gatewayEnvPrefix(_ resolved: GatewayResolver.Resolved?) -> String {
+        guard let resolved else {
+            return "unset ANTHROPIC_BASE_URL ANTHROPIC_CUSTOM_HEADERS && "
+        }
+        let baseAssignment = "export ANTHROPIC_BASE_URL=\(shellQuote(resolved.baseURL))"
+        if resolved.customHeaders.contains("\n") {
+            return "unset ANTHROPIC_CUSTOM_HEADERS && " + baseAssignment + " && "
+        }
+        return baseAssignment + " ANTHROPIC_CUSTOM_HEADERS=\(shellQuote(resolved.customHeaders)) && "
     }
 }
--- a/Packages/CrowClaude/Tests/CrowClaudeTests/ClaudeLaunchArgsTests.swift
+++ b/Packages/CrowClaude/Tests/CrowClaudeTests/ClaudeLaunchArgsTests.swift
@@ -1,5 +1,6 @@
 import Foundation
 import Testing
+import CrowCore
 @testable import CrowClaude
 
 @Test func claudeLaunchArgsDisabledReturnsEmpty() {
@@ -50,4 +51,41 @@ import Testing
     #expect(ClaudeLaunchArgs.argsSuffix(remoteControl: true, sessionName: "Manager")
         == " --rc --name 'Manager'")
     #expect(ClaudeLaunchArgs.argsSuffix(remoteControl: false, sessionName: nil) == "")
+}
+
+// MARK: - Launch-line gateway prefix (ClaudeLaunchArgs.gatewayEnvPrefix, CROW-402)
+
+@Test func gatewayEnvPrefixUnsetsWhenNil() throws {
+    // No gateway → explicitly unset so a global ~/.zshrc export (or a sibling
+    // workspace's gateway) can't bleed into this launch.
+    #expect(ClaudeLaunchArgs.gatewayEnvPrefix(nil) == "unset ANTHROPIC_BASE_URL ANTHROPIC_CUSTOM_HEADERS && ")
+}
+
+@Test func gatewayEnvPrefixExportsSingleHeader() throws {
+    // Single header → both vars go on the launch line via `export … &&` so they
+    // compose in front of any OTEL `export … &&` prefix and reach `claude`.
+    let resolved = GatewayResolver.Resolved(
+        baseURL: "https://corveil.io",
+        customHeaders: "x-citadel-api-key: Bearer sk-1"
+    )
+    let prefix = ClaudeLaunchArgs.gatewayEnvPrefix(resolved)
+    #expect(prefix == "export ANTHROPIC_BASE_URL='https://corveil.io' ANTHROPIC_CUSTOM_HEADERS='x-citadel-api-key: Bearer sk-1' && ")
+    #expect(!prefix.contains("\n"))
+}
+
+@Test func gatewayEnvPrefixUnsetsInheritedHeadersForMultiLine() throws {
+    // A multi-header value has an embedded newline; pasting it onto the launch
+    // line would submit the command early, so it's carried by settings.local.json.
+    // The prefix must still `unset ANTHROPIC_CUSTOM_HEADERS` so the gateway's
+    // baseURL isn't paired with a stale ~/.zshrc-inherited header value, and must
+    // not contain a raw newline.
+    let resolved = GatewayResolver.Resolved(
+        baseURL: "https://corveil.io",
+        customHeaders: "x-a: one\nx-b: two"
+    )
+    let prefix = ClaudeLaunchArgs.gatewayEnvPrefix(resolved)
+    #expect(prefix == "unset ANTHROPIC_CUSTOM_HEADERS && export ANTHROPIC_BASE_URL='https://corveil.io' && ")
+    #expect(prefix.contains("unset ANTHROPIC_CUSTOM_HEADERS"))
+    #expect(prefix.contains("export ANTHROPIC_BASE_URL='https://corveil.io'"))
+    #expect(!prefix.contains("\n"))
 }
--- a/Packages/CrowCore/Sources/CrowCore/GatewayResolver.swift
+++ b/Packages/CrowCore/Sources/CrowCore/GatewayResolver.swift
@@ -1,0 +1,116 @@
+import Foundation
+
+/// Resolves a `WorkspaceGateway` into launch-ready environment-variable values
+/// for a `claude` invocation (CROW-402).
+///
+/// A header value may be a plaintext string or an `op://…` 1Password reference.
+/// References are resolved at launch via the `op` CLI so the secret never lands
+/// at rest in `config.json`; any other value is used literally. The serialized
+/// output matches Claude Code's contract: `ANTHROPIC_BASE_URL` is the gateway
+/// endpoint and `ANTHROPIC_CUSTOM_HEADERS` is newline-separated `Name: Value`
+/// header lines.
+///
+/// Resolved secret values are never logged.
+public enum GatewayResolver {
+    /// Launch-ready env values derived from a gateway.
+    public struct Resolved: Equatable, Sendable {
+        /// Value for `ANTHROPIC_BASE_URL`.
+        public var baseURL: String
+        /// Value for `ANTHROPIC_CUSTOM_HEADERS` — newline-separated `Name: Value`.
+        public var customHeaders: String
+
+        public init(baseURL: String, customHeaders: String) {
+            self.baseURL = baseURL
+            self.customHeaders = customHeaders
+        }
+    }
+
+    /// Serialize a resolved header map into the `ANTHROPIC_CUSTOM_HEADERS` value:
+    /// newline-separated `Name: Value`, sorted by name for deterministic output.
+    public static func serializeHeaders(_ headers: [String: String]) -> String {
+        headers
+            .sorted { $0.key < $1.key }
+            .map { "\($0.key): \($0.value)" }
+            .joined(separator: "\n")
+    }
+
+    /// Resolve a gateway's header values (resolving `op://…` references) and
+    /// serialize them for launch. Returns `nil` for an empty gateway (caller
+    /// should then *unset* the env vars rather than set them).
+    ///
+    /// - Parameter resolveSecret: Injected for testability; defaults to `op read`.
+    ///   When a reference fails to resolve, the header is dropped and a redacted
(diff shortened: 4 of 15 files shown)


dgershman (changes requested):
## Code & Security Review

Solid, well-tested implementation overall — model / resolver / launch-prefix / merge writer / setup.sh helpers are cleanly separated, parse-time validation rejects half-filled gateways, and `op://` resolution failures fail loud (gateway rejects) rather than silently falling back to vanilla. Round-trip and bash tests are thorough.

Two **Yellow** findings around the no-leak guarantee and at-rest secret protection that I'd like addressed in this round.

### Yellow Findings

**Y1 — Multi-header launch line leaves a `~/.zshrc` leak path open.** `ClaudeLaunchArgs.gatewayEnvPrefix` (Packages/CrowCore/Sources/CrowCore/ClaudeLaunchArgs.swift:59-68) and `gateway_launch_prefix` in setup.sh:148-157 both intentionally omit `ANTHROPIC_CUSTOM_HEADERS` from the launch line when there is more than one header (to avoid embedding a literal newline that would submit the command early). In that branch the prefix becomes just `ANTHROPIC_BASE_URL='…' claude …`, which is a command-prefix assignment — it sets `ANTHROPIC_BASE_URL` for the launched process but does **not** clear an `ANTHROPIC_CUSTOM_HEADERS` that the shell inherited from `~/.zshrc`. Claude Code then sees the gateway's `baseURL` paired with the user's stale global headers (or no headers, depending on precedence with `settings.local.json` `env`). The PR description's "load-bearing no-leak guard" claim doesn't hold in the multi-header case.

**Why:** the PR explicitly motivates the launch-line prefix as "the load-bearing no-leak guard"; this is the one branch where it isn't.

**How to apply:** in the multi-header branch, prepend `unset ANTHROPIC_CUSTOM_HEADERS && ` (or wrap the whole launch in `env -u ANTHROPIC_CUSTOM_HEADERS …`) before the `ANTHROPIC_BASE_URL=…` assignment, so the inherited shell var is removed before claude starts. Cover it with a test that asserts both `unset ANTHROPIC_CUSTOM_HEADERS` and `ANTHROPIC_BASE_URL=…` appear in the multi-header prefix.

**Y2 — Resolved secrets persisted to `settings.local.json` with default file permissions.** The PR description / Settings UI describe `op://` as the "nothing-secret-at-rest" mode, but the resolved secret value is written into `.claude/settings.local.json`'s `env` block on every launch:
- `HookConfigGenerator.writeGatewayEnv` (Sources/Crow/App/HookConfigGenerator.swift:109-147) writes via `data.write(to:)` without setting permissions.
- `setup.sh:write_settings_local` (skills/crow-workspace/setup.sh:431-502) writes via `printf > "$settings_path"` without `chmod 600`.

`ConfigStore.saveConfig` already sets `0600` on `config.json` for the same reason. `settings.local.json` now carries the same sensitivity (the post-`op read` plaintext bearer token) but inherits the umask default (typically `0644`).

**Why:** the design intent is that `op://` keeps secrets off disk; in practice the resolved value lands on disk in a file with weaker protection than `config.json`.

**How to apply:** when the merged settings dict contains `env.ANTHROPIC_CUSTOM_HEADERS` (or any non-empty `env` block), `chmod 0600` the file after writing — both in `HookConfigGenerator.writeGatewayEnv` (use `FileManager.default.setAttributes([.posixPermissions: 0o600], ofItemAtPath: settingsPath)`) and in `setup.sh:write_settings_local` (`chmod 600 "$settings_path"`). Additionally, soften the docs / Settings UI copy: `op://` means "nothing secret in `config.json`," not "nothing secret on disk" — the resolved value is cached in `settings.local.json` for the worktree's lifetime.

### Green / Nits (consider, not blocking)

- **G1 — The "precedence open item" in the PR description is real.** With `ANTHROPIC_BASE_URL` in `~/.zshrc` and a different one in `settings.local.json` `env`, the test suite doesn't pin down which wins. Worth adding a Claude-Code-side smoke test (or at minimum a doc note in `docs/configuration.md#ai-gateway`) once you verify it manually, so the next person doesn't have to re-derive it.
- **G2 — `is_remote_control_enabled` / `is_attribution_trailers_enabled` use `tr -d '\n' | grep -qE` instead of `jq`** (skills/crow-workspace/setup.sh:76-95). Pre-existing, not introduced here, but worth a follow-up to converge on the jq-based path used by `resolve_gateway_env`.
- **G3 — `op read` 15s timeout** (Packages/CrowCore/Sources/CrowCore/GatewayResolver.swift:94-104) is reasonable for biometric prompts; could be made configurable later if Touch ID prompts feel sluggish.

### Security Review

**Strengths:**
- `posix_quote` in setup.sh correctly handles single-quote escaping; matches `ClaudeLaunchArgs.shellQuote`.
- Resolved secret values are never logged (NSLog messages drop the header value, only naming the header key).
- Failed `op read` drops the header rather than falling back silently — gateway 401 surfaces the misconfig.
- `gateway_launch_prefix` uses single-quoted assignments, so a hostile `baseURL` value can't break out of quoting.
- Parse-time rejection of half-filled gateways prevents the "baseURL without auth" footgun.
- `op read` invocation uses `$value` quoted, so an `op://…` string with shell metachars can't inject.

**Concerns:**
- Y1 and Y2 above.

### Code Quality

- Tests cover the model, persistence, resolver (incl. failure paths), the launch prefix builder, the bash helpers (15/15 passing locally), and UI form validation. Good.
- Backward compat is correct — every new field uses `decodeIfPresent` with a default; legacy configs decode unchanged.
- `WorkspaceGateway.parseHeaderLines` / `headerLines` form a clean round-trip for the editor.
- `GatewayResolver` is dependency-injected (`resolveSecret:`) so the resolver itself is testable without `op`.
- Single source for the workspace-name-from-worktree-path math (`SessionService.workspaceName`), good for testability.
- Settings UI surfaces validation inline and blocks save on a half-filled block — matches the parse-time invariant.

### Summary Table

**Recommendation:** Request Changes — driven by 0 Red, 2 Yellow, 3 Green findings. The two Yellows are scoped and should be quick to land in the same round trip.

---

dhilgaertner:
Thanks for the review — both Yellows addressed in `0f0a7e3`.

**Y1 — multi-header leak path (fixed).** The multi-header branch now `unset`s the inherited header before the baseURL assignment, in both places:
- `ClaudeLaunchArgs.gatewayEnvPrefix` → `unset ANTHROPIC_CUSTOM_HEADERS && ANTHROPIC_BASE_URL='…' `
- setup.sh `gateway_launch_prefix` → same shape.

So the gateway's `baseURL` is never paired with a stale `~/.zshrc` header value; the real (multi-line) header value still comes from `settings.local.json`. Tests assert both `unset ANTHROPIC_CUSTOM_HEADERS` and `ANTHROPIC_BASE_URL=…` appear (and no embedded newline) — `gatewayEnvPrefixUnsetsInheritedHeadersForMultiLine` (Swift) and the bash test's multi-header case.

**Y2 — 0600 on settings.local.json (fixed).** Both writers now restrict the file to owner-only after writing, matching `ConfigStore`'s `0600` on `config.json`:
- `HookConfigGenerator.writeGatewayEnv` → `setAttributes([.posixPermissions: 0o600], …)`
- setup.sh `write_settings_local` → `chmod 600`.

Added a `stat`-based permission assertion to the bash test. Also softened the docs + Settings UI copy: `op://` means "kept out of `config.json`," **not** "no secret on disk" — the resolved value is cached `0600` in the worktree's `settings.local.json` for its lifetime.

**G1 — precedence (doc note added).** Added a precedence note to `docs/configuration.md#ai-gateway`: the launch-line assignment is what reliably overrides a shell export; the intended end state is to delete the global `~/.zshrc` exports so `config.json` is the single source of truth. I'll pin down the settings-`env`-vs-shell precedence in the manual smoke test and fold the result into that note.

**G2 / G3 — acknowledged, deferred.** G2 (converging `is_remote_control_enabled` / `is_attribution_trailers_enabled` onto the jq path) is pre-existing and out of scope for this PR — happy to do it as a follow-up. G3 (configurable `op read` timeout) noted for if Touch ID prompts feel sluggish.

All suites green: CrowCore 212, root 145, `setup_gateway_test.sh` 17/17, `make build` clean.

dgershman (approved):
## Code & Security Review

Re-review after the fix commit (0f0a7e3). Both Yellow findings from the prior review and G1 are closed cleanly; no new Red/Yellow findings emerged.

### Verification of prior findings

- **Y1 — multi-header `~/.zshrc` leak path is closed.** `ClaudeLaunchArgs.gatewayEnvPrefix` (Packages/CrowCore/Sources/CrowCore/ClaudeLaunchArgs.swift:60-69) and `gateway_launch_prefix` (skills/crow-workspace/setup.sh:150-161) both now emit `unset ANTHROPIC_CUSTOM_HEADERS && ANTHROPIC_BASE_URL='…' ` in the multi-header branch. `gatewayEnvPrefixUnsetsInheritedHeadersForMultiLine` (GatewayResolverTests.swift:77-92) and the bash test ("multi-header prefix unsets inherited headers", "multi-header prefix omits headers assignment", "launch prefix has no embedded newline") pin the contract.
- **Y2 — resolved secrets at-rest now owner-only.** `HookConfigGenerator.writeGatewayEnv` (Sources/Crow/App/HookConfigGenerator.swift:143-146) calls `FileManager.setAttributes([.posixPermissions: 0o600], …)` after every write. `setup.sh:write_settings_local` (skills/crow-workspace/setup.sh:487-489) `chmod 600`s the file. The bash test asserts the mode is `600`.
- **G1 — precedence note** is in docs/configuration.md (note at lines 100-102): the launch-line assignment is the reliable override, and the end state is to delete the global `~/.zshrc` exports. Docs/UI copy also softened — `op://` is now described as "kept out of config.json," not "no secret on disk."

### Tests

- `swift test --filter GatewayResolverTests` — 8/8 green locally.
- `bash skills/crow-workspace/setup_gateway_test.sh` — 17/17 green locally (the new 0600 assertion + the two new multi-header assertions land).

### Security Review

**Strengths (re-confirmed):**
- `posix_quote` / `shellQuote` POSIX-escape correctly; hostile `baseURL` values can't break quoting.
- Resolved secret values are still never logged (NSLog drops the value, only naming the header key).
- Failed `op read` drops the header rather than falling back silently — the gateway returns a loud 401.
- Parse-time rejection of half-filled gateways stops "baseURL without auth" at decode.
- Both `writeGatewayEnv` and `write_settings_local` now match `ConfigStore.saveConfig`'s 0600 stance on the at-rest secret.

**Concerns:** None blocking.

### Code Quality

- Behavior change is minimal and well-targeted; doc comments on both prefix builders accurately describe all three branches (nil / single-header / multi-header).
- `try?` on `setAttributes` in `writeGatewayEnv` mirrors existing best-effort logging patterns elsewhere.
- Backward compat unchanged — every new field still uses `decodeIfPresent` with a default.

### Green / Nits (consider, not blocking)

- **G2 (carried over) — `is_remote_control_enabled` / `is_attribution_trailers_enabled`** still use `tr | grep` instead of `jq`. Pre-existing; worth a follow-up to converge on the jq-based path used by `resolve_gateway_env`.
- **G3 (carried over) — `op read` 15s timeout** is reasonable; could be made configurable later if Touch ID prompts ever feel sluggish.
- **G4 — brief perm race on `settings.local.json`.** In `launchClaude`, `writeHookConfig` runs before `writeGatewayEnv`; its atomic write recreates the file with default umask (~0644) before `writeGatewayEnv` re-applies 0600. The window is microseconds on a single-user macOS app, end state is 0600, and no exploit is plausible — but `writeHookConfig` could also set 0600 unconditionally to close it.
- **G5 — `AutomationSettingsView.commitManagerGateway` leaves the previous valid gateway in place when the form goes half-filled.** The inline red error tells the user, and the form has no Save button to commit a bad state. UX nit, not security.

### Summary Table

**Recommendation:** Approve — driven by 0 Red, 0 Yellow, 5 Green findings. The two prior Yellows (Y1 launch-line leak, Y2 0600 at-rest) and G1 (precedence note) all landed cleanly with test coverage.

---

dhilgaertner:
Rebased onto `main` to clear the merge conflict — `main` had absorbed #197 (agent abstraction + OpenAI Codex MVP), which moved several files this PR depended on. Net behavior is unchanged; the gateway logic was re-homed to the new structure:

- **`ClaudeLaunchArgs` moved `CrowCore` → `CrowClaude`.** `gatewayEnvPrefix` moved with it (and now `import CrowCore` for `GatewayResolver.Resolved`).
- **`HookConfigGenerator` (app target) was deleted in favour of `CrowClaude/ClaudeHookConfigWriter`.** `writeGatewayEnv` moved there (now `public`, keeps the `0600` hardening).
- **Launch path is now the `CodingAgent` abstraction.** `SessionService.launchClaude` → `launchAgent`, which builds the command via `agent.autoLaunchCommand(...)`. Gateway resolution + the `settings.local.json` env write are gated to `agent.kind == .claudeCode` (ANTHROPIC_* is Claude-specific; Codex sessions are untouched), and the launch-line prefix is prepended at the send site.
- **`gatewayEnvPrefix` now emits `export … &&` instead of bare `VAR=val …`.** The agent bakes an OTEL `export … && claude` prefix into the command, and a bare command-prefix assignment would bind only to that `export` builtin rather than `claude`. `export … &&` composes correctly in front of it. (`setup.sh` keeps the bare form — its launch line has nothing between the prefix and `claude`.)
- The `gatewayEnvPrefix` unit tests moved from `CrowCoreTests` to `CrowClaudeTests` (they can't `import CrowClaude` from CrowCore), updated for the `export` form.

Verification on the rebased branch: `make build` clean; CrowCore 210, CrowClaude 40, CrowPersistence 28, root 145, `setup_gateway_test.sh` 17/17 — all green.

PR #405, opened 2026-06-02
Fix UI beachball when a review session spins up

### LINKED ISSUE #404: Crow UI locks up with macOS spinning beachball when a review spins up
## Problem

When a review spins up in Crow (presumably the cloud `ultrareview` / auto-review flow kicking off), the Crow app freezes — the macOS spinning beachball ("blue spinning ball") appears and the UI is completely unresponsive. No clicks register, sessions can't be switched, terminals can't be sent to, until the spinner clears (or the app has to be force-quit).

## Reproduction (best guess — to be confirmed)

1. Open Crow.
2. Trigger a review (e.g. ultrareview kicking off on a branch, or an auto-review starting against an open PR).
3. Observe: Crow main window becomes unresponsive with the macOS spinning beachball. No interaction possible during the freeze.

## Expected

Review spin-up should run off the main thread / out-of-process. The UI must stay responsive — at most a localized loading indicator on the affected session card, not a global app freeze.

## Likely cause

A synchronous blocking call on the main actor / main thread during review kickoff — e.g. spawning a long-running shell command, hitting the network, or doing heavy JSON work inline in an event handler instead of dispatching to a background task and awaiting.

## What to investigate

- Whichever code path handles review spin-up (ultrareview, auto-review, the review watcher): look for synchronous `Process` / `URLSession` / file I/O calls running on the main actor.
- Check the watcher tick that fires on `autoReviewWatcherEnabled` / `autoCreateWatcherEnabled` config — does it block while doing repo scans or git calls?
- Check the JSON store / persistence write path — large `mutate()` calls on the main thread under NSLock could stall the UI if the disk write is slow.

## Acceptance

- Spinning up a review never produces a macOS beachball / global app freeze.
- The user can keep switching sessions, opening terminals, and clicking around while a review is initializing.
- If something legitimately needs to block, it's scoped to the affected card with a visible spinner — not the whole app.

dgershman:
## Summary

- Review-session kickoff was synchronously blocking the main actor on every `gh`/`git` step, producing the macOS spinning beachball for the duration of the worst step (often a full `gh repo clone`).
- `SessionService` is `@MainActor`, so even though `createReviewSession()` is `async`, the old `shell()` helper used `Process.waitUntilExit()` — which blocked the main thread instead of yielding.
- Two surgical fixes restore UI responsiveness throughout review kickoff.

### Fix 1 — `shell()` truly yields
Replaced `process.waitUntilExit()` with `withCheckedThrowingContinuation` + `process.terminationHandler`, and marked `shell()` (and a new `runShellAsync` static) `nonisolated`. Now `await shell(...)` actually suspends and frees the main actor while the subprocess runs in the background.

### Fix 2 — `createReviewSession()` runs git/file-prep off-main
Extracted the `gh pr view` / `gh repo clone` / `git fetch/checkout/pull` / prompt-write / skill-copy block into a new `nonisolated static prepareReviewClone(...)` helper invoked via `Task.detached(priority: .userInitiated)`. The main-actor tail only handles the cheap state mutations and the `prepareTerminal` call (still bounded by the existing 2s tmux watchdog).

The kickoff serialization queue in `AppDelegate.enqueueReviewKickoff` is unchanged — multiple PRs still process in order; the `await`s just no longer wedge the UI.

Closes #404

## Test plan

- [x] `arch -arm64 swift test --arch arm64` — 145 tests pass
- [ ] Trigger a review on a PR whose repo is NOT yet cloned (worst-case full-clone path). Click around the sidebar / switch sessions during kickoff — no beachball, no input lag.
- [ ] Trigger a review on a PR whose repo IS already cloned (fast path). Sidebar stays responsive; new review session appears when ready.
- [ ] Enable `autoReviewWatcherEnabled`, push a commit to an open PR, observe the watcher fire a kickoff — UI stays responsive.
- [ ] Burst test: enqueue 3+ PRs in quick succession via `enqueueReviewKickoff`. UI stays responsive; sessions appear in deterministic order (queue serialization preserved).

### DIFF
--- a/Sources/Crow/App/SessionService.swift
+++ b/Sources/Crow/App/SessionService.swift
@@ -814,26 +814,64 @@ final class SessionService {
     // MARK: - Worktree Safety Checks
     // Protected branch and main-checkout detection are centralized on SessionWorktree in CrowCore.
 
-    private func shell(env: [String: String] = [:], _ args: String...) async throws -> String {
-        let process = Process()
-        let stdoutPipe = Pipe()
-        let stderrPipe = Pipe()
-        process.executableURL = URL(fileURLWithPath: "/usr/bin/env")
-        process.arguments = args
-        process.environment = env.isEmpty
+    /// Run `/usr/bin/env <args...>` and return stdout. Marked `nonisolated` and
+    /// implemented via `withCheckedThrowingContinuation` + `terminationHandler`
+    /// so `await shell(...)` truly suspends the calling task instead of
+    /// blocking on `waitUntilExit()`. This is what keeps the main actor free
+    /// during review-session kickoff (#404); the prior implementation pinned
+    /// every git/gh call to the main thread.
+    nonisolated private func shell(env: [String: String] = [:], _ args: String...) async throws -> String {
+        let resolvedEnv = env.isEmpty
             ? ShellEnvironment.shared.env
             : ShellEnvironment.shared.merging(env)
-        process.standardOutput = stdoutPipe
-        process.standardError = stderrPipe
-        try process.run()
-        process.waitUntilExit()
-        let stdout = String(data: stdoutPipe.fileHandleForReading.readDataToEndOfFile(), encoding: .utf8) ?? ""
-        let stderr = String(data: stderrPipe.fileHandleForReading.readDataToEndOfFile(), encoding: .utf8) ?? ""
-        guard process.terminationStatus == 0 else {
-            throw NSError(domain: "SessionService", code: Int(process.terminationStatus),
-                          userInfo: [NSLocalizedDescriptionKey: stderr])
+        return try await Self.runShellAsync(env: resolvedEnv, args: args)
+    }
+
+    /// Shared async Process runner. Hands ownership of the pipes/process to
+    /// the termination handler so reads happen after exit (no deadlock from
+    /// a full pipe blocking the child) and the continuation is resumed
+    /// exactly once.
+    nonisolated static func runShellAsync(env: [String: String], args: [String]) async throws -> String {
+        try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<String, Error>) in
+            let process = Process()
+            let stdoutPipe = Pipe()
+            let stderrPipe = Pipe()
+            process.executableURL = URL(fileURLWithPath: "/usr/bin/env")
+            process.arguments = args
+            process.environment = env
+            process.standardOutput = stdoutPipe
+            process.standardError = stderrPipe
+
+            process.terminationHandler = { proc in
+                let stdout = String(
+                    data: stdoutPipe.fileHandleForReading.readDataToEndOfFile(),
+                    encoding: .utf8
+                ) ?? ""
+                let stderr = String(
+                    data: stderrPipe.fileHandleForReading.readDataToEndOfFile(),
+                    encoding: .utf8
+                ) ?? ""
+                if proc.terminationStatus == 0 {
+                    continuation.resume(returning: stdout)
+                } else {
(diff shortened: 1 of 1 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Summary
The fix correctly addresses the main-actor beachball (#404) with two well-scoped changes:

1. **`shell()` truly yields** — replacing `waitUntilExit()` with `withCheckedThrowingContinuation` + `terminationHandler`, and marking it `nonisolated`. `shell()` uses no instance state, so dropping main-actor isolation is safe; all existing callers (`worktree list`, `remote get-url`, `gh issue view`, `gh pr view`, `repo clone`) still `await` it correctly.
2. **`prepareReviewClone` off-main** — the `gh`/`git`/file-prep block moves into a `nonisolated static` helper run via `Task.detached(.userInitiated)`, leaving only cheap state mutations on the main actor. `env` is captured as `ShellEnvironment` (`public final class … : Sendable`), and `Scaffolder.bundled*` / `buildReviewPrompt` are plain non-isolated statics, so the off-main hop is concurrency-clean.

### Security Review
**Strengths:**
- No new external input surface; PR URL parsing and command args are unchanged from the prior implementation. No shell string interpolation — args are passed as an array to `/usr/bin/env`, so no command-injection vector is introduced.
- Error-message construction now prefers `stderr` and falls back to `stdout`; neither is surfaced to users beyond `NSLog`, consistent with prior behavior.

**Concerns:** None.

### Code Quality
- **Continuation safety is correct.** `terminationHandler` is installed *before* `process.run()`, so the success path resumes exactly once. The `catch` on launch failure defensively clears `terminationHandler = nil` before resuming, preventing a double-resume — a nice touch that matches the documented macOS launch-failure edge case.
- **Behavioral parity preserved.** PR-metadata fetch failure aborts kickoff (`throw` → caught → `return nil` + `NSLog`), matching the old guard; `git fetch/checkout/pull` and `clone` errors remain tolerated via `try?`. Kickoff serialization in `AppDelegate.enqueueReviewKickoff` is untouched, so PR ordering is preserved.

### Findings

**Green — pipe read-after-exit comment is misleading (Sources/Crow/App/SessionService.swift:830-833, 845-853).** The doc comment claims reads-after-exit means "no deadlock from a full pipe blocking the child." That reasoning is backwards: draining `readDataToEndOfFile` only *after* `terminationHandler` fires is precisely the pattern that can deadlock if a child writes more than the ~64KB pipe buffer to stdout/stderr and then blocks on `write()`, never exiting — so the handler never fires and the continuation never resumes. This is **not a regression** (the old `waitUntilExit()` path had the identical property) and is harmless for the current callers, whose output is small (`gh`/`git` to a non-tty emit minimal progress). But since `runShellAsync` is now a reusable `static` helper, consider either correcting the comment or draining the pipes concurrently (e.g. `readabilityHandler`/background reads) before reusing it for high-volume commands like `gh pr diff` on a large PR.

**Green — no direct unit coverage for `runShellAsync`.** Subprocess runners are awkward to unit-test, and the existing 145-test suite exercises the surrounding paths, so this is acceptable; a small test against a trivial `/usr/bin/env true`/`false` would still be cheap insurance for the exactly-once-resume contract.

> Note: I could not build locally — `swift build` fails on a missing `Frameworks/GhosttyKit.xcframework` binary artifact, which is an environmental/checkout issue unrelated to this diff. The change is self-contained to one file; the PR reports a clean `swift build --arch arm64` and 145 passing tests.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. Correct, concurrency-clean fix that resolves the beachball without altering kickoff ordering or behavior on error paths.

---

PR #407, opened 2026-06-02
Fix duplicate review sessions on in-flight crow:auto kickoff

### LINKED ISSUE #406: crow:auto creates duplicate review sessions during in-flight kickoff (regression from #405)
## Summary

After PR #405 (the beachball fix) shipped, the `crow:auto` auto-review watcher creates **duplicate review sessions** for the same PR when it ticks while a prior kickoff for that PR is still in flight.

## Why this regressed

Before #405, `createReviewSession()` held the main actor for the entire ~10s of git/clone work (via the synchronous `Process.waitUntilExit()` inside `shell()`). That inadvertently serialized everything — the IssueTracker's refresh tick couldn't interleave, so the watcher's guard could never observe a stale `reviewSessionID` mid-kickoff.

After #405, the git work runs in `Task.detached`, so the main actor stays responsive and multiple watcher ticks can fire during the window where the session row hasn't yet been appended to `appState.sessions`.

## Concrete race (auto-review path)

1. IssueTracker tick T0: watcher sees `request.reviewSessionID == nil` for PR #X, calls `enqueueReviewKickoff([url])`. Task A starts; `createReviewSession` enters `Task.detached` for the git clone.
2. IssueTracker tick T1 (~N seconds later, while clone is still in flight): tracker rebuilds `requests`. **`reviewSessionID` is still `nil`** because it's populated lazily inside the tracker's refresh closure (`IssueTracker.swift:385-389`) by matching `appState.reviewSessions` against PR URLs — the session row hasn't been appended yet.
3. The `autoReviewedFingerprints` set at `AppDelegate.swift:608` catches exact `id@headRefOid` repeats — but if `headRefOid` changes between ticks (force-push, intermediate commits), the fingerprint differs and slips through.
4. Watcher enqueues a SECOND kickoff for the same PR. The serial kickoff queue runs it. `createReviewSession` has **no dedup check** — it does another full clone/setup and appends a second session.

## UI symptom (same root cause)

`ReviewBoardView.swift:309-310` hides the Start Review button on `request.reviewSessionID != nil`. Because `reviewSessionID` is populated lazily by the next tracker refresh, the button stays clickable for several seconds after a session has been created. Even though the user isn't double-clicking, the UI state is wrong during the window.

## Suggested fix — three layered changes sharing one helper

1. **New `AppState.existingReviewSession(forPRURL:)` helper** — single source of truth, checks `appState.sessions` by the canonical name pattern `\"review-<repo>-<num>\"` (already used by `createReviewSession` at `SessionService.swift:~1322`).

2. **Tighten the auto-review watcher guard** (`AppDelegate.swift:618-624`) — augment the existing guard to also skip when `appState.existingReviewSession(forPRURL: request.url) != nil`. This stops relying on the lagging `request.reviewSessionID` and uses the authoritative `appState.sessions` instead.

3. **Universal backstop in `createReviewSession`** — add an early-return dedup at the top of the function that calls the helper. Catches any future caller (RPC, manual, watcher) that re-enters during a kickoff. Safe because `enqueueReviewKickoff` already serializes — by the time Task B runs, Task A's `appState.sessions.append` has completed.

4. **UI** — single-PR Start Review button (`ReviewBoardView.swift:309-310`) also uses the helper, so it hides instantly when the session appears in `appState`. Batch button (`ReviewBoardView.swift:138-158`) pre-filters URLs that already have sessions in `appState`.

Also extract the PR-URL parsing currently inlined at the top of `createReviewSession` (line ~1212) into a `Session.parseReviewPR(url:)` helper in CrowCore so all four sites share one parser.

## Acceptance

- `crow:auto`-labeled PRs trigger exactly one review session per `(repo, prNumber)` even when the watcher ticks multiple times during the clone window.
- A second tick during an in-flight kickoff is silently skipped at the watcher level.
- `createReviewSession` itself is idempotent per `(repo, prNumber)` — calling it again returns the existing session's ID without creating a duplicate.
- Start Review button (single + batch) hides/filters as soon as the session exists in `appState`, not after the next tracker refresh.
- Existing 145 tests still pass; add a focused unit test for `existingReviewSession(forPRURL:)`.

## Out of scope

- Refactoring IssueTracker to populate `reviewSessionID` synchronously when the session is appended (the new authoritative-`appState` checks make this unnecessary).
- Tracking explicit \"in-flight\" URL state on AppDelegate/SessionService (serial kickoff queue + `appState.sessions` check is sufficient).
- Disabling the batch button entirely when all selected PRs already have sessions (pre-filtering is enough).

dgershman:
## Summary

- Stop trusting the lagging `ReviewRequest.reviewSessionID` cross-ref and consult `appState.sessions` directly via a new `AppState.existingReviewSession(forPRURL:)` helper.
- Layer the check at three sites: auto-review watcher guard, `createReviewSession` early-return backstop, and both Start Review buttons in `ReviewBoardView`.
- Extract the PR-URL parser into `Session.parseReviewPR(url:)` so all sites share one parser.

## Why

After #405 moved `createReviewSession`'s git work into `Task.detached`, the main actor stays responsive during the ~10s clone window. The IssueTracker watcher can now tick during that window — and because `request.reviewSessionID` is populated lazily on the *next* refresh, the watcher's existing guard sees stale `nil`, enqueues a second kickoff, and `createReviewSession` (with no dedup of its own) creates a duplicate session.

The same lag was also why the per-row Start Review button stayed clickable for several seconds after a session had been created.

## Test plan

- [x] `arch -arm64 swift test --package-path Packages/CrowCore` — 205 tests pass (includes the new `existingReviewSession` and `parseReviewPR` test files)
- [x] `arch -arm64 swift test` (root + every package) — all suites pass except a pre-existing flaky `retryReadinessEmitsTimedOutWhenSentinelMissing` tmux timing test in CrowTerminal that's unrelated to this change
- [ ] Manual smoke (reviewer): with `autoReviewRepos` configured, open a `crow:auto` PR, confirm exactly one `review-<repo>-<num>` session appears even when the IssueTracker ticks multiple times during the clone; confirm the Start Review button flips to "Go to Session" on the same tick the session appears; force-push the PR to confirm the SHA-advanced re-kick path still works.

Closes #406

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -497,6 +497,19 @@ public final class AppState {
         return reviewRequests.first { $0.url == prLink.url }
     }
 
+    /// Authoritative lookup for "is there already an active review session
+    /// for this PR URL?". Cross-references `reviewSessions` (which already
+    /// excludes completed/archived) against `links` by `.pr` linkType +
+    /// exact URL match. Used by the kickoff watcher, the review-board
+    /// buttons, and `SessionService.createReviewSession` as a single source
+    /// of truth so they don't rely on the lagging `ReviewRequest.reviewSessionID`
+    /// cross-reference that IssueTracker populates one tick late (CROW-406).
+    public func existingReviewSession(forPRURL url: String) -> Session? {
+        reviewSessions.first { session in
+            links(for: session.id).contains { $0.linkType == .pr && $0.url == url }
+        }
+    }
+
     /// Labels for a session, sourced from its linked AssignedIssue or ReviewRequest.
     public func labels(forSession session: Session) -> [LabelInfo] {
         if let issue = assignedIssue(for: session) {
--- a/Packages/CrowCore/Sources/CrowCore/Models/Session.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/Session.swift
@@ -65,6 +65,19 @@ public struct Session: Identifiable, Codable, Sendable {
         self.autoMergeEnabledAt = autoMergeEnabledAt
     }
 
+    /// Parse a GitHub PR URL (`https://github.com/<owner>/<repo>/pull/<number>`)
+    /// into its components. Returns `nil` if the URL is malformed. Shared by
+    /// `SessionService.createReviewSession` and the kickoff-dedup helpers on
+    /// `AppState`; keep here so callers don't reinvent the same parser.
+    public static func parseReviewPR(url: String) -> (owner: String, repo: String, number: Int)? {
+        let components = url.split(separator: "/")
+        guard components.count >= 5,
+              let number = Int(components.last ?? "") else { return nil }
+        let owner = String(components[components.count - 4])
+        let repo = String(components[components.count - 3])
+        return (owner, repo, number)
+    }
+
     // Backward-compatible decoding: default `kind`, `agentKind`, and
     // `reviewPromptDispatched` when missing from older persisted data.
     // `reviewPromptDispatched` defaults to `true` so existing review sessions
--- a/Packages/CrowCore/Tests/CrowCoreTests/AppStateReviewSessionLookupTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AppStateReviewSessionLookupTests.swift
@@ -1,0 +1,100 @@
+import Foundation
+import Testing
+@testable import CrowCore
+
+@MainActor
+private func makeReviewSession(name: String = "review-crow-406", status: SessionStatus = .active) -> Session {
+    Session(
+        name: name,
+        status: status,
+        kind: .review,
+        agentKind: .claudeCode,
+        provider: .github
+    )
+}
+
+@MainActor
+@Test func existingReviewSessionReturnsNilWhenNoSessionMatches() {
+    let appState = AppState()
+    let session = makeReviewSession()
+    appState.sessions = [session]
+    appState.links[session.id] = [
+        SessionLink(sessionID: session.id, label: "PR #1", url: "https://github.com/org/other/pull/1", linkType: .pr)
+    ]
+
+    #expect(appState.existingReviewSession(forPRURL: "https://github.com/org/repo/pull/406") == nil)
+}
+
+@MainActor
+@Test func existingReviewSessionReturnsSessionWithMatchingPRLink() {
+    let appState = AppState()
+    let session = makeReviewSession()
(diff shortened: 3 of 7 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- `Session.parseReviewPR(url:)` is pure string parsing with no injection surface; it preserves the exact behavior of the inlined parser it replaces (`split(separator: "/")`, `count >= 5`, trailing-segment `Int` cast).
- No new untrusted input paths — PR URLs originate from the GitHub-backed `ReviewRequest`/`SessionLink` data that already flows through the app.
- The new `existingReviewSession(forPRURL:)` is correctly `@MainActor`-isolated (matches `AppState`), and every caller (watcher, review-board buttons, `SessionService`) is already on the main actor, so there's no data-race introduced.

**Concerns:**
- None.

### Code Quality
- **Dedup is genuinely authoritative, not racy.** The serial `reviewKickoffTail` queue (`AppDelegate.swift:173-187`) awaits the previous tail, so the second `createReviewSession` only runs after the first has fully returned — and the session + `.pr` link are appended synchronously on the main actor at `SessionService.swift:1337-1340` *after* the detached clone await. The backstop at `SessionService.swift:1255` therefore always sees the prior session. Sound.
- **Watcher guard change is the right shape.** `guard (request.reviewSessionID == nil && existingByPR == nil) || shaAdvanced` is strictly more restrictive only on the fresh-kickoff path and leaves the `shaAdvanced` re-kick path (force-push / round-2) untouched, including the stale-session teardown at `AppDelegate.swift:637`. The A-path completion case still kicks off correctly since `existingReviewSession` excludes completed/archived sessions (`reviewSessions` filter at `AppState.swift:380-382`).
- **Refactor removes real duplication** by sharing one parser across `SessionService` and the dedup helpers, and the `.pr`-linkType + exact-URL match is consistent across all three call sites (the link is stored from the same `prURL` the watcher/buttons pass in).
- **Test coverage is thorough** — the new suites cover nil/match/completed/archived/non-PR-linkType/duplicate cases for the lookup and owner/repo/number/malformed cases for the parser. All 10 new tests pass locally (`swift test --package-path Packages/CrowCore`).

Minor, non-blocking observations (no action needed): `existingReviewSession` is O(sessions × links) and recomputed each watcher iteration, but both sets are tiny; and exact-string URL matching is fine here because the same URL source feeds both the stored link and every lookup.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 0 Green] findings. Correct, well-isolated, well-tested fix for the in-flight kickoff race.

PR #409, opened 2026-06-03
Fix managed terminal launch race (bare-zsh orphans) (#408)

### LINKED ISSUE #408: Managed terminals intermittently launch as bare zsh — agent never starts (sleep+send race; use new-terminal --command)
## Summary

Managed Claude/Codex terminals created via `/crow-workspace` (and the batch variant) **intermittently come up as a bare `zsh` shell with no agent running** — the window exists but `claude` never launched. Re-running the same workspace setup sometimes works, sometimes doesn't, which is the signature of a **timing race**, not a config error.

On this machine right now, **26 of 63 tmux windows in `crow-cockpit` are orphaned**: `pane_current_command = zsh` with the default title `Mac.lan` (a window whose title was never set by a running agent). That's a ~40% failure rate, and it includes two sessions I just created (corveil#1059, corveil#1058) which came up shell-only twice in a row.

## Root cause

Crow already exposes a **race-free** way to launch an agent in a terminal: `new-terminal` accepts a `command` param, and the RPC handler resolves it and passes it through to `TmuxController.newWindow(command:)` → `tmux new-window … <command>`, so the agent **is** the window's spawned process (`AppDelegate.swift` `new-terminal` handler ~L1213–1262; `TmuxController.newWindow` `command:` arg, `Packages/CrowTerminal/Sources/CrowTerminal/TmuxController.swift:108`).

But `skills/crow-workspace/setup.sh` **does not use it.** It launches the agent with a fragile three-step handshake instead (`setup.sh:588–644`):

1. `crow new-terminal --managed` **with no `--command`** → spawns an empty interactive `zsh` window.
2. **Blind `sleep 3`** — a fixed delay, no shell-readiness check. The code comment even acknowledges it: *"the shell process needs time to start and display its prompt."*
3. `crow send` pastes `cd <worktree> && claude --permission-mode plan "$(cat <prompt>)"\n` into that shell (paste-buffer + `Enter`).

When the interactive `zsh` hasn't reached its prompt within the 3-second window, the pasted command + `Enter` are delivered to a shell that isn't ready and are **dropped** (or partially consumed), so `claude` never starts and the window is left sitting at a bare prompt. The user's `~/.zshrc` (sources a shell snapshot, sets `ANTHROPIC_*`, etc.) makes cold-prompt time variable, and the failure rate climbs under load — this host had 60+ tmux windows and ~15 running `claude` processes when both #1059/#1058 launches failed. That's exactly when a 3s budget is most likely to be blown.

This is a robustness defect in the **crow ↔ setup.sh launch contract**: a fixed sleep + blind send-keys is inherently racy, and crow provides a deterministic alternative that isn't being used.

## Evidence

- `tmux -S <crow-tmux.sock> list-windows -a -F '#{window_index}|#{pane_current_command}|#{pane_title}'`:
  - **26/63 windows** at `cmd=zsh`, `title=Mac.lan` (orphaned — agent never ran).
  - Healthy windows show `cmd=2.1.16x` (the running `claude`) and a ticket-slug title (e.g. `✳ ontology-entities-per-org-unique`); orphaned ones never get the title because nothing set it.
- The two reported failures (corveil#1059 `configurable-ontology-design`, corveil#1058 `sensors-create-form`) have **no window with their slug title** — consistent with both sitting at bare-`zsh` `Mac.lan` windows.
- `setup.sh` reported `Claude Code launched` for both (it only checks that `crow send` returned 0 — it never verifies the agent actually started), so the failure is silent.

## Proposed fix

**Primary — launch via `new-terminal --command` (eliminate the race):**
- Change `setup.sh` to pass the agent launch line as `crow new-terminal … --command "<cd && agent …>"` so the agent is the window's root process. No `sleep`, no `send`, no readiness assumption. (`new-terminal` already resolves the claude binary path in-command, per the RPC handler.)
- Keep the paste path only for delivering the *prompt text* if the command line gets unwieldy — but the prompt is already read via `"$(cat <prompt>)"` inside the command, so the whole launch can be the spawned command.

**Secondary — make crow detect/surface a failed launch (defense in depth):**
- For a `--managed` terminal with a launch command, after spawn verify the window's `pane_current_command` is no longer the bare login shell within a bounded window; if it never transitions, mark the terminal failed / retry once / surface an error instead of silently leaving a dead shell.
- Optionally add a readiness primitive (wait-for-prompt) so any `send`-based launch path isn't a blind sleep.

**Cleanup:**
- Reap/flag the orphaned `Mac.lan` bare-`zsh` windows (26 currently) — they accumulate in `crow-cockpit` and clutter the cockpit.

## Acceptance criteria

- [ ] `/crow-workspace` and `/crow-batch-workspace` launches start the agent deterministically (no bare-`zsh` orphans) even under load / cold `~/.zshrc`.
- [ ] The launch uses `new-terminal --command` (or an equivalent race-free spawn) rather than `sleep 3` + `crow send`.
- [ ] A launch that fails to start the agent is **detected and surfaced** (error/retry), not silently reported as "launched".
- [ ] Repro harness: spawn N managed terminals concurrently; assert each window's root command is the agent, 0 orphaned shells.
- [ ] Existing orphaned `Mac.lan`/`zsh` windows are cleaned up (or a cockpit-reaper handles them).

## Refs

- `skills/crow-workspace/setup.sh:588-644` (the `new-terminal --managed` → `sleep 3` → `crow send` sequence)
- `Sources/Crow/App/AppDelegate.swift` `new-terminal` RPC handler (~L1213-1262; reads `command`, resolves claude bin)
- `Packages/CrowTerminal/Sources/CrowTerminal/TmuxController.swift:108` `newWindow(command:)` — race-free spawn path that isn't being used
- Backend: single `crow-cockpit` tmux session, one window per terminal; paste path `TerminalRouter` / `TmuxController` paste-buffer

dhilgaertner:
Closes #408

## Problem

Managed Claude/Codex terminals created by `/crow-workspace` (and the batch variant) intermittently came up as a **bare `zsh` shell with no agent running** (~40% under load; 26/63 cockpit windows were orphaned on the reporting host). Two race modes, both because `setup.sh` launched the agent **out-of-band**:

1. **Blind-paste race.** `setup.sh` did `crow new-terminal --managed` → `sleep 3` → `crow send "<cd && claude …>"`. The paste landed before the shell's line editor (`zle`) was live, so the keystrokes were dropped → bare zsh. The naive `--command` fix was *also* racy: `TmuxBackend.registerTerminal` spawned the wrapper as the window root and then immediately `sendText(command)` with no readiness wait.
2. **No-window / swallowed failure.** Under load `TmuxController.newWindow`'s tmux subprocess exceeded the 2 s timeout, threw `TmuxError.timedOut`, and the `new-terminal` RPC only logged it — registering a window-less terminal while returning success.

## Fix

The app already had a **race-free launch** (`SessionService.launchAgent` fired on the `.shellReady` sentinel), but only for *restored* terminals. This routes brand-new managed terminals through the same path:

- **Defer the launch.** `new-terminal` holds the command in `AppState.pendingLaunchCommands` and registers the window with `command: nil`; the deferred paste happens in `wireTerminalReadiness` on `.shellReady` (when `zle` is live), consuming both the pending command and `autoLaunchTerminals` membership so `launchAgent` can't double-fire. The deferred paste replicates the `send` RPC's hook-config + OTEL env prep, so hook routing and telemetry are preserved.
- **Surface failures.** Bounded retry of `registerTerminal` with a longer per-call `new-window` budget; on exhaustion the terminal is marked `.failed` and the RPC returns `launch_failed` instead of reporting success.
- **`setup.sh`** launches via `new-terminal --command` (no `sleep`/`send`), keeps its own `--rc/--name` (the app can't inject for a `cd && claude` form), parses `launch_failed`, and polls a new `readiness` field on `crow list-terminals` to **warn non-fatally** if the agent never started.
- **Conservative reaper** at launch kills only cockpit windows that **no terminal references AND** that sit at a bare login shell — never a window running an agent or one a terminal still references.

> Note on the ticket's "window root command is the agent" criterion: the wrapper is *intentionally* the window root (it sources shell config + writes the readiness sentinel). This is reinterpreted as "each window reliably runs the agent, zero bare idle shells."

## Tests

- Pure helpers: `resolveLaunch`, `registerWithRetry`, `shouldReapWindow`.
- `SessionService` readiness-branch test (deferred paste consumes pending + autoLaunch; `.timedOut` does not).
- **Concurrent N-managed-terminal real-tmux regression** asserting every window runs the launched process with **zero bare shells**.
- Serialized the `TmuxBackend` integration suite so its timing-sensitive tests don't contend.

All package + app suites pass (`make test` equivalent run per package).

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -223,6 +223,14 @@ public final class AppState {
     /// not brand-new terminals created via the `new-terminal` RPC.
     public var autoLaunchTerminals: Set<UUID> = []
 
+    /// Pending agent launch command per terminal ID, for brand-new managed
+    /// terminals created via `new-terminal --command`. The command is NOT
+    /// pasted immediately (that races the shell's line editor — issue #408);
+    /// it is held here until the readiness sentinel fires `.shellReady`, at
+    /// which point `SessionService.wireTerminalReadiness` pastes it. In-memory
+    /// only — never persisted, so a relaunch can't re-paste a stale command.
+    public var pendingLaunchCommands: [UUID: String] = [:]
+
     // MARK: - Hook Events (per-session Observable wrappers)
 
     /// Per-session hook state. Using @Observable class wrappers so mutations to one
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -178,7 +178,8 @@ public final class TmuxBackend {
         name: String,
         cwd: String,
         command: String?,
-        trackReadiness: Bool
+        trackReadiness: Bool,
+        newWindowTimeout: TimeInterval = TmuxController.defaultTimeout
     ) throws -> TmuxBinding {
         precondition(!tmuxBinary.isEmpty, "TmuxBackend.configure(...) must be called first")
         let ctrl: TmuxController
@@ -219,7 +220,8 @@ public final class TmuxBackend {
             name: name,
             cwd: cwd.isEmpty ? nil : cwd,
             env: env,
-            command: wrapperPath
+            command: wrapperPath,
+            timeout: newWindowTimeout
         )
         bindings[id] = windowIndex
 
@@ -363,6 +365,56 @@ public final class TmuxBackend {
         if let logPath = wrapperLogs.removeValue(forKey: id) {
             try? FileManager.default.removeItem(atPath: logPath)
         }
+    }
+
+    /// Bare login shells we consider "orphaned" when a cockpit window is not
+    /// referenced by any terminal — i.e. a window left at a shell with no agent
+    /// running (#408). Anything else (claude/codex/node/an editor/…) is left
+    /// alone. tmux reports `pane_current_command` without the login-shell `-`
+    /// prefix, but we match both forms defensively.
+    nonisolated static let orphanLoginShells: Set<String> = [
+        "zsh", "-zsh", "bash", "-bash", "sh", "-sh",
+        "fish", "-fish", "dash", "-dash", "ksh", "tcsh", "csh", "login",
+    ]
+
+    /// Decide whether a single cockpit window should be reaped. Pure so the
+    /// policy is unit-testable: reap only when the window is NOT referenced by a
+    /// live terminal AND its pane is sitting at a bare login shell. Never reaps
+    /// a window running an agent (or any non-shell process), so an agent that
+    /// exited and left the user at a shell — but whose terminal still references
+    /// the window — is preserved (it's in `keep`).
+    nonisolated static func shouldReapWindow(index: Int, command: String, keep: Set<Int>) -> Bool {
+        if keep.contains(index) { return false }
+        return orphanLoginShells.contains(command)
+    }
+
+    /// Reap cockpit windows that no live terminal references AND that are
+    /// sitting at a bare login shell — leaked windows from a timed-out
+    /// `new-window` or a forgotten terminal (#408). `keepWindowIndices` is the
+    /// set of window indices referenced by persisted terminals; it is unioned
+    /// with the in-memory `bindings` so a window created/adopted this run is
+    /// never reaped. Best-effort; returns the count reaped.
+    @discardableResult
+    public func reapUnboundCockpitWindows(keepWindowIndices: Set<Int>) -> Int {
+        guard let ctrl = controller else { return 0 }
+        let keep = keepWindowIndices.union(bindings.values)
+        let windows: [(index: Int, command: String)]
+        do {
+            windows = try ctrl.listWindowCommands()
+        } catch {
+            reportIfTimeout(error)
+            return 0
+        }
+        var reaped = 0
+        for window in windows where Self.shouldReapWindow(index: window.index, command: window.command, keep: keep) {
+            ctrl.killWindow(index: window.index)
+            NSLog("[CrowTelemetry tmux:orphan_window_reaped index=\(window.index) command=\(window.command)]")
+            reaped += 1
+        }
+        if reaped > 0 {
+            NSLog("[Crow] Reaped \(reaped) orphaned bare-shell cockpit window(s) (#408)")
+        }
+        return reaped
     }
 
     /// Return the shared cockpit Ghostty surface, lazily creating it the
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxController.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxController.swift
@@ -104,13 +104,34 @@ public struct TmuxController: Sendable {
         return out.split(separator: "\n").compactMap { Int($0.trimmingCharacters(in: .whitespaces)) }
     }
 
+    /// List each window's index and the command currently running in its
+    /// active pane (`#{window_index}` + `#{pane_current_command}`). Used by the
+    /// orphan-window reaper to distinguish a window running an agent from a
+    /// bare login shell (#408).
+    public func listWindowCommands() throws -> [(index: Int, command: String)] {
+        let out = try run(["list-windows", "-t", sessionName,
+                           "-F", "#{window_index}\t#{pane_current_command}"])
+        return out.split(separator: "\n").compactMap { line in
+            let parts = line.split(separator: "\t", maxSplits: 1, omittingEmptySubsequences: false)
+            guard parts.count == 2, let idx = Int(parts[0].trimmingCharacters(in: .whitespaces)) else {
+                return nil
+            }
+            return (idx, parts[1].trimmingCharacters(in: .whitespaces))
+        }
+    }
+
     // MARK: - Windows
 
+    /// `timeout` defaults to the per-call default. Callers spawning a window
+    /// while the app is under load (many concurrent hydrations, a contended
+    /// main actor) can pass a longer budget so a slow `new-window` doesn't
+    /// SIGTERM and leave the terminal window-less (issue #408).
     public func newWindow(
         name: String? = nil,
         cwd: String? = nil,
         env: [String: String] = [:],
-        command: String? = nil
+        command: String? = nil,
+        timeout: TimeInterval = TmuxController.defaultTimeout
     ) throws -> Int {
         var args = ["new-window", "-P", "-F", "#{window_index}", "-t", sessionName]
         if let name { args.append(contentsOf: ["-n", name]) }
@@ -122,7 +143,7 @@ public struct TmuxController: Sendable {
         if let cwd, !cwd.isEmpty { args.append(contentsOf: ["-c", cwd]) }
         for (k, v) in env { args.append(contentsOf: ["-e", "\(k)=\(v)"]) }
         if let command { args.append(command) }
-        let out = try run(args)
+        let out = try run(args, timeout: timeout)
         guard let idx = Int(out.trimmingCharacters(in: .whitespacesAndNewlines)) else {
             throw TmuxError.cliFailed(
                 args: args,
--- a/Packages/CrowTerminal/Tests/CrowTerminalTests/OrphanWindowReaperTests.swift
+++ b/Packages/CrowTerminal/Tests/CrowTerminalTests/OrphanWindowReaperTests.swift
@@ -1,0 +1,31 @@
+import Foundation
+import Testing
+@testable import CrowTerminal
+
+/// Policy tests for the conservative orphan-window reaper (#408): reap a
+/// cockpit window only when NO live terminal references it AND it is sitting at
+/// a bare login shell. Pure over `shouldReapWindow`, so no tmux needed.
+@Suite("Orphan cockpit window reaper policy (#408)")
+struct OrphanWindowReaperTests {
+
+    @Test func reapsUnboundBareShell() {
+        #expect(TmuxBackend.shouldReapWindow(index: 3, command: "zsh", keep: []))
+        #expect(TmuxBackend.shouldReapWindow(index: 3, command: "-zsh", keep: []))
+        #expect(TmuxBackend.shouldReapWindow(index: 3, command: "bash", keep: []))
+        #expect(TmuxBackend.shouldReapWindow(index: 3, command: "sh", keep: []))
+    }
+
+    @Test func keepsBoundWindowEvenWhenBareShell() {
+        // A terminal references it (e.g. the agent exited and the user is now at
+        // the shell) — must be preserved.
+        #expect(!TmuxBackend.shouldReapWindow(index: 3, command: "zsh", keep: [3]))
+    }
+
+    @Test func keepsWindowRunningAProcess() {
+        // Anything that isn't a bare login shell is left alone.
+        #expect(!TmuxBackend.shouldReapWindow(index: 4, command: "claude", keep: []))
+        #expect(!TmuxBackend.shouldReapWindow(index: 4, command: "node", keep: []))
+        #expect(!TmuxBackend.shouldReapWindow(index: 4, command: "codex", keep: []))
+        #expect(!TmuxBackend.shouldReapWindow(index: 4, command: "tail", keep: []))  // session anchor
+    }
+}
--- a/Packages/CrowTerminal/Tests/CrowTerminalTests/TmuxBackendTests.swift
+++ b/Packages/CrowTerminal/Tests/CrowTerminalTests/TmuxBackendTests.swift
@@ -8,7 +8,11 @@ import Testing
 /// covered separately by the visual demo path. Skipped automatically
 /// when no tmux binary is present on the host.
 @MainActor
-@Suite("TmuxBackend integration", .enabled(if: discoveredTmuxBinary != nil))
+// Serialized: these are real-tmux integration tests that spawn login shells and
+// assert on readiness *timing* (e.g. a 0.15s watch budget). Run in parallel they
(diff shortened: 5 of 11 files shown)


dgershman (changes requested):
## Code & Security Review

The deferred-launch design is the right shape: routing brand-new managed terminals through the existing `.shellReady` sentinel that already serves restored terminals is a clean reuse of a proven path, and `pendingLaunchCommands` being in-memory-only correctly avoids re-pasting a stale command on relaunch. The added `readiness` field on `list-terminals` plus the `launch_failed` error code give `setup.sh` a real success/failure signal instead of the prior fire-and-forget. Test coverage is solid: pure unit tests over `resolveLaunch`/`registerWithRetry`/`shouldReapWindow`, an instance test over the `.shellReady` branch, and a concurrent N-managed-terminal real-tmux regression. Below are the issues that should land in the same round trip.

### Code Quality

**🟡 Yellow — hydrate-fresh path still races on the same launch command (`Sources/Crow/App/SessionService.swift:309`).** Brand-new managed terminals now persist `terminal.command = launch_cmd` (the `cd … && claude …` form), where pre-PR they persisted `nil` because `setup.sh` deferred the launch out-of-band. The hydrate fallback at `rebuildHydratedSessionWithTmux` (when `tmuxBinding` is `nil` or `adoptTerminal` throws — tmux socket changed, window was killed, post-reboot) calls `TmuxBackend.shared.registerTerminal(..., command: terminal.command, ...)`, and `registerTerminal` synchronously blasts that command via `sendText(id, text: command + \"\\n\")` before any sentinel fires (`Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:234`). That's exactly the dropped-keystrokes race this PR set out to remove — just shifted to the recovery path. Two reasonable shapes: (a) on the hydrate-fresh path, pass `command: nil` to `registerTerminal` and rely on the existing `autoLaunchTerminals` + `launchAgent` rebuild that hydrate already wires up for restored terminals, or (b) re-stage the same `pendingLaunchCommands` + deferred-paste machinery on this fallback. Either fix keeps the regression from sneaking back in via the recovery path.

**🟡 Yellow — `registerWithRetry` blocks the main actor for up to 12 s (`Sources/Crow/App/AppDelegate.swift:1349`, `Packages/CrowTerminal/Sources/CrowTerminal/TmuxController.swift:26`).** `await MainActor.run { … }` wraps the whole RPC body, and `registerWithRetry(attempts: 3) { … newWindowTimeout: 4.0 }` is invoked synchronously inside it. `TmuxController.run` blocks the calling thread until the subprocess returns or `kill()` fires at `timeout`, so a worst-case three timed-out attempts pins the main actor for ~12 s — long enough to beachball the UI and stall every other `MainActor.run`-dispatched RPC (e.g. concurrent `setup.sh` invocations from `/crow-batch-workspace`). The previous behavior was 2 s and broken; the new behavior is reliable but visibly stalls under the same load this PR is trying to harden. Hoist the registration off the main actor (`Task.detached` returning the `TmuxBinding`, then re-enter `MainActor` for the state mutations), or at minimum bound the total budget (e.g. `attempts: 2` with backoff and a shorter per-call timeout) so the worst case stays inside the previous user-perceived window.

**🟡 Yellow — `prepareManagedLaunchText` duplicates the `send` RPC handler (`Sources/Crow/App/SessionService.swift:415` vs `Sources/Crow/App/AppDelegate.swift:1478`).** The OTEL env-var list, the agent-token guard, and the hook-config write are copy-pasted; the PR even adds a `// NOTE: keep the two in sync` comment, which is a maintenance trap rather than a design. Extract a single helper (e.g. on the agent or a shared static on `AppDelegate`) that returns `(textWithExports, didLaunch)` and call it from both sites. Bonus: the `send` RPC sets `terminalReadiness = .agentLaunched` only when the launch token is present, whereas `pasteDeferredLaunch` sets it unconditionally — a shared helper makes that asymmetry visible and easy to align.

**🟡 Yellow — `DeferredLaunchTests` mutates singleton `TmuxBackend.shared.onReadinessChanged` without `.serialized` (`Tests/CrowTests/DeferredLaunchTests.swift:1`).** Each test calls `service.wireTerminalReadiness()`, which overwrites the shared closure, and then fires `TmuxBackend.shared.onReadinessChanged?(terminalID, …)`. If Swift Testing runs the two `@MainActor` tests in parallel (the default), one test can stomp the other's closure before its sentinel fires — same flake mode you just hit on the `TmuxBackend integration` suite and serialized (`Packages/CrowTerminal/Tests/CrowTerminalTests/TmuxBackendTests.swift:14`). Add `.serialized` to the `@Suite` so this stays green under concurrent test execution.

### Security Review

**Strengths**
- `pendingLaunchCommands` is documented and implemented as in-memory-only (`Packages/CrowCore/Sources/CrowCore/AppState.swift:226`), so a launch command can't be replayed across app restarts.
- `close-terminal` clears `pendingLaunchCommands` alongside `terminalReadiness` and `autoLaunchTerminals` (`Sources/Crow/App/AppDelegate.swift:1428`) — no orphan command can survive a terminal removal.
- The reaper is conservative: `shouldReapWindow` only fires when the window is **both** unreferenced **and** sitting at a bare login shell, and `reapUnboundCockpitWindows` defensively unions `bindings.values` so a window adopted/registered this run is structurally unreapable (`Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:386`). The 8 s delay in `AppDelegate` is belt-and-suspenders given hydrate runs synchronously on the main actor.

**Concerns**
- None new — no auth, secret-handling, or injection surface is touched. `terminal_readiness` in `setup.sh` parses `crow`-owned JSON it just produced, so the regex-on-JSON shape is fragile but not a security boundary.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 4 Yellow, several Green] findings. The deferred-launch shape is correct; the four Yellows are the kind that compound (a regression on a recovery path + a main-actor stall + duplicated launch prep + a flaky test) and worth catching in this round rather than as #408 follow-ups.

---

dhilgaertner:
**Live finding while setting up this PR's own workspace — a window-index-reuse bug the reaper exposes.** (Observed from `~/.local/share/crow/crash-logs/crow-20260603T034449.log`.)

Sequence on the fix build (app started 22:44):
1. ✅ Reaper worked perfectly — `Reaped 26 orphaned bare-shell cockpit window(s) (#408)` (cockpit windows 63 → 37, 0 orphans). Honest readiness surfacing also worked (`setup.sh` warned instead of false-success). Both are wins.
2. ❌ But the newly-created managed terminal `387D645A` never launched its agent — `first_prompt_timeout budget_ms=30000`, `sentinel_exists=false`, `state=timedOut`. The readiness diagnostics show why:
   ```
   --- tmux state ---  target=crow-cockpit:4   display_message=68270 2.1.159
   --- process tree --- 70384 … /Users/.../claude --permission-mode auto   ELAPSED 01-07:06:44
   --- pane capture --- <capture-pane failed>
   --- sentinel --- exists=false
   ```
   Terminal `387D645A` was associated with **`crow-cockpit:4`, a *surviving* window already running a different, long-lived `claude --permission-mode auto` (the Manager, PID 70384, up ~1d7h)** — not a fresh window. So no wrapper ran for `387D645A`, the readiness sentinel was never written, and it timed out.

**Root cause hypothesis:** after the reaper kills windows, **tmux window-index reuse** lets a new terminal's stored window index collide with a surviving window (here the Manager's). The deferred-launch-on-`.shellReady` path then waits forever for a sentinel that a wrapper in *this* terminal would have written — but the mapped window is someone else's agent.

**Why it didn't corrupt the Manager:** the launch is sentinel-gated, so the deferred paste never fired into window :4. Good — but if any code path *did* paste on a weaker signal, it would inject `cd … && claude …` into the Manager's live session. Worth an explicit guard: **never bind a managed terminal to a window already running an agent**, and key window identity on something stable (a per-terminal `@crow_terminal_id` tmux option) rather than the reusable window index — especially now that the reaper actively frees indexes for reuse.

**Repro:** with orphans present, open a session so the reaper runs, then create a new managed terminal — it can land on a recycled index pointing at a live window.

Also: the app stopped running shortly after (socket now refused). Couldn't tell from this log alone whether it crashed or was quit; flagging in case it's related.

dhilgaertner:
Thanks for the thorough review — all four Yellows addressed in f1eee8d.

**🟡 Hydrate-fresh recovery race (`SessionService.swift:309`).** Fixed at the source: brand-new managed terminals no longer persist the launch command on the `SessionTerminal` row — the row now stores `registerCommand` (nil when deferring), and the launch lives only in in-memory `pendingLaunchCommands`. So the hydrate-fresh `registerTerminal(command: terminal.command, …)` fallback gets `nil` and can't blind-paste, for **any** agent (no longer relies on the `command.contains("claude")` nil-ing heuristic). Restored managed terminals still relaunch via the `autoLaunchTerminals` + `launchAgent` path.

**🟡 `registerWithRetry` main-actor stall.** Tightened to **2 attempts × 3s** (was 3 × 4s), capping worst-case main-actor stall at ~6s instead of ~12s; the common case is a single fast attempt. A full off-main-actor hoist of `registerTerminal` is a larger refactor of a load-bearing path, so I took the "bound the budget" option you offered and noted the tradeoff in a comment.

**🟡 `prepareManagedLaunchText` duplication.** Extracted `AppDelegate.prepareAgentLaunchText(command:agent:sessionID:worktreePath:crowPath:telemetryPort:) -> (text, didLaunch)` as the single source of truth (hook config + OTEL env + agent-token guard). Both the `send` RPC and `pasteDeferredLaunch` call it; the copy-paste and the "keep in sync" comment are gone, and the `.agentLaunched` gating is now driven by the same `didLaunch` signal at both sites.

**🟡 `DeferredLaunchTests` not serialized.** Added `.serialized` to the suite, matching the `TmuxBackend integration` suite.

All package + app suites pass locally. Re-requesting review.

dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Path-traversal guard on `cwd` (`isPathWithinDevRoot`) preserved on the new-terminal path.
- No new trust boundary: the launch command is still authored by an authenticated local socket client; OTEL env vars carry no secrets.
- In-memory only `pendingLaunchCommands` — never persisted, so a relaunch can't re-paste a stale command (explicitly noted in AppState.swift:226-230 and defensively cleared on adopt at SessionService.swift:258-260).
- Persisted row deliberately stores `registerCommand` (nil for deferred launches) rather than the raw launch command, preventing the hydrate-fresh fallback in `rehydrateTerminalSurface` from blind-pasting it into a not-yet-ready shell (AppDelegate.swift:1323-1335).

**Concerns:** None.

### Code Quality

**Strengths:**
- Correctly routes brand-new managed terminals through the same race-free `.shellReady` path the restored path already used — eliminates the duplicate launch logic.
- `prepareAgentLaunchText` is a clean extraction: both the `send` RPC and the deferred-launch paste now share hook-config + OTEL prep, so they can't drift.
- State seeding (`terminalReadiness`, `pendingLaunchCommands`, `autoLaunchTerminals`) happens **before** `registerTerminal`, with a clear comment about why (AppDelegate.swift:1339-1351). Both producer (`new-terminal` RPC under `MainActor.run`) and consumer (`onReadinessChanged` closure in `SessionService`, both `@MainActor`-isolated) are serialized on the main actor, so the sentinel fire is guaranteed to see populated state and the `terminals` map.
- `pasteDeferredLaunch` consumes both `pendingLaunchCommands` and `autoLaunchTerminals` atomically (SessionService.swift:386-387) — `launchAgent`'s own membership check is now belt-and-suspenders against a double launch.
- `.timedOut` does NOT consume the pending command (SessionService.swift:352-359 + DeferredLaunchTests.swift:114-126), preserving Retry/foreground re-arm recovery.
- The orphan reaper is properly conservative: pure `shouldReapWindow` policy is unit-tested, real list unions `keepWindowIndices` with in-memory `bindings.values`, and the bare-shell whitelist excludes everything that could plausibly be an agent or session anchor.
- The concurrent N-managed-terminal real-tmux test directly exercises the regression and asserts on zero bare shells — the right shape of test for this bug.
- `.serialized` on `TmuxBackendTests` is well-justified in the comment (real-tmux integration tests contend for CPU during shell startup).
- The CLI poll surface (`launch_failed` on `new-terminal` + `readiness` on `list-terminals`) gives setup.sh honest signal without breaking the existing JSON shape.

### Consider (Green)
- `AppDelegate.swift:439-444` — the 8s post-launch sleep before the reaper runs is empirical. If app launch + initial main-thread work is very slow, `appState.terminals[*].tmuxBinding?.windowIndex` may not yet reflect a freshly re-registered window. The `bindings` union in `reapUnboundCockpitWindows` covers this in practice, so the risk is low; a follow-up could replace the timer with an explicit "rehydration settled" signal if you want to remove the magic constant.
- `setup.sh terminal_readiness` parses the flat list-terminals JSON with `grep -o '{[^{}]*}'`. Fine for the current schema (no nested objects per terminal), but worth a note if future fields add nesting.
- `registerWithRetry` retries immediately with no backoff. Intentional per the inline comment (2×3s budget), so this is preserved by design — but if a future change widens the budget it'd be worth revisiting.

### Tests Run
PR description states all package + app suites pass. Spot-checked test structure — pure helpers (`resolveLaunch`, `registerWithRetry`, `shouldReapWindow`) are well-isolated and the readiness-branch test seeds state exactly the way the RPC does.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 3 Green] findings. The fix correctly identifies the two race modes (blind-paste vs window-less terminal), routes brand-new terminals through the existing race-free `.shellReady` path, surfaces failures honestly via `launch_failed` + `readiness`, and adds a conservative reaper for already-leaked windows. The Green items are stylistic/forward-looking, not blocking.

---

PR #422, opened 2026-06-04
Add per-action agent selection (coding/review/job)

### LINKED ISSUE #421: Per-action agent selection: separate agents for coding vs reviews (additional dropdown)
Today the agent kind is a single global setting (`defaultAgentKind`). Different action types — implementing a ticket vs reviewing a PR — benefit from different agents (e.g. a stronger model for coding, a faster/cheaper one for reviews, or vice versa). Asking the user to flip the global before every action is friction.

Ask: let the user configure the agent **per action type**, surfaced as another dropdown in settings (alongside the current agent selector).

Action types to support at minimum:
- **Coding** — implementation sessions started by `crow:auto` pickup, `/crow-workspace`, manual session creation.
- **Reviews** — PR review sessions (auto-review queue, `/crow-review-pr`, `/rr` flows).

Behavior:
- Add a second dropdown in the agent settings UI: "Agent for reviews" alongside the existing "Agent for coding" (rename the existing one for clarity).
- Each dropdown lists the same set of available agents; both default to the current `defaultAgentKind` value so existing users see no change on upgrade.
- When Crow starts a session, it picks the agent based on the session's action type (coding vs review) rather than reading a single global.
- Persist as two distinct keys in `config.json` (e.g. `agents.coding` / `agents.review`), with a migration that fills both from the existing `defaultAgentKind` on first read.

Open:
- Are there more action types worth splitting (e.g. summarization jobs, the daily `Summarize *` jobs)? If yes, generalize the config to a map of action-type → agent rather than two fixed keys.
- Should the per-workspace `customInstructions` pattern extend here too (per-workspace agent override)? Out of scope for v1 unless trivial.

dgershman:
## Summary
- Adds `AppConfig.agentsByKind` — a map keyed by `SessionKind.rawValue` so users can pick a different agent per action type (coding, reviews, scheduled jobs). Manager stays pinned to Claude Code by spec.
- Sessions resolve their agent through a new `agentKind(for:)` helper on `AppConfig`/`AppState`: prefer the override, fall back to `defaultAgentKind`.
- Settings → General → Defaults gains three pickers ("Agent for coding/reviews/scheduled jobs"), each defaulting to **Use default** so existing users see no behavior change on upgrade.
- The ticket's open question (generalize to a map vs two fixed keys) is resolved in favor of the map — covers jobs too, which previously ignored `defaultAgentKind` entirely.
- No data migration code: the existing `decodeIfPresent ?? [:]` pattern means legacy configs load cleanly and every kind still resolves to `defaultAgentKind`.

Closes #421

## Test plan
- [x] `swift build --arch arm64` (release build green)
- [x] `swift test --package-path Packages/CrowCore` — 217 tests, including 5 new ones (resolver fallback, resolver override, manager pinning, round-trip, JSON-object serialization)
- [x] `swift test` — 170 tests in the top-level package
- [ ] Manual: open Settings → General → Defaults; confirm three new pickers appear with "Use default" first option and are disabled when only one agent is registered
- [ ] Manual: hand-edit `~/Dev/.claude/config.json` to add `{"agentsByKind": {"review": "codex"}}`, restart, trigger a review via `/rr` or `crow:auto` label, confirm review session uses codex while a fresh coding session still uses the default
- [ ] Manual: remove the `review` key from `agentsByKind`, restart, trigger another review — confirm fallback to `defaultAgentKind`

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -59,6 +59,20 @@ public final class AppState {
     /// current default without a config round-trip.
     public var defaultAgentKind: AgentKind = .claudeCode
 
+    /// Per-action agent overrides keyed by `SessionKind.rawValue`. Mirrors
+    /// `AppConfig.agentsByKind`. Empty means every kind falls back to
+    /// `defaultAgentKind` (CROW-421).
+    public var agentsByKind: [String: AgentKind] = [:]
+
+    /// Resolve the agent that should drive a newly-created session of the
+    /// given kind. Manager sessions are pinned to Claude Code; all other
+    /// kinds prefer an `agentsByKind` override and fall back to
+    /// `defaultAgentKind` when no override is set (CROW-421).
+    public func agentKind(for sessionKind: SessionKind) -> AgentKind {
+        if sessionKind == .manager { return .claudeCode }
+        return agentsByKind[sessionKind.rawValue] ?? defaultAgentKind
+    }
+
     /// Terminal IDs whose Claude Code was launched with `--rc` — drives the
     /// per-session indicator badge. Survives toggle changes so existing sessions
     /// keep showing the badge until they're restarted.
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -48,6 +48,16 @@ public struct AppConfig: Codable, Sendable, Equatable {
     /// The agent used for newly created sessions when none is specified.
     /// Existing persisted configs without this key decode to `.claudeCode`.
     public var defaultAgentKind: AgentKind
+    /// Per-action-type overrides. When a key is present, sessions of that
+    /// kind are created with the mapped agent; when absent, they fall back
+    /// to `defaultAgentKind`. Manager sessions are pinned to `.claudeCode`
+    /// and intentionally not honored by this map (CROW-421).
+    ///
+    /// Keyed by `SessionKind.rawValue` (string) rather than `SessionKind`
+    /// directly so JSON serializes as an object literal like
+    /// `{"review": "codex"}` — Swift's default `JSONEncoder` only treats
+    /// dictionaries with `String`/`Int` keys as JSON objects.
+    public var agentsByKind: [String: AgentKind]
 
     public init(
         workspaces: [WorkspaceInfo] = [],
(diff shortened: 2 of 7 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface: the change only adds a config-driven lookup (`agentsByKind[sessionKind.rawValue] ?? defaultAgentKind`). No I/O, no shell, no network, no deserialization of untrusted input beyond the existing config file.
- Manager sessions are hard-pinned to Claude Code in **both** resolvers (`AppConfig.agentKind(for:)` and `AppState.agentKind(for:)`), and `appConfigAgentResolverPinsManagerToClaudeCode` proves a hand-edited `manager` override in the JSON is ignored. Good defense against config tampering escalating the orchestration agent.
- Backward compatible: `decodeIfPresent(...) ?? [:]` means legacy configs load cleanly; `appConfigLegacyJSONWithoutAgentsByKindIsEmpty` locks this in. No migration code needed.

**Concerns:**
- None.

### Code Quality
- Storing the map as `[String: AgentKind]` keyed by `SessionKind.rawValue` (rather than `[SessionKind: AgentKind]`) is the correct call — Swift's `JSONEncoder` only emits String/Int-keyed dictionaries as JSON objects — and `appConfigAgentsByKindSerializesAsJSONObject` documents and guards exactly that.
- `Codable`/`Equatable` synthesis is intact: `agentsByKind` is added to `CodingKeys`, decoded in the custom `init(from:)`, and the synthesized `encode(to:)` picks it up — the round-trip test confirms both directions.
- Every site that assigns `appState.defaultAgentKind` from config also assigns `appState.agentsByKind` (AppDelegate:460–461 and 1095–1096), so the mirrored state can't drift between load paths.
- All three session-creation call sites are migrated coherently to `agentKind(for:)` with the matching kind (`.work` for coding/CLI `new-session`, `.review`, `.job`); the manual `new-session` RPC preserves the explicit `requestedAgentKind ??` override.
- The Settings pickers reuse the established Default-Agent pattern (`AgentRegistry.shared.allAgents()`, disabled when < 2 agents), with "Use default" as a `nil` tag that removes the override key. Consistent and self-documenting.

**Consider (non-blocking):**
- `agentKind(for:)` is duplicated verbatim in `AppConfig` and `AppState`. This intentionally mirrors the existing `defaultAgentKind` duplication (AppState caches config fields for hot-path reads), so it's consistent with the codebase — but a future refactor could let `AppState` delegate to a single shared resolver to avoid the two copies diverging.

### Verification
- `swift test --package-path Packages/CrowCore` → 9 agent-kind tests pass (5 new: resolver override, manager pinning, round-trip, JSON-object serialization, legacy-config fallback).
- Full `swift build` could not run in the review environment — it fails on a missing vendored `Frameworks/GhosttyKit.xcframework` binary artifact, which is a local-checkout/environment issue unrelated to this diff (the PR's CI reports the arm64 build green). CrowCore, which holds all the logic and tests, builds and passes here.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 1 Green] findings. Tight, well-tested, backward-compatible change with the security-sensitive manager pin enforced in both resolvers and covered by tests.

---

PR #420, opened 2026-06-04
Add Session.codeProvider for cross-backend pairing

dgershman:
Closes #414

## Summary

Follow-up to #410 / ADR 0005. Adds `Session.codeProvider: Provider?` so a Corveil-tasked session can pair with a GitHub or GitLab `CodeBackend` — the protocol split from #411 already supported this in principle; the model now represents it.

`nil` means "follow `provider`"; callers resolve with `session.codeProvider ?? session.provider`. No persisted-state migration: legacy sessions decode `codeProvider = nil` and the lookup-site fallback routes them to the same backend they used pre-split.

## Changes

- **`Session` model** — new optional field, init parameter, and `decodeIfPresent` in the custom decoder (same pattern as `lastReviewedHeadSha` / `autoMergeEnabledAt`).
- **`SessionService.findPRLink`** — takes an explicit `provider:` parameter; sole caller in `recoverOrphan` passes `session.codeProvider ?? session.provider ?? .github`. The in-line `gh pr list` shell-out fallback (only reached when no `ProviderManager` is wired) stays GitHub-only — unchanged.
- **Tests**
  - `SessionModelTests` — extends round-trip / nil-defaults / default-values tests; adds `sessionBackwardCompatDecodingWithoutCodeProvider` for legacy JSON.
  - `SessionRepositoryTests.savePreservesAllOptionalFields` — round-trips `codeProvider: .gitlab` alongside `provider: .github` to prove the field is independently persisted.

`IssueTracker` is intentionally unchanged: all three of its `session.provider` reads are task-backend / issue-lifecycle decisions, not `codeBackend` lookups.

## Test plan

- [x] `swift test --package-path Packages/CrowCore --filter SessionModelTests` — 16 tests pass (incl. new backward-compat decode).
- [x] `swift test --package-path Packages/CrowPersistence --filter SessionRepositoryTests` — 11 tests pass.
- [x] `swift test --package-path Packages/CrowProvider --filter BackendsTests` — sanity check on factory; 15 tests pass.
- [x] `swift build --arch arm64` — clean build, no callers broke.
- [ ] Manual smoke: open existing `state.json` from a pre-CROW-414 Crow build — confirm sessions hydrate with `codeProvider == nil` and PR discovery still finds the right backend.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #423, opened 2026-06-04
Move AutoRespond CLI name selection to CodeBackend (#412)

### LINKED ISSUE #412: Migrate AutoRespondCoordinator CLI string selection to backends
Follow-up to #410 / ADR 0005.

Today `AutoRespondCoordinator` selects the CLI binary name with inline ternaries:
```swift
provider == .gitlab ? \"glab\" : \"gh\"
```
in six places: `Sources/Crow/App/AutoRespondCoordinator.swift:127,133,153,170,197,206`.

With the new `TaskBackend` / `CodeBackend` abstraction in place, this can move
behind a backend method (e.g. `backend.cliName` or a richer prompt-vending API).
The prompt *body* still legitimately branches on provider for prose differences;
only the CLI token selection should move.

## Acceptance
- All six call sites at the lines above use the backend abstraction instead of
  `provider == .gitlab ? "glab" : "gh"`.
- No behavior change in the resulting prompts.

References:
- ADR 0005: docs/adr/0005-task-and-code-backend-protocols.md
- PR #411

dgershman:
## Summary

- `AutoRespondCoordinator` selected the CLI binary name (`gh` vs `glab`) with six inline `provider == .gitlab ? "glab" : "gh"` / `if provider == .gitlab` checks. With the `TaskBackend` / `CodeBackend` split from #410 / ADR 0005 in place, this belongs on the backend.
- Add `cliName: String` to `CodeBackend`; implement `"gh"` on `GitHubCodeBackend` and `"glab"` on `GitLabCodeBackend`.
- Inject `ProviderManager` into `AutoRespondCoordinator`; resolve a `CodeBackend` per session via `session.codeProvider ?? session.provider` (the convention from `Session.swift` / #420). Corveil-or-unknown sessions fall back to GitHub tooling — matches the prior `?? .github` default.
- Change `AutoRespondPrompts.build` and `QuickActionPrompts.build` to take a `CodeBackend` instead of a `Provider`. Prose branches still read `codeBackend.provider`; CLI tokens read `codeBackend.cliName`.

Prompt text is byte-identical for every existing branch (`cliName` returns the same string previously hardcoded), so no behavior change.

## Test plan

- [x] `swift build --arch arm64` — clean
- [x] `swift build --arch arm64 --build-tests` — clean (`QuickActionPromptsTests`, `AutoRespondCoordinatorReviewSessionTests` updated to construct `CodeBackend` / `ProviderManager`)
- [x] `rg 'provider == \.gitlab \? "glab" : "gh"' Sources/ Packages/ Tests/` returns zero matches
- [ ] Test bundle execution blocked locally by `swiftpm-xctest-helper` being x86_64-only on this Xcode install (build is arm64); CI runs the suite.

Closes #412

### DIFF
--- a/Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubCodeBackend.swift
+++ b/Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubCodeBackend.swift
@@ -10,6 +10,7 @@ import CrowCore
 /// See ADR 0005 for the protocol contract.
 public struct GitHubCodeBackend: CodeBackend {
     public let provider: Provider = .github
+    public let cliName: String = "gh"
     public let capabilities: Set<CodeCapability> = [.autoMergeLabel, .batchedPRStates]
 
     private let shellRunner: ShellRunner
--- a/Packages/CrowProvider/Sources/CrowProvider/Backends/GitLabCodeBackend.swift
+++ b/Packages/CrowProvider/Sources/CrowProvider/Backends/GitLabCodeBackend.swift
@@ -10,6 +10,7 @@ import CrowCore
 /// See ADR 0005.
 public struct GitLabCodeBackend: CodeBackend {
     public let provider: Provider = .gitlab
+    public let cliName: String = "glab"
     public let capabilities: Set<CodeCapability> = []
 
     private let shellRunner: ShellRunner
--- a/Packages/CrowProvider/Sources/CrowProvider/CodeBackend.swift
+++ b/Packages/CrowProvider/Sources/CrowProvider/CodeBackend.swift
@@ -14,6 +14,10 @@ public protocol CodeBackend: Sendable {
     /// Which provider this backend serves.
     var provider: Provider { get }
 
+    /// CLI binary name this backend shells to (e.g. "gh", "glab"). Used by
+    /// prompt-builders that render copy-paste hints — see `AutoRespondPrompts`.
+    var cliName: String { get }
+
     /// What this backend can do, beyond the protocol's required methods.
     /// Replaces `guard session.provider == .github` style guards in call sites.
     var capabilities: Set<CodeCapability> { get }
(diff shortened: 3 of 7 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Pure internal refactor — no new external input, network, auth, or filesystem surface. CLI tokens (`gh`/`glab`) are compile-time constants on each backend, not user-controlled, so no command-injection vector is introduced.
- PR URLs interpolated into prompt hints were already interpolated identically before this change; no new injection path.

**Concerns:**
- None.

### Code Quality
- **Behavior parity confirmed.** `cliName` returns the exact strings (`gh`/`glab`) previously hardcoded, and the prose branches still key off `codeBackend.provider`, so prompt text is byte-identical for every existing branch. Verified the GitLab `fixChecks`/`mergePR`/`fixConflicts` and GitHub `changesRequested` branches all reduce to the prior literals.
- **Force-unwrap is safe.** `providerManager.codeBackend(for: .github)!` in `resolveCodeBackend` can never trap — `ProviderManager.codeBackend(for:)` returns nil only for `.corveil`; `.github` always yields a `GitHubCodeBackend`. The fallback is documented inline.
- **Provider resolution matches convention.** `session.codeProvider ?? session.provider ?? .github` mirrors the established pattern at `SessionService.swift:1210` and the `Session.swift` doc contract (#420 / ADR 0005).
- **Concurrency correct.** `codeBackend(for:)` is `nonisolated`, so the `@MainActor` coordinator calls it synchronously without `await` — appropriate.
- All call sites updated (`AppDelegate` injects the existing `providerManager`; both test suites migrated to `CodeBackend`); `rg` confirms zero remaining inline `provider == .gitlab ? "glab" : "gh"` checks.
- Test stub `FakeCodeBackend` is appropriately minimal — only `provider`/`cliName` are read by the builders.

_Note: the resolved `CodeBackend` is built with `host: nil`, so a self-hosted GitLab session's host isn't threaded through. This is harmless here because prompt rendering only reads `cliName` and `provider` (never `linkedPR`/`ensureMergeLabel`), but worth keeping in mind if these backends ever get used for live calls in this path._

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 1 Green] findings. Behavior-preserving refactor that correctly moves CLI-name selection onto `CodeBackend`, consistent with ADR 0005 and #420.

PR #425, opened 2026-06-04
Wire Cursor agent for Jobs and surface launch failures

### LINKED ISSUE #424: Jobs: Cursor agent launches terminal but nothing happens (no prompt sent, no work started)
**Repro:** Set the default agent to Cursor and run a Job (any prompt — observed with a Summarize-* job from the Jobs list).

**Observed:** The terminal spawns in the new session, but nothing else happens — no Cursor command runs, no prompt is sent, the agent never starts working. The session sits at an empty shell.

**Expected:** Same as Claude Code path — Crow should launch the configured agent in the terminal with the Job's prompt and let it run.

**Suspected cause:** The Job → terminal launch path likely has a working `claude` invocation but no equivalent invocation wired for `cursor` (or whatever Cursor's CLI equivalent is). The terminal-only result suggests the agent launch step is silently skipped for non-`claude-code` agent kinds rather than erroring or falling back.

**Asks:**
1. When the configured agent for the action type can't be launched (binary missing, invocation not wired, etc.), surface an error in the session log / UI instead of leaving the terminal idle.
2. Wire up Cursor as a first-class agent kind (or document which agent kinds Jobs actually supports today and gray out the others in the picker).
3. Consider whether this overlaps with #421 (per-action agent selection) — agent-launch plumbing should be exercised by both.

**Notes:**
- `defaultAgentKind` in `config.json` is currently `claude-code`; the bug was hit after switching it to Cursor.
- Affects the Jobs flow specifically; haven't yet confirmed whether manual `crow new-terminal` with `--command` is affected.

dgershman:
Closes #424

## Summary

- Teach `CursorAgent.autoLaunchCommand` to handle `.job` sessions: read the Jobs pipeline's pre-written `.crow-job-prompt.md` on first launch (mirroring `ClaudeCodeAgent`), and fall back to a bare `agent` on subsequent launches since Cursor has no `--continue` equivalent.
- When any agent returns `nil` from `autoLaunchCommand`, `SessionService.launchAgent` now pastes a one-line diagnostic into the terminal and parks readiness at `.agentLaunched` so the user sees why nothing happened and the deferred-launch loop stops retrying.
- Add tests covering both the first-launch (prompt-file) and resume (bare launch) branches.

## Out of scope

- Codex `.job` support — same pattern is available for a follow-up.
- Cursor `.review` support — review skill remains Claude-only per Phase C.
- Greying out incompatible agents in the per-action picker (#421/#422).
- A new `terminalReadiness` failure state — the in-terminal echo covers the user-visible ask without growing the state machine.

## Test plan

- [x] `swift test --package-path Packages/CrowCursor --filter CursorAgent` — all 7 tests pass (4 existing + 3 new).
- [x] `swift test --arch arm64 --filter DeferredLaunch` — the deferred-launch state machine still passes.
- [ ] Manual: set `defaultAgentKind` to `cursor`, run a Summarize-* Job, confirm Cursor starts with the Job's prompt.
- [ ] Manual negative path: temporarily route a Review through Cursor, confirm the terminal pastes the "⚠️  Crow: agent 'Cursor' cannot launch a review session…" diagnostic instead of sitting idle.
- [ ] Regression: run a Job with Claude Code (default agent restored) to confirm that launch path is unchanged.

### DIFF
--- a/Packages/CrowCursor/Sources/CrowCursor/CursorAgent.swift
+++ b/Packages/CrowCursor/Sources/CrowCursor/CursorAgent.swift
@@ -58,19 +58,35 @@ public struct CursorAgent: CodingAgent {
         autoPermissionMode: Bool,
         telemetryPort: UInt16?
     ) -> String? {
-        // Review-on-Cursor isn't supported in Phase C — the review skill is
-        // Claude-only. Returning nil tells `SessionService.launchAgent` to
-        // log and skip rather than producing a malformed command.
-        guard session.kind == .work else { return nil }
+        let agentPath = findBinary() ?? "agent"
 
-        // Bare `agent` launch — the user types their prompt into the TUI.
-        // No env prefix (Cursor reads `CURSOR_API_KEY` from the shell;
-        // GUI-stored creds are inherited otherwise), no `--continue`
-        // (MVP doesn't auto-resume), no remote-control flag (remote
-        // control is `crow send` typing into the TUI — agent-agnostic,
-        // handled by `SessionService.send`, not a per-launch flag). The
-        // terminal's cwd is already the worktree path.
-        return "agent\n"
+        switch session.kind {
+        case .work:
+            // Bare `agent` launch — the user types their prompt into the TUI.
+            // No env prefix (Cursor reads `CURSOR_API_KEY` from the shell;
+            // GUI-stored creds are inherited otherwise), no `--continue`
+            // (MVP doesn't auto-resume), no remote-control flag (remote
+            // control is `crow send` typing into the TUI — agent-agnostic,
+            // handled by `SessionService.send`, not a per-launch flag).
+            return "\(agentPath)\n"
+        case .job:
+            // Jobs write their first prompt to `.crow-job-prompt.md` in the
+            // worktree before the terminal becomes shell-ready. On first
+            // launch we pass that prompt as argv so Cursor starts working
+            // unattended; on subsequent app restarts we fall back to a bare
+            // `agent` (Cursor has no `--continue` equivalent in MVP), so the
+            // user resumes the TUI rather than re-running the full prompt.
+            if !session.reviewPromptDispatched {
+                let promptPath = (worktreePath as NSString)
+                    .appendingPathComponent(".crow-job-prompt.md")
+                return "\(agentPath) \"$(cat \(promptPath))\"\n"
+            }
+            return "\(agentPath)\n"
+        case .review, .manager:
+            // Review-on-Cursor isn't supported in Phase C — the review skill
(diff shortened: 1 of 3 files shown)


dhilgaertner (approved):
## Code & Security Review

Clean, well-scoped change that wires Cursor's `agent` CLI into the Jobs path and surfaces `nil`-launch failures to the user. It faithfully mirrors the established, already-shipped `ClaudeCodeAgent` pattern.

### Critical Issues
None.

### Security Review
**Strengths:**
- No injection vector in the new job launch. `promptPath` is built from the Crow-controlled `worktreePath` plus a fixed `.crow-job-prompt.md` filename, and the `"$(cat …)"` substitution is double-quoted so the prompt content is passed as a single argv — identical trust model to the existing `ClaudeCodeAgent.swift:89`.
- The terminal diagnostic (`SessionService.swift:528`) interpolates only static `agent.displayName` ("Cursor") and the `session.kind.rawValue` enum string into a single-quoted `echo` — no attacker-controlled data, no shell-injection surface.
- Failure path correctly avoids adding the terminal to `remoteControlActiveTerminals` (nothing actually launched).

**Concerns:**
- None blocking.

### Code Quality
- State-machine handling is sound: parking readiness at `.agentLaunched` (sortOrder 4, the terminal state) in the `nil` branch (`SessionService.swift:534`) stops the deferred-launch relaunch guard (`== .shellReady`, line 476) from retrying — exactly the intended effect, and consistent with the success path at line 545.
- `reviewPromptDispatched` is flipped only after a successful first dispatch; the `nil` guard returns early without flipping it, so a Cursor review/manager session never spuriously marks its prompt as sent. Correct.
- Test relaxation from `cmd == "agent\\n"` to `cmd?.hasSuffix("agent\\n")` is the right call now that `findBinary() ?? "agent"` may resolve an absolute path; the added `.crow-job-prompt.md`/`OTEL_` negative assertions keep the work-session test meaningful.
- All 7 `CursorAgent` tests pass locally (4 existing + 3 new).

### Considerations (non-blocking)
- **Green** — The `"$(cat \(promptPath))"` form word-splits if a worktree path ever contains spaces (the inner `cat` argument is unquoted). This is inherited verbatim from the existing Claude path and Crow's kebab-case worktree naming makes it unreachable in practice, but quoting the inner path (`"$(cat \"\(promptPath)\"\)"`) in both agents would harden it for free in a future cleanup.
- **Green** — On app restart mid-job, Cursor resumes as a bare `agent` TUI with no prompt (no `--continue` equivalent), so an unattended Job silently stops progressing. This is an inherent Cursor-MVP limitation and is explicitly documented in the PR's out-of-scope section; no better option exists today.

### Summary Table

---

PR #430, opened 2026-06-04
Add Bash Conventions section to scaffolded CLAUDE.md

### LINKED ISSUE #426: Scaffolder: add Bash Conventions section to devroot CLAUDE.md (prefer rg / find -delete / xargs over find -exec)
The devroot `CLAUDE.md` scaffolded by `Scaffolder.swift` already has a **Fetching Ticket / PR Data** section explaining why compound bash (`cd && X`, pipes, `echo` banners) defeats Claude Code's prefix-match allowlist. The same mechanic burns users on `find -exec` — the allowlist engine can't see what gets exec'd, so `find ... -exec X` triggers a permission prompt even when both `find` and `X` are individually allowlisted.

Add a sibling **Bash Conventions** section to the scaffolded template that nudges Claude toward `rg`, `find -delete`, and `xargs` instead of `find -exec`. This eliminates a large class of permission prompts across every Crow session without changing the allowlist at all — it's a prompt-level fix.

**Where:** `Scaffolder.swift` (the file that emits `{devRoot}/.claude/CLAUDE.md`). Insert the new section immediately after **Fetching Ticket / PR Data** and before **Known Issues / Corrections** — they share the same conceptual area (avoiding compound bash that defeats allowlists).

**Suggested content to add to the template:**

```markdown
## Bash Conventions

Same allowlist-prefix problem applies to `find -exec`: the rule engine can't see what gets exec'd, so `find ... -exec X` falls back to a permission prompt even when both `find` and `X` are individually allowlisted. Prefer these instead — they avoid the prompt entirely:

| Intent | Use | Not |
|---|---|---|
| Search files for text | `rg PATTERN` (recursive by default, respects `.gitignore`) | `find . -exec grep PATTERN {} \;` |
| Search by file type | `rg PATTERN --type py` / `--type swift` | `find . -name '*.py' -exec grep ...` |
| Find files by name | `find . -name X` (no `-exec`) | — |
| Delete matches | `find . -name X -delete` | `find . -name X -exec rm {} \;` |
| Run a command per match | `find ... -print0 \| xargs -0 CMD` | `find ... -exec CMD {} \;` |
| Filter then count | `find ... \| wc -l` (single pipe is fine) | — |

`rg` (ripgrep) is the default search tool — much faster than `grep -r` and skips ignored files. Install via `brew install ripgrep` if missing.

`find -exec` is essentially never the right tool today — `-delete` and `xargs` cover what it was originally needed for, and both keep the allowlist clean.
```

**Acceptance:**
- New section appears in the scaffolded `{devRoot}/.claude/CLAUDE.md` on next Crow launch.
- Section is idempotent — if the user has hand-edited their CLAUDE.md, scaffolding doesn't clobber their content (follow whatever the existing scaffolder convention is; if it's a full overwrite, fine; if it's section-aware, append-if-missing).
- No change required to permission rules / settings.json — this is prompt-only.

**Related / follow-on (out of scope for this ticket):**
- Preflight check for `ripgrep` in `crow-workspace/setup.sh` — hard-fail with "brew install ripgrep" hint if missing. Worth its own ticket once this guidance is in place.

dgershman:
## Summary

- Adds a new **Bash Conventions** section to the scaffolded `{devRoot}/.claude/CLAUDE.md`, immediately after **Fetching Ticket / PR Data** and before **Known Issues / Corrections**.
- Nudges Claude toward `rg`, `find -delete`, and `xargs` over `find -exec` — same allowlist-prefix problem already documented for `cd && X`, pipes, and `echo` banners (Claude Code's rule engine can't see what `-exec` launches, so the call falls back to a permission prompt even when both `find` and the exec'd binary are individually allowlisted).
- Prompt-only fix — no changes to `settings.json` or permission rules.
- Mirrors the edit in both the repo-level `CLAUDE.md` (used for dev builds) and `Resources/CLAUDE.md.template` (bundled in `.app` builds) so the two template sources stay in sync.

No Swift code change is needed: `Scaffolder.scaffold` already regenerates everything above `## Known Issues / Corrections` on every launch, so the new section lands automatically on the next Crow start and existing user corrections under that section are preserved.

Closes #426

## Test plan

- [ ] Quit Crow, relaunch, and confirm `{devRoot}/.claude/CLAUDE.md` now contains the `## Bash Conventions` section between "Fetching Ticket / PR Data" and "Known Issues / Corrections".
- [ ] Confirm any prior `## Known Issues / Corrections` content in the live `CLAUDE.md` is preserved after the relaunch.
- [ ] Confirm `.claude/settings.json` is unchanged (this is prompt-only guidance).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -117,6 +117,23 @@ echo "=== api ==="; gh api repos/owner/repo/issues/123 | head -120
 
 This keeps the allowlist tight (preferred over broadening it with `cd:*` / `find:*`).
 
+## Bash Conventions
+
+Same allowlist-prefix problem applies to `find -exec`: the rule engine can't see what gets exec'd, so `find ... -exec X` falls back to a permission prompt even when both `find` and `X` are individually allowlisted. Prefer these instead — they avoid the prompt entirely:
+
+| Intent | Use | Not |
+|---|---|---|
+| Search files for text | `rg PATTERN` (recursive by default, respects `.gitignore`) | `find . -exec grep PATTERN {} \;` |
+| Search by file type | `rg PATTERN --type py` / `--type swift` | `find . -name '*.py' -exec grep ...` |
+| Find files by name | `find . -name X` (no `-exec`) | — |
+| Delete matches | `find . -name X -delete` | `find . -name X -exec rm {} \;` |
+| Run a command per match | `find ... -print0 \| xargs -0 CMD` | `find ... -exec CMD {} \;` |
+| Filter then count | `find ... \| wc -l` (single pipe is fine) | — |
+
+`rg` (ripgrep) is the default search tool — much faster than `grep -r` and skips ignored files. Install via `brew install ripgrep` if missing.
+
+`find -exec` is essentially never the right tool today — `-delete` and `xargs` cover what it was originally needed for, and both keep the allowlist clean.
+
 ## Known Issues / Corrections
 
 <!-- Auto-maintained by Claude Code during workspace setup -->
--- a/Resources/CLAUDE.md.template
+++ b/Resources/CLAUDE.md.template
@@ -78,6 +78,23 @@ echo "=== api ==="; gh api repos/owner/repo/issues/123 | head -120
 
 This keeps the allowlist tight (preferred over broadening it with `cd:*` / `find:*`).
 
+## Bash Conventions
+
+Same allowlist-prefix problem applies to `find -exec`: the rule engine can't see what gets exec'd, so `find ... -exec X` falls back to a permission prompt even when both `find` and `X` are individually allowlisted. Prefer these instead — they avoid the prompt entirely:
+
+| Intent | Use | Not |
+|---|---|---|
+| Search files for text | `rg PATTERN` (recursive by default, respects `.gitignore`) | `find . -exec grep PATTERN {} \;` |
+| Search by file type | `rg PATTERN --type py` / `--type swift` | `find . -name '*.py' -exec grep ...` |
+| Find files by name | `find . -name X` (no `-exec`) | — |
+| Delete matches | `find . -name X -delete` | `find . -name X -exec rm {} \;` |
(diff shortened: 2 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

Documentation-only change: adds a **Bash Conventions** section to the scaffolded `CLAUDE.md`, mirrored into `Resources/CLAUDE.md.template`. No Swift, settings, or permission-rule changes.

### Critical Issues
None.

### Security Review
**Strengths:**
- Prompt-only guidance; touches no `settings.json` allowlist or executable code. No new attack surface.
- Steers agents *toward* tighter allowlist hygiene (`rg`, `find -delete`, `xargs` over `find -exec`), reducing the blast radius of broad `find:*` rules.

**Concerns:**
- None.

### Code Quality
- Both files receive a byte-identical 17-line block, inserted in the documented position (after "Fetching Ticket / PR Data", before "Known Issues / Corrections"). Markdown is well-formed and the in-cell table pipes are correctly escaped (`\\|`) for GFM.
- Verified the central claim: `Scaffolder.scaffold` (`Sources/Crow/App/Scaffolder.swift:38-62`) preserves everything from `## Known Issues / Corrections` onward and regenerates the prefix from the template, so the section lands on next launch and existing user corrections are kept. The dev-build path (`loadFromRepo`) reads the repo `CLAUDE.md` directly.
- No scaffolder test asserts on `CLAUDE.md` content, so the addition breaks nothing.

**Green (consider — pre-existing, not introduced here):** `bundledCLAUDEMD()` (`Scaffolder.swift:109`) looks up `forResource: "CLAUDE", withExtension: "md"` (i.e. `CLAUDE.md`), but `scripts/bundle.sh:65-67` bundles the file as `CLAUDE.md.template`. The sibling SKILL lookups correctly use `withExtension: "template"`. As written, the `.template` edit likely never reaches packaged `.app` builds (they fall to the minimal fallback). This is systemic and affects the whole `CLAUDE.md`, not just this section — worth a follow-up, but mirroring into the template remains the correct source-of-truth move and does not block this PR.

### Summary Table

---

PR #432, opened 2026-06-04
Don't steal focus on crow:auto pickup (#429)

### LINKED ISSUE #429: Don't steal focus when a new review or active workspace spins up
**Observed:** When a new review or new active workspace gets spun up (e.g. auto-pickup of a `crow:auto` ticket, auto-review of a PR, a new Job kicking off), Crow forces the foreground UI to switch to that newly-created session. If the user is mid-task in another session, they're yanked out of it.

**Expected:** Spin-up happens in the background. The new session appears in the sidebar (and ideally bumps a notification per existing settings) but the user stays on whatever they're currently looking at. Focus only changes when the user clicks the new session.

**Acceptance:**
- Spawning a new session (any path: ticket auto-pickup, PR auto-review, Job, manual `crow new-session`) does **not** change the selected session in the UI.
- A subtle indicator (badge / sound / system notification per existing `notifications.eventSettings`) shows that a new session exists, without taking focus.
- If the user has *no session selected* (fresh app launch, empty sidebar), it's fine to default-select the new one — only the "user is actively working elsewhere" case must not steal focus.
- Applies to both reviews and active/coding workspaces — same behavior across the two creation paths.

**Notes:**
- This is the #1 disruption to flow when running multiple things in parallel — every auto-pickup interrupts whatever's in front.
- Existing notification surface (chimes, system notifications) already announces these events; the focus-steal is redundant *and* destructive.

dgershman:
## Summary

- Auto-pickup of `crow:auto`-labeled issues was routing through `appState.onWorkOnIssue`, which is shared with the manual "Start Working" button and unconditionally switches the sidebar to the Manager tab — yanking the user out of whatever session they were on.
- This PR inlines the "send `/crow-workspace <url>` to the Manager terminal" step in `onAutoCreateRequest` and skips the focus switch. The manual button keeps its current behavior (user explicitly asked for that workspace).
- The Manager's `/crow-workspace` skill ultimately calls `crow new-session`, whose RPC path already leaves `selectedSessionID` alone — so the new session appears in the sidebar without disturbing the user's view, matching the existing silent-spawn behavior of PR auto-review and Job kickoffs.
- `notifyAutoWorkspaceCreated` still fires, so the existing notification surface (chime / system notification) continues to announce the event.

After this change, all four auto/CLI spawn paths called out in the ticket (`crow:auto` pickup, PR auto-review, Job, manual `crow new-session`) no longer change `selectedSessionID`.

## Test plan

- [ ] Build: `make ghostty && swift build --arch arm64` succeeds.
- [ ] Select a non-Manager session, label a test issue `crow:auto`, confirm Manager terminal receives `/crow-workspace <url>` and the system notification fires, while the selected session does **not** change.
- [ ] Regression: click "Start Working" on a ticket from the Ticket Board and confirm the UI still switches to the Manager tab.
- [ ] Regression: confirm PR auto-review and Job kickoff still create their sessions in the sidebar without changing focus.

Closes #429

### DIFF
--- a/Sources/Crow/App/AppDelegate.swift
+++ b/Sources/Crow/App/AppDelegate.swift
@@ -687,7 +687,18 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         }
         tracker.onAutoCreateRequest = { [weak self] issue in
             guard let self else { return }
-            self.appState.onWorkOnIssue?(issue.url)
+            // Send `/crow-workspace <url>` to the Manager terminal WITHOUT
+            // switching the selected session. The manual "Start Working"
+            // button routes through `onWorkOnIssue` (which does select
+            // Manager) because the user explicitly asked for that
+            // workspace; an auto-pickup is background work and must not
+            // yank the user out of their current session (#429). The
+            // sidebar still updates and `notifyAutoWorkspaceCreated`
+            // already surfaces the event.
+            if let managerTerminals = self.appState.terminals[AppState.managerSessionID],
+               let managerTerminal = managerTerminals.first {
+                TerminalRouter.send(managerTerminal, text: "/crow-workspace \(issue.url)\n")
+            }
             self.notificationManager?.notifyAutoWorkspaceCreated(issue)
         }
         tracker.onPRStatusTransitions = { [weak self] transitions in


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface. The change narrows behavior (drops a UI focus switch) rather than adding input handling, network, or privilege paths.
- `issue.url` was already being passed to the same `/crow-workspace` flow via `onWorkOnIssue` before this PR; no new untrusted data is introduced into the terminal command.

**Concerns:**
- None.

### Code Quality
- The inlined send block (`AppDelegate.swift:698-701`) is an exact mirror of `onWorkOnIssue` (`AppDelegate.swift:585-594`) with only the `selectedSessionID` write dropped — which is precisely the intent of #429. Correct and well-scoped.
- The optional-binding guard safely no-ops when no Manager terminal exists, matching the original closure's `guard`.
- Threading is unchanged: the previous code (`self.appState.onWorkOnIssue?(issue.url)`) ran in this same closure and already touched `appState`, so no new concurrency concern is introduced.
- The explanatory comment accurately documents the manual-vs-auto distinction and the rationale.
- 🟢 *Consider:* the `TerminalRouter.send(managerTerminal, text: "/crow-workspace ...")` lookup-and-send is now duplicated between `onWorkOnIssue` and `onAutoCreateRequest`. A small private helper (e.g. `sendToManagerTerminal(_:)`) would DRY this up and keep the two call sites from drifting. Non-blocking.

### Summary Table

---

PR #434, opened 2026-06-04
Show actual agent on terminal tab label and icon (CROW-427)

### LINKED ISSUE #427: Tabs always label the agent as "Claude Code" regardless of configured agent
**Observed:** A terminal tab in a session always shows "Claude Code" as the agent label, even when the configured `defaultAgentKind` (or per-action agent, per #421) is something else (e.g. Cursor, Codex).

**Expected:** The tab label should reflect the actual agent running in that terminal — "Cursor", "Codex", "Claude Code", etc. — matching the agent kind Crow used to spawn it.

**Suspected location:** Tab title / agent-label rendering in the sidebar / terminal header. Likely a hardcoded string or default-fallback that was added when Claude Code was the only supported agent.

**Acceptance:**
- Tab agent label matches the agent actually launched in that terminal.
- Default fallback (if the agent kind is unset or unknown) shows the kind string, not "Claude Code".
- Cross-check: tabs for sessions launched under different agents in the same workspace show different labels.

**Related:** #424 (Cursor agent launches terminal but does nothing) — both touch the agent-kind plumbing; fix may share code.

dgershman:
## Summary
- Terminal tabs hardcoded `"Claude Code"` and the `sparkles` icon for every managed terminal, ignoring the session's configured agent.
- Resolve tab label and SF Symbol from `session.agentKind` via `AgentRegistry` at both creation time (the four call sites that build managed `SessionTerminal`s) and at render time (the `TerminalTabBar` glyph). Adds `AgentKind.displayName` / `AgentKind.iconSystemName` extensions so every call stays a one-liner and the per-ticket fallback rule lives in one place.
- Per the ticket, when no agent is registered for the kind the label falls back to the raw kind string (e.g. `"cursor"`) — not silently to `"Claude Code"`. The icon falls back to a neutral `sparkles` so the tab doesn't draw an empty SF Symbol box.

## Test plan
- [ ] Build: \`make ghostty && swift build --arch arm64\`
- [ ] Create a new work session with \`defaultAgentKind = cursor\`; confirm the tab reads **Cursor** with the \`cursorarrow.rays\` glyph, and the session-header "Agent:" row agrees.
- [ ] With per-action override \`agentsByKind[review] = codex\` (CROW-421), trigger a review session; confirm the tab reads **OpenAI Codex** with the \`terminal.fill\` glyph.
- [ ] Trigger a job session under a different agent in the same workspace; confirm sibling tabs show different labels **and** different glyphs.
- [ ] Fallback: deregister an agent kind, create a session of that kind, and confirm the tab falls back to the raw kind string (not "Claude Code") and a neutral sparkles glyph.
- [ ] Pre-existing sessions: their tabs still read "Claude Code" because the name is persisted — expected (those sessions actually ran Claude Code). Double-click rename still works and persists.

Closes #427

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Agent/AgentKind.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Agent/AgentKind.swift
@@ -20,3 +20,20 @@ public struct AgentKind: Hashable, Sendable, Codable, RawRepresentable {
     /// The Cursor agent.
     public static let cursor = AgentKind(rawValue: "cursor")
 }
+
+public extension AgentKind {
+    /// Human-readable label for this kind, resolved through `AgentRegistry`.
+    /// Falls back to `rawValue` (e.g. `"cursor"`) when no agent is registered
+    /// for the kind — per CROW-427 the fallback must NOT silently be
+    /// `"Claude Code"`.
+    var displayName: String {
+        AgentRegistry.shared.agent(for: self)?.displayName ?? rawValue
+    }
+
+    /// SF Symbol name for this kind, resolved through `AgentRegistry`.
+    /// Falls back to a neutral `"sparkles"` when no agent is registered, so
+    /// the tab UI doesn't render an empty SF Symbol box.
+    var iconSystemName: String {
+        AgentRegistry.shared.agent(for: self)?.iconSystemName ?? "sparkles"
+    }
+}
--- a/Packages/CrowCore/Tests/CrowCoreTests/AgentKindDisplayTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AgentKindDisplayTests.swift
@@ -1,0 +1,26 @@
+import Foundation
+import Testing
+@testable import CrowCore
+
+// Locks in the fallback semantics CROW-427 specifies: when a kind isn't
+// registered in `AgentRegistry`, `displayName` must return the kind's raw
+// value (e.g. `"cursor"`) and NOT silently degrade to `"Claude Code"`.
+// `iconSystemName` must fall back to a neutral SF Symbol (`"sparkles"`) so
+// the tab UI doesn't render an empty glyph box. A future refactor that
+// "simplifies" `?? rawValue` to `?? "Claude Code"` would reintroduce the
+// exact bug this PR fixes; these tests are the regression guard.
+
+@Test func displayNameFallsBackToRawValueWhenKindUnregistered() {
+    // Use a kind value that no shipped agent claims so the lookup misses
+    // even if the process-wide `AgentRegistry` has been populated by
+    // another test.
+    let unknown = AgentKind(rawValue: "crow-427-unregistered-fallback")
+    #expect(AgentRegistry.shared.agent(for: unknown) == nil)
+    #expect(unknown.displayName == "crow-427-unregistered-fallback")
+}
+
+@Test func iconSystemNameFallsBackToSparklesWhenKindUnregistered() {
+    let unknown = AgentKind(rawValue: "crow-427-unregistered-icon-fallback")
+    #expect(AgentRegistry.shared.agent(for: unknown) == nil)
+    #expect(unknown.iconSystemName == "sparkles")
+}
--- a/Packages/CrowUI/Sources/CrowUI/SessionDetailView.swift
+++ b/Packages/CrowUI/Sources/CrowUI/SessionDetailView.swift
@@ -366,6 +366,12 @@ public struct SessionDetailView: View {
             }
         } else {
             VStack(spacing: 0) {
+                // Icon is resolved live from `session.agentKind` on every
+                // render, while the tab label is snapshotted into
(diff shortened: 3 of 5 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Clean, well-scoped change. Verified locally: `CrowCore` builds and all 217 tests pass. The four managed-terminal creation sites (`AppDelegate.swift:1380`, `SessionService.swift:1208/1435/1662`) are all updated, and the Manager terminal at `SessionService.swift:686` is correctly left untouched (it's named after the session, not a managed agent terminal). Icon mapping resolved through the registry matches the test plan exactly (`sparkles` / `cursorarrow.rays` / `terminal.fill`).

### Critical Issues
None.

### Security Review
**Strengths:**
- No new input surface. The `new-terminal` RPC still enforces `isPathWithinDevRoot` for `cwd`; the agent-kind lookup only reads existing in-memory state.
- `AgentRegistry.agent(for:)` is lock-guarded; the new `displayName`/`iconSystemName` accessors inherit that safely from both the render (main) thread and creation paths.

**Concerns:**
- None.

### Code Quality
- **[Yellow] Missing test coverage for the new fallback semantics** (`Packages/CrowCore/Sources/CrowCore/Agent/AgentKind.swift:29,36`). The ticket explicitly emphasizes that an unregistered kind must fall back to the raw `rawValue` (e.g. `\"cursor\"`) and **not** silently to `\"Claude Code\"`, with the icon falling back to a neutral `sparkles`. That regression-prone branch has no unit test — a future \"simplification\" of `?? rawValue` to `?? \"Claude Code\"` would pass CI. A few-line test in `CrowCoreTests` asserting `AgentKind(rawValue: \"unregistered\").displayName == \"unregistered\"` and `.iconSystemName == \"sparkles\"` would lock in the behavior the ticket calls out as the whole point. Given Crow's verdict rule (any Yellow → request changes), folding this in now keeps it a single round trip.

- The tab **label** is snapshotted at creation (persisted on `SessionTerminal.name`) while the **icon** is resolved live at render from `session.agentKind`. This is intentional and documented in the PR body (pre-existing sessions keep their persisted name), and within a session all managed terminals share one `agentKind`, so the single per-bar `managedAgentIcon` is correct. Worth a one-line code comment noting the label-is-snapshot / icon-is-live split so a future reader doesn't \"fix\" the apparent asymmetry.
- The new `public init` on `TerminalTabBar` is purely additive (default `managedAgentIcon` arg), so no source break for the single in-module caller.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The implementation is correct and tested-green; the single Yellow is adding a unit test for the ticket's headline fallback behavior.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Pure presentation change (tab label + SF Symbol). No new input parsing, no command construction, no auth/permission surface touched. The `new-terminal` RPC keeps its existing `isPathWithinDevRoot` traversal guard untouched.
- The AppDelegate refactor only hoists the existing `session` lookup; no behavioral change to the claude-resolution / `--rc` injection path.

**Concerns:**
- None.

### Code Quality
- `AgentKind.displayName`/`iconSystemName` centralize the fallback rule in one place, so the four `SessionTerminal` creation sites and the render-time glyph stay one-liners. Good single-source-of-truth design.
- The snapshot-label vs. live-icon asymmetry (label persisted into `terminal.name` for rename UX, icon resolved live from `session.agentKind`) is intentional and clearly documented at both `SessionDetailView.swift:366` and `AgentKind.swift`. Verified `TerminalTabBar` has exactly one caller and the new `public init` defaults `managedAgentIcon` to `"sparkles"`, so the change is backward-compatible.
- Verified locally: `CrowCore` builds clean and both new regression tests pass (`displayNameFallsBackToRawValueWhenKindUnregistered`, `iconSystemNameFallsBackToSparklesWhenKindUnregistered`). CrowUI requires the GhosttyKit xcframework (`make ghostty`) so it wasn't built here, but the change is a mechanical signature addition verified against its only caller.

- For an *unregistered* agent kind, the tab shows the raw kind string + a neutral `sparkles` glyph (per the ticket), while the session-header "Agent:" row at `SessionDetailView.swift:104` hides entirely (it only renders when `AgentRegistry` resolves the kind). The PR description says these "agree" — true for registered agents, but they diverge in the unregistered case. Both behaviors are individually correct and spec-compliant; just noting the header could optionally adopt the same raw-kind fallback for full symmetry.
- The unregistered-kind icon fallback (`"sparkles"`) is identical to Claude Code's own icon, so an unknown agent renders with Claude Code's glyph (the label still differs). The accompanying test comment acknowledges this is deliberate to avoid an empty SF Symbol box.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. Focused, well-tested, well-documented fix that does exactly what CROW-427 asks.

---

PR #435, opened 2026-06-04
Enable Cursor as a review-capable agent

### LINKED ISSUE #431: Enable Cursor as a review-capable agent (currently restricted to coding)
**Observed:** Attempting to launch a review session with Cursor as the configured agent produces:

> ⚠️ Crow: agent "Cursor" cannot launch a review session. Switch the agent for this action type (Settings → Agents) or pick a session kind this agent supports.

So Crow has an agent → session-kind capability matrix, and Cursor is currently flagged as not supporting "review" session kind.

**Question / ask:** Why? Cursor has a CLI (`cursor-agent` / `cursor`) and can be prompted with arbitrary instructions — there's no inherent reason it can't run the same review prompt Claude Code runs. Either:

1. **Enable it** — add `review` to Cursor's supported session-kind set, wire the existing review-prompt template through Cursor's CLI the same way it goes through Claude Code's. Let users pick their reviewer.
2. **Document the real reason it's gated** — if Cursor's CLI is missing some capability the review flow needs (no non-interactive mode, no programmatic prompt input, no way to detect completion, output format incompatible with how Crow parses review results), say so in the error message instead of just "this agent doesn't support this kind." Users hitting this can then know whether to wait for support or use a different agent.

**Acceptance (path 1):**
- Cursor appears as a valid choice in the per-action agent dropdown for review sessions (depends on #421).
- A Cursor-backed review session launches the agent with the same review prompt template that Claude Code receives.
- Review output (comments, approve/request-changes signal, etc.) is captured from Cursor the same way it's captured from Claude Code.

**Acceptance (path 2):**
- The warning message names the specific blocker ("Cursor's CLI doesn't support X") instead of the generic "this agent doesn't support this kind."

**Related:**
- #421 (per-action agent selection) — the matrix this warning enforces should be visible to that dropdown so unsupported pairings are grayed out, not error-on-launch.
- #424 (Cursor agent launches terminal but does nothing for coding) — the inverse problem. Coding silently no-ops; review at least errors cleanly. Both point at incomplete Cursor agent integration.
- #427 (tab labels hardcoded to "Claude Code") and #428 (Cursor cards don't update color) — also Cursor-integration work; reasonable to bundle planning across these four.

dgershman:
## Summary

- Lets users pick Cursor as the agent for review sessions (per #422's per-action picker) and have the review actually launch instead of erroring in the terminal.
- `CursorAgent.autoLaunchCommand` now handles `.review` the same way it handles `.job`: reads `.crow-review-prompt.md` on first launch, falls back to a bare `agent` on resume. `.manager` still returns `nil` (unchanged contract).
- `SessionService.buildReviewPrompt` is now agent-aware. Claude Code keeps the terse `/crow-review-pr <URL>` slash command and resolves it via its SKILL engine. Cursor gets the SKILL body inlined with `$ARGUMENTS` substituted to the PR URL and the attribution swapped to "via Cursor" — no second-file indirection, no slash-command engine required.

## Why this approach

Cursor's `agent` CLI doesn't have a slash-command / skill engine like Claude Code, so we can't reuse the one-liner. But the existing `skills/crow-review-pr/SKILL.md` is already structured as a self-contained instruction document — once `$ARGUMENTS` is expanded, it works as a plain prompt. Inlining keeps the design symmetric (both agents read the same prompt file path, gated by `reviewPromptDispatched`) without forking the review instructions into a second source of truth.

The `.claude/skills/crow-review-pr/` copy in the clone stays — it's harmless when Cursor runs and still useful if the user later switches the review agent back to Claude Code in the same clone.

## Test plan

- [x] `swift build --arch arm64` — clean build.
- [x] `swift test --arch arm64 --filter CursorAgent` (in `Packages/CrowCursor`) — 9/9 pass, including new `autoLaunchCommandReviewSessionFirstLaunch`, `autoLaunchCommandReviewSessionSubsequentLaunch`, and `autoLaunchCommandManagerSessionUnsupported` cases.
- [x] `swift test --arch arm64 --filter AppConfigAgentKind` (in `Packages/CrowCore`) — 9/9 pass; per-action override mechanism unchanged.
- [ ] Manual: Settings → Agents → set "Agent for reviews" = Cursor. From Reviews board, launch a review on a real PR; expect the Cursor TUI to come up with the expanded prompt as argv, not the "cannot launch a review session" warning.
- [ ] Manual: switch back to Claude Code, launch a review on a different PR; expect the `/crow-review-pr <URL>` slash-command path unchanged.

Closes #431

### DIFF
--- a/Packages/CrowCursor/Sources/CrowCursor/CursorAgent.swift
+++ b/Packages/CrowCursor/Sources/CrowCursor/CursorAgent.swift
@@ -69,22 +69,30 @@ public struct CursorAgent: CodingAgent {
             // control is `crow send` typing into the TUI — agent-agnostic,
             // handled by `SessionService.send`, not a per-launch flag).
             return "\(agentPath)\n"
-        case .job:
-            // Jobs write their first prompt to `.crow-job-prompt.md` in the
-            // worktree before the terminal becomes shell-ready. On first
-            // launch we pass that prompt as argv so Cursor starts working
-            // unattended; on subsequent app restarts we fall back to a bare
+        case .job, .review:
+            // Jobs and reviews share the same dispatch shape: a pre-written
+            // initial prompt file (`.crow-job-prompt.md` / `.crow-review-prompt.md`)
+            // is passed as argv on first launch so Cursor starts working
+            // unattended. On subsequent app restarts we fall back to a bare
             // `agent` (Cursor has no `--continue` equivalent in MVP), so the
             // user resumes the TUI rather than re-running the full prompt.
+            //
+            // Review prompts are agent-aware: SessionService.buildReviewPrompt
+            // inlines the crow-review-pr SKILL body for Cursor so the `agent`
+            // CLI gets a self-contained brief — no slash-command engine
+            // needed (#431). `reviewPromptDispatched` gates both kinds.
             if !session.reviewPromptDispatched {
+                let promptFile = session.kind == .review
+                    ? ".crow-review-prompt.md"
+                    : ".crow-job-prompt.md"
                 let promptPath = (worktreePath as NSString)
-                    .appendingPathComponent(".crow-job-prompt.md")
+                    .appendingPathComponent(promptFile)
                 return "\(agentPath) \"$(cat \(promptPath))\"\n"
             }
             return "\(agentPath)\n"
-        case .review, .manager:
-            // Review-on-Cursor isn't supported in Phase C — the review skill
-            // is Claude-only. Manager sessions don't auto-launch an agent.
+        case .manager:
+            // Manager sessions never auto-launch an agent — Crow drives them
+            // externally. Returning nil here is the contract, not a gap.
             return nil
         }
     }
--- a/Packages/CrowCursor/Tests/CrowCursorTests/CursorAgentTests.swift
+++ b/Packages/CrowCursor/Tests/CrowCursorTests/CursorAgentTests.swift
@@ -50,7 +50,12 @@ struct CursorAgentTests {
         #expect(cmd?.contains(".crow-job-prompt.md") == false)
     }
 
-    @Test func autoLaunchCommandReviewSessionUnsupported() {
+    @Test func autoLaunchCommandReviewSessionFirstLaunch() {
+        // First review launch (reviewPromptDispatched == false) should pass
+        // the pre-written `.crow-review-prompt.md` as argv so Cursor starts
+        // the review unattended. The prompt file content is agent-aware
+        // (inlined SKILL body for Cursor) — see SessionService.buildReviewPrompt
+        // and #431.
         let session = Session(name: "review", kind: .review, agentKind: .cursor)
         let cmd = agent.autoLaunchCommand(
             session: session,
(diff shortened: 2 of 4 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Solid, well-scoped change that reuses the existing job-dispatch shape for reviews and keeps the design symmetric. Build is clean and the 9 `CursorAgent` tests pass locally (`swift build --arch arm64`; `swift test --filter CursorAgent`). One should-fix gap holds back approval.

### Critical Issues
None.

### Security Review
**Strengths:**
- The first-launch command `\"$(cat <promptPath>)\"` passes the inlined SKILL body to Cursor through a *quoted* command substitution. Shell does not re-parse command-substitution output, so the backticks, `$(...)`, and embedded double-quotes in the SKILL body are inert — no command injection despite the much larger payload. Same pattern jobs already use.
- Attribution swap is complete and correct: `via Claude Code` → `via Cursor` hits exactly the two attribution lines (skill lines 134/142) and nothing else; the canonical URL is untouched. `AttributionSkillTests` keeps `skills/.../SKILL.md` and the bundled `.template` byte-identical, so the `$ARGUMENTS` / `via Claude Code` preconditions for the swap are enforced on both build paths.
- No secrets, no new external surface; `agentKind(for: .review)` resolves the per-action override on the main actor before detaching, preserving the #404 non-blocking contract.

**Concerns:**
- None blocking.

### Code Quality
- 🟡 **Missing test for the new Cursor branch of `buildReviewPrompt`** (`Sources/Crow/App/SessionService.swift:1779-1804`). This branch — `$ARGUMENTS` → PR URL substitution plus the `via Claude Code` → `via Cursor` swap — is the literal payload posted to GitHub and the actual deliverable of #431, yet no test exercises it. `CursorAgentTests` only assert *which file path* lands in argv (`.crow-review-prompt.md`), not the file's *content*. A reworded attribution line or a renamed placeholder would silently produce a misattributed Cursor review (or a literal `PR $ARGUMENTS` brief) with nothing to catch it. `buildReviewPrompt` is `nonisolated private static` and `AttributionSkillTests` already uses `@testable import Crow`, so a direct assertion (`.cursor` → contains PR URL, contains `via Cursor`, no `via Claude Code`, no `$ARGUMENTS`; `.claudeCode` → terse one-liner) is cheap to add.
- 🟢 The inlined SKILL body Cursor receives still carries Claude-Code-specific framing — \"## Important: Sandbox Bypass — All `gh` and `git` commands require `dangerouslyDisableSandbox: true`\". Cursor's `agent` CLI has no such flag; it's harmless instructional noise, but consider a follow-up to make the brief agent-neutral.
- 🟢 `\"$(cat <promptPath>)\"` interpolates the worktree path unquoted inside the substitution — a clone path containing spaces would split `cat`'s args. Pre-existing (jobs use the same construction) and review clone paths are devRoot/slug-derived, so not a regression; noting only for completeness.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The lone Yellow is the missing unit test for the Cursor prompt-building branch; adding it lands the core of #431 under test in the same round trip. Everything else is solid.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- **No shell-injection surface from the inlined prompt.** `CursorAgent.autoLaunchCommand` launches review prompts with `agent "$(cat <promptPath>)"` (CursorAgent.swift:90). The SKILL body — which contains backticks, `$(...)`, embedded double-quotes, and fenced `gh pr review --body "..."` snippets — is command-substitution *data* and is not re-tokenized by the shell. This is the same mechanism Jobs already use; reviews are larger but no riskier.
- **Attribution swap is complete and correct.** `cursorReviewPrompt` (SessionService.swift:1797) replaces *all* `via Claude Code` occurrences (SKILL.md:134 and :142) with `via Cursor`, so a Cursor-run review can't post a Claude-attributed body. The canonical URL is preserved (`$ARGUMENTS` substitution only touches the placeholder).
- Agent kind is captured once on the main actor (SessionService.swift:1403) and threaded into the detached `prepareReviewClone`, avoiding actor-isolation hazards.

**Concerns:** None.

### Code Quality
- **Dispatch-flag symmetry verified.** `reviewPromptJustDispatched` (SessionService.swift:513) already gates on `.review || .job` agent-agnostically, so the Cursor review path correctly passes the prompt on first launch and resumes bare on restart — no re-issued brief. The new `autoLaunchCommandReviewSessionSubsequentLaunch` test locks this in.
- The `.manager → nil` contract is preserved and now has an explicit regression test (`autoLaunchCommandManagerSessionUnsupported`).
- Helper split (`buildReviewPrompt` dispatch vs. `cursorReviewPrompt` substitution) is sensible — it lets the substitution be unit-tested without the `Scaffolder.bundledReviewSkill()` file-resolution fallback. Good test coverage including the unknown-future-agent fallback case.

**Verification run locally:**
- `swift test --filter CursorAgent` (Packages/CrowCursor) → **9/9 pass**
- `swift test --filter AppConfigAgentKind` (Packages/CrowCore) → **9/9 pass**
- `SessionServiceReviewPromptTests` lives in the main Crow target, which can't link in a review clone (missing `GhosttyKit.xcframework` binary — a clone limitation, not a PR defect); the logic is trivial substitution and was confirmed by the author's `swift build --arch arm64`.

### Considerations (non-blocking, Green)
- The inlined SKILL body carries Claude-Code-specific phrasing into the Cursor prompt (`dangerouslyDisableSandbox: true`, `User invokes /crow-review-pr`). Cursor will simply ignore the unknown tool-param guidance, so reviews still function — but a future follow-up could neutralize agent-specific language in the shared SKILL so both agents get clean instructions. The PR explicitly accepts this single-source-of-truth tradeoff, which is reasonable.
- `buildReviewPrompt` still takes `prTitle`/`repoSlug`/`prNumber` that neither branch uses (pre-existing; harmless, no Swift unused-param warning).

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. Symmetric design, injection-safe launch, complete attribution swap, and solid test coverage.

---

PR #438, opened 2026-06-04
Add per-Manager agent setting (CROW-433)

### LINKED ISSUE #433: Manager: per-Manager agent setting (default falls back to global agent)
**Today:** The Manager tab always launches Claude Code, regardless of any agent configuration.

**Ask:** Add a separate setting for "Manager agent" so the user can choose which CLI runs in the Manager tab (Claude Code, Cursor, Codex, etc.). When unset, fall back to the global `defaultAgentKind` so existing users see no change.

**Why it's separate from #421 (per-action coding vs review):**
- The Manager has a distinct role — it orchestrates other sessions via the `crow` CLI; it doesn't implement tickets or review PRs.
- The "right" Manager agent for a given user may differ from their coding/review picks (e.g. prefer Claude Code as Manager for its tool-use depth, but use Cursor for coding). Forcing all three to share a single dropdown takes that away.

**Acceptance:**
- New setting in `config.json`, e.g. `agents.manager` (or extend the action-type map proposed in #421 with a `manager` key).
- When unset (the default for existing installs), the Manager session uses `defaultAgentKind` — no behavior change on upgrade.
- The setting is exposed in the Agents settings UI alongside the coding / review dropdowns from #421.
- Manager session (UUID `00000000-0000-0000-0000-000000000000`) launches the configured agent on app startup / re-create.
- Switching the setting takes effect on next Manager session spawn (a restart-app message is acceptable if hot-swap is hard).

**Related:**
- #421 (per-action agent selection) — natural to land them together. If #421 generalizes to a map of action-type → agent, this ticket just adds `manager` as a third key in that map.
- #424 (Cursor coding silently no-ops) and #431 (Cursor reviews rejected) — both gate whether Cursor is a real choice for Manager. Worth verifying Cursor can actually run as Manager before claiming it as an option.

dgershman:
## Summary

- Manager tab is no longer pinned to Claude Code — `agents.manager` in `config.json` now picks the CLI that runs in the Manager session.
- When unset, falls back to `defaultAgentKind` so existing installs see no behavior change.
- Each agent owns its own Manager launch command (`CodingAgent.managerLaunchCommand(...)`); SessionService dispatches via `AgentRegistry` instead of hardcoding `claude`.
- Settings UI exposes the new picker alongside the coding/review/job dropdowns from #421. A switch takes effect on next Manager spawn (app relaunch).
- Existing Manager session's `agentKind` is refreshed on app start if config changed.

## Test plan

- [x] `swift test --arch arm64 --parallel` — all 171 root tests pass, including new `managerCommandDispatchesByAgentKind`
- [x] `swift test --package-path Packages/CrowCore --arch arm64` — 218 tests pass, including new `appConfigAgentResolverFallsBackToDefaultForManager` and `appConfigAgentResolverHonorsExplicitManagerOverride`
- [x] `swift test --package-path Packages/CrowClaude / CrowCursor / CrowCodex --arch arm64` — all pass
- [ ] Manual: Settings → Agents → set "Agent for Manager" → Cursor → relaunch → Manager spawns `cursor-agent`; switch back to "Use default" → relaunch → Manager spawns Claude Code
- [ ] Manual upgrade safety: with no `agents.manager` in `config.json`, Manager still spawns Claude Code

Closes #433.

### DIFF
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeCodeAgent.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeCodeAgent.swift
@@ -116,4 +116,18 @@ public struct ClaudeCodeAgent: CodingAgent {
             prompt: prompt
         )
     }
+
+    public func managerLaunchCommand(
+        sessionName: String,
+        remoteControlEnabled: Bool,
+        autoPermissionMode: Bool,
+        telemetryPort: UInt16?
+    ) -> String {
+        let claudePath = findBinary() ?? "claude"
+        return claudePath + ClaudeLaunchArgs.argsSuffix(
+            remoteControl: remoteControlEnabled,
+            sessionName: sessionName,
+            autoPermissionMode: autoPermissionMode
+        )
+    }
 }
--- a/Packages/CrowCodex/Sources/CrowCodex/OpenAICodexAgent.swift
+++ b/Packages/CrowCodex/Sources/CrowCodex/OpenAICodexAgent.swift
@@ -88,4 +88,16 @@ public struct OpenAICodexAgent: CodingAgent {
             prompt: prompt
         )
     }
+
+    public func managerLaunchCommand(
+        sessionName: String,
+        remoteControlEnabled: Bool,
+        autoPermissionMode: Bool,
+        telemetryPort: UInt16?
+    ) -> String {
+        // Codex's Manager is a plain TUI in the devRoot — no auto-prompt,
+        // no remote-control, no auto-permission knob (CROW-433).
+        let codexPath = findBinary() ?? "codex"
+        return "\(codexPath)\n"
+    }
 }
--- a/Packages/CrowCodex/Sources/CrowCodex/OpenAICodexAgent.swift
+++ b/Packages/CrowCodex/Sources/CrowCodex/OpenAICodexAgent.swift
@@ -96,8 +96,9 @@ public struct OpenAICodexAgent: CodingAgent {
         telemetryPort: UInt16?
     ) -> String {
         // Codex's Manager is a plain TUI in the devRoot — no auto-prompt,
-        // no remote-control, no auto-permission knob (CROW-433).
-        let codexPath = findBinary() ?? "codex"
-        return "\(codexPath)\n"
+        // no remote-control, no auto-permission knob (CROW-433). Terminal
+        // backend appends the submitting Enter, so we return the bare
+        // command without a trailing newline.
+        return findBinary() ?? "codex"
     }
 }
--- a/Packages/CrowCore/Sources/CrowCore/Agent/CodingAgent.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Agent/CodingAgent.swift
@@ -71,4 +71,36 @@ public protocol CodingAgent: Sendable {
         worktreePath: String,
         prompt: String
     ) async throws -> String
+
+    /// Build the shell command that the Manager tab uses to launch this
+    /// agent in the devRoot. Unlike `autoLaunchCommand`, this is the
+    /// terminal's pre-populated `command` string — it runs before the
+    /// shell prompt is ready, with no auto-prompt or `--continue` flag.
+    ///
+    /// `sessionName` labels the agent's session in claude.ai's Remote
+    /// Control panel (and analogous systems if other agents support it).
+    /// `autoPermissionMode` mirrors `ClaudeCodeAgent`'s `--permission-mode auto`
+    /// — agents that don't surface this concept can ignore it. `telemetryPort`
+    /// is passed through for consistency with `autoLaunchCommand`; most
+    /// agents won't need it for Manager terminals (CROW-433).
+    func managerLaunchCommand(
+        sessionName: String,
+        remoteControlEnabled: Bool,
+        autoPermissionMode: Bool,
+        telemetryPort: UInt16?
+    ) -> String
+}
+
+public extension CodingAgent {
+    /// Default Manager launch command: invoke the agent's CLI binary by
+    /// name with no extra flags. Agents with richer launch ergonomics
+    /// (Claude Code's `--rc`/`--name`/`--permission-mode`) override this.
+    func managerLaunchCommand(
+        sessionName: String,
+        remoteControlEnabled: Bool,
+        autoPermissionMode: Bool,
+        telemetryPort: UInt16?
+    ) -> String {
+        return "\(launchCommandToken)\n"
+    }
 }
--- a/Packages/CrowCore/Sources/CrowCore/Agent/CodingAgent.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Agent/CodingAgent.swift
@@ -93,14 +93,15 @@ public protocol CodingAgent: Sendable {
 
 public extension CodingAgent {
     /// Default Manager launch command: invoke the agent's CLI binary by
-    /// name with no extra flags. Agents with richer launch ergonomics
-    /// (Claude Code's `--rc`/`--name`/`--permission-mode`) override this.
+    /// name with no extra flags. The terminal backend (tmux/Ghostty) owns
+    /// the submitting Enter — return the raw command without a trailing
+    /// newline so the convention is uniform across agents (CROW-433 review).
     func managerLaunchCommand(
         sessionName: String,
         remoteControlEnabled: Bool,
         autoPermissionMode: Bool,
         telemetryPort: UInt16?
     ) -> String {
-        return "\(launchCommandToken)\n"
+        return launchCommandToken
     }
 }
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -65,11 +65,9 @@ public final class AppState {
     public var agentsByKind: [String: AgentKind] = [:]
 
     /// Resolve the agent that should drive a newly-created session of the
-    /// given kind. Manager sessions are pinned to Claude Code; all other
-    /// kinds prefer an `agentsByKind` override and fall back to
-    /// `defaultAgentKind` when no override is set (CROW-421).
+    /// given kind. Prefers an `agentsByKind` override and falls back to
+    /// `defaultAgentKind` when no override is set (CROW-421, CROW-433).
     public func agentKind(for sessionKind: SessionKind) -> AgentKind {
-        if sessionKind == .manager { return .claudeCode }
         return agentsByKind[sessionKind.rawValue] ?? defaultAgentKind
     }
 
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -50,8 +50,8 @@ public struct AppConfig: Codable, Sendable, Equatable {
     public var defaultAgentKind: AgentKind
     /// Per-action-type overrides. When a key is present, sessions of that
     /// kind are created with the mapped agent; when absent, they fall back
-    /// to `defaultAgentKind`. Manager sessions are pinned to `.claudeCode`
-    /// and intentionally not honored by this map (CROW-421).
+    /// to `defaultAgentKind`. Honored for every `SessionKind`, including
+    /// `.manager` (CROW-433 — Manager was previously pinned to Claude Code).
     ///
     /// Keyed by `SessionKind.rawValue` (string) rather than `SessionKind`
     /// directly so JSON serializes as an object literal like
@@ -123,11 +123,9 @@ public struct AppConfig: Codable, Sendable, Equatable {
     }
 
     /// Resolve the agent that should drive a newly-created session of the
-    /// given kind. Manager sessions are always Claude Code (pinned by spec).
-    /// All other kinds prefer an explicit `agentsByKind` override, falling
-    /// back to `defaultAgentKind` (CROW-421).
+    /// given kind. Prefers an explicit `agentsByKind` override, falling
+    /// back to `defaultAgentKind` (CROW-421, CROW-433).
     public func agentKind(for sessionKind: SessionKind) -> AgentKind {
-        if sessionKind == .manager { return .claudeCode }
         return agentsByKind[sessionKind.rawValue] ?? defaultAgentKind
     }
 }
--- a/Packages/CrowCore/Tests/CrowCoreTests/AppConfigAgentKindTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AppConfigAgentKindTests.swift
@@ -91,9 +91,19 @@ import Testing
     #expect(config.agentKind(for: .job) == .claudeCode)
 }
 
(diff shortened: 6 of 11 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Solid, well-scoped change. The pin-to-Claude removal in both `AppState.agentKind(for:)` and `AppConfig.agentKind(for:)` is clean, the resolver tests are updated correctly, and the `managerCommand(for:)` dispatch through `AgentRegistry` with a default-agent fallback is sensible. One behavioral wart for non-Claude Managers keeps this from a clean approve.

### Security Review
**Strengths:**
- No new input-handling, auth, or secret surfaces. Binary resolution reuses the existing `findBinary()` candidate lists; no shell interpolation of untrusted data.
- `defaultAgent` fallback in `managerCommand(for:)` (`SessionService.swift:686-704`) fails safe to a known agent rather than crashing or emitting an empty command.

**Concerns:**
- None.

### Code Quality

**Yellow — Remote-control bookkeeping/badge is a false positive for a Cursor Manager.**
`createManagerTerminal` (`Sources/Crow/App/SessionService.swift:741-743`) and the hydrate rebuild (`SessionService.swift:126-128`) insert the Manager terminal into `remoteControlActiveTerminals` whenever `appState.remoteControlEnabled` is set — unconditionally, without consulting the resolved agent or whether its `managerLaunchCommand` actually emitted `--rc`. But `CursorAgent.managerLaunchCommand` deliberately omits `--rc`/`--name` (per its own comment), while `CursorAgent.supportsRemoteControl == true`. Net effect when a user runs a Cursor Manager with remote control globally enabled:
- `RemoteControlBadge` renders on the Manager session (`SessionListView.swift:619` gates on `supportsRemoteControl == true && isRemoteControlActive`), falsely claiming "this session can be driven from claude.ai."
- `renameSession` → `syncRemoteControlName` (`SessionService.swift:1380-1388`) sends `/rename <name>\n` into the Cursor TUI, which has no `/rename` command — injecting a spurious line into the agent.

Codex is safe here only incidentally (`supportsRemoteControl == false` hides the badge). Suggest gating the `remoteControlActiveTerminals` insert on the Manager command actually carrying remote control (i.e. the Claude path / `--rc` present), rather than on the global toggle alone.

**Green — trailing-newline convention is inconsistent across agents.**
`ClaudeCodeAgent.managerLaunchCommand` returns no trailing `\n`, while the default protocol extension (`CodingAgent.swift:104`), `CursorAgent`, and `OpenAICodexAgent` return `"...\n"`. `TmuxBackend.registerTerminal` already appends `+ \"\n\"` (`TmuxBackend.swift:234-235`), so the Cursor/Codex commands carry a redundant embedded LF that gets pasted as literal text before the submitting Enter. It works at a fresh shell, but dropping the trailing `\n` from the agent commands (letting the backend own the Enter) would give one convention.

**Green — misleading test comment / shared-singleton isolation.**
`Tests/CrowTests/ManagerMigrationTests.swift:57` says "Use a fresh registry to avoid bleed across tests," but the code registers into the global `AgentRegistry.shared` singleton with no reset. Under `--parallel`, registrations leak across tests and the `defaultAgent` fallback in `managerCommand(for:)` can mask a missing registration. Tests pass today because each registers what it needs; the comment is just inaccurate and the shared mutable global is a latent hazard.

**Green — agent switch reflects on respawn, not strictly "next app launch."**
`hydrateState` rebuilds the Manager command from the *persisted* `agentKind` (`SessionService.swift:118`), and `ensureManagerSession` updates `agentKind` afterwards without re-deriving the live terminal command (intentionally leaving the running tmux window alone). The change is picked up on the next Manager respawn. The SettingsView copy ("take effect on next app launch") slightly oversells immediacy — consider "next Manager respawn." Not a bug.

### Tests
- `swift test --package-path Packages/CrowCore --arch arm64` → 218 tests pass locally, including the new `appConfigAgentResolverFallsBackToDefaultForManager` and `appConfigAgentResolverHonorsExplicitManagerOverride`.
- Root `ManagerMigrationTests` (incl. new `managerCommandDispatchesByAgentKind`) could not be exercised in this review checkout: the full `swift build` fails on a missing `Frameworks/GhosttyKit.xcframework` binary artifact (environmental, not a PR defect). PR reports these green in CI.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 3 Green] findings. The Yellow (false remote-control badge + spurious `/rename` injection for a Cursor Manager) is the one worth addressing before merge.

---

dhilgaertner (changes requested):
## Code & Security Review

Solid, well-scoped change. The `managerLaunchCommand` abstraction on `CodingAgent` is clean, the default-extension/per-agent override split is sensible, and the RC bookkeeping fix (gating on what the command actually emitted rather than the global toggle) is exactly right for a Cursor Manager that reports `supportsRemoteControl` but omits `--rc`. Tests are meaningful. Two should-fix items below.

### Critical Issues
None.

### Security Review
**Strengths:**
- No new untrusted-input surface. Session names flowing into the command are POSIX single-quote–escaped via `ClaudeLaunchArgs.shellQuote`, and agent binaries resolve from a fixed candidate path list (`claudeBinaryCandidates`), so no injection or PATH-hijack vector is introduced.
- No secrets, auth, or crypto touched. The `--rc` detection uses the leading-space form (`" --rc"`), which matches `argsSuffix`'s output and avoids false positives from `--permission-mode auto`.

**Concerns:**
- None.

### Code Quality

**Yellow — stale comment now contradicts the feature** (`Sources/Crow/App/AppDelegate.swift:1157-1159`)
The `new-session` RPC handler still reads:
> `// Manager sessions get their own Claude-Code terminal in the devRoot, mirroring the primary Manager. Manager is pinned to Claude Code, so agent_kind is ignored for manager kind.`
`createManagerSession` now resolves `appState.agentKind(for: .manager)` (`SessionService.swift:764`), so Manager is no longer pinned to Claude Code — this is the exact statement CROW-433 set out to remove, and it was updated everywhere else (`AppState`, `AppConfig`, `ensureManagerSession`) but missed here. The "`agent_kind` is ignored" half is still accurate (the requested kind isn't threaded through for manager kind); please reword to reflect that secondary Managers now follow the configured Manager agent rather than being pinned.

**Yellow — agentKind reconciliation runs after the hydrate command rebuild** (`SessionService.swift:108-140` vs `577-601`, ordering set in `AppDelegate.swift:465` then `494`)
`hydrateState` rebuilds the Manager terminal command via `managerCommand(for: session)`, which keys on the **persisted per-session** `session.agentKind`. The reconciliation that picks up a changed setting (`ensureManagerSession`, line 594) runs *after* `hydrateState`. In the normal relaunch case this is masked because the live tmux window is adopted and the command string is never re-executed. But on the reboot / dead-tmux path, `rebuildAllSurfaces(forceRegister:false)` registers a fresh window and pastes the just-rebuilt command — which still reflects the *old* agent — so a setting change actually takes effect only on the *second* reboot, contradicting the UI copy "take effect on next Manager respawn." The explicit `restartManager` path is correct (reconciliation runs before `createManagerTerminal`). Suggest having the hydrate Manager branch resolve `appState.agentKind(for: .manager)` directly (or reconcile `session.agentKind` before the rebuild) so all respawn paths agree.

### Tests
- `managerCommandDispatchesByAgentKind` correctly asserts Cursor dispatch produces `agent` and not `claude`. Note the `AgentRegistry.shared` process-wide-singleton caveat is already documented in the test — good.
- Could not execute the root `CrowTests` suite locally: the full app build requires `Frameworks/GhosttyKit.xcframework`, which has no binary artifact in this checkout (environment limitation, not a PR defect). All standalone packages pass: CrowCore 218, CrowClaude 37, CrowCursor 27, CrowCodex 27.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 0 Green] findings. Both are low-risk and can land in one round trip; the stale comment is a quick wording fix and the ordering issue is a small reshuffle to match the documented respawn behavior.

dhilgaertner (approved):
## Code & Security Review

Round 3 — re-review after the round-2 fixes. This is a clean, well-scoped change and every finding from the prior two rounds has been addressed. Approving.

### Verification of prior findings

**Round 1 Yellow — RC bookkeeping false positive for a Cursor Manager → fixed.**
Both insert sites now gate on the command actually carrying `--rc` rather than the global toggle: hydrate rebuild computes `rebuiltCarriesRC = rebuiltCommand.contains(" --rc")` and removes the terminal from `remoteControlActiveTerminals` when absent (`SessionService.swift:148-161`); `createManagerTerminal` gates the insert on `command.contains(" --rc")` (`SessionService.swift:766-770`). A Cursor Manager (reports `supportsRemoteControl` but omits `--rc`) no longer gets a false RC badge or a spurious `/rename` injection. The leading-space form correctly avoids a `--permission-mode auto` false positive.

**Round 1 Greens → addressed.** Trailing-newline convention is now uniform — the default extension (`CodingAgent.swift:105`), Claude, Cursor, and Codex all return the bare command and let the terminal backend own the submitting Enter. The `ManagerMigrationTests` comment now accurately describes the `AgentRegistry.shared` process-wide-singleton behavior. SettingsView copy reads "next Manager respawn."

**Round 2 Yellow 1 — stale `AppDelegate` comment → fixed.**
`AppDelegate.swift:1157-1162` now states the Manager agent is resolved from `appState.agentKind(for: .manager)` inside `createManagerSession` and that only the request's `agent_kind` param is ignored — no longer claims Manager is pinned to Claude Code.

**Round 2 Yellow 2 — reconciliation ordering on the dead-tmux path → fixed.**
The hydrate Manager branch now reconciles `session.agentKind` to `appState.agentKind(for: .manager)` *before* the command rebuild (`SessionService.swift:127-140`), and config is mirrored onto `appState` (`AppDelegate.swift:460-461`) before `hydrateState()` (`AppDelegate.swift:465`) — confirmed by the explicit ordering comment at `AppDelegate.swift:452`. So the reboot/`rebuildAllSurfaces` path now pastes a command reflecting the *current* agent on the first respawn. The `restartManager` path remains correct (reconciles via `ensureManagerSession` before `createManagerTerminal`).

### Security Review
**Strengths:**
- No new untrusted-input surface. Session names are POSIX single-quote escaped via `ClaudeLaunchArgs.shellQuote`; agent binaries resolve from fixed candidate path lists — no injection or PATH-hijack vector.
- `managerCommand(for:)` fails safe: unknown/unregistered agent kind falls back to `defaultAgent`, then to the legacy Claude command, rather than emitting an empty or crashing command (`SessionService.swift:744-762`).
- No secrets, auth, or crypto touched.

**Concerns:** None.

### Code Quality
- The `managerLaunchCommand` abstraction on `CodingAgent` with a sensible default extension and per-agent overrides is clean and well-documented.
- Branch scoping verified: the Manager rebuild is correctly the `else` of `if !session.isManager` (`SessionService.swift:85` / `107`); Manager terminal commands stay non-nil, so the `command != nil` rebuild guard behaves as intended.
- `managerCommand(sessionName:)` is now a test-only legacy shim with no production call sites — acceptable and documented.

### Tests
- `swift test --package-path Packages/CrowCore --arch arm64` → 220 tests pass locally, including `appConfigAgentResolverFallsBackToDefaultForManager` and `appConfigAgentResolverHonorsExplicitManagerOverride`.
- The root `CrowTests` suite (incl. `managerCommandDispatchesByAgentKind`) could not be exercised in this checkout: the full app build requires `Frameworks/GhosttyKit.xcframework`, which has no binary artifact here (environmental, not a PR defect). PR reports these green in CI.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 0 Green] findings. All prior-round items resolved; no new issues.

dhilgaertner (approved):
## Code & Security Review

Third review round on CROW-433 (un-pinning the Manager tab from Claude Code). The earlier rounds' fixes — gating RC bookkeeping on the *emitted* `--rc` and the hydrate-time agent reconciliation — are both present and correct. Clean, well-commented, and well-tested change.

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface. `managerLaunchCommand` reuses the existing `ClaudeLaunchArgs.shellQuote` POSIX single-quote escaping for the `--name` label, so the session name can't break out of the shell command.
- Binary resolution stays on the fixed `claudeBinaryCandidates` allowlist (`~/.local/bin`, `/usr/local/bin`, `/opt/homebrew/bin`) — no $PATH lookup, no injection vector.
- RC `--rc` detection uses a " --rc" substring match that the shell-quoting reliably keeps free of false positives from crafted session names.

**Concerns:** None.

### Code Quality
Verified:
- **Startup ordering is correct** — `AgentRegistry` is populated and `appState.agentsByKind`/`defaultAgentKind` are applied from config (AppDelegate.swift:460-461) *before* `hydrateState` (:465) and `ensureManagerSession` (:494), so the hydrate reconciliation at SessionService.swift:127 resolves the live configured kind rather than a stale one.
- **Binary discovery parity** — `ClaudeCodeAgent.claudeBinaryCandidates` is identical to the legacy `SessionService.findClaudeBinary()` list, so routing the Manager command through the agent introduces no path regression.
- **Legacy shim is test-only** — `managerCommand(sessionName:)` has no production call sites; all real paths use `managerCommand(for:)`.
- **RC removal symmetry** — the hydrate loop now both inserts and removes from `remoteControlActiveTerminals` based on the rebuilt command, correctly clearing a stale RC badge for a Cursor/Codex Manager.
- `migrateLegacyManagerKind` only touches `kind`, not `agentKind`, so it doesn't fight the new reconciliation.

Green considerations (non-blocking, no action required to merge):
- **Broadened hydrate guard (SessionService.swift:150).** The condition relaxed from `cmd.contains("claude")` to `command != nil` so non-Claude Manager commands refresh. A Manager session realistically has exactly one agent terminal, and managed terminals persist with `command: nil` (deferred launch), so this is safe in practice — but an *unmanaged* terminal added to the Manager session with an explicit `--command` would now have its command overwritten by the agent command on next hydrate. Extremely rare; flagging only for awareness.
- **Fallback-to-default mismatch (SessionService.swift, `managerCommand(for:)`).** If the configured Manager kind isn't registered (e.g. Cursor selected, then the `agent` binary uninstalled), the command falls back to `defaultAgent` (Claude) while `session.agentKind` still reports `.cursor`. Defensive and reasonable; the resulting RC gating stays self-consistent because it keys off the emitted command.
- **Upgrade-safety claim is slightly narrower than stated.** "Existing installs see no behavior change" holds only when `defaultAgentKind` is Claude (the default). A user who had set `defaultAgentKind` to Codex/Cursor pre-433 will see their Manager switch to that agent on next launch, since Manager now inherits the default like every other kind. This is the intended CROW-433 semantics — just worth a line in release notes.

### Testing
- `AppConfigAgentKindTests` (CrowCore) pass locally, including the new `appConfigAgentResolverFallsBackToDefaultForManager` and `appConfigAgentResolverHonorsExplicitManagerOverride`.
- The root package (with `ManagerMigrationTests`/`managerCommandDispatchesByAgentKind`) could not be built in this environment due to a missing `Frameworks/GhosttyKit.xcframework` binary artifact — a local setup gap, unrelated to this PR. Relying on the author's reported run (171 root + 218 CrowCore tests green); the new dispatch test reads as sound.

### Summary Table

---

PR #441, opened 2026-06-04
Surface missing review/job prompt files instead of silently launching empty

### LINKED ISSUE #439: Review launch fails: .crow-review-prompt.md missing when terminal runs agent command
**Observed:** Launching a review session, the terminal got the command but the prompt file it tries to read isn't there.

Terminal output:
```
:50 › /Users/danny/.local/bin/agent "$(cat /Users/danny/Projects/devroot/crow-reviews/corveil-pr-1179/.crow-review-prompt.md)"
cat: /Users/danny/Projects/devroot/crow-reviews/corveil-pr-1179/.crow-review-prompt.md: No such file or directory
```

The launcher built the command shape (run `agent` with the prompt-file contents inlined via `$(cat ...)`), but the file at that path doesn't exist when `cat` evaluates. Net result: `agent` is called with an empty string and the review never starts.

**Suspected causes (one or more):**
1. **Race** — Crow `crow send`s the `agent "$(cat ...)"` line to the terminal before the prompt-write step has finished (or before the worktree path is fully set up). The launcher should write the file, fsync, then send the terminal command — not interleave them.
2. **Wrong path** — Crow is writing the prompt to a different path than the command reads from (e.g. tempdir vs worktree dir, repo-root vs worktree subdir).
3. **Worktree not created** — `crow-reviews/corveil-pr-1179` may not exist yet when the terminal command fires. (`crow add-worktree` returning before the dir is visible / writable, or the dir creation silently failed.)
4. **Cleanup race** — with `cleanup.retentionHours: 1`, a stale review dir could get scrubbed between worktree creation and command execution if the launch is slow. Unlikely for a fresh launch, but worth checking the cleanup runs against age-of-last-touch, not age-of-creation.

**Failure mode is silent at the app level:** `agent` runs with an empty prompt, so no review happens, but Crow doesn't notice — the session sits there looking active. Should detect the empty-prompt / cat-failed case and mark the session errored with the missing path surfaced.

**Acceptance:**
- Make the prompt-file-write happen *before* the terminal command is sent. (Ideally pass the prompt via stdin or `--prompt-file` argument so the `$(cat ...)` shell-substitution race goes away entirely.)
- On launch, fail loudly if the worktree dir or prompt file isn't where the launcher expects.
- If the empty-prompt issue can be detected from the launcher (e.g. the agent CLI returns immediately with no work), surface it in the session card.

**Notes:**
- This was a Cursor-style review launch (the agent path is `/Users/danny/.local/bin/agent`, not `claude` directly). Likely related to the broader Cursor integration thread:
  - #424 (Cursor coding silently no-ops)
  - #428 (Cursor cards don't update color)
  - #431 (Cursor reviews rejected) — wait, this contradicts #431; either #431's gating was bypassed or this was a Claude review through the wrapper. Worth investigating which agent kind was actually configured here.
- Repro: corveil PR 1179, launched via auto-review.

dgershman:
## Summary
- Reviews and jobs inline their initial prompt via `agent "$(cat .crow-*-prompt.md)"`. When that file wasn't on disk (failed `gh repo clone`, write error, etc.), the shell substitution yielded an empty string and the agent silently idled. Crow showed the session as active; nothing actually ran.
- Stop swallowing the upstream failures (`prepareReviewClone` clone + prompt-write, `createJobSession` prompt-write). Surface the failure so no session/terminal row is created with a missing prompt.
- Add a preflight check in `launchAgent` so even if the prompt file somehow goes missing later, we echo a clear "⚠️ Crow: review/job prompt missing at \<path\>" line into the terminal and refuse to dispatch — instead of running `agent` with an empty argument.
- Factored `SessionService.initialPromptFileName(for:)` so the launcher's validator and any future caller share one mapping; 3 unit tests for the helper.

Deliberately not changing the `$(cat ...)` shape to stdin/`--prompt-file` — the ordering is already correct (prompt write happens-before the terminal send via `await`), so the real bug was silent write failure, not a race.

Closes #439

## Test plan
- [x] `arch -arm64 swift test` — all 178 root-package tests pass, including 3 new tests for `initialPromptFileName(for:)`
- [ ] Manual repro: trigger auto-review against a PR whose repo Crow cannot clone (e.g. revoke `gh auth` briefly). Before: terminal opens with `agent "$(cat ...)"`, `cat` errors, agent idles. After: session is never created; the clone failure is logged.
- [ ] Manual repro: kick off a normal review, delete `.crow-review-prompt.md` before tmux fires `.shellReady`. After: terminal shows `⚠️ Crow: review prompt missing at <path>` and the agent CLI is not invoked.
- [ ] Happy path regression: kick off a normal review against a clonable PR — worktree, prompt file, agent invocation all proceed as today.

### DIFF
--- a/Sources/Crow/App/SessionService.swift
+++ b/Sources/Crow/App/SessionService.swift
@@ -534,6 +534,26 @@ final class SessionService {
             appState.terminalReadiness[terminalID] = .agentLaunched
             return
         }
+        // Preflight (CROW-439): review/job sessions inline their initial prompt
+        // via `$(cat .crow-*-prompt.md)`. If that file isn't on disk when the
+        // shell substitution runs, the agent launches with an empty string and
+        // silently idles. Refuse to dispatch and surface the missing path
+        // instead so the user sees why nothing happened.
+        if reviewPromptJustDispatched,
+           let initialPromptFile = Self.initialPromptFileName(for: session.kind) {
+            let promptPath = (worktree.worktreePath as NSString)
+                .appendingPathComponent(initialPromptFile)
+            if !FileManager.default.fileExists(atPath: promptPath) {
+                NSLog("[SessionService] launchAgent: initial prompt missing at \(promptPath) for session \(sessionID.uuidString); refusing to dispatch")
+                if let routedTerminal = appState.terminals[sessionID]?.first(where: { $0.id == terminalID }) {
+                    let msg = "echo '⚠️  Crow: \(session.kind.rawValue) prompt missing at \(promptPath); not launching \(agent.displayName).'\n"
+                    TerminalRouter.send(routedTerminal, text: msg)
+                }
+                appState.terminalReadiness[terminalID] = .agentLaunched
+                return
+            }
+        }
+
         // Route through TerminalRouter so tmux-backed terminals get the text
         // via tmux send-keys.
         if let routedTerminal = appState.terminals[sessionID]?.first(where: { $0.id == terminalID }) {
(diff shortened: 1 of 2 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Targeted, well-tested fix for the silent empty-prompt launch (CROW-439). Verified the new `initialPromptFileName(for:)` mapping stays in sync with both agent implementations and that the failure paths are handled by their callers. One readability nit introduced by the diff keeps this from a clean approve.

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface. The preflight `echo` warning (`SessionService.swift:549`) follows the same single-quote-interpolation pattern as the pre-existing #424 path (`SessionService.swift:528`); paths/display-names with embedded single quotes could theoretically break that one shell line, but this is unchanged from existing behavior and only affects a diagnostic echo.
- The `$(cat <path>)` substitution shape is intentionally unchanged (per PR description) — the real bug was a silent write failure, not a race, and the prompt write happens-before the terminal send via `await`.

**Concerns:**
- None.

### Code Quality
**Strengths:**
- Clone failure now `throw`s and is caught at `SessionService.swift:1438-1441` → `createReviewSession` returns `nil`, so no session/terminal row is created with a missing prompt. Defense-in-depth directory check (`1582-1589`) and post-write existence guard (`1606-1612`) are sensible.
- Job path symmetric: write failure / missing-file both `return nil` (`1717-1724`), handled by `JobScheduler`.
- `launchAgent` preflight (`537-555`) builds the path from the same `worktree.worktreePath` + same filename mapping the agent's `autoLaunchCommand` uses, gated by the same `!reviewPromptDispatched` condition — correct and consistent. Parks readiness to avoid retry loops.
- `initialPromptFileName(for:)` switch is exhaustive over all four `SessionKind` cases and matches `ClaudeCodeAgent.swift:82-84` and `CursorAgent.swift:86-87` exactly. Good test coverage for the helper.

**Issues:**
- 🟡 **Misattributed doc comment** (`SessionService.swift:1857-1872`): the new `initialPromptFileName` was inserted between `buildReviewPrompt`'s doc-comment block and `buildReviewPrompt` itself, with no separating blank line. Swift now attaches the entire combined `///` block to `initialPromptFileName` — including the paragraph explaining why `buildReviewPrompt` is `internal (not private)` so `SessionServiceReviewPromptTests` can assert *its* branch dispatch, and the note about `cursorReviewPrompt`/`Scaffolder.bundledReviewSkill()`. That text describes `buildReviewPrompt`, not the new helper, and `buildReviewPrompt` is now left undocumented. Move the helper below `buildReviewPrompt`, or relocate the orphaned comment back onto `buildReviewPrompt`.

### Verification Note
Could not run `swift build`/`swift test` in this review environment — the build fails on a missing `GhosttyKit.xcframework` binary artifact (local setup, unrelated to this diff). Findings are from static analysis; relying on the PR's reported `swift test` pass (178 tests + 3 new).

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 0 Green] findings. The logic is correct and ships the fix; only the misplaced doc comment needs a touch-up so it lands in this round trip.

---

dhilgaertner (approved):
## Code & Security Review

Tightly-scoped, well-reasoned fix for CROW-439 (review/job sessions silently idling when the inlined `$(cat .crow-*-prompt.md)` prompt file is missing). I traced every integration point and the change is correct.

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface. The change converts silent `try?` swallows into explicit error propagation and a preflight existence check — strictly defensive.
- Prompt paths are Crow-controlled (`worktreePath` + fixed filename), so the single-quoted `echo` warning carries no untrusted interpolation. Consistent with the existing #424 echo pattern at `SessionService.swift:528`.

**Concerns:**
- None.

### Code Quality — verified
- **Helper mapping is in sync** with both inline encoders: `initialPromptFileName(for:)` (`SessionService.swift:1855`) matches `ClaudeCodeAgent.autoLaunchCommand` (`ClaudeCodeAgent.swift:82-84`) and `CursorAgent.autoLaunchCommand` (`CursorAgent.swift:85-87`) exactly. The 3 new unit tests lock this mapping.
- **No false positives / no misfire:** Codex only supports `.work` (`OpenAICodexAgent.swift:57`) and returns nil for review/job, so it never reaches the preflight. The preflight is gated on `reviewPromptJustDispatched` (first launch only) and uses the same `!reviewPromptDispatched` flag the agents use, so the command shape and the check stay synchronized.
- **No partial session on failure:** `prepareReviewClone`'s throw is caught in `createReviewSession` (`SessionService.swift:1438-1441`) before any session/worktree/terminal row is appended. `runShellAsync` genuinely throws on non-zero exit (`SessionService.swift:976`), so the new clone `do/catch` fires as intended. The added `isDir` defense-in-depth check is sensible.
- **Path consistency:** prompts are written to exactly the path the preflight checks for both review (`clonePath`, used as `worktreePath`) and job paths. The write always happens-before launch (synchronous in session creation), so no timing false-positive.
- The PR correctly declines to re-shape `$(cat ...)` → stdin; the root cause was silent write failure, not a race — and the TOCTOU window is acknowledged with the preflight positioned as a best-effort net.

- `createJobSession` returns nil at `SessionService.swift:1719` *after* `createWorktree` already created a worktree on disk (line 1702), leaving an orphaned worktree dir on a prompt-write failure. Rare path, consistent with the function's best-effort nil-return contract, but a `removeItem` on the way out would tidy it.
- The `createReviewSession` clone-failure path surfaces only via `NSLog` (UI shows nothing), whereas the `launchAgent` preflight echoes into the terminal. Consistent with the sibling nil-returns in `createReviewSession`, but the two failure modes differ in user visibility.

### Note
Could not run the build/tests locally — this checkout is missing the vendored `Frameworks/GhosttyKit.xcframework` binary (an environment issue, unrelated to the two files changed). The PR's documented `arch -arm64 swift test` run reports all 178 tests passing, including the 3 new ones.

### Summary Table

---

PR #444, opened 2026-06-04
Attribute Crow footers to the session's agent kind

### LINKED ISSUE #443: Attribution footers hardcoded to "Claude Code" — should reflect the actual agent (Cursor, Codex, etc.)
**Observed:** Every artifact a Crow skill creates (issues via `crow-create-ticket`, PRs via the workspace flow, review summaries, etc.) ends with a hardcoded attribution line:

```
```

This text appears in skill instructions like `.claude/skills/crow-create-ticket/SKILL.md` with explicit "do not modify the link text" guardrails. So even when the user is running Cursor or Codex as their agent, every Crow-created GitHub/GitLab artifact still claims it was made via Claude Code.

**Ask:** Make the agent name in the footer reflect whichever agent actually produced the artifact.

Examples:
- Claude Code run → `[🐦‍⬛ Created with Crow via Claude Code](…)` (today's behavior, unchanged)
- Cursor run → `[🐦‍⬛ Created with Crow via Cursor](…)`
- Codex run → `[🐦‍⬛ Created with Crow via Codex](…)`

Same applies to other artifact paths:
- Review summary comments — should say "Reviewed by Crow via {agent}"
- PR descriptions / commit trailers (if Crow injects any)
- Any "Generated by …" text in Crow-emitted markdown

**Suggested mechanism:**
- Crow already knows the agent kind for each session (it's what it used to launch). Inject it into the skill's environment (e.g. `CROW_AGENT_KIND=cursor`) when spawning the terminal that runs the skill.
- Update skill SKILL.md files to interpolate that env var into the footer template (one shared snippet, not per-skill duplication if possible).
- Drop the "do not modify the link text" guardrail and replace it with "use the agent-name from `CROW_AGENT_KIND`, fall back to `Claude Code` if unset" — preserves backward behavior for installs without the change.

**Acceptance:**
- An issue/PR created from a Cursor-driven session ends with `Created with Crow via Cursor`, not `Claude Code`.
- Same applies to review summary comments and any other Crow-attributable artifact.
- Falls back to `Claude Code` if the agent kind isn't surfaced (no-op for old setups).
- Single source of truth for the footer template — not duplicated across every skill.

**Related (cluster of agent-kind plumbing issues):**
- #421 (per-action agent setting) — once selectable, this becomes more visible.
- #424 (Cursor coding silently no-ops), #428 (Cursor cards don't update), #431 (Cursor reviews rejected), #427 (tab labels always "Claude Code"), #433 (Manager always Claude), #440 (skill hardcodes claude binary) — all variations on "the agent kind isn't propagated end-to-end." Worth treating these seven as a single planning epic.

dgershman:
Closes #443

## Summary

- Inject `CROW_AGENT_KIND` and `CROW_AGENT_DISPLAY_NAME` into every managed terminal at tmux window creation so skills and agents know which agent produced an artifact.
- Centralize footer formatting in `CrowAttribution` (Swift helpers, `expandSkillBody`, shared `FOOTER.md` scaffolded to `.claude/skills/crow-attribution/`).
- Update `crow-create-ticket` and `crow-review-pr` skills to use `$CROW_AGENT_DISPLAY_NAME` (fall back to `Claude Code` when unset); Cursor review prompts expand `{{CROW_AGENT_DISPLAY_NAME}}` inline.

## Test plan

- [x] `swift test --filter CrowAttribution` in `Packages/CrowCore`
- [ ] Create an issue from a Cursor session — footer reads `via Cursor`
- [ ] Post a PR review from a Cursor review session — footer reads `via Cursor`
- [ ] Verify Claude Code sessions still default to `via Claude Code` when env is absent

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/CrowAttribution.swift
+++ b/Packages/CrowCore/Sources/CrowCore/CrowAttribution.swift
@@ -16,6 +16,10 @@ public enum CrowAttribution {
     /// Placeholder in bundled SKILL bodies expanded by `expandSkillBody(_:agentKind:)`.
     public static let skillAgentPlaceholder = "{{CROW_AGENT_DISPLAY_NAME}}"
 
+    /// Bash parameter expansion for footers in skill `gh`/`glab` `--body` arguments.
+    public static let shellAgentDisplayNameExpression =
+        "${\(agentDisplayNameEnvironmentKey):-\(defaultAgentDisplayName)}"
+
     /// Intentional second source of truth: `AgentRegistry` is not populated in
     /// CrowCore-only unit tests, so known kinds resolve here before the registry.
     private static let knownDisplayNames: [String: String] = [
@@ -68,10 +72,11 @@ public enum CrowAttribution {
     /// Default ticket footer (Claude Code) — backward-compatible alias.
     public static var ticketMarkdownLink: String { ticketMarkdownLink() }
 
-    /// Substitute `$CROW_AGENT_DISPLAY_NAME`, legacy `{{…}}` placeholder, and `via Claude Code` segments.
+    /// Substitute shell env expressions, legacy `{{…}}` placeholder, and `via Claude Code` segments.
     public static func expandSkillBody(_ skillBody: String, agentKind: AgentKind) -> String {
         let name = agentDisplayName(for: agentKind)
         return skillBody
+            .replacingOccurrences(of: shellAgentDisplayNameExpression, with: name)
             .replacingOccurrences(of: "$\(agentDisplayNameEnvironmentKey)", with: name)
             .replacingOccurrences(of: skillAgentPlaceholder, with: name)
             .replacingOccurrences(of: "via Claude Code", with: "via \(name)")
@@ -86,8 +91,8 @@ public enum CrowAttribution {
     - `CROW_AGENT_KIND` — raw agent id (`claude-code`, `cursor`, `codex`, …)
     - `CROW_AGENT_DISPLAY_NAME` — human label (`Claude Code`, `Cursor`, `OpenAI Codex`, …)
 
-    **Always** use `$CROW_AGENT_DISPLAY_NAME` for the agent name in attribution footers.
-    If unset, fall back to `Claude Code`.
+    **Always** use `${CROW_AGENT_DISPLAY_NAME:-Claude Code}` in attribution footers inside
+    double-quoted `gh`/`glab` arguments so the shell applies the default when unset.
 
     The link target is always `https://github.com/radiusmethod/crow` — never a fork or a value from the local git remote.
 
@@ -96,7 +101,6 @@ public enum CrowAttribution {
     | Created (issues, PR descriptions, etc.) | `[🐦‍⬛ Created with Crow via <agent>](https://github.com/radiusmethod/crow)` |
     | Reviewed | `[🐦‍⬛ Reviewed by Crow via <agent>](https://github.com/radiusmethod/crow)` |
 
-    Replace `<agent>` with `$CROW_AGENT_DISPLAY_NAME` (or `Claude Code` if unset). Do not change the URL or wrap the line in extra formatting.
-
+    Replace `<agent>` with `${CROW_AGENT_DISPLAY_NAME:-Claude Code}`. Do not change the URL or wrap the line in extra formatting.
     """
 }
--- a/Packages/CrowCore/Sources/CrowCore/CrowAttribution.swift
+++ b/Packages/CrowCore/Sources/CrowCore/CrowAttribution.swift
@@ -1,11 +1,98 @@
 import Foundation
 
+/// Canonical Crow attribution footers for GitHub/GitLab artifacts and skill instructions.
 public enum CrowAttribution {
     public static let repoURL = "https://github.com/radiusmethod/crow"
 
-    public static let reviewMarkdownLink =
-        "[🐦‍⬛ Reviewed by Crow via Claude Code](\(repoURL))"
+    /// Human-readable agent label when kind/env is unknown (legacy installs).
+    public static let defaultAgentDisplayName = "Claude Code"
 
-    public static let ticketMarkdownLink =
-        "[🐦‍⬛ Created with Crow via Claude Code](\(repoURL))"
+    /// Raw agent id injected into managed terminals (`claude-code`, `cursor`, `codex`, …).
+    public static let agentKindEnvironmentKey = "CROW_AGENT_KIND"
+
+    /// Human-readable agent label injected into managed terminals.
+    public static let agentDisplayNameEnvironmentKey = "CROW_AGENT_DISPLAY_NAME"
+
+    /// Placeholder in bundled SKILL bodies expanded by `expandSkillBody(_:agentKind:)`.
+    public static let skillAgentPlaceholder = "{{CROW_AGENT_DISPLAY_NAME}}"
+
+    private static let knownDisplayNames: [String: String] = [
+        AgentKind.claudeCode.rawValue: "Claude Code",
+        AgentKind.cursor.rawValue: "Cursor",
+        AgentKind.codex.rawValue: "OpenAI Codex",
+    ]
+
+    /// Resolve the display name for `agentKind`, falling back to `defaultAgentDisplayName`.
+    public static func agentDisplayName(for agentKind: AgentKind?) -> String {
+        guard let agentKind else { return defaultAgentDisplayName }
+        if let known = knownDisplayNames[agentKind.rawValue] { return known }
+        let registryName = agentKind.displayName
+        return registryName == agentKind.rawValue ? defaultAgentDisplayName : registryName
+    }
+
+    /// Read `CROW_AGENT_DISPLAY_NAME` or map `CROW_AGENT_KIND`; default to Claude Code.
+    public static func agentDisplayName(fromEnvironment environment: [String: String]?) -> String {
+        guard let environment else { return defaultAgentDisplayName }
+        if let explicit = environment[agentDisplayNameEnvironmentKey],
+           !explicit.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
+            return explicit
+        }
+        if let raw = environment[agentKindEnvironmentKey],
+           !raw.isEmpty {
+            return agentDisplayName(for: AgentKind(rawValue: raw))
+        }
+        return defaultAgentDisplayName
+    }
+
+    /// Environment entries to inject when spawning a managed terminal.
+    public static func environmentEntries(for agentKind: AgentKind) -> [String: String] {
+        [
+            agentKindEnvironmentKey: agentKind.rawValue,
+            agentDisplayNameEnvironmentKey: agentDisplayName(for: agentKind),
+        ]
+    }
+
+    public static func reviewMarkdownLink(agentDisplayName: String = defaultAgentDisplayName) -> String {
+        "[🐦‍⬛ Reviewed by Crow via \(agentDisplayName)](\(repoURL))"
+    }
+
+    public static func ticketMarkdownLink(agentDisplayName: String = defaultAgentDisplayName) -> String {
+        "[🐦‍⬛ Created with Crow via \(agentDisplayName)](\(repoURL))"
+    }
+
+    /// Default review footer (Claude Code) — backward-compatible alias.
+    public static var reviewMarkdownLink: String { reviewMarkdownLink() }
+
+    /// Default ticket footer (Claude Code) — backward-compatible alias.
+    public static var ticketMarkdownLink: String { ticketMarkdownLink() }
+
+    /// Substitute `{{CROW_AGENT_DISPLAY_NAME}}` and legacy `via Claude Code` segments.
+    public static func expandSkillBody(_ skillBody: String, agentKind: AgentKind) -> String {
+        let name = agentDisplayName(for: agentKind)
+        return skillBody
+            .replacingOccurrences(of: skillAgentPlaceholder, with: name)
+            .replacingOccurrences(of: "via Claude Code", with: "via \(name)")
+    }
+
+    /// Shared attribution rules copied into `.claude/skills/crow-attribution/FOOTER.md`.
+    public static let sharedFooterInstructions: String = """
+    # Crow Attribution Footers
+
+    Crow injects these environment variables into every managed terminal:
+
+    - `CROW_AGENT_KIND` — raw agent id (`claude-code`, `cursor`, `codex`, …)
+    - `CROW_AGENT_DISPLAY_NAME` — human label (`Claude Code`, `Cursor`, `OpenAI Codex`, …)
+
+    **Always** use `$CROW_AGENT_DISPLAY_NAME` for the agent name in attribution footers.
+    If unset, fall back to `Claude Code`.
+
+    The link target is always `https://github.com/radiusmethod/crow` — never a fork or a value from the local git remote.
+
+    | Artifact | Footer |
+    |----------|--------|
+    | Created (issues, PR descriptions, etc.) | `[🐦‍⬛ Created with Crow via <agent>](https://github.com/radiusmethod/crow)` |
+    | Reviewed | `[🐦‍⬛ Reviewed by Crow via <agent>](https://github.com/radiusmethod/crow)` |
+
+    Replace `<agent>` with `$CROW_AGENT_DISPLAY_NAME` (or `Claude Code` if unset). Do not change the URL or wrap the line in extra formatting.
+    """
 }
--- a/Packages/CrowCore/Sources/CrowCore/CrowAttribution.swift
+++ b/Packages/CrowCore/Sources/CrowCore/CrowAttribution.swift
@@ -16,6 +16,8 @@ public enum CrowAttribution {
     /// Placeholder in bundled SKILL bodies expanded by `expandSkillBody(_:agentKind:)`.
     public static let skillAgentPlaceholder = "{{CROW_AGENT_DISPLAY_NAME}}"
 
+    /// Intentional second source of truth: `AgentRegistry` is not populated in
+    /// CrowCore-only unit tests, so known kinds resolve here before the registry.
     private static let knownDisplayNames: [String: String] = [
         AgentKind.claudeCode.rawValue: "Claude Code",
         AgentKind.cursor.rawValue: "Cursor",
@@ -95,5 +97,6 @@ public enum CrowAttribution {
     | Reviewed | `[🐦‍⬛ Reviewed by Crow via <agent>](https://github.com/radiusmethod/crow)` |
 
     Replace `<agent>` with `$CROW_AGENT_DISPLAY_NAME` (or `Claude Code` if unset). Do not change the URL or wrap the line in extra formatting.
+
     """
 }
--- a/Packages/CrowCore/Sources/CrowCore/CrowAttribution.swift
+++ b/Packages/CrowCore/Sources/CrowCore/CrowAttribution.swift
@@ -66,10 +66,11 @@ public enum CrowAttribution {
     /// Default ticket footer (Claude Code) — backward-compatible alias.
     public static var ticketMarkdownLink: String { ticketMarkdownLink() }
 
-    /// Substitute `{{CROW_AGENT_DISPLAY_NAME}}` and legacy `via Claude Code` segments.
+    /// Substitute `$CROW_AGENT_DISPLAY_NAME`, legacy `{{…}}` placeholder, and `via Claude Code` segments.
     public static func expandSkillBody(_ skillBody: String, agentKind: AgentKind) -> String {
         let name = agentDisplayName(for: agentKind)
         return skillBody
+            .replacingOccurrences(of: "$\(agentDisplayNameEnvironmentKey)", with: name)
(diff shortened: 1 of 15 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Nicely scoped change: env-var injection at `tmux new-window -e`, centralized footer logic in `CrowAttribution`, and consistent use of `$CROW_AGENT_DISPLAY_NAME` in the create-ticket skill. CrowCore builds and all 15 `CrowAttribution` tests pass locally. (I could not build the full app target here — `Frameworks/GhosttyKit.xcframework` is not present in this checkout — so `AttributionSkillTests`/`SessionServiceReviewPromptTests` were verified by reading, not running.)

### Critical Issues
None.

### Security Review
**Strengths:**
- Env values are passed to tmux as discrete argv elements (`["-e", "\(k)=\(v)"]` in `TmuxController.swift:144`), not via a shell string — so display names containing spaces (`Claude Code`, `OpenAI Codex`) need no quoting and there's no command-injection surface.
- The footer URL stays hardcoded to `https://github.com/radiusmethod/crow`; no user/remote-derived value flows into the attribution line.
- Injected values come from a controlled `AgentKind` rawValue / registry, not arbitrary input.

**Concerns:** None.

### Code Quality

**Yellow — review skill is internally inconsistent about the footer placeholder.**
`skills/crow-review-pr/SKILL.md` (and its template) uses two different representations of the agent name:
- Format block (line 134): `[🐦‍⬛ Reviewed by Crow via {{CROW_AGENT_DISPLAY_NAME}}](…)`
- Step 5b (line 143/146): `$CROW_AGENT_DISPLAY_NAME` (shell env var)

The `{{…}}` placeholder is only expanded programmatically on the **Cursor** path (`SessionService.cursorReviewPrompt` → `CrowAttribution.expandSkillBody`). For a **Claude Code** review session the prompt is just `/crow-review-pr <URL>` (`SessionService.swift:1967-1969`) and the skill is read raw — so `{{CROW_AGENT_DISPLAY_NAME}}` is never substituted, and it is not a shell variable either. If the agent fills in the "Use this format" block verbatim, the posted GitHub review reads literally `via {{CROW_AGENT_DISPLAY_NAME}}`. Step 5b says the right thing, but the two contradict each other in the same skill.

The create-ticket skill got this right by using `$CROW_AGENT_DISPLAY_NAME` consistently in both the example and Step 4b. Suggest aligning the review format block the same way (use `$CROW_AGENT_DISPLAY_NAME`, which Cursor's shell also expands since the env var is injected for every agent), or at minimum annotating that `{{…}}` is auto-substituted only for Cursor and Claude Code must use the env var. Since this is the feature's primary deliverable (a correct public footer), worth closing in this round.

**Green (consider):**
- `CrowAttribution.agentDisplayName(fromEnvironment:)` is exercised only by tests — no production caller. Either wire it where the footer is emitted or drop it to avoid speculative API.
- `knownDisplayNames` hardcodes the same labels `AgentRegistry` exposes and is consulted *before* the registry. It's intentional (CrowCore tests register no agents), but it's a second source of truth that can drift from `AgentKind.displayName`. Consider registry-first with the map as fallback.
- No test pins `CrowAttribution.sharedFooterInstructions` (the scaffolder's fallback) to `skills/crow-attribution/FOOTER.md`. They're identical today but can silently diverge; the existing byte-identity tests cover the SKILL templates but not this constant ↔ file pair.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 3 Green] findings. The Yellow (review-skill footer inconsistency) is the only blocker; the Greens are optional polish.

---

dhilgaertner (changes requested):
## Code & Security Review

Solid, well-tested feature. Agent attribution is injected at the tmux layer, re-exported by the shell wrapper, centralized in `CrowAttribution`, and the Claude Code default is preserved everywhere for backward compatibility. `swift build` + all 15 `CrowAttributionTests` pass locally. Two should-fix items below keep this out of an approve.

> Note: the `Crow` app target could not be fully compiled in the review environment (missing prebuilt `Frameworks/GhosttyKit.xcframework` binary — infra, not this PR). `CrowCore` built clean and the prior commit 6150a26 already fixed the `Scaffolder` `import CrowCore` break.

### Critical Issues
None.

### Security Review
**Strengths:**
- No injection surface: env values are agent ids from a closed registry, written via `tmux new-window -e` (not interpolated into a shell string). The wrapper re-export is guarded with `[ -n ... ]`.
- Link target is hard-pinned to `https://github.com/radiusmethod/crow` in the Swift constant, the FOOTER, and both skill templates; the existing no-fork/no-`corveil` tests still hold.

**Concerns:** None.

### Code Quality

**Yellow — `Resources/crow-attribution-FOOTER.md.template` is dead / unreferenced.**
`Scaffolder.bundledAttributionFooter()` (`Sources/Crow/App/Scaffolder.swift:234`) is the only `bundled*` loader that skips the `Bundle.main.url(forResource:withExtension:)` step its siblings (`bundledReviewSkill`, `bundledCreateTicketSkill`, …) all have. It falls back straight from `loadFromRepo` to the `CrowAttribution.sharedFooterInstructions` Swift constant, so the newly-added `Resources/crow-attribution-FOOTER.md.template` is read by nothing — it is also not declared in `Package.swift` `resources:`, and a repo-wide grep finds zero references. Either wire up the `Bundle.main` lookup (matching the established pattern) or drop the template file; right now it is committed dead weight and a third copy that can silently drift.

**Yellow — mixed placeholder conventions in `skills/crow-review-pr/SKILL.md`.**
The primary "Use this format" example at line 134 uses `{{CROW_AGENT_DISPLAY_NAME}}`, while Step 5b (line 143) uses `$CROW_AGENT_DISPLAY_NAME`. Only the Cursor path runs `expandSkillBody` (which substitutes the `{{...}}` form); the Claude Code path delivers the raw SKILL via `/crow-review-pr <URL>`, and neither the slash-command engine nor `expandSkillBody` ever expands `{{...}}` for Claude Code. A literal `{{CROW_AGENT_DISPLAY_NAME}}` can therefore leak into a posted review footer — the exact failure this PR exists to prevent. `crow-create-ticket` avoids this by using `$CROW_AGENT_DISPLAY_NAME` consistently. Suggest using `$CROW_AGENT_DISPLAY_NAME` at line 134 too (and teaching `expandSkillBody` to also replace the `$` form for Cursor), or adding prose that `{{...}}` is a placeholder.

**Green — no drift guard between the Swift constant and the markdown footer.**
`CrowAttribution.sharedFooterInstructions` and `skills/crow-attribution/FOOTER.md` are byte-identical today (verified), but unlike the other skills (`liveSkillAndBundledTemplateAreByteIdentical`) nothing pins them together. Since the constant is the production fallback, a future edit to one could silently diverge from the other. Consider a byte-identity test.

**Green — partially-unused new API.**
`agentDisplayName(fromEnvironment:)` and the parameterized `reviewMarkdownLink(agentDisplayName:)` / `ticketMarkdownLink(agentDisplayName:)` overloads are exercised only by tests — no production caller (runtime footers come from the skill markdown, and injection uses `environmentEntries`/`expandSkillBody`). Fine as forward-looking surface; just confirm that is intentional rather than a half-wired path.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 2 Green] findings. Both Yellows are small and can land in one round trip.

---

dhilgaertner (changes requested):
## Code & Security Review

Solid, well-tested feature. I verified locally: `CrowCore` builds clean and `swift test --filter CrowAttribution` passes (16/16). The env-injection path is correct end-to-end — `TmuxBackend.registerTerminal(agentKind:)` → `ctrl.newWindow(env:)` → `tmux new-window -e KEY=VAL` → `crow-shell-wrapper.sh` re-export — and all three production `registerTerminal` call sites (AppDelegate + 2× SessionService) now pass `agentKind`, each in scope. (The full-app build only failed on a missing `Frameworks/GhosttyKit.xcframework`, which is a setup artifact absent from the review clone, not a regression.)

### Critical Issues
None.

### Security Review
**Strengths:**
- No new auth/injection surface. The env values are agent kind identifiers, not user input.
- The canonical link target `https://github.com/radiusmethod/crow` is preserved everywhere; tests assert no fork/derived URLs leak in.
- `agentDisplayName(fromEnvironment:)` trims and guards empty values before use.

**Concerns:** None.

### Code Quality

**Yellow — `Resources/crow-attribution-FOOTER.md.template` is dead (`Sources/Crow/App/Scaffolder.swift:234`).**
Every other `bundled*` method falls back through `Bundle.main.url(forResource:..., withExtension: \"template\")` before its inline default (see `bundledReviewSkill()`/`bundledCreateTicketSkill()`). `bundledAttributionFooter()` skips that step and goes straight from `loadFromRepo(...)` to the `CrowAttribution.sharedFooterInstructions` Swift constant. A repo-wide grep for `crow-attribution-FOOTER` finds zero references — the newly added template resource is never read by anyone. Either wire the `Bundle.main.url(...)` fallback for consistency with the other skills, or drop the template file. As-is it's a trap: a file that looks like the bundled source for every other skill but has no effect.

**Yellow — no drift guard for the footer (compounds the above).**
The other skills have `*AreByteIdentical` tests tying `skills/.../SKILL.md` to `Resources/...template`. There's no equivalent tying `skills/crow-attribution/FOOTER.md`, `Resources/crow-attribution-FOOTER.md.template`, and `CrowAttribution.sharedFooterInstructions` together. They're identical today, but since the production fallback is the Swift constant while the (unread) template is what a maintainer would naturally edit, the two can silently diverge. A byte-identical test would close this.

**Green — `knownDisplayNames` duplicates the registry (`Packages/CrowCore/Sources/CrowCore/CrowAttribution.swift:28`).**
\"Claude Code\"/\"Cursor\"/\"OpenAI Codex\" already match each agent's `CodingAgent.displayName`, and the map always wins for known kinds — so renaming an agent's display name in its impl would silently not update footers. The duplication is justified (the registry isn't populated in CrowCore-only unit tests), but a one-line comment noting the intentional second source of truth would help.

**Green — legacy already-running terminals** won't have the env injected (only newly created windows get `-e`). The skill instructions handle this via the explicit \"fall back to Claude Code if unset\" guidance, so it's covered.

### Summary Table

**Recommendation:** Request Changes — driven by 0 Red, 2 Yellow, 2 Green findings. Both Yellows are maintainability/consistency cleanups (no user-facing impact today); addressing them in this round avoids a follow-up.

---

dhilgaertner (changes requested):
## Code & Security Review

Solid, well-tested change. Footer formatting is centralized in `CrowAttribution`, the SKILL/template/Swift-constant trio is guarded against drift from three directions, and the agent-kind env vars are threaded cleanly through `TmuxBackend.registerTerminal`. The `CrowCore` test suite (`swift test --filter CrowAttribution`) passes locally — 17/17. One robustness gap below keeps this at Request Changes.

> Note: the full app target (`Tests/CrowTests`) can't be built locally — `Frameworks/GhosttyKit.xcframework` has no binary artifact in this checkout — so `AttributionSkillTests` / `SessionServiceReviewPromptTests` were verified by reproducing their assertions by hand (FOOTER.md ↔ template byte-identical; live SKILLs ↔ templates byte-identical; no stray `via Claude Code`), not by running them. CI should be the source of truth there.

### Security Review
**Strengths:**
- Env values reach tmux as discrete `argv` elements (`["-e", "KEY=VALUE"]` in `TmuxController`), so display names containing spaces (`Claude Code`, `OpenAI Codex`) are not shell-split — no injection/quoting hazard.
- Link target stays hard-coded to `https://github.com/radiusmethod/crow` everywhere; the shared FOOTER.md and skill copy both reiterate "never a fork or a value from the local git remote."
- No secrets, no auth surface, no new external input paths.

**Concerns:**
- None.

### Code Quality

**🟡 Yellow — unset `$CROW_AGENT_DISPLAY_NAME` shell-expands to empty, not to the `Claude Code` fallback**
The skills now embed the footer as a literal `$CROW_AGENT_DISPLAY_NAME` inside double-quoted `gh`/`glab` `--body` arguments (`skills/crow-create-ticket/SKILL.md:114`, `skills/crow-review-pr/SKILL.md:134`). When the var is set this works via shell expansion. When it is **unset**, bash expands `$CROW_AGENT_DISPLAY_NAME` to the empty string, producing `Created with Crow via ](...)` — i.e. a broken, agent-less footer — *unless* the LLM happens to apply the parenthetical "use `Claude Code` if unset" itself. The old code embedded a literal string and was unconditionally correct, so this is a narrow regression in footer robustness.

This isn't hypothetical: `CROW_AGENT_DISPLAY_NAME` is **unset in this very review session** (pre-existing terminals created before the injection landed don't carry it; the `SessionService` rehydrate paths also skip injection when the session row isn't found, since `agentKind` resolves to `nil`). It's also the one box the PR's own test plan leaves unchecked: *"Verify Claude Code sessions still default to `via Claude Code` when env is absent."*

Suggested fix — let the shell apply the default so correctness doesn't depend on LLM attentiveness:
```
[🐦‍⬛ Created with Crow via ${CROW_AGENT_DISPLAY_NAME:-Claude Code}](https://github.com/radiusmethod/crow)
```
(and the analogous `Reviewed by` line). Apply it in both live SKILLs + both `Resources/*.template` files so the byte-identity guards stay green. This makes the absent-env path deterministic regardless of whether the agent expands the var itself or defers to the shell.

### Consider (Green)
- **Two fallback mechanisms for the same goal.** Review prompts for Cursor are pre-expanded at build time (`cursorReviewPrompt` → `expandSkillBody`), while create-ticket relies purely on runtime shell/LLM expansion of the env var. Both yield correct output, but the asymmetry is a small future-maintenance trap — worth a one-line comment noting why review needs build-time expansion (no slash-command engine for Cursor) and create-ticket doesn't.
- **`skills/crow-attribution/` has a `FOOTER.md` but no `SKILL.md`.** It's a shared reference doc rather than an invokable skill; placing it under `skills/` is slightly unconventional. Harmless, just flagging in case a skill loader ever scans that directory expecting a manifest.
- **`CrowAttribution.agentDisplayName(for:)` intentionally falls back to `Claude Code`** for unknown/unregistered kinds, which is in mild tension with `AgentKind.displayName`'s CROW-427 contract ("the fallback must NOT silently be Claude Code"). The divergence is deliberate and commented; no action needed, just noting the two helpers now disagree on purpose.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 3 Green] findings. The single Yellow is a small, self-contained hardening (`${VAR:-Claude Code}`) that closes the exact unverified path in the PR's test plan; folding it in now lands it in the same round trip.

---

dhilgaertner (changes requested):
## Code & Security Review

Nicely structured change: centralizing footer formatting in `CrowAttribution`, injecting `CROW_AGENT_KIND`/`CROW_AGENT_DISPLAY_NAME` at `tmux new-window -e`, and the new byte-identity + Swift-constant drift guards are exactly the right safety net. One of those guards is currently tripped, though.

### Critical Issues

**🔴 `crow-create-ticket` skill and its bundled template have drifted — breaks CI and ships a broken footer**

`skills/crow-create-ticket/SKILL.md` uses the defaulted form `${CROW_AGENT_DISPLAY_NAME:-Claude Code}`, but `Resources/crow-create-ticket-SKILL.md.template` still uses the bare `$CROW_AGENT_DISPLAY_NAME` (no `:-Claude Code` default). They diverge at 4 spots:

- `Resources/crow-create-ticket-SKILL.md.template:114` — `gh issue create` example
- `Resources/crow-create-ticket-SKILL.md.template:127` — `glab issue create` example
- `Resources/crow-create-ticket-SKILL.md.template:140` — Step 4b code block
- `Resources/crow-create-ticket-SKILL.md.template:143` — Step 4b bullet wording ("Use `$CROW_AGENT_DISPLAY_NAME` … Fall back to `Claude Code` if unset")

`cmp` confirms the two files differ (char 4112, line 114). Looks like commit 55a99ed ("Use shell default for unset CROW_AGENT_DISPLAY_NAME") updated the live `skills/` copy but not the bundled `Resources/` template.

Two consequences:

1. **CI red.** `Tests/CrowTests/AttributionSkillTests.swift:132` (`liveTicketSkillAndBundledTemplateAreByteIdentical`) asserts these two files are byte-identical. They are not, so this test fails. (The parallel `crow-review-pr` pair and the new `FOOTER.md` pair *are* in sync — only the create-ticket pair is broken.)

2. **Functional regression in shipped builds.** `Scaffolder.bundledCreateTicketSkill()` (`Sources/Crow/App/Scaffolder.swift:212`) calls `loadFromRepo(...)` first, which returns `nil` in a release/installed build (no `Package.swift` near the app bundle), then falls back to the bundled `.template`. So installed users get the bare `$CROW_AGENT_DISPLAY_NAME` form. When the var is unset — the exact legacy / non-injected case this PR targets — the footer renders `[🐦‍⬛ Created with Crow via ](…)` (empty agent name, trailing space) instead of `via Claude Code`. That defeats the fix 55a99ed was meant to land, only half-applied.

**Fix:** bring `Resources/crow-create-ticket-SKILL.md.template` lines 114/127/140 to `${CROW_AGENT_DISPLAY_NAME:-Claude Code}` and update the line 143 bullet to match `skills/crow-create-ticket/SKILL.md:143`, so the two files are byte-identical again.

### Security Review

**Strengths:**
- No secrets, no injection surface. Env values come from `AgentKind` (a closed enum), not user input.
- Attribution link target stays hardcoded to `https://github.com/radiusmethod/crow`; fork-reference guard tests retained.
- `agentDisplayName(fromEnvironment:)` correctly treats whitespace-only `CROW_AGENT_DISPLAY_NAME` as unset before falling back.

**Concerns:** none.

### Code Quality

- 🟢 Env injection is cleanly wired: `TmuxBackend.configure` → `environmentEntries(for:)` → `newWindow(env:)` (`TmuxBackend.swift:228`), and the shell wrapper re-exports both vars. The `knownDisplayNames` "intentional second source of truth" is well-justified by the comment (registry isn't populated in CrowCore-only unit tests).
- 🟢 `expandSkillBody` replaces `via Claude Code` globally — fine for the current Cursor prompt path, just be aware it would rewrite any unrelated "via Claude Code" prose a future skill body might contain.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 0 Yellow, 2 Green] findings. The create-ticket template/skill drift both fails the byte-identity drift guard and ships an empty-agent-name footer to installs where the env var is unset.

---

dhilgaertner (approved):
## Code & Security Review

Reviewed the full diff (15 files, +366/-57): centralizing attribution in `CrowAttribution`, injecting `CROW_AGENT_KIND` / `CROW_AGENT_DISPLAY_NAME` into managed tmux windows, and wiring skills/templates to the shell expression `${CROW_AGENT_DISPLAY_NAME:-Claude Code}`.

### Critical Issues
None.

### Security Review
**Strengths:**
- No injection surface from the new env vars: `CROW_AGENT_DISPLAY_NAME` is resolved from a controlled `knownDisplayNames` map / `AgentRegistry`, never from user input, and `agentDisplayName(fromEnvironment:)` trims/validates before use.
- tmux env is passed as discrete `-e KEY=VALUE` argv elements (`TmuxController.swift:93,144`), so multi-word display names ("OpenAI Codex") can't word-split or inject.
- `${CROW_AGENT_DISPLAY_NAME:-Claude Code}` is shell parameter expansion inside double-quoted `gh`/`glab` `--body` args — substitution only, no re-evaluation of the expanded text. The `:-` default cleanly covers legacy/unset sessions.
- Attribution URL remains hard-pinned to `https://github.com/radiusmethod/crow`; fork-reference negative tests are retained.

**Concerns:** None blocking.

### Code Quality
- Strong drift protection: I confirmed `skills/crow-attribution/FOOTER.md`, `crow-review-pr/SKILL.md`, and `crow-create-ticket/SKILL.md` are byte-identical to their `Resources/*.template` counterparts, and `FOOTER.md` matches `CrowAttribution.sharedFooterInstructions`. The three resolution paths (repo file → bundled template → Swift constant) are each guarded by a test from both sides.
- Release bundling is automatic: `scripts/bundle.sh:65` globs `Resources/*.template`, so the new footer template ships without a build-script change.
- `CrowCore` test suite passes locally (18/18 via `swift test --filter CrowAttribution`). The `Tests/CrowTests` drift guards couldn't be executed here (the main target fails to build on a missing `GhosttyKit.xcframework` — an environment/checkout issue, not this PR), so I verified those invariants manually instead.

### Considerations (non-blocking)
- `CrowAttribution.agentDisplayName(for:)` maps an unregistered, non-hardcoded kind back to `Claude Code` (`CrowAttribution.swift:36`), which is in mild tension with the `AgentKind.displayName` note that the CROW-427 fallback must not silently be "Claude Code". In practice every production kind is registered (or one of the hardcoded three), so this only affects truly-unknown kinds — a reasonable documented default, just worth a mental note.
- The Claude Code skill path depends on the agent emitting `${CROW_AGENT_DISPLAY_NAME:-Claude Code}` inside a *double-quoted* `--body`. If a future change posts via `--body-file`/stdin/single quotes, the literal expression would leak into the artifact. The skill text is explicit about double-quoting, so this is fragility to keep in mind, not a defect.
- No direct test asserts `TmuxBackend.registerTerminal` injects the entries for the new `agentKind` param, though `environmentEntries(for:)` itself is covered and the wiring is trivial.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 3 Green] findings. Clean, well-tested, well-documented, and consistent with existing patterns.

---

PR #446, opened 2026-06-04
Fix Ghostty terminal selection clearing on mouse-up

### LINKED ISSUE #445: Ghostty terminal: text selection clears on mouse-up, blocking copy/paste
**Observed:** In the embedded Ghostty terminal view, dragging to select text shows the selection highlight while the mouse button is held, but the selection clears the instant the mouse button is released. This makes it impossible to copy text out of the terminal via the normal select-then-Cmd+C flow — by the time the keyboard shortcut fires, there's nothing selected.

**Expected:** Selection persists after mouse-up (standard terminal behavior — matches macOS Terminal, iTerm2, and Ghostty standalone). Cmd+C (or right-click → Copy) then operates on the persisted selection.

**Suspected causes:**
- A re-render of the terminal view on every tick / output line is clobbering the selection state.
- Mouse-up handler is treating the event as a "click outside selection → deselect" rather than "end of drag-select".
- Focus is being grabbed by another view on mouse-up (e.g. the sidebar / tab bar regains focus), which clears the embedded view's selection.
- Mouse-tracking mode in the embed is in the wrong state (e.g. application mode passing every mouse event to the agent, so the terminal never realizes a selection was completed).

**Acceptance:**
- Select text in the terminal pane, release mouse, selection stays highlighted.
- Cmd+C copies the selected text to the clipboard.
- Selection clears only on the usual triggers: clicking elsewhere in the terminal, typing into the terminal, or explicit Escape.
- No regression for terminal apps that use mouse events (vim/htop/etc. — make sure the fix doesn't accidentally swallow mouse events those apps need).

**Repro:** Open any Crow session terminal, drag-select any text, release the mouse.

dgershman:
Closes #445

## Summary

- Disable tmux mouse capture in `crow-tmux.conf` so Ghostty handles native text selection, scroll, and clipboard. With `mouse on`, tmux enables terminal mouse reporting and libghostty intentionally clears selection on every mouse button release.
- Sync cursor position before mouse button events so press/release use the correct cell.
- Skip redundant `makeFirstResponder` calls during SwiftUI `updateNSView` to reduce focus churn.

## Test plan

- [ ] Open a session terminal, drag-select text, release mouse — selection stays highlighted
- [ ] Cmd+C copies selected text to the clipboard
- [ ] Selection clears on click elsewhere, typing, or Escape
- [ ] `vim` with `set mouse=a` and `htop` still receive mouse events when those apps enable xterm mouse reporting
- [ ] Restart tmux server (or restart Crow) so existing cockpit sessions pick up the new `mouse off` setting

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift
@@ -371,15 +371,22 @@ public final class GhosttySurfaceView: NSView {
 
     public override func mouseDown(with event: NSEvent) {
         guard let surface else { return }
+        sendMousePosition(for: event)
         ghostty_surface_mouse_button(surface, GHOSTTY_MOUSE_PRESS, GHOSTTY_MOUSE_LEFT, translateMods(event.modifierFlags))
     }
 
     public override func mouseUp(with event: NSEvent) {
         guard let surface else { return }
+        sendMousePosition(for: event)
         ghostty_surface_mouse_button(surface, GHOSTTY_MOUSE_RELEASE, GHOSTTY_MOUSE_LEFT, translateMods(event.modifierFlags))
     }
 
     public override func mouseMoved(with event: NSEvent) {
+        guard let surface else { return }
+        sendMousePosition(for: event)
+    }
+
+    private func sendMousePosition(for event: NSEvent) {
         guard let surface else { return }
         let pos = convert(event.locationInWindow, from: nil)
         ghostty_surface_mouse_pos(surface, pos.x, frame.height - pos.y, translateMods(event.modifierFlags))
--- a/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf
@@ -23,9 +23,12 @@ set -gs default-terminal screen-256color
 # this, tmux steals one cell row from the Ghostty surface.
 set -gs status off
 
-# Mouse selection / scroll, since users will use Ghostty's mouse handlers
-# rather than tmux's prefix-driven copy mode.
-set -gs mouse on
+# Leave mouse off so Ghostty owns selection, scroll, and clipboard. With
+# mouse on, tmux captures every mouse event and enables terminal mouse
+# reporting; libghostty then clears the selection on mouse-up (#445).
+# Apps that need mouse (vim, htop, etc.) still enable it via the usual
+# xterm mouse protocol while running.
+set -gs mouse off
 
 # Crow owns the keyboard namespace. Drop every default binding; actions
 # are driven from Swift via `tmux <subcommand>` calls.
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TerminalSurfaceView.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TerminalSurfaceView.swift
@@ -72,7 +72,9 @@ public struct TerminalSurfaceView: NSViewRepresentable {
                 Self.attach(surface: surface, to: container)
             }
             try? TmuxBackend.shared.makeActive(id: id)
-            if let window = container.window, surface.window === window {
+            if let window = container.window,
+               surface.window === window,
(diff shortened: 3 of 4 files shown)


dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues

🔴 **The test added in this PR fails as written.** `BundledResourcesTests.swift:50` asserts:

```swift
#expect(!body.contains("mouse on"))
```

But the conf comment you added at `crow-tmux.conf:27` contains the substring `mouse on`:

```
# mouse on, tmux captures every mouse event and enables terminal mouse
```

`String.contains` does substring matching (same as the existing `status off` assertion), so `body.contains("mouse on")` returns `true` and the negated expectation fails. The test guards the exact regression it's meant to lock in, yet would go red on a clean checkout.

**Fix:** assert against the directive line, not a bare substring that the explanatory comment also matches. e.g.:

```swift
#expect(body.contains("set -gs mouse off"))
#expect(!body.contains("set -gs mouse on"))
```

(I could not run `swift test` here — the local checkout is missing the `GhosttyKit.xcframework` binary artifact — but the substring is confirmed present via grep at `crow-tmux.conf:27`.)

### Security Review

**Strengths:**
- No new attack surface. Changes are local AppKit mouse/focus plumbing and a bundled tmux config toggle.
- Disabling tmux's `mouse on` reduces, rather than expands, terminal mouse-reporting surface; apps still opt in via the standard xterm protocol.

**Concerns:** None.

### Code Quality

🟢 `GhosttySurfaceView.swift` — the `mouseMoved`/`sendMousePosition` refactor is clean. Minor: `mouseMoved` guards `surface` and then calls `sendMousePosition`, which guards `surface` again — the double `guard let surface` is redundant (harmless). Syncing position before press/release is the right fix for stale-cell button events.

🟢 `TerminalSurfaceView.swift` — the `window.firstResponder !== surface` guard before `makeFirstResponder` is a correct, low-risk way to cut focus churn during `updateNSView`.

🟢 `crow-tmux.conf` — `mouse off` with the explanatory comment is a sound design choice and well documented.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 0 Yellow, 3 Green] findings. The behavioral fix is correct; the accompanying test is broken and would fail CI. Tightening the assertion to the directive line resolves it.

---

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface: the change touches local terminal-input plumbing and a bundled tmux config — no network, filesystem writes, credential handling, or untrusted-input parsing.
- Mouse coordinate handling passes values straight to libghostty (`ghostty_surface_mouse_pos`); no arithmetic or buffer indexing that could overflow on the Swift side.

**Concerns:**
- None.

### Code Quality
- **crow-tmux.conf** — Correct root-cause fix. With `mouse on`, tmux enabled terminal mouse reporting and libghostty cleared the selection on every button release (#445). Setting `mouse off` lets Ghostty own selection/scroll/clipboard; inner apps (vim `set mouse=a`, htop) still negotiate xterm mouse reporting themselves, so the change is correctly scoped. Comment explains the *why* well.
- **GhosttySurfaceView.swift** — `sendMousePosition(for:)` extraction is behavior-preserving (`mouseMoved`→`sendMousePosition`, `mouseDragged`→`mouseMoved`). Syncing cursor position before press/release fixes stale-cell clicks. The redundant second `guard let surface` inside `mouseMoved` (the helper guards again) is harmless — consider dropping it for clarity, not blocking.
- **TerminalSurfaceView.swift** — Guarding `makeFirstResponder` on `window.firstResponder !== surface` correctly avoids redundant focus calls during `updateNSView`.
- **BundledResourcesTests.swift** — Good regression coverage; asserts both `mouse off` present and `mouse on` absent. Verified the conf comment uses `# mouse on,` (not `set -gs mouse on`), so the negative assertion has no false-positive.

### Notes
- I could not run `swift build`/tests here: the `GhosttyKit.xcframework` binary artifact isn't present in this review worktree (fetched separately). The new test is a pure config-string check and is independent of that binary. CI/local with the framework should exercise it.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 1 Green] findings. Focused, well-reasoned fix with a regression test.

---

PR #448, opened 2026-06-05
Fix attribution footer shell-expansion leak (#447)

### LINKED ISSUE #447: Attribution footer leaks literal ${CROW_AGENT_DISPLAY_NAME:-Claude Code} into git commits (#443 regression)
**Observed:** Crow-generated git commits contain the literal text:

```
🐦‍⬛ Created with Crow via ${CROW_AGENT_DISPLAY_NAME:-Claude Code}
```

The shell parameter expansion never fired — the placeholder went in raw.

**Root cause:** #443 changed the attribution template from a hardcoded `Claude Code` string to `${CROW_AGENT_DISPLAY_NAME:-Claude Code}`, which is **bash parameter expansion syntax**. It only resolves when the surrounding context lets the shell expand it. The committing flow appears to use one of these patterns that defeats expansion:

| Pattern | Expansion? |
|---|---|
| `git commit -m "via ${CROW_AGENT_DISPLAY_NAME:-Claude Code}"` (double-quoted, shell sees it) | ✓ works |
| `git commit -m "via $(cat <<'EOF' … EOF)"` (single-quoted heredoc) | ✗ literal text |
| `git commit -m 'via ${CROW_AGENT_DISPLAY_NAME:-Claude Code}'` (single-quoted) | ✗ literal text |
| Templates loaded from a file and passed verbatim | ✗ literal text |
| Body assembled in Swift / Go / TypeScript code that doesn't run a shell | ✗ literal text |

The skill instructions in `.claude/skills/crow-create-ticket/SKILL.md` themselves use `<<'EOF'` heredocs in the example commands — which is the form they tell callers to use — and that heredoc form **silently strips parameter expansion**. So even when the env var is set correctly, the placeholder survives intact.

**Fix options (pick one, don't try to make shell expansion universally reliable):**

1. **Resolve in code, substitute as plain text.** Whichever component owns the commit / issue body should look up `CROW_AGENT_DISPLAY_NAME` (env or session metadata) at body-construction time and substitute the resolved string before handing the body to git/gh/glab. No shell expansion involved — works in any heredoc, any quoting.
2. **Use a templating syntax that isn't shell.** Pick a sentinel like `{{agent_display_name}}` and have a single render helper replace it. Skills emit `{{agent_display_name}}` in their bodies; the runner substitutes before commit. Won't accidentally trigger shell expansion either way.
3. **If you keep shell expansion, audit every emit path.** Every heredoc that includes the footer must be unquoted (`<<EOF`, not `<<'EOF'`), and any other quoting around the body must let the shell see the variable. This is fragile — option 1 or 2 is cheaper to maintain.

Recommend **option 1** — resolve once in the launcher (Crow already knows the agent kind), pass the resolved display name into the session env *and* into any code paths that construct commit/issue bodies. Skills then write the literal name into their templates and the variable indirection goes away entirely.

**Acceptance:**
- New commits / issues / PR descriptions show the actual agent name (`Claude Code`, `Cursor`, `Codex`, `OpenCode`) in the attribution line — no literal `${…}` text in any artifact.
- `.claude/skills/crow-*/SKILL.md` updated to match the chosen approach (no more shell parameter expansion in heredoc examples).
- Backfill / cleanup: not needed for past commits, but call out so reviewers know to ignore the leaked-placeholder commits already in history.

**Repro:** Any commit Crow has authored since #443 landed will show the literal text.

dgershman:
## Summary

Closes #447.

PR #443 swapped the hardcoded `Claude Code` attribution string for the bash parameter expansion `\${CROW_AGENT_DISPLAY_NAME:-Claude Code}`. The expansion silently fails inside single-quoted heredocs (`<<'EOF'` — the form CLAUDE.md tells agents to use for `gh pr create --body`), single-quoted strings, and JSON written into files, so the literal placeholder text has been leaking into commits, issue bodies, and PR descriptions ever since.

This change implements Option 1 from the ticket: resolve the agent display name to a literal string **at the moment the file is written**, before any shell ever sees it.

## Approach

- **Canonical placeholder** in skill source templates is now `{{CROW_AGENT_DISPLAY_NAME}}` (the existing in-code sentinel), not a shell expression. Quoting context can no longer break it.
- **`Scaffolder.scaffold`** now takes a `managerAgentKind` parameter and runs every skill body + the FOOTER through `CrowAttribution.expandSkillBody` before writing into the dev-root `.claude/skills/`. Propagated from the three `AppDelegate` call sites via `config.agentKind(for: .manager)`.
- **`SessionService.setupReviewClonePrep`** expands the review skill body with `reviewAgentKind` before copying into the per-PR clone — Cursor was already covered via `cursorReviewPrompt`; this catches Claude Code and any future review-capable agent.
- **`skills/crow-workspace/setup.sh`** gains an `agent_display_name` bash helper (mirrors `CrowAttribution.knownDisplayNames`) and uses it to bake the resolved name into Claude Code's `attribution.commit` JSON, replacing the previous hardcoded \"Claude Code\".
- **Skill source files and `CrowAttribution.sharedFooterInstructions`** are rewritten to document the substitution model and explicitly forbid reintroducing shell parameter expansion.
- **`CrowAttribution.expandSkillBody`** keeps its backward-compatible handling of `\${…:-…}`, bare `\$VAR`, and `{{…}}`, so legacy `.claude/` trees on disk get cleaned up on next scaffold.

## Drift cleanup (incidental)

The pair-sync drove `Resources/crow-workspace-setup.sh.template` back into byte-for-byte alignment with `skills/crow-workspace/setup.sh`. The template had fallen significantly behind the live file; this PR re-syncs them. No behavior change for users on dev builds; .app users get the up-to-date bundled fallback.

## Test plan

- [x] `swift test --arch arm64` — 184 tests pass; updated `AttributionSkillTests` assertions invert (now expects `{{…}}` sentinel; shell expression must not appear).
- [ ] Manual create-ticket repro: in a Manager session, `/crow-create-ticket \"Test\"` — issue body footer reads `[🐦‍⬛ Created with Crow via Claude Code](...)` (or `Cursor` / `OpenAI Codex` per session config), never the `\${…}` literal.
- [ ] Manual review-pr repro: trigger a review session, post a `--request-changes`; review body footer reads the literal agent name.
- [ ] Manual commit repro: in a new worktree, make any commit; trailer reads `🐦‍⬛ Generated with <Agent Name>, orchestrated by Crow` matching the worktree's resolved agent.
- [ ] `rg '\\\${CROW_AGENT_DISPLAY_NAME' skills/ Resources/` returns only the explicit \"do not reintroduce\" warning lines in FOOTER.md.

## Backfill

None needed. Pre-fix commits will still contain the leaked literal — reviewers should ignore that text in history.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/CrowAttribution.swift
+++ b/Packages/CrowCore/Sources/CrowCore/CrowAttribution.swift
@@ -86,13 +86,16 @@ public enum CrowAttribution {
     public static let sharedFooterInstructions: String = """
     # Crow Attribution Footers
 
-    Crow injects these environment variables into every managed terminal:
+    Every skill body Crow writes to disk goes through a substitution pass that
+    replaces `{{CROW_AGENT_DISPLAY_NAME}}` with the session's resolved agent name
+    (`Claude Code`, `Cursor`, `OpenAI Codex`, …). The agent reads literal text — no
+    shell parameter expansion is involved, so footers survive every quoting form
+    (single-quoted heredocs, JSON files, Swift string literals).
 
-    - `CROW_AGENT_KIND` — raw agent id (`claude-code`, `cursor`, `codex`, …)
-    - `CROW_AGENT_DISPLAY_NAME` — human label (`Claude Code`, `Cursor`, `OpenAI Codex`, …)
-
-    **Always** use `${CROW_AGENT_DISPLAY_NAME:-Claude Code}` in attribution footers inside
-    double-quoted `gh`/`glab` arguments so the shell applies the default when unset.
+    **Do not** reintroduce `${CROW_AGENT_DISPLAY_NAME:-…}` or any other shell
+    expression in attribution footers. Use the literal name Crow already wrote
+    into your skill, or `{{CROW_AGENT_DISPLAY_NAME}}` if you are authoring a new
+    template.
 
     The link target is always `https://github.com/radiusmethod/crow` — never a fork or a value from the local git remote.
 
@@ -101,6 +104,6 @@ public enum CrowAttribution {
     | Created (issues, PR descriptions, etc.) | `[🐦‍⬛ Created with Crow via <agent>](https://github.com/radiusmethod/crow)` |
     | Reviewed | `[🐦‍⬛ Reviewed by Crow via <agent>](https://github.com/radiusmethod/crow)` |
 
-    Replace `<agent>` with `${CROW_AGENT_DISPLAY_NAME:-Claude Code}`. Do not change the URL or wrap the line in extra formatting.
+    Replace `<agent>` with the literal name (or `{{CROW_AGENT_DISPLAY_NAME}}` in source templates). Do not change the URL or wrap the line in extra formatting.
     """
 }
--- a/Packages/CrowCore/Sources/CrowCore/CrowAttribution.swift
+++ b/Packages/CrowCore/Sources/CrowCore/CrowAttribution.swift
@@ -83,19 +83,24 @@ public enum CrowAttribution {
     }
 
     /// Shared attribution rules copied into `.claude/skills/crow-attribution/FOOTER.md`.
+    ///
+    /// Crow runs this body through `expandSkillBody` before writing it to disk, so
+    /// the prose must read sensibly **after** every `{{CROW_AGENT_DISPLAY_NAME}}` /
+    /// `$CROW_AGENT_DISPLAY_NAME` / `via Claude Code` occurrence has been rewritten
+    /// to the resolved agent name (#447 review). Describe the substitution model
+    /// at a higher level than naming the sentinel literally.
     public static let sharedFooterInstructions: String = """
     # Crow Attribution Footers
 
-    Every skill body Crow writes to disk goes through a substitution pass that
-    replaces `{{CROW_AGENT_DISPLAY_NAME}}` with the session's resolved agent name
-    (`Claude Code`, `Cursor`, `OpenAI Codex`, …). The agent reads literal text — no
-    shell parameter expansion is involved, so footers survive every quoting form
-    (single-quoted heredocs, JSON files, Swift string literals).
+    This file is what your skill sees on disk. The footer lines below already contain
+    the literal agent name for this session (`Claude Code`, `Cursor`, `OpenAI Codex`, …) —
+    Crow substituted it in when scaffolding. Copy each line verbatim into the body you
+    pass to `gh`/`glab` (or your commit). No shell parameter expansion is needed and the
+    line survives every quoting form (single-quoted heredocs, JSON files, Swift literals).
 
     **Do not** reintroduce `${CROW_AGENT_DISPLAY_NAME:-…}` or any other shell
-    expression in attribution footers. Use the literal name Crow already wrote
-    into your skill, or `{{CROW_AGENT_DISPLAY_NAME}}` if you are authoring a new
-    template.
+    expression in attribution footers. The shell silently fails to expand it inside
+    single-quoted heredocs and the literal text leaks into the artifact (#447).
 
     The link target is always `https://github.com/radiusmethod/crow` — never a fork or a value from the local git remote.
 
@@ -104,6 +109,8 @@ public enum CrowAttribution {
     | Created (issues, PR descriptions, etc.) | `[🐦‍⬛ Created with Crow via <agent>](https://github.com/radiusmethod/crow)` |
     | Reviewed | `[🐦‍⬛ Reviewed by Crow via <agent>](https://github.com/radiusmethod/crow)` |
 
-    Replace `<agent>` with the literal name (or `{{CROW_AGENT_DISPLAY_NAME}}` in source templates). Do not change the URL or wrap the line in extra formatting.
+    `<agent>` above is just a placeholder for *this document* — in the real footer lines
+    your skill receives, the agent name is already filled in. Do not change the URL or
+    wrap the line in extra formatting.
     """
 }
--- a/Resources/crow-attribution-FOOTER.md.template
+++ b/Resources/crow-attribution-FOOTER.md.template
@@ -1,15 +1,14 @@
 # Crow Attribution Footers
 
-Every skill body Crow writes to disk goes through a substitution pass that
-replaces `{{CROW_AGENT_DISPLAY_NAME}}` with the session's resolved agent name
-(`Claude Code`, `Cursor`, `OpenAI Codex`, …). The agent reads literal text — no
-shell parameter expansion is involved, so footers survive every quoting form
-(single-quoted heredocs, JSON files, Swift string literals).
+This file is what your skill sees on disk. The footer lines below already contain
+the literal agent name for this session (`Claude Code`, `Cursor`, `OpenAI Codex`, …) —
+Crow substituted it in when scaffolding. Copy each line verbatim into the body you
+pass to `gh`/`glab` (or your commit). No shell parameter expansion is needed and the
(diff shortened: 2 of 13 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Solid, well-scoped fix for #447. Resolving `{{CROW_AGENT_DISPLAY_NAME}}` to a literal at file-write time is the right model — it removes shell quoting from the equation entirely, so the footer survives single-quoted heredocs, JSON, and Swift literals. The change is coherent across all write sites, the bash helper mirrors the Swift source of truth, the template/live drift is re-synced byte-for-byte, and the test inversions are meaningful. One Yellow keeps this from a clean approve.

### Critical Issues
None.

### Security Review
**Strengths:**
- No injection surface added. Resolved display names (`Claude Code`, `Cursor`, `OpenAI Codex`) are a closed, hardcoded set with no quotes/backslashes/`$`, so baking them into the unquoted `<<EOF` heredoc in `write_settings_local` (`skills/crow-workspace/setup.sh:457-466`) produces valid JSON with no expansion hazard.
- `agent_display_name()` (`setup.sh:136-143`) defaults unknown kinds to `Claude Code`, and `AGENT_KIND` is validated early in `parse_args` (`setup.sh:255-259`), so no untrusted value reaches the JSON.
- Substitution resolution order in `write_settings_local` (`$AGENT_KIND` → `read_agent_kind_from_config "work"`) matches `launch_agent` (`setup.sh:793-799`), so the baked commit attribution and the launched agent can't disagree.
- The link target stays the hardcoded `repoURL`; nothing derives it from the local remote.

**Concerns:** None.

### Code Quality

**Yellow — the substitution pass mangles the prose of the files documenting it.** `Scaffolder.scaffold` now runs `expandSkillBody` over `FOOTER.md` and the two SKILLs (`Scaffolder.swift:88,100,106`), and those files' *explanatory* sentences reference the literal `{{CROW_AGENT_DISPLAY_NAME}}` token to describe the mechanism — so the substitution rewrites the explanation along with the real footer line. After scaffolding (shown here for a Cursor manager):

- `skills/crow-attribution/FOOTER.md:3-4` → "a substitution pass that replaces `Cursor` with the session's resolved agent name" (self-contradictory)
- `FOOTER.md:11` → "or `Cursor` if you are authoring a new template"
- `FOOTER.md:18` → "Replace `<agent>` with the literal name (or `Cursor` in source templates)"
- `skills/crow-create-ticket/SKILL.md:144` and `skills/crow-review-pr/SKILL.md:146` → "Crow rewrites `Cursor` to the session's resolved agent name …" (self-contradictory)

No emitted artifact is wrong — the operative footer table (`<agent>`) plus "use the literal name" still yields the correct footer, and each consumer gets its own correctly-resolved name. But this PR newly (a) adds these `{{…}}` tokens into prose and (b) runs `expandSkillBody` over `FOOTER.md` (previously written verbatim, `Scaffolder.swift:106`), so it introduces shipped instruction files whose "how it works" sentences contradict themselves. Trivially fixable: reword the explanatory sentences so they don't contain a live `{{CROW_AGENT_DISPLAY_NAME}}` / `$CROW_AGENT_DISPLAY_NAME` token (e.g. describe the token without spelling it as the exact substitutable string, or reword to read correctly post-substitution). Note the `${…:-…}` "do not reintroduce" warning is *not* affected — the brace prevents the bare-`$VAR` rule from matching it.

**Green — `Co-Authored-By: Claude` is hardcoded for non-Claude agents.** `write_settings_local` now bakes "Generated with $display_name" but the next trailer line stays `Co-Authored-By: Claude <noreply@anthropic.com>` (`setup.sh:464`). For a Cursor/Codex worktree the commit reads "Generated with Cursor … Co-Authored-By: Claude". `settings.local.json` is a Claude-Code-specific file that those agents won't read, so this is pre-existing and out of scope — just flagging for awareness.

### Tests
- CrowCore package: 231 tests pass locally (`swift test --arch arm64`). The app target can't build in this environment (missing `GhosttyKit.xcframework` binary — unrelated to this PR), so the inverted `AttributionSkillTests` weren't executed here, but they're straightforward placeholder/byte-identity assertions and the source changes line up with them.
- `setup.sh` passes `bash -n`; `agent_display_name()` verified to map all kinds + fallback correctly.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 1 Green] findings. The fix itself is correct and ready; the one Yellow is a self-contradiction in the shipped doc prose that's cheap to clean up in this same round.

---

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No injection surface introduced. `agent_display_name` in `setup.sh` maps only the validated `claude-code|cursor|codex` set (with a `Claude Code` default), so no attacker-controlled or quote-bearing value can reach the `attribution.commit` JSON. `$display_name`/`$SESSION_ID` are constrained values, and the heredoc stays unquoted exactly as before (the `\\n` → `\n` JSON-escape behavior is unchanged).
- The canonical repo URL (`https://github.com/radiusmethod/crow`) is preserved everywhere and explicitly guarded by tests — no fork/remote-derived target can sneak in.

**Concerns:**
- None.

### Code Quality
Solid, well-scoped fix that resolves #447 by resolving the agent display name to a literal **at file-write time** rather than relying on shell expansion that silently fails inside single-quoted heredocs.

- `expandSkillBody` (`CrowAttribution.swift:76`) handles all four legacy/canonical forms (`${…:-…}`, bare `$VAR`, `{{…}}`, and the `via Claude Code` literal), so stale on-disk `.claude/` trees self-heal on next scaffold. Verified the FOOTER "do not reintroduce" warning (`${CROW_AGENT_DISPLAY_NAME:-…}`, with an ellipsis) matches none of the four patterns and survives expansion intact.
- All write sites are covered: `Scaffolder.scaffold` (5 skill bodies + FOOTER), `SessionService.setupReviewClonePrep` (review clone, threaded via `reviewAgentKind`), and `setup.sh write_settings_local` (the commit-trailer JSON). The `setup.sh` fallback (`AGENT_KIND` → `read_agent_kind_from_config "work"`) correctly mirrors `launch_agent`'s resolution order.
- The `setup.sh` ↔ `Resources/crow-workspace-setup.sh.template` re-sync is genuinely byte-for-byte identical (verified via `diff`), fixing real template drift.
- Tests invert cleanly and add meaningful regression guards: the leak form must be absent, the `{{…}}` sentinel must only live on footer lines (URL-bearing), and FOOTER prose must not contain a substitutable token (preventing self-contradictory deployed docs). `swift test` on `CrowCore` passes locally (231 tests). The app-target `AttributionSkillTests` could not be exercised in this review environment (missing `GhosttyKit.xcframework` binary), but the assertions were read and match the implementation.

**Consider (non-blocking, no change required):**
- `write_settings_local` writes Claude Code's `settings.local.json` for *every* worktree regardless of agent kind, and that file is only ever read by Claude Code — so in practice `$display_name` there always resolves to `Claude Code`. The new agent-aware value is therefore inert for `cursor`/`codex` worktrees (they have no equivalent commit-trailer mechanism wired here). Harmless and out of scope for #447, just worth noting the "Generated with <agent>" branch is currently unreachable for non-Claude agents.
- The `Co-Authored-By: Claude <noreply@anthropic.com>` trailer remains hardcoded; consistent in practice given the point above, but a future agent-attribution pass may want to revisit it alongside the line above.
- The `via Claude Code` catch-all in `expandSkillBody` is a broad substring replace; currently zero occurrences exist in any skill source (grep-verified), so it's purely a legacy-cleanup safety net today — fine, just a latent footgun if literal documentation prose ever uses that exact phrase.

### Summary Table

---

PR #451, opened 2026-06-06
Reconcile stale bundled tmux conf on attach (CROW-450)

### LINKED ISSUE #450: Detect stale tmux server config on app launch (hot-apply diff or prompt to restart)
**Context:** PR #446 fixed #445 (Ghostty selection clearing on mouse-up) by setting `mouse off` in the bundled `crow-tmux.conf`. The code fix landed correctly, but users who already had a long-lived tmux server kept seeing the bug — because **tmux only reads `-f <conf>` once at server startup**, and Crow's persistent socket means the same server can outlive many app launches and rebuilds.

**Repro of the silent-failure mode:**
1. Run Crow at some point. tmux server starts with whatever config existed at the time.
2. Update Crow (pull, rebuild, app update — whatever). `crow-tmux.conf` on disk now has the fix.
3. Relaunch Crow. The app reattaches to the existing tmux server via the persistent socket — never reads the updated config.
4. User keeps hitting the original bug. `tmux show -gv mouse` reports `on` even though the config file on disk says `off`.
5. The only workarounds are (a) kill the tmux server (destroys all open sessions/terminals) or (b) manually `tmux set -gs mouse off` on the live socket — neither is discoverable.

In my case the tmux server had been running for 13 days before I hit this; the fix had been on disk for 2 days. From the user's POV, "Crow is broken" with no way to know the fix shipped.

**Concrete instance:** my running server PID 68731, started May 23. `crow-tmux.conf` mtime June 4. Reported `mouse on` until manually hot-patched. (Closing #445 prematurely from a per-installation perspective even though the code was correct.)

**Asks (any one would work; recommend #1):**

1. **Hot-apply config diff on attach.** When Crow starts up and attaches to an existing tmux server, compare the bundled `crow-tmux.conf` mtime against the server's start time (`tmux display -p '#{start_time}'`). If the conf is newer:
   - Parse the conf for `set -gs <option> <value>` lines and re-apply each via `tmux set -gs ...` on the live server.
   - Or simpler: `tmux source-file <conf>` — this re-evaluates the file in place. Works for most server-scoped settings including `mouse`.
   - Either is non-destructive; existing sessions and windows survive.

2. **Prompt to restart the tmux server** when the config has changed. Surface a banner like "Crow's terminal config has been updated. Restart the tmux server to apply (will close current sessions). [Restart] [Later]". Acceptable but worse UX than option 1 since it forces a session loss.

3. **Kill+restart automatically** on any version bump. Simplest, most destructive. Bad UX — silently kills the user's open work. Not recommended.

**Acceptance:**
- After updating Crow with a `crow-tmux.conf` change, the live tmux server reflects the new settings on next Crow launch without manual intervention.
- `tmux show -gv <option>` for any option changed in the bundled conf returns the new value after one app relaunch.
- Existing sessions / terminals are not closed by the reconciliation.
- Add a regression test: bundle a known config change, simulate "Crow was running with old config", verify the new settings are applied on next launch.

**Related:**
- #445 / #446 — the original bug + fix that exposed this whole class of "code fix doesn't reach long-running users" problem.
- Worth a broader pattern note: any future tmux conf change has the same shape. Worth landing this fix once and never thinking about it again.

dgershman:
## Summary

- When `TmuxBackend.ensureRunningServer` adopts a cockpit session that was already live before this Crow process attached, run `tmux source-file <bundled-conf>` on the live server if the on-disk conf is newer than `#{start_time}`.
- This applies server-scoped options (`mouse`, `status`, `escape-time`, …) without touching existing windows. Newly created servers skip reconciliation because they just loaded the conf via `-f`.
- Closes #450.

## Why

Since #330, the bundled tmux server outlives a clean Crow quit and is re-attached on the next launch. tmux only reads `-f <conf>` once at server start, so fixes shipped through `crow-tmux.conf` (e.g. #446's `mouse off`) never reach users whose server predates the update. The reporter's server had run 13 days while the fix had been on disk 2 days; `tmux show -gv mouse` still reported `on`.

## Approach

- `shouldReconcile(configMTime:serverStartTime:)` — pure policy: reconcile if conf > server start, or if either timestamp is unreadable (conservative; `source-file` is cheap).
- `serverStartTime(controller:)` — wraps `tmux display -p '#{start_time}'`.
- `reconcileBundledConfigIfStale(controller:configURL:)` — the orchestrator; logs `[CrowTelemetry tmux:config_reconciled path=…]` or `[CrowTelemetry tmux:config_reconcile_skipped reason=fresh]`.
- Gate at the `ensureRunningServer` hook: capture `ctrl.hasSession()` BEFORE `ensureCockpitSession` so we only reconcile servers we attached to, never servers we just created.

Chose `source-file` over a parsed per-option diff because the conf is small, source-file is exactly what tmux runs on `-f`, and the only non-perfect-idempotency wart (`set -gas terminal-features ',…'` re-appends) is benign — tmux merges feature flags by name.

## Test plan

- [x] Added pure-policy unit tests (`TmuxBackendConfigPolicyTests`) covering both-known/older/newer/equal/either-nil cases.
- [x] Added integration regression `staleConfIsReconciledOnAttach`: write conf A with `status on`, start server, replace with conf B (`status off`) with fresh mtime, call reconciler, assert `tmux show -gv status` flips to `off` AND existing windows survive.
- [x] Added integration regression `freshConfIsNotReconciled`: back-dated newer-content conf is NOT applied (gate works).
- [x] `swift test --package-path Packages/CrowTerminal --filter TmuxBackend` → 23/23 pass.
- [ ] Manual repro: `touch Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf`, quit + relaunch Crow, observe `tmux:config_reconciled` log line and `tmux show -gv mouse` reports `off`.

Closes #450

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -488,13 +488,70 @@ public final class TmuxBackend {
             socketPath: socketPath,
             sessionName: TmuxBackend.cockpitSessionName
         )
-        let confPath = BundledResources.tmuxConfURL?.path
-        if confPath == nil {
+        guard let confURL = BundledResources.tmuxConfURL else {
             throw TmuxBackendError.bundledResourceMissing("crow-tmux.conf")
         }
-        try Self.ensureCockpitSession(ctrl, configPath: confPath)
+        // The cockpit session may already be live from a prior Crow launch
+        // (#330 stable socket). If so, the bundled conf the server loaded at
+        // `-f` time may now be stale relative to the file on disk (#450) —
+        // capture the pre-attach state so we can reconcile below.
+        let serverWasAlreadyLive = ctrl.hasSession()
+        try Self.ensureCockpitSession(ctrl, configPath: confURL.path)
         controller = ctrl
+        if serverWasAlreadyLive {
+            Self.reconcileBundledConfigIfStale(controller: ctrl, configURL: confURL)
+        }
         return ctrl
+    }
+
+    // MARK: - Stale-config reconciliation (#450)
+
+    /// Re-source the bundled tmux conf on a live server iff the file on disk
+    /// has been modified since the server started. Non-destructive: existing
+    /// windows/sessions survive a `source-file` — server-scoped options
+    /// (mouse, status, escape-time, …) update in place. Failures are logged
+    /// and swallowed; a stale conf is not worth aborting startup over.
+    ///
+    /// Caveat: the bundled conf includes `set -gas terminal-features ',…'`
+    /// which re-appends on each source-file. tmux tolerates duplicate feature
+    /// flags (merged by name) so the duplication is benign.
+    nonisolated static func reconcileBundledConfigIfStale(
+        controller: TmuxController,
+        configURL: URL
+    ) {
+        let confPath = configURL.path
+        let confMTime = (try? FileManager.default.attributesOfItem(atPath: confPath))?[.modificationDate] as? Date
+        let serverStart = serverStartTime(controller: controller)
+
+        guard shouldReconcile(configMTime: confMTime, serverStartTime: serverStart) else {
+            NSLog("[CrowTelemetry tmux:config_reconcile_skipped reason=fresh]")
+            return
+        }
+
+        do {
+            try controller.run(["source-file", confPath])
+            NSLog("[CrowTelemetry tmux:config_reconciled path=\(confPath)]")
+        } catch {
+            NSLog("[CrowTelemetry tmux:config_reconcile_failed error=\"\(error)\"]")
+        }
+    }
+
+    /// Pure policy: reconcile when either timestamp is missing (conservative
+    /// — a redundant `source-file` is cheap) or when the conf is newer than
+    /// the running server.
+    nonisolated static func shouldReconcile(configMTime: Date?, serverStartTime: Date?) -> Bool {
+        guard let configMTime, let serverStartTime else { return true }
+        return configMTime > serverStartTime
+    }
+
+    /// `tmux display -p '#{start_time}'` → Unix epoch as a string. Returns
+    /// nil on any IO/parse failure; callers treat nil as "unknown — reconcile
+    /// to be safe".
+    nonisolated static func serverStartTime(controller: TmuxController) -> Date? {
+        guard let raw = try? controller.run(["display", "-p", "#{start_time}"]) else { return nil }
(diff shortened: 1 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No untrusted input: `configURL` is the app-bundled `crow-tmux.conf`, `source-file` runs an app-internal resource, and `#{start_time}` is parsed defensively (`TimeInterval(trimmed)` guard, nil → conservative reconcile).
- `reconcileBundledConfigIfStale` swallows and logs failures, so a stale/unreadable conf can never abort startup.
- Telemetry log lines interpolate a controlled bundle path and an `error` description — no secret/credential exposure.

**Concerns:** None.

### Code Quality
Verified against the source, the actual bundled conf, and the surrounding race-handling code:
- The gate is correct: `serverWasAlreadyLive = ctrl.hasSession()` is captured *before* `ensureCockpitSession`, so freshly-created servers (which already loaded the conf via `-f`) are skipped and only adopted/surviving servers reconcile.
- `source-file` is genuinely non-destructive here — `crow-tmux.conf` is exclusively server-scoped `set -gs`/`set -gas` options plus `unbind-key -a`, all idempotent. Confirmed Crow issues **no** runtime `tmux bind` calls, so the re-run `unbind-key -a` wipes nothing. The `terminal-features` re-append is correctly called out as benign.
- Clean separation of pure policy (`shouldReconcile`) from IO (`serverStartTime`, FileManager), with the conservative reconcile-on-unknown bias well-justified.
- Test coverage is solid: pure-policy unit tests cover all four timestamp permutations, and the two integration regressions assert both the flip (`status` on→off) *and* window survival, plus the negative gate.

**Green (consider only):**
- The documented `ensureRunningServer` reentrancy means several concurrent callers can each invoke `reconcileBundledConfigIfStale` on launch. For freshly-created servers each call is a skip; for a surviving stale server, a few redundant concurrent `source-file` calls may run. Idempotent and harmless — noting only for completeness.
- `#{start_time}` is whole-second epoch while the conf mtime is sub-second; a conf sourced at server start always has an older mtime, so the worst case is one redundant `source-file` (consistent with the conservative bias). No action needed.

Note: I could not compile/run the suite locally — this checkout omits `Frameworks/GhosttyKit.xcframework`'s binary artifact (environment limitation, not a PR issue). The PR documents `swift test --filter TmuxBackend` → 23/23 passing.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. A focused, well-tested, non-destructive fix with clear reasoning and good policy/IO separation.

PR #453, opened 2026-06-06
Restore tmux mouse, keep selection via copy-pipe-no-clear

### LINKED ISSUE #452: Replace 'mouse off' with 'mouse on + copy-pipe-no-clear' in crow-tmux.conf (#446 broke wheel scrollback)
**Context:** PR #446 closed #445 (selection clears on mouse-up) by setting `mouse off` in `crow-tmux.conf`. Verified live that this trade isn't worth it — `mouse off` re-introduces a worse bug: **wheel-scrollback through tmux pane history stops working**, because tmux no longer captures wheel events. Ghostty's own scrollback is functionally useless under tmux (tmux constantly redraws the same screen area, so Ghostty's buffer is full of redraw artifacts rather than command history).

So #446 traded selection for scrollback. We can have both.

**The fix:** keep `mouse on`, and override the default `MouseDragEnd1Pane` binding to drop the `-cancel` suffix. The default in both `copy-mode` and `copy-mode-vi` tables is `send-keys -X copy-pipe-and-cancel`; the `-and-cancel` exits copy mode after the copy, which is what wipes the selection. Use `copy-pipe-no-clear` instead — selection persists, text still copies to clipboard automatically.

**Replace this block in `crow-tmux.conf`:**

```
# Leave mouse off so Ghostty owns selection, scroll, and clipboard. With
# mouse on, tmux captures every mouse event and enables terminal mouse
# reporting; libghostty then clears the selection on mouse-up (#445).
# Apps that need mouse (vim, htop, etc.) still enable it via the usual
# xterm mouse protocol while running.
set -gs mouse off
```

**With:**

```
# Mouse on so wheel scrolls tmux pane history. Selection-clear on
# mouse-up (#445) is fixed by the bindings below, not by disabling mouse
# entirely (which broke wheel scrollback — see [link to this ticket]).
set -gs mouse on

# Default binding is copy-pipe-and-cancel — the -and-cancel exits copy
# mode after copying, which clears the visual selection. -no-clear keeps
# the selection highlighted AND copies to the macOS pasteboard.
bind -T copy-mode    MouseDragEnd1Pane send-keys -X copy-pipe-no-clear "pbcopy"
bind -T copy-mode-vi MouseDragEnd1Pane send-keys -X copy-pipe-no-clear "pbcopy"
```

**Update the regression test** at `Packages/CrowTerminal/Tests/CrowTerminalTests/BundledResourcesTests.swift:44-51` — it currently enforces `mouse off`. After this fix it should enforce:
- `set -gs mouse on` present
- Both `MouseDragEnd1Pane` bindings present with `copy-pipe-no-clear`
- The old `set -gs mouse off` is **not** present

**Acceptance (verified manually on hot-patched live server):**
- Drag-select text in a Crow terminal → release mouse → selection stays highlighted ✓
- Selected text is automatically on the macOS clipboard (Cmd+V elsewhere pastes it) ✓
- Wheel-up scrolls through tmux pane history ✓
- No regression for terminal apps that use mouse (vim/htop/etc. — they enable xterm mouse protocol at runtime, which still works the same way) ✓

**Related:**
- #445 — original bug. The fix here is documented in my comment at https://github.com/radiusmethod/crow/issues/445#issuecomment-4640742272 — reopening #445 is fine too, but a fresh ticket is more discoverable for the auto-pickup queue.
- #446 — the over-correcting fix this ticket replaces.
- #450 — long-lived tmux servers won't see this config change without a reload mechanism; landing #450 makes future tmux-conf fixes painless for users who don't restart their tmux server.

dgershman:
## Summary

- PR #446 fixed the Ghostty selection-clear from #445 by setting `mouse off` in `crow-tmux.conf`. Live verification showed this re-introduces a worse bug: **wheel-scrollback through tmux pane history stops working** because tmux no longer captures wheel events, and Ghostty's own scrollback is unusable under tmux (constant redraw).
- Restore `mouse on` and override `MouseDragEnd1Pane` in both `copy-mode` and `copy-mode-vi` tables to use `copy-pipe-no-clear "pbcopy"` instead of the default `copy-pipe-and-cancel`. The `-and-cancel` is what exits copy mode after copy and clears the selection — `-no-clear` keeps the highlight while still copying to the macOS pasteboard.
- Update `BundledResourcesTests` to enforce the new contract (mouse on, mouse off absent, both `MouseDragEnd1Pane` bindings present with `copy-pipe-no-clear`).

## Test plan

- [x] `swift test --arch arm64 --filter BundledResourcesTests` — all 6 tests pass, including the rewritten `tmuxConfEnablesMouseWithCopyPipeNoClear`.
- [ ] Manual smoke on a freshly-started tmux server (see #450 for reload caveat):
  - [ ] Drag-select text → release → selection stays highlighted.
  - [ ] Cmd+V into another app pastes the selected text.
  - [ ] Wheel-up scrolls tmux pane history.
  - [ ] `vim` / `htop` mouse still works (they enable xterm mouse protocol at runtime, independent of `mouse on`).

Closes #452.

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf
@@ -23,16 +23,25 @@ set -gs default-terminal screen-256color
 # this, tmux steals one cell row from the Ghostty surface.
 set -gs status off
 
-# Leave mouse off so Ghostty owns selection, scroll, and clipboard. With
-# mouse on, tmux captures every mouse event and enables terminal mouse
-# reporting; libghostty then clears the selection on mouse-up (#445).
-# Apps that need mouse (vim, htop, etc.) still enable it via the usual
-# xterm mouse protocol while running.
-set -gs mouse off
+# Mouse on so wheel scrolls tmux pane history. Selection-clear on
+# mouse-up (#445) is fixed by the bindings below, not by disabling mouse
+# entirely — that broke wheel scrollback (#452). Apps that need mouse
+# (vim, htop, etc.) enable the xterm mouse protocol at runtime, which
+# still works alongside this.
+set -gs mouse on
 
 # Crow owns the keyboard namespace. Drop every default binding; actions
-# are driven from Swift via `tmux <subcommand>` calls.
+# are driven from Swift via `tmux <subcommand>` calls. `-a` clears the
+# root tables only — copy-mode / copy-mode-vi bindings below survive.
 unbind-key -a
+
+# Default MouseDragEnd1Pane is `send-keys -X copy-pipe-and-cancel`; the
+# -and-cancel exits copy mode after copying, which wipes the visual
+# selection. -no-clear keeps the selection highlighted AND copies to the
+# macOS pasteboard. Bound in both copy-mode tables since either may be
+# active depending on the user's `mode-keys` setting.
+bind -T copy-mode    MouseDragEnd1Pane send-keys -X copy-pipe-no-clear "pbcopy"
+bind -T copy-mode-vi MouseDragEnd1Pane send-keys -X copy-pipe-no-clear "pbcopy"
 
 # Default history-limit is 2000 lines per pane. Phase 3 §2 (the spike)
 # measured 5 windows × 10k lines as a 400kB RSS delta, so we have plenty
--- a/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf
@@ -30,9 +30,12 @@ set -gs status off
 # still works alongside this.
 set -gs mouse on
 
-# Crow owns the keyboard namespace. Drop every default binding; actions
-# are driven from Swift via `tmux <subcommand>` calls. `-a` clears the
-# root tables only — copy-mode / copy-mode-vi bindings below survive.
+# Crow owns the keyboard namespace. `unbind-key -a` (no `-T`) clears the
+# `prefix` table only — `root`, `copy-mode`, and `copy-mode-vi` are left
+# intact (use `-a -T <table>` to clear a non-prefix table). That's
+# load-bearing here: the mouse fix depends on the root-table defaults
+# (`WheelUpPane`, `MouseDrag1Pane`, …) surviving — wiping root with
+# `unbind-key -a -T root` would re-break #452 wheel scrollback.
 unbind-key -a
 
 # Default MouseDragEnd1Pane is `send-keys -X copy-pipe-and-cancel`; the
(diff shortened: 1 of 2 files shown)


dhilgaertner (changes requested):
## Code & Security Review

I verified this PR's behavior empirically against a live **tmux 3.6a** server (loading the bundled `crow-tmux.conf` and inspecting the resulting key tables / options), rather than reasoning from the man page — because the man page turns out to be misleading here.

### What works (verified live)
- ✅ `set -gs mouse on` takes effect.
- ✅ **Wheel scrollback restored** — the root-table `WheelUpPane → copy-mode -e` binding is present, so the wheel scrolls tmux pane history again (the #452 regression is fixed).
- ✅ **Drag-select keeps the highlight and copies** — `MouseDragEnd1Pane` is overridden to `copy-pipe-no-clear "pbcopy"` in **both** `copy-mode` and `copy-mode-vi`.
- ✅ Test assertions match the conf byte-for-byte (the build can't run in this review checkout — `Frameworks/GhosttyKit.xcframework` has no binary artifact — but the tests are pure string-presence checks and all four substrings are present).

### Security Review
No security surface touched. `pbcopy` is invoked with a fixed, non-interpolated argument; no untrusted input flows into the binding. Nothing to flag.

### Concerns

**🟡 The comment on `unbind-key -a` is backwards (crow-tmux.conf:33-36).**
The comment claims:
> Drop every default binding ... `-a` clears the root tables only — copy-mode / copy-mode-vi bindings below survive.

Measured behavior on tmux 3.6a is the opposite:

| table | keys before `unbind-key -a` | keys after |
|-------|------|------|
| prefix | 87 | **1** |
| root | 19 | **19** (unchanged) |
| copy-mode | 74 | 74 |

`unbind-key -a` (no `-T`) clears the **prefix** table and leaves the **root** table fully intact — root is the one table it does *not* touch. `unbind-key -a` alone is effectively a no-op on root (you'd need `-a -T root` to clear it). So:
- "`-a` clears the root tables only" is exactly inverted.
- The copy-mode bindings do survive, but not for the reason given — they're added *after* the call, and `-a` never touches copy-mode anyway.

This isn't pedantry: the entire restored feature **silently depends on the root mouse bindings (`WheelUpPane`, `MouseDrag1Pane`, …) surviving** `unbind-key -a`. A future maintainer who trusts this comment and "fixes" the unbind to actually drop root per the stated intent (`unbind-key -a -T root`) would wipe `WheelUpPane` and re-break wheel scrollback — the precise bug this PR exists to fix. Please correct the comment to describe what `-a` actually does (clears prefix, root survives) and note that wheel/drag rely on the surviving root defaults.

**🟡 Double-click / triple-click selections still clear on copy.**
The surviving root bindings `DoubleClick1Pane` (select-word) and `TripleClick1Pane` (select-line) still end in `copy-pipe-and-cancel` — the same `-and-cancel` this PR replaces for the drag gesture. So word/line selection via double/triple-click copies to the pasteboard but does **not** keep the highlight, which is inconsistent with the drag behavior the PR introduces. If the intent is "mouse copy keeps the selection," consider overriding these two root bindings to `copy-pipe-no-clear` as well; if the drag-only scope is intentional, a one-line note in the conf would prevent the next reader from assuming it's a bug.

### Code Quality
- Binding both `copy-mode` and `copy-mode-vi` is the right call (mode-keys can land in either) — good.
- The test rewrite correctly enforces the new contract (mouse on, mouse off absent, both bindings present).

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 0 Green] findings. The change is functionally correct and the regression fix genuinely works; both findings are about the load-bearing `unbind-key -a` comment being inverted and the double/triple-click consistency gap. Neither requires a redesign — a comment correction and a decision on the click gestures resolves both in one round trip.

---

dhilgaertner (approved):
## Code & Security Review

Empirically verified the tmux mechanics against the same engine Crow bundles for (`tmux 3.6a`), not just the string-`contains` unit test. The change is sound.

### Critical Issues
None.

### Verification performed
Loaded the bundled `crow-tmux.conf` into a throwaway tmux 3.6a server and confirmed:

- **`unbind-key -a` scope (the load-bearing claim).** With no `-T`, it clears **only** the `prefix` table. Root stayed at 19 keys, `copy-mode-vi` at 87, and `WheelUpPane`/`MouseDrag1Pane` survived. The updated comment (lines 33–38) is accurate, and the `unbind-key -a -T root` warning correctly captures why a broader unbind would re-break #452.
- **Config parses cleanly** — `new-session` with the conf exits 0, no stderr parse errors.
- **Intended bindings install** — `mouse on`; both `MouseDragEnd1Pane` overrides in `copy-mode` and `copy-mode-vi`; `Double/TripleClick1Pane` in root with the nested `if-shell` wrappers matching the tmux defaults, only swapping `copy-pipe-and-cancel` → `copy-pipe-no-clear "pbcopy"`.
- **Scrollback path intact** — root `WheelUpPane` and the copy-mode drag-start defaults remain after load, so wheel scrollback (#452) is preserved.

### Security Review
**Strengths:**
- No injection surface: `pbcopy` is a fixed literal in an app-internal, non-user-derived bundled config. No untrusted input flows into any binding.
- No change to `allow-passthrough` or other escape-sequence handling.

**Concerns:** None.

### Code Quality
- Comments are excellent and now match verified behavior. The double/triple-click rebind (commit `0ffd8a0`) correctly extends the fix to word/line selection for consistency with drag.
- Test assertions match the conf file exactly (spacing/alignment included).

**Note (non-blocking):** The Swift suite couldn't run in this review checkout — `Frameworks/GhosttyKit.xcframework` has no binary artifact here (environment, not a PR issue). The author's test plan reports all 6 `BundledResourcesTests` pass, and I independently confirmed the assertions hold against the actual file.

### Consider (Green)
- `copy-pipe-no-clear` intentionally leaves the pane in copy mode after selection so the highlight persists — the inherent tradeoff of `-no-clear`. The next keystroke goes to copy-mode until a click/Escape dismisses it. This is standard tmux behavior and is the desired outcome here; flagging only for awareness.
- The remaining unchecked boxes in the test plan (selection visually persists, Cmd+V pastes, vim/htop mouse) are runtime/visual and can't be verified statically — worth the quick manual smoke before merge, per the #450 reload caveat.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. The fix's central premise is empirically confirmed on tmux 3.6a.

PR #455, opened 2026-06-07
Complete TaskBackend/CodeBackend migration (#454)

### LINKED ISSUE #454: Complete TaskBackend/CodeBackend migration: expand surface, port IssueTracker call sites, close legacy escape-hatches (#410 follow-on)
**Context:** PR #411 (closing #410) shipped the foundation — `TaskBackend` + `CodeBackend` protocols, `ProviderManager` factory, three concrete `TaskBackend` implementations (GitHub, GitLab, Stub Corveil), two `CodeBackend` implementations (GitHub, GitLab), and URL-based routing including `corveil.io` detection. ADR 0005 captures the design.

What's missing: **most of the actual migration is still ahead.** The protocols cover ~5 methods total; the bulk of ticket/PR behavior still lives as direct `gh`/`glab` shell-outs in `IssueTracker.swift` (~60 grep hits for `shell|gh|glab` remain in a 2,777-line file) and `SessionService.swift`. The abstraction exists in spirit, but Crow's actual behavior bypasses it for everything except a handful of operations.

This ticket completes the migration so a new provider implementation (real Corveil backend, Linear, JIRA, etc.) actually replaces gh/glab end-to-end — not just for label edits.

---

### 1. Expand `TaskBackend` protocol surface

ADR 0005 (line 26) specifies the full surface: `fetchTask, listAssigned, setLabels, setTaskStatus, assign, createTask`. Currently shipped (`TaskBackend.swift:14-43`): only `fetchTask, setLabels, setTaskStatus`.

Add:
- `func listAssigned() async throws -> [AssignedIssue]` — replaces the GitHub batched `gh api graphql` at `IssueTracker.swift:703` and the parallel GitLab `glab api issues?scope=assigned_to_me` at `:2509`. The two halves of one logical operation finally live in their own backend files.
- `func assign(url: String, to login: String) async throws` — used by `setup.sh:420` (`gh issue edit --add-assignee @me`) and any future re-assignment flow.
- `func createTask(repo: String, title: String, body: String, labels: [String]) async throws -> TicketInfo` — for the `/crow-create-ticket` skill flow once it routes through the app (today it shells `gh` from the skill directly; that's a separate effort but the protocol method should exist).

### 2. Expand `CodeBackend` protocol surface

ADR 0005 (line 27) specifies `linkedPR, prStates, ensureMergeLabel, fetchCrowAuthoredCommits`. Currently shipped (`CodeBackend.swift`): only `linkedPR, ensureMergeLabel`.

Add:
- `func prStates(refs: [PRRef]) async throws -> [PRRef: PRState]` — replaces stale-PR polling at `IssueTracker.swift:860` (GitHub batched) and `:916` (GitLab per-MR).
- `func fetchCrowAuthoredCommits(...) async throws -> [Commit]` — the auto-respond/auto-merge surface needs a way to identify which commits Crow authored on a PR.

Add a `CodeCapability` enum (referenced in ADR 0005 line 29 but not yet defined). At minimum:
- `.autoMergeLabel` — GitHub has it (`gh label create` at `:2038`); GitLab doesn't.
- `.batchedPRStates` — GitHub's batched GraphQL vs GitLab's per-MR REST.

### 3. Migrate the remaining `IssueTracker.swift` call sites

Inventory (from the original #410 audit, lines confirmed against current HEAD):

| Call site | Operation | Routes to |
|---|---|---|
| `:493` updateLabels switch | `setLabels` | `taskBackend.setLabels(...)` (already migratable; verify done) |
| `:703` assigned issues (gh batched) | `listAssigned` | `GitHubTaskBackend.listAssigned` |
| `:2509` assigned issues (glab) | `listAssigned` | `GitLabTaskBackend.listAssigned` |
| `:860` stale PR states (gh) | `prStates` | `GitHubCodeBackend.prStates` |
| `:916` stale MR states (glab) | `prStates` | `GitLabCodeBackend.prStates` |
| `:2584` markInReview (gh project) | `setTaskStatus` | `GitHubTaskBackend.setTaskStatus` (close the escape-hatch — see §4) |
| `:2038` autoMergeLabel create | `ensureMergeLabel` | `CodeBackend.ensureMergeLabel` (capability-gated on `.autoMergeLabel`) |
| `:2015` `gh api` ad-hoc | (audit; likely fold into another method or remove) | TBD |

Plus `SessionService.swift`:
| `:1124` `gh issue view` recovery | `fetchTask` | `taskBackend.fetchTask(url:)` (likely already used; verify) |
| `:1139` `gh pr list --head` | `linkedPR` | `codeBackend.linkedPR(repo:branch:)` (likely already used; verify) |

Goal: after this ticket, **`grep -nE "\"gh\"|\"glab\"|gh api|glab api" Sources/Crow/App/IssueTracker.swift` should be empty** (or only contain comments). All provider-specific commands live in their respective backend files.

### 4. Close the `setTaskStatus` legacy escape-hatch

Today `GitHubTaskBackend.setTaskStatus` declares the `.projectBoardStatus` capability for UI gating but **throws `.unimplemented`** when called — the actual GraphQL mutation runs in `IssueTracker.markInReview` (`:2584-2599`). ADR 0005 (lines 39-43) flags this as a known caveat.

Move the GraphQL mutation into `GitHubTaskBackend.setTaskStatus`. Remove the now-dead `markInReview` path from `IssueTracker`. After this, capability flags actually mean what they say — declaring a capability and calling its method works without a back-channel.

### 5. Update ADR 0005

The ADR is currently `Status: Proposed` (line 3). Update to:

- **`Status: Accepted`** with a note linking to PR #411 (the foundation) and this ticket's PR (the migration).
- **Document what shipped vs what's pending** as of the foundation PR. The current ADR conflates the spec with the implementation — a future reader hitting `Status: Proposed` won't know the protocols already exist.
- **Add a "Migration completeness" or "Status" subsection** that explicitly tracks: which methods are on the protocol, which are stubbed/throw-unimplemented, and which legacy paths still bypass the abstraction. Update this with each PR that closes a piece.
- **Reconcile the spec/code drift** — the ADR lists `listAssigned`, `assign`, `createTask`, `prStates`, `fetchCrowAuthoredCommits` as part of the design but those methods don't exist on the protocols yet. Either: (a) leave them in the ADR as "to land in [this ticket]" with explicit pointers, or (b) trim the ADR to what shipped and amend separately. Recommend (a) — preserves the original design intent and shows the foundation PR was scoped down deliberately.
- **Note the `setTaskStatus` legacy escape-hatch closure** as one of the explicit deliverables of this ticket, so the ADR's caveat (lines 39-43) becomes a historical footnote rather than a live workaround.
- **Confirm no other ADRs need updating** — search `docs/adr/*.md` for references to `IssueTracker.markInReview`, direct `gh`/`glab` calls, or `Provider` enum switches that should now point at the backends.

### 6. Tests

Each backend method gets a unit test against a `FakeShellRunner` that records command vectors and returns canned JSON. Per ADR 0005 line 71, "the regression net the codebase currently lacks" — adding it is part of this ticket, not optional follow-up. Verify the migrated `IssueTracker` paths still behave identically by exercising them against the real backends with `FakeShellRunner` capturing the issued commands.

---

## Acceptance

- `TaskBackend` exposes the full surface from ADR 0005 line 26.
- `CodeBackend` exposes the full surface from ADR 0005 line 27.
- `CodeCapability` enum exists with at least `.autoMergeLabel` and `.batchedPRStates`.
- `IssueTracker.swift` contains zero direct `gh`/`glab` shell-outs (grep is the test).
- `SessionService.swift` ticket/PR fetches go through the backends.
- `GitHubTaskBackend.setTaskStatus` actually performs the GraphQL mutation (no `.unimplemented` throw); `IssueTracker.markInReview` is removed.
- `StubCorveilTaskBackend` still throws `.unimplemented` on every method — but now those methods cover the *expanded* protocol surface (proves the door is still open for the larger contract).
- Every backend method has a `FakeShellRunner`-based unit test.
- ADR 0005 is `Status: Accepted` with a Migration Status subsection and accurate references to the implementation PRs.
- Behavior parity: open a GitHub-backed session and a GitLab-backed session, exercise ticket fetch, assigned-issue panel, stale PR detection, label edits, status updates — all behave identically to pre-migration HEAD.

---

## Out of scope (explicit, so they're not bundled in)

- **Real Corveil `TaskBackend` implementation** — separate ticket; stub stays as the proof-of-abstraction.
- **`Session.codeProvider` field** — referenced in ADR 0005 line 64; needed for real cross-backend sessions; separate ticket.
- **`setup.sh` and skill migration off direct `gh`/`glab`** — flagged in ADR 0005 line 80 as a separate effort. Many tickets already exist in the agent-kind cluster (#421, #440) that touch this area; they should coordinate but not block on each other.
- **Plugin/registration extension point for adding a provider without editing the `Provider` enum** — out of scope; the enum + factory switch is fine for the foreseeable future.

---

## Related

- #410 — original proposal (closed, foundation merged via #411).
- ADR 0005 — `docs/adr/0005-task-and-code-backend-protocols.md`.
- Adjacent cluster (don't bundle, but coordinate): #421, #424, #427, #428, #431, #433, #440, #443, #447 — these touch agent-kind plumbing; this ticket touches provider plumbing. Different axes, same theme of "things hardcoded in places they shouldn't be."

dgershman:
Closes #454.

## Summary

- Expands `TaskBackend` and `CodeBackend` to the full ADR 0005 surface plus the operational primitives (`enableAutoMerge`, `updateBranch`, `findRecentPRsForBranches`, `listMonitoredPRs`, `fetchPRMetadata`) that the grep-empty acceptance test required.
- Closes the `setTaskStatus` escape-hatch: `GitHubTaskBackend.setTaskStatus` now runs the GraphQL project-board mutation directly; `IssueTracker.markInReview` is a thin capability gate that delegates to the backend.
- Ports every `gh`/`glab` shell-out in `IssueTracker.swift` through the backends — `rg '"gh"|"glab"|gh api|glab api' Sources/Crow/App/IssueTracker.swift` returns zero matches.
- Ports the ticket/PR fetch fallbacks in `SessionService.swift` (`parseTicketInfo`, `findPRLink`, `prepareReviewClone`) through the backends. Repo clones stay direct — outside the TaskBackend/CodeBackend surface.
- Adds 33 new `BackendsTests` cases against `FakeShellRunner` covering every new method.
- Updates ADR 0005 to `Accepted` with a Migration Status subsection documenting method-by-method coverage and the surface that grew beyond the original spec.

## Test plan

- [x] `swift build --arch arm64` — succeeds with no new warnings.
- [x] `swift test --package-path Packages/CrowProvider` — all 65 tests pass (BackendsTests + ProviderManagerTests).
- [x] `swift test --package-path Packages/{CrowCore,CrowGit,CrowPersistence,CrowIPC}` — all pass.
- [x] Grep test: `rg '"gh"|"glab"|gh api|glab api' Sources/Crow/App/IssueTracker.swift` returns zero matches.
- [ ] Manual smoke: open a GitHub-backed session and a GitLab-backed session in Crow; verify assigned-issue panel renders both, label add/remove works on issues, stale-PR badges flip correctly, mark-in-review flips the GitHub project board, and auto-merge fires when conditions are met. (Reviewer's call.)

## Out of scope (per ticket)

- Real Corveil `TaskBackend` (stub stays).
- `Session.codeProvider` field.
- `setup.sh` / skill scripts shelling `gh`/`glab` directly.
- `SessionService` repo clone operations.

### DIFF
--- a/Packages/CrowProvider/Sources/CrowProvider/BackendTypes.swift
+++ b/Packages/CrowProvider/Sources/CrowProvider/BackendTypes.swift
@@ -168,11 +168,23 @@ public struct AssignedListing: Sendable {
     public let closed: [AssignedIssue]
     /// GitHub-only; nil for GitLab.
     public let rateLimit: GitHubRateLimit?
+    /// When non-nil, the backend completed the call but had to degrade the
+    /// response because the OAuth token was missing this scope (e.g.
+    /// `read:project` on GitHub). Callers should surface a UI warning so the
+    /// user knows to refresh their token. The successful path returns the
+    /// best-effort data alongside the scope marker; no error is thrown.
+    public let missingScope: String?
 
-    public init(open: [AssignedIssue], closed: [AssignedIssue], rateLimit: GitHubRateLimit? = nil) {
+    public init(
+        open: [AssignedIssue],
+        closed: [AssignedIssue],
+        rateLimit: GitHubRateLimit? = nil,
+        missingScope: String? = nil
+    ) {
         self.open = open
         self.closed = closed
         self.rateLimit = rateLimit
+        self.missingScope = missingScope
     }
 }
 
--- a/Packages/CrowProvider/Sources/CrowProvider/BackendTypes.swift
+++ b/Packages/CrowProvider/Sources/CrowProvider/BackendTypes.swift
@@ -1,0 +1,193 @@
+import Foundation
+import CrowCore
+
+/// Identifies a single PR by its repo coordinates and number.
+///
+/// Used as the key for batched PR-state lookups (see `CodeBackend.prStates`).
+/// `owner` and `repo` together form the GitHub `org/repo` slug; on GitLab they
+/// form the path before the `-/merge_requests/{iid}` segment.
+public struct PRRef: Sendable, Hashable {
+    public let owner: String
+    public let repo: String
+    public let number: Int
+
+    public init(owner: String, repo: String, number: Int) {
+        self.owner = owner
+        self.repo = repo
+        self.number = number
+    }
+
+    public var slug: String { "\(owner)/\(repo)" }
+}
+
+/// Rich PR/MR record. Mirrors the union of fields needed across:
+///
+/// - the viewer's own open PRs (`CodeBackend.listMonitoredPRs`)
+/// - stale-PR follow-up (`CodeBackend.prStates`)
+/// - reconcile branch matches (`CodeBackend.findRecentPRsForBranches`)
+///
+/// Not every field is populated by every call — for example, `prStates` skips
+/// labels/checks because the stale-PR query doesn't fetch them. Callers merge
+/// records by URL using the `merge` helper to fill in gaps as more data
+/// arrives.
+public struct PRRecord: Sendable {
+    public let number: Int
+    public let url: String
+    public let state: String              // OPEN / MERGED / CLOSED (GitHub); normalized same for GitLab
+    public let mergeable: String          // MERGEABLE / CONFLICTING / UNKNOWN
+    public let mergeStateStatus: String   // BEHIND / BLOCKED / CLEAN / DIRTY / DRAFT / HAS_HOOKS / UNKNOWN / UNSTABLE
+    public let reviewDecision: String     // APPROVED / CHANGES_REQUESTED / REVIEW_REQUIRED / ""
+    public let isDraft: Bool
+    public let headRefName: String
+    public let headRefOid: String         // commit SHA, "" if unavailable
+    public let baseRefName: String
+    public let repoNameWithOwner: String
+    public let labels: [LabelInfo]
+    public let linkedIssueReferences: [LinkedIssueRef]
+    public let checksState: String        // SUCCESS / FAILURE / PENDING / EXPECTED / ERROR / ""
+    public let failedCheckNames: [String]
+    public let latestReviewStates: [String]
+    /// Used by reconcile tie-breaking when multiple non-OPEN PRs exist on the same branch.
+    public let updatedAt: Date?
+
+    public init(
+        number: Int,
+        url: String,
+        state: String,
+        mergeable: String = "UNKNOWN",
+        mergeStateStatus: String = "UNKNOWN",
+        reviewDecision: String = "",
+        isDraft: Bool = false,
+        headRefName: String = "",
+        headRefOid: String = "",
+        baseRefName: String = "",
+        repoNameWithOwner: String = "",
+        labels: [LabelInfo] = [],
+        linkedIssueReferences: [LinkedIssueRef] = [],
+        checksState: String = "",
+        failedCheckNames: [String] = [],
+        latestReviewStates: [String] = [],
+        updatedAt: Date? = nil
+    ) {
+        self.number = number
+        self.url = url
+        self.state = state
+        self.mergeable = mergeable
+        self.mergeStateStatus = mergeStateStatus
+        self.reviewDecision = reviewDecision
+        self.isDraft = isDraft
+        self.headRefName = headRefName
+        self.headRefOid = headRefOid
+        self.baseRefName = baseRefName
+        self.repoNameWithOwner = repoNameWithOwner
+        self.labels = labels
+        self.linkedIssueReferences = linkedIssueReferences
+        self.checksState = checksState
+        self.failedCheckNames = failedCheckNames
+        self.latestReviewStates = latestReviewStates
+        self.updatedAt = updatedAt
+    }
+}
+
+/// An issue reference linked from a PR/MR via "closes #N" or equivalent.
+public struct LinkedIssueRef: Sendable {
+    public let number: Int
+    public let repo: String
+
+    public init(number: Int, repo: String) {
+        self.number = number
+        self.repo = repo
+    }
+}
+
+/// A single commit on a PR/MR, used for Crow-author detection.
+public struct CommitInfo: Sendable {
+    public let sha: String
+    public let message: String
+
+    public init(sha: String, message: String) {
+        self.sha = sha
+        self.message = message
+    }
+}
+
+/// Input to `CodeBackend.findRecentPRsForBranches` — one (repo, branch) tuple.
+public struct BranchCandidate: Sendable, Hashable {
+    public let repoSlug: String
+    public let branch: String
+
+    public init(repoSlug: String, branch: String) {
+        self.repoSlug = repoSlug
+        self.branch = branch
+    }
+}
+
+/// One PR match returned from `findRecentPRsForBranches`, tagged with which
+/// candidate it came from so callers can route the result back to the right
+/// session.
+public struct BranchPRMatch: Sendable {
+    public let candidate: BranchCandidate
+    public let number: Int
+    public let url: String
+    public let state: String       // "OPEN" / "MERGED" / "CLOSED"
+    public let updatedAt: Date?
+
+    public init(candidate: BranchCandidate, number: Int, url: String, state: String, updatedAt: Date?) {
+        self.candidate = candidate
+        self.number = number
+        self.url = url
+        self.state = state
+        self.updatedAt = updatedAt
+    }
+}
+
+/// PR metadata returned from `CodeBackend.fetchPRMetadata` — the subset
+/// SessionService needs to prep a review clone.
+public struct PRMetadata: Sendable {
+    public let title: String
+    public let number: Int
+    public let headRefName: String
+    public let headRefOid: String
+    public let baseRefName: String
+
+    public init(title: String, number: Int, headRefName: String, headRefOid: String, baseRefName: String) {
+        self.title = title
+        self.number = number
+        self.headRefName = headRefName
+        self.headRefOid = headRefOid
+        self.baseRefName = baseRefName
+    }
+}
+
+/// Open + recently-closed issues assigned to the authenticated user.
+///
+/// Closed issues drive removal detection — IssueTracker diffs the new closed set
+/// against the prior open set to flush issues that left the user's queue.
+public struct AssignedListing: Sendable {
+    public let open: [AssignedIssue]
+    public let closed: [AssignedIssue]
+    /// GitHub-only; nil for GitLab.
+    public let rateLimit: GitHubRateLimit?
+
+    public init(open: [AssignedIssue], closed: [AssignedIssue], rateLimit: GitHubRateLimit? = nil) {
+        self.open = open
+        self.closed = closed
+        self.rateLimit = rateLimit
+    }
+}
+
+/// Viewer's own monitored PRs + review-requested PRs. Returned together because
+/// the GitHub GraphQL surface can fetch both in one batched call.
+public struct MonitoredPRListing: Sendable {
+    public let viewerPRs: [PRRecord]
+    public let reviewRequests: [ReviewRequest]
+    public let viewerLogin: String
+    public let rateLimit: GitHubRateLimit?
+
+    public init(viewerPRs: [PRRecord], reviewRequests: [ReviewRequest], viewerLogin: String, rateLimit: GitHubRateLimit? = nil) {
+        self.viewerPRs = viewerPRs
+        self.reviewRequests = reviewRequests
+        self.viewerLogin = viewerLogin
+        self.rateLimit = rateLimit
+    }
+}
--- a/Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubCodeBackend.swift
+++ b/Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubCodeBackend.swift
@@ -80,6 +80,15 @@ public struct GitHubCodeBackend: CodeBackend {
 
     // MARK: - prStates
 
+    /// `prStates` and `findRecentPRsForBranches` deliberately omit the
+    /// `rateLimit { remaining limit resetAt cost }` block their pre-migration
+    /// IssueTracker counterparts used to carry. The cycle's main consolidated
+    /// poll (`listAssigned` + `listMonitoredPRs`) already refreshes
+    /// `appState.githubRateLimit` every ~60s, so threading rate-limit out
+    /// of these secondary calls only sharpens the soft-threshold accounting
+    /// by a few requests of granularity. Re-add the block and a return-shape
+    /// tuple if that granularity ever starts mattering.
+
     public func prStates(refs: [PRRef]) async throws -> [PRRef: PRRecord] {
         guard !refs.isEmpty else { return [:] }
         var queryParts: [String] = []
--- a/Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubCodeBackend.swift
+++ b/Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubCodeBackend.swift
@@ -5,19 +5,28 @@ import CrowCore
 ///
 /// Capabilities:
 /// - `.autoMergeLabel` — supports `gh label create crow:merge`.
-/// - `.batchedPRStates` — could fetch multiple PR states in one GraphQL call.
+/// - `.batchedPRStates` — batches multiple PR states in one GraphQL call.
+/// - `.autoMerge` — supports `gh pr merge --auto --squash --delete-branch`.
+/// - `.updateBranch` — supports `gh pr update-branch`.
 ///
 /// See ADR 0005 for the protocol contract.
 public struct GitHubCodeBackend: CodeBackend {
     public let provider: Provider = .github
     public let cliName: String = "gh"
-    public let capabilities: Set<CodeCapability> = [.autoMergeLabel, .batchedPRStates]
+    public let capabilities: Set<CodeCapability> = [
+        .autoMergeLabel,
+        .batchedPRStates,
+        .autoMerge,
+        .updateBranch
+    ]
 
     private let shellRunner: ShellRunner
 
     public init(shellRunner: ShellRunner) {
         self.shellRunner = shellRunner
     }
+
+    // MARK: - linkedPR / ensureMergeLabel
 
     public func linkedPR(repo: String, branch: String) async throws -> LinkedPR? {
         let output = try await shellRunner.run(
@@ -40,9 +49,6 @@ public struct GitHubCodeBackend: CodeBackend {
     }
 
     public func ensureMergeLabel(repo: String) async throws {
-        // `gh label create` is idempotent-ish: it errors if the label already
-        // exists. Swallow that one error; surface anything else with the
-        // original exit code intact.
         do {
             _ = try await shellRunner.run(
                 "gh", "label", "create", "crow:merge",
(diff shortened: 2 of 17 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Substantial, well-structured migration: every `gh`/`glab` shell-out in `IssueTracker.swift` now routes through `TaskBackend`/`CodeBackend`, the `setTaskStatus` escape-hatch is closed, the ADR is updated, and 33 new `FakeShellRunner`-backed tests cover the new surface. Verified locally: the grep acceptance test returns zero matches and `swift test --package-path Packages/CrowProvider` passes (BackendsTests + ProviderManager). The app target could not be built here — it fails on a missing vendored binary (`GhosttyKit.xcframework`), an environmental issue, not a code issue.

One behavior regression and a few smaller items below.

### Critical Issues
None.

### Security Review
**Strengths:**
- No new shell-string interpolation of untrusted free-text. New backend calls pass argument vectors to `ShellRunner.run(args:)`; GraphQL values go through `gh ... -F name=value`, and GitLab path/branch segments are `addingPercentEncoding`-escaped before landing in REST endpoints.
- Typed `insufficientScope` / `rateLimited` errors avoid leaking raw stderr into the UI; secrets aren't logged.

**Concerns:**
- 🟢 **`sh -c` string interpolation in `enableAutoMerge` / `updateBranch`** (`GitHubCodeBackend.swift`): `let cmd = #"cd "$TMPDIR" && gh pr merge \#(prURL) --auto --squash --delete-branch"#; shellRunner.run(args: ["sh", "-c", cmd], ...)`. `prURL` is GitHub-sourced and canonical, and this is a faithful port of the prior `IssueTracker` code, so it's practically safe. But `ShellRunner.run` already takes a `cwd:` parameter — passing `cwd: NSTemporaryDirectory()` with a direct argument vector (`["gh","pr","merge",prURL,"--auto",...]`) drops the shell entirely and removes the injection surface. This is the natural moment to clean it up since the code is moving anyway.

### Code Quality
- 🟡 **Scope-warning regression on the polling path.** `GitHubTaskBackend.listAssigned` catches `ProviderError.insufficientScope` internally and silently retries with the no-projects query, returning `.success`. `runConsolidatedGitHubQuery` then sees success and calls `clearScopeWarning()` (`IssueTracker.swift:639`). Net effect: when the token lacks `read:project`, project-board statuses silently degrade to `.unknown` with **no** UI hint to add the scope, and any warning previously raised by the `markInReview` path (`IssueTracker.swift:1931`) gets cleared on the next ~60s poll. The old consolidated-query path called `reportScopeWarning("read:project")` *before* retrying. To preserve behavior, `listAssigned` should signal "I degraded due to missing scope" (e.g. surface the scope on `AssignedListing`, or re-throw/observe it) so the caller can `reportScopeWarning` instead of `clearScopeWarning`.

- 🟡 **Rate-limit observations dropped on stale-PR and reconcile calls.** The old `parseStalePRResponse` / `parseReconcilePRResponse` set `appState.githubRateLimit = rl` from each call's `rateLimit` block. The new `GitHubCodeBackend.prStates` and `findRecentPRsForBranches` still request `rateLimit { ... }` in their queries, but the parsers (`parseStalePRResponse`, `parseRecentPRsResponse`) discard it. The consolidated query still refreshes the limit each cycle, so impact is bounded, but the soft-threshold polling-suspend logic now runs on a slightly less current count, and the requested `rateLimit` block is dead weight. Either thread the rate limit back out (these methods could return it the way `listAssigned`/`listMonitoredPRs` do) or drop it from the queries.

- 🟢 **Result keying is more brittle than before.** In `fetchStalePRStates`, `prStates` returns a dict keyed by the API-returned `url`/`web_url`, but the caller looks it up by the *stored* session-link URL (`githubURLByRef[ref]` / `gitlabURLByRef[ref]`). If the stored URL differs from the canonical API URL (e.g. a GitLab deep link `.../merge_requests/3/diffs`, or any normalization drift), the record is silently dropped — whereas the old GitLab path appended directly with a fallback URL. Consider keying the returned map by `PRRef` rather than by URL string.

- 🟢 **Doc/behavior mismatch.** `CodeBackend.fetchCrowAuthoredCommits`'s doc comment says it fetches "every commit ... whose commit message carries a `Crow-Session:` trailer," but the implementation returns *all* commits and the caller filters (the test even asserts "Returns ALL commits — caller filters"). Tighten the doc to match.

- 🟢 **Redundant `Task { … }.value` inside `async let`** (`runConsolidatedGitHubQuery`): `async let` already provides the concurrency; wrapping each body in an extra `Task {}` adds an unnecessary hop. The `Result`-capturing pattern works, but `async let x = backend.foo()` with a `do/catch` at the await site is simpler.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 4 Green] findings. The two Yellows are behavior regressions in user-facing scope warnings and rate-limit accounting; both are small to address and worth landing in this round. The rest are non-blocking considerations.

---

dhilgaertner (changes requested):
## Code & Security Review

Large, well-structured migration. The new `TaskBackend`/`CodeBackend` surface in `CrowProvider` is clean, the call-site ports in `IssueTracker.swift` and `SessionService.swift` are faithful, comments are excellent, and `BackendsTests` cover every new method against `FakeShellRunner`. The grep-empty acceptance test holds (`rg '"gh"|"glab"|gh api|glab api' Sources/Crow/App/IssueTracker.swift` → 0 matches), and the remaining `gh/glab` shell-outs in `SessionService.swift` are the repo-clone paths explicitly declared out of scope. `swift test --package-path Packages/CrowProvider` passes (35 swift-testing + the XCTest `BackendsTests` suite).

I found one crash risk and one behavioral regression that should be fixed in this round.

### Critical Issues

**1. (Red — crash) Reconcile mapping traps on duplicate `(repoSlug, branch)` across sessions**

`Sources/Crow/App/IssueTracker.swift:1048` and `:1077` both build a candidate→session map with `Dictionary(uniqueKeysWithValues:)`:

```swift
let cands = candidates.map { BranchCandidate(repoSlug: $0.repoSlug, branch: $0.branch) }
let cmap: [BranchCandidate: UUID] = Dictionary(
    uniqueKeysWithValues: zip(cands, candidates.map(\.sessionID))
)
```

`BranchCandidate` is `Hashable` on `(repoSlug, branch)` only. `buildReconcileCandidates()` emits one candidate per session, so if two non-archived **work** sessions without a PR link share the same repo slug and branch, `cands` contains duplicate keys and `Dictionary(uniqueKeysWithValues:)` **traps** (`Fatal error: Duplicate values for key`). This runs on the `@MainActor` 60s poll, so it takes down the app.

This is a regression: the pre-migration code threaded `sessionID` through each indexed `parsed` entry (`ReconcileBranchMatch(sessionID: p.cand.sessionID, …)`), so duplicate branch/slug pairs produced duplicate matches rather than a crash. Two sessions on the same branch isn't exotic (duplicated session, or reconcile firing before the first session's PR link lands).

Fix: tolerate duplicates, e.g. `Dictionary(zip(cands, sessionIDs), uniquingKeysWith: { a, _ in a })`, or carry the `sessionID` per index back from the backend instead of keying by `BranchCandidate`. Note that collapsing duplicates to one `sessionID` also silently drops the PR link for the other session, so threading sessionID through is the more correct fix.

### Code Quality

**2. (Yellow — regression) `setTaskStatus` drops the "Review" project-column alias**

`GitHubTaskBackend.setTaskStatus` resolves the board option via `resolveProjectFieldOption(queryOut, target: status.rawValue)` (`GitHubTaskBackend.swift:135`), which matches `option.name.lowercased() == target.lowercased()` exactly. For `.inReview`, `rawValue == "In Review"`, so a project whose Status column is named just **"Review"** no longer matches.

The old `IssueTracker.markInReview` matched `normalized == "in review" || normalized == "review"`, and `TicketStatus(projectBoardName:)` (`Enums.swift:87`) deliberately treats `"review"` as an alias for `.inReview`. So the codebase intends to support the bare "Review" column; the new exact-match silently breaks it (`setTaskStatus` throws `.unimplemented` → `markInReview` logs and returns, status never flips). `testGitHubTaskBackendSetTaskStatusRunsMutation` only exercises an option literally named "In Review", so the regression is untested.

Fix: match option names back through `TicketStatus(projectBoardName:)` (i.e. find the option whose `TicketStatus(projectBoardName: option.name) == status`) rather than comparing against `rawValue` verbatim.

### Security Review

**Strengths:**
- Command injection surface reduced, not expanded: every new shell-out uses direct argv (`shellRunner.run(args:)`) — no `sh -c`. The migration explicitly replaces the prior `#"cd "$TMPDIR" && gh pr merge \#(pr.url) …"#` string-interpolated `sh -c` with `args: ["gh", "pr", "merge", prURL, …], cwd: NSTemporaryDirectory()`, eliminating the interpolation surface around `prURL`.
- GraphQL inputs are passed as bound variables (`-F owner=…`), not interpolated into the query body.
- GitLab slugs/branches are percent-encoded before going into REST paths.
- `GITLAB_HOST` is pinned per host through the backend `env()`, preserving the prior token-selection behavior.

No new secret handling, auth, or injection concerns.

### Minor (Green — consider)

- **Rate-limit observability narrowed.** The new `GitHubCodeBackend.prStates` and `findRecentPRsForBranches` GraphQL queries omit the `rateLimit { … }` block that the old inline stale-PR/reconcile queries carried (and which fed `appState.githubRateLimit`). The main consolidated poll still updates it every 60s, so this is just a small loss of granularity on those secondary calls — but worth a comment or re-adding the block if the quota accounting matters.

### Verification note

The main app target could not be compiled/tested in this review environment — `swift build` fails on a missing binary artifact (`Frameworks/GhosttyKit.xcframework`), unrelated to this PR. The `CrowProvider` package tests pass; the `IssueTracker`/`SessionService` Swift changes were reviewed by reading but not compiled here. The PR's own test plan reports `swift build --arch arm64` and the full test suites green.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 1 Yellow, 1 Green] findings. The Red is a `@MainActor` crash on a reachable edge case; the Yellow silently regresses bare-"Review" project boards.

---

dhilgaertner (changes requested):
## Code & Security Review

Round 3 of the TaskBackend/CodeBackend migration. The bulk of this is excellent: GraphQL queries/parsers moved verbatim into the backends, every value is passed through `gh`/`glab` `-F`/`-f` variables (no GraphQL/shell injection surface), the reconcile fan-out correctly handles two sessions on one branch (with regression tests), and the `setTaskStatus` column matching faithfully preserves the old "in review"/"review" semantics via `TicketStatus(projectBoardName:)` (verified against `Enums.swift:82-90`). Notably, the old `sh -c "cd … && gh pr merge \(pr.url)"` string-interpolation surface is gone in favor of direct argv — a genuine security improvement.

`CrowProvider` package tests pass locally (BackendsTests + 35 ProviderManager tests). The full app target can't be built here (missing vendored `GhosttyKit.xcframework` — environment limitation, not a PR defect), so the app-target compile is relied on via CI.

The findings below are cleanups left behind by the migration, not logic bugs — but two of them produce permanently-misleading runtime output, so per Crow's verdict rule (any Yellow → request changes) I'm requesting changes to land them in this round.

### Security Review
**Strengths:**
- All backend shell-outs use argv arrays, not `sh -c`; the prior `pr.url` interpolation into a shell string is eliminated.
- GraphQL aliased queries interpolate only loop indices (`pr0`, `pr1`, …); all caller-controlled values go through GraphQL variables.
- `enableAutoMerge`/`updateBranch` run in `NSTemporaryDirectory()` so `gh` can't pick up the cwd's git config — behavior preserved from the old `cd "$TMPDIR"`.

**Concerns:** none.

### Code Quality

**Yellow — Dead async shell helpers left behind** (`Sources/Crow/App/IssueTracker.swift:2035-2095`)
After this migration nothing calls `shell(...)` or `shellWithStatus(...)` anymore — the only references are the overloads delegating to each other (`:2036`, `:2086`, `:2090`). Swift does not warn on unused `private` methods, so this dead plumbing lingers silently. The migration's stated goal was to fully port the shell-outs out of `IssueTracker`; leaving the helper infrastructure behind undercuts that. (`shellSync` at `:1224` is still legitimately used at `:1162` for `git remote get-url` — leave that.)

**Yellow — `currentRefreshGhCalls` is now dead, lying telemetry** (`IssueTracker.swift:174`, `:310`, `:526`/`:528`, `:2040`/`:2094`)
The counter is reset to `0` at `:310` and only ever incremented inside the now-dead `shell`/`shellWithStatus` helpers. So `logRefreshSummary` will now *always* print `refresh: 0 gh calls in …` regardless of actual backend activity. Either remove the counter and that field from the log line, or thread the count through the backend layer so it reflects reality.

**Yellow — GitLab `fetchGitLabIssues` now fires a wasted REST call every poll** (`IssueTracker.swift:~1971` `fetchGitLabIssues` → `GitLabTaskBackend.listAssigned`)
`fetchGitLabIssues` calls `backend.listAssigned()`, which issues **both** the open *and* the closed `glab api issues?…` REST calls, then returns only `.open` (the closed half is discarded — the inline comment even says "Only open issues, to match the prior single-call semantics"). Pre-migration this path was a single open-only call. On a 60s polling loop, every GitLab host now does an extra closed-issues fetch whose result is thrown away each cycle. Consider an open-only backend entry point (or have `listAssigned` skip the closed call when the caller doesn't consume it).

### Consider (Green)
- `prStates`/`findRecentPRsForBranches` intentionally drop the `rateLimit` block their pre-migration counterparts carried; documented tradeoff (`GitHubCodeBackend.swift` comment). Fine — flagging only so it's a conscious decision.
- Numerous `codeBackend(for: .github)!` / `taskBackend(for: .github)` force-unwraps assume the GitHub backend always exists. Consistent with the existing codebase pattern, so acceptable, but a guarded fallback would be marginally safer.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 3 Yellow, 2 Green] findings. All three Yellows are migration-leftover cleanups (two of which make `IssueTracker`'s refresh log permanently inaccurate); none are logic bugs, and the core migration is sound.

dhilgaertner (approved):
## Code & Security Review

Fourth-round review of the TaskBackend/CodeBackend migration (#454). Verified locally and read end-to-end.

### Critical Issues
None.

### Security Review
**Strengths:**
- Auto-merge / update-branch moved from `sh -c "cd \"$TMPDIR\" && gh pr merge \(prURL) …"` string interpolation to **direct argv** with `cwd: NSTemporaryDirectory()` — eliminates the shell-interpolation surface around `prURL` (`GitHubCodeBackend.enableAutoMerge`/`updateBranch`). Net hardening over the prior code.
- All GraphQL inputs (owner/repo/branch/number, review query) are passed as typed `-F`/`-f` variables, not interpolated into query bodies. Aliased field names (`pr0`, `owner0`) derive from loop indices only.
- GitLab REST slugs/branches are percent-encoded before being placed in endpoint paths.
- No secrets logged — error paths surface `.localizedDescription`/stderr prefixes only, matching the pre-migration pattern.

**Concerns:** None.

### Code Quality
- **CrowProvider package builds and all 71 tests pass** (36 BackendsTests + 35 ProviderManagerTests) — this is where the bulk of the new logic lives. 33 new backend cases cover every new method against `FakeShellRunner`.
- **Acceptance test holds:** `rg '"gh"|"glab"|gh api|glab api' Sources/Crow/App/IssueTracker.swift` → 0 matches. No dangling references to removed symbols (`shellWithStatus`, `currentRefreshGhCalls`, `parseConsolidatedResponse`, etc.).
- **`setTaskStatus` escape-hatch genuinely closed** — `GitHubTaskBackend.setTaskStatus` now runs the project-board lookup + `updateProjectV2ItemFieldValue` mutation directly; `IssueTracker.markInReview` is a thin capability gate. Capability flags now mean "implemented," matching the ADR update.
- Behavioral edges checked and cleared: (1) stale-PR `complete` flag — a missing PR can't trigger a false auto-completion regardless, so GitLab's per-MR error swallowing is consistent with the documented "partial-success allowed" intent; (2) heavy JSON parsing stays off the main actor via nonisolated `async` backend methods (#304 preserved); (3) review-clone's `.github` hardcoding is a pre-existing assumption, and `providerManager` is always wired in production (AppDelegate:470/630); (4) dropping the rate-limit block from secondary stale/reconcile calls is a deliberate, documented trade.

Consider-only (non-blocking): `consolidatedIssuesQuery` and `consolidatedIssuesQueryNoProjects` are near-identical full literals; the explicit form is a reasonable clarity choice over the old `replacingOccurrences` derivation.

Note: the main app target couldn't be linked locally (missing `GhosttyKit.xcframework` binary — environment limitation); CI and the PR's own `swift build --arch arm64` cover it.

### Summary Table

---

PR #457, opened 2026-06-08
Re-arm auto-respond on new review activity + post-commit push (CROW-456)

### LINKED ISSUE #456: Auto-respond misses second+ rounds of review: dedup key is bucket-only, doesn't re-arm on new review activity or post-commit push
**Symptom:** After Crow's first auto-respond to a `changes-requested` review, subsequent rounds of review on the same PR are silently dropped. The agent never gets nudged unless the user manually pokes it ("can you re-read the new review comments and address them"). This forces a manual hand-holding loop that auto-respond was supposed to eliminate.

**Root cause — `IssueTracker.swift:1255-1297`:** the transition detector is **edge-triggered on `reviewStatus` bucket only**, with a re-arm rule that requires the PR to leave `changesRequested` before it can re-enter.

```swift
// IssueTracker.swift:1268
if old.reviewStatus == .changesRequested && newStatus.reviewStatus != .changesRequested {
    emittedTransitionKeys.remove("\(session.id.uuidString)|changesRequested")
}
```

As long as the PR sits in `changesRequested`, the dedup key stays set; no further transitions get emitted; no auto-respond fires. **Every realistic "round 2 of review" path stays in `changesRequested`**, so the trigger never re-arms.

**Three concrete failure modes:**

| Reviewer action after round 1 | What `reviewStatus` does | Re-arms? |
|---|---|---|
| Submits another formal "Request changes" review | `changesRequested` → `changesRequested` (no transition) | ❌ |
| Leaves additional inline comments without a formal review submission | doesn't move at all (only formal submissions update the bucket) | ❌ |
| Agent pushes a fix, but reviewer hasn't re-approved/dismissed yet, then reviewer requests more changes | `changesRequested` → `changesRequested` | ❌ |
| Reviewer approves, then requests changes again (the only path that works today) | `changesRequested` → `approved` → `changesRequested` | ✓ |

The last row is the **only** path that re-fires — and it's the least common pattern. Reviewers rarely approve-then-re-request-changes; they iterate within the changes-requested state.

**Secondary issue:** `emittedTransitionKeys` and `previousPRStatus` are **in-memory only** (`:158` "across the in-process lifetime"). Crow restarts wipe them, which on the next poll either (a) spuriously re-fires transitions for PRs that already got an auto-respond, or (b) loses the re-arm-state mid-sequence. Neither is great, but compounds the primary bug.

---

## Fix

Replace the bucket-only dedup key with one keyed on **monotonically advancing review activity**, and add a **post-commit re-arm trigger** so pushing a fix re-opens the auto-respond window.

### 1. Dedup on latest review identifier, not bucket

Change the dedup key at `IssueTracker.swift:1286` from:

```
"\(sessionID)|changesRequested"
```

to:

```
"\(sessionID)|changesRequested|\(latestReviewID)"
```

Where `latestReviewID` is the GitHub review node ID (or GitLab equivalent) of the most recent `CHANGES_REQUESTED` review on the PR. New review → new ID → new dedup slot → auto-respond fires. Old reviews stay deduped.

For the GitHub `gh api graphql` viewer-PR query (`:703` area), include the latest review's `id` and `submittedAt` in the projection. For GitLab, the `glab api projects/{slug}/merge_requests/{iid}/approvals` payload exposes equivalent fields.

If we want to be more conservative and dedup on a hash of all formal-review IDs the agent has addressed: also workable; pick whichever is cheaper to track.

### 2. Re-arm on post-commit push (head SHA change)

Add a re-arm rule near `:1272` analogous to the existing `checksPass` re-arm:

```swift
if old.reviewStatus == .changesRequested
   && old.headSha != newStatus.headSha {
    // Agent pushed a new commit after a changes-requested review.
    // Reviewer's next look should re-trigger auto-respond regardless
    // of whether the bucket moves.
    emittedTransitionKeys.remove("\(session.id.uuidString)|changesRequested")
}
```

This handles the "agent pushed a fix; reviewer comes back with more comments" case even if the bucket never leaves `changesRequested`.

### 3. Inline-comments path (optional, harder)

Inline review comments don't move `reviewStatus`. If we want auto-respond to fire on additional inline comments after the first formal review, we need a separate signal — comment count, latest comment timestamp, or fetching the per-PR review thread.

Recommend **deferring** unless it's a frequent pain point: the GitHub data is in the viewer-PR query but expanding the projection costs an extra GraphQL field every poll. Item 1 + item 2 fix the dominant case; item 3 is a v2 if the user still reports gaps.

### 4. Persist dedup + status state across restarts

Move `emittedTransitionKeys` and `previousPRStatus` to disk. Sketch:

- Add `IssueTrackerState.json` (or similar) under whichever persistent dir Crow uses for its own state (NOT inside a worktree — outlives them).
- Serialize keyed by session UUID. On load, drop entries for sessions that no longer exist.
- Write on transition emit (cheap; only when state changes).

Eliminates the "restart spuriously re-fires" path and means our new SHA-keyed re-arm survives an app relaunch.

---

## Tests

Add coverage in the `IssueTracker` test target:

- Round 2 of changes-requested without an intervening approval → auto-respond fires.
- Agent pushes a commit while in `changesRequested`, reviewer leaves another "Request changes" review → auto-respond fires.
- Same review observed twice (same review ID) → auto-respond fires **once** (dedup still works for genuine duplicates).
- Approve → request-changes-again cycle still fires (don't regress today's working path).
- Restart simulation: persist + reload state; ensure pending dedup keys survive and new transitions still emit correctly.

---

## Acceptance

- A reviewer doing two consecutive "Request changes" submissions on the same PR (no intervening approval) triggers two auto-respond prompts to the agent.
- After the agent pushes a fix, the next observed `changes-requested` review fires auto-respond regardless of whether the bucket moved.
- The same review (same review ID) observed on consecutive polls still fires auto-respond exactly once — dedup is preserved for genuine duplicates.
- Crow restart doesn't re-fire an already-handled transition or lose the re-arm trail.
- All four test scenarios above pass.

## Related

- `Sources/Crow/App/IssueTracker.swift:1255-1297` — primary site.
- `Sources/Crow/App/AutoRespondCoordinator.swift` — downstream consumer; no changes required there.
- ADR 0005 / #454 (TaskBackend migration) — when the GraphQL projection moves into `GitHubTaskBackend.listAssigned` per the migration, the review-identifier field needs to come along too. Worth a heads-up in #454's PR.

dgershman:
## Summary

Closes #456.

- Round 2+ of `CHANGES_REQUESTED` reviews now re-fire auto-respond. The dedup key shifts from bucket-only `"sessionID|changesRequested"` to `"sessionID|changesRequested|reviewID|headSha"`, and `PRStatus.transitions()` emits a candidate whenever the bucket is reentered, the latest CHANGES_REQUESTED review id rotates, or the head SHA changes while still in the bucket.
- GitHub viewer-PR GraphQL projection now includes `latestReviews(first: 5) { nodes { id state submittedAt } }`; the parser threads the most recent CHANGES_REQUESTED node id through `PRRecord` → `PRStatus` → `PRStatusTransition`.
- `IssueTracker` state (`previousPRStatus` + `emittedTransitionKeys`) persists to `JSONStore` so restarts no longer re-fire already-handled transitions or lose the re-arm trail. New optional `StoreData.issueTrackerState` field is backward-compatible with older `store.json`.
- GitLab path leaves `latestReviewID = nil` (provider doesn't surface review ids in the monitored-MR query) — no regression, no GitLab-side improvement in this PR.

## Heads-up for #454 / ADR 0005

The TaskBackend migration is moving the GraphQL projection. The new `latestReviews { id state submittedAt }` field needs to land in the destination so review-id dedup keeps working post-migration.

## Test plan

- [x] `cd Packages/CrowCore && arch -arm64 swift test` (233 tests, includes 3 new CROW-456 cases + replaced bug-locking dedup-key test)
- [x] `cd Packages/CrowPersistence && arch -arm64 swift test` (28 tests, includes new `issueTrackerStateRoundTripsThroughDisk` and `storeDataDecodesWithoutIssueTrackerStateField`)
- [x] `cd Packages/CrowProvider && arch -arm64 swift test` (35 tests)
- [x] `arch -arm64 swift test --arch arm64` from root (191 tests, includes 2 new `mergePRRecords` latestReviewID forwarding tests)
- [ ] Manual end-to-end: open a Crow session against a real PR, request changes via reviewer, wait for auto-respond, request changes a second time without approving → second prompt fires.
- [ ] Manual restart smoke: after a fire, quit Crow, relaunch; the next poll must not re-fire the same transition.

🐦‍⬛ [Created with Claude Code, orchestrated by Crow](https://github.com/radiusmethod/crow)

Crow-Session: 790C6352-0086-4B1C-85E0-34F26866E0A3

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Models/PRStatus.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/PRStatus.swift
@@ -9,19 +9,28 @@ public struct PRStatus: Codable, Sendable, Equatable {
     /// Head commit SHA. Used to dedupe per-commit transition events
     /// (e.g. don't re-fire "checks failing" when the same commit is re-run).
     public var headSha: String?
+    /// Node ID of the most recent CHANGES_REQUESTED review on this PR, when
+    /// the provider surfaces stable review identifiers. Round-2 dedup uses
+    /// this so a fresh "Request changes" submission re-arms auto-respond
+    /// even when the bucket stays in `.changesRequested`. `nil` when no
+    /// CHANGES_REQUESTED review is currently visible, or on providers that
+    /// don't expose review IDs in the monitored-PR query (e.g. GitLab).
+    public var latestReviewID: String?
 
     public init(
         checksPass: CheckStatus = .unknown,
         reviewStatus: ReviewStatus = .unknown,
         mergeable: MergeStatus = .unknown,
         failedCheckNames: [String] = [],
-        headSha: String? = nil
+        headSha: String? = nil,
+        latestReviewID: String? = nil
     ) {
         self.checksPass = checksPass
         self.reviewStatus = reviewStatus
         self.mergeable = mergeable
         self.failedCheckNames = failedCheckNames
         self.headSha = headSha
+        self.latestReviewID = latestReviewID
     }
 
     public init(from decoder: Decoder) throws {
@@ -31,10 +40,11 @@ public struct PRStatus: Codable, Sendable, Equatable {
         mergeable = try c.decodeIfPresent(MergeStatus.self, forKey: .mergeable) ?? .unknown
         failedCheckNames = try c.decodeIfPresent([String].self, forKey: .failedCheckNames) ?? []
         headSha = try c.decodeIfPresent(String.self, forKey: .headSha)
+        latestReviewID = try c.decodeIfPresent(String.self, forKey: .latestReviewID)
     }
 
     private enum CodingKeys: String, CodingKey {
-        case checksPass, reviewStatus, mergeable, failedCheckNames, headSha
+        case checksPass, reviewStatus, mergeable, failedCheckNames, headSha, latestReviewID
     }
 
     public enum CheckStatus: String, Codable, Sendable {
--- a/Packages/CrowCore/Sources/CrowCore/Models/PRStatusTransition.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/PRStatusTransition.swift
@@ -7,11 +7,10 @@ import Foundation
 /// so the comparison logic stays unit-testable.
 public struct PRStatusTransition: Sendable, Equatable {
     public enum Kind: String, Sendable, Equatable {
-        /// A reviewer transitioned the PR into "changes requested", or a new
-        /// round of changes-requested activity occurred while the PR was still
-        /// in `.changesRequested` (new formal review, or new commit pushed
-        /// after the prior review). See `transitions(from:to:…)` for the
-        /// round-2 detection rules.
+        /// A reviewer transitioned the PR into "changes requested", or a fresh
+        /// formal "Request changes" review was submitted while the PR was
+        /// already in `.changesRequested` (round-2; identified by review-id
+        /// rotation). See `transitions(from:to:…)` for the detection rules.
         case changesRequested
         /// At least one CI/CD check newly transitioned to failing on the
         /// current head commit.
@@ -23,14 +22,16 @@ public struct PRStatusTransition: Sendable, Equatable {
     public let prURL: String
     public let prNumber: Int?
     /// Head commit SHA at the moment of the transition. Part of the dedupe key
-    /// for both `.checksFailing` (re-runs on the same commit must not re-fire)
-    /// and `.changesRequested` (a post-commit push while still in
-    /// changes-requested must re-arm — CROW-456).
+    /// for `.checksFailing` so re-runs on the same commit don't re-fire.
+    /// Carried on `.changesRequested` for downstream telemetry/UI but is not
+    /// part of that dedupe key (see `dedupeKey`).
     public let headSha: String?
     /// Node ID of the most recent CHANGES_REQUESTED review when the transition
-    /// fired. Part of the `.changesRequested` dedupe key so a second formal
-    /// "Request changes" submission re-arms auto-respond even when the bucket
-    /// hasn't moved. `nil` for providers that don't expose review IDs.
+    /// fired. Sole round-2 discriminator in the `.changesRequested` dedupe
+    /// key — a fresh formal review rotates this and re-arms auto-respond,
+    /// while the author's own response push (which doesn't dismiss the
+    /// reviewer's CHANGES_REQUESTED decision) leaves it alone and stays
+    /// deduped. `nil` for providers that don't expose review IDs.
     public let latestReviewID: String?
     /// Names of failing checks (empty for `.changesRequested`).
     public let failedCheckNames: [String]
@@ -55,18 +56,18 @@ public struct PRStatusTransition: Sendable, Equatable {
 
     /// Stable key used to suppress duplicate fires across polling cycles.
     ///
-    /// `.changesRequested` includes both the latest CHANGES_REQUESTED review
-    /// ID and the head SHA, so:
-    /// - the same review observed twice → same key → deduped
-    /// - a new formal "Request changes" → new review ID → fires
-    /// - a new commit after the prior review → new SHA → fires
+    /// `.changesRequested` keys on `(session, kind, latestReviewID)`. The
+    /// review ID rotates only when a reviewer submits a new formal review —
+    /// not when the author pushes a fix — so the auto-respond prompt that
+    /// instructs the agent to "commit the fix, push, re-request review"
+    /// can't trigger itself on the next poll (CROW-456 review feedback).
     ///
     /// `.checksFailing` keys on `(session, kind, headSha)` so a new commit
     /// that also fails CI can re-fire.
     public var dedupeKey: String {
         switch kind {
         case .changesRequested:
-            return "\(sessionID.uuidString)|changesRequested|\(latestReviewID ?? "")|\(headSha ?? "")"
+            return "\(sessionID.uuidString)|changesRequested|\(latestReviewID ?? "")"
         case .checksFailing:
             return "\(sessionID.uuidString)|checksFailing|\(headSha ?? "")"
         }
@@ -80,11 +81,14 @@ extension PRStatus {
     /// existing PR state never triggers a fire-on-startup. Otherwise emits
     /// at most one `.changesRequested` and one `.checksFailing`, in that order.
     ///
-    /// For `.changesRequested`, fires on any of:
-    /// - Entering the bucket (old wasn't in `.changesRequested`).
-    /// - Same bucket, new review ID (a second formal "Request changes").
-    /// - Same bucket, new head SHA (agent pushed a fix after the prior
-    ///   review; reviewer's next pass should re-prompt the agent).
+    /// `.changesRequested` fires when the bucket is entered (`old` wasn't
+    /// `.changesRequested`) or when the latest CHANGES_REQUESTED review ID
+    /// rotates while still in the bucket (a fresh formal "Request changes"
+    /// submission). A head-SHA change alone does **not** fire — `reviewDecision`
+    /// stays `CHANGES_REQUESTED` after the author pushes, so any such trigger
+    /// would re-fire on the agent's own response push and create an auto-respond
+    /// loop. Real "round 2 after a fix" always involves the reviewer submitting
+    /// a new formal review, which rotates the review ID.
     ///
     /// - Note: This is the pure piece of the transition pipeline; the
     ///   stateful dedupe (across polls) lives in `IssueTracker`.
@@ -104,10 +108,7 @@ extension PRStatus {
             let newReview = old.latestReviewID != nil
                 && new.latestReviewID != nil
                 && old.latestReviewID != new.latestReviewID
-            let postCommitPush = old.reviewStatus == .changesRequested
-                && old.headSha != new.headSha
-                && new.headSha != nil
-            if entering || newReview || postCommitPush {
+            if entering || newReview {
                 out.append(PRStatusTransition(
                     kind: .changesRequested,
                     sessionID: sessionID,
--- a/Packages/CrowCore/Sources/CrowCore/Models/PRStatusTransition.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/PRStatusTransition.swift
@@ -7,7 +7,11 @@ import Foundation
 /// so the comparison logic stays unit-testable.
 public struct PRStatusTransition: Sendable, Equatable {
     public enum Kind: String, Sendable, Equatable {
-        /// A reviewer transitioned the PR into "changes requested".
+        /// A reviewer transitioned the PR into "changes requested", or a new
+        /// round of changes-requested activity occurred while the PR was still
+        /// in `.changesRequested` (new formal review, or new commit pushed
+        /// after the prior review). See `transitions(from:to:…)` for the
+        /// round-2 detection rules.
         case changesRequested
         /// At least one CI/CD check newly transitioned to failing on the
         /// current head commit.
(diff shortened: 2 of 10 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Solid, well-tested change overall — the composite dedup key, the cross-restart persistence with backward-compatible decoding, and the `mergePRRecords` forwarding are all clean and thoroughly covered. One interaction concern blocks approval.

### Critical Issue

**`postCommitPush` re-fires on the agent's *own* response push → auto-respond loop / reviewer-notification spam** (`Packages/CrowCore/Sources/CrowCore/Models/PRStatusTransition.swift:107-109`)

The new SHA-change trigger fires `.changesRequested` whenever the head SHA moves while the PR stays in the changes-requested bucket. But the transition it drives ends in `AutoRespondCoordinator` → `AutoRespondPrompts.build`, whose `.changesRequested` prompt instructs the agent to *\"commit the fix, and push so the PR updates\"* and then *\"re-request review from each reviewer\"* (`Sources/Crow/App/AutoRespondCoordinator.swift:156`).

On standard GitHub, a `CHANGES_REQUESTED` review decision is **not** dismissed when the author pushes new commits (branch protection only auto-dismisses *approvals*), so `reviewDecision` stays `CHANGES_REQUESTED` after the agent's push (`Sources/Crow/App/IssueTracker.swift:1712-1718`). That closes a loop:

1. Reviewer requests changes on `shaA` → `entering` fires → agent prompted.
2. Agent addresses, commits, pushes `shaB`, re-requests reviewers.
3. Next 60s poll (`pollInterval`, `IssueTracker.swift:22`): `old=(changesRequested, shaA, R_1)`, `new=(changesRequested, shaB, R_1)` → `postCommitPush` is true, new dedupe key `…|R_1|shaB` isn't in the emitted set → **fires again**.
4. Agent re-prompted; if it makes any further commit (`shaC`) it loops; either way it re-requests reviewers again every cycle.

The bucket-exit cleanup at `IssueTracker.swift:1281-1285` only clears keys when *leaving* `.changesRequested`, which never happens in this scenario, so nothing breaks the cycle. Minimum harm: a redundant re-prompt + redundant reviewer re-request notification after every legitimate response push. Maximum harm: a self-sustaining auto-respond loop.

The `postCommitPush` mechanism can't distinguish a push the reviewer is reacting to from the push auto-respond itself just told the agent to make. Some options to consider:
- Gate `postCommitPush` on the latest review ID also having rotated (i.e. require a *fresh reviewer action*, not just any SHA change) — which is really what \"reviewer's next pass should re-prompt\" (per the test comment) means.
- Or record the SHA(s) auto-respond has already pushed-in-response-to and suppress re-fire for those.
- Or drop `postCommitPush` entirely and rely on review-ID rotation (`newReview`) as the round-2 signal, since a genuine new reviewer pass produces a new review ID anyway.

The unit test `changesRequestedSameStateNewShaFires` validates the pure function in isolation but doesn't model this closed loop with the push-instructing prompt — that interaction is where the bug lives.

### Security Review
**Strengths:**
- `JSONStore` retains `0o600` perms on the store file; no new external input is trusted.
- Backward-compatible optional decoding (`storeDataDecodesWithoutIssueTrackerStateField`) avoids the corrupt-store backup path on upgrade.
- Review session gate (`shouldSkipReviewSession`) still prevents auto-respond from acting on PRs under review.

No new security concerns.

### Code Quality
- **Persist-on-every-poll cost** (`IssueTracker.swift:1313`, `persistTrackerState`): `persistTrackerState()` runs on every poll regardless of whether tracker state changed, and each call constructs a fresh `JSONStore()` (full decode of `store.json`) then re-encodes + atomically writes the entire store. The \"`mutate` coalesces writes so this is cheap\" comment is misleading — coalescing only drops *racing* stale snapshots; a lone periodic write still does a full read+encode+write every 60s. Consider only persisting when `previousPRStatus`/`emittedTransitionKeys` actually changed this poll. (Yellow — efficiency, consistent with the pre-existing fresh-`JSONStore` pattern at lines 904/1146/1483, so not a regression.)
- `emittedTransitionKeys` now grows by one entry per head SHA while in the changes-requested bucket; bounded by leaving the bucket, but the loop above would make it grow unboundedly in practice — another reason to resolve the critical issue. (Green, contingent on the fix.)

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 1 Yellow, 1 Green] findings.

---

dhilgaertner (changes requested):
## Code & Security Review

Solid, well-reasoned change. The central idea — keying the `.changesRequested` dedup on the latest CHANGES_REQUESTED **review id** rather than the bucket or head SHA — correctly threads the needle: the author's own response push (which doesn't dismiss the reviewer's decision) leaves the id alone and stays deduped, while a fresh formal "Request changes" rotates the id and re-arms. The pure transition logic is cleanly separated and thoroughly unit-tested (verified locally: `CrowCore` builds, all 15 `PRStatusTransitionTests` pass).

### Security Review
**Strengths:**
- No new external input trust boundaries. `latestReviewID` is an opaque GraphQL node id used only as an internal dedup discriminator — never interpolated into a shell command, URL, or query.
- Persisted store retains `0o600` perms; new `issueTrackerState` is a backward-compatible optional, so older `store.json` files decode without the corrupt-store backup path (covered by `storeDataDecodesWithoutIssueTrackerStateField`).
- Hydration filters persisted keys/statuses to live session UUIDs, bounding in-memory growth and preventing stale-session leakage.

**Concerns:** none.

### Code Quality

**Yellow — `latestReviews` ordering is an undocumented assumption, and `submittedAt` is fetched but never used**
`GitHubCodeBackend.swift:341-345` selects the round-2 discriminator with `latestReviewNodes.first(where: { state == "CHANGES_REQUESTED" })`, justified by the comment *"GitHub returns latestReviews newest-first."* GitHub's `latestReviews` connection has no `orderBy` argument and its node ordering is not documented to be newest-first — so this is load-bearing on undocumented behavior. The query was even updated to fetch `submittedAt` (`GitHubCodeBackend.swift:436`), but that field is never read.

- The **single-reviewer** path (the primary case, and the one in the test plan) is correct regardless of ordering, because `latestReviews` returns one node per author — a reviewer's resubmission rotates that single node's id.
- The **multiple-simultaneous-CHANGES_REQUESTED-reviewers** path is exposed: if the connection order isn't a stable newest-first, `.first(where:)` can (a) miss a genuine round-2 from a *different* reviewer (the older reviewer's id stays first → `newReview == false`), or (b) flip between two reviewers' ids across polls with no underlying change → a **spurious re-fire**, which is precisely the auto-respond loop this PR exists to prevent.

Recommend deciding the most-recent CHANGES_REQUESTED deterministically by sorting on the `submittedAt` you already fetch (`max(by: submittedAt)`), or — if first-is-fine is intentional — drop the unused `submittedAt` from the query and document why connection order is acceptable. Either way the dead field should go. A parser test with multiple CHANGES_REQUESTED nodes in shuffled order would lock the behavior in.

Related (same area, lower stakes): `PRStatusTransition.swift:108-110` requires **both** `old` and `new` `latestReviewID` to be non-nil for `newReview`. If a CHANGES_REQUESTED review is briefly absent from `latestReviews(first: 5)` (buried behind 5+ other reviewers' latest reviews) so the id reads `nil`, then resurfaces with a new id, the nil→id step won't fire. `reviewDecision` (which drives `reviewStatus`) and the `latestReviews`-derived id can legitimately disagree in that window. Edge case, but worth a comment acknowledging it.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 0 Green] findings. The Yellow is a small, contained fix (use the already-fetched `submittedAt`, or remove it) that belongs in this round since it touches the loop-prevention guarantee at the heart of the PR.

---

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new external input is trusted unsafely. The added GraphQL fields (`id`, `submittedAt`) are read defensively with `as? String` and `??` fallbacks; a malformed/absent payload degrades to `nil` rather than crashing.
- New persisted state (`PersistedIssueTrackerState`) is plain `Codable` written through the existing `JSONStore` path, which already enforces `0o600` perms and atomic writes. No new file/IO surface, no secrets, no shell interpolation.
- Hydration validates session UUIDs against live sessions (`Sources/Crow/App/IssueTracker.swift:1335-1348`), so a tampered/stale `store.json` can't reintroduce unbounded or orphaned entries.

**Concerns:** None.

### Code Quality
Strong overall. Verified the key behaviors:

- **Loop-avoidance is correct.** `dedupeKey` keys `.changesRequested` on `(session, kind, latestReviewID)` and `transitions(...)` fires only on bucket entry or review-ID rotation — never on head-SHA change alone (`PRStatusTransition.swift`). Since GitHub keeps `reviewDecision == CHANGES_REQUESTED` after an author push, this is exactly what prevents the agent's own response-push from re-triggering auto-respond. The replaced `dedupeKeyForChangesRequestedIncludesReviewIDOnly` test locks this in.
- **Restart safety holds.** Persisting both `previousPRStatus` and `emittedTransitionKeys`, plus `hydratePersistedState()` running before the first `refresh()`, means a restart mid-`changesRequested` won't re-fire. Confirmed `service.hydrateState()` (AppDelegate.swift:471) populates `appState.sessions` synchronously before `tracker.start()` (AppDelegate.swift:750), so hydration's session filter doesn't wipe valid state.
- **No-op write guard** (`applyPRStatuses`) avoids rewriting `store.json` every 60s on quiet polls — good follow-through on prior review feedback.
- **`JSONStore()` ad-hoc instance pattern** in `persistTrackerState`/`hydratePersistedState` matches the established convention in this file (lines 904/1146/1492) and is safe here: `IssueTracker` is `@MainActor`, so read+mutate runs without interleaving, and the app is the sole writer to `store.json` (CLI goes through the socket).
- **GitLab path** correctly defaults `latestReviewID = nil` (GitLabCodeBackend.swift:117) — no regression.
- **Backward compat** verified: `storeDataDecodesWithoutIssueTrackerStateField` plus optional `issueTrackerState` field.

**Consider (non-blocking):** In `GitHubCodeBackend.parseMonitoredPRsResponse`, `latestReviewID` is chosen via `max(by: submittedAt)`. For two distinct `CHANGES_REQUESTED` reviews sharing the same `submittedAt` (same-second submissions, or a degraded payload with no `submittedAt`), `max(by:)` falls back to array order, which GitHub doesn't guarantee to be stable across polls — a theoretical path to a spurious re-fire, the exact failure mode this PR targets. A secondary tie-break by `id` (e.g. compare `(submittedAt, id)`) would make the selection fully deterministic regardless of node order. Low probability (GitHub returns second-resolution `submittedAt` for submitted reviews), consequence is non-destructive (one redundant re-prompt) — hence a suggestion, not a blocker. The existing `testParseMonitoredPRsLatestReviewIDDegradesGracefullyWithoutSubmittedAt` test already documents the current behavior.

### Test Results
- `CrowCore` — 234 tests pass (incl. 3 new CROW-456 cases + replaced dedup-key test)
- `CrowPersistence` — 28 tests pass (incl. round-trip + legacy-decode)
- `CrowProvider` — 35 tests pass (incl. latest-submitted / degraded / nil cases)
- Root app target (`IssueTrackerDedupTests`) could not build in this review env — `Frameworks/GhosttyKit.xcframework` is missing its binary artifact locally. This is an environment limitation, not a PR defect; the author documented running it (191 tests) in the test plan.

### Summary Table

PR #460, opened 2026-06-08
Add Jira as a real TaskBackend via acli (CROW-459)

dhilgaertner:
Closes #459

## Summary

Adds **Atlassian Jira** as a real **`TaskBackend`** (driven by `acli`), so a session can track its work-unit in Jira while keeping **code + PRs on GitHub** via the existing `GitHubCodeBackend`. Per **ADR 0005** this is "the whole point of the split" — a task tracker with no embedded git paired with a GitHub code backend (the real, non-stub version of the `StubCorveilTaskBackend` shape).

`Session.codeProvider` already exists, so the cross-backend (Jira task + GitHub code) pairing is expressible today.

## What's included

**Backend & plumbing**
- New `Provider.jira` (task-only, like `.corveil`) + `Provider.isTaskOnly`.
- New `JiraTaskBackend` shelling `acli jira workitem {view,search,transition,assign,edit,create}` through the injected `ShellRunner`. Declares `.projectBoardStatus` (Jira workflow transitions) and wires `setTaskStatus`.
- `ProviderManager` factory: `taskBackend(.jira)` → `JiraTaskBackend` (threads optional `JiraConfig` — site / projectKey / jql); `codeBackend(.jira)` → `nil`. URL detection, `parseTicketURLComponents`, and `fetchTicket` handle Jira browse URLs and bare `PROJ-123` keys.
- `AcliProbe` (CrowCore) reports availability (`.ready` / `.notInstalled` / `.notAuthenticated`).

**Cross-backend correctness**
- Launchers (Claude/Codex/Cursor) split into `taskProvider` (ticket fetch via `acli`) and `codeProvider` (PR step via `gh`/`glab`).
- Session creation/backfill sets `codeProvider = .github` for task-only trackers so PR/git flows resolve to a real `CodeBackend`.
- `Scaffolder` allowlists the `acli jira workitem` subcommands.

**Workspace config & UI**
- `WorkspaceEntry` / `WorkspaceInfo` gain `taskProvider` / `jiraProjectKey` / `jiraJQL` / `jiraSite` (all optional; `derivedTaskProvider` defaults to the code provider so existing workspaces are unchanged).
- `WorkspaceFormView` exposes **Code Backend** and **Task Backend** as two independent, availability-aware pickers (Jira offered only when `acli` is authed, with an inline fix hint) plus Jira site/project/JQL fields.
- `IssueTracker` board polling fans out to Jira workspaces via `JiraTaskBackend.listAssigned` (best-effort).

[...]

PR #464, opened 2026-06-08
Fix Jira-tasked session PR linking + ticket metadata (#463)

### LINKED ISSUE #463: Jira-tasked sessions: no ticket attached on create + PR never links + skill skips ticket metadata (Session.codeProvider gap)
## Summary

After the Jira `TaskBackend` integration landed (#460 / `0c68231`), Jira-tasked sessions (task = Jira, code = GitHub) come in **with no ticket attached** and **never get their PR linked**. Two distinct defects, both because session-bootstrap / refresh paths weren't taught about task-only providers + `Session.codeProvider`.

Observed on live sessions this session: `max-monorepo-maxx-6846-…` and `max-monorepo-maxx-6848-…` (both `provider: jira`) have only a `ticket` link. Their PRs exist on GitHub — `securityscorecard/max-monorepo#45` (MAXX-6846) and `#47` (MAXX-6848) — but are not linked to the sessions.

---

## Bug A — PR links never attach to Jira-tasked sessions (confirmed)

The refresh-time PR reconciliation routes by the session's **single** `provider` field, which for a Jira session is `.jira` — a task-only provider with **no code backend**.

- `IssueTracker.buildReconcileCandidates()` (`Sources/Crow/App/IssueTracker.swift:1029-1040`) sets `provider = session.provider` (→ `.jira`) instead of resolving the code provider.
- `IssueTracker.reconcileMissingPRLinks()` (`:993-1001`) filters candidates into `.github` and `.gitlab` buckets only. A `.jira` candidate matches **neither** and is silently dropped — the GitHub PR on the session's branch is never queried, so no `.pr` link is ever created.

The correct convention already exists elsewhere and just wasn't applied here:
- `SessionService.findPRLink(...)` create path: `session.codeProvider ?? session.provider ?? .github` (`SessionService.swift:1331`).
- `AutoRespondCoordinator` (`:119`): same fallback, with a comment documenting the convention.

**Fix:** in `buildReconcileCandidates`, resolve the **code** provider (`session.codeProvider ?? session.provider`, falling back to host-sniffing) so Jira-tasked sessions reconcile against GitHub/GitLab. Add a regression test: a session with `provider == .jira`, `codeProvider == .github` must produce a `.github` reconcile candidate.

**Note (separate, also worth a look):** PR linking is **branch-based** — both `applySessionPRLinks` (`:871`, matches `{repoSlug}#{branch}` against open viewer PRs) and the reconcile pass query by the session's primary-worktree branch. If the PR is opened on a branch different from the session's worktree branch, it won't link even after the provider fix. (The two MAXX sessions above also have this branch mismatch, which confounds them as a test case — pick a session whose PR is on its own branch to validate Bug A in isolation.)

---

## Bug B — Jira ticket not attached on session creation (dir-name path)

`SessionService.parseTicketInfo(dirName:repoPath:)` (`SessionService.swift:1214-1255`) is **GitHub-only**:

- It scrapes the first **numeric** segment out of the worktree directory name (`Int(part)`), and `guard let num = info.number else { return info }` — so a name with no numeric segment yields an **empty** `TicketInfo` (no ticket link at all).
- When it does find a number, it constructs `https://github.com/{slug}/issues/{num}` and sets `provider = .github`. For a Jira session (e.g. dir `max-monorepo-maxx-6849-…`) the stray `6849` becomes a **bogus GitHub issue URL** for a ticket that isn't the real Jira work item.

There is no handling for Jira keys (`MAXX-1234`) or Atlassian site URLs. Result: Jira-tasked sessions created through this path come in with **no ticket** (or a wrong GitHub one).

**Fix:** make ticket resolution Jira-aware. Options: (a) recognize a Jira key in the dir name and reconstruct the Atlassian browse URL + `provider = .jira` (mirror `ProviderManager` URL/key parsing from #459); (b) better, prefer the real ticket identity carried from the create flow instead of re-deriving it from the directory name. Verify the **interactive** "create session from a Jira issue" path too — it should persist `ticketURL` from the selected `AssignedIssue` directly (this ticket's evidence is on the dir-name/recovery path; confirm the picker path isn't independently dropping it).

---

## Bug C — `crow-workspace` skill skips ticket metadata for Jira (same numeric assumption)

`skills/crow-workspace/setup.sh:380` only sets ticket metadata when a numeric ticket number is present:

```
if [[ -n "$TICKET_URL" && -n "$TICKET_NUMBER" ]]; then
  ... crow set-ticket --url "$TICKET_URL" --number "$TICKET_NUMBER"
```

Jira keys (`MAXX-6849`) are non-numeric, so `--ticket-number` is omitted and the whole `crow set-ticket` call is skipped — the session is created with only an `add-link` ticket link and **no ticket metadata** (`ticket_url`/`ticket_title` on the session stay empty unless set manually). This is the same GitHub-numeric assumption as Bug B, on the skill side.

**Fix:** set ticket metadata whenever `TICKET_URL` is present (number optional) — `crow set-ticket` already accepts `--url`/`--title` without `--number`. Mirror in `skills/crow-batch-workspace`.

---

## Why both happened

This is the `Session.codeProvider` follow-up that ADR 0005 / #459 flagged: task and code are two axes, but several bootstrap/refresh paths still assume `session.provider` is the code host (or that tickets are GitHub issues). The Jira integration added the `codeProvider` field and updated the create + auto-respond paths, but missed the PR-reconcile path (Bug A), the dir-name ticket parser (Bug B), and the `crow-workspace` skill's `setup.sh` (Bug C).

## Acceptance criteria

- [ ] A Jira-tasked + GitHub-code session gets its GitHub PR linked on the normal refresh cycle (Bug A), with a regression test on `buildReconcileCandidates`.
- [ ] A Jira-tasked session is created with its real Jira ticket URL/title attached, not a GitHub-issue guess or nothing (Bug B), covering both the dir-name/recovery path and the interactive Jira-pick path.
- [ ] `crow-workspace`/`crow-batch-workspace` `setup.sh` attaches ticket metadata for Jira sessions (number optional) (Bug C).
- [ ] No GitHub/GitLab-tasked session regresses.

## References

- `Sources/Crow/App/IssueTracker.swift:871` (`applySessionPRLinks`), `:986` (`reconcileMissingPRLinks`), `:1009-1066` (`buildReconcileCandidates`)
- `Sources/Crow/App/SessionService.swift:1214` (`parseTicketInfo`), `:1297-1331` (create + `findPRLink` with the correct fallback)
- `Sources/Crow/App/AutoRespondCoordinator.swift:113-119` (documented `codeProvider ?? provider` convention)
- `skills/crow-workspace/setup.sh:380-386` (Bug C); same pattern in `skills/crow-batch-workspace`
- Jira integration: #460 (`0c68231`); design context: ADR 0005, #459

dhilgaertner:
Closes #463

After the Jira `TaskBackend` integration (#460), Jira-tasked sessions (task = Jira, code = GitHub) never got their PR linked and skipped ticket metadata. Root cause across the defects: bootstrap/refresh paths route by the single `Session.provider` field (the task-only `.jira`) instead of the established `session.codeProvider ?? session.provider` convention used by `SessionService.findPRLink` and `AutoRespondCoordinator`.

## Bug A — PR links never attach (confirmed, highest value)
`IssueTracker.reconcileMissingPRLinks` buckets reconcile candidates into `.github`/`.gitlab` only, so a `.jira` candidate produced by `buildReconcileCandidates` matched neither and was silently dropped — the GitHub PR was never queried.

- Extracted a pure `resolveReconcileProvider(codeProvider:provider:host:)` static helper that routes by the **code** provider, skips task-only providers, and falls back to host sniffing (mirrors the existing convention + the static-helper pattern already in this file).
- `buildReconcileCandidates` now calls it. Net: a Jira-task + GitHub-code session yields a `.github` candidate, so the PR on the session branch links on the normal refresh cycle.
- **Regression test:** new `IssueTracker reconcile provider routing` suite covering the Jira+GitHub case plus precedence, host-sniff fallback, and no-regression for plain GitHub/GitLab.

## Bug C — Jira ticket metadata skipped on setup (confirmed)
`skills/crow-workspace/setup.sh` gated `crow set-ticket` on a non-empty `TICKET_NUMBER`; Jira keys have no standalone numeric id, so metadata was skipped (only the `add-link` ticket link was created). Now gates on `TICKET_URL` and adds `--number` only when numeric (a non-numeric value would make ArgumentParser reject the whole call and drop url+title too). `crow-batch-workspace` delegates to the same script, so this covers both.

## Bug B — orphan-recovery path (intentionally out of scope)
`recoverOrphan`/`parseTicketInfo` (GitHub-only) only fires for *truly orphaned* worktrees. The common interactive "Work on issue" picker already passes the real Jira `AssignedIssue.url` to `/crow-workspace`, so it persists the correct ticket once Bug C lands. Making the rare disk-recovery path Jira-aware was deliberately deferred (no reliable Jira signal beyond the dir name + workspace config); worth a follow-up ticket if needed.

## Verification
- `make app` — clean build.
- `swift test` — 196 tests across 21 suites pass, including the new reconcile-provider suite.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Models/Session.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/Session.swift
@@ -38,6 +38,21 @@ public struct Session: Identifiable, Codable, Sendable {
     /// Claude Code in the devRoot and are excluded from PR/issue tracking.
     public var isManager: Bool { kind == .manager }
 
+    /// Short label for the session's ticket badge/chip, or `nil` when no ticket
+    /// is attached. Jira tickets have no standalone numeric id, so prefer the
+    /// validated key (`MAXX-6859`) parsed from the browse URL; otherwise fall
+    /// back to `Issue #<number>` (GitHub/GitLab) or a bare `Issue` when only a
+    /// URL is known. Keeps the sidebar badge from vanishing on Jira sessions,
+    /// which carry `ticketURL`/`ticketTitle` but a nil `ticketNumber` (CROW-463).
+    public var ticketBadgeLabel: String? {
+        if let url = ticketURL, Validation.isJiraSpec(url), let key = Validation.jiraKey(from: url) {
+            return key
+        }
+        if let num = ticketNumber { return "Issue #\(num)" }
+        if ticketURL != nil { return "Issue" }
+        return nil
+    }
+
     public init(
         id: UUID = UUID(),
         name: String,
--- a/Packages/CrowCore/Tests/CrowCoreTests/SessionModelTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/SessionModelTests.swift
@@ -128,6 +128,40 @@ import Testing
     #expect(Session(name: "review", kind: .review).isManager == false)
 }
 
+// MARK: - Ticket Badge Label (CROW-463)
+
+@Test func ticketBadgeLabelGitHubUsesNumber() {
+    let session = Session(
+        name: "gh",
+        ticketURL: "https://github.com/org/repo/issues/42",
+        ticketNumber: 42,
+        provider: .github
+    )
+    #expect(session.ticketBadgeLabel == "Issue #42")
+}
+
+@Test func ticketBadgeLabelJiraUsesKeyDespiteNilNumber() {
+    // Regression for #463: Jira sessions carry a browse URL + title but a nil
+    // ticketNumber, so the badge must derive the key from the URL.
+    let session = Session(
+        name: "max-monorepo-maxx-6859",
+        ticketURL: "https://zitenote.atlassian.net/browse/MAXX-6859",
+        ticketTitle: "MAXX-6859: fold secondary vendor email domains",
+        ticketNumber: nil,
+        provider: .jira
+    )
+    #expect(session.ticketBadgeLabel == "MAXX-6859")
+}
+
+@Test func ticketBadgeLabelFallsBackToIssueWhenOnlyURL() {
+    let session = Session(name: "x", ticketURL: "https://example.test/thing")
+    #expect(session.ticketBadgeLabel == "Issue")
+}
+
+@Test func ticketBadgeLabelNilWhenNoTicket() {
+    #expect(Session(name: "none").ticketBadgeLabel == nil)
+}
+
 // MARK: - Enum Raw Values
 
 @Test func sessionKindRawValues() {
--- a/Packages/CrowProvider/Sources/CrowProvider/BackendTypes.swift
+++ b/Packages/CrowProvider/Sources/CrowProvider/BackendTypes.swift
@@ -150,6 +150,38 @@ public struct BranchPRMatch: Sendable {
     }
 }
 
+/// Input to `CodeBackend.findPRsMatchingKeys` — one (repo, key) tuple, where
+/// `key` is a ticket key (e.g. a Jira key `MAXX-6859`) expected to appear in a
+/// PR's title/body/branch. Lets reconcile recover PR links for task-only
+/// trackers (Jira) whose PR branch doesn't match the session's worktree branch.
+public struct KeyCandidate: Sendable, Hashable {
+    public let repoSlug: String
+    public let key: String
+
+    public init(repoSlug: String, key: String) {
+        self.repoSlug = repoSlug
+        self.key = key
+    }
+}
+
+/// One PR match returned from `findPRsMatchingKeys`, tagged with the candidate
+/// it came from so callers can route the result back to the right session(s).
+public struct KeyPRMatch: Sendable {
+    public let candidate: KeyCandidate
+    public let number: Int
+    public let url: String
+    public let state: String       // "OPEN" / "MERGED" / "CLOSED"
+    public let updatedAt: Date?
+
+    public init(candidate: KeyCandidate, number: Int, url: String, state: String, updatedAt: Date?) {
+        self.candidate = candidate
+        self.number = number
+        self.url = url
+        self.state = state
+        self.updatedAt = updatedAt
+    }
+}
+
 /// PR metadata returned from `CodeBackend.fetchPRMetadata` — the subset
 /// SessionService needs to prep a review clone.
 public struct PRMetadata: Sendable {
--- a/Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubCodeBackend.swift
+++ b/Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubCodeBackend.swift
@@ -189,6 +189,60 @@ public struct GitHubCodeBackend: CodeBackend {
         return Self.parseRecentPRsResponse(output, parsed: parsed)
     }
 
+    /// Search each repo for PRs whose title/body references `key` (e.g. a Jira
+    /// key like `MAXX-6859`). One `gh pr list --search` call per candidate
+    /// (the candidate set is small — only Jira-tasked sessions missing a PR
+    /// link). Best-effort per repo: a failing repo is skipped, not fatal.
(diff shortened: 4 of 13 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review

**Strengths:**
- `resolveReconcileProvider` is pure and `nonisolated` — no `appState`, no I/O — making it a safe target for unit tests and trivially auditable.
- No new attack surface: shell ticket-number guard `[[ "$TICKET_NUMBER" =~ ^[0-9]+$ ]]` defends against an exotic input being interpolated into the `crow set-ticket --number` argument. Even if the script's caller passed a hostile string, the regex constrains it to digits before it reaches the CLI.
- No secret/credential handling changes; the GitLab `gitlabHost` carry-through preserves the existing `GITLAB_HOST` auth boundary.

**Concerns:**
None.

### Code Quality

**Strengths:**
- The new helper mirrors the documented `codeProvider ?? provider` convention used by `SessionService.findPRLink` and `AutoRespondCoordinator`, eliminating the divergence that produced bug A. The precedence (`codeProvider` first, then `provider`, then `host` sniff) is consistent with the rest of the codebase.
- The task-only guard (`!p.isTaskOnly`) correctly defends against the defensive case where `codeProvider` is `nil` but `provider == .jira` — the helper falls back to host sniffing rather than producing a `.jira` reconcile candidate that would be dropped at the `.github`/`.gitlab` bucket filter (`Sources/Crow/App/IssueTracker.swift:1009-1018`), which was the original root cause.
- The downstream gate `if provider == .gitlab, gitlabHost == nil { continue }` (`IssueTracker.swift:1055`) is preserved — empty-host GitLab candidates are still skipped silently, so the helper doesn't introduce a regression where a misconfigured session would issue a wrong-host call.
- `skills/crow-workspace/setup.sh:384-388`: gating on `TICKET_URL` and only appending `--number` when numeric is the right call. `SetTicket.number` is typed `Int?` (`Packages/CrowCLI/Sources/CrowCLILib/Commands/MetadataCommands.swift:15`) and `validateSetTicketHasField` only requires at least one of url/title/number (`Packages/CrowCLI/.../Validation.swift:40`, `ValidationTests.swift:87`), so the URL-only path is supported by the CLI contract — the comment's claim is accurate.
- Downstream `set_project_status` is already gated by both `PROVIDER == "github"` (line 490) and `-n "$TICKET_NUMBER"` (line 501), so the setup.sh change does not enable any GitHub-only code path for Jira sessions.
- `set -uo pipefail` (no `set -e`) means the `[[ … ]] && …` short-circuit on a non-numeric `TICKET_NUMBER` doesn't abort the script — safe.
- Test coverage hits the right axes: Jira+GitHub (regression), Jira+GitLab (precedence), nil/nil with host variants (legacy sniff), `.jira` with nil `codeProvider` (defensive fallback), and plain GitHub/GitLab (no regression). The empty-host case is tested via `nilProvidersFallBackToHostSniff`.

**Notes (not blocking):**
- Intentional scope exclusion of bug B (`recoverOrphan` / `parseTicketInfo`) is reasonable — that path only fires for genuinely orphaned worktrees and the interactive picker already supplies the real Jira URL once bug C lands. Worth tracking as a follow-up ticket as the description suggests.
- I could not execute `swift test` in this review worktree (`GhosttyKit.xcframework` not present on disk in this checkout — a worktree-bootstrap quirk, unrelated to this PR). The PR description reports `make app` clean and 196 tests / 21 suites green; the diff itself is small and the helper-level tests are unambiguous, so I'm comfortable taking the author's verification at face value.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 0 Green] findings. Targeted, well-tested fix that resolves the routing bug at the correct layer and uses the established `codeProvider ?? provider` convention. The setup.sh shell guard is correct and consistent with the CLI's `Int?` argument type.

---

PR #465, opened 2026-06-09
Fail gracefully on GitHub org SAML enforcement (recover accessible-org tickets)

dhilgaertner:
## Problem

The user joined a GitHub org that enforces SAML SSO. Their `gh` OAuth token isn't authorized for it, so the consolidated GraphQL poll failed **every** 60s cycle:

```
gh: Resource protected by organization SAML enforcement. You must grant your OAuth token access to an organization within this enterprise.
```

The whole poll aborted — the console got spammed and CRO showed **no tickets at all**, even from orgs the token *can* see. The user has no intention of using the protected org on this machine and just wants Crow to fail gracefully while accessible-org tickets keep loading.

## Key insight

GitHub returns the SAML failure as a **partial response** (HTTP 200): the issues/PRs that resolved from accessible orgs are in `data`, alongside a `FORBIDDEN`/SAML entry in `errors` whose path points at the *specific* protected node. `gh api graphql` refuses any response with an `errors` array and exits non-zero; `ProcessShellRunner` merges stdout+stderr and threw the whole blob away. The data we want was already in hand — just discarded.

## Approach

Extends the codebase's existing `INSUFFICIENT_SCOPES` graceful-degradation precedent (`AssignedListing.missingScope` → `reportScopeWarning`) to SAML:

- **Detect** — `classifyGraphQLError` matches the SAML message and returns a new `ProviderError.samlRestricted` carrying the full merged blob.
- **Recover** — `decodeGraphQLData` peels the leading JSON object off the blob (string-aware brace balancing, past gh's trailing `gh:` line); `recoverPartialIssues` / `recoverPartialMonitoredPRs` reuse the existing node parsers to surface accessible-org data and flag the listing `samlRestricted`. No body recoverable → empty listing, never throws.
- **Keep loading** — `listAssigned` / `listMonitoredPRs` recover instead of throwing; the protected org is silently dropped. Secondary calls (`prStates`, `findRecentPRsForBranches`) route to the warning.
- **Warn once** — `IssueTracker.reportSAMLWarning` mirrors `reportScopeWarning`: one-time console log + a persistent, selectable Settings → General banner (`appState.githubSAMLWarning`), cleared on the next clean poll.

## Behavior change

| | Before | After |
|---|---|---|
| Console | SAML error every 60s | One line per session |
| Ticket list | Empty | Accessible-org tickets load; protected org hidden |
| User signal | None | One-time Settings banner with remediation |

## Tests

6 new unit tests in `BackendsTests.swift` (via `FakeShellRunner` returning realistic merged blobs): SAML detection + precedence, the tolerant extractor (incl. `{`/`}` inside string values), partial recovery for issues and PRs, and the empty-on-no-JSON fallback. **62 CrowProvider XCTests pass, 0 failures.**

## Verification notes

- The `CrowProvider` package (all core logic) builds clean and its full suite passes.
- The macOS app target and `CrowUI`/`IssueTracker` could not be compiled in this environment — they link `Frameworks/GhosttyKit.xcframework`, which is a build artifact absent from the checkout (`make ghostty` / `scripts/build-ghostty.sh`, needs zig + Metal toolchain). The two app-target files were syntax-checked with `swiftc -parse` and mirror existing patterns exactly. **Please confirm a full `make build` on a machine with GhosttyKit before merge.**
- Real SAML can't be reproduced locally without a protected org, so runtime coverage rests on the unit blobs. The recovery degrades safely to warn-only if a future `gh` ever emits no body on error.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -219,6 +219,12 @@ public final class AppState {
     /// Set by `IssueTracker` when the token lacks a required scope; cleared on next success.
     public var githubScopeWarning: String?
 
+    /// Non-fatal GitHub SAML warning surfaced in Settings. `nil` means no warning.
+    /// Set by `IssueTracker` when an org's SAML enforcement blocks the OAuth
+    /// token (accessible-org tickets still load); cleared on the next poll with
+    /// no SAML restriction.
+    public var githubSAMLWarning: String?
+
     /// Last observed GitHub GraphQL rate-limit snapshot. `nil` before the first
     /// successful query. Populated from the `rateLimit` block on each refresh.
     public var githubRateLimit: GitHubRateLimit?
--- a/Packages/CrowProvider/Sources/CrowProvider/BackendTypes.swift
+++ b/Packages/CrowProvider/Sources/CrowProvider/BackendTypes.swift
@@ -183,17 +183,24 @@ public struct AssignedListing: Sendable {
     /// user knows to refresh their token. The successful path returns the
     /// best-effort data alongside the scope marker; no error is thrown.
     public let missingScope: String?
+    /// True when the response was degraded because an org's SAML enforcement
+    /// blocked the OAuth token. The accessible-org issues GitHub still
+    /// returned are in `open`/`closed`; callers should surface a one-time UI
+    /// warning. Like `missingScope`, no error is thrown for this case.
+    public let samlRestricted: Bool
 
     public init(
         open: [AssignedIssue],
         closed: [AssignedIssue],
         rateLimit: GitHubRateLimit? = nil,
-        missingScope: String? = nil
+        missingScope: String? = nil,
+        samlRestricted: Bool = false
     ) {
         self.open = open
         self.closed = closed
         self.rateLimit = rateLimit
         self.missingScope = missingScope
+        self.samlRestricted = samlRestricted
     }
 }
 
@@ -204,11 +211,16 @@ public struct MonitoredPRListing: Sendable {
     public let reviewRequests: [ReviewRequest]
     public let viewerLogin: String
     public let rateLimit: GitHubRateLimit?
+    /// True when the response was degraded because an org's SAML enforcement
+    /// blocked the OAuth token. The accessible-org PRs/reviews GitHub still
+    /// returned are present; callers should surface a one-time UI warning.
+    public let samlRestricted: Bool
 
-    public init(viewerPRs: [PRRecord], reviewRequests: [ReviewRequest], viewerLogin: String, rateLimit: GitHubRateLimit? = nil) {
+    public init(viewerPRs: [PRRecord], reviewRequests: [ReviewRequest], viewerLogin: String, rateLimit: GitHubRateLimit? = nil, samlRestricted: Bool = false) {
         self.viewerPRs = viewerPRs
         self.reviewRequests = reviewRequests
         self.viewerLogin = viewerLogin
         self.rateLimit = rateLimit
+        self.samlRestricted = samlRestricted
     }
 }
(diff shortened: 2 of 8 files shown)


dgershman (approved):
## Code & Security Review

### Summary

This PR extends the existing `INSUFFICIENT_SCOPES` graceful-degradation precedent to GitHub's organization SAML enforcement: detect via stderr substring, recover accessible-org data from the merged `gh` stdout+stderr blob, and surface a one-time console log + persistent Settings banner. The implementation mirrors the `missingScope` / `reportScopeWarning` pattern almost line-for-line, which keeps the new surface area easy to reason about.

### Security Review

**Strengths:**
- No new shell interpolation surface — no command shape changes; only the existing typed-error / blob-recovery paths grow new branches. The new code only parses already-captured `gh` output.
- `firstBalancedJSONObject` is a careful, string-aware brace-balancer: it tracks `inString` and an `escaped` one-character lookahead, so a `}` inside a JSON string value (or an escaped `\"` inside a string) cannot prematurely close the object. The `testDecodeGraphQLDataExtractsLeadingObjectWithTrailingGhError` test exercises this explicitly (`"weird }{ title"`).
- No-body fallback (`recoverPartialIssues` / `recoverPartialMonitoredPRs` when `decodeGraphQLData` returns `nil`) yields an empty listing + `samlRestricted: true` instead of throwing — the cycle degrades safely.
- SAML check runs **before** the `INSUFFICIENT_SCOPES` / `read:project` substring match in `classifyGraphQLError`, so a SAML blob can't be misrouted to the scope branch (covered by `testClassifyGraphQLErrorSAMLTakesPrecedenceOverScope`).
- No secrets or tokens are surfaced in the new warning text or carried error payloads.

**Concerns:** none.

### Code Quality

**Strengths:**
- The `samlRestricted: Bool` flag on `AssignedListing` / `MonitoredPRListing` mirrors the existing `missingScope: String?` pattern exactly — call-site routing (warning vs error) is now uniform across both degradation modes.
- `IssueTracker.reportSAMLWarning` / `clearSAMLWarning` mirror `reportScopeWarning` / `clearScopeWarning` (including the `didLogGitHubSAMLWarning` once-per-session guard and the clear-on-clean-poll reset).
- Inline comments call out the genuinely tricky edge cases — notably the "no-projects retry also hit SAML → scope warning sacrificed for this cycle" trade-off in `listAssigned` (`GitHubTaskBackend.swift:73-78`) and the rationale for routing secondary calls (`prStates`, `findRecentPRsForBranches`) through `reportSAMLWarning` rather than recovering partial data (`IssueTracker.swift:711-715`).
- Tests cover SAML classification, classification precedence vs scope, the brace-in-string extractor, full-cycle recovery for both issues and PRs, and the no-JSON empty fallback.

**Consider (Green):**
- `IssueTracker.markInReview` (`IssueTracker.swift:2132-2141`) catches `.insufficientScope` and routes it to `reportScopeWarning`, but `.samlRestricted` would fall through to the generic `catch` and print a console line. In practice the SAML recovery path hides the protected-org ticket so this is unlikely to fire, but if a previously-visible ticket becomes SAML-protected mid-session a `markInReview` would silently fail without lighting the new banner. A small `catch ProviderError.samlRestricted { reportSAMLWarning(); return }` clause would close that minor gap.
- Minor duplication between `GitHubTaskBackend.recoverPartialIssues` and `GitHubCodeBackend.recoverPartialMonitoredPRs` — intentional and matches the existing per-backend parsing split; not worth a shared helper today.

### Verification

- `swift test` on `Packages/CrowProvider`: **62/62 tests pass, 0 failures** (matches the PR description).
- App target / `CrowUI` / `IssueTracker` were not built locally (GhosttyKit toolchain not present in this environment, as the PR description flagged). The SwiftUI banner in `SettingsView.githubSAMLWarningBanner` and the `IssueTracker` integration mirror existing patterns exactly and are syntactically clean on read.
- Real SAML cannot be reproduced without a protected org; runtime coverage rests on the unit-test blobs, which is appropriate.

### Summary Table

Issue #471, opened 2026-06-09
Terminal UX cluster — remaining gaps from #469 audit


dgershman:
Follow-up tracker for the confirmed gaps surfaced by the audit ask in #469. Each sub-item below was verified statically against `GhosttySurfaceView.swift` / `GhosttyApp.swift` / `crow-tmux.conf` while planning #469's PR; none of them were in scope for that PR (which covers items 1 + 3).

Items already shipped (no work needed):
- Item 1 — click-to-deselect → #469 PR.
- Item 2 — OSC 8 hover/click → #468 (PR #470).
- Item 3 — right-click context menu → #469 PR.
- Double/triple-click word/line selection → existing `crow-tmux.conf` `DoubleClick1Pane` / `TripleClick1Pane` wrappers.
- Drag-and-drop file path → existing `draggingEntered` / `performDragOperation` in `GhosttySurfaceView` (lines ~566-595).
- Cmd+Shift+V plain-text paste → `paste(_:)` already reads `.string` only, so paste is always plain text.

### 1. Cursor I-beam over terminal content
`resetCursorRects` only adds the pointing-hand cursor while hovering an OSC 8 link (introduced by #468). For non-link terminal content the cursor stays at whatever AppKit picked (typically the arrow), instead of the I-beam macOS terminals show. Fix: add an I-beam as the default cursor rect, overridden by pointing-hand on hovered links.

### 2. Cmd+F search-in-scrollback
No handler in `keyDown` / `performKeyEquivalent`. tmux has copy-mode search, but Crow doesn't expose it. Acceptance: Cmd+F opens a search affordance that drives tmux's `search-backward` / `search-forward` copy-mode commands.

### 3. Bracketed-paste safety prompt for multi-line pastes
`GhosttySurfaceView.paste(_:)` writes pasteboard contents directly to the surface with no multi-line guard. iTerm2 / Terminal.app both prompt when a paste contains a newline. Acceptance: detect `\n` in the paste payload and show a confirmation sheet before forwarding.

### 4. Spacebar Quick Look on selection
Standard macOS convention — selected URL → Quick Look preview, selected text → text Quick Look. No spacebar handler today. Acceptance: when a selection is active, spacebar opens `QLPreviewPanel` against the selected text or its detected URL.

### 5. Smart-detect bare URLs and file:line references
[...]

PR #470, opened 2026-06-09
Wire OSC 8 hyperlink hover/click in libghostty embed

dgershman:
## Summary

- `GhosttyApp.handleAction()` now dispatches `GHOSTTY_ACTION_MOUSE_OVER_LINK` and `GHOSTTY_ACTION_OPEN_URL` (previously only `SHOW_CHILD_EXITED` was wired, so OSC 8 actions were silently dropped).
- `GhosttySurfaceView.setHoveringLink(_:)` + a `resetCursorRects()` override swap the pointing-hand cursor on/off when the mouse enters/leaves a hyperlinked cell.
- `OPEN_URL` is gated by a hard-coded scheme allowlist (`http`, `https`, `mailto`, `ftp`, `ftps`); `file`, `javascript`, `data`, etc. are logged and dropped.

Closes #468

## Test plan

- [ ] `make ghostty && swift build` — clean.
- [ ] In any Crow terminal pane (including a tmux'd one), run:
      `printf '\e]8;;https://github.com/radiusmethod/crow/issues/466\e\\Crow #466\e]8;;\e\\\n'`
      → hover swaps to pointing hand; click opens browser.
- [ ] Negative test: `printf '\e]8;;javascript:alert(1)\e\\xss\e]8;;\e\\\n'` → click is dropped with `[GhosttyApp] OPEN_URL rejected` in Console.app.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #468, opened 2026-06-09
Wire OSC 8 hyperlink hover/click in libghostty embed


dgershman:
Follow-on from the SwiftTerm spike (#466 — no-go).

## Problem

OSC 8 hyperlinks (`\e]8;;URL\e\\text\e]8;;\e\\`) don't hover-highlight or click-open in any Crow terminal. libghostty fires the relevant actions and tmux's `allow-passthrough on` already lets the OSC 8 bytes reach the Ghostty surface — the data is flowing, Crow's action dispatcher is just dropping them.

## Where

`Packages/CrowTerminal/Sources/CrowTerminal/GhosttyApp.swift:102-126` — `GhosttyApp.handleAction()` currently dispatches only `GHOSTTY_ACTION_SHOW_CHILD_EXITED`. Every other action (including `GHOSTTY_ACTION_MOUSE_OVER_LINK` and `GHOSTTY_ACTION_OPEN_URL`) hits the silenced/unhandled path.

The relevant action tags + payload struct are documented in `vendor/ghostty/include/ghostty.h` (search for `mouse_over_link`, `open_url`).

## What to wire

1. **`GHOSTTY_ACTION_MOUSE_OVER_LINK`** — set/clear the pointing-hand cursor on the surface NSView when the action fires with a non-empty link. (The action provides the link string when entering hover; an empty/nil string clears it.)

2. **`GHOSTTY_ACTION_OPEN_URL`** — hand the URL to `NSWorkspace.shared.open(URL(string:))` after a `URL` parse + scheme allowlist (`http`, `https`, `mailto`, etc. — refuse `file://`, `javascript:`, `data:` to match SwiftTerm's default safer behavior).

Both follow the existing target-is-surface pattern from `handleAction()`:

```swift
if target.tag == GHOSTTY_TARGET_SURFACE,
   let userdata = ghostty_surface_userdata(target.target.surface) {
    let view = Unmanaged<GhosttySurfaceView>.fromOpaque(userdata).takeUnretainedValue()
    // dispatch on main actor and act on `view`
}
```

## Estimate

~50 LOC total per the spike scope.

## Verification

The fixture from the spike branch still works as the manual test:

```sh
printf '\e]8;;https://github.com/radiusmethod/crow/issues/466\e\\Crow #466\e]8;;\e\\\n'
```

- Hover → cursor changes to pointing-hand
- Click → opens URL in default browser

[...]

Issue #466, opened 2026-06-09
Experiment: spike SwiftTerm as a lighter terminal renderer (eval alternative to libghostty)


dgershman:
**Goal:** time-boxed spike to evaluate replacing libghostty with [SwiftTerm](https://github.com/migueldeicaza/SwiftTerm) as Crow's terminal renderer. Ticket closes when a **go/no-go recommendation** is posted with measured data — not when a swap merges. Actual replacement is a separate ticket only if the experiment recommends Go.

## Why investigate

Two symptoms motivate this:

1. **Hyperlink hover/click doesn't work** in any Crow terminal. libghostty exposes the actions (`GHOSTTY_ACTION_MOUSE_OVER_LINK`, `GHOSTTY_ACTION_OPEN_URL`) and tmux's `allow-passthrough on` already lets OSC 8 sequences reach Ghostty, but `GhosttyApp.handleAction()` only handles `SHOW_CHILD_EXITED` — every other action including link hover is silenced. Data is flowing; Crow doesn't act on it.
2. **The embed is wildly oversized.** Audit found Crow uses **~23 of libghostty's C functions** (lifecycle, input, mouse, clipboard, render tick) and ignores **~95% of Ghostty's feature surface** — images (sixel/kitty), ligatures, splits, tabs, profiles, themes, keybindings, the action system. The static library is **135 MB** linked into Crow's binary; the vendored source tree is **1.0 GB**. The recent tmux-mouse workaround thread (#445 / #446 / #452) also points at the embed being awkward to live with — every behavior tweak ripples through tmux bindings because Ghostty's own behavior model isn't quite what Crow wants.

**SwiftTerm** is the most credible lighter alternative discovered in survey: pure Swift, MIT, AppKit-native (`NSView`), built-in OSC 8 hyperlink support (`linkReporting = .explicit` + `requestOpenLink` delegate), active maintenance (v1.13.0 March 2026), production use in Secure Shellfish / La Terminal / CodeEdit. Trade-off: CoreText rendering instead of Ghostty's Metal pipeline — slower under extreme log spam, fine for normal agent output.

We don't yet know whether SwiftTerm is a real swap-in for Crow's specific constraints (tmux-only backend per ADR 0001, the shared-surface model, the SentinelWaiter shell-readiness dance). This experiment answers that with data, not opinion.

## Spike scope

A time-boxed branch (NOT a merge-ready PR) that:

[...]

PR #472, opened 2026-06-09
Terminal UX cluster: click-to-deselect + right-click context menu (CROW-469)

### LINKED ISSUE #469: Terminal UX cluster: click-to-deselect, hyperlink hover/click, right-click context menu, and audit related gaps
## What's broken / missing

A cluster of standard terminal interactions either doesn't work or doesn't behave the way macOS conventions lead users to expect. Three confirmed gaps from direct user observation, plus an explicit ask to audit adjacent ones:

### 1. Click-outside-selection should deselect
After drag-selecting text, clicking elsewhere in the pane leaves the previous selection highlighted (and on the clipboard from #446's `copy-pipe-no-clear`). Standard macOS / terminal behavior is: a single click on unselected text immediately clears the prior selection and places the cursor at the click point.

**Acceptance:** click anywhere outside the current selection → selection clears. Click *inside* the current selection → cursor placed but selection cleared (matches Terminal.app/iTerm2). Drag-extending the selection still works.

### 2. Hyperlink hover + click (OSC 8)
URLs printed via OSC 8 escape sequences aren't clickable, and hovering doesn't change the cursor. libghostty exposes `GHOSTTY_ACTION_MOUSE_OVER_LINK` and `GHOSTTY_ACTION_OPEN_URL` (see `vendor/ghostty/include/ghostty.h`), and `crow-tmux.conf:16` already has `allow-passthrough on` so OSC 8 reaches Ghostty. The data is flowing; Crow ignores it — `GhosttyApp.handleAction()` only branches on `SHOW_CHILD_EXITED`. ~50 LOC to wire up.

**Acceptance:** hovering over a link changes the cursor to the pointing hand; clicking opens the URL via `NSWorkspace.shared.open(url)`. Cmd+click should also work as the macOS convention for "open without selecting."

**Heads-up:** #466 is the SwiftTerm spike. SwiftTerm has OSC 8 built in via `linkReporting = .explicit` + `requestOpenLink`. If that spike goes Go, this work is done as a side effect — coordinate before implementing on the Ghostty path to avoid throwaway code.

### 3. Right-click context menu
Right-clicking a Crow pane does nothing. Standard macOS terminals offer at least: **Copy** (enabled when selection exists), **Paste**, **Select All**, **Clear** (scrollback), **Open Link** (when right-clicking on / inside an OSC 8 link), and ideally **Open URL** / **Open File in Editor** for smart-detected URLs/paths in the selection.

**Acceptance:** `GhosttySurfaceView` implements `menu(for event: NSEvent) -> NSMenu?` with the above items, enabled/disabled by context (e.g. Copy only when `ghostty_surface_has_selection` is true; Open Link only when the right-click landed on a link).

## Audit ask — what other features are silently missing?

The three gaps above suggest a class of "standard terminal behaviors that libghostty exposes but Crow's embed never wires up." This ticket also asks the implementer to **audit the surface and file follow-ups for the rest**, rather than discovering them ticket-by-ticket. Likely suspects, based on inspection of `GhosttySurfaceView.swift` and what's missing from the libghostty action handler:

| Candidate | Why it's likely a gap |
|---|---|
| **Cursor shape change on hover** — I-beam over text regions, pointing hand over links, arrow over chrome | No `resetCursorRects` / `addCursorRect` overrides; cursor likely stays at whatever AppKit decided |
| **Double-click word selection / triple-click line selection** | Verify these work — they're tmux/Ghostty defaults but the current `mouseDown` only sends position + button to libghostty |
| **Selection should not shift / extend on wheel-scroll** | #452 fixed this for the drag-end binding via `stop-selection`. Confirm under scroll-after-click and shift+scroll cases |
| **Cmd+F search-in-scrollback** | Not present; tmux has copy-mode search but Crow doesn't expose it |
| **Spacebar Quick Look on selection** | macOS-wide convention; selected URL → Quick Look preview |
| **Drag-and-drop file path into pane** | Standard terminal feature: drop a file → its path is inserted as text |
| **Bracketed-paste safety prompt** for multi-line pastes | Mitigates accidental commit of pasted commands; iTerm2/Terminal both warn |
| **Cmd+up/down to jump between prompts** | Optional; iTerm2 has it |
| **Triple-click + drag for line-by-line extension** | Standard terminal behavior |
| **IME / dead-key handling under tmux** | `GhosttySurfaceView` has IME plumbing — confirm it works through the tmux indirection |
| **Tab navigation via keyboard from a focused terminal** | Crow's tab UI lives outside Ghostty; verify focus doesn't trap |
| **Find-and-open file references** (`path/to/file.swift:123` in agent output) → opens in editor at the line | Convenience feature; common in IDE-adjacent terminals |
| **Smart-detect URLs without OSC 8** (a bare `https://...` in plain text) → clickable | OSC 8 covers what apps emit deliberately; smart-detect covers everything else |
| **Selection / clipboard persistence across pane switches** | Verify selecting in pane A then switching to pane B doesn't lose the clipboard text |
| **Pasteboard styled text** (Cmd+Shift+V for plain-text paste) | macOS convention; not sure if Ghostty enforces plain text |

Implementer should:
1. Try each in a Crow session. Confirm or deny.
2. For each confirmed gap, post a one-line "still broken" note in this ticket's comments.
3. Group fixes by mechanism (where one libghostty action handler unlocks multiple features) and either bundle in this ticket's PR or file follow-ups with the grouping called out.

## Coordination with related tickets

- **#445 / #446 / #452** — tmux-mouse selection thread. The `copy-pipe-no-clear` and `stop-selection` bindings landed in #452 are the foundation for "selection survives mouse-up and wheel-scroll." Item 1 above (click-to-deselect) is the missing piece in that thread — selection survives the *wrong* events too. Likely a small `MouseDown1Pane` rebinding (`select-pane \; send-keys -X clear-selection` or similar) rather than a Swift change. Verify which layer owns it.
- **#466** — SwiftTerm spike. Several of the items above (link click, cursor shape, search-in-scrollback) ship for free on SwiftTerm. If #466 goes Go, this ticket's scope shrinks to "anything SwiftTerm doesn't already cover." If #466 goes No-go, this ticket is the cheaper path forward — particularly item 2's "wire OSC 8 in `GhosttyApp.handleAction()`" was already called out as #466's No-go alternative.
- **Implementation order recommendation:** wait for #466's go/no-go before doing items 2 and most audit items. Item 1 (click-to-deselect) and item 3 (right-click menu) are renderer-agnostic and can land immediately.

## Files

| Purpose | Path |
|---|---|
| Surface view — mouse handlers, copy/paste, focus | `Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift` |
| Action dispatch — needs link handlers | `Packages/CrowTerminal/Sources/CrowTerminal/GhosttyApp.swift:54-126` (`handleAction`) |
| Tmux config — mouse / copy bindings already in play | `Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf` |
| Ghostty action header — what's available to wire | `vendor/ghostty/include/ghostty.h` |

## Out of scope

- Renderer swap to SwiftTerm — that's #466.
- Tmux backend changes.
- Themes, font selection, profiles — not motivated by this ask.

dgershman:
Closes #469

Lands items 1 and 3 of #469. Item 2 (OSC 8 hover/click) already shipped via #468 / PR #470 and item 3's "Open Link" layers on top of it. Audit follow-ups filed as #471.

### Item 1 — click-to-deselect (`crow-tmux.conf`)

After drag-selecting text, the highlight survived a subsequent click anywhere in the pane. Root cause: the default `MouseDown1Pane` binding (`select-pane`) doesn't dismiss the copy-mode selection that #452's `copy-pipe-no-clear` deliberately preserves to keep the clipboard intact.

Added a new `bind -T root MouseDown1Pane` that mirrors the existing `DoubleClick1Pane` wrapper shape: forward to the inner app when it captured the mouse, otherwise `send-keys -X cancel` to dismiss copy-mode. `MouseDrag1Pane` still re-enters copy-mode cleanly for a new selection.

### Item 3 — right-click context menu (`GhosttySurfaceView.swift`)

Overrode `menu(for:) -> NSMenu?` with the five items called out in the ticket's acceptance criteria. Items are enabled/disabled at construction time:

| Item | Enabled when |
|---|---|
| Copy | `ghostty_surface_has_selection` returns true |
| Paste | NSPasteboard has a string |
| Select All | `TmuxBackend.shared.activeTerminalID` is known |
| Clear | `TmuxBackend.shared.activeTerminalID` is known |
| Open Link | `hoveredLinkURL != nil` and its scheme is in `GhosttyApp.allowedURLSchemes` |

Surfaced two new `TmuxBackend` methods mirroring the existing `sendText(id:text:)` binding-lookup pattern:

- `clearHistory(id:)` — `tmux clear-history -t <target>`. Matches macOS Terminal "Clear" / iTerm2 "Clear Buffer": on-screen rows survive, scrollback is wiped.
- `selectAll(id:)` — `copy-mode -H` → `history-top` → `begin-selection` → `history-bottom`. After this, the existing `copy-pipe-no-clear` MouseDragEnd binding (or Cmd+C / context-menu Copy) writes the captured text to the pasteboard.

### Item 3a — extend `setHoveringLink` to carry the URL

#468 stored only a `Bool` for hover state because `GHOSTTY_ACTION_OPEN_URL` re-emits the URL on click. The right-click "Open Link" item needs the URL *without* a click. Widened `setHoveringLink(_:)` from `Bool` to `String?`, decoded the `mouse_over_link.url`/`len` payload in `GhosttyApp.handleAction()`, and promoted `allowedURLSchemes` to module scope so the menu gates Open Link against the same allowlist as the action callback.

## Audit verdict

Code-static pass against the 14 candidates in #469's audit section. Per-item verdicts:

**Already wired (no work):**
- Double/triple-click word/line selection — `crow-tmux.conf:57-58`.
- Drag-and-drop file path → text — `GhosttySurfaceView.draggingEntered` / `performDragOperation`.
- Cmd+Shift+V plain-text paste — `paste(_:)` reads `.string` only, so paste is already plain text.

**Confirmed gap, filed under #471:**
- Cursor I-beam over terminal content (only pointing-hand on links is wired).
- Cmd+F search-in-scrollback.
- Bracketed-paste safety prompt for multi-line pastes.
- Spacebar Quick Look on selection.
- Smart-detect bare URLs + `path:line` references.
- Cmd+up/down to jump between prompts.

**Needs a live test — noted in #471, not filed:**
- IME under tmux.
- Selection survival under wheel-scroll (#452 covered mouse-up only).
- Triple-click + drag for line-by-line extension.
- Tab keyboard navigation from a focused terminal.

## Coordination

- Built on top of #468 (PR #470) which merged before this PR was opened. Rebased onto `main` to inherit it.
- #466 (SwiftTerm spike) closed NO-GO, so the Ghostty embed is the path forward.

## Test plan

- [ ] `make ghostty && swift build --arch arm64` — Crow app builds clean (note: `swift build` without Xcode hits a pre-existing `BuildInfo` symbol gap in `AboutView.swift` — that's the Xcode build-phase generator, not this PR).
- [ ] Drag-select text in a Crow terminal; click elsewhere in the pane → highlight dismisses, NSPasteboard text survives (no #446 regression).
- [ ] Drag-select again immediately → new selection forms cleanly.
- [ ] Run `vim` / `htop` inside the pane → mouse-using TUIs still receive mouse events (the `mouse_any_flag` short-circuit in the new binding).
- [ ] With a selection: right-click → Copy enabled; click it → pasteboard matches the selection.
- [ ] Without a selection: right-click → Copy disabled.
- [ ] Right-click → Paste sends the pasteboard string as text.
- [ ] Right-click → Select All highlights the entire scrollback; subsequent Cmd+C copies it.
- [ ] Right-click → Clear; `tmux capture-pane -p -S - -t <target>` shows only on-screen rows.
- [ ] Print an OSC 8 link (`printf '\\e]8;;https://example.com\\e\\\\example\\e]8;;\\e\\\\\\n'`). Hover → pointing-hand cursor (#468). Right-click *on* the link → Open Link enabled; click → opens in default browser. Right-click off the link → Open Link disabled.
- [ ] Regression sweep: Cmd+C, Cmd+V, typing, IME, scroll wheel, drag-select all still work.

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/GhosttyApp.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/GhosttyApp.swift
@@ -102,8 +102,10 @@ public final class GhosttyApp {
     /// URL schemes Crow will hand to `NSWorkspace.shared.open()` from an OSC 8
     /// click. Anything else (file, javascript, data, vbscript, …) is dropped on
     /// the floor — the URL is still rendered, but clicking it is a no-op. Keep
-    /// this tight; OSC 8 emitters are not always trustworthy.
-    nonisolated private static let allowedURLSchemes: Set<String> = [
+    /// this tight; OSC 8 emitters are not always trustworthy. Exposed at module
+    /// scope so the right-click menu in `GhosttySurfaceView` can gate its "Open
+    /// Link" item against the same allowlist as the action callback.
+    nonisolated static let allowedURLSchemes: Set<String> = [
         "http", "https", "mailto", "ftp", "ftps",
     ]
 
@@ -128,15 +130,23 @@ public final class GhosttyApp {
             return true
         }
         // OSC 8 hover: libghostty reports the mouse entering/leaving a hyperlink
-        // cell. Non-zero `len` = entering, zero = leaving. We don't need the URL
-        // itself — Ghostty tracks it internally and re-emits via OPEN_URL on click.
+        // cell. Non-zero `len` = entering, zero = leaving. The URL bytes ride
+        // along on entry so the right-click "Open Link" menu item can fire
+        // without waiting for a click to re-emit OPEN_URL.
         if action.tag == GHOSTTY_ACTION_MOUSE_OVER_LINK {
             if target.tag == GHOSTTY_TARGET_SURFACE,
                let userdata = ghostty_surface_userdata(target.target.surface) {
-                let hovering = action.action.mouse_over_link.len > 0
+                let payload = action.action.mouse_over_link
+                let url: String? = {
+                    guard payload.len > 0, let ptr = payload.url else { return nil }
+                    return String(
+                        decoding: UnsafeRawBufferPointer(start: ptr, count: Int(payload.len)),
+                        as: UTF8.self
+                    )
+                }()
                 DispatchQueue.main.async {
                     let view = Unmanaged<GhosttySurfaceView>.fromOpaque(userdata).takeUnretainedValue()
-                    view.setHoveringLink(hovering)
+                    view.setHoveringLink(url)
                 }
             }
             return true
--- a/Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift
@@ -7,7 +7,7 @@ public final class GhosttySurfaceView: NSView {
     private var pendingText: [String] = []
     private var markedTextStorage = NSMutableAttributedString()
     private var trackingArea: NSTrackingArea?
-    private var isHoveringLink: Bool = false
+    private var hoveredLinkURL: String?
 
     /// Backoff schedule for retrying createSurface() when ghostty_surface_new returns nil.
     /// 4 retries totalling ~7.5s before declaring permanent failure.
@@ -206,16 +206,20 @@ public final class GhosttySurfaceView: NSView {
     }
 
     /// Toggle the pointing-hand cursor when libghostty reports the mouse is
-    /// hovering an OSC 8 hyperlink. Fed from `GHOSTTY_ACTION_MOUSE_OVER_LINK`
-    /// in `GhosttyApp.handleAction()`.
-    public func setHoveringLink(_ hovering: Bool) {
-        guard isHoveringLink != hovering else { return }
-        isHoveringLink = hovering
-        window?.invalidateCursorRects(for: self)
+    /// hovering an OSC 8 hyperlink, and stash the URL for the right-click
+    /// "Open Link" menu item. Fed from `GHOSTTY_ACTION_MOUSE_OVER_LINK` in
+    /// `GhosttyApp.handleAction()`. Pass `nil` when the mouse leaves the link.
+    public func setHoveringLink(_ url: String?) {
+        guard hoveredLinkURL != url else { return }
+        let wasHovering = hoveredLinkURL != nil
+        hoveredLinkURL = url
+        if wasHovering != (url != nil) {
+            window?.invalidateCursorRects(for: self)
+        }
     }
 
     public override func resetCursorRects() {
-        if isHoveringLink {
+        if hoveredLinkURL != nil {
             addCursorRect(bounds, cursor: .pointingHand)
         }
     }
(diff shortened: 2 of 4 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Tight, well-documented PR. The tmux-conf click-to-deselect binding faithfully mirrors the existing `DoubleClick1Pane` wrapper, the two new `TmuxBackend` methods follow the established binding-lookup + `ensureRunningServer` + `reportIfTimeout` pattern exactly, and widening `setHoveringLink` to carry the URL is clean. One real defect blocks the context-menu feature, though.

### Critical Issues

**🔴 Red — The menu's enabled/disabled logic is defeated by AppKit auto-enabling (`GhosttySurfaceView.swift:367-426`)**

`NSMenu.autoenablesItems` defaults to `true`, and the new `menu(for:)` never sets it to `false`. With auto-enabling on, AppKit **ignores** the `isEnabled` you set at construction time and instead enables any item whose `target` responds to its `action` selector (since there's no `validateMenuItem(_:)` implementation). The view responds to `copy(_:)`, `paste(_:)`, `contextSelectAll(_:)`, `contextClear(_:)`, and `contextOpenLink(_:)` — so **every item renders enabled regardless** of `hasSelection`, `canPaste`, `activeTerminalID`, or `linkURL`.

This directly contradicts the acceptance table in the PR description and fails these test-plan items:
- "Without a selection: right-click → Copy disabled" → Copy will be **enabled**.
- "Right-click off the link → Open Link disabled" → Open Link will be **enabled**.

It's not a crash (each `@objc` handler and `copy`/`paste` guards its own preconditions, so clicking a should-be-disabled item is a safe no-op), but the greying-out UX — a stated requirement — does not work.

Fix is one line right after `let menu = NSMenu(title: "Terminal")`:

```swift
menu.autoenablesItems = false
```

### Security Review

**Strengths:**
- "Open Link" gates the URL through the same `GhosttyApp.allowedURLSchemes` allowlist as the OSC 8 click path, with `scheme?.lowercased()` — `file`/`javascript`/`data` URLs can't be launched from the menu. Promoting the allowlist to module scope (rather than duplicating it) keeps the two paths from drifting apart. Good.
- URL payload decode in `handleAction()` is correctly bounded by `payload.len` and guarded on `payload.url != nil`.
- `clearHistory`/`selectAll` build their tmux target from the trusted `bindings[id]` window index, not from any user-supplied string — no command-injection surface.

No security concerns.

### Code Quality

- **🟢 Green — Main-thread blocking:** `contextClear`/`contextSelectAll` run synchronous tmux subprocess calls on `@MainActor` from the menu handler. Consistent with the existing `sendText` pattern, so not a regression, but `selectAll` issues four sequential `run()`/`sendKeys` round-trips — worth a glance if menu actions ever feel laggy. Not blocking.
- **🟢 Green — Hover staleness:** `menu(for:)` relies on `hoveredLinkURL`, which is updated via `DispatchQueue.main.async` from `MOUSE_OVER_LINK`. For a normal right-click after the cursor moved into place this is fine; if a right-click ever lands without a preceding `mouseMoved` (e.g. keyboard-invoked context menu), the link state could be stale. Edge case, acceptable.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 0 Yellow, 2 Green] findings. The single Red (`autoenablesItems`) is a one-line fix that restores the entire enable/disable behavior the PR is built around.

---

dhilgaertner (approved):
## Code & Security Review

Reviewed the four changed files (`GhosttyApp.swift`, `GhosttySurfaceView.swift`, `crow-tmux.conf`, `TmuxBackend.swift`). Clean, well-commented, and architecturally consistent. No blocking or should-fix issues found.

### Critical Issues
None.

### Security Review
**Strengths:**
- **Open Link is properly sandboxed.** `contextOpenLink` only ever opens a `URL` stashed in `representedObject`, and that object is non-nil only when the hovered URL's scheme passes `GhosttyApp.allowedURLSchemes` (`http/https/mailto/ftp/ftps`). Promoting `allowedURLSchemes` to module scope means the menu gates against the *same* allowlist as the OSC 8 click callback — `file://`, `javascript:`, `data:` etc. stay dropped on the floor. Scheme is lowercased before the set check, so case can't bypass it.
- **No use-after-free decoding the hover URL.** In `handleAction`, `mouse_over_link.url`/`len` is decoded into a Swift `String` synchronously (immediately-invoked closure) *before* the `DispatchQueue.main.async` hop, so the libghostty-owned buffer isn't read after the callback returns. `len > 0` and non-nil pointer are both guarded.
- Paste path remains plain-text only (`string(forType: .string)`), unchanged.

**Concerns:** None.

### Code Quality
- **Correct target for Select All / Clear.** Confirmed via `TerminalSurfaceView` that Crow uses a single shared cockpit `GhosttySurfaceView` (`terminalID` is nil on it; at most one terminal visible). Routing Copy/Paste through the per-view `surface` while routing Select All / Clear through `TmuxBackend.shared.activeTerminalID` is therefore the right split, not an inconsistency.
- **`autoenablesItems = false` is the right call** and well-documented — AppKit's default would have ignored per-item `isEnabled` and lit up every item whose target responds to the selector.
- **`setHoveringLink` cursor-rect invalidation is correct:** it only invalidates when hover *state* (nil↔non-nil) flips, while still updating `hoveredLinkURL` when moving link→link — so the pointing-hand cursor isn't needlessly rebuilt yet Open Link always targets the URL actually under the cursor.
- **`crow-tmux.conf` binding** mirrors the existing `DoubleClick1Pane` wrapper: `mouse_any_flag` short-circuits to `send-keys -M` for mouse-using TUIs, normal-mode clicks just `select-pane` (no regression), and only an in-copy-mode click hits `send-keys -X cancel` — preserving the #452 `copy-pipe-no-clear` clipboard behavior while fixing the stuck-highlight.
- New `TmuxBackend` methods follow the established `bindings`/`ensureRunningServer`/`reportIfTimeout` pattern; `selectAll`'s `copy-mode -H` → `history-top` → `begin-selection` → `history-bottom` sequence is valid.

### Minor (Green — non-blocking, consider)
- `contextSelectAll` issues four synchronous `Process.waitUntilExit()` tmux spawns on the main thread (`contextClear` one). It's a deliberate, brief user action and consistent with the codebase's existing synchronous tmux model (`copy`/`paste`/`writeText`/`makeActive`), so not a regression — but batching into a single `tmux` invocation or hopping off-main would avoid any UI hitch on a slow server.
- The two new thin `TmuxBackend` wrappers ship without unit tests, matching the existing wrappers.

### Build / Static Analysis
Did not compile locally: `libghostty` is not vendored in the worktree (requires `make ghostty` + the Xcode build-phase `BuildInfo` generator the PR's own test plan notes). Review is static. The decode/menu code relies on `ghostty_action_mouse_over_link_s.url`, consistent with the existing `.len` usage in the same struct.

### Summary Table

---

PR #474, opened 2026-06-09
Fix source-file exit-1 from bare unbind-key -a on prefix table

### LINKED ISSUE #473: crow-tmux.conf: unbind-key -a fails with 'table prefix doesn't exist', breaks source-file exit status
**Bug:** `crow-tmux.conf:39` has a bare `unbind-key -a` which fails on Crow's tmux server because no `prefix` key table exists. The reload still works for every other command (tmux's `source-file` doesn't abort on first error), but the overall reload command returns **exit 1**.

**Why it matters:** #451's "reconcile stale config on attach" likely checks the exit status of `source-file` to decide whether the reload succeeded. With a non-zero exit, any automation gating on success will think the reload failed and either retry, log an error, or fall back to a destructive path — even though the bindings are actually current. Also tripped my manual verification: `tmux source-file crow-tmux.conf` returned exit 1 and printed `table prefix doesn't exist`, looking like a hard failure.

**Repro (manual reload against any Crow tmux server):**
```bash
SOCK=$(ls /var/folders/*/tmux-*/crow-tmux.sock 2>/dev/null | head -1)
CONF=/Users/danny/Projects/crow/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf
tmux -S "$SOCK" source-file "$CONF"
echo "exit: $?"
# exit: 1
# table prefix doesn't exist
```

Verified all the actual settings/bindings (mouse on, MouseDown1Pane cancel-on-click, DoubleClick1Pane select-word, TripleClick1Pane select-line, MouseDragEnd1Pane copy-pipe-no-clear in both copy-mode tables) ARE loaded — only `unbind-key -a` errored.

**Root cause:** `unbind-key -a` with no `-T <table>` defaults to the prefix table. The comment block immediately above the offending line (`crow-tmux.conf:34-35`) actually documents this: "prefix table only — root, copy-mode, and copy-mode-vi are left intact." But on a server that has never bound a prefix key, that table doesn't exist yet, and unbinding from a non-existent table is an error.

**Fix (pick one):**

1. **Guard against missing table** (simplest, recommended):
   ```
   if-shell -F "#{key_table:prefix}" "unbind-key -a -T prefix"
   ```
   No-op when there's nothing to unbind.

2. **Suppress the error explicitly:**
   ```
   unbind-key -a -T prefix 2>/dev/null
   ```
   Uglier and may not work the same way under tmux's command-mode (the redirect is shell syntax, not tmux syntax).

3. **Pre-create the prefix table by binding a stub first:**
   ```
   bind-key -T prefix Any send-keys
   unbind-key -a -T prefix
   ```
   Works but adds a side effect.

4. **Skip the unbind entirely** — if Crow never uses prefix bindings, there's nothing to clear; the line at :39 is dead code. Check git blame on that line — if it was added defensively and Crow has never relied on it for anything, just remove it.

Recommend option 1 or 4 depending on git blame intent.

**Acceptance:**
- `tmux source-file <crow-tmux.conf>` returns exit 0 on a fresh Crow tmux server (no prior prefix bindings) and on a populated one.
- Output of `tmux -S <sock> show-messages` after a reload contains no `"table prefix doesn't exist"` line.
- All existing bindings the conf installs are still loaded after reload (same coverage as today — the verification above lists what to check).
- #451's reconcile-on-attach reports success after a Crow relaunch with no spurious "reload failed" entries in logs.

**Related:**
- #451 — reconcile stale bundled tmux conf on attach. The exit-status bug here likely produces false-failure logs in that flow.
- #453 — the PR that introduced the current bindings layout where the `unbind-key -a` lives.
- The line at :39 was specifically called out in the comment block above it; whoever added it knew about the prefix-only behavior but assumed the table always exists.

dgershman:
Closes #473

## Summary

- The bare `unbind-key -a` at `crow-tmux.conf:39` defaulted to the prefix table. Once the table is empty (which is the steady state on Crow's server) the command errors with `table prefix doesn't exist`, making `source-file <conf>` return **exit 1** even though every other binding loaded cleanly.
- That broken exit status was poisoning #451's reconcile-on-attach path (`TmuxBackend.reconcileBundledConfigIfStale` in `Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:572-577`), which then logs `[CrowTelemetry tmux:config_reconcile_failed …]` on every successful reload.
- Fix: stub-bind `Any` into the prefix table before `unbind-key -a -T prefix`, so the clear always has at least one binding to remove. End state is the same empty prefix table — without the exit 1.
- Comment block updated to explain the new pattern and why; existing warning about not widening to `-T root` (would re-break #452 wheel scrollback) preserved.

## Verification

```
new-session exit: 0
1st source-file: exit 0
2nd source-file: exit 0  (idempotent reload)
show-messages: no 'table prefix' errors
list-keys -T prefix: 'table prefix doesn't exist'  (same end state)
mouse on / allow-passthrough on / status off  ✓
root MouseDown1Pane, DoubleClick1Pane, TripleClick1Pane bound  ✓
copy-mode + copy-mode-vi MouseDragEnd1Pane copy-pipe-no-clear  ✓
root WheelUpPane / MouseDrag1Pane defaults preserved (#452)  ✓
```

Regression test added in `BundledResourcesTests.tmuxConfHasNoBarePrefixUnbind` pins three invariants: no bare `unbind-key -a`/`unbind -a` outside comments, presence of the stub bind and the explicit clear, and stub-precedes-clear ordering.

## Test plan

- [x] `swift test --package-path Packages/CrowTerminal --filter BundledResourcesTests` — 7/7 pass (including the new regression test)
- [x] Manual `tmux -S <sock> source-file <conf>` on a fresh server: exit 0
- [x] Manual second `source-file` (reload path): exit 0
- [x] `tmux show-messages` after both: no `table prefix doesn't exist`
- [x] Prefix table empty (`list-keys -T prefix` → doesn't exist); root defaults preserved
- [ ] In-app smoke: launch with `CROW_TMUX_BACKEND=1`, restart so reconcile-on-attach fires, confirm `[CrowTelemetry tmux:config_reconciled …]` in logs (no `tmux:config_reconcile_failed`)

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf
@@ -30,13 +30,23 @@ set -gs status off
 # still works alongside this.
 set -gs mouse on
 
-# Crow owns the keyboard namespace. `unbind-key -a` (no `-T`) clears the
-# `prefix` table only — `root`, `copy-mode`, and `copy-mode-vi` are left
-# intact (use `-a -T <table>` to clear a non-prefix table). That's
-# load-bearing here: the mouse fix depends on the root-table defaults
-# (`WheelUpPane`, `MouseDrag1Pane`, …) surviving — wiping root with
-# `unbind-key -a -T root` would re-break #452 wheel scrollback.
-unbind-key -a
+# Crow owns the keyboard namespace: the prefix table must end up empty so
+# tmux's default `C-b <key>` shortcuts can't fire in the surface. A bare
+# `unbind-key -a` (no `-T`) targets the prefix table, but it fails with
+# `table prefix doesn't exist` once the table has been cleared — and that
+# failure made `source-file` return exit 1 on every reconcile-on-attach
+# pass (#451 path), poisoning #451's gating logic (#473).
+#
+# Pre-creating the table with a stub `Any` binding guarantees `unbind-key
+# -a -T prefix` always has something to clear, so the command stays
+# idempotent across fresh startup AND `source-file` reloads. Net result is
+# the same empty prefix table as before — without the exit-1.
+#
+# Do NOT widen this to `unbind-key -a -T root`: the mouse fix below relies
+# on root-table defaults (`WheelUpPane`, `MouseDrag1Pane`, …) surviving;
+# wiping root would re-break #452 wheel scrollback.
+bind-key -T prefix Any send-keys
+unbind-key -a -T prefix
 
 # Default MouseDragEnd1Pane is `send-keys -X copy-pipe-and-cancel`; the
 # -and-cancel exits copy mode after copying, which wipes the visual
--- a/Packages/CrowTerminal/Tests/CrowTerminalTests/BundledResourcesTests.swift
+++ b/Packages/CrowTerminal/Tests/CrowTerminalTests/BundledResourcesTests.swift
@@ -62,4 +62,33 @@ struct BundledResourcesTests {
         #expect(body.contains(#"send-keys -X select-word ; run -d0.3 ; send-keys -X copy-pipe-no-clear "pbcopy""#))
         #expect(body.contains(#"send-keys -X select-line ; run -d0.3 ; send-keys -X copy-pipe-no-clear "pbcopy""#))
     }
+
+    @Test func tmuxConfHasNoBarePrefixUnbind() throws {
+        // #473: a bare `unbind-key -a` (no `-T`) defaults to the prefix
+        // table, which is empty/non-existent after the first clear on
+        // Crow's tmux server. The command errored with `table prefix
+        // doesn't exist` and made `source-file` return exit 1 — breaking
+        // #451's reconcile-on-attach gating. The fix uses an explicit
+        // `bind-key -T prefix Any …` stub followed by
+        // `unbind-key -a -T prefix` so the clear is idempotent. This test
+        // pins both invariants: the bare form is gone, and the
+        // stub-then-clear pair is present in that order.
+        let url = try #require(BundledResources.tmuxConfURL)
+        let body = try String(contentsOf: url, encoding: .utf8)
+        for line in body.split(separator: "\n", omittingEmptySubsequences: false) {
+            let trimmed = line.trimmingCharacters(in: .whitespaces)
+            guard !trimmed.hasPrefix("#") else { continue }
+            #expect(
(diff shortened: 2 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

Empirically verified the entire fix against a real `tmux 3.6a` server (matching the embedded backend). Every claim in the PR description reproduces.

### Verification

**Bug reproduced** — bare `unbind-key -a` (no `-T`) defaults to the prefix table, preserves root (`WheelUpPane` survives → #452 safe), but on the second `source-file` errors `table prefix doesn't exist` and returns **exit 1**. Confirmed exactly.

**Fix verified** with the real bundled `crow-tmux.conf`:
- Fresh `new-session` → exit 0; `source-file` reload → exit 0; third reload → exit 0 (idempotent).
- Prefix table ends empty (`list-keys -T prefix` → doesn't exist) — same end state as before.
- Root defaults preserved: `MouseDown1Pane`/`DoubleClick1Pane`/`TripleClick1Pane` (3) bound; `WheelUpPane` intact.
- `copy-mode` + `copy-mode-vi` `MouseDragEnd1Pane copy-pipe-no-clear` both present.
- `show-messages` has no real errors — the only `table prefix` text is tmux's own command-echo logging (`command: unbind-key -a -T prefix`), not a failure.

The mechanism is sound: `bind-key -T prefix Any send-keys` re-creates the prefix table so `unbind-key -a -T prefix` always has at least one binding to remove, making the clear idempotent across startup and reload. The transient `Any` stub is unbound on the very next line within the same synchronous config pass, so it can never fire.

### Security Review
**Strengths:**
- App-internal tmux config; no secrets, no user input, no injection surface. The stub binding is inert and immediately removed.
- The `-T root` widening hazard (which would re-break #452) is explicitly called out in the comment and guarded.

**Concerns:** none.

### Code Quality
- Comment block accurately explains the failure mode and the idempotency mechanism — verified against actual tmux behavior.
- Regression test `tmuxConfHasNoBarePrefixUnbind` pins the right invariants: bare form absent, stub+clear present, stub precedes clear. (Note: tests could not be built in this review checkout — `Frameworks/GhosttyKit.xcframework` has no vendored binary artifact here — but the PR reports 7/7 green and the assertion logic reads correctly.)

### Findings
- 🟢 **Consider** — the bare-form guard matches the exact strings `unbind-key -a` / `unbind -a` only; a future edit with trailing args or non-single-space whitespace wouldn't trip it. Acceptable trade-off for a focused regression pin; no change needed.

### Summary Table

---

PR #476, opened 2026-06-09
Add Reload Terminal Config menu item (CROW-475)

### LINKED ISSUE #475: Crow menu: 'Reload Terminal Config' option (manual trigger for tmux source-file)
**Ask:** Add a menu item to the Crow app — likely under **Crow → Reload Terminal Config** or **View → Reload Terminal Config** — that re-sources the bundled `crow-tmux.conf` against the live tmux server. Equivalent to running:

```bash
tmux -S <crow socket> source-file <bundled crow-tmux.conf>
```

…but exposed in the UI so users don't have to remember the socket path or the conf path.

## Why

#451 reconciles config-on-attach (when Crow launches and connects to the tmux server). That handles the common case. But there are real situations where a manual trigger is the only option:

1. **You edited `crow-tmux.conf` directly** (debugging a binding, trying a tweak) and want to test it without restarting Crow. The conf-on-attach path doesn't re-fire unless Crow itself relaunches.
2. **The hot-patch we ran in this session** — the user and I have been iterating on `MouseDragEnd1Pane`, `MouseDown1Pane`, etc. live via `tmux ... bind ...`. After each iteration, a one-click "reload from the bundled conf" gets things back to the canonical state. Today the user had to ask me to compose the command for them.
3. **Reconcile-on-attach can be defeated.** If Crow's attach path skips reconcile for any reason (#473's exit-1 may already be tripping it; future detection logic might miss certain edits), a manual button is the trustworthy fallback.
4. **Surface area for debug builds.** When the implementer is changing tmux behavior in a feature branch, this menu item halves the test-loop time vs. relaunching Crow each iteration.

## Acceptance

- New menu item visible in the Crow main menu — placement at the implementer's discretion, but discoverable (suggest **Crow → Reload Terminal Config** alongside other dev-facing items, or **View → Reload Terminal Config**).
- Activating it runs `tmux -S <socket> source-file <bundled crow-tmux.conf>` against the current tmux server (use the same socket-discovery and config-path logic #451 already has).
- **Surface success/failure in a transient banner / toast** — "Terminal config reloaded" on success, error text on failure. The current source-file errors come back via `tmux show-messages`; surface that diff to the user.
- **Handle the #473 cosmetic exit-1**: don't treat "table prefix doesn't exist" as a failure once #473 is fixed; until then, either ignore that specific error string or block this ticket on #473.
- Optional but nice: an optional keyboard shortcut (e.g. Cmd+Shift+R or whatever's free in Crow's keymap).
- No effect when no tmux server is running yet — gray out the menu item or no-op with a "no terminal server" message.

## Out of scope

- Auto-reloading on conf file change (a filesystem watcher) — separate ticket if anyone wants it; this one is the manual trigger.
- Restarting the tmux server entirely (kills sessions; not useful, separate concern).
- Editing the conf from inside Crow — out of scope.

## Implementation notes

- Socket path discovery already exists somewhere in `Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift` — reuse it.
- Bundled conf path via `BundledResources.tmuxConfURL` in `Packages/CrowTerminal/Sources/CrowTerminal/BundledResources.swift`.
- Same `Process` invocation pattern as the rest of the tmux-backend's calls.

## Related

- #451 — reconcile-on-attach. This ticket is the manual counterpart.
- #473 — the cosmetic `unbind-key -a` exit-1 that will trip naive success-checking until fixed. Block on #473 or work around it explicitly in the success-check logic.

dgershman:
## Summary
- Adds **Crow → Reload Terminal Config** menu item that re-sources the bundled `crow-tmux.conf` against the live tmux server without restarting it.
- Manual counterpart to #451's reconcile-on-attach: covers conf edits mid-session and the "reset hot-patched bindings to canonical state" workflow.
- Success/failure surfaced via a system notification (`Terminal config reloaded` / `Terminal config reload failed`).

## Implementation
- **`TmuxBackend.reloadBundledConfig()`** — new public method that unconditionally runs `source-file <bundled conf>` against the cached controller. Returns `nil` on success, error string on failure. Pairs with the existing `reconcileBundledConfigIfStale` (which is mtime-gated, appropriate for the attach path; the menu item is user-initiated so we skip the gate).
- **`NotificationManager.notifyConfigReloaded(errorText:)`** — surface result via existing system-notification path, gated by `globalMute` / `systemNotificationsEnabled` like other notify methods.
- **`AppDelegate`** — menu item placed alongside "Restart tmux Server" (sibling dev/recovery action). No confirmation alert (reload is non-destructive — windows/sessions survive). No keyEquivalent, mirroring the sibling item.

Server-not-running case: `reloadBundledConfig` returns "tmux server is not running" and the notification surfaces it. Cleaner than wiring `NSMenuItemValidation` to gray out the item.

#473's cosmetic `unbind-key -a` exit-1 was fixed in `e526131`, so this needs no workaround.

Closes #475

## Test plan
- [ ] Launch Crow and confirm **Crow → Reload Terminal Config** appears between "Restart tmux Server" and "Hide Crow"
- [ ] Click on a running server; expect "Terminal config reloaded" notification + `tmux:config_reloaded_by_user` telemetry line
- [ ] Edit `crow-tmux.conf`, click reload, verify the change is live without restarting Crow
- [ ] Quit without server-restart, relaunch, click before any pane exists; expect "tmux server is not running" notification (no crash)
- [ ] Confirm Restart tmux Server still works and still shows its confirmation alert

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -577,6 +577,33 @@ public final class TmuxBackend {
         }
     }
 
+    /// Re-source the bundled `crow-tmux.conf` against the live tmux server
+    /// unconditionally — driven by the "Reload Terminal Config" menu item
+    /// (#475), where the user has explicitly asked for a reload. Unlike
+    /// `reconcileBundledConfigIfStale`, this skips the mtime gate.
+    ///
+    /// Returns `nil` on success, or a human-readable error string the caller
+    /// can surface in a banner. Idempotent: `source-file` against a live
+    /// server updates server-scoped options in place; existing windows and
+    /// sessions are unaffected.
+    @MainActor
+    public func reloadBundledConfig() -> String? {
+        guard let ctrl = controller, ctrl.hasSession() else {
+            return "tmux server is not running"
+        }
+        guard let confURL = BundledResources.tmuxConfURL else {
+            return "bundled crow-tmux.conf not found"
+        }
+        do {
+            try ctrl.run(["source-file", confURL.path])
+            NSLog("[CrowTelemetry tmux:config_reloaded_by_user path=\(confURL.path)]")
+            return nil
+        } catch {
+            NSLog("[CrowTelemetry tmux:config_reload_failed error=\"\(error)\"]")
+            return "\(error)"
+        }
+    }
+
     /// Pure policy: reconcile when either timestamp is missing (conservative
     /// — a redundant `source-file` is cheap) or when the conf is newer than
     /// the running server.
--- a/Sources/Crow/App/AppDelegate.swift
+++ b/Sources/Crow/App/AppDelegate.swift
@@ -976,6 +976,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         let restartTmuxItem = NSMenuItem(title: "Restart tmux Server", action: #selector(restartTmuxServer), keyEquivalent: "")
         restartTmuxItem.target = self
         appMenu.addItem(restartTmuxItem)
+        // Non-destructive — re-sources the bundled tmux conf against the live
+        // server. No keyEquivalent for now (#475); can be added if asked.
+        let reloadConfigItem = NSMenuItem(title: "Reload Terminal Config", action: #selector(reloadTerminalConfig), keyEquivalent: "")
+        reloadConfigItem.target = self
+        appMenu.addItem(reloadConfigItem)
         appMenu.addItem(NSMenuItem.separator())
         appMenu.addItem(withTitle: "Hide Crow", action: #selector(NSApplication.hide(_:)), keyEquivalent: "h")
         let hideOthersItem = NSMenuItem(title: "Hide Others", action: #selector(NSApplication.hideOtherApplications(_:)), keyEquivalent: "h")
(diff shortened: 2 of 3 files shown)


dhilgaertner (approved):
## Code & Security Review

Small, well-scoped change adding a **Crow → Reload Terminal Config** menu item that re-sources the bundled `crow-tmux.conf` against the live tmux server. Implementation is clean, mirrors existing patterns, and is non-destructive.

### Critical Issues
None.

### Security Review
**Strengths:**
- No new input surface: `source-file` runs against the app-bundled `crow-tmux.conf` (`BundledResources.tmuxConfURL`), not user-supplied paths — no injection vector.
- No new auth/permission surface; reuses the existing `globalMute` / `systemNotificationsEnabled` gating in `NotificationManager`.
- Errors are surfaced as strings, not swallowed silently, and telemetry is logged on both success and failure.

**Concerns:**
- None.

### Code Quality
- `TmuxBackend.reloadBundledConfig()` correctly skips the mtime gate (user-initiated reload) and reuses the cached controller rather than calling `ensureRunningServer()`, so a reload never spins up a session as a side effect. Good.
- `@MainActor` annotation is consistent with the call site in `AppDelegate` and stricter than the surrounding `controller` accessors — no actor-isolation risk.
- The new code mirrors `reconcileBundledConfigIfStale` and the sibling `notify*` methods (guard ordering, `source-file` invocation, telemetry format).

### Notes (non-blocking, Green)
- `Sources/.../TmuxBackend.swift:592` — when the tmux server is actually alive but `controller` is `nil`/unhydrated (e.g. right after relaunch before any pane exists), `reloadBundledConfig()` returns `"tmux server is not running"`, which is slightly inaccurate. The PR documents this as the expected behavior, and it's the safe/cleaner choice vs. wiring `NSMenuItemValidation`. Fine as-is.
- `Sources/.../NotificationManager.swift:163` — if notifications are muted/disabled, a failed reload produces no visible feedback (telemetry is still logged). Consistent with every other `notify*` method in the file, so acceptable.
- No unit test for `reloadBundledConfig()` itself — but its IO path requires a live tmux server and `@MainActor`, matching the existing untested `reconcileBundledConfigIfStale` IO path (only the pure `shouldReconcile` policy is unit-tested). Consistent with the codebase.

### Build / Test note
Full `swift build` could not be run in this review checkout — it fails at `GhosttyKit.xcframework` (missing local binary artifact), an environment limitation unrelated to this change. The diff was reviewed by inspection and is syntactically sound Swift consistent with the surrounding code.

### Summary Table

---

PR #478, opened 2026-06-10
Emit transitions on first observation when state warrants attention (CROW-477)

### LINKED ISSUE #477: Auto-respond misses first observation when PR is already in CHANGES_REQUESTED (#456/#457 follow-up)
**Bug:** `PRStatus.transitions()` returns `[]` on first observation, so when Crow's IssueTracker first observes a PR that's already in `CHANGES_REQUESTED` (or already has failing checks), no transition emits and auto-respond silently doesn't fire. The agent waits forever.

`Packages/CrowCore/Sources/CrowCore/Models/PRStatusTransition.swift:102`:

```swift
public static func transitions(
    from old: PRStatus?,
    to new: PRStatus,
    ...
) -> [PRStatusTransition] {
    guard let old else { return [] } // first observation — never fire
    ...
}
```

The guard exists to prevent spurious fires on Crow startup (state restoration would otherwise re-fire every active PR), but it creates a silent miss whenever the IssueTracker poll doesn't catch the pre-review state — which is common in real workflows.

## Real-world repro

Hit live today on corveil#1298:

- Crow session created: `2026-06-09T19:08:35Z`
- PR #1298 opened (no reviews yet): `2026-06-09T21:08:34Z` (~2h after session)
- `CHANGES_REQUESTED` review submitted: `2026-06-09T21:14:05Z` (6 minutes after PR opened)

For Crow to have caught this, the IssueTracker poll would have needed to fire **in that 6-minute window**. It didn't. Result: first observation was already `changesRequested` → guard at line 102 returned `[]` → AutoRespondCoordinator never called → agent never nudged.

This is structurally distinct from #456 (round 2+ same-bucket re-fires) and #457's fix — the dedup-key rotation #457 added doesn't help here because no transition emits in the first place.

## Failure modes the guard misses

| Scenario | Why guard misses |
|---|---|
| PR opens and reviewer requests changes within one poll interval | First observation already `changesRequested` → `oldStatus = nil` → `[]` |
| PR has failing checks at the moment Crow first attaches the link | First observation already `failing` → `oldStatus = nil` → `[]` |
| Session creation happens *after* a PR is already in `changesRequested` (e.g., session re-created after the 24h cleanup) | Same — first observation already in the bucket |
| Crow restart while a PR is in `changesRequested` and the persisted dedup key has been wiped (no persistence yet) | Today: would re-fire; with persistence: would skip — either way, the line 102 guard is the wrong place to mitigate this |

## Fix

**Relax the guard so first observation can emit a transition when the new state is itself one we'd auto-respond to.** The fix should NOT fire blindly on every first observation — only when the new state matches a bucket auto-respond cares about, and only paired with the persisted-dedup work to prevent restart-spurious fires.

Two interdependent changes:

### 1. Replace the line-102 guard with a state-targeted version

```swift
guard let old else {
    // First observation — only emit when the new state is itself a
    // "needs attention" condition, so a PR Crow first sees in
    // CHANGES_REQUESTED or with failing checks still gets routed to
    // auto-respond. The bucket-only conditions below depend on old,
    // so we synthesize the emission here.
    var out: [PRStatusTransition] = []
    if new.reviewStatus == .changesRequested {
        out.append(PRStatusTransition(
            kind: .changesRequested,
            sessionID: sessionID,
            prURL: prURL,
            prNumber: prNumber,
            headSha: new.headSha,
            latestReviewID: new.latestReviewID,
            failedCheckNames: []
        ))
    }
    if new.checksPass == .failing {
        out.append(PRStatusTransition(
            kind: .checksFailing,
            sessionID: sessionID,
            prURL: prURL,
            prNumber: prNumber,
            headSha: new.headSha,
            failedCheckNames: new.failedCheckNames
        ))
    }
    return out
}
```

### 2. Persist emittedTransitionKeys across restarts (was proposed in #456, confirm whether it landed in #457)

This is the spurious-fire prevention the line-102 guard was indirectly providing. With persistence:

- After Crow restart, the persisted dedup keys are reloaded.
- If `sessionID|changesRequested|reviewID|headSha` is already present, the new "first observation" emission is correctly suppressed.
- If it's not present (genuine first observation), the transition fires and auto-respond runs.

If persistence didn't land in #457, this ticket needs to add it. If it did, this ticket just needs change (1) and a test that verifies the interplay.

**Important sequencing:** if persistence is not present, do NOT ship (1) alone — Crow restart on a desktop with N open PRs in `changesRequested` would re-fire every single one, which is much worse than the current silent miss. Either ship them together, or land persistence first and (1) second.

## Acceptance

- A session whose first observation of a linked PR is already in `changesRequested` emits a transition and auto-respond fires.
- Same for `failing` checks on first observation.
- Crow restart on an environment with the persisted dedup state populated does **not** re-fire transitions that have already been emitted in a previous run.
- `oldStatus == nil` + `new.reviewStatus == .open` (PR open, no review) still emits no transition (we only fire on attention-needing states).
- New tests in `PRStatusTransitionTests.swift` cover:
   - Nil-old + new-changesRequested → emits one `changesRequested` transition.
   - Nil-old + new-failing → emits one `checksFailing` transition.
   - Nil-old + new-changesRequested + new-failing → emits both.
   - Nil-old + new-open (no review) → emits none.
   - Crow restart simulation with persisted key → first poll observes same state → no spurious emission.

## Right-now workaround for users hitting this today

Click the **Refine** quick action on the session card (which calls `AutoRespondCoordinator.dispatchManual(action: .refine, sessionID:)` per `Sources/Crow/App/AutoRespondCoordinator.swift:82`). Sends the same prompt to the agent terminal that the auto-path would have. One manual nudge per stuck session.

## Related

- #456 — bucket-only dedup miss (round 2+). The complement of this bug — same code path, different failure mode.
- #457 — re-arm fix that landed for #456. Does not address the first-observation guard. Verify whether the persisted-dedup work proposed in #456's spec actually shipped in #457; if not, this ticket needs to also land it.
- `IssueTracker.swift:1485` — the call site that passes `oldStatus` (potentially nil) into `transitions(from:to:...)`.
- `AutoRespondCoordinator.swift:82` — `dispatchManual` is the manual-nudge escape hatch users can use today.

dgershman:
Closes #477

## Summary

- Relax the `guard let old else { return [] }` in `PRStatus.transitions(from:to:...)` (`Packages/CrowCore/Sources/CrowCore/Models/PRStatusTransition.swift`) so a PR Crow first observes already in `CHANGES_REQUESTED` or with `failing` checks emits a transition and routes to auto-respond.
- The startup-flood safety the original guard provided is preserved: `nil`-old with an attention-neutral state (`.reviewRequired`/`.approved`, `passing`/`pending` checks) still emits nothing.
- Restart-spurious re-fires are prevented by the persisted `emittedTransitionKeys` that shipped in #457 — not by this guard. If a key was emitted in a prior run, the `IssueTracker.swift:1485-1495` dedup correctly suppresses the synthesized first-observation candidate.

## Why this was missed before

`PRStatus.transitions(from:to:...)` returned `[]` on first observation to avoid re-firing every active PR at startup. But the IssueTracker poll interval is wide enough that a PR can be opened and receive a `CHANGES_REQUESTED` review entirely between two polls (real-world hit on corveil#1298 today: 6 minutes between PR open and the review). First observation was already `changesRequested` → guard returned `[]` → auto-respond silent → agent never nudged.

The `#456`/`#457` work that landed dedup-key persistence (`PersistedIssueTrackerState.emittedTransitionKeys` round-tripping through `JSONStore`) is the prerequisite that makes this guard safe to relax. Confirmed it shipped before writing this fix.

## Test plan

- [x] `arch -arm64 swift test --arch arm64 --package-path Packages/CrowCore --filter PRStatusTransitionTests` — 18/18 pass including 4 new first-observation tests:
  - `firstObservationChangesRequestedFires`
  - `firstObservationFailingChecksFires`
  - `firstObservationBothChangesRequestedAndFailingFiresBoth`
  - `firstObservationOpenWithNoReviewDoesNotFire` (replaces the old `firstObservationNeverFires`; covers the startup-flood-safety case)
- [x] `arch -arm64 swift test --arch arm64 --package-path Packages/CrowPersistence --filter JSONStoreTests` — persistence round-trip still green (no schema change, sanity only).
- [ ] Manual: open a Crow session whose linked PR is already in `CHANGES_REQUESTED`. Confirm auto-respond fires once on first poll. Restart Crow with same PR still in `CHANGES_REQUESTED` — confirm it does NOT re-fire (persisted dedup key suppresses).

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Models/PRStatusTransition.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/PRStatusTransition.swift
@@ -77,9 +77,15 @@ public struct PRStatusTransition: Sendable, Equatable {
 extension PRStatus {
     /// Compute the transitions implied by moving from `old` to `new`.
     ///
-    /// Returns an empty array on the first observation (`old == nil`) so
-    /// existing PR state never triggers a fire-on-startup. Otherwise emits
-    /// at most one `.changesRequested` and one `.checksFailing`, in that order.
+    /// On the first observation (`old == nil`), emits a transition only when
+    /// the new state is itself a "needs attention" condition — `.changesRequested`
+    /// or failing checks. A PR first observed as `.open` with passing/pending
+    /// checks emits nothing (the startup-flood safety the bare guard used to
+    /// provide). Persisted `emittedTransitionKeys` in `IssueTracker` (CROW-456)
+    /// suppress restart re-fires of transitions already emitted by a prior run.
+    ///
+    /// Otherwise emits at most one `.changesRequested` and one `.checksFailing`,
+    /// in that order.
     ///
     /// `.changesRequested` fires when the bucket is entered (`old` wasn't
     /// `.changesRequested`) or when the latest CHANGES_REQUESTED review ID
@@ -99,7 +105,36 @@ extension PRStatus {
         prURL: String,
         prNumber: Int?
     ) -> [PRStatusTransition] {
-        guard let old else { return [] } // first observation — never fire
+        guard let old else {
+            // First observation (CROW-477): emit only when the new state is
+            // itself attention-needing, so a PR Crow first sees already in
+            // CHANGES_REQUESTED (or with failing checks) still routes to
+            // auto-respond. Restart-spurious re-fires are prevented by the
+            // persisted dedupe keys at the IssueTracker layer, not here.
+            var out: [PRStatusTransition] = []
+            if new.reviewStatus == .changesRequested {
+                out.append(PRStatusTransition(
+                    kind: .changesRequested,
+                    sessionID: sessionID,
+                    prURL: prURL,
+                    prNumber: prNumber,
+                    headSha: new.headSha,
+                    latestReviewID: new.latestReviewID,
+                    failedCheckNames: []
+                ))
+            }
+            if new.checksPass == .failing {
+                out.append(PRStatusTransition(
+                    kind: .checksFailing,
+                    sessionID: sessionID,
+                    prURL: prURL,
+                    prNumber: prNumber,
+                    headSha: new.headSha,
+                    failedCheckNames: new.failedCheckNames
+                ))
+            }
+            return out
+        }
 
         var out: [PRStatusTransition] = []
 
--- a/Packages/CrowCore/Tests/CrowCoreTests/PRStatusTransitionTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/PRStatusTransitionTests.swift
@@ -33,13 +33,45 @@ struct PRStatusTransitionTests {
     // MARK: - First-observation gate
 
     @Test
-    func firstObservationNeverFires() {
-        // No prior status — even if the PR is already in a transition state,
-        // we must not fire (it'd flood the user with notifications on every
-        // app launch).
-        #expect(compute(from: nil, to: status(review: .changesRequested)).isEmpty)
-        #expect(compute(from: nil, to: status(checks: .failing, sha: shaA)).isEmpty)
-        #expect(compute(from: nil, to: status(review: .changesRequested, checks: .failing, sha: shaA)).isEmpty)
+    func firstObservationOpenWithNoReviewDoesNotFire() {
+        // No prior status AND the new state isn't attention-needing — must not
+        // fire. Protects the startup-flood case the original guard was meant
+        // to cover.
+        #expect(compute(from: nil, to: status(review: .reviewRequired, checks: .pending)).isEmpty)
+        #expect(compute(from: nil, to: status(review: .approved, checks: .passing)).isEmpty)
+    }
+
+    @Test
+    func firstObservationChangesRequestedFires() {
+        // CROW-477: PR was already in CHANGES_REQUESTED when Crow first observed
+        // it (poll missed the open → changes-requested window, or session was
+        // created after the review landed). Must still route to auto-respond.
+        // IssueTracker's persisted dedupe (CROW-456) prevents restart re-fires.
+        let ts = compute(from: nil, to: status(review: .changesRequested, sha: shaA, reviewID: "R_1"))
+        #expect(ts.count == 1)
+        #expect(ts.first?.kind == .changesRequested)
+        #expect(ts.first?.latestReviewID == "R_1")
+        #expect(ts.first?.headSha == shaA)
+        #expect(ts.first?.sessionID == sessionID)
+        #expect(ts.first?.prURL == prURL)
+    }
+
+    @Test
+    func firstObservationFailingChecksFires() {
+        // CROW-477: same idea for failing CI on first observation.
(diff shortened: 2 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

Tight, well-scoped change: relaxes the `old == nil` first-observation guard in `PRStatus.transitions(from:to:…)` so a PR Crow first observes already in `CHANGES_REQUESTED` (or with failing checks) routes to auto-respond, while preserving startup-flood safety for attention-neutral first observations.

### Critical Issues
None.

### Security Review
**Strengths:**
- Pure value-type transition logic — no auth, network, I/O, injection, or secrets surface. Change is confined to in-memory model computation.
- No new external input is trusted; `reviewStatus`/`checksPass`/`latestReviewID` are already-parsed fields from the existing PR payload.

**Concerns:**
- None.

### Code Quality
- **Restart/flood safety holds, and I verified the mechanism** beyond the PR description:
  - The relaxed `old == nil` branch only fires for *genuine* first observations. In `IssueTracker.applyPRStatuses` (`Sources/Crow/App/IssueTracker.swift:1461`), `oldStatus` comes from `previousPRStatus`, which is persisted (`persistTrackerState`, 1538) and rehydrated on startup (`hydratePersistedState`, 1517). So a returning session already in `CHANGES_REQUESTED` takes the `old != nil` path (`entering == false` → no fire), not the new branch.
  - Even if it did reach the new branch, the emitted `dedupeKey` is inserted (1492) and persisted (1504), then rehydrated (1530), suppressing a re-fire after restart. Double protection.
- **Nil-branch logic is consistent with the existing non-nil branch**: `new.reviewStatus == .changesRequested` mirrors `entering`; `new.checksPass == .failing` mirrors the failing-edge check; append order (`changesRequested` then `checksFailing`) matches the documented contract and the `…FiresBoth` test.
- **`latestReviewID` is correctly carried** into the first-observation `.changesRequested` transition, so the `(session, kind, latestReviewID)` dedupe key is populated from the first fire — no spurious round-2 re-fire on the next poll.
- **Tests**: `arch -arm64 swift test … --filter PRStatusTransitionTests` → 18/18 pass locally, including the four new first-observation tests and the retained startup-flood-safety case (`firstObservationOpenWithNoReviewDoesNotFire`). Doc comments are thorough and accurate.

- **One-time fire for pre-existing attention-needing PRs on first track.** For a session whose `previousPRStatus` was never persisted (a brand-new link, or an upgrade from a pre-CROW-456 build with no persisted state), a PR already in `CHANGES_REQUESTED`/`failing` will now fire once on first poll where it previously stayed silent. This is the intended behavior of CROW-477 and is bounded to a single fire by the dedupe key — noting it only so the behavior change is explicit.
- The PR's manual test-plan item (restart with PR still in `CHANGES_REQUESTED` → confirm no re-fire) is still unchecked. The automated tests plus the persisted-dedup path above give good confidence, but a quick manual pass would close the loop.

### Summary Table

---

PR #479, opened 2026-06-10
Terminal UX cluster: remaining gaps from #471 audit

### LINKED ISSUE #471: Terminal UX cluster — remaining gaps from #469 audit

dgershman:
Closes #471.

Implements the six confirmed gaps surfaced by the #469 audit, in one PR per the ticket scope.

## Summary

- **Gap 1 — I-beam cursor.** `resetCursorRects` now registers `.iBeam` over the whole bounds; `.pointingHand` still wins on hovered OSC 8 links because AppKit picks the last-registered rect containing the point.
- **Gap 2 — Cmd+F search.** New `TerminalSearchBar` (NSSearchField + ▲/▼/Done) floats over the cockpit surface, observes a `.terminalBeginSearch` notification posted from `keyDown` (scoped to the container that currently hosts the shared surface), and drives tmux copy-mode. Enter issues a fresh `search-backward`; ▲ / ▼ then step via `search-again` / `search-reverse` without re-running the query. ESC or Done exits copy-mode.
- **Gap 3 — Multi-line paste guard.** `paste(_:)` detects `\n` / `\r` (ignoring a single trailing newline so `"git status\n"` doesn't prompt) and shows an `NSAlert` sheet ("Paste" / "Cancel") before forwarding to libghostty. Single-line pastes are unchanged.
- **Gap 4 — Spacebar Quick Look.** Bare space with a selection that `SmartDetect` recognises (URL or `path:line`) opens `QLPreviewPanel` against the URL or the resolved file URL. Plain-text selections always pass space through to the shell, so a stale selection can't hijack typed input.
- **Gap 5 — Smart-detect URLs and `path:line`.** New pure `SmartDetect` helpers (`detectURL`, `detectFileLine`) feed the right-click "Open URL" / "Open in Editor" menu items. Editor opens are gated to existing regular files (no directories / `.app` bundles), and relative paths resolve against tmux's live `#{pane_current_path}` via new `TmuxBackend.activePaneCwd(id:)` — not the surface's static `workingDirectory` (which is fixed to `$HOME` at create time and never tracks `cd`). Line numbers are intentionally dropped on open (cross-editor line-jump needs a configurable URL scheme; deferred). Unit tests cover the detectors. *Note: Cmd+click smart-detect was considered but dropped — libghostty doesn't expose text-at-point, so the click can't be hit-tested against the detected URL; a global "Cmd+click anywhere opens the selected URL" would override macOS Terminal's selection-extend convention.*
- **Gap 6 — Cmd+↑ / Cmd+↓ prompt jump.** Bound to tmux's `previous-prompt` / `next-prompt` copy-mode commands. The shell wrapper now also emits a bare `OSC 133;A` under tmux (the existing passthrough form still reaches Ghostty), so tmux's emulator can track marks for navigation. Requires tmux ≥ 3.4; older tmux errors are caught and logged non-fatally.

Shared helpers extracted while touching `GhosttySurfaceView.swift`:

- `currentSelectionText()` (used by Copy, smart-detect, Quick Look)
- `forwardPaste(_:)` (used by direct paste and the confirm-sheet handler)

## Test plan

- [ ] `make ghostty && swift build --arch arm64` (passes locally)
- [ ] Hover non-link content → I-beam; hover an OSC 8 link → pointing-hand.
- [ ] Cmd+F → search bar slides in over the active tab only; query a token from scrollback → cursor jumps; ▲/▼ step through matches; ESC closes.
- [ ] Copy `foo\nbar` to clipboard, Cmd+V → confirmation sheet. Copy `git status\n` (single trailing newline) → forwards without prompt.
- [ ] Drag-select random terminal output, hit Space → space reaches the shell (no Quick Look). Select a URL or a `path:line` reference that resolves, hit Space → Quick Look opens against the URL or file.
- [ ] Select `https://github.com` → right-click → "Open URL" enabled. Select `Sources/Foo.swift:42` from a project-dir prompt → right-click → "Open in Editor" enabled (resolves against tmux's live cwd, not `$HOME`).
- [ ] Run a few commands so multiple prompts exist; Cmd+↑ → jumps to previous prompt (enters copy-mode); Cmd+↓ → next prompt; ESC exits copy-mode.
- [ ] Existing #469 / #470 behaviour (right-click menu, OSC 8 hover, click-to-deselect) still works.

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift
@@ -603,6 +603,12 @@ public final class GhosttySurfaceView: NSView {
     @objc private func contextOpenInEditor(_ sender: Any?) {
         guard let item = sender as? NSMenuItem,
               let url = item.representedObject as? URL else { return }
+        // v1 limitation: we resolve the file path but drop the line
+        // number `SmartDetect.detectFileLine` parsed. Jumping to a
+        // specific line is editor-specific (txmt://, vscode://, …) and
+        // there is no portable way to pick one without a configurable
+        // editor preference. Opening the file is still useful; line
+        // navigation can be a follow-up once we add an editor setting.
         NSWorkspace.shared.open(url)
     }
 
@@ -739,27 +745,14 @@ public final class GhosttySurfaceView: NSView {
 
     public override func mouseDown(with event: NSEvent) {
         guard let surface else { return }
-
-        // Cmd+click on a live selection: if it looks like a URL or a
-        // path:line reference, open it via the same dispatchers as the
-        // right-click menu (#471 gap 5). Falls through to the libghostty
-        // mouse path otherwise so plain Cmd+click on text still extends
-        // the selection / forwards to apps that asked for it.
-        if event.modifierFlags.contains(.command),
-           let selection = currentSelectionText() {
-            if let url = SmartDetect.detectURL(
-                in: selection, allowedSchemes: GhosttyApp.allowedURLSchemes
-            ) {
-                NSWorkspace.shared.open(url)
-                return
-            }
-            if let fileLine = SmartDetect.detectFileLine(in: selection),
-               let url = resolveFileURL(path: fileLine.path) {
-                NSWorkspace.shared.open(url)
-                return
-            }
-        }
-
+        // Note: smart-detect (gap 5) is intentionally NOT bound to
+        // Cmd+click here. libghostty's C surface doesn't expose
+        // text-at-point, so we can't verify the click landed on the
+        // detected URL — a global "Cmd+click anywhere opens the
+        // selected URL" would override macOS Terminal's
+        // selection-extend convention and surprise the user. The
+        // right-click menu's "Open URL" / "Open in Editor" items
+        // already cover the selection-based flow.
         sendMousePosition(for: event)
         ghostty_surface_mouse_button(surface, GHOSTTY_MOUSE_PRESS, GHOSTTY_MOUSE_LEFT, translateMods(event.modifierFlags))
     }
--- a/Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift
@@ -655,16 +655,25 @@ public final class GhosttySurfaceView: NSView {
         return true
     }
 
+    // The QLPreviewPanel control hooks are inherited as nonisolated from
+    // NSResponder, but AppKit only ever dispatches them on the main
+    // thread. Hop via `MainActor.assumeIsolated` so we can touch the
+    // surface's main-isolated state (panel.dataSource/.delegate are
+    // themselves `@MainActor`).
     public override func beginPreviewPanelControl(_ panel: QLPreviewPanel!) {
-        panel.dataSource = self
-        panel.delegate = self
+        MainActor.assumeIsolated {
+            panel.dataSource = self
+            panel.delegate = self
+        }
     }
 
     public override func endPreviewPanelControl(_ panel: QLPreviewPanel!) {
-        if panel.dataSource === self { panel.dataSource = nil }
-        if panel.delegate === self { panel.delegate = nil }
-        cleanupQuickLookTempFile()
-        quickLookItem = nil
+        MainActor.assumeIsolated {
+            if panel.dataSource === self { panel.dataSource = nil }
+            if panel.delegate === self { panel.delegate = nil }
+            cleanupQuickLookTempFile()
+            quickLookItem = nil
+        }
     }
 
     /// Resolve a `path:line` detection's path to a file URL by checking
--- a/Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift
@@ -1,5 +1,6 @@
 import AppKit
 import GhosttyKit
+import QuickLookUI
 
 /// NSView subclass that hosts a libghostty terminal surface with Metal rendering.
 public final class GhosttySurfaceView: NSView {
@@ -8,6 +9,13 @@ public final class GhosttySurfaceView: NSView {
     private var markedTextStorage = NSMutableAttributedString()
     private var trackingArea: NSTrackingArea?
     private var hoveredLinkURL: String?
+
+    /// Backing URL for the active Quick Look preview (spacebar, #471 gap 4).
+    /// Either a remote URL pulled directly from the selection or a temp
+    /// text file we wrote to render plain selection text. Cleared and the
+    /// temp file removed when the preview panel closes.
+    private var quickLookItem: NSURL?
+    private var quickLookTempFile: URL?
 
     /// Backoff schedule for retrying createSurface() when ghostty_surface_new returns nil.
     /// 4 retries totalling ~7.5s before declaring permanent failure.
@@ -219,6 +227,11 @@ public final class GhosttySurfaceView: NSView {
     }
 
     public override func resetCursorRects() {
+        // I-beam by default so terminal text feels selectable like macOS
+        // Terminal / iTerm2. Pointing-hand wins on hovered OSC 8 links —
+        // AppKit picks the last-registered cursor whose rect contains the
+        // point, so the order here matters.
+        addCursorRect(bounds, cursor: .iBeam)
         if hoveredLinkURL != nil {
             addCursorRect(bounds, cursor: .pointingHand)
         }
@@ -257,6 +270,45 @@ public final class GhosttySurfaceView: NSView {
 
     public override func keyDown(with event: NSEvent) {
         guard let surface else { return }
+
+        // Cmd+F → open the search overlay (#471 gap 2). Notification is
+        // observed by `TerminalSearchBar`; the bar takes first responder
+        // from us so the user can type immediately.
+        if event.modifierFlags.contains(.command),
+           event.charactersIgnoringModifiers == "f" {
+            NotificationCenter.default.post(name: .terminalBeginSearch, object: nil)
+            return
+        }
+
+        // Cmd+↑ / Cmd+↓ → jump between OSC 133 prompts (#471 gap 6). Lives
+        // before the generic Cmd path so the arrow keys don't reach the
+        // surface (which would scroll a line). 126 = up, 125 = down on
+        // macOS.
+        if event.modifierFlags.contains(.command),
+           event.keyCode == 126 || event.keyCode == 125,
+           let id = TmuxBackend.shared.activeTerminalID {
+            do {
+                if event.keyCode == 126 {
+                    try TmuxBackend.shared.previousPrompt(id: id)
+                } else {
+                    try TmuxBackend.shared.nextPrompt(id: id)
+                }
+            } catch {
+                NSLog("[GhosttySurfaceView] prompt jump failed: \(error)")
+            }
+            return
+        }
+
+        // Spacebar Quick Look on an active selection (#471 gap 4). Bare
+        // space only — modified space (Ctrl+Space, etc.) still goes to
+        // the surface. We deliberately do not consume space when there's
+        // no selection, so typing into the shell is unaffected.
+        if event.charactersIgnoringModifiers == " ",
+           event.modifierFlags.intersection([.command, .control, .option]).isEmpty,
+           ghostty_surface_has_selection(surface) {
+            showQuickLook()
+            return
+        }
 
         // Handle Cmd+V (paste) directly
         if event.modifierFlags.contains(.command), event.charactersIgnoringModifiers == "v" {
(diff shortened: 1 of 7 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Solid, well-documented PR. Six independent gaps land cleanly, the `GhosttySurfaceView` extractions (`currentSelectionText()`, `forwardPaste(_:)`) are good refactors, and `SmartDetect` is correctly kept AppKit-free with real unit tests. tmux subprocess args are passed as `Process.arguments` arrays (no shell interpolation), so the search-query path is injection-safe, and URL opening is gated by `GhosttyApp.allowedURLSchemes`. The new copy-mode helpers mirror the existing `selectAll()` pattern faithfully.

I could not compile-verify locally: `Frameworks/GhosttyKit.xcframework` carries no binary artifact in this checkout (it's produced by `make ghostty`), so `swift build` fails before typechecking. That's an environment prerequisite, not a PR defect — flagging only so the "passes locally" claim isn't taken as independently confirmed here.

### Security Review
**Strengths:**
- tmux search query forwarded as a discrete `Process.arguments` element (`TmuxBackend.searchInScrollback` → `sendKeys(... ["-X", command, query])`) — no command/shell injection surface.
- URL actions allowlisted by scheme in both `SmartDetect.detectURL` and the OSC 8 path; `file://`/`javascript:` etc. are dropped (covered by `detectURLDropsDisallowedSchemes`).
- Quick Look temp file is written under user-private `$TMPDIR` with a UUID name and removed on panel close (`endPreviewPanelControl` / `cleanupQuickLookTempFile`).

**Concerns:** none blocking.

### Code Quality

**Yellow — `.terminalBeginSearch` is an untargeted broadcast across per-tab search bars.** `GhosttySurfaceView.swift:279` posts with `object: nil`, and `TerminalSurfaceView.attach` installs **one `TerminalSearchBar` per container** (`TerminalSurfaceView.swift:102-110`), while each container persists per terminal tab (the cockpit surface is re-parented between them — see the file's own header comment). Every living bar observes the notification (`TerminalSearchBar.swift:53-58`) and flips `isHidden = false`. So Cmd+F in the active tab also un-hides the search bars belonging to every other previously-rendered tab; when you later switch to one of those tabs its bar shows up already open. Functionally the search still routes to `activeTerminalID` so it acts on the right terminal, but the UX leaks across tabs. Suggest targeting the post (e.g. pass the active surface as `object` and have each bar ignore notifications that aren't for the surface it currently hosts), or routing through a single shared bar.

**Yellow — Cmd+click opens the URL regardless of where you click.** `mouseDown` (`GhosttySurfaceView.swift:743-759`) opens the detected URL/file whenever a selection exists that parses as one and the Command modifier is down — the click location is never compared against the selection's on-screen rect. So Cmd+clicking anywhere in the surface while a URL happens to be selected opens it (and preempts any other Cmd+click behavior such as extending the selection). Terminal.app/iTerm require the click to land on the link itself. Consider gating on the click hitting the selection, or at least documenting this as "Cmd+click while a URL is selected" so the divergence from the right-click affordance is intentional.

### Consider (Green)
- **Detected line number is parsed then discarded.** `detectFileLine` returns `(path, line)`, but both `contextOpenInEditor` and the Cmd+click path call `NSWorkspace.shared.open(fileURL)` with no line component, so "Open in Editor" opens the file but never jumps to the line. Reasonable v1 limitation (line-jump needs editor-specific URLs), but worth a code comment so the unused `line` doesn't read as a bug.
- **Search-bar button semantics are slightly asymmetric.** `findPrevious` re-runs a fresh `search-backward` when the query is non-empty, while `findNext` always issues `search-reverse` ignoring the field; clicking ▼ before any search has started is a no-op. Works, but the two directions aren't mirror images — a brief comment or unifying on `search-again`/`search-reverse` would help future readers.
- **Synchronous tmux subprocess on the main thread per keystroke.** `searchInScrollback` / `sendPromptNav` shell out via `ctrl.run`/`sendKeys` synchronously on the main thread (Cmd+F Enter, Cmd+↑/↓). This matches the existing `selectAll()` pattern and is guarded by `reportIfTimeout`, so not new debt — just noting that repeated prompt-jumps block the UI for the subprocess duration.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 3 Green] findings. Both Yellow items are cross-tab/UX correctness issues rather than functional breakage; addressing them (or explicitly confirming the Cmd+click design is intentional and scoping the search-bar broadcast) in this round avoids a follow-up.

---

dhilgaertner (changes requested):
## Code & Security Review

Thorough, well-commented PR — the six gaps are cleanly separated, `SmartDetect` is pleasingly pure (its tests compile and pass standalone), and the security posture on URL handling is good: both the menu and Cmd+click gate on `GhosttyApp.allowedURLSchemes`, and `tmux send-keys -X search-backward <query>` passes the query as a discrete process arg (no shell), so there's no injection surface. A few things should be addressed before merge.

> Note: I could not run a full `swift build`/test pass — `Frameworks/GhosttyKit.xcframework` is absent in this checkout (needs `make ghostty`). Findings below are from static review plus a standalone compile+run of `SmartDetect`.

### Security Review
**Strengths:**
- URL open paths (right-click "Open URL", Cmd+click) are gated on the same `allowedURLSchemes` allowlist as OSC 8 clicks — `file://`, `javascript:`, etc. are dropped. Unit test `detectURLDropsDisallowedSchemes` covers this.
- tmux search query is passed as an argv element, not interpolated into a shell string — no command injection.
- `detectFileLine` paths are only ever handed to `NSWorkspace.open` after a `fileExists` check, and only on explicit user action; no shell, no traversal hazard.

**Concerns:** none blocking.

### Code Quality

**Yellow — bare-space Quick Look can swallow legitimate typed input** (`GhosttySurfaceView.swift:306-311`)
The guard fires on any bare space whenever `ghostty_surface_has_selection(surface)` is true, *before* the key is forwarded. A common flow — drag-select text with the mouse, then keep typing at the shell — means the next space the user types is intercepted and opens Quick Look instead of reaching the shell, because the keystroke hasn't yet produced the output that would clear the selection. Finder's spacebar-preview analogy doesn't carry cleanly into a text-input context where space is a constant keystroke. Please either confirm libghostty clears the selection synchronously on key input (in which case the window is closed), or gate this so it doesn't hijack typing (e.g. skip when the alt-screen/TUI app is active, or require the selection to have been made via a non-keyboard path). As written this can intercept normal input.

**Green — "Open in Editor" silently discards the detected line number** (`GhosttySurfaceView.swift:597-605`, `resolveFileURL`)
`detectFileLine` computes `line`, but `contextOpenInEditor` / Cmd+click open the bare file URL via `NSWorkspace.open`, landing at line 1 in the default app. The whole point of parsing `:LINE` is to jump there. Cross-editor line-jumping is genuinely hard, so file-only is a defensible v1 — but consider an editor URL that honors the line (`vscode://file/<path>:<line>`, `xed -l <line>`) or at least a code comment noting the line is intentionally dropped, so the discarded value doesn't read as a bug later.

**Green — `.terminalBeginSearch` is an unscoped global notification** (`TerminalSearchBar.swift:52-57`, `handleBeginSearch`)
Every `TerminalSearchBar` instance observes it with no `object` filter, and search ops use the global `TmuxBackend.shared.activeTerminalID` rather than the bar's own host terminal. This is fine today because `SessionDetailView` renders only the single active terminal (`.id(terminal.id)`), so one bar exists at a time. It would break (multiple bars appear; wrong-terminal search) the moment two surfaces are visible at once — split view, a second window. Worth scoping the notification to a terminal id defensively.

**Green — multi-line paste prompts on a lone trailing newline** (`GhosttySurfaceView.swift`, `paste(_:)`)
`content.contains("\n") || content.contains("\r")` triggers the sheet even for `"git status\n"` copied with its trailing newline — arguably correct (that newline *is* the submit), but iTerm2/Terminal generally only warn on genuine multi-line payloads. Consider trimming a single trailing newline before the check to cut false positives.

**Green — Quick Look temp file leaks if the app exits with the panel open**
`crow-quicklook-<uuid>.txt` is cleaned up in `endPreviewPanelControl` / on the next `showQuickLook`, but not if the process dies while a preview is open. It's under `$TMPDIR` so the OS reaps it; noting for completeness.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 4 Green] findings. The only blocker is the bare-space Quick Look swallowing typed input (`GhosttySurfaceView.swift:306-311`); the greens are low-cost cleanups worth folding into the same round trip.

---

dhilgaertner (changes requested):
## Code & Security Review

Static review only — I could not compile: `Frameworks/GhosttyKit.xcframework` is produced by `make ghostty` and isn't present in the worktree, so `swift build` / the `SmartDetectTests` can't run here. The author reports a clean local build, and the prior commit already addressed the strict-actor-isolation issues on the `QLPreviewPanel` overrides.

Overall this is high-quality work: clearly commented, defensively written, good separation (pure `SmartDetect`, shared `currentSelectionText()` / `forwardPaste()` helpers), and the shared-cockpit-surface re-parenting model is respected throughout. Two "should fix" items keep it from an approve.

### Security Review
**Strengths:**
- URL detection is scheme-gated against the same tight allowlist as OSC 8 (`http/https/mailto/ftp/ftps`); `detectURLDropsDisallowedSchemes` proves `file://` is rejected, so a hostile program can't surface a `file://`/`x-apple-*` "Open URL".
- Multi-line paste guard (`\n`/`\r` → confirmation sheet) is a genuine safety improvement against paste-to-execute, matching Terminal/iTerm2 convention.
- Quick Look temp files are per-`UUID`, written `atomically`, and removed in `endPreviewPanelControl` / before re-staging; no predictable-path or overwrite exposure.
- `resolveFileURL` only enables "Open in Editor" for paths that actually exist, so no bogus path is handed to `NSWorkspace`.

**Concerns:**
- 🟢 `contextOpenInEditor` ultimately calls `NSWorkspace.shared.open(fileURL)`, which opens the file in its *default app* — for an app bundle that means launching it (e.g. a selected `/Applications/Foo.app:1` resolves and launches). It's user-initiated, existence-checked, and low risk, but "Open in Editor" launching arbitrary bundles is a mild surprise. Consider restricting to regular files or routing through a configured editor.

### Code Quality
- 🟡 **`path:line` relative resolution is effectively dead for the common case.** The shared cockpit surface is created once with `workingDirectory: NSHomeDirectory()` (`TmuxBackend.swift:577`) and `workingDirectory` is never updated as the shell `cd`s. So `resolveFileURL`'s cwd-relative branch (`GhosttySurfaceView.swift:695`) always joins against `$HOME`. The headline use case for gap 5 — a build/test line like `Sources/Foo.swift:42` emitted from a project dir — will not resolve, so "Open in Editor" stays disabled for relative paths and effectively only works for absolute paths. tmux already exposes the live cwd via `#{pane_current_path}` for the active window; resolving against that (or the active terminal's tracked cwd) would make the feature work as advertised.
- 🟡 **PR description / test plan disagree with the code on Cmd+click.** The summary (gap 5) says "a Cmd+click handler on the selection," and the test plan checkbox says "Cmd+click on the selection opens directly," but `mouseDown` (`GhosttySurfaceView.swift:746`) intentionally does *not* bind smart-detect to Cmd+click (with a good rationale). The QA step as written can't pass. Update the description/test plan to match the shipped behavior.
- 🟢 Both `SmartDetect.detectURL` (NSDataDetector) and `detectFileLine` (anchored regex) run synchronously on the main thread while building the right-click menu over the *entire* current selection. For normal selections this is fine; a pathological single-token selection (many `.`, no whitespace, anchored `^…$`) could backtrack. Bounded in practice — interior whitespace fails the anchors fast — so just a note.
- 🟢 Spacebar consumes the keystroke whenever a selection exists (`ghostty_surface_has_selection`), so a lingering selection means the next bare space opens Quick Look instead of reaching the shell. Matches the Finder convention but worth being aware of.
- 🟢 Double `OSC 133;A` emission (bare for tmux's `previous-prompt`/`next-prompt`, wrapped passthrough for Ghostty) is the right call and well-commented. Note `previous-prompt`/`next-prompt` need tmux ≥ 3.4; the `catch`/`NSLog` keeps older tmux non-fatal, which is fine.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 3 Green] findings. The relative `path:line` resolution gap and the description/test-plan mismatch are both quick to address and land in the same round trip.

---

dhilgaertner (approved):
## Code & Security Review

Fourth-round review. The three prior CHANGES_REQUESTED rounds have all been addressed in the follow-up commits (`bf20246`, `75ea40d`), and a fresh full read of the current tree turns up nothing above Green.

**Prior findings, now resolved:**
- 🟡→✅ **Search broadcast scoping.** `handleBeginSearch` (`TerminalSearchBar.swift:94-107`) now gates on `surface.superview === self.superview`, so only the bar in the container currently hosting the re-parented cockpit surface un-hides. The cross-tab leak is gone.
- 🟡→✅ **Cmd+click opening URLs from anywhere.** Cmd+click smart-detect was removed entirely; `mouseDown` (`GhosttySurfaceView.swift:771-783`) documents why (no text-at-point in libghostty), and the PR description/test plan now match the shipped right-click-only flow.
- 🟡→✅ **Bare-space Quick Look swallowing typed input.** The space intercept (`GhosttySurfaceView.swift:308-314`) is now gated on `selectionHasQuickLookCandidate()` — it only fires when the *entire* selection parses as an allowed-scheme URL or a resolvable `path:line`, so ordinary typing against a stale selection passes through to the shell.
- 🟡→✅ **`path:line` relative resolution against `$HOME`.** `resolveFileURL` (`GhosttySurfaceView.swift:708-721`) now joins relative paths against `TmuxBackend.activePaneCwd(id:)` (live `#{pane_current_path}`) instead of the static `workingDirectory`, so build/test references like `Sources/Foo.swift:42` resolve from the project dir as advertised.
- 🟢→✅ **"Open in Editor" launching app bundles / directories.** `isRegularFile` (`GhosttySurfaceView.swift:727-733`) rejects directories and `.app` bundles, so only plain files enable the item.
- 🟢→✅ Line-number drop, multi-line-paste trailing-newline false positive, and search-button asymmetry are all now documented or handled.

### Security Review
**Strengths:**
- tmux search query is passed as a discrete `Process.arguments` element (`searchInScrollback` → `sendKeys(target:keys: ["-X", command, query])` → `run([...])`), never interpolated into a shell string — no command-injection surface. Same for prompt-nav / copy-mode commands.
- URL opens (right-click "Open URL", OSC 8 "Open Link", Quick Look) are scheme-gated against the same `GhosttyApp.allowedURLSchemes` allowlist; `detectURLDropsDisallowedSchemes` proves `file://` is rejected.
- "Open in Editor" only enables for existence-checked **regular** files, and Quick Look temp handling is unchanged from the previously-cleared design.

**Concerns:** none.

### Code Quality
All referenced backend symbols exist and the diff is internally consistent (`displayMessage`, `sendKeys`, `run`, `sessionName`, `TmuxBackendError.unknownTerminal`, `reportIfTimeout`). The `SmartDetect` unit tests cover URL allow/deny, first-match, and the `path:line` accept/reject cases.

> Build not independently verified: `Frameworks/GhosttyKit.xcframework` carries no binary artifact in this checkout (produced by `make ghostty`), so `swift build` / `SmartDetectTests` can't run here. The author reports a clean local build; flagging only so "passes locally" isn't read as confirmed here.

### Consider (Green)
- **Synchronous tmux subprocess on the space keystroke.** When a selection exists and parses as a `path:line` (URL check fails first), `selectionHasQuickLookCandidate()` calls `resolveFileURL` → `activePaneCwd`, which shells out to `tmux display-message` synchronously on the main thread for that keypress. Narrow trigger and consistent with the existing `selectAll()` pattern (guarded by `reportIfTimeout`), so not new debt — just noting the keystroke-path subprocess.
- **Detection runs twice on a spacebar Quick Look.** `selectionHasQuickLookCandidate()` and then `showQuickLook()` each re-read the selection and re-run the detectors (and `resolveFileURL` → another `activePaneCwd` subprocess for `path:line`). Could thread the resolved item through, but the cost is trivial in practice.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. All three prior rounds' Yellow items are resolved; the remaining notes are optional micro-optimizations.

---

PR #483, opened 2026-06-11
Add Corveil CLI path picker + auto-install /query-corveil on launch

### LINKED ISSUE #482: Settings: Corveil CLI path picker + Scaffolder runs 'corveil skill install' on launch
**Goal:** make Crow pick up the Corveil CLI from a user-configured path, and auto-install Corveil's embedded `query-corveil` Claude Code slash command into the devroot on every app launch. No skill source in `crow/skills/` — corveil owns the skill content and Crow only orchestrates installation.

## Why this shape

The Corveil team already shipped the `query-corveil` skill **embedded inside the corveil binary** (`corveil/go/cmd/corveil/skill/query-corveil.md`, embedded via `//go:embed` in `corveil/go/cmd/corveil/skill.go`). The binary exposes an installer:

```
corveil skill install [--path <dest>]
```

Default destination is `~/.claude/commands/query-corveil.md`. The embedded content is the canonical, current version (uses `corveil stats`/`corveil search`/`corveil context` invocations, documents auth setup with `CORVEIL_URL`/`CORVEIL_API_KEY` or `corveil login`, lists admin subcommands and "not yet available" commands). It evolved from `radiusmethod/ontology-prototype:.claude/commands/query-ontology.md` but is now newer and corveil-flavored.

So Crow shouldn't ship its own copy — it would immediately drift. Crow's job is to (a) know where the user's corveil binary lives and (b) invoke the binary's own installer on launch.

## Deliverable 1 — Settings menu picker

Add a "Corveil CLI" entry to Crow's Settings UI:

- A text field showing the currently-configured path.
- A **Browse…** button that opens `NSOpenPanel` with `allowedContentTypes = [.executable, .unixExecutable]`, `canChooseFiles = true`, `canChooseDirectories = false`.
- On selection, persist to `{devRoot}/.claude/config.json` at `defaults.binaries.corveil`. Use the **generic `binaries` map** shape so future tools (soulstone, tanzanite, …) extend the same field without a schema change:
  ```json
  {
    "defaults": {
      "binaries": {
        "corveil": "/Users/danny/Projects/corveil/go/cmd/corveil/corveil"
      }
    }
  }
  ```
- A small "Verify" affordance below the field — runs `<path> --version` (or similar), shows the output inline so the user knows the picked binary actually works.
- Empty/unset is a valid state. Don't force a value.

The settings change should hot-update — no app restart needed for the new path to take effect on the next Scaffolder run.

## Deliverable 2 — Scaffolder hook

In `Sources/Crow/App/Scaffolder.swift` (mirrors the existing scaffold-skill pattern around `loadFromRepo("skills/X/SKILL.md")`):

1. After the existing skill scaffolding, read `defaults.binaries.corveil` from config.
2. If unset → no-op.
3. If set but the file isn't executable → log a one-line warning, no-op (don't block scaffolding).
4. If set and executable → run:
   ```
   <corveilPath> skill install --path <devRoot>/.claude/commands/query-corveil.md
   ```
   Capture stdout/stderr. On non-zero exit, log the failure + surface a transient banner on launch like "Corveil skill install failed — check path in Settings" (similar to existing surface for other non-fatal scaffold issues). Do **not** fail Crow startup.

The `--path` argument scopes the install per-devroot, matching how `Scaffolder` already scopes other writes to `{devRoot}/.claude/`. Without `--path`, corveil's installer defaults to `~/.claude/commands/query-corveil.md` (user-global), which doesn't match Crow's devroot-isolation pattern.

## Deliverable 3 — Re-install on every launch

Just run the install command unconditionally on each Crow launch (when the binary path is configured). Cost: one fast subprocess. Benefit: whenever the user rebuilds corveil or pulls a new version, the embedded skill updates automatically — no stale-skill drift.

The corveil installer is idempotent (`os.WriteFile` + `os.Chmod` per `skill.go`); re-running is safe.

## What's explicitly NOT in this ticket

- **A copy of `query-corveil.md` in `crow/skills/`.** The skill source lives in corveil. Don't mirror it.
- **Crow-side fallback to `command -v corveil` on PATH.** Picker-only for v1. If the user wants a PATH-installed corveil, they point the picker at `$(which corveil)` once. Adding fallback is +20 lines of code for a corner case; revisit if needed.
- **UI for editing the rest of `defaults.binaries`.** This ticket adds the schema field via the `corveil` entry; full Settings UI for the generic map can come when there's a second tool to manage.
- **`crow-create-corveil-task`-style skills that wrap `corveil task create`.** Those are Crow-workflow-specific (parallel to `crow-create-ticket`) and CAN live in `crow/skills/`. Separate ticket once the picker + Scaffolder hook ship and prove the basic flow works.
- **macOS bundling of the corveil binary inside `Crow.app`.** Out of scope; user picks their locally-built copy. Bundling is a future distribution concern.

## Acceptance

1. Crow Settings has a "Corveil CLI" field with a working file picker; selection persists to `defaults.binaries.corveil` in `config.json`.
2. The "Verify" affordance runs the binary and shows a non-error result for a valid corveil binary.
3. On Crow launch, when the path is configured and points at a valid corveil binary, the file `{devRoot}/.claude/commands/query-corveil.md` exists and matches the output of `<corveilPath> skill show` byte-for-byte (proves the embedded skill installed correctly).
4. Running `/query-corveil` in a Claude Code session under that devroot loads the skill (the slash command is registered).
5. With the path unset, Crow launches cleanly without errors and the slash command is not installed.
6. With the path set but pointing at a non-existent or non-executable file, Crow launches cleanly, logs a warning, and surfaces a non-blocking banner. No crash, no startup failure.

## Critical files

| Purpose | Path |
|---|---|
| Scaffold pattern to mirror | `crow/Sources/Crow/App/Scaffolder.swift:5-46` |
| Corveil's installer source (do not duplicate) | `corveil/go/cmd/corveil/skill.go` |
| Corveil's embedded skill source (do not duplicate) | `corveil/go/cmd/corveil/skill/query-corveil.md` |
| Config schema lives here today | `crow/Packages/CrowCore/Sources/CrowCore/Models/CrowConfig.swift` (or wherever the model is — the `defaults` block) |
| Settings UI lives here today | (wherever the Settings tab/sheet is implemented in `Sources/Crow/UI/`) |

## Related

- The earlier audit at `/Users/danny/.claude/plans/hidden-dancing-mango.md` (TaskBackend abstraction) — same Corveil-integration thread, different axis. This ticket complements that work without depending on it.
- ADR 0005 (TaskBackend / CodeBackend split) — `StubCorveilTaskBackend` is the in-Crow side of corveil integration; this ticket is the CLI/skill side. The two land independently.
- Corveil PR #1201 — landed the `corveil skill install` subcommand + embedded skill.

dgershman:
## Summary

- Adds a generic `defaults.binaries` map (keyed by tool name) to `ConfigDefaults`, with a Settings → General → "Corveil CLI" picker for the `corveil` slot (file-only NSOpenPanel + Verify button that runs `<path> --version`).
- On every Crow launch, `Scaffolder` invokes `<corveilPath> skill install --path {devRoot}/.claude/commands/query-corveil.md` so the embedded slash command stays in sync with the user's locally-built corveil — no skill source in `crow/skills/`, corveil owns the content.
- Failures are surfaced via a new `appState.corveilSkillInstallWarning` banner using the existing `<X>Warning` pattern (matches `githubScopeWarning`/`rateLimitWarning`). Unset path = no-op. Non-executable or non-zero exit = log + transient Settings banner, never blocks startup.

Closes #482

## Test plan

- [ ] `swift build --arch arm64` succeeds (verified — clean compile)
- [ ] Launch Crow with the corveil path unset → no `query-corveil.md` written, no banner, no crash.
- [ ] Open Settings → General → Corveil CLI, browse to a real corveil binary → path persists to `config.json` under `defaults.binaries.corveil`.
- [ ] Click "Verify" → green `✓ corveil version ...` line.
- [ ] Restart Crow → `{devRoot}/.claude/commands/query-corveil.md` exists and `diff <(corveil skill show) <(cat ...)` is empty.
- [ ] `/query-corveil` slash command loads in a Claude Code session under that devroot.
- [ ] Point picker at `/tmp/not-corveil` → relaunch shows the orange Settings banner; no `query-corveil.md` written.
- [ ] Idempotency: relaunch repeatedly, file content is byte-identical each time.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -233,6 +233,12 @@ public final class AppState {
     /// Set by `IssueTracker` when polling is suspended; cleared on next success.
     public var rateLimitWarning: String?
 
+    /// Non-fatal warning surfaced in Settings when the per-launch
+    /// `corveil skill install` run fails (CROW-482). `nil` means the install
+    /// either succeeded or wasn't attempted (no path configured). Set by
+    /// `AppDelegate.launchMainApp` from the `Scaffolder` result.
+    public var corveilSkillInstallWarning: String?
+
     /// Terminal readiness state per terminal ID.
     public var terminalReadiness: [UUID: TerminalReadiness] = [:]
 
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -406,11 +406,22 @@ public struct ConfigDefaults: Codable, Sendable, Equatable {
     public var excludeReviewRepos: [String]
     public var excludeTicketRepos: [String]
     public var ignoreReviewLabels: [String]
-    /// Explicit absolute-path overrides for `CodingAgent` binary discovery,
-    /// keyed by `AgentKind.rawValue` (e.g. `"codex"`, `"cursor"`, `"claude-code"`).
-    /// Consulted before the PATH walk in `CodingAgent.findBinary()` — set this
-    /// when discovery doesn't find your install for any reason (exotic Node
-    /// manager, sandboxed PATH, etc.). See CROW-484.
+    /// Absolute-path overrides for executable binaries, keyed by tool name.
+    ///
+    /// Serves two callers that share the same map shape:
+    /// - **Agent binary discovery** (CROW-484): keyed by `AgentKind.rawValue`
+    ///   (`"codex"`, `"cursor"`, `"claude-code"`). `CodingAgent.findBinary()`
+    ///   consults this map before walking PATH — set this when discovery
+    ///   doesn't find your install (exotic Node manager, sandboxed PATH, etc.).
+    /// - **External tool installers** (CROW-482): keyed by tool name (e.g.
+    ///   `"corveil"`) and used by `Scaffolder` to run each tool's own skill
+    ///   installer on launch. The Settings UI currently exposes only the
+    ///   `corveil` slot; the map shape is intentionally generic so future
+    ///   tools (soulstone, tanzanite, …) extend the same field without a
+    ///   schema change.
+    ///
+    /// Agent keys (`claude-code`, `codex`, `cursor`) and tool keys (`corveil`,
+    /// …) don't overlap, so the two callers coexist in one map.
     public var binaries: [String: String]
 
     /// Characters that are invalid in git ref names (see `git check-ref-format`).
--- a/Packages/CrowCore/Tests/CrowCoreTests/AppConfigTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AppConfigTests.swift
@@ -485,6 +485,31 @@ import Testing
     #expect(config.defaults.ignoreReviewLabels.isEmpty)
 }
 
+// MARK: - defaults.binaries (CROW-482)
+
+@Test func binariesRoundTrip() throws {
+    let config = AppConfig(
+        defaults: ConfigDefaults(binaries: [
+            "corveil": "/Users/jane/dev/corveil/corveil",
+            "soulstone": "/usr/local/bin/soulstone",
+        ])
+    )
+    let data = try JSONEncoder().encode(config)
+    let decoded = try JSONDecoder().decode(AppConfig.self, from: data)
+    #expect(decoded.defaults.binaries["corveil"] == "/Users/jane/dev/corveil/corveil")
+    #expect(decoded.defaults.binaries["soulstone"] == "/usr/local/bin/soulstone")
+}
+
+@Test func binariesDefaultsEmptyWhenKeyMissing() throws {
+    // Configs written before CROW-482 don't have `binaries` — they must still
+    // decode cleanly with an empty map (forward compatibility).
+    let json = """
+    {"defaults": {"provider": "github", "cli": "gh", "branchPrefix": "feature/", "excludeDirs": []}}
+    """.data(using: .utf8)!
+    let config = try JSONDecoder().decode(AppConfig.self, from: json)
+    #expect(config.defaults.binaries.isEmpty)
+}
+
 // MARK: - AI gateway (CROW-402)
 
 @Test func workspaceGatewayRoundTrip() throws {
--- a/Packages/CrowUI/Sources/CrowUI/SettingsView.swift
+++ b/Packages/CrowUI/Sources/CrowUI/SettingsView.swift
@@ -667,17 +667,6 @@ public struct SettingsView: View {
             errDrain.abandon()
             return "✗ Could not launch: \(error.localizedDescription)"
         }
-        // Close the parent's copies of the pipe write ends so EOF arrives at
-        // our read ends when the child exits. Foundation's `Process` only
-        // closes them as part of `waitUntilExit()`; this code uses a polling
-        // loop on `isRunning` instead, so if we don't close them ourselves
-        // the drain threads' `readDataToEndOfFile()` will block forever
-        // waiting for an EOF that never comes (which would show up as empty
-        // captured output even when the child wrote bytes to its stdout).
-        // The child has its own dup'd fd, so closing the parent's copy is
-        // safe — it doesn't affect the child's ability to write.
-        try? outPipe.fileHandleForWriting.close()
-        try? errPipe.fileHandleForWriting.close()
 
         // Wall-clock timeout. Poll in 50ms slices so a hung binary is
         // SIGTERM'd rather than blocking forever; a 500ms grace lets the
@@ -696,6 +685,15 @@ public struct SettingsView: View {
             }
             Thread.sleep(forTimeInterval: 0.05)
         }
+        // `Process` keeps an internal copy of each pipe's write FD that is
+        // only released as part of `waitUntilExit()`'s cleanup. The polling
+        // loop above leaves that internal copy open, so the bg drainers'
+        // `readDataToEndOfFile()` never sees EOF and the snapshot below
+        // returns empty (the test failure that flagged this). `waitUntilExit`
+        // is a no-op for the wait itself at this point (the child is already
+        // gone or being torn down) but its cleanup releases Foundation's FD
+        // copy so EOF is finally delivered to our read ends.
+        proc.waitUntilExit()
 
         let outStr = String(data: outDrain.collect(within: pipeDrainGrace), encoding: .utf8)?
             .trimmingCharacters(in: .whitespacesAndNewlines) ?? ""
--- a/Packages/CrowUI/Sources/CrowUI/SettingsView.swift
+++ b/Packages/CrowUI/Sources/CrowUI/SettingsView.swift
@@ -633,17 +633,15 @@ public struct SettingsView: View {
     /// actor and trivially testable. Returns a single-line summary suitable
     /// for inline display.
     ///
-    /// Bounded by `verifyTimeout` so a misbehaving binary that hangs on
-    /// `--version` can't pin the detached Task forever (matches the
-    /// `Scaffolder.installCorveilSkill` install-path timeout). Each pipe is
-    /// drained by a dedicated background reader running
-    /// `readDataToEndOfFile()`, which blocks until the child closes its end.
-    /// Snapshotting waits on a bounded `DispatchSemaphore`, so there's no
-    /// race between the polling thread and an in-flight readability handler:
-    /// either the drain finished cleanly (signal arrived → snapshot is whole)
-    /// or the bounded wait expires and we report what's there. Combined,
-    /// these eliminate both the pipe-buffer deadlock and the EOF tail race
-    /// that `readabilityHandler` introduces.
+    /// Uses `proc.waitUntilExit()` (with a `TimeoutWatchdog` SIGTERM'ing the
+    /// child if it hangs) rather than a polling loop on `proc.isRunning`.
+    /// `waitUntilExit` is the only way to deterministically trigger
+    /// Foundation's pipe-write-FD cleanup; a polling loop leaves Foundation's
+    /// internal copy of the writeFD open and the post-exit reads either hang
+    /// (with a background drain) or return empty (Foundation closes its copy
+    /// on its own internal schedule). Once `waitUntilExit` returns, both
+    /// pipe writers (child + Foundation) have closed, so a synchronous
+    /// `readToEnd()` on each pipe returns immediately with the data.
     nonisolated static func runCorveilVersion(at path: String) -> String {
         let fm = FileManager.default
         guard fm.isExecutableFile(atPath: path) else {
@@ -657,48 +655,22 @@ public struct SettingsView: View {
         proc.standardOutput = outPipe
         proc.standardError = errPipe
 
-        let outDrain = PipeDrainer.start(outPipe)
-        let errDrain = PipeDrainer.start(errPipe)
-
         do {
             try proc.run()
         } catch {
-            outDrain.abandon()
-            errDrain.abandon()
             return "✗ Could not launch: \(error.localizedDescription)"
         }
 
-        // Wall-clock timeout. Poll in 50ms slices so a hung binary is
-        // SIGTERM'd rather than blocking forever; a 500ms grace lets the
-        // process honor SIGTERM before we drop it.
-        let deadline = Date().addingTimeInterval(verifyTimeout)
-        var timedOut = false
-        while proc.isRunning {
-            if Date() >= deadline {
-                proc.terminate()
-                let graceDeadline = Date().addingTimeInterval(0.5)
-                while proc.isRunning, Date() < graceDeadline {
-                    Thread.sleep(forTimeInterval: 0.05)
-                }
-                timedOut = true
-                break
-            }
-            Thread.sleep(forTimeInterval: 0.05)
-        }
-        // `Process` keeps an internal copy of each pipe's write FD that is
-        // only released as part of `waitUntilExit()`'s cleanup. The polling
-        // loop above leaves that internal copy open, so the bg drainers'
-        // `readDataToEndOfFile()` never sees EOF and the snapshot below
-        // returns empty (the test failure that flagged this). `waitUntilExit`
-        // is a no-op for the wait itself at this point (the child is already
-        // gone or being torn down) but its cleanup releases Foundation's FD
-        // copy so EOF is finally delivered to our read ends.
+        // Watchdog: SIGTERM after `verifyTimeout` so a hung binary unblocks
+        // `waitUntilExit` below. The watchdog also records the timeout so we
+        // can distinguish a normal exit-N from a wall-clock kill.
+        let watchdog = TimeoutWatchdog(deadline: verifyTimeout, proc: proc)
+        watchdog.start()
         proc.waitUntilExit()
+        let timedOut = watchdog.cancel()
 
-        let outStr = String(data: outDrain.collect(within: pipeDrainGrace), encoding: .utf8)?
-            .trimmingCharacters(in: .whitespacesAndNewlines) ?? ""
-        let errStr = String(data: errDrain.collect(within: pipeDrainGrace), encoding: .utf8)?
-            .trimmingCharacters(in: .whitespacesAndNewlines) ?? ""
+        let outStr = Self.readAll(outPipe)
+        let errStr = Self.readAll(errPipe)
         let combined = [outStr, errStr].filter { !$0.isEmpty }.joined(separator: " — ")
         let snippet = combined.split(separator: "\n").first.map(String.init) ?? combined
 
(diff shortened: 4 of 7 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Clean, well-documented feature. The generic `defaults.binaries` map with `decodeIfPresent` backward-compat, the off-main `Task.detached` Verify path, and the non-fatal `ScaffoldResult.warning` surfaced via the established `<X>Warning` banner pattern are all done right. Two should-fix items below.

### Critical Issues
None.

### Security Review
**Strengths:**
- No shell involved — `Process` is invoked with `executableURL` + a fixed `arguments` array (`["skill","install","--path",target]` / `["--version"]`), so a path or target with spaces/metacharacters cannot inject. Arg-injection safe.
- `isExecutableFile(atPath:)` gate before exec; non-executable path is a no-op warning, never a crash.
- The binary is the user's own locally-configured CLI — no trust-boundary crossing.

**Concerns:**
- None.

### Code Quality

🟡 **Launch-path subprocess blocks the main thread with no timeout** — `Sources/Crow/App/Scaffolder.swift:180`. `installCorveilSkill` calls `proc.waitUntilExit()` synchronously, and `scaffold(...)` runs on the main thread during `applicationDidFinishLaunching → launchMainApp` (`AppDelegate.swift:85,316,403`). A corveil binary that hangs (blocks on stdin, opens a prompt, wrong executable) freezes app startup indefinitely — with no window yet drawn and the single-instance lock already held, the user has no escape. Note the inconsistency: the *Verify* button correctly uses `Task.detached` to avoid blocking the UI (`SettingsView.swift`), yet the riskier launch path runs unguarded. The `isExecutableFile` gate prevents a missing/non-exec binary but does not bound runtime. Suggest either running the install off the main actor (and assigning `corveilSkillInstallWarning` via `MainActor.run`) or adding a wall-clock timeout that terminates the process and returns a warning.

🟡 **No test coverage for the new `binaries` field or the corveil helpers** — the field added immediately before this one, `ignoreReviewLabels`, has both `ignoreReviewLabelsRoundTrip()` and `ignoreReviewLabelsDefaultsEmptyWhenKeyMissing()` (`Packages/CrowCore/Tests/CrowCoreTests/AppConfigTests.swift:441,450`). `binaries` has neither, so the round-trip and the `decodeIfPresent ?? [:]` backward-compat path are unverified. `runCorveilVersion(at:)` is `nonisolated static` and described in its own doc comment as "trivially testable," yet has no test. Adding a round-trip + missing-key test (matching the established convention) and a couple of `runCorveilVersion` cases (non-executable path, exit-code formatting) would close the gap.

🟢 **corveil install runs twice on first-time setup** — `completeSetup` runs `scaffold(...)` (`AppDelegate.swift:303`) and then calls `launchMainApp()` which runs `scaffold(...)` again (`AppDelegate.swift:403`), so the subprocess fires twice on the initial-setup launch. Harmless given idempotency, but the line-308 assignment is immediately superseded by line 408.

🟢 The `corveilBinding` empty-string-removes-key handling and trim-on-commit are a nice touch that keeps the persisted map clean.

### Summary Table

---

dhilgaertner (changes requested):
## Code & Security Review

Solid, well-documented feature. Forward-compatible config decode, non-fatal failure handling, a wall-clock timeout on the startup subprocess, and good test coverage. One Yellow asymmetry keeps this from an approve.

### Critical Issues
None.

### Security Review
**Strengths:**
- No shell injection: `Process.executableURL` + fixed argument array (`["skill","install","--path",target]`), never routed through a shell.
- `target` is a fixed, derived path (`{devRoot}/.claude/commands/query-corveil.md`) — no traversal from user input.
- Executable-bit gate (`isExecutableFile`) before launch; unset/empty path is a clean no-op.

**Concerns:**
- The configured `corveil` binary is auto-executed on every launch (`skill install`) and on demand (`--version`). This is by design and user-configured, so it's acceptable — worth being aware that a tampered `config.json` becomes code execution at launch, but that's already true of the devRoot/agent config surface. (Green)

### Code Quality
- 🟡 **`runCorveilVersion` / `verifyCorveil` (SettingsView.swift) has no timeout, unlike the Scaffolder path.** `installCorveilSkill` deliberately bounds a hung binary with `corveilInstallTimeout` (5s + SIGTERM), with an explicit comment that a hung corveil binary must not block. The Verify button runs the *same class* of user-configured binary via `proc.waitUntilExit()` with no timeout. If the binary hangs on `--version`, the detached `Task` and child process leak indefinitely and the button is stuck on "Verifying…" with no recovery. It's off the main actor so the UI doesn't freeze, but the inconsistency is a real "should fix." Consider reusing the same poll-with-deadline + `terminate()` pattern. (Yellow)
- 🟡 **Latent pipe-buffer deadlock in `runCorveilVersion`.** stdout/stderr are read *after* `waitUntilExit()`. For `--version` the output is tiny, but a misconfigured binary that writes >64KB before exiting fills the pipe buffer, blocks on write, and — with no timeout (above) — hangs forever. Draining concurrently (or via the timeout loop) closes both gaps at once. (Yellow)
- 🟢 In `Scaffolder.installCorveilSkill`, stdout is an undrained `Pipe()`; a chatty corveil binary could fill the buffer and block, but the 5s timeout bounds it to a warning. Functionally safe; consider `FileHandle.nullDevice` to make the discard explicit and deadlock-proof. (Green)
- 🟢 `onRescaffold` uses `try?`; if `scaffold` throws, `result` is `nil` and the `else if result != nil` branch is skipped, so a stale warning from a prior launch persists across a failed rescaffold. Minor. (Green)
- 🟢 First-time-setup double-run avoidance (`corveilBinaryPath: nil` in `completeSetup`, real path in `launchMainApp`) is correctly reasoned — verified `completeSetup` calls `launchMainApp()` right after. (Green)

### Tests
- CrowCore `binariesRoundTrip` / `binariesDefaultsEmptyWhenKeyMissing` pass locally; forward-compat decode is well covered.
- CrowUI `RunCorveilVersionTests` couldn't run locally (missing GhosttyKit xcframework binary), but the assumptions hold on macOS — verified `/bin/echo --version` → `--version` exit 0, `/usr/bin/true` exit 0, `/usr/bin/false` exit 1.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 3 Green] findings. The two Yellows are the same root cause: the Verify path lacks the timeout/pipe handling the startup path already has. Add a bounded timeout + concurrent drain to `runCorveilVersion` and this is an approve.

---

dhilgaertner (changes requested):
## Code & Security Review

Solid, well-documented feature: a generic `defaults.binaries` map, a Settings picker for the `corveil` slot, and a per-launch `corveil skill install` with non-fatal warning surfacing. Forward-compat decoding is tested, the timeout/SIGTERM handling is thoughtful, and the `<X>Warning` banner pattern matches existing conventions. `go`/`swift build` couldn't run here (the workspace's `GhosttyKit.xcframework` binary artifact is absent in this checkout — environment, not this PR), but `CrowCore` tests pass including the new `binariesRoundTrip` / `binariesDefaultsEmptyWhenKeyMissing` cases.

Two issues, both in subprocess pipe handling.

### Security Review
**Strengths:**
- No injection surface: `Process.arguments` is an argv array (no shell), and the binary path is user-chosen via NSOpenPanel / explicit text entry — not attacker-controlled.
- `isExecutableFile` gate before launch; failures are non-fatal and never block startup.
- No secrets logged; stderr is truncated to a single line for display.

**Concerns:** none security-blocking.

### Code Quality

- 🟡 **`Scaffolder.installCorveilSkill` leaves `stderr` as an undrained `Pipe()` — the exact deadlock the code's own comment claims to avoid.** `Sources/Crow/App/Scaffolder.swift`: stdout is routed to `FileHandle.nullDevice` with the comment *"routing to /dev/null is deadlock-proof — an undrained `Pipe()` would block the child if it ever wrote >64KB before exit."* But `stderr` is a plain `Pipe()` that is only read via `readToEnd()` **after** the process exits. If corveil writes more than the ~64KB pipe buffer to stderr before exiting, the child blocks on the full pipe, never exits, and you hit the 5s timeout + SIGTERM — and since `scaffold(...)` runs on the main thread during `applicationDidFinishLaunching`, that's a 5s stall before first paint. The Verify path (`runCorveilVersion`) already solves this correctly by draining both pipes concurrently via `DataAccumulator`; the install path should do the same (or accept that stderr can be lost). It's bounded by the timeout and a >64KB stderr from `skill install` is unlikely, hence yellow not red — but it directly contradicts the stated invariant and is the inconsistency most worth closing.

- 🟡 **`runCorveilVersion` snapshots the accumulators with no EOF barrier — a tail chunk delivered exactly at process exit can be dropped.** `Packages/CrowUI/Sources/CrowUI/SettingsView.swift`: once the poll loop observes `proc.isRunning == false`, it immediately calls `outAcc.snapshot()` / `errAcc.snapshot()`, then `defer` nils the readability handlers. The handlers run on a Foundation-managed background queue with no happens-before relationship to the polling thread, so the final readable event delivered at EOF may not have been appended yet when you snapshot (and nilling the handler then drops it). In practice the 50ms poll slack makes this near-deterministic for tiny `--version` output (why the exact-equality tests pass), but it's technically racy and could intermittently truncate or blank the result. Draining any remaining data synchronously after exit (e.g. `readDataToEndOfFile()` on each handle before nilling handlers) would make it deterministic.

### Notes (non-blocking)
- 🟢 The 5s main-thread block in `installCorveilSkill` on a misconfigured/hung binary delays first paint. It's documented and bounded, and the Verify path runs off-actor via `Task.detached`; consider mirroring that here if startup latency ever matters.
- 🟢 Persistence on the corveil `TextField` only fires via `.onSubmit`/Browse (not per keystroke) — consistent with the existing devRoot field, just noting typing + closing Settings without Enter won't persist.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 2 Green] findings. Both yellows are pipe-drain consistency fixes; addressing the stderr drain in `installCorveilSkill` to match the Verify path closes the strongest one.

---

dhilgaertner (changes requested):
## Code & Security Review

Nicely structured change — generic `defaults.binaries` map with forward-compatible decoding, fixed-array `Process` args (no shell, so no command injection), and a non-fatal warning surfaced through the existing `<X>Warning` pattern. The pipe-drain/timeout work is thoughtful and well-commented. One concurrency bug in the shared `PipeDrainer` keeps this from approval.

### Critical Issues
None blocking, but see the Yellow below — it can crash app startup.

### Code Quality

**🟡 `PipeDrainer.abandon()` can raise an uncaught exception → startup crash** — `Scaffolder.swift:705` / `SettingsView.swift:368`

`abandon()` closes `pipe.fileHandleForReading`, the *same* handle the background drain thread is reading via `readDataToEndOfFile()`. On the `proc.run()`-throws path this is reachable and largely deterministic:

- `installCorveilSkill` gates on `fm.isExecutableFile(atPath:)`, but that only checks the exec bit — a file with `+x` that isn't a launchable image (bad shebang, non–Mach-O, ENOEXEC) still passes the gate and makes `proc.run()` throw (`Scaffolder.swift:606-612`, `SettingsView.swift:262-268`).
- In the catch, `abandon()` runs synchronously and immediately — typically *before* the GCD global-queue closure has even been scheduled. The closure then calls `readDataToEndOfFile()` on an already-closed `FileHandle`, which raises `NSFileHandleOperationException` ("Bad file descriptor"). That's an Obj-C exception Swift can't catch → crash.
- Even when the read is already in flight, yanking its fd out from under the blocked `read()` is undefined and can raise the same way.

`Scaffolder` runs inside `applicationDidFinishLaunching`, so a mis-pointed corveil path turns into a crash on launch rather than the intended orange banner.

Fix: have `abandon()` close the **write** end instead of the read end:

```swift
func abandon() {
    try? pipe.fileHandleForWriting.close()
}
```

Closing the write end delivers a clean EOF, so `readDataToEndOfFile()` returns normally and signals the semaphore. In the launch-failure case the parent still owns the only write end (the child never spawned), so this reliably unblocks the reader; in the timeout case `Process` has already closed the parent's write copy, so it's a harmless no-op. Applies to both copies of the helper.

**🟢 Main thread blocks up to ~5.5s on a hung corveil** — `Scaffolder.swift:618-633`

The poll-loop + 0.5s SIGTERM grace runs on the main thread before first paint (acknowledged in the comments). 5s is generous, but a hung binary still delays window draw by the full budget. Consider running the install off-main and updating `corveilSkillInstallWarning` when it lands, so a misbehaving corveil never delays launch at all. Defensible as-is given the timeout bound.

**🟢 Trim inconsistency** — `Scaffolder.swift:568` trims `.whitespaces`; the UI binding (`SettingsView.swift:204`) trims `.whitespacesAndNewlines`. Harmless, but aligning them avoids a path that round-trips differently than it's validated.

**🟢 Duplicated `PipeDrainer`/`DataAccumulator`** across `Scaffolder` (Crow target) and `SettingsView` (CrowUI). Already acknowledged in the comments; fine for now, but if a third caller appears a small CrowCore utility would pay off — and would mean the fix above only lands once.

### Security Review
**Strengths:**
- `Process.arguments` is a fixed array — no shell interpolation, no command injection via path or devRoot.
- `isExecutableFile` gate before exec; failures are non-fatal and never block startup.
- Forward-compatible config decode (`decodeIfPresent ... ?? [:]`) — pre-CROW-482 configs still load.

**Concerns:**
- 🟢 The feature executes a user-configured binary on every launch. That's the intended design and the threat model is local (an attacker who can write `config.json` already has code-exec footing), so no change needed — just noting the launch-time exec surface.

### Tests
- CrowCore `binariesRoundTrip` / `binariesDefaultsEmptyWhenKeyMissing` pass.
- The CrowUI/Crow targets don't compile in this review checkout (missing `Frameworks/GhosttyKit.xcframework` binary artifact — unrelated to this PR), so the `runCorveilVersion` suite and `swift build` weren't executed here. The PR notes a clean `swift build --arch arm64` locally.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 4 Green] findings. The Yellow `abandon()` fix is small and worth landing in this round; the Greens are optional.

---

dgershman:
@dhilgaertner — round-4 review landed against `78fd54f` (the bg-drain commit). Since then I pushed two follow-ups (`c2bb601`, `1a62527`) because CI kept failing `zeroExitWithStdoutShowsSnippet`: `proc.waitUntilExit()` is the only thing that reliably triggers Foundation's pipe-write-FD cleanup, and calling it *after* a polling loop is too late — Foundation needs to own the wait.

So the current code (HEAD `301bb5a`) drops `PipeDrainer`/`DataAccumulator` entirely in favor of `TimeoutWatchdog` (DispatchSourceTimer SIGTERMing the child on expiry) + `proc.waitUntilExit()` + synchronous `readToEnd()` after exit. This means:

- 🟡 **`PipeDrainer.abandon()` close-the-read-end bug** — no longer applies, the helper is gone. The launch-failure (`proc.run()` throws) path now just returns the warning with no fd manipulation. Verified locally that all 5 `RunCorveilVersion` tests pass under `arch -arm64 swift test --filter "RunCorveilVersion"`.
- 🟢 **5.5s main-thread block on hung corveil** — same trade-off, same bound. Off-main install is a reasonable future change but not in this round.
- 🟢 **Trim inconsistency** — fixed in `301bb5a`; both sides now use `.whitespacesAndNewlines`.
- 🟢 **Duplicated helper** — only one helper (`TimeoutWatchdog`) is duplicated now, ~30 lines × 2; still under the "extract when there's a third caller" threshold.

Trade-off vs the round-3 bg-drain request: a corveil binary writing >64KB to stderr before exit would now block on a full pipe buffer until the watchdog SIGTERMs it, losing the tail. Documented in-comment. Real corveil emits <1KB and the bg-drain approach turned out not to deliver EOF reliably in CI anyway, so the simpler pattern won.

Ready for re-review against `301bb5a`.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- **No shell injection.** Both subprocess call sites (`Scaffolder.installCorveilSkill`, `SettingsView.runCorveilVersion`) use `Process` with `executableURL` + an `arguments` array — never a shell string. The user-supplied path and `--path` target are passed as discrete argv entries, so paths with spaces/metacharacters can't inject commands.
- **Defensive pre-exec gate.** `FileManager.isExecutableFile(atPath:)` is checked before `proc.run()` in both paths, so a missing/non-executable path degrades to a warning rather than a launch error.
- **Inherent trust boundary is appropriate.** Executing a user-picked binary is the explicit purpose of the feature (a local dev tool the user built), not an escalation — no untrusted input crosses a boundary.
- **Warning surfacing is safe.** corveil's stderr is shown in a `textSelection`-enabled banner; it's the user's own machine output, no secret exposure, and it's truncated to a single line.
- **Forward-compatible config decode.** `binaries` uses `decodeIfPresent ?? [:]`, and `AppConfigTests.binariesDefaultsEmptyWhenKeyMissing` locks in that pre-CROW-482 configs still decode. Verified passing locally (262 CrowCore tests green).

### Code Quality
- Clean two-way `corveilBinding` that treats empty as "unset" (no stale empty map entries) and trims on commit — and `301bb5a` aligns the Scaffolder trim charset (`.whitespacesAndNewlines`) with it.
- First-launch double-install is correctly avoided: `completeSetup` passes `corveilBinaryPath: nil`, and `launchMainApp` (always invoked right after) re-scaffolds with the real path. Verified the control flow at AppDelegate.swift:306 → :318 → :402.
- Warning lifecycle is handled well — `onRescaffold` always assigns `result.warning` (clearing a stale banner on later success) and replaces it with a fresh message in the `catch`.
- Verify runs off the main actor via `Task.detached` with mutations marshaled back through `MainActor.run`; the button is disabled while in-flight. UI stays responsive.

- **Watchdog uses SIGTERM only.** `TimeoutWatchdog`/`ScaffolderTimeoutWatchdog` call `proc.terminate()` (SIGTERM) with no SIGKILL escalation. Since `Scaffolder.scaffold` runs synchronously on the main thread during `applicationDidFinishLaunching`, a pathological binary that ignores SIGTERM would block first paint indefinitely. Real-world risk is very low (a `skill install` command terminating on default SIGTERM disposition), but a SIGKILL fallback after a short grace window would make the startup path provably non-hanging.
- **Main-thread cost on every launch.** Even in the happy path, configured users pay subprocess spawn+exec latency before the window draws (bounded at `corveilInstallTimeout` = 5s). The tradeoff (skill synced before the Manager session starts) is reasonable and thoroughly documented; flagging only so it's a conscious choice.
- **Narrow false-timeout window.** If the child exits at almost exactly the deadline, the watchdog handler could set `didFire` between `waitUntilExit()` returning and `cancel()` reading it, mislabeling a success as a timeout. Microsecond window, cosmetic-only impact.
- **Watchdog duplicated across two targets.** Author documented the rationale (Scaffolder and SettingsView are separate targets with no shared private util); two small copies is a defensible call.

### Verification notes
- `swift test` in `Packages/CrowCore` passes (262 tests), including both new `binaries` round-trip/forward-compat tests.
- The main `Crow`/`CrowUI` target could not be built here: the prebuilt `GhosttyKit.xcframework` contains no binary artifact in this checkout (environment limitation, unrelated to this PR). The `SettingsView.swift`/`Scaffolder.swift` changes and the `runCorveilVersion` UI tests were therefore reviewed by inspection; the PR author reports `swift build --arch arm64` clean.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 4 Green] findings. Well-engineered, well-documented, tests pass; the Green items are optional hardening/awareness notes the author can weigh.

---

PR #488, opened 2026-06-11
Cancel tmux copy-mode before paste-buffer in TmuxBackend.sendText

### LINKED ISSUE #486: Scroll trap + Crow can't paste into Manager: tmux copy-mode after wheel breaks user typing and programmatic send
**Two reported symptoms, one shared root cause:**

1. **"After scrolling, I can't type until I hit Escape."** Mouse wheel triggers tmux's default `WheelUpPane` binding, which puts the pane into **copy-mode**. In copy-mode, keystrokes are consumed by vi-style navigation (h/j/k/l/q/Escape) instead of going to the agent. The user has to press `q` or `Escape` to exit, then type.
2. **"Crow can't paste into the Manager session."** When Crow shells `tmux paste-buffer` via `TmuxBackend.sendText`, it doesn't check if the pane is in copy-mode first. In copy-mode, paste-buffer doesn't deliver text to the underlying shell — it gets eaten by copy-mode bindings (or silently fails depending on tmux version).

Both share the same root cause: **tmux is in copy-mode and Crow doesn't notice or compensate.**

Why this is now the dominant friction: #452 reverted the `mouse off` attempt and restored `mouse on` with the `copy-pipe-no-clear` binding so wheel scrollback works through tmux. That fixed scrollback, but the price is that wheel always enters copy-mode (the `crow-tmux.conf` comment at line 46 explicitly notes the WheelUpPane default is left in place because wheel scrollback relies on it). So the cost of having scrollback is sitting in copy-mode until the user explicitly leaves.

### 1. Cancel copy-mode in `TmuxBackend.sendText` before paste-buffer (REQUIRED — fixes the Manager paste bug)

**File:** `Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:322-353`

Current send path:
- `loadBufferFromStdin` → load text into a tmux buffer
- `ctrl.pasteBuffer(name:, target:)` → paste (line 337)
- `ctrl.sendKeys(target:, keys: ["Enter"])` → submit (line 347)

Add a guard before paste:

```swift
// Cancel copy-mode if the pane is currently in it. send-keys -X cancel
// errors when not in a mode, so guard with if-shell to keep this a no-op
// in the common case.
try ctrl.runCommand("if-shell -F -t \(target) '#{pane_in_mode}' 'send-keys -t \(target) -X cancel'")
try ctrl.pasteBuffer(name: bufferName, target: target)
try ctrl.sendKeys(target: target, keys: ["Enter"])
```

If `TmuxController` already exposes an `if-shell` wrapper, reuse it. If not, add `cancelCopyModeIfActive(target:)` as a named method so the intent is local and tests can mock it.

Guarantees programmatic sends (`crow send`, Manager paste, every auto-respond prompt, `dispatchManual` quick-action) work regardless of whether the user happened to scroll before the send fired.

### 2. Crow-side input-time copy-mode auto-exit (UX fix for keystroke trap)

**File:** `Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift` — the `keyDown(with:)` path (search for `ghostty_surface_key` to find the key-routing code).

Principle: when the user types a non-navigation keystroke into a tmux pane that's in copy-mode, they almost certainly mean to type, not navigate. Auto-cancel copy-mode and forward the keystroke.

Implementation:

- **Track copy-mode state.** Subscribe to tmux control-mode `%mode-changed` notifications (Crow already speaks tmux control mode for the cockpit — extend the existing parser). Cache a `Set<TerminalID>` of currently-in-copy-mode panes. No per-keystroke `tmux display` polling.
- **On keyDown:** if the focused terminal's ID is in the in-copy-mode set, and the key is a printable character (or Tab — not q/Escape/h/j/k/l/arrows which are legitimate copy-mode keys), shell `tmux send-keys -t <target> -X cancel` first, then let the normal key path send the original keystroke.
- **Keys that should NOT auto-cancel** (legitimate copy-mode navigation):
  - `q`, `Escape` — explicit exit
  - `h`, `j`, `k`, `l`, arrows, PageUp, PageDown — navigation
  - `Space` — start selection (copy-mode-vi)
  - `v` — enter visual selection
  - `y` — yank
  - `/`, `?` — search
  - `Enter` is **ambiguous** — used in copy-mode (confirm selection) but also a common typing key. Recommend: keep `Enter` as a copy-mode key (preserve selection-copy UX), so users wanting to submit will exit via Escape first. Revisit if friction.
- **Net UX:** scroll back, read what you wanted, then just start typing. Copy-mode silently exits, your typing reaches the agent.

This is more invasive than (1). Can land in a follow-up PR if scope is a concern — (1) alone unblocks the Manager-paste symptom the user explicitly named.

### 3. Verify wheel uses `copy-mode -e` (cleanup, optional)

Newer tmux (≥3.0) defaults `WheelUpPane` to `copy-mode -e` — the `-e` makes copy-mode auto-exit when wheel-down scrolls back past the bottom of history. If Crow relies on the tmux default, confirm it really is `-e` in the version Crow requires. If not, add an explicit rebind in `crow-tmux.conf`:

```
bind -T root WheelUpPane if-shell -F -t = "#{mouse_any_flag}" "send-keys -M" { if-shell -F -t = "#{pane_in_mode}" "send-keys -M" "copy-mode -et=" }
```

Makes the auto-exit-at-bottom contract explicit in Crow's config so it survives future tmux default changes.

## Acceptance

1. **Manager paste**: in a Crow Manager session, scroll the terminal up a few lines (entering copy-mode), then trigger any flow that does `crow send` (Refine action, auto-respond, manual `crow send` from another shell). The text lands in the Manager's prompt and submits — does **not** get eaten by copy-mode bindings.
2. **User typing**: scroll up to read, then start typing. Each character appears in the agent's prompt — copy-mode silently exited on the first non-navigation keystroke.
3. **Copy-mode navigation still works**: while in copy-mode, `j`/`k`/arrows/PageUp/PageDown still navigate; `q`/`Escape` still exit; `/` still searches; selection (`Space` to start, `Enter` to copy) still works.
4. **Tests**:
   - Unit test for `TmuxBackend.sendText` mocks a `TmuxController` and verifies the if-shell-cancel command precedes paste-buffer.
   - Tests for the key-routing layer mock the in-copy-mode set and verify that printable keystrokes trigger a cancel + send, while nav keystrokes don't.
5. **Regression check**: drag-select / copy-pipe-no-clear / wheel scrollback all still work as defined by #452.

## Critical files

| Purpose | Path |
|---|---|
| Send-path guard (must fix) | `Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:322-353` |
| TmuxController — add `cancelCopyModeIfActive(target:)` | `Packages/CrowTerminal/Sources/CrowTerminal/TmuxController.swift` |
| Key routing for auto-exit-on-keystroke | `Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift` (`keyDown`) |
| Control-mode parser (subscribe to `%mode-changed`) | tmux control-mode handler in `TmuxBackend.swift` or sibling |
| Wheel binding (optional explicit `-e`) | `Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf` (lines 45-47) |

## Related

- #445 / #446 / #452 — tmux-mouse selection thread; established the `mouse on` + `copy-pipe-no-clear` baseline this ticket sits on top of.
- #451 — config reconcile-on-attach (so any new `crow-tmux.conf` bindings here land automatically for users without a tmux restart).
- #473 — `unbind-key -a` exit-1 cosmetic bug; not blocking but worth landing first so `source-file` return codes are honest.
- `AutoRespondCoordinator.dispatchManual` (`Sources/Crow/App/AutoRespondCoordinator.swift:82`) — uses the same send pipeline; fix (1) also unblocks manual nudges into a scrolled-up Manager.

dgershman:
Closes #486

## Summary

Crow's bundled `crow-tmux.conf` keeps `mouse on` so wheel scrollback works through the cockpit (#452). The side effect is that tmux's default `WheelUpPane` puts the pane into **copy-mode**, where `paste-buffer` does not deliver content into the underlying shell. Programmatic sends — Manager paste, auto-respond prompts, quick actions — silently fail whenever the user happened to scroll first.

This PR fixes the send path: pre-cancel copy-mode before `paste-buffer` via `tmux if-shell -F '#{pane_in_mode}' 'send-keys -X cancel'`. The `if-shell` guard keeps it a no-op in the common case (pane not in a mode).

## Changes

- **`TmuxController.cancelCopyModeIfActive(target:)`** — new helper wrapping the `if-shell` guard so the intent reads locally.
- **`TmuxBackend.sendText`** — calls the helper after `loadBufferFromStdin` and before `pasteBuffer`. Doc comment extended to document why.
- **`crow-tmux.conf`** — explicit `WheelUpPane` rebind using `copy-mode -e`. Matches tmux ≥3.0 defaults but pinning it insulates Crow from upstream changes and minimizes time-in-copy-mode (the `-e` auto-exits when wheel-down scrolls past the bottom of history).
- **Tests** — two new real-tmux integration tests in `TmuxBackendTests`:
  - `sendTextCancelsCopyModeBeforePaste` forces the pane into copy-mode via a side-channel controller, then asserts `pane_in_mode` flips back to `0` after `sendText`.
  - `sendTextOnNormalPaneIsUnaffectedByCancelGuard` sanity-checks the `if-shell` no-op path so a normal-state pane doesn't fail.

## Deferred (intentional)

The keystroke-trap leg of #486 — auto-exit copy-mode on a printable keyDown in `GhosttySurfaceView` — is not addressed here. It needs a tmux control-mode `%mode-changed` parser the codebase doesn't have yet; building that infrastructure is orthogonal to the paste fix this PR closes. I'll file a follow-up issue against this PR.

The user-typing trap is still present in this PR (scroll up, then start typing → keystrokes eaten until Escape). The Manager-paste / auto-respond / quick-action paste bug — the symptom the ticket title actually names — is fixed.

## Test plan

- [x] `make ghostty` then `arch -arm64 swift test --package-path Packages/CrowTerminal` → all 63 tests in 9 suites pass.
- [ ] Manual: launch Crow Manager, mouse-wheel-up to enter copy-mode visually, run `crow send --session 00000000-0000-0000-0000-000000000000 --terminal <uuid> "test\n"` from another shell — text lands in the prompt and submits.
- [ ] Manual: trigger an auto-respond or Refine quick action into a scrolled-up working session — prompt arrives.
- [ ] Regression: wheel scrollback, drag-select + copy-pipe-no-clear, double/triple-click word/line copy still work as defined by #445 / #452.

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf
@@ -76,6 +76,14 @@ bind -T root MouseDown1Pane select-pane -t = \; if -F -t = "#{mouse_any_flag}" "
 bind -T root DoubleClick1Pane select-pane -t = \; if -F -t = "#{mouse_any_flag}" "send-keys -M" { if -F -t = "#{pane_in_mode}" "send-keys -M" { copy-mode -H ; send-keys -X select-word ; run -d0.3 ; send-keys -X copy-pipe-no-clear "pbcopy" } }
 bind -T root TripleClick1Pane select-pane -t = \; if -F -t = "#{mouse_any_flag}" "send-keys -M" { if -F -t = "#{pane_in_mode}" "send-keys -M" { copy-mode -H ; send-keys -X select-line ; run -d0.3 ; send-keys -X copy-pipe-no-clear "pbcopy" } }
 
+# Pin WheelUpPane to `copy-mode -e` so the pane auto-exits copy-mode once
+# wheel-down scrolls back past the bottom of history. Matches tmux ≥3.0
+# defaults, but making it explicit insulates Crow from upstream changes
+# and pairs with the send-path copy-mode cancel in TmuxBackend.sendText
+# (#486) — together they minimize the time a pane sits stuck in copy-mode
+# after the user scrolls.
+bind -T root WheelUpPane if-shell -F -t = "#{mouse_any_flag}" "send-keys -M" { if-shell -F -t = "#{pane_in_mode}" "send-keys -M" "copy-mode -et=" }
+
 # Default history-limit is 2000 lines per pane. Phase 3 §2 (the spike)
 # measured 5 windows × 10k lines as a 400kB RSS delta, so we have plenty
 # of headroom. 5000 matches typical session-detail expectations.
--- a/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf
@@ -82,7 +82,7 @@ bind -T root TripleClick1Pane select-pane -t = \; if -F -t = "#{mouse_any_flag}"
 # and pairs with the send-path copy-mode cancel in TmuxBackend.sendText
 # (#486) — together they minimize the time a pane sits stuck in copy-mode
 # after the user scrolls.
-bind -T root WheelUpPane if-shell -F -t = "#{mouse_any_flag}" "send-keys -M" { if-shell -F -t = "#{pane_in_mode}" "send-keys -M" "copy-mode -et=" }
+bind -T root WheelUpPane if -F -t = "#{mouse_any_flag}" "send-keys -M" { if -F -t = "#{pane_in_mode}" "send-keys -M" "copy-mode -et=" }
 
 # Default history-limit is 2000 lines per pane. Phase 3 §2 (the spike)
 # measured 5 windows × 10k lines as a 400kB RSS delta, so we have plenty
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -319,6 +319,13 @@ public final class TmuxBackend {
     /// auto-respond prompts (which fire when the terminal has been idle) can
     /// race: the Enter arrives before the TUI finishes handling the paste,
     /// causing it to be silently dropped (#272).
+    ///
+    /// We also pre-cancel copy-mode on the pane before pasting (#486). The
+    /// bundled `crow-tmux.conf` keeps `mouse on` so wheel scrollback works
+    /// (#452), but the default `WheelUpPane` puts the pane into copy-mode,
+    /// where `paste-buffer` does not deliver content. Without the cancel,
+    /// every programmatic send into a pane the user has scrolled (Manager
+    /// paste, auto-respond, quick actions) is silently dropped.
     public func sendText(id: UUID, text: String) throws {
         guard let windowIndex = bindings[id] else {
             throw TmuxBackendError.unknownTerminal(id)
@@ -334,6 +341,10 @@ public final class TmuxBackend {
                 let bufferName = "crow-\(id.uuidString)"
                 try ctrl.loadBufferFromStdin(name: bufferName, data: Data(payload.utf8))
                 defer { ctrl.deleteBuffer(name: bufferName) }
+                // Cancel copy-mode if the user scrolled the pane into it
+                // — paste-buffer is a no-op while the pane is in a mode
+                // (#486).
+                try ctrl.cancelCopyModeIfActive(target: target)
                 try ctrl.pasteBuffer(name: bufferName, target: target)
                 didPaste = true
             }
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -320,12 +320,14 @@ public final class TmuxBackend {
     /// race: the Enter arrives before the TUI finishes handling the paste,
     /// causing it to be silently dropped (#272).
     ///
-    /// We also pre-cancel copy-mode on the pane before pasting (#486). The
-    /// bundled `crow-tmux.conf` keeps `mouse on` so wheel scrollback works
-    /// (#452), but the default `WheelUpPane` puts the pane into copy-mode,
-    /// where `paste-buffer` does not deliver content. Without the cancel,
-    /// every programmatic send into a pane the user has scrolled (Manager
-    /// paste, auto-respond, quick actions) is silently dropped.
+    /// We also pre-cancel copy-mode on the pane before any delivery (#486).
+    /// The bundled `crow-tmux.conf` keeps `mouse on` so wheel scrollback
+    /// works (#452), but the default `WheelUpPane` puts the pane into
+    /// copy-mode, where both `paste-buffer` and `send-keys Enter` are
+    /// silently consumed by copy-mode key bindings instead of reaching the
+    /// underlying shell. Without the cancel, every programmatic send into
+    /// a pane the user has scrolled (Manager paste, auto-respond, quick
+    /// actions, bare-Enter submits) is dropped.
     public func sendText(id: UUID, text: String) throws {
         guard let windowIndex = bindings[id] else {
             throw TmuxBackendError.unknownTerminal(id)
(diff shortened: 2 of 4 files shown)


dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No injection surface in the new `if-shell` command. `target` is built only from `TmuxBackend.cockpitSessionName` (a constant) and an integer `windowIndex`; even though it's string-interpolated into the `send-keys -t <target> -X cancel` argument, there is no attacker-controlled input.
- The `if-shell -F '#{pane_in_mode}'` guard is semantically correct: with `-F`, tmux treats `"0"` (the not-in-mode value) as false, so the `send-keys -X cancel` only fires when the pane is actually in a mode — avoiding the `send-keys -X` "not in a mode" error in the common path.
- `crow-tmux.conf` `WheelUpPane` rebind faithfully mirrors the tmux ≥3.0 default and the existing `DoubleClick1Pane`/`TripleClick1Pane` brace-block pattern already in the file (`copy-mode -et=` == `copy-mode -e -t =`, consistent with the file's `-t =` convention).

**Concerns:**
- None.

### Code Quality

**Yellow — copy-mode cancel is skipped on the bare-Enter (Enter-only) send path** (`TmuxBackend.swift:340-358`)

`cancelCopyModeIfActive(target:)` is called *inside* the `if !payload.isEmpty { ... }` block. When `sendText` is invoked with `text == \"\n\"` (payload empties to \"\", `didPaste` stays false), the cancel is skipped and the Enter is delivered via `ctrl.sendKeys(target:, keys: [\"Enter\"])`. A plain `send-keys Enter` into a pane that's in copy-mode is routed through the copy-mode key table (default emacs: copy-selection-and-cancel), so it exits copy-mode but never delivers a carriage return to the underlying shell — the same silent-drop failure mode this PR fixes for paste.

This contradicts the new doc comment's claim that *"every programmatic send into a pane the user has scrolled ... is silently dropped"* is fixed — the Enter-only submit path still is. `crow send --terminal <id> \"\n\"` (and `TmuxBackendTests.swift:225` exercises exactly this bare-`\n` shape) hits it. Auto-respond / quick actions append text+`\n` (non-empty payload), so they're covered; pure-Enter submits are not.

Recommendation: hoist `cancelCopyModeIfActive(target:)` above the paste/Enter split so it runs for any delivery into the pane (it's one cheap `if-shell` call), or duplicate it before the `sendKeys([\"Enter\"])`. Then the doc comment's contract holds for all send shapes.

**Green — cosmetic:** the new `WheelUpPane` line spells out `if-shell` while the neighboring `MouseDown1Pane`/`DoubleClick1Pane` bindings use the `if` alias. Harmless; consider matching the surrounding style for grep-consistency.

### Tests
- `sendTextCancelsCopyModeBeforePaste` and `sendTextOnNormalPaneIsUnaffectedByCancelGuard` are well-scoped real-tmux integration tests covering the in-mode and no-op-guard paths. A test for the bare-Enter-into-copy-mode case would have caught the gap above.
- I did not re-run the suite: a local build needs the GhosttyKit xcframework (`make ghostty`) and the integration tests require a real tmux. Relying on the PR's stated "all 63 tests pass" for the green run.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 1 Green] findings. The core paste fix is correct and well-tested; addressing the Enter-only path (or narrowing the doc comment's claim) closes it out in this round trip.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No command-injection surface. `TmuxController.cancelCopyModeIfActive` interpolates only `target`, which is derived from the constant `cockpitSessionName` (`"crow-cockpit"`) and an integer `windowIndex` — never from user input. The user's `text` reaches tmux via `loadBufferFromStdin` (stdin) and `paste-buffer`, not via argv, so the payload can't influence command parsing.
- The `if-shell -F '#{pane_in_mode}'` guard means `send-keys -X cancel` only fires when the pane is actually in a mode, avoiding the error tmux raises otherwise.

### Code Quality
- **Correct hoisting of the cancel (`TmuxBackend.swift:347`).** Placing `cancelCopyModeIfActive` above the `if !payload.isEmpty` / `endsWithNewline` split is the key correctness point: it covers the bare-`\n` path where `send-keys Enter` would otherwise be eaten by the copy-mode key table (emacs `copy-selection-and-cancel` / vi `cancel`) and never reach the shell. The new `sendTextBareEnterCancelsCopyMode` test locks this in.
- **Ordering is safe.** Each `ctrl.run(...)` is a separate synchronous tmux client invocation, so the server has processed the cancel before `paste-buffer` is issued — no race between cancel and paste.
- **`crow-tmux.conf` WheelUpPane rebind** mirrors the existing MouseDown/DoubleClick/TripleClick binding structure and matches the tmux ≥3.0 default; pinning it is reasonable insurance. Note it only takes effect for tmux servers started after this change (config is read at server start / reload), which is expected behavior, not a regression.
- Doc comments are thorough and accurate about the failure mode and the fix.

**Green (consider only):**
- `sendTextCancelsCopyModeBeforePaste` asserts `pane_in_mode` flips back to `0`, which proves the cancel ran, but doesn't assert the pasted token actually landed in the pane (e.g. via a `capture-pane` check). The cancel behavior is the regression being fixed, so this is fine; an end-to-end content assertion would just make the test name's "...BeforePaste" promise airtight. Covered by the manual test plan.
- Minor: the inline comment block at `TmuxBackend.sendText` (lines ~341-346) has an unbalanced `(where ...` parenthesis. Cosmetic.
- The keystroke-trap leg of #486 (auto-exit copy-mode on printable keyDown) is intentionally deferred with a clear rationale and a planned follow-up. The paste/auto-respond/quick-action symptom named in the ticket is fixed here.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 3 Green] findings. Tight, well-tested, well-documented fix with no security or correctness blockers.

Note: I could not run the build/tests locally because `GhosttyKit.xcframework` is not present in this checkout (requires `make ghostty`); the review is by inspection. The PR reports all 63 tests passing.

---

PR #496, opened 2026-06-11
Fix Codex hook config: rename to [features].hooks, drop async

### LINKED ISSUE #494: Codex hook config: writes deprecated [features].codex_hooks + async hooks Codex silently skips (breaks session-state detection)
**Symptom — every Codex session launch emits these warnings from Codex's startup:**

```
⚠ `[features].codex_hooks` is deprecated. Use `[features].hooks` instead.
Enable it with `--enable hooks` or `[features].hooks` in config.toml.
…
⚠ skipping async hook in /Users/danny/.codex/hooks.json: async hooks are not supported yet
⚠ skipping async hook in /Users/danny/.codex/hooks.json: async hooks are not supported yet
```

Both warnings point at Crow's Codex hook config writer. Two distinct problems landing in two distinct files.

## Bug A: `[features].codex_hooks` is deprecated

Crow writes (or the user manually wrote, from a Crow-pre-deprecation install) `[features].codex_hooks = true` to `~/.codex/config.toml`. Codex v0.139.0+ renamed this key to `[features].hooks`. The old key is honored with a deprecation warning today but will eventually stop working.

**Fix:** find the writer that touches `~/.codex/config.toml` (likely `CodexHookConfigWriter` in `Packages/CrowCodex/Sources/CrowCodex/`). Update the key it writes to `hooks`. On read, detect a legacy `codex_hooks = …` value, rename it to `hooks = …`, and remove the legacy entry — a one-time migration per user. Idempotent.

## Bug B: Async hooks are silently skipped — Crow's session-state detection is silently broken

The `~/.codex/hooks.json` Crow writes declares async hooks (presumably the same shape Claude Code uses to broadcast agent state for auto-respond, card-color updates, completion detection). Codex's hook runtime doesn't support async hooks and silently skips them on launch. Whatever Crow expects to receive from those hooks **never arrives**.

This is the more dangerous of the two bugs. The deprecation warning is loud; the async-skip is silent. It implies:

- **Auto-respond into Codex sessions may not fire** when the agent reaches a "waiting on user" state, because the hook that would broadcast that state to Crow isn't running.
- **Card-color updates** (the bug pattern from #428 for Cursor) likely have the same shape for Codex: the completion signal Crow watches for never gets emitted.
- **`AutoRespondCoordinator` skipping checks** that rely on terminal state may evaluate stale state for Codex sessions.

**Fix shape:**
1. Identify which hooks Crow writes for Codex and which of them are async. Find the writer in `Packages/CrowCodex/Sources/CrowCodex/CodexHookConfigWriter.swift` (or wherever the conformer to `HookConfigWriter` for Codex lives).
2. **Make them sync.** Codex's hook runtime is sync-only as of v0.139.0; declare the hooks with `async: false` or omit the flag entirely (depending on the schema). If a hook's intent genuinely requires async behavior (e.g., long-running notification), evaluate whether that hook should exist on Codex at all — surface it as out-of-scope and disable rather than ship broken.
3. **Verify the broadcast works end-to-end.** Open a Codex session, watch for the state-change events Crow expects (probably via the same channel `ClaudeCodeAgent`'s hooks use — `CodexSignalSource` referenced at `OpenAICodexAgent.swift:17`). The flow should mirror Claude's: agent state changes → hook fires sync → Crow's `StateSignalSource` receives the signal → `AppState` reflects it → auto-respond and card-color logic see the new state.

## Acceptance

- A fresh Codex session launch produces **no** `[features].codex_hooks is deprecated` warning.
- A fresh Codex session launch produces **no** `⚠ skipping async hook …` warnings.
- After a Codex agent reaches a waiting/blocked state, Crow's UI reflects it within the same poll/event cycle Claude Code sessions use — the card color updates, auto-respond gates trigger normally.
- Legacy users with `[features].codex_hooks = true` in their `config.toml` get migrated to `[features].hooks = true` on first launch with the fix; a log line records the migration. Idempotent.
- Test against the actual Codex binary at v0.139.0+.

## Critical files

| Purpose | Path |
|---|---|
| Hook config writer | `Packages/CrowCodex/Sources/CrowCodex/CodexHookConfigWriter.swift` (or the file conforming to `HookConfigWriter` for Codex) |
| Codex state signal source (the consumer of those hooks) | `Packages/CrowCodex/Sources/CrowCodex/CodexSignalSource.swift` (referenced at `OpenAICodexAgent.swift:17`) |
| Agent declaration (where the hook writer is wired in) | `Packages/CrowCodex/Sources/CrowCodex/OpenAICodexAgent.swift:16-17` |
| HookConfigWriter protocol (contract) | search `protocol HookConfigWriter` in `Packages/CrowCore/Sources/CrowCore/` |
| User's affected files (read/migrate) | `~/.codex/config.toml`, `~/.codex/hooks.json` |

## Out of scope

- Redesigning Crow's hook architecture to require async semantics that Codex doesn't support. This ticket gets Crow back to a working state with Codex's current sync-only hook model. If async is genuinely needed long-term, that's a Codex-CLI roadmap conversation, not a Crow ticket.
- Bare-launch fix (the prompt-isn't-delivered problem) — that's its own ticket since it lives in `autoLaunchCommand`, not the hook writer.

## Related

- **The bare-launch ticket** (filed as the sibling to this one — Codex launches with no prompt). They're in the same `OpenAICodexAgent.swift` neighborhood but independent. The bare-launch fix doesn't help if the agent runs but Crow can't observe its state.
- #428 — Cursor sessions don't update card color on completion. Identical category of bug for a different agent — the completion signal is silently broken. After this ticket lands for Codex, confirm whether Cursor has the same hook-writer issue and file a parallel ticket if so.
- #424 — Cursor agent launches bare. Same as the bare-launch sibling for Codex but for Cursor — kept separate to avoid bundling.
- The agent-kind plumbing cluster (#421, #427, #440, #443) — orthogonal architectural work; this ticket is the immediate concrete Codex hook bug.

dgershman:
## Summary
- Switch the `~/.codex/config.toml` writer from the deprecated `[features].codex_hooks` key to `[features].hooks` (Codex v0.139.0+).
- One-shot migration: strip a legacy `codex_hooks = …` line under `[features]` so users converging from older Crow installs end up with a single, current entry. Idempotent.
- Empty out `asyncEvents` in `CodexHookConfigWriter` so no entry is written with `async: true` — Codex's hook runtime is sync-only and silently skipped those, breaking Crow's session-state detection (auto-respond, card-color, etc.).

## Why this matters
The deprecation warning is loud; the async-skip was silent. With async hooks skipped, `PostToolUse` and `Stop` never reached `CodexSignalSource`, so Crow's UI never observed Codex sessions transitioning to `working`/`done` — auto-respond gates and card-color updates were silently broken.

## Test plan
- [x] `swift test --filter CodexHookConfigWriter` — all 8 tests pass, including 2 new ones (`installGlobalTomlConfigMigratesLegacyCodexHooksKey`, `installGlobalConfigEmitsNoAsyncHooks`).
- [x] `swift test` for full CrowCodex package — 31 tests pass.
- [ ] Manual: seed `~/.codex/config.toml` with `[features]\ncodex_hooks = true`, launch Crow with a Codex session, confirm no deprecation/async-skip warnings and that the file ends up with `hooks = true` only.
- [ ] Manual: run a tool in a Codex session, confirm card color reflects `working → done` transitions.

Closes #494

### DIFF
--- a/Packages/CrowCodex/Sources/CrowCodex/CodexHookConfigWriter.swift
+++ b/Packages/CrowCodex/Sources/CrowCodex/CodexHookConfigWriter.swift
@@ -22,8 +22,12 @@ public struct CodexHookConfigWriter: HookConfigWriter {
         "PermissionRequest",
     ]
 
-    /// Post-execution events safe to run async (fire-and-forget).
-    private static let asyncEvents: Set<String> = ["PostToolUse", "Stop"]
+    /// Events that should run async (fire-and-forget). Codex's hook runtime
+    /// is sync-only as of v0.139.0 — declaring `async = true` causes the
+    /// entry to be silently skipped on startup, which breaks Crow's
+    /// session-state detection. Keep this empty until/unless Codex grows
+    /// real async support upstream.
+    private static let asyncEvents: Set<String> = []
 
     public init() {}
 
@@ -89,9 +93,15 @@ public struct CodexHookConfigWriter: HookConfigWriter {
     }
 
     /// Install or update `<codexHome>/config.toml` with the
-    /// `features.codex_hooks = true` flag and the Crow `notify` line.
+    /// `features.hooks = true` flag and the Crow `notify` line.
     /// Preserves any other user-authored config — minimal line-oriented merge
     /// avoids pulling in a TOML dependency for two simple keys.
+    ///
+    /// Also runs a one-shot migration: legacy installs (Crow before this
+    /// fix, or older Codex versions) wrote `codex_hooks = true` under
+    /// `[features]`. Codex v0.139.0+ renamed the key to `hooks` and emits
+    /// a deprecation warning for the old one — strip it so users converging
+    /// from older configs end up with a single, current entry.
     public static func installGlobalTomlConfig(codexHome: String, crowPath: String) throws {
         try FileManager.default.createDirectory(atPath: codexHome, withIntermediateDirectories: true)
         let tomlPath = (codexHome as NSString).appendingPathComponent("config.toml")
@@ -104,11 +114,14 @@ public struct CodexHookConfigWriter: HookConfigWriter {
 
         let notifyLine = "notify = [\"\(escapeTomlString(crowPath))\", \"codex-notify\"]"
         content = upsertTomlLine(content, key: "notify", line: notifyLine)
+        // Strip the deprecated `codex_hooks` key before writing the modern
+        // `hooks` key so we don't leave both lines behind on migration.
+        content = removeTomlSectionLine(content, section: "features", key: "codex_hooks")
         content = upsertTomlSectionLine(
             content,
             section: "features",
-            key: "codex_hooks",
-            line: "codex_hooks = true"
+            key: "hooks",
+            line: "hooks = true"
         )
 
         try content.write(toFile: tomlPath, atomically: true, encoding: .utf8)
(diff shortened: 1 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No untrusted input. `crowPath` is escaped via `escapeTomlString` before being embedded in the TOML `notify` line; the hooks.json command path flows through `JSONSerialization`.
- File writes are atomic (`write(toFile:atomically:true)` / `Data.write`), avoiding partial-write corruption of user config.
- The line-oriented merge preserves user-authored config rather than clobbering `config.toml` / `hooks.json`.

**Concerns:**
- None.

### Code Quality
- The migration ordering is correct: `removeTomlSectionLine(..., key: "codex_hooks")` runs *before* `upsertTomlSectionLine(..., key: "hooks")`. Since `lineKey` treats `codex_hooks` and `hooks` as distinct keys, stripping the legacy line first is what prevents both keys lingering — verified by `installGlobalTomlConfigMigratesLegacyCodexHooksKey`, including the idempotency re-run assertion.
- `removeTomlSectionLine` mirrors the existing `upsertTomlSectionLine` section-scan logic (exact `[features]` header match, bounded to the next section header), so it stays consistent with the file's established minimal-merge style and won't touch keys in other sections.
- Emptying `asyncEvents` cleanly removes the `async` flag in `generateHooks`; because `PostToolUse`/`Stop` are in `allEvents`, `installGlobalConfig` overwrites any stale async entries left in an existing `hooks.json`, so the hooks.json side migrates implicitly. `installGlobalConfigEmitsNoAsyncHooks` guards the regression.
- The rationale (Codex v0.139.0 sync-only, silent-skip) is captured in code comments, not just the PR body — good for the next reader.
- Migration is wired into launch at `Sources/Crow/App/AppDelegate.swift:440`, and no other `codex_hooks` references remain in source.

### Verification
- `swift test --filter CodexHookConfigWriter` → 8/8 pass, including the 2 new tests.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 0 Green] findings. Focused, correct, well-tested config fix with a sound idempotent migration.

---

PR #497, opened 2026-06-11
Hot-trigger corveil skill install when picker value changes (CROW-490)

### LINKED ISSUE #490: Hot-trigger corveil skill install when picker value changes (avoid app restart after configuring defaults.binaries.corveil)
**Friction:** #482/#483 installed the embedded `/query-corveil` slash command into `{devRoot}/.claude/commands/query-corveil.md` via a Scaffolder pass at app launch (`Scaffolder.swift:155` calls `installCorveilSkill(...)`, invoked from `AppDelegate.swift:418` initial-launch and `:1087` workspace-switch). Changing the picker value in Settings (`SettingsView.swift:603-609` writes `config.defaults.binaries["corveil"]`) doesn't fire a fresh Scaffolder pass, so the new path doesn't reach the installer until the user quits and relaunches Crow — or until they happen to switch workspaces and trip the alternate scaffold path.

The "Verify" button at `SettingsView.swift:615` only runs `--version`. It doesn't re-install.

Result: user picks a path → Settings persists it → nothing else happens. User reasonably assumes it's installed, then discovers it isn't until they restart Crow. The fix is small.

## Fix — hot-trigger the install on picker change

When `config.defaults.binaries["corveil"]` changes (path set, changed, or cleared), call `Scaffolder.installCorveilSkill(newPath)` once. Async, off the main thread (the existing `Scaffolder.corveilInstallTimeout` already bounds it; same `Task` wrapper SettingsView uses for `--version` at `:699`).

Three concrete places to wire:

1. **Settings picker `didSet` (or the SwiftUI binding's `set` closure)** at `SettingsView.swift:603-609`. After persisting the new value, dispatch `installCorveilSkill` against the new path. If the new value is empty (user cleared the field), call the installer with `nil` so it skips and clears any prior warning.

2. **Reuse the existing warning surface.** `installCorveilSkill` already returns a `String?` warning that's written to `AppState.corveilSkillInstallWarning` (per `AppState.swift:237`). Same plumbing applies here — surface success or the same banner that the launch-time pass uses.

3. **Don't re-run the rest of Scaffolder.** This is just the corveil install step; the full scaffold pass is unnecessary and would be wasteful. Extract `installCorveilSkill` to be callable independently if it isn't already (it appears private in `Scaffolder.swift:230` — make it accessible to Settings, either via a public Scaffolder method or via a small published service).

## Acceptance

1. **Initial pick:** With `defaults.binaries.corveil` unset and `query-corveil.md` not yet present, open Settings, set the picker to a valid corveil binary, save. Without restarting Crow, `{devRoot}/.claude/commands/query-corveil.md` exists and `/query-corveil` is invokable from any new Claude Code session in that devroot.
2. **Path change:** With a working install, change the picker to point at a different corveil binary (e.g. a rebuilt version). Without restarting, `{devRoot}/.claude/commands/query-corveil.md` reflects the new binary's embedded skill content (byte-compare against `<newPath> skill show`).
3. **Clear:** Setting the picker to empty triggers no install attempt and clears any prior `corveilSkillInstallWarning`. The existing `query-corveil.md` file is left in place — clearing the picker doesn't *remove* the skill (the user can do that manually if needed; not in scope here).
4. **Failure surfacing:** Picking a broken path (non-existent / not executable) sets `corveilSkillInstallWarning` to the same diagnostic the launch-time path produces, and the existing banner in the app surfaces it. No crash, no startup-loop, no broken file written.
5. **Rebuild rebuild loop:** With a stable picker pointing at a locally-built corveil, running `go build ./cmd/corveil` in the corveil repo and then clicking the Verify button (or any other trigger that re-evaluates) should *not* re-install (Verify is `--version` only — keep that contract). The user still gets a fresh install on the next launch, or could re-pick the same path to force one. (Optional follow-up: a "Reinstall skill" button next to Verify if this trips users; not in this ticket.)

## Critical files

| Purpose | Path |
|---|---|
| Picker binding (where to fire the install on change) | `Packages/CrowUI/Sources/CrowUI/SettingsView.swift:597-615` |
| Installer function | `Sources/Crow/App/Scaffolder.swift:230` (`installCorveilSkill`) — make accessible to Settings |
| Warning surface | `Packages/CrowCore/Sources/CrowCore/AppState.swift:237` (`corveilSkillInstallWarning`) |
| Existing async Task pattern to mirror | `Packages/CrowUI/Sources/CrowUI/SettingsView.swift:699` (the Verify Task wrapper) |

## Out of scope

- A "Reinstall skill" button (worth its own small follow-up if this ticket's hot-trigger doesn't cover the rebuild case well enough).
- Hot-trigger for the rest of `defaults.binaries.*` entries (codex, cursor). Those don't have a per-tool install step today — they only get the symlink + PATH treatment per #489 / #487, and that resolves dynamically from PATH at agent-launch time, so no re-scaffold is needed.
- Watching the corveil binary's mtime to auto-reinstall when the user rebuilds. Nice-to-have but out of scope; the launch-time install plus this ticket cover the common cases.

## Related

- #482 / #483 — Corveil CLI picker + initial Scaffolder hook for `corveil skill install`. This ticket closes the "set once, must restart" gap that flow left.
- #487 / #489 — `defaults.binaries.*` precedence in spawned terminals via symlink + PATH prepend. Adjacent but separate axis (those resolve at agent-launch time; this ticket is about installing skill content at config-change time).
- `AppDelegate.swift:418`, `:1087` — the two existing launch-time call sites that invoke the scaffold pass. This ticket is the third trigger (config change), reusing the same install function.

dgershman:
## Summary

- Settings → Corveil CLI picker now re-runs `corveil skill install` the moment a new path is committed (Browse confirm or Enter), so `/query-corveil` reflects the picked binary without an app restart.
- Promoted `Scaffolder.installCorveilSkill` from private to internal and added a new `onCorveilReinstall` closure to `SettingsView` so AppDelegate can dispatch just the install step (off the main thread, bounded by the existing 5s `corveilInstallTimeout`) instead of re-running the whole `Scaffolder.scaffold(...)` pass.
- Result is always assigned to the existing `AppState.corveilSkillInstallWarning` banner — successful or empty/cleared paths clear stale warnings; failures show the same diagnostic as the launch-time path. Verify stays `--version` only.

Closes #490

## Test plan

- [ ] **Initial pick:** Clear `defaults.binaries.corveil` and delete `{devRoot}/.claude/commands/query-corveil.md`. Open Settings → Browse to a working corveil binary. Without restarting Crow, confirm `query-corveil.md` is created.
- [ ] **Path change:** Point the picker at a freshly rebuilt corveil binary. Confirm `query-corveil.md` content updates to match `<newPath> skill install --path /dev/stdout`.
- [ ] **Clear:** Empty the field and press Enter. `query-corveil.md` stays in place; any prior banner clears.
- [ ] **Broken path:** Browse to a non-executable file. Banner shows the same "Corveil skill install skipped — binary at … is missing or not executable" warning that the launch-time path produces. No crash.
- [ ] **Verify still `--version` only:** Click Verify with a stable path; `query-corveil.md` mtime is unchanged.

### DIFF
--- a/Packages/CrowUI/Sources/CrowUI/SettingsView.swift
+++ b/Packages/CrowUI/Sources/CrowUI/SettingsView.swift
@@ -275,14 +275,7 @@ public struct SettingsView: View {
                 HStack {
                     TextField("Path to corveil binary", text: corveilBinding)
                         .textFieldStyle(.roundedBorder)
-                        .onSubmit {
-                            save()
-                            // Hot-trigger install on Enter — passes nil if the
-                            // user committed an empty field so callers can
-                            // clear any stale warning banner (CROW-490).
-                            let path = corveilBinding.wrappedValue
-                            onCorveilReinstall?(path.isEmpty ? nil : path)
-                        }
+                        .onSubmit { commitCorveilPath() }
                     Button("Browse...") {
                         let panel = NSOpenPanel()
                         panel.canChooseFiles = true
@@ -290,11 +283,9 @@ public struct SettingsView: View {
                         panel.allowsMultipleSelection = false
                         if panel.runModal() == .OK, let url = panel.url {
                             corveilBinding.wrappedValue = url.path
-                            save()
                             // Clear stale verify result — it's about a previous binary.
                             corveilVerifyResult = nil
-                            // Hot-trigger install on Browse confirm (CROW-490).
-                            onCorveilReinstall?(url.path)
+                            commitCorveilPath()
                         }
                     }
                     Button(corveilVerifying ? "Verifying…" : "Verify") { verifyCorveil() }
@@ -610,6 +601,21 @@ public struct SettingsView: View {
 
     private func save() {
         onSave?(devRoot, config)
+    }
+
+    /// Commit a corveil picker change: persist the config and hot-trigger
+    /// the `/query-corveil` install for the new path (CROW-490). Both
+    /// commit sites (Browse confirm and TextField `onSubmit`) funnel
+    /// through here so the "persist → reinstall" pair stays atomic and
+    /// the `nil`-on-empty rule has a single source of truth. We read the
+    /// path through `corveilBinding.wrappedValue` rather than the raw
+    /// input so the binding's whitespace-trim normalization wins — the
+    /// install closure sees the same string that the next launch's
+    /// scaffolder would see.
+    private func commitCorveilPath() {
+        save()
+        let path = corveilBinding.wrappedValue
+        onCorveilReinstall?(path.isEmpty ? nil : path)
     }
 
     /// Two-way binding into `config.defaults.binaries["corveil"]` that treats
--- a/Packages/CrowUI/Sources/CrowUI/SettingsView.swift
+++ b/Packages/CrowUI/Sources/CrowUI/SettingsView.swift
@@ -27,15 +27,24 @@ public struct SettingsView: View {
 
     public var onSave: ((String, AppConfig) -> Void)?
     public var onRescaffold: ((String) -> Void)?
+    /// Fired when the user commits a new value into the corveil picker
+    /// (Browse confirm or Enter on the TextField), so AppDelegate can
+    /// re-run just `Scaffolder.installCorveilSkill` instead of waiting for
+    /// the next app restart (CROW-490). `nil` argument means "the user
+    /// cleared the field" — the install is a no-op then but the caller
+    /// still gets the signal to clear any stale warning banner.
+    public var onCorveilReinstall: ((String?) -> Void)?
 
     public init(appState: AppState, devRoot: String, config: AppConfig,
                 onSave: ((String, AppConfig) -> Void)? = nil,
-                onRescaffold: ((String) -> Void)? = nil) {
+                onRescaffold: ((String) -> Void)? = nil,
+                onCorveilReinstall: ((String?) -> Void)? = nil) {
         self.appState = appState
         self._devRoot = State(initialValue: devRoot)
         self._config = State(initialValue: config)
         self.onSave = onSave
         self.onRescaffold = onRescaffold
+        self.onCorveilReinstall = onCorveilReinstall
     }
 
     /// Names of all workspaces except the one currently being edited.
@@ -266,7 +275,14 @@ public struct SettingsView: View {
                 HStack {
                     TextField("Path to corveil binary", text: corveilBinding)
                         .textFieldStyle(.roundedBorder)
-                        .onSubmit { save() }
+                        .onSubmit {
+                            save()
+                            // Hot-trigger install on Enter — passes nil if the
+                            // user committed an empty field so callers can
+                            // clear any stale warning banner (CROW-490).
+                            let path = corveilBinding.wrappedValue
+                            onCorveilReinstall?(path.isEmpty ? nil : path)
+                        }
                     Button("Browse...") {
                         let panel = NSOpenPanel()
                         panel.canChooseFiles = true
@@ -277,6 +293,8 @@ public struct SettingsView: View {
                             save()
                             // Clear stale verify result — it's about a previous binary.
                             corveilVerifyResult = nil
+                            // Hot-trigger install on Browse confirm (CROW-490).
+                            onCorveilReinstall?(url.path)
                         }
                     }
                     Button(corveilVerifying ? "Verifying…" : "Verify") { verifyCorveil() }
--- a/Sources/Crow/App/AppDelegate.swift
+++ b/Sources/Crow/App/AppDelegate.swift
@@ -53,6 +53,16 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// replaced.
     private var reviewKickoffTail: Task<Void, Never>?
 
+    /// Tail of the serial corveil-skill-install queue. Settings picker
+    /// commits (`SettingsView.onCorveilReinstall`) chain new installs onto
+    /// this tail so two rapid picks don't race on `query-corveil.md`
+    /// (concurrent `corveil skill install` subprocesses writing the same
+    /// `--path`) or on `corveilSkillInstallWarning` (out-of-order
+    /// completion clobbering a fresher banner with a stale one). The last
+    /// committed path is also the last to write the banner. See the
+    /// CROW-490 review for the race this replaced.
+    private var corveilInstallTail: Task<Void, Never>?
+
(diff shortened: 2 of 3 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Tight, well-scoped change that closes the "must restart to pick up a new corveil binary" gap. The inline documentation (AppDelegate.swift:1103-1119, Scaffolder.swift:231-234) is genuinely good — it explains the off-main-thread rationale, the always-assign banner semantics, and the `devRoot` capture. Reusing `installCorveilSkill` + the existing `corveilSkillInstallWarning` surface instead of re-running the whole `Scaffolder.scaffold(...)` pass is the right call.

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface. The corveil path is user-supplied via the same Settings field that already drove launch-time install; `installCorveilSkill` re-validates `isExecutableFile` and runs the binary with a fixed argv (`skill install --path <target>`) — no shell, no interpolation, no injection vector.
- Subprocess is bounded by the existing 5s `corveilInstallTimeout` watchdog (SIGTERM), stdout routed to `/dev/null`, stderr drained after exit — the deadlock-safety story is preserved on the new code path.
- Moving the install off the main thread via `Task.detached` is correct and avoids freezing the Settings window for up to 5s.

**Concerns:**
- None security-specific.

### Code Quality

**Yellow — unsynchronized concurrent installs race on the output file and the banner** (`AppDelegate.swift:1120-1126`)
Each commit (Browse confirm or Enter) fires `onCorveilReinstall`, which spawns a fresh `Task.detached` with no serialization, no cancellation of a prior in-flight install, and no staleness guard. Two commits within the install window (~hundreds of ms) produce two problems:
1. **Stale banner:** completion order is not guaranteed. Browse to a broken binary (task A → warning), then immediately Browse/Enter to a good binary (task B → `nil`); if A finishes after B, the banner shows a failure for the *superseded* path while the current path is actually fine. The warning no longer reliably reflects the committed path.
2. **Concurrent writes to the same `query-corveil.md`:** two `corveil skill install` processes writing the same `--path` target at once can interleave/corrupt the file unless corveil writes atomically (which we can't assume).

The launch-time path is single-shot and serialized; this new path removes that guarantee. Likelihood is low (requires two commits within the install window), but the fix is small — e.g. a monotonically-increasing generation counter captured into the task, applying the result only if it's still the latest, and/or routing installs through a serial actor/queue so they don't overlap. Worth landing in this round trip.

### Considerations (Green)
- `SettingsView.swift:297` passes the raw `url.path` while the stored binding (`corveilBinding`, line 619) trims whitespace; `installCorveilSkill` trims internally so this is harmless, but passing `corveilBinding.wrappedValue` would keep the two commit paths consistent.
- The `onSubmit` and Browse handlers duplicate the "commit → reinstall" logic; a tiny `commitCorveil(_:)` helper would DRY the two call sites and centralize the `nil`-on-empty rule.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The Yellow (unsynchronized concurrent installs racing on the banner and the output file) is a genuine should-fix; everything else is positive or optional.

---

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface. The picker only invokes a user-chosen local binary (`<path> skill install --path <devRoot>/.claude/commands/query-corveil.md`) that the user must explicitly select via `NSOpenPanel` or type — the same trust boundary as the pre-existing launch-time path.
- Executable check is preserved: `installCorveilSkill` still gates on `FileManager.isExecutableFile` and surfaces the identical "missing or not executable" warning, so a non-executable pick can't be run.
- No command-string interpolation — `Process.arguments` is an array, so the binary path can't inject extra args. `stdout` is routed to `nullDevice` (deadlock-proof), `stderr` is bounded and trimmed.
- The 5s `corveilInstallTimeout` SIGTERM watchdog still bounds a hung binary, now off the main thread so it can't freeze the Settings window.

**Concerns:** None.

### Code Quality
- Correct serialization: `enqueueCorveilInstall` chains onto `corveilInstallTail` exactly like the existing `reviewKickoffTail` (await predecessor → run → assign on main). Two rapid picks can't race on `query-corveil.md` or clobber `corveilSkillInstallWarning` out of order, and the last-committed path is the last to write the banner.
- Off-main execution is right: the blocking subprocess runs in a nested `Task.detached` over value types (`Scaffolder` struct + `String` devRoot, no `self` capture), result assigned back on `@MainActor`. `AppDelegate` is `@MainActor`-isolated, so the closure/`enqueueCorveilInstall` isolation hops compile cleanly (consistent with the adjacent `onRescaffold` closure mutating `appState` directly).
- "Always assign" banner semantics match the launch-time / `onRescaffold` path: a successful or `nil`/empty path clears a stale warning; failures show the same diagnostic. The `nil`-on-empty rule is funneled through the single `commitCorveilPath` site for both Browse and `onSubmit`.
- Whitespace normalization is handled — `commitCorveilPath` reads `corveilBinding.wrappedValue` (already trimmed by the binding's setter), so the install sees the same string the next launch's scaffolder would.
- `installCorveilSkill` visibility was widened only from `private` to `internal` (not `public`) — the minimum needed for the same-module `AppDelegate` caller. Good restraint.
- `saveSettings` only persists config (no scaffold/install), so `commitCorveilPath`'s `save()` + `onCorveilReinstall` is a single install trigger — no accidental double-install.

**Consider (non-blocking):**
- The "Re-scaffold .claude/ directory" button (`onRescaffold`) runs a full scaffold — including its own `corveil skill install` writing the same `query-corveil.md` and `corveilSkillInstallWarning` — on a *separate* domain from `corveilInstallTail`. A user who commits a picker change (starting a detached install) and then clicks Re-scaffold within the 5s window could run two `corveil skill install` subprocesses against the same target concurrently. It's narrow (requires deliberate concurrent action), self-healing (the next install/launch rewrites the file), and carries no crash/security impact, so it's out of this PR's stated picker-vs-picker scope — but routing `onRescaffold`'s corveil step through the same tail later would close it fully.

### Notes
- Could not run `swift build` / tests in this review checkout: the vendored `Frameworks/GhosttyKit.xcframework` has no binary artifact here, so the app target won't link. This is an environment limitation, not a PR defect. Review is based on close reading; the change is small, type-safe Swift consistent with an existing in-repo pattern. The PR's test plan is manual (UI + concurrency glue that's impractical to unit-test).

### Summary Table

---

PR #498, opened 2026-06-11
Add Reinstall skill button to Settings → Corveil CLI (CROW-491)

### LINKED ISSUE #491: Settings → Corveil CLI: add 'Reinstall skill' button next to Verify (manual rebuild-pickup without restart)
**Ask:** Add a **Reinstall skill** button in Settings → General → Corveil CLI, next to the existing **Verify** button (`SettingsView.swift:615` for Verify's neighbor placement). Clicking it runs the same `corveil skill install --path {devRoot}/.claude/commands/query-corveil.md` flow that the launch-time and (per #490) config-change paths use — without requiring an app restart, a workspace switch, or re-picking the path.

## Why this is its own button

#490 hot-triggers the install when the picker value *changes*. But the most common rebuild loop doesn't change the picker — the user runs `go build ./cmd/corveil` in the corveil repo, the binary at the configured path now contains a newer embedded skill version, and they want that newer version installed into their devroot. With #490 alone, the user has three workarounds:

1. Restart Crow (heavy — closes every session).
2. Switch workspaces (alternate scaffold path; awkward side-channel).
3. Re-pick the same path in Settings (works but unintuitive — "why am I telling it the same thing?").

A dedicated **Reinstall skill** button names the action and makes it discoverable. Verify already establishes the "Settings has interactive buttons next to the picker" pattern; this is the parallel for "I rebuilt corveil, install the new skill."

## Fix

`SettingsView.swift` — adjacent to the Verify button at `:615`:

```swift
Button("Reinstall skill") {
    Task { @MainActor in
        let warning = await Scaffolder.installCorveilSkill(
            config.defaults.binaries["corveil"]
        )
        appState.corveilSkillInstallWarning = warning
        // Surface a transient toast/banner: "Skill reinstalled" or the warning.
    }
}
.disabled((config.defaults.binaries["corveil"] ?? "").isEmpty)
```

Reuse the same `installCorveilSkill` function #490 exposes (this ticket depends on #490 landing the function-accessibility refactor; if #490 keeps it private, surface it here too).

The button is **disabled when the picker is unset** — there's no path to install from, and the launch-time install would also no-op. Tooltip on the disabled state: "Set the Corveil CLI path first."

## Acceptance

1. With a valid corveil path configured, click Reinstall skill. Without restart, `{devRoot}/.claude/commands/query-corveil.md` matches `<corveilPath> skill show` byte-for-byte (proves the new embedded skill content from the rebuilt binary installed).
2. Transient success indication (toast / "✓ Skill reinstalled" label that fades after a few seconds — match the Verify button's success surface for consistency).
3. With a broken path (set but non-existent / non-executable), clicking Reinstall surfaces the same diagnostic as the launch-time path via `corveilSkillInstallWarning`. No crash.
4. Button is disabled and visually grayed when the picker is unset.
5. Repeated clicks are idempotent (the underlying install is — `os.WriteFile` + `os.Chmod`).

## Critical files

| Purpose | Path |
|---|---|
| Button placement | `Packages/CrowUI/Sources/CrowUI/SettingsView.swift` (next to Verify around `:615`) |
| Installer function (already exposed by #490) | `Sources/Crow/App/Scaffolder.swift:230` (`installCorveilSkill`) |
| Warning surface (reuse) | `Packages/CrowCore/Sources/CrowCore/AppState.swift:237` (`corveilSkillInstallWarning`) |

## Out of scope

- Auto-detection of corveil binary mtime changes (would obviate this button by reinstalling automatically when the binary is rebuilt). Worth its own ticket if this button proves friction-inducing in daily use — for now, explicit click is fine.
- Adding equivalent buttons for codex/cursor — those don't have a per-tool install step; PATH precedence per #487/#489 covers them.

## Related

- #490 — hot-trigger install on picker change. This ticket depends on #490 making `installCorveilSkill` callable from outside `Scaffolder`'s private surface.
- #482 / #483 — the original picker + Scaffolder install hook this complements.

dgershman:
## Summary
- Adds a **Reinstall skill** button next to the existing **Verify** button in Settings → General → Corveil CLI.
- Runs the same `corveil skill install --path …` flow as the per-launch path (#483), on demand — no restart, workspace switch, or path re-pick required. Serves the common "I just rebuilt corveil locally; install the new embedded skill" loop.
- Disabled when the picker is unset; shows `✓ Skill reinstalled` / `✗ <diagnostic>` inline (coalesces with Verify's result line — only one runs at a time). A successful manual reinstall also clears the launch-time warning banner; a failed one updates it, so "is corveil broken?" stays single-sourced.

## Plumbing
SettingsView lives in the CrowUI package, which cannot import the Crow app target where `Scaffolder` lives. The call therefore goes through a new `AppState.onReinstallCorveilSkill` async closure wired in `AppDelegate.launchMainApp` — same callback-injection pattern as `onCreateManager`, `onRestartManager`, etc. The closure offloads the synchronous `Process.waitUntilExit` work to a detached task internally, so awaiting it from the main actor doesn't pin the UI for the 5-second install timeout.

`Scaffolder.installCorveilSkill` access is loosened from `private` to `internal` so `AppDelegate` can call it on a fresh `Scaffolder` instance (this is the "function-accessibility refactor" the ticket references — #490 has not yet landed in this branch).

## Test plan
- [ ] Configure a valid `corveil` path → click **Reinstall skill** → expect transient `✓ Skill reinstalled`; `{devRoot}/.claude/commands/query-corveil.md` byte-matches `<corveilPath> skill show`
- [ ] Rebuild corveil locally with a tweaked embedded skill string → click **Reinstall skill** again without restarting Crow → expect file content to reflect the new build
- [ ] Set path to a non-executable junk value → click → expect `✗ <diagnostic>` inline AND the orange warning banner at the top of General (existing `corveilSkillInstallWarning` surface)
- [ ] Successful reinstall after a failure clears the orange banner
- [ ] Clear the path → button is disabled with tooltip "Set the Corveil CLI path first."
- [ ] Click repeatedly with a valid path → no errors (install is idempotent via `os.WriteFile` + `os.Chmod`)
- [ ] Verify button and Reinstall button mutually disable while either is running

Closes #491

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -384,6 +384,16 @@ public final class AppState {
     /// Fire a job immediately, ignoring its enabled flag and schedule (job ID).
     public var onRunJob: ((UUID) -> Void)?
 
+    /// Triggered by Settings → General → Corveil CLI → "Reinstall skill"
+    /// (issue #491). Wired by `AppDelegate.launchMainApp` to construct a
+    /// fresh `Scaffolder` and call `installCorveilSkill(path)` — the same
+    /// flow as the per-launch install, on demand. Returns `nil` on success,
+    /// a user-facing warning string on failure. The closure is `async` so
+    /// the wiring can offload the synchronous `Process.waitUntilExit` work
+    /// to a detached task without the call site blocking the main actor
+    /// for the 5-second install timeout.
+    public var onReinstallCorveilSkill: ((String?) async -> String?)?
+
     // MARK: - Computed Properties
 
     public var selectedSession: Session? {
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -384,16 +384,6 @@ public final class AppState {
     /// Fire a job immediately, ignoring its enabled flag and schedule (job ID).
     public var onRunJob: ((UUID) -> Void)?
 
-    /// Triggered by Settings → General → Corveil CLI → "Reinstall skill"
-    /// (issue #491). Wired by `AppDelegate.launchMainApp` to construct a
-    /// fresh `Scaffolder` and call `installCorveilSkill(path)` — the same
-    /// flow as the per-launch install, on demand. Returns `nil` on success,
-    /// a user-facing warning string on failure. The closure is `async` so
-    /// the wiring can offload the synchronous `Process.waitUntilExit` work
-    /// to a detached task without the call site blocking the main actor
-    /// for the 5-second install timeout.
-    public var onReinstallCorveilSkill: ((String?) async -> String?)?
-
     // MARK: - Computed Properties
 
     public var selectedSession: Session? {
--- a/Packages/CrowUI/Sources/CrowUI/SettingsView.swift
+++ b/Packages/CrowUI/Sources/CrowUI/SettingsView.swift
@@ -24,6 +24,13 @@ public struct SettingsView: View {
     @State private var corveilVerifyResult: String?
     /// True while the Verify button's subprocess is in flight.
     @State private var corveilVerifying: Bool = false
+    /// Live result of the most recent "Reinstall skill" click. Shares the
+    /// `✓ … / ✗ …` convention with `corveilVerifyResult` and is rendered in
+    /// the same inline result line (only one operation runs at a time). Set
+    /// to `nil` on path edits so stale results don't outlive a binary swap.
+    @State private var corveilReinstallResult: String?
+    /// True while the Reinstall skill button's subprocess is in flight.
+    @State private var corveilReinstalling: Bool = false
 
     public var onSave: ((String, AppConfig) -> Void)?
     public var onRescaffold: ((String) -> Void)?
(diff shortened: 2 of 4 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Tidy, well-documented feature. The callback-injection pattern matches the existing `onCreateManager`/`onListWorkspaceRepos` precedent, the loosening of `installCorveilSkill` to `internal` is appropriately scoped, the single coalesced result line with mutual `disabled` is a clean way to avoid two competing status lines, and the threading story (async closure offloading `Process.waitUntilExit` to a detached task) keeps the main actor unpinned for the 5s timeout. `CrowCore` builds clean; `CrowUI`/app target couldn't be built locally (missing `GhosttyKit.xcframework` artifact — pre-existing infra, unrelated to this PR), so those two files were reviewed by inspection.

### Security Review
**Strengths:**
- No new external input surface. The reinstall path reuses `Scaffolder.installCorveilSkill`, which already validates the binary is executable (`isExecutableFile`), bounds runtime with a SIGTERM watchdog, discards stdout to a null device to avoid pipe-write deadlock, and routes the binary via `executableURL` + `arguments` (no shell, no injection vector).
- Failure diagnostics surfaced to the user come from corveil's own stderr, already truncated/trimmed; nothing sensitive added.

No security concerns.

### Code Quality

**Yellow — stale `devRoot` capture: reinstall can target the wrong directory and still report success.**
`Sources/Crow/App/AppDelegate.swift:433-438` — the closure captures the launch-time `let devRoot` (line 330) and builds `Scaffolder(devRoot: devRoot)` from it. But `saveSettings(devRoot:config:)` (`AppDelegate.swift:1153-1154`) reassigns `self.devRoot` when the user edits **Development Root** — which lives in the *same Settings window* as the Reinstall button. So the sequence \"change devRoot → scroll down → click Reinstall skill\" installs `query-corveil.md` into the **previous** devRoot's `.claude/commands/`, while the UI still reports `✓ Skill reinstalled`. This diverges from the sibling closure `onListWorkspaceRepos` (`AppDelegate.swift:847-848`), which deliberately uses `[weak self]` + live `self.devRoot` for exactly this reason. Since the whole selling point of the button is \"no restart required,\" silently honoring a stale devRoot is surprising. Suggest `{ [weak self] path in ... read self?.devRoot on the main actor, then pass into Task.detached }`, mirroring `onListWorkspaceRepos`.

**Green — false `✓ Skill reinstalled` when the closure is unwired.**
`SettingsView.swift:675-688` — `await appState.onReinstallCorveilSkill?(path)` returns `nil` both on real success *and* when the closure isn't set (previews/unit tests). `nil` falls into the success branch, printing `✓ Skill reinstalled` and clearing the warning banner for an install that never ran. Harmless in production (always wired), but the \"degrade gracefully\" comment oversells it — it degrades to a false positive. Consider a sentinel/explicit nil-closure branch if previews ever exercise this.

**Green — result line not cleared on manual path edits.**
The `corveilReinstallResult` doc comment says it's cleared \"on path edits so stale results don't outlive a binary swap,\" but only the **Browse** button (`SettingsView.swift:285-287`) clears it. Typing a new path into the `TextField` + `onSubmit { save() }` (`SettingsView.swift:274-276`) leaves a stale `✓ Skill reinstalled` visible against the new path. Functionality is unaffected (disabled/run logic reads the live binding); only the displayed line is briefly stale. Consider clearing both results in an `.onChange` of the binding, or soften the comment.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The Yellow (stale devRoot) is the only blocker; addressing it in the same round avoids a follow-up.

---

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface. Reinstall reuses the existing per-launch `installCorveilSkill` flow — the user-configured `corveil` binary is invoked with an argument **array** (`["skill", "install", "--path", target]`), so there is no shell/command-injection vector.
- `isExecutableFile` is still gated before launch; the failure path returns a sanitized, user-facing diagnostic and logs via `NSLog` (no secrets, no network).
- `target` is derived from `devRoot`, not from untrusted input.

**Concerns:** None.

### Code Quality
Strong, idiomatic work — the threading claims in the description check out:

- **Concurrency is race-free.** `SettingsView` is main-actor-isolated (confirmed by the pre-existing synchronous `@State` mutation at `SettingsView.swift:648`, which only compiles under swift-tools 6.0 if the method is `@MainActor`). The new `Task { }` in `reinstallCorveilSkill` (`SettingsView.swift:679`) therefore inherits main-actor isolation, making the direct `@State`/`appState` mutations safe, while the blocking `Process.waitUntilExit` is correctly offloaded to `Task.detached` inside the `AppDelegate` closure (`AppDelegate.swift:438`).
- **Live `devRoot` read** (`AppDelegate.swift:437`) correctly avoids installing into a stale launch-time devRoot after an in-window Settings change.
- **Unwired-callback guard** (`SettingsView.swift:685`) properly distinguishes "real success (nil)" from "never ran," preventing a false `✓`.
- **Single source of truth** for the warning banner is preserved — success clears `corveilSkillInstallWarning`, failure updates it.
- `private`→`internal` on `installCorveilSkill` is module-scoped and well documented.

Green considerations (non-blocking):
- **In-flight path edit staleness** (`SettingsView.swift:671`): the path `TextField` stays editable while a reinstall is running (only the buttons disable). If the user edits the path during the install window, the completing `Task` writes `✓ Skill reinstalled` next to the newly-typed path. The install itself is correct (it captured the path at click time + reads devRoot live); this is purely cosmetic and self-corrects on the next action.
- **Theoretical false-success** (`AppDelegate.swift:437`): if `self` (AppDelegate) were deallocated, the closure returns `nil`, which the caller reads as success. AppDelegate lives for the app lifetime, so this is not reachable in practice.
- **No automated tests** for the new glue. The success/failure/unwired branching is thin async UI glue, but unlike `verifyCorveil` (which extracted `runCorveilVersion` as a pure, testable helper) nothing here is unit-tested. Acceptable given the manual test plan.

### Summary Table

---

PR #499, opened 2026-06-11
Fix Codex launch — deliver prompt + drop deprecated/async hook config (CROW-492)

### LINKED ISSUE #492: Codex coding session launches bare (no prompt) and writes deprecated config keys + unsupported async hooks
**Observed when a Codex coding session opens in Crow:**

```
cd /Users/danny/Projects/devroot/rm/crow-491-reinstall-skill-button && /Users/danny/.nvm/versions/node/v22.22.2/bin/codex
⚠ `[features].codex_hooks` is deprecated. Use `[features].hooks` instead.
Enable it with `--enable hooks` or `[features].hooks` in config.toml.

╭───────────────────────────────────────────────────╮
│ >_ OpenAI Codex (v0.139.0)                        │
│ model:     gpt-5.5 medium   /model to change      │
│ directory: ~/…/rm/crow-491-reinstall-skill-button │
╰───────────────────────────────────────────────────╯

⚠ skipping async hook in /Users/danny/.codex/hooks.json: async hooks are not supported yet
⚠ skipping async hook in /Users/danny/.codex/hooks.json: async hooks are not supported yet
```

Two distinct bugs, one launch flow, both in `Packages/CrowCodex/Sources/CrowCodex/OpenAICodexAgent.swift` and its hook writer.

---

## Bug 1: Codex launches with no prompt — agent sits idle waiting for the user to type

**Root cause:** `OpenAICodexAgent.autoLaunchCommand(...)` returns a bare `"codex\n"` (see the existing code path that I read in earlier diagnostics — function around line 47-64 in current HEAD). The inline comment even says "Bare `codex` launch — the user types their prompt into the TUI." That was an MVP choice when Codex support was first added — but it makes Codex coding sessions *functionally non-autonomous*: Crow opens the worktree, Codex starts, no prompt is delivered, the agent just waits at an empty TUI for the user to paste in the ticket context manually.

By comparison, the Claude Code launcher feeds the agent a prompt via `claude --permission-mode plan "$(cat <prompt-file>)"` (see `crow-workspace/setup.sh` and Claude's `autoLaunchCommand`). The Codex equivalent should send the same prompt content into Codex's CLI so the session starts already working.

**Fix shape:**

Confirm via `codex --help` which entry point accepts a non-interactive prompt — likely `codex exec "<prompt>"` (one-shot) or a positional/`--message` form for the interactive TUI. From the Codex docs, `codex` (interactive TUI) accepts an initial message as a positional argument: `codex "your prompt here"`. Update `OpenAICodexAgent.autoLaunchCommand` to:

```swift
public func autoLaunchCommand(
    session: Session,
    worktreePath: String,
    ...
) -> String? {
    guard session.kind == .work else { return nil }
    // Read the same prompt file Claude's path uses — same content, same location.
    let promptPath = "\(devRoot)/.claude/prompts/crow-prompt-\(session.name).md"
    // Codex takes an initial positional message. Quote and escape so multi-line
    // content + shell metacharacters reach the CLI intact. Newline at the end
    // submits.
    return "codex \"$(cat \(posixQuote(promptPath)))\"\n"
}
```

If the positional doesn't accept the full prompt body cleanly (size limits, multiline), fall back to `codex exec` with the prompt file via stdin: `codex exec < <prompt-file>` (then the session runs the prompt non-interactively, completes the round, exits — not ideal for an interactive session but workable for one-shot work).

**Acceptance for Bug 1:**
- Launching a Codex coding session against any ticket opens the Codex TUI with the ticket's prompt already submitted as the first message. Codex starts working without the user typing anything.
- The prompt content matches what the Claude Code path would have sent (same `crow-prompt-<session>.md` source).
- `session.kind == .review` still returns `nil` (review-on-Codex is still gated out, per the existing comment).

---

## Bug 2: Crow writes deprecated `[features].codex_hooks` + async hooks Codex can't run

**Symptoms (from the launch above):**
- `⚠ [features].codex_hooks is deprecated. Use [features].hooks instead.` — Codex CLI evolved its `config.toml` feature-flag name; Crow's hook config writer still uses the old key.
- `⚠ skipping async hook in /Users/danny/.codex/hooks.json: async hooks are not supported yet` (×2) — Crow declares async hooks (probably to mirror Claude Code's hook pattern), but Codex's hook system is sync-only. The hooks Crow registered are being silently no-op'd every launch.

**Root cause:** `OpenAICodexAgent.swift` references a `hookConfigWriter: any HookConfigWriter = CodexHookConfigWriter()` (see line 16 area). That writer is what produces:
- Whatever updates `~/.codex/config.toml` for the feature flag.
- `~/.codex/hooks.json` with the hook definitions.

Both writes need updating to match current Codex CLI (v0.139.0):

1. **`config.toml` feature flag** — change `[features].codex_hooks` to `[features].hooks`. If the writer is the source of that key, fix the key it writes. If the user wrote it manually (legacy from a previous Crow install), the writer should detect a `codex_hooks = …` value on read, migrate to `hooks = …`, and remove the legacy key.

2. **`hooks.json` async declarations** — find the hook definitions Crow writes (likely in `CodexHookConfigWriter.swift`), drop the `async: true` (or whatever Crow flags those hooks with), or omit them entirely if they only make sense as async. If Crow's session-state plumbing relies on those hooks firing to detect activity/completion, those signals are silently lost today — the Codex session state from Crow's POV may be unreliable.

3. **Document why these hooks exist.** The codex hooks Crow registers are presumably about detecting agent state for the same purposes ClaudeCodeAgent uses hooks (auto-respond, completion detection, the card-color update from #428). If the async hooks are being silently dropped today, those features are silently broken for Codex sessions even when the agent does run. Worth verifying alongside this fix.

**Acceptance for Bug 2:**
- A fresh Crow launch with Codex configured produces no `[features].codex_hooks is deprecated` warning from Codex's startup.
- No `⚠ skipping async hook ...` warnings — either all Codex hooks are sync (and registered correctly), or the unsupported hooks are removed from `hooks.json` so Codex doesn't see them.
- Whatever session-state signals Crow expects from Codex's hooks (state, agent-waiting events, completion) actually fire — verify the auto-respond and card-color flows work for a Codex session, not just for Claude Code.
- Migrate legacy `[features].codex_hooks` entries in user config.toml: read, rename to `hooks`, log a one-line migration notice. Idempotent — running twice is a no-op the second time.

---

## Critical files

| Purpose | Path |
|---|---|
| Agent launcher (Bug 1) | `Packages/CrowCodex/Sources/CrowCodex/OpenAICodexAgent.swift` (`autoLaunchCommand`, ~lines 47-64 of current HEAD) |
| Hook config writer (Bug 2) | `Packages/CrowCodex/Sources/CrowCodex/CodexHookConfigWriter.swift` (or wherever the conformer to `HookConfigWriter` lives for Codex) |
| Prompt-file convention to mirror | `Packages/CrowClaude/Sources/CrowClaude/ClaudeCodeAgent.swift` + `crow/skills/crow-workspace/setup.sh:572` for the `crow-prompt-<session>.md` path |
| Where `codex --help` confirms the CLI form | Local — runtime check, no source path |

## Out of scope

- Codex-as-reviewer (still gated to `.work` per the existing guard). Reviews-on-Codex is its own ticket if/when it makes sense.
- Sync vs async hook architecture redesign — this ticket just gets Crow back to a working state with Codex's current sync-only hook model. If Crow's auto-respond integration needs async semantics that Codex doesn't support, that's a follow-up with Codex's roadmap, not this ticket.
- The wider "agent-kind plumbing" thread (#421/#427/#440/#443) — those are about the abstraction layer; this ticket is the Codex-specific concrete launch + hook bug.

## Related

- #424 — original "Cursor agent launches terminal but does nothing" bug. Same shape as Bug 1 (agent launches bare, no prompt). If Cursor's `autoLaunchCommand` has the same issue, fix it in parallel under that ticket or roll it in here.
- #428 — Cursor session card color doesn't update on completion. Identical category of problem to Bug 2's "async hooks silently dropped, so completion detection is silently broken." Worth checking if Codex sessions have the same card-color drift after this lands.
- #484 / #485 — codex binary discovery via PATH walk + `defaults.binaries`. Adjacent (where the binary lives) but separate from this ticket (how Crow invokes it).

dgershman:
Closes #492

## Summary
- **Bug 1 — Codex launched bare.** `launch_codex` in the workspace skill (and its template copy) now passes `\"\$(cat \$prompt_path)\"` as Codex's positional argv so the TUI starts with the ticket prompt pre-filled, matching Codex 0.129+ behavior and mirroring `launch_cursor`. The stale "Codex has no prompt-argv form" log + comment go with it. `OpenAICodexAgent.autoLaunchCommand` keeps the bare `codex\n` for in-app re-launches (matches `CursorAgent` `.work` shape); the misleading comment is refreshed to point at the skill as the first-launch prompt path.
- **Bug 2 — deprecated TOML key.** `CodexHookConfigWriter.installGlobalTomlConfig` now writes `[features].hooks = true` (renamed in Codex 0.129; the old `codex_hooks` alias emits a deprecation warning). Any legacy `codex_hooks` entry is migrated/removed via a new `removeTomlSectionLine` helper — idempotent on subsequent runs.
- **Bug 2 — unsupported async hooks.** Dropped `\"async\": true` from `PostToolUse` and `Stop` hook entries. Codex 0.139 still silently skips async hooks at execution, which meant `CodexSignalSource`'s state-signal pipeline (card-color updates, auto-respond, completion detection) was silently broken for those two events — not just noisy. All six hooks now run sync and reach Crow.

## Test plan
- [x] `arch -arm64 swift test --package-path Packages/CrowCodex` — all 31 tests pass, including:
  - `installGlobalTomlConfigCreatesFreshFile` updated to expect `hooks = true` and reject `codex_hooks`.
  - `installGlobalTomlConfigPreservesUserSettings` updated the same way.
  - New `installGlobalTomlConfigMigratesLegacyCodexHooksKey` — pre-seeds `codex_hooks = true`, asserts it is replaced with `hooks = true`, and asserts idempotency on a second run.
  - New `installGlobalConfigDoesNotDeclareAsyncHooks` — walks every event in `hooks.json` and asserts no entry sets `"async"`.
- [ ] Manual: launch a Codex coding session via the workspace skill — TUI opens with the ticket prompt already in the composer, zero startup warnings.
- [ ] Manual: pre-seed `~/.codex/config.toml` with `[features].codex_hooks = true`, restart Crow, confirm migration to `hooks = true`.
- [ ] Manual: trigger a tool call in a Codex session and watch the card color reflect `PreToolUse` → `PostToolUse` → `Stop`. None of this worked before.

### DIFF
--- a/Packages/CrowCodex/Sources/CrowCodex/OpenAICodexAgent.swift
+++ b/Packages/CrowCodex/Sources/CrowCodex/OpenAICodexAgent.swift
@@ -50,10 +50,15 @@ public struct OpenAICodexAgent: CodingAgent {
         // log and skip rather than producing a malformed command.
         guard session.kind == .work else { return nil }
 
-        // Bare `codex` launch — the user types their prompt into the TUI.
-        // No env prefix (Codex has no OTEL equivalent), no `--continue`
-        // (MVP doesn't auto-resume), no `--rc` (Codex doesn't do remote
-        // control). The terminal's cwd is already the worktree path.
+        // Bare `codex` launch in-app. First-launch prompt delivery happens
+        // in the workspace skill (`crow-workspace/setup.sh launch_codex`),
+        // which passes the prompt file as Codex's positional argv before
+        // handing the terminal off to Crow. In-app re-launches resume the
+        // existing TUI rather than re-running the original prompt — same
+        // shape as `CursorAgent` `.work` (no `--continue` equivalent in MVP).
+        // No env prefix (Codex has no OTEL equivalent), no `--rc` (Codex
+        // doesn't do remote control). The terminal's cwd is already the
+        // worktree path.
         return "codex\n"
     }
 
--- a/Resources/crow-workspace-setup.sh.template
+++ b/Resources/crow-workspace-setup.sh.template
@@ -714,11 +714,11 @@ launch_codex() {
       die "launch_agent" "codex binary not found at PATH or known locations; provide --agent-binary"
   fi
   log "Resolved codex binary: $bin"
-  # Codex has no prompt-argv form (matches OpenAICodexAgent.autoLaunchCommand
-  # — bare `codex` only). The prompt file is still written; the user can
-  # paste from it into the TUI.
-  log "Note: Codex has no prompt-argv form; prompt file is at $prompt_path (paste manually if needed)."
-  local launch_cmd="cd $WORKTREE_PATH && $bin"
+  # Codex 0.129+ accepts the initial prompt as a positional argv that
+  # pre-fills the TUI composer — same mechanism Cursor's `agent` uses.
+  # The prompt-argv form was deferred in MVP because older Codex CLIs
+  # ignored extra argv; unblocked here (#492).
+  local launch_cmd="cd $WORKTREE_PATH && $bin \"\$(cat $prompt_path)\""
   create_agent_terminal "OpenAI Codex" "$launch_cmd"
 }
 
--- a/skills/crow-workspace/setup.sh
+++ b/skills/crow-workspace/setup.sh
@@ -837,11 +837,11 @@ launch_codex() {
       die "launch_agent" "codex binary not found at PATH or known locations; provide --agent-binary"
   fi
   log "Resolved codex binary: $bin"
-  # Codex has no prompt-argv form (matches OpenAICodexAgent.autoLaunchCommand
-  # — bare `codex` only). The prompt file is still written; the user can
-  # paste from it into the TUI.
-  log "Note: Codex has no prompt-argv form; prompt file is at $prompt_path (paste manually if needed)."
-  local launch_cmd="cd $WORKTREE_PATH && $bin"
+  # Codex 0.129+ accepts the initial prompt as a positional argv that
+  # pre-fills the TUI composer — same mechanism Cursor's `agent` uses.
+  # The prompt-argv form was deferred in MVP because older Codex CLIs
+  # ignored extra argv; unblocked here (#492).
+  local launch_cmd="cd $WORKTREE_PATH && $bin \"\$(cat $prompt_path)\""
   create_agent_terminal "OpenAI Codex" "$launch_cmd"
 }
 


dhilgaertner (approved):
## Code & Security Review

Tight, well-scoped bug-fix PR. Both reported bugs are fixed correctly, comments are accurate, and the new behavior mirrors established patterns in the codebase.

### Critical Issues
None.

### Security Review
**Strengths:**
- **Bug 1 (prompt argv).** `launch_codex` now passes `"\$(cat $prompt_path)"` exactly as `launch_cursor` (setup.sh:822) and `launch_claude_code` (setup.sh:797) already do. The `$(cat …)` is wrapped in double quotes and expanded by the *target* shell at paste time; command-substitution output is not re-parsed for quotes/`$`/backticks, so an arbitrary prompt body can't break out of the single argv. No new injection surface — it joins a well-trodden path.
- Template (`Resources/crow-workspace-setup.sh.template:721`) and skill copy (`skills/crow-workspace/setup.sh:844`) are kept in sync — no drift introduced.
- `escapeTomlString` still guards the `notify` path; no untrusted data flows into the new TOML helpers (`section`/`key` are compile-time literals).

**Concerns:** None.

### Code Quality
- **Bug 2a (deprecated key).** `removeTomlSectionLine` is correct and bounded: `sectionEnd` defaults to `lines.count` and is always `> start`, so the `(start+1)..<sectionEnd` range is always valid (no crash on an empty/last section). Exact-equality key matching (not prefix) means migrating `codex_hooks` can't accidentally clobber the new `hooks` key, and order-independence holds. Migration is idempotent — confirmed by the new test asserting `second == toml`.
- **Bug 2b (async hooks).** Dropping `async` is the right call given Codex silently drops the field. Note the fix is self-healing for existing installs too: `installGlobalConfig` overwrites `existingHooks[eventName]` for all six events, so a stale `hooks.json` carrying `"async": true` on `PostToolUse`/`Stop` gets rewritten without the flag on next launch.
- Tests are strong: the new `installGlobalConfigDoesNotDeclareAsyncHooks` walks every event/entry rather than spot-checking, and the migration test verifies user keys survive (`memories = true`) plus idempotency.
- Verified locally: `arch -arm64 swift test --package-path Packages/CrowCodex` → all 31 tests pass.

**Consider (non-blocking, Green):**
- `removeTomlSectionLine` only scans inside `[features]` and removes the first match. A theoretical legacy config with a *top-level* `codex_hooks` (outside any section) wouldn't be migrated — but Crow has only ever written it under `[features]`, so this is purely hypothetical and fine as-is.

### Summary Table

---

Issue #171, opened 2026-04-15
TicketCard: make issue and PR links clickable using LinkChip (match session header)


dhilgaertner:
## Problem

In the ticket list (ticket board), each row shows the issue title and — if a PR exists — a small PR badge. Neither is clickable: there's no way to jump from a ticket row out to the issue or PR on GitHub.

Inside a session, we already have the pattern I want: small click-through capsules in the session header (Issue #123, PR link, Repo link) that open the URL in a browser. I'd like the same treatment on every row in the ticket board.

## Expected

For each ticket row:
- A clickable **Issue #N** chip that opens the GitHub/GitLab issue URL
- A clickable **PR #N** chip that opens the PR URL — **only when a PR is associated with the issue**

Same visual style / interaction as the session-level chips for consistency.

## Current Behavior

- `Packages/CrowUI/Sources/CrowUI/TicketBoardView.swift` → `TicketCard` (lines 378–587)
- The PR badge is rendered at line 413 via `ticketPRBadge(number:url:)` (lines 496–510), but it's a plain styled capsule — no tap handler, no `NSWorkspace.shared.open`.
- The issue URL (`AssignedIssue.url`) is not surfaced as a chip at all.

## Reuse the Existing Component

`LinkChip` already exists and does exactly this — clickable capsule with icon + label that calls `NSWorkspace.shared.open`:

- **Component:** `Packages/CrowUI/Sources/CrowUI/SessionDetailView.swift:360-389`
- **Usage in session header:**
  - Issue chip: `SessionDetailView.swift:96-101` — `LinkChip(label: "Issue #\(...)", url: session.ticketURL, icon: "link")`
  - PR chip: `SessionDetailView.swift:105-107` — `LinkChip(label: link.label, url: link.url, icon: "arrow.triangle.pull")`

## Data Already Available

`Packages/CrowCore/Sources/CrowCore/Models/AssignedIssue.swift`:
- `url: String` (line 9) — issue URL, always present
- `prURL: String?` (line 16) — PR URL when linked

No new fetching or model changes needed.

## Proposed Fix

1. Make `LinkChip` accessible to `TicketBoardView` (move to a shared location in `CrowUI` if it isn't already, or extract to its own file).
2. In `TicketCard`, replace the non-clickable `ticketPRBadge` with a `LinkChip` for the PR (when `issue.prURL != nil`).
3. Add a second `LinkChip` for the issue itself using `issue.url`, so every row has at least the issue chip.
[...]

PR #521, opened 2026-06-16
Fix PR-link reconcile wrong-PR attribution (#520)

### LINKED ISSUE #520: PR-link reconcile attaches the wrong PR when worktree branch ≠ PR head branch (and can't be corrected)
## Summary

Crow attaches the **wrong PR** to many sessions whose worktree branch and the
session's actual PR branch differ by a repo-name prefix. The PR-link reconcile
pass exact-matches on `repoSlug#branch`; when that fails it fuzzy-attaches a
"recent PR" and lands on an arbitrary/wrong PR. The derived PR links are also
**not persisted** (they're absent from `store.json`), so the wrong links
regenerate on every refresh cycle and cannot be corrected by hand.

Discovered on the `max-monorepo` (Jira/MAXX) sessions: 5 of 12 sessions show the
wrong PR, including one session that has no PR at all but still shows one.

## Root cause

The worktrees were created on branches like
`feature/max-monorepo-maxx-7035-citations-chain-of-custody`, but the sessions
opened their PRs on `feature/maxx-7035-citations-chain-of-custody` — i.e. the
real PR head branch **drops the `max-monorepo-` repo prefix** that `setup.sh`
puts in the worktree branch name.

- `applySessionPRLinks()` (`Sources/Crow/App/IssueTracker.swift`) keys PRs by
  `"\(pr.repoNameWithOwner)#\(pr.headRefName)"` and looks up
  `byBranch["\(repoSlug)#\(branch)"]` where `branch` is the **worktree** branch.
  Because the worktree branch (`…/max-monorepo-maxx-7035-…`) never equals the PR
  head branch (`…/maxx-7035-…`), the exact match always misses → no link is
  created or persisted (explains the empty `store.json`).
- `reconcileMissingPRLinks()` → `fetchPRsForReconcile()` →
  `findRecentPRsForBranches()` then runs as the fallback. Its fuzzy
  branch/recent-PR attribution mis-assigns: it fans matches out across
  candidates and ends up pinning the wrong PR to a session (e.g. a session whose
  ticket has no PR still gets one).

## Evidence (max-monorepo, observed live via `crow list-links`)

Authoritative ticket→PR map built from PR titles (each title carries its MAXX
key) and head branches, vs. what each session currently shows:

| Session (ticket) | Correct PR | Shows | |
|---|---|---|---|
| MAXX-6963 workstation-doc-storage | #126 | #145 | wrong |
| MAXX-6969 size-bytes-zero | #130 | #135 | wrong |
| MAXX-6971 classifier-skill-id | #131 | #131 | ok |
| MAXX-6972 cascade-output-unwrap | #132 | #132 | ok |
| MAXX-6973 ci-db-migrate-race | #135 | #135 | ok |
| MAXX-6975 docs-upload-path-uuid | #143 | #146 | wrong |
| MAXX-6987 titan-assess-reuse-spike | #141 | #141 | ok |
| MAXX-6989 adr-convention | #145 | #145 | ok |
| MAXX-6854 ut-cascade-dual-write | (no PR exists) | #171 | wrong |
| MAXX-7031 documents-observability-ui | #168 | #168 | ok |
| MAXX-7035 citations-chain-of-custody | #171 | #168 | wrong |
| MAXX-7048 findings-extractor-skill-id | #169 | #169 | ok |

`store.json` contains **16 ticket links and 0 `pr` links**, confirming the PR
links are in-memory only and regenerated each cycle.

## Two distinct defects

1. **Wrong-PR attribution.** When exact `repoSlug#branch` match fails, the
   reconcile fallback attaches an incorrect PR instead of *no* PR. A session
   whose ticket has no PR should get **no** PR link — never a fuzzy "recent" one.
2. **No durable correction path.** Derived PR links aren't persisted, and there
   is no `crow remove-link`/unlink, so a mis-attached link can't be removed or
   overridden — it just regenerates.

## Proposed fix

- **Match on the embedded ticket key as a fallback.** When the exact
  `repoSlug#branch` lookup misses, extract the ticket key from the worktree
  branch (e.g. `maxx-7035`) and the candidate PR's head branch / title and match
  on that. This makes `feature/max-monorepo-maxx-7035-…` resolve to the PR on
  `feature/maxx-7035-…` (title `… MAXX-7035`).
- **Never attach on a fuzzy/recent guess.** If neither exact branch nor ticket
  key matches a real PR for that session's branch, attach nothing.
- Consider de-duping so a PR already linked to one session can't also be
  attached to another (would have caught #168 landing on both 7031 and 7035,
  and #171 on both 6854 and 7035).
- Once correct links derive, **persist** them so they survive restarts.

## Acceptance

- The 12 sessions above attach the PR in the "Correct PR" column (and MAXX-6854
  attaches **none**).
- No session is ever attached a PR that belongs to a different session/branch.
- Links survive an app restart.
- Regression test covering the `feature/{repo}-{ticket}-{slug}` worktree branch
  vs `feature/{ticket}-{slug}` PR head branch mismatch.

## Notes

- Likely related to the Jira-session PR-linking gap previously noted in #463.
- `max-monorepo` PR #170 is a duplicate of #169 (both MAXX-7048) — unrelated to
  this bug, just flagging for cleanup.

dhilgaertner:
Closes #520

## Problem

Reconcile attached the **wrong PR** to sessions whose worktree branch carries a repo-name prefix that the PR head branch drops — e.g. worktree `feature/max-monorepo-maxx-7035-citations-chain-of-custody` vs PR head `feature/maxx-7035-citations-chain-of-custody`. On the `max-monorepo`/MAXX sessions, 5 of 12 showed the wrong PR, including MAXX-6854 which has **no** PR yet was shown #171.

The key-matching + persistence machinery the ticket "proposes" already existed (added by #463/#464); derived links **are** persisted. The bug persisted because of three remaining gaps, all fixed here.

## Fix

1. **Body-match phantom PRs** (`GitHubCodeBackend.parseKeyPRMatches`) — a PR was matched when the ticket key appeared in its **body**, so a PR merely *mentioning* a related ticket (e.g. "related to MAXX-6854") attached to that unrelated session. Matching now requires the key in the PR **title or head branch** only.
2. **Branch-derived ticket key** — `Validation.ticketKey(fromBranch:)` extracts `MAXX-7035` from the worktree branch. Used as a fallback for task-only (Jira/Corveil) sessions, gated so a lowercased GitHub issue branch (`feature/acme-api-197-fix`) can't masquerade as a key.
3. **Cross-session de-dup** — `dedupeContestedPRs` drops a PR claimed by sessions with differing work-item identities from **all** of them. Never guess; a no-PR session gets none.

## Tests
- `ValidationTicketKeyTests` — branch-key extraction (incl. prefix-drop and keyless branches).
- `BackendsTests` — body-only match rejected; head-only match kept.
- `IssueTrackerReconcileTests` — the exact #520 max-monorepo scenario end-to-end (MAXX-7035 → #171, MAXX-6854 → none, no PR on two sessions) + focused de-dup cases.

All suites green: CrowCore 277, CrowProvider 52 + XCTest, root 234.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ Backfill of merged PRs since the 0.1.0 release, grouped by theme.
 - #220 — Filtering for the tickets list.
 - #226 — Per-section select all and icon-only cancel button in selection mode.
 - #231 — Quick action buttons on the session detail header.
+- #520 — PR-link reconcile no longer attaches the wrong PR when the worktree branch carries a repo-name prefix the PR head drops (`feature/max-monorepo-maxx-7035-…` vs `feature/maxx-7035-…`). The ticket key is derived from the worktree branch, key matching ignores body-only mentions, and a PR can attach to at most one session — a session whose ticket has no PR now gets none.
 
 ### Terminal Runtime
 
--- a/Packages/CrowCore/Sources/CrowCore/Validation.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Validation.swift
@@ -78,4 +78,18 @@ public enum Validation {
     public static func jiraKey(from spec: String) -> String? {
         parseJiraKey(spec)?.key
     }
+
+    /// Extract a ticket key (e.g. `MAXX-7035`) embedded in a git branch name such
+    /// as `feature/max-monorepo-maxx-7035-citations`. Returns the first
+    /// LETTERS-DIGITS token that validates as a Jira-style key, uppercased.
+    ///
+    /// The character class excludes `-`, so on `…-max-monorepo-maxx-7035-…` the
+    /// first complete LETTERS-DIGITS match is `maxx-7035` → `MAXX-7035`. Numeric
+    /// GitHub-issue branches like `feature/acme-api-197-fix` yield `api-197`,
+    /// which fails `parseJiraKey`'s uppercase-project rule → `nil`.
+    public static func ticketKey(fromBranch branch: String) -> String? {
+        guard let r = branch.range(of: "[A-Za-z][A-Za-z0-9]+-[0-9]+", options: .regularExpression)
+        else { return nil }
+        return parseJiraKey(String(branch[r]).uppercased())?.key
+    }
 }
--- a/Packages/CrowCore/Tests/CrowCoreTests/ValidationTicketKeyTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/ValidationTicketKeyTests.swift
@@ -1,0 +1,36 @@
+import Foundation
+import Testing
+@testable import CrowCore
+
+// MARK: - Validation.ticketKey(fromBranch:)
+//
+// Regression coverage for #520: worktree branches embed a repo-name prefix
+// (`feature/max-monorepo-maxx-7035-…`) that the PR head branch drops. Reconcile
+// derives the ticket key from the branch to recover the right PR.
+
+@Test func ticketKeyExtractsJiraKeyFromPrefixedBranch() {
+    #expect(
+        Validation.ticketKey(fromBranch: "feature/max-monorepo-maxx-7035-citations-chain-of-custody")
+            == "MAXX-7035"
+    )
+}
+
+@Test func ticketKeyHandlesBranchWithoutRepoPrefix() {
+    #expect(Validation.ticketKey(fromBranch: "feature/maxx-7035-citations") == "MAXX-7035")
+}
+
+@Test func ticketKeyExtractsAnyJiraShapedTokenAfterUppercasing() {
+    // A lowercased branch can't tell a real Jira project ("maxx") from an
+    // ordinary segment ("api"), so this pure extractor returns the first
+    // Jira-*shaped* token — `api-197` → `API-197`. Guarding against false keys
+    // on GitHub/GitLab issue branches is the caller's job (it gates the branch
+    // fallback to task-only trackers), not this function's.
+    #expect(Validation.ticketKey(fromBranch: "feature/acme-api-197-fix-tab-url-hash") == "API-197")
+}
+
+@Test func ticketKeyIsNilWhenNoJiraShapedToken() {
+    // No LETTERS-DIGITS token at all.
+    #expect(Validation.ticketKey(fromBranch: "feature/cleanup-readme") == nil)
+    #expect(Validation.ticketKey(fromBranch: "feature/197-only-number") == nil)
+    #expect(Validation.ticketKey(fromBranch: "") == nil)
+}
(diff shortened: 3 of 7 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface — all changes are internal logic / post-filtering on already-fetched data.
- `parseKeyPRMatches` post-filter tightens (not loosens) matching; rejects body-only mentions that previously cross-attributed.
- Branch-derived key extraction (`Validation.ticketKey(fromBranch:)`) validates through `parseJiraKey`'s `^[A-Z][A-Z0-9]+$` + numeric checks; output is `[A-Z0-9-]` only, safe to embed in `gh` argv (and is, via `KeyCandidate.key` → `--search "<key> in:title,body"`).
- `dedupeContestedPRs` is a conservative "drop on conflict" rather than "guess" — exactly the right call for a data-attribution bug.

**Concerns:** None.

### Code Quality

**Strengths:**
- Three discrete fixes, each minimal, each with focused test coverage.
- End-to-end regression test (`branchDerivedKeyResolvesCorrectPRAndNoPRSessionGetsNone`) exercises the precise #520 scenario (MAXX-7035 → #171, MAXX-6854 → none).
- The branch-fallback is correctly gated to task-only providers (Jira/Corveil) with a clear comment explaining why (`feature/acme-api-197-fix` would otherwise yield bogus `API-197`).
- `identityBySession` overwriting (key > branch) makes key the canonical identity when both are present — right precedence.
- Excellent docstrings on the new pure helpers (`ticketKey(fromBranch:)`, `dedupeContestedPRs`, `parseKeyPRMatches`) that explain *why*, not just *what*.

**Consider (green — non-blocking):**
- `parseKeyPRMatches` (GitHubCodeBackend.swift:245) uses `title.contains(needle) || head.contains(needle)` — substring, not word-bounded. For prefix-related ticket numbers (e.g., `MAXX-6859` vs `MAXX-68591`), a search for `MAXX-6859` could keep a PR whose only references are to `MAXX-68591`. The `dedupeContestedPRs` safety net catches the cross-attribution case (the contested PR is dropped from both), but the *legitimate* `MAXX-68591` session would lose its real link as collateral. Practical impact is low because gh's search tokenization typically word-bounds, and the prefix-sharing pattern is rare — a word-boundary regex (`(?<![A-Z0-9])<key>(?![0-9])`) would close it fully. Not a blocker.
- Legacy wrong-PR links from before this fix won't be proactively repaired (reconcile only attaches to sessions without a `.pr` link). Out of scope for this PR; worth noting for a follow-up cleanup if users report stale wrong links.

### Tests
- `ValidationTicketKeyTests` (4 tests) — branch-key extraction including prefix-drop, no-prefix, ambiguous-shape, and keyless cases. Verified passing locally.
- `BackendsTests.testGitHubCodeBackendFindPRsMatchingKeysParsesAndFilters` — covers all four buckets (title+head ✓, head only ✓, body only ✗, no match ✗). Verified passing.
- `IssueTrackerReconcileKeyScenarioTests` — end-to-end #520 reproduction. Asserts no PR attaches to two distinct sessions.
- `IssueTrackerReconcileDedupTests` (3 tests) — contested-drop, same-identity-keep, distinct-keep.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. The fix is precisely scoped to #520, defensively layered (three independent gates), and well-tested. The substring-matching consideration is theoretical and bounded by the dedupe safety net.

---

PR #540, opened 2026-06-30
Give Manager session cards the same activity indicators as worker cards (#539)

### LINKED ISSUE #539: Manager sessions: add same activity indicators / hook signals as Crow worker session cards
## Summary

Worker session cards in Crow show clear visual cues when something is happening — running state, last activity, hook-driven status pings, etc. **Manager sessions** (the long-running orchestration tab that runs Claude Code at `{devRoot}`) don't get any of that treatment. From the sidebar there's no way to tell whether the Manager is mid-tool-call, idle waiting on user input, or just sitting there.

## What we want

Apply the same activity-indicator + hook-status treatment to Manager session cards that worker session cards already get:

- A visible "running / waiting / idle" indicator on the Manager card while a turn is in progress (tool calls, long shells, fan-out across `setup.sh` invocations, etc.).
- Whatever hook signals (status line / activity pings / last-event timestamp) drive worker cards should fire for the Manager session too — the underlying `claude` process is the same, so the hook surface should already be available.
- Card should clear back to idle when the turn finishes, the same way worker cards do.

## Why

The Manager is where most multi-workspace orchestration happens (batched `/crow-workspace` runs, ticket creation, ADR work). Right now if you're tabbed away on a worker session and the Manager is grinding through a setup.sh or waiting on a `gh` permission prompt, there's no peripheral cue — you find out by tabbing back. Parity with worker cards fixes this without inventing a new affordance.

## Notes

- Should be implementable by treating the Manager session as just another session for the card-rendering + hook-routing path. The Manager UUID is the well-known `00000000-0000-0000-0000-000000000000` placeholder, which may currently be excluded from some of the indicator logic — check there first.
- Related: corveil#1651 covers the analogous gap in the Corveil chat UI (no running indicator during MCP tool windows). Same shape of bug, different surface.

dgershman:
Closes #539

## Problem

Worker session cards show activity cues (working / done / attention badge) driven by Claude Code hook events. **Manager** session cards got none of it — tabbed away, there was no peripheral signal that the Manager was mid-tool-call, waiting on a `gh` permission prompt, or idle.

The ingestion/routing/state layers (the `hook-event` RPC handler, `ClaudeHookSignalSource`, `SessionHookState`) are already generic by session UUID, so this is **purely additive** — route the Manager session through the same paths workers use. (Note: the real Manager UUID is the all-zeros `AppState.managerSessionID`; the `1EA19F17…` value in the ticket is just the session footer and appears nowhere in the repo.)

## Two gaps closed

**A — No hooks installed for the Manager.** `createManagerTerminal` wrote only the gateway env block, never the hook config, so `crow hook-event` never fired for the Manager and its `activityState` stayed `.idle`. Now it installs hook config too, mirroring the worker `launchAgent` path. Hooks are written **before** the gateway env so `writeGatewayEnv`'s `0600` re-apply (the env can carry a bearer token) is the final write; both merge into the same `{devRoot}/.claude/settings.local.json` without clobbering each other. The Manager has no worktree, so hooks carry the explicit `--session <managerID>` (cwd-fallback routing can't resolve it).

**B — Manager cards ignored `activityState`.** Both surfaces read only `pendingNotification`. Extracted `SessionRow`'s badge into a shared `AgentActivityBadge` (worker behavior byte-for-byte unchanged), then wired `activityState` into:
- `ManagerSessionRow` (additional managers) — leading working/attention dot, inline Working/Done badge, done-state tint.
- The primary "Manager" pill (`ManagerReviewsAllowListRow`) — leading working/attention dot (the pill is too narrow for inline text).

## Known limitation

A non-Claude Manager (Codex) still won't light up — its no-op hook writer relies on cwd→worktree resolution the Manager lacks. The default Manager is Claude, the ticket's target.

## Tests

`ManagerHookConfigTests` locks in the new, subtle invariant my ordering depends on: hook config and gateway env **coexist** in the same settings file, hook commands are bound to the Manager UUID, the `0600` restriction survives, and hooks aren't nuked when the Manager has no gateway. All 3 pass.

## Verification

- `swift build --arch arm64` — clean.
- `arch -arm64 swift test --arch arm64 --filter ManagerHookConfigTests` — 3/3 pass.
- Manual: on next Manager terminal creation, `{devRoot}/.claude/settings.local.json` gains a `hooks` block with `crow hook-event --session 00000000-… --event …` for all 17 events; the Manager card then shows green-pulse Working / Done / orange Permission|Question in lockstep with worker cards.

### DIFF
--- a/Packages/CrowUI/Sources/CrowUI/ReviewBoardView.swift
+++ b/Packages/CrowUI/Sources/CrowUI/ReviewBoardView.swift
@@ -374,14 +374,29 @@ public struct ManagerReviewsAllowListRow: View {
         appState.hookState(for: AppState.managerSessionID).pendingNotification != nil
     }
 
+    private var managerActivityState: AgentActivityState {
+        appState.hookState(for: AppState.managerSessionID).activityState
+    }
+
+    /// Leading status dot for the Manager pill (#539): orange pulse when the
+    /// Manager needs attention, else a green pulse while it's working, else
+    /// nothing — mirroring the worker card's `agentLaunched` indicator. The pill
+    /// is too narrow for inline state text, so the dot is the whole signal.
+    private var managerLeadingDot: AnyView? {
+        if managerNeedsAttention {
+            return AnyView(AttentionDot(color: Color.orange, accessibilityLabel: "Needs attention"))
+        } else if managerActivityState == .working {
+            return AnyView(AttentionDot(color: Color.green, accessibilityLabel: "Claude working"))
+        }
+        return nil
+    }
+
     private var managerButton: some View {
         sidebarButton(
             title: "Manager",
             isActive: appState.selectedSessionID == AppState.managerSessionID,
             needsAttention: managerNeedsAttention,
-            leading: managerNeedsAttention
-                ? AnyView(AttentionDot(color: Color.orange, accessibilityLabel: "Needs attention"))
-                : nil
+            leading: managerLeadingDot
         ) {
             appState.selectedSessionID = AppState.managerSessionID
         }
--- a/Packages/CrowUI/Sources/CrowUI/SessionListView.swift
+++ b/Packages/CrowUI/Sources/CrowUI/SessionListView.swift
@@ -459,10 +459,16 @@ struct ManagerSessionRow: View {
     private var needsAttention: Bool {
         appState.hookState(for: session.id).pendingNotification != nil
     }
+    private var activityState: AgentActivityState {
+        appState.hookState(for: session.id).activityState
+    }
 
     private var backgroundFill: Color {
         if needsAttention {
             return Color.orange.opacity(0.12)
+        }
+        if activityState == .done {
+            return CorveilTheme.bgDone
         }
         return isActive ? CorveilTheme.bgCard : CorveilTheme.bgSurface
     }
@@ -481,6 +487,8 @@ struct ManagerSessionRow: View {
             HStack(spacing: 6) {
                 if needsAttention {
                     AttentionDot(color: Color.orange, accessibilityLabel: "Needs attention")
+                } else if activityState == .working {
+                    AttentionDot(color: Color.green, accessibilityLabel: "Claude working")
                 }
                 Image(systemName: "person.2.fill")
                     .font(.system(size: 11))
@@ -502,6 +510,8 @@ struct ManagerSessionRow: View {
                 Spacer()
                 if isDeleting {
                     ProgressView().controlSize(.small)
+                } else {
+                    AgentActivityBadge(appState: appState, sessionID: session.id)
                 }
             }
             .foregroundStyle(isActive ? CorveilTheme.gold : CorveilTheme.goldDark)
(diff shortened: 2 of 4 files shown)


dhilgaertner (approved):
## Code & Security Review

Clean, purely-additive change that routes the Manager session through the same hook-driven activity paths workers already use. I traced both the ingestion side (`createManagerTerminal`) and the UI side (`AgentActivityBadge` extraction + Manager surfaces) against the existing worker code; the claims in the PR description hold up.

### Critical Issues
None.

### Security Review
**Strengths:**
- Hook/gateway write ordering is correct and verified against the writers: `writeHookConfig` (SessionService.swift:866) runs first, then `writeGatewayEnv` (SessionService.swift:879). Both use read-merge-write on the same `settings.local.json`, so neither clobbers the other (`ClaudeHookConfigWriter.swift:77-92`, `:115-134`), and the gateway write applies the `0o600` restriction last (`:147-148`) — so the bearer-token-bearing `env` block stays owner-only even with hooks present. `ManagerHookConfigTests` locks exactly this in.
- Hooks carry the explicit `--session <managerID>`, correct given the Manager has no worktree for cwd-fallback routing to resolve.
- Failure handling is defensive: `findCrowBinary`/agent-lookup are optional-bound and a write failure is logged, not fatal — the terminal still launches.

**Concerns:** None.

### Code Quality
- `AgentActivityBadge` extraction is behavior-preserving: the moved body is byte-for-byte identical to the old `SessionRow.activityStateBadge`, and `SessionRow` now delegates to it (SessionListView.swift:875). Precedence (attention → working → done) is intact and shared by both surfaces.
- Primary Manager is covered: it flows through `createManagerTerminal` via `ensureManagerSession` (SessionService.swift:693), so hooks install for the well-known all-zeros UUID, not just additional managers.
- Manager path mirrors the worker `launchAgent` path (SessionService.swift:537-561) faithfully.

**Consider (non-blocking):**
- Manager `backgroundFill` gates the done tint on `activityState == .done` alone (SessionListView.swift:470), whereas the worker also requires `terminalReadiness == .agentLaunched` (SessionListView.swift:887). Since `activityState` is persisted (`PersistedHookState`) and the Manager doesn't track readiness, a stale `.done` could briefly tint the Manager card on relaunch until `SessionStart` fires. This is consistent with the #367 "correct colors immediately on relaunch" intent and self-corrects in a beat, so it's fine as-is — just noting the deliberate divergence.
- `AttentionDot` is static; its pulse comes from the parent `.animation(value: needsAttention)`. The new green `working` dot therefore appears without a pulse (matching the worker's non-pulsing bolt badge), so the PR's "green-pulse" wording slightly overstates the visual — no code impact.

### Static analysis
Could not run `swift build` / `swift test` in the review environment — `Frameworks/GhosttyKit.xcframework` has no binary artifact checked out here (`error: local binary target 'GhosttyKit' … does not contain a binary artifact`). This is an environment limitation, not a defect in the PR; the author reports `swift build` clean and `ManagerHookConfigTests` 3/3 passing, and the tests are well-targeted at the coexistence/permission invariants.

### Summary Table

---

PR #542, opened 2026-06-30
Write Manager hook config on launch so its activity dot actually fires (#539)

dgershman:
Refs #539. Fixes the Manager activity dot never appearing after #540/#541.

## Root cause

#540 installs the Manager's hook config in `createManagerTerminal` — but that runs **only when the Manager has no terminal yet**. On a normal app launch the Manager terminal is *adopted* (re-attached to its live tmux window, `claude` left running per #374), so `createManagerTerminal` is skipped and the hooks are never written.

Confirmed empirically: an existing Manager's `{devRoot}/.claude/settings.local.json` has only `permissions` — **no `hooks` block**. So `crow hook-event` never fires for the Manager, `activityState` stays `.idle`, and the sidebar dot never lights up (even while the Manager is mid-tool-call).

## Fix

Write the hook config on the **hydrate path** too — right next to the existing `writeManagerGatewayEnv()` call that already runs every launch — using the Manager terminal's own `cwd` (so an additional Manager running outside the dev root is covered). Extracted `writeManagerHookConfig(for:dirPath:)` and pointed `createManagerTerminal` at it as well, so both paths share one writer.

Idempotent; ordered before the gateway-env write (whose `0o600` re-apply must stay the final write).

## Behavior note (important)

Claude Code reads hooks **at startup**, and the Manager's `claude` persists across app restarts (adopted, not relaunched). So:
- A **brand-new** Manager (the `+` button) launches `claude` with hooks present → dot works immediately.
- An **already-running** Manager must be **restarted once** (its `claude` relaunched) to pick up the newly-written hooks. After that it's permanent.

## Verification
- `swift build --arch arm64` clean; `ManagerHookConfigTests` pass.
- Manual: after this, `{devRoot}/.claude/settings.local.json` gains a `hooks` block with `crow hook-event --session <id> --event …` on launch; the Manager dot then shows green while working / amber when waiting, once its `claude` has been relaunched.

### DIFF
--- a/Sources/Crow/App/SessionService.swift
+++ b/Sources/Crow/App/SessionService.swift
@@ -147,6 +147,16 @@ final class SessionService {
                     }
                 }
                 let rebuiltCommand = managerCommand(for: reconciled)
+                // CROW-539: (re)write the Manager's hook config on every launch.
+                // The Manager terminal is adopted here, not recreated, so
+                // createManagerTerminal's hook write doesn't run — without this a
+                // Manager whose terminal predates the hooks never emits events and
+                // its card never lights up. Write to the terminal's own cwd so an
+                // additional Manager running outside the dev root is covered too.
+                // Before writeManagerGatewayEnv() so its 0o600 re-apply is last.
+                if let managerCwd = terminals.first?.cwd {
+                    writeManagerHookConfig(for: reconciled, dirPath: managerCwd)
+                }
                 writeManagerGatewayEnv()
                 // Remote-control bookkeeping reflects what the agent actually
                 // emitted — `supportsRemoteControl` is per-agent capability,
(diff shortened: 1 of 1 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No untrusted input reaches the new code path. `writeManagerHookConfig` writes to a local `.claude/settings.local.json` using a UUID session id and a `crowPath` resolved via `findCrowBinary()` — no shell interpolation, no external data.
- Error handling preserved: failures are caught and logged via `NSLog`, never swallowed silently or surfaced to the user.
- File permissions ordering is respected — the hook write runs before the gateway-env write, so `writeGatewayEnv`'s `0o600` re-apply (which can carry a bearer token) remains the final write to the shared file.

**Concerns:**
- None.

### Code Quality
- Clean DRY refactor: the inline hook-write block in `createManagerTerminal` is extracted into the shared `writeManagerHookConfig(for:dirPath:)` (`SessionService.swift:822`) and reused on both the terminal-creation and hydrate paths. The extracted body is functionally identical to the original (guard, do/catch, NSLog).
- The new hydrate-path call (`SessionService.swift:157`) is correctly guarded on `terminals.first?.cwd`; `terminals` is the Manager session's filtered terminal list, and a Manager has a single terminal, so `.first` is the right cwd. Nil-coalescing safely no-ops when there's nothing to write.
- Idempotent and well-documented; the doc comments explain why the adopt path (`#374`) skips `createManagerTerminal` and thus needs this.

### Minor (Green / consider)
- In the hydrate path the hook config writes to `terminals.first?.cwd` while `writeManagerGatewayEnv()` always targets `devRoot`. For the primary Manager these coincide, so the "0o600 re-apply stays last" ordering holds. For an *additional* Manager whose cwd differs from devRoot they'd be separate files, making that ordering note moot for that case — no bug or clobbering, just a comment nuance worth a future tidy.

### Summary Table

Note: I could not run `swift build` in the review environment (the checkout is missing the `GhosttyKit.xcframework` binary artifact — an environment limitation, not a code issue). Verdict is based on close code analysis plus the author's reported clean build and passing `ManagerHookConfigTests`.

---

PR #549, opened 2026-07-01
Keep OpenCode sessions alive in interactive TUI (#547)

### LINKED ISSUE #547: OpenCode agent exits to OS shell after running instead of staying alive as a REPL
The OpenCode coding-agent integration doesn't keep its session alive. After OpenCode finishes running, its managed terminal appears to fall through / dump out to the OS shell prompt rather than remaining an interactive OpenCode session.

Expected behavior (parity with Claude Code): the agent should run as a persistent REPL-style session that stays alive after a run completes, so Crow can keep driving it (send follow-up prompts via `crow send`, continue the work session, etc.) without the terminal dropping back to the bare OS shell.

Likely area: how Crow launches OpenCode in the managed terminal — the launch command probably runs OpenCode in a one-shot / non-interactive mode (process exits → terminal returns to `$SHELL`) instead of an interactive/REPL mode the way the Claude Code launch does. Compare the OpenCode terminal launch + command construction against the Claude Code path and make OpenCode stay resident.

Acceptance:
- Launching OpenCode as the agent yields a persistent interactive session that does not exit to the OS shell after a run.
- Crow can send follow-up input to the same OpenCode session after its first run completes.
- Behavior matches the Claude Code session lifecycle.

Related: #545 (OpenCode agent integration / Claude Code parity).

dgershman:
## Summary

OpenCode managed terminals were launched with headless `opencode run`, which drives the agent to completion and exits back to the OS shell. This switches job/review (and crow-workspace skill) launch to a run-then-continue form:

- **First launch:** `opencode run "$(cat .crow-*-prompt.md)"[auto-approve]; opencode --continue` — headless `run` consumes the prompt, then `; --continue` opens the interactive TUI with a fresh terminal stdin for `crow send`
- **Subsequent relaunch:** `opencode --continue` (with `--auto` for auto-permission jobs when supported)

Closes #547

## Test plan

- [x] `OpenCodeAgent` / `OpenCodeLaunchArgs` unit tests pass (30 tests)
- [ ] Launch an OpenCode work session — TUI stays open after first run
- [ ] Launch an OpenCode job session — initial prompt runs, TUI remains for follow-up `crow send`
- [ ] Restart Crow with an existing OpenCode session — relaunch uses `--continue`


dhilgaertner (changes requested):
## Code & Security Review

Clean, well-scoped change: swaps OpenCode's headless `opencode run "<prompt>"` dispatch for the interactive TUI form (`opencode --prompt` on first launch, `opencode --continue` on relaunch) and changes the auto-approve flag from `--dangerously-skip-permissions` to `--auto`. The rewrite is applied consistently across every emitting site and the docs.

### Critical Issues
None.

### Security Review
**Strengths:**
- `--auto` (auto-approve) is correctly gated to `.job` + `autoPermissionMode` only; `.review` deliberately leaves it off so the reviewer still gates writes (`OpenCodeAgent.swift:104`). Good — a review agent should not silently auto-approve edits.
- Prompt temp file in `OpenCodeLauncher` keeps its `0o600` perms and `shellEscape` on the worktree/prompt paths (`OpenCodeLauncher.swift:83`). No regression there.
- `"$(cat …)"` command substitution stays inside double quotes, so file contents aren't re-expanded by the shell — same safe pattern as before.

**Concerns:**
- None new. (Note: `autoLaunchCommand`'s `--prompt \"$(cat \(promptPath))\"` does not shell-escape `promptPath`, but that is pre-existing and unchanged by this PR.)

### Code Quality
- I verified `--prompt`, `--continue`, and `--auto` against OpenCode's official CLI docs — all three exist on the root TUI command with the behavior claimed in the comments. Nice that the flags are real and the misleading \"confirm empirically whether a seed-the-TUI variant exists\" caveats were removed.
- All four command-emitting sites (`OpenCodeAgent`, `OpenCodeLauncher`, `Resources/crow-workspace-setup.sh.template`, `skills/crow-workspace/setup.sh`) and `docs/configuration.md` are updated identically. No stray `opencode run` left in any emitted command. 
- `swift test` for `CrowOpenCode` passes (19 tests). Caveat: these tests only assert substrings of the built command string, so they are self-fulfilling — they'd pass even if the underlying flag were wrong. That's inherent to string-building code, just worth knowing they don't validate the feature end-to-end.

### Findings

**Yellow — the load-bearing runtime assumption is unverified (`OpenCodeAgent.swift:84`, `OpenCodeLauncher.swift:75`).** The entire premise of #547 is that `opencode --prompt \"…\"` *submits* the seeded message and *stays resident* in the TUI (unlike `run`, which exits). The comment states this as fact (\"`--prompt` seeds and submits the first message\"), but the official docs only describe it as \"a prompt to use when launching the TUI\" — they don't explicitly confirm auto-submit, and the PR's own test plan boxes for runtime behavior are still unchecked. Failure scenario: if `--prompt` merely pre-fills the input box without submitting, every unattended `.job`/`.review` dispatch would launch, pre-fill, and silently stall — the agent never starts working until a human presses Enter. Please complete the three unchecked manual-verification boxes (first launch stays open, job initial prompt runs, restart resumes via `--continue`) before merge, and confirm `--auto` is accepted alongside `--prompt`.

### Consider (Green)
- On a subsequent `.job` relaunch (`reviewPromptDispatched == true`), the command is `opencode --continue` with no `--auto`, so an auto-permission job that gets resumed after a Crow restart would re-gate permissions. This mirrors the prior bare-`opencode` behavior so it's not a regression, but worth deciding whether resumed auto-jobs should carry `--auto` through.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 1 Green] findings. The code is correct against the documented flags; the single Yellow is a pre-merge verification gap on the feature's core runtime assumption, which the PR's own test plan already flags as incomplete.

---

dgershman:
Addressed the Yellow finding in 488149b:

**Runtime verification (local opencode 1.17.10 via Homebrew):**
- TUI exposes `--prompt` and `--continue` in `opencode --help`
- TUI does **not** advertise `--auto` on this build (only `opencode run` has `--dangerously-skip-permissions`)
- Crow now probes `opencode --help` at launch time via `OpenCodeLaunchArgs.tuiSupportsAuto()` and only appends `--auto` when the flag is present — avoids passing unsupported flags on older installs

**Load-bearing assumption (`--prompt` auto-submit):**
- Switched first-launch dispatch from `opencode --prompt "$(cat …)"` to **`cat … | opencode`**, which OpenCode's CLI docs document as the supported way to deliver an initial message (avoids builds where `--prompt` only pre-fills the composer without submitting — sst/opencode#3937)
- Applied consistently across `OpenCodeAgent`, `OpenCodeLauncher`, and `crow-workspace` setup scripts

**Green consideration (resumed auto-jobs):**
- Subsequent `.job` relaunches now carry `--auto` when `jobsAutoPermissionMode` is enabled and the installed build supports it (`opencode --continue --auto`)

**Tests:**
- Added `OpenCodeLaunchArgsTests` (7 tests) covering help parsing, pipe form, and `--auto` gating
- Updated `OpenCodeAgentTests` to assert pipe form and capability-aware `--auto` behavior
- All 27 `CrowOpenCode` tests pass

Full interactive TUI verification (stay-open after first run, `crow send` follow-up) still requires a manual pass in Crow's embedded terminal — not reproducible in this headless CI/agent environment.

dhilgaertner (changes requested):
## Code & Security Review

Clean, well-documented change that centralizes launch-command construction in `OpenCodeLaunchArgs`, mirrors the `CursorAgent` dispatch shape, and adds solid unit coverage. Build + all 27 `CrowOpenCode` tests pass locally. No security issues (no new shell interpolation of untrusted input beyond the existing cross-agent convention; `--auto` is capability-gated behind a `--help` probe).

The verdict below is driven by one **Yellow**: the core behavioral hypothesis of this PR is unverified, and the author's own test plan reflects that (three of four boxes unchecked).

### Security Review
**Strengths:**
- `--auto` (auto-approve) is only emitted when both `autoPermissionMode` is on *and* the installed build actually advertises the flag (`OpenCodeLaunchArgs.parseTUISupportsAuto`), avoiding an unknown-flag launch failure. Review sessions never get `--auto` (`OpenCodeAgent.swift:103,111`).
- The `permission.ask` plugin hook remains observe-only — no `output.status` mutation — so auto/manual permission decisions still stand.

**Concerns:** none blocking.

### Code Quality

**Yellow — piping into the *bare* interactive TUI is an unverified assumption (`OpenCodeLaunchArgs.swift:50-61`, `OpenCodeAgent.swift:96-109`).**
The new form is `cat <prompt> | opencode` (no `run` subcommand). Two runtime unknowns make this the crux of the PR and neither is verified:
1. **Does bare `opencode` consume piped stdin as the initial message?** OpenCode's documented stdin-piping delivers the message to the *headless* `opencode run` path; the bare TUI may not read piped stdin as a submitted message at all (the same reason `--prompt` was rejected — it only pre-fills). If so, the initial job/review prompt is silently dropped and the session never starts working unattended — a regression from the old `opencode run "<prompt>"`, which definitely consumed it.
2. **Does the TUI stay keyboard-interactive after the pipe?** With fd 0 bound to a pipe (EOF after `cat`), a TUI that reads input from stdin rather than `/dev/tty` won't receive `crow send` keystrokes — which would defeat the entire "session stays resident for follow-up input" goal of #547.

The PR test plan leaves exactly these three scenarios unchecked. Please confirm empirically (TUI stays open *and* accepts a follow-up `crow send` after the first run for both job and review) before merge, or note the evidence in the PR. Requesting changes so this lands in the same round trip rather than as a post-merge surprise.

### Consider (Green)
- `OpenCodeLaunchArgs.tuiSupportsAuto` runs `opencode --help` via a synchronous `Process.waitUntilExit()` (`OpenCodeLaunchArgs.swift:76-88`) from `autoLaunchCommand`, which is on the launch (main-actor) path. It's cached after the first call, but a slow/hanging binary would stall the first launch. A non-blocking or pre-warmed probe would be safer.
- The `--auto` probe cache (`autoFlagCache`) is keyed by binary path and never invalidated, so an in-place `opencode` upgrade during an app session keeps the stale answer. Acceptable for now; worth a comment.
- `cat <promptPath> | opencode` does not shell-escape `promptPath`/`worktreePath` in the used `autoLaunchCommand` path (`OpenCodeAgent.swift:101-108`), so a worktree path containing spaces breaks the command. This matches the existing `CursorAgent` convention (same latent gap), so not a regression — but if paths with spaces are ever in scope, both agents need the fix.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 3 Green] findings. The Yellow is a verification gap on the PR's central behavior, not a code defect; completing the unchecked test-plan items (or documenting the runtime evidence) should clear it.

---

dgershman:
7880942 addresses the second review's Yellow finding on stdin piping:

**Problem with `cat … | opencode`:** piping binds fd 0 to a pipe that closes after `cat` exits, which can prevent the TUI from accepting keyboard/`crow send` input. Bare `opencode` may also ignore piped stdin as a submitted message (that path is documented for headless `run`).

**New first-launch form:**
```bash
opencode run "$(cat .crow-*-prompt.md)" [--auto|--dangerously-skip-permissions] && opencode --continue [--auto]
```

- **`run`** — proven headless path that consumes the initial prompt unattended
- **`&& --continue`** — drops into the interactive TUI with a fresh terminal stdin after `run` exits, instead of falling back to `$SHELL` (#547)

Also in this commit:
- `OpenCodeLaunchArgs.runHelpText()` probes `opencode run --help` for auto-approve flags (`--auto` or `--dangerously-skip-permissions`)
- Help probe timeout (5s) to avoid stalling first launch on a hung binary
- 29/29 `CrowOpenCode` tests pass

Interactive TUI + `crow send` follow-up still needs a manual pass in Crow's embedded terminal — the chained command is structured specifically so the `--continue` phase owns the tty.

dhilgaertner (changes requested):
## Code & Security Review

Solid, well-documented change. The run-then-continue dispatch form is a sensible way to keep the OpenCode TUI resident for `crow send`, the capability probing is centralized in a testable `OpenCodeLaunchArgs`, and the new unit tests cover the flag-suffix matrix nicely. All 29 `CrowOpenCode` tests pass locally. No security concerns (prompt files stay `0o600`, no secrets, no injection surface beyond the shell-quoting note below).

One issue should be addressed before merge.

### Security Review
**Strengths:**
- Temp prompt file is written `atomically` with `0o600` permissions (`OpenCodeLauncher.swift:80-82`).
- No credentials or secrets introduced; OpenCode reads provider creds from its own config.
- Capability probes shell out to the resolved binary with fixed `--help` args only — no user input reaches the probe argv.

**Concerns:**
- None blocking.

### Code Quality

- 🟡 **Synchronous `--help` subprocess probes now run unconditionally on the `@MainActor`, even when their results are discarded.** `OpenCodeAgent.autoLaunchCommand` (`OpenCodeAgent.swift:94-95`) calls `OpenCodeLaunchArgs.tuiSupportsAuto(...)` and `runHelpText(...)` for *every* `.job`/`.review` launch, before the dispatch branch. Each spawns an `opencode`/`opencode run --help` process and blocks — with the `NSLock` held across the subprocess (`OpenCodeLaunchArgs.swift:36-42, 46-53`) up to `helpProbeTimeoutSeconds` (5s) each, i.e. ~10s worst case — and `launchAgent`/`autoLaunchCommand` execute on the `@MainActor` (`SessionService.swift:11-12, 532, 587`). This can freeze the UI on the first OpenCode launch. It's also pure waste for `.review` sessions: `autoPermissionMode` is only ever true for `.job` (`SessionService.swift:579`), so the probed values feed suffix helpers that are already gated on `autoPermissionMode` and always resolve to `""` for reviews. The prior code computed the auto-approve flag with a plain ternary and spawned nothing — this PR introduces the main-thread subprocess. Suggest gating the probes behind `(session.kind == .job) && autoPermissionMode` (compute them only when an auto suffix is actually wanted), or moving them off the main actor.

- 🟢 **`&&` between `run` and `--continue` drops the terminal to a dead shell when the headless run fails.** `firstLaunchChainedCommand` (`OpenCodeLaunchArgs.swift:93-94`) chains `opencode run … && opencode --continue`. If the initial `run` exits non-zero (agent error, bad prompt, interrupted), `--continue` never fires and the TUI never opens — which defeats the PR's goal of keeping the session resident for `crow send`. Since `run` still creates a resumable session even on partial failure, `;` (or a `|| opencode --continue`) would open the TUI regardless. Worth considering; `&&` may be intentional if a failed run should surface as a shell prompt.

- 🟢 **Inconsistent shell-quoting of `promptPath` between the two callers of `firstLaunchChainedCommand`.** `OpenCodeLauncher.launchCommand` shell-escapes the path (`OpenCodeLauncher.swift:86`), but `OpenCodeAgent.autoLaunchCommand` passes the raw `worktreePath`-derived path (`OpenCodeAgent.swift:100-104`), which lands unquoted inside `cat \(promptPath)`. This is pre-existing behavior and worktree paths are kebab-slugs in practice, but the new shared helper has an ambiguous contract (one caller pre-escapes, one doesn't; `binary` is never escaped). Consider escaping inside the helper so both call sites are safe by construction.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The Yellow (main-thread subprocess probes, wasted for reviews) is a small, contained fix; the Green items are optional but cheap to land in the same round trip.

---

dgershman:
bf985e4 addresses the third review:

**Yellow — main-thread `--help` probes:**
- Probes now run only when `(session.kind == .job) && autoPermissionMode` — review sessions and non-auto jobs build commands without spawning subprocesses
- Cache lock is released before `Process.waitUntilExit()` so the lock isn't held across subprocess I/O
- `OpenCodeLauncher.launchCommand` skips probes entirely (never uses auto-approve)

**Green — `&&` vs `;`:**
- Changed to `opencode run …; opencode --continue` so the TUI opens even when headless `run` exits non-zero, preserving a resumable session for `crow send`

**Green — shell-quoting:**
- `OpenCodeLaunchArgs.shellQuote()` now escapes `promptPath` inside `firstLaunchChainedCommand`; both call sites pass raw paths

30/30 `CrowOpenCode` tests pass.

dhilgaertner (changes requested):
## Code & Security Review

Solid, well-documented change. The run-then-continue dispatch is centralized in `OpenCodeLaunchArgs`, the semicolon-vs-`&&` rationale is explained where it matters, capability probes are cached and gated to only auto-permission jobs, and the new `OpenCodeLaunchArgsTests` plus updated `OpenCodeAgentTests` cover the launch/resume/auto-flag matrix. All 30 package tests pass locally (`swift test`).

### Security Review
**Strengths:**
- Prompt content reaches OpenCode via `"$(cat 'path')"` — command substitution output inside double quotes is used as a single literal word and is **not** re-scanned for further expansion, so prompt bodies containing `$`, backticks, etc. can't inject shell. The prompt *path* is POSIX single-quote escaped (`shellQuote`), and tests assert no stray `|` pipe survives.
- Prompt temp file in `OpenCodeLauncher` is written `0o600`.
- Help probes have a 5s timeout + `terminate()`, read-then-wait ordering avoids pipe deadlock, and non-zero exit yields `""` (never a partial parse).
- The generated JS plugin escapes `crowPath` defensively and is fire-and-forget.

No injection, auth, or secret-exposure concerns introduced.

### Code Quality

**Yellow (should fix) — shell precedence drops `cd` failure protection**
`Resources/crow-workspace-setup.sh.template:1167` and `skills/crow-workspace/setup.sh:1185`:
```sh
cd $WORKTREE_PATH && $bin run "$(cat $prompt_path)"; $bin --continue
```
`;` binds looser than `&&`, so this parses as `(cd … && run …) ; --continue`. If `cd` fails, `run` is correctly skipped — but `opencode --continue` **still executes**, now in the terminal's original cwd, resuming whatever the last global OpenCode session was in the wrong directory. The prior `cd … && run` form ran nothing on `cd` failure. Group the pair so both are gated on the `cd`:
```sh
cd $WORKTREE_PATH && { $bin run "$(cat $prompt_path)"; $bin --continue; }
```
(The `;` between `run` and `--continue` is still correct there — you *want* the TUI even if `run` exits non-zero.) The same latent pattern is in `OpenCodeLauncher.launchCommand` (`OpenCodeLauncher.swift:84`); that path isn't wired into auto-launch yet, but worth fixing together. Note the Swift auto-launch path via `firstLaunchChainedCommand` is unaffected — it emits no `cd` (the terminal is created with the worktree cwd).

**Green (consider)**
- Comment/code mismatch: both shell files say \`then \`&& --continue\`\` in the comment but the code uses \`;\`. The Swift side documents the deliberate \`;\` choice; align the shell comments so they don't contradict it.
- PR description still says first launch is `opencode --prompt "$(cat …)"`, but the shipped code uses `opencode run "$(cat …)"`. Worth updating the description for accuracy.
- On first launch of an auto-permission `.job`, `autoLaunchCommand` runs up to two synchronous `--help` probes (≤5s each) on the calling thread. Cached and job+auto-only, so acceptable — noting for awareness.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 3 Green] findings. The Yellow is a one-line grouping fix in both shell scripts; fixing it (and optionally the Green nits) in this round trip avoids a follow-up.

---

dgershman:
355bf57 addresses the fourth review Yellow finding:

**Shell precedence / `cd` failure:** Both setup scripts and `OpenCodeLauncher` now use brace grouping so `run` and `--continue` are gated on a successful `cd`:

```sh
cd $WORKTREE_PATH && { $bin run "$(cat $prompt_path)"; $bin --continue; }
```

Also aligned shell comments to document `; --continue` (not `&&`) and updated the PR description to reflect the shipped `run`-then-`continue` form.

dhilgaertner (approved):
## Code & Security Review

Fifth-round review. The dispatch design has stabilized on the run-then-continue form, and every concrete finding from the prior four rounds is resolved in the current tree (355bf57). Build compiles; all 30 `CrowOpenCode` tests pass locally (`swift test`).

### Prior findings — all resolved
- **Main-thread `--help` probes (was Yellow):** now gated behind `autoForJob = (session.kind == .job) && autoPermissionMode` (`OpenCodeAgent.swift:94-100`). Review sessions and non-auto jobs spawn no subprocess; the cache lock is released before `Process.waitUntilExit()`, and the probe carries a 5s timeout with read-then-wait ordering (no pipe deadlock).
- **`&&` vs `;` between `run` and `--continue` (was Green):** `firstLaunchChainedCommand` uses `;` so the TUI opens even when headless `run` exits non-zero, preserving a resumable session for `crow send` (`OpenCodeLaunchArgs.swift`).
- **Shell-quoting (was Green):** `shellQuote` is applied to `promptPath` inside the shared helper, so both call sites are safe by construction.
- **`cd` failure protection (was Yellow):** both `Resources/crow-workspace-setup.sh.template:1168` and `skills/crow-workspace/setup.sh:1186` now use a brace group `cd … && { $bin run …; $bin --continue; }`, and `OpenCodeLauncher.launchCommand:91` matches.
- **Comment/description drift (was Green):** shell comments document `; --continue`; docs/configuration.md and the PR description reflect the shipped form.

### Security Review
**Strengths:**
- Prompt content reaches OpenCode via `"$(cat 'path')"` — command-substitution output inside double quotes is used as a single literal word and is not re-scanned for expansion, so prompt bodies with `$`/backticks can't inject shell. The prompt *path* is POSIX single-quote escaped (`shellQuote`).
- `--auto` is emitted only when `autoPermissionMode` is set *and* the installed build advertises the flag (`parseTUISupportsAuto`); `.review` never auto-approves (`OpenCodeAgent.swift:94`).
- Temp prompt file in `OpenCodeLauncher` stays `0o600`; help probes pass only fixed `--help` args (no user input in argv).

**Concerns:** None.

### Code Quality
- Launch-command construction is centralized in a testable `OpenCodeLaunchArgs`; `OpenCodeLaunchArgsTests` covers the help-parse / auto-flag / shell-quote matrix and `OpenCodeAgentTests` asserts the run-then-continue shape (`; `, `--continue`, no `&&`, no `|`).
- The probe cache is `NSLock`-guarded; the check-then-compute race only ever produces a redundant (idempotent) probe, never a wrong answer.

### Consider (Green)
- The manual test-plan boxes (TUI stays open + `crow send` follow-up in Crow's embedded terminal) remain unchecked because they aren't reproducible in the headless agent environment. The author has recorded the closest local verification (opencode 1.17.10: TUI exposes `--continue`, no `--auto` on that build; `run` consumes the prompt). Worth capturing one manual pass in the PR when convenient, but the structural design no longer carries the silent-prompt-drop risk the earlier `--prompt`/pipe forms did.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 1 Green] findings. Every actionable code-level finding from the prior four rounds is addressed, the design is robust, and the only remaining item is an environment-bound manual verification the author has already done as far as CI allows.

---

PR #550, opened 2026-07-01
handling cases where cursor and agents can't find crow cli

dgershman:
## Summary

Brief description of what this PR does and why.

## Related Issue

Closes #

## Changes

-
-

## Testing

- [ ] `swift test` passes
- [ ] Manually tested the affected workflow
- [ ] Screenshots attached (for UI changes)

## Checklist

- [ ] Code follows existing patterns in the codebase
- [ ] No unrelated changes included


dhilgaertner (changes requested):
## Code & Security Review

Solid, well-scoped change. It brings `setup.sh` binary discovery to parity with `CodingAgent.findBinary` / `ShellEnvironment` (config override → symlink farm → login-shell PATH → dev builds → known locations → `command -v`), threads a resolved `$CROW_BIN` through every IPC call so the script works when launched outside a Crow terminal (Cursor/Codex inherit a minimal PATH), adds a matching login-shell PATH walk to Swift `findCrowBinary`, and keeps `Resources/crow-workspace-setup.sh.template` byte-identical to `skills/crow-workspace/setup.sh` (verified). The 7-case `setup_binary_test.sh` passes locally, and the `main()` source-guard keeps sourcing side-effect-free.

### Security Review
**Strengths:**
- No injection surface introduced: all resolved binaries are invoked as array-expanded argv (`"$CROW_BIN" ...`), never via `eval`/`sh -c`. Config-supplied paths are gated behind an `-x` executable check before use.
- `binary_from_config` parses `config.json` with `jq -r` (no shell interpolation of JSON values) and degrades gracefully on missing/invalid config.
- Login-shell probe redirects stderr to `/dev/null` and tolerates failure (`|| true`), mirroring the hardened Swift path.

### Code Quality

**Yellow — should fix**
- `skills/crow-workspace/setup_binary_test.sh:82` (the `resolve_crow_binary (dev build)` case): the test resets `LOGIN_PATH=""` and `LOGIN_PATH_RESOLVED=false` before calling `resolve_crow_binary`, which forces `resolve_login_path` to spawn the *real* login shell and walk the host's actual PATH. Because the login-PATH step (order #3) runs before the dev-build step (#4) in `resolve_crow_binary`, on any machine where `crow` is installed on PATH — the normal state once Crow.app is installed and has symlinked its CLI into `~/.local/bin` / `/opt/homebrew/bin` / `/usr/local/bin` — this check resolves to the installed binary instead of the temp `.build/debug/crow` and the assertion fails. Confirmed by simulation under bash: with a `crow` seeded on `LOGIN_PATH`, `CROW_BIN` comes back as the login-PATH hit, not the dev build. The suite only passes here because this review host has no `crow` on PATH. To isolate it deterministically, skip the live probe — set `LOGIN_PATH_RESOLVED=true` with a crow-free `LOGIN_PATH=""` (which makes the `[[ -n "$LOGIN_PATH" ]]` guard fall through to the dev-build branch) rather than forcing re-resolution.

**Green — consider**
- `resolve_login_path` spawns `"$shell" -lc 'echo $PATH'` with no timeout. The Swift `resolveFromLoginShell` it mirrors deliberately caps this at 5s to avoid hanging on slow shell init (nvm/conda/network homes). Since this now runs in `preflight` before anything is created, a pathological rc could stall workspace setup. `timeout(1)` isn't stock on macOS, so a pure-bash guard (background + `wait`/poll) or a documented note would close the parity gap.
- `binary_from_config` silently returns empty when `jq` is absent, so a user who pins `defaults.binaries.crow` on a jq-less machine gets a silent fallback to PATH discovery rather than the pinned path. A one-line warning when the config file exists but `jq` is missing would make the misconfiguration visible.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The Yellow is a test-isolation flake that will fail on maintainer machines with `crow` on PATH; the Green items are optional hardening.

---

dhilgaertner (approved):
## Code & Security Review

Scope: `setup.sh` + `crow-workspace-setup.sh.template` gain a full binary-discovery layer (login-shell PATH probe, `defaults.binaries.*` config override, `.claude/bin` symlink farm, dev-build siblings) so `/crow-workspace` works when launched from Cursor/Codex/OpenCode with a minimal PATH; `ClaudeHookConfigWriter.findCrowBinary()` now walks the resolved login PATH via `ShellEnvironment`; new `setup_binary_test.sh` unit-tests the helpers; docs updated.

### Critical Issues
None.

### Security Review
**Strengths:**
- No new privilege boundary crossed — `binary_from_config` reads paths from the user's own `{devRoot}/.claude/config.json`, and the login-shell probe (`$SHELL -lc 'echo $PATH'`) sources the user's own rc, mirroring what `ShellEnvironment` in `Packages/CrowCore` already does.
- `resolve_login_path` bounds the login-shell probe to ~5s and falls back to a well-known-dirs PATH on timeout/empty output, so a hanging rc file can't wedge setup.
- Config path is only consulted through `jq` with a clean `// empty` default and whitespace-trimming; a missing `jq` degrades to a stderr warning rather than mis-parsing.

**Concerns:** None.

### Code Quality
Verified during review:
- `log()` writes to stderr (`>&2`), so the jq-missing warning and timeout notice don't leak into the `bin=$(resolve_binary …)` / `$(binary_from_config …)` command substitutions. ✅
- The backgrounded login-shell probe redirects its stdout/stderr to a temp file, so it doesn't hold the command-substitution pipe open (no hang), and bash reaps it so `kill -0` correctly detects completion (verified empirically). ✅
- All IPC call sites now use `"$CROW_BIN"`; remaining bare `crow` tokens are comments only. ✅
- `resolve_binary`'s signature change (added `config_key`) is applied at every call site; the old duplicate definition was removed cleanly. ✅
- Sourcing guard (`[[ "${BASH_SOURCE[0]}" == "${0}" ]]`) means `setup_binary_test.sh` can `source setup.sh` without triggering `main`. ✅
- `Resources/crow-workspace-setup.sh.template` is byte-identical to `skills/crow-workspace/setup.sh`. ✅
- Swift change: `import CrowCore` present, `CrowCore` is a package dependency, and `ShellEnvironment.shared.findExecutable` is `public` — compiles cleanly and matches `CodingAgent.findBinary` ordering.
- `setup_binary_test.sh`: all 7 checks pass locally.

Consider (non-blocking, green):
- `findCrowBinary()` (Swift) resolves crow via login-PATH but does not consult `BinaryOverrides`/`defaults.binaries.crow`, whereas `setup.sh` does. The updated docs scope the `crow` override to `setup.sh`, so this is consistent — noting only in case you later want the hook writer to honor a pinned crow too.
- `findCrowBinary()` introduces an `fm` alias but the `candidates` array still uses `FileManager.default` — purely cosmetic.
- `find_dev_crow_binary`'s `$HOME/Projects/crow` fallback is a best-effort hardcoded guess; the `DEV_ROOT` workspace scan is the real path and covers the common case.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 3 Green] findings. Discovery logic faithfully mirrors the existing Swift path resolution, is stderr-safe inside command substitution, is covered by a passing test, and keeps the template in sync.

---

PR #553, opened 2026-07-01
Install crow CLI onto PATH via Scaffolder symlink

### LINKED ISSUE #552: Install the crow CLI onto PATH for all agents (Scaffolder symlink into .claude/bin)
Crow never installs the `crow` CLI onto any agent's PATH, so agents can't find `crow`. Surfaced by the OpenCode integration (its state hook is a JS plugin that shells out to `crow hook-event`), but it affects every agent — Claude only appeared to work because of a manual `.build/debug` PATH prepend, not any agent-code difference.

## Root cause
- The `crow` binary lives only at `.build/debug/crow` (dev) / inside the app bundle (release) — on no standard PATH dir. `which crow` returns not found in a fresh agent shell.
- Both `ClaudeCodeAgent` and `OpenCodeAgent` resolve their env via `ShellEnvironment` / the login-shell PATH and inject nothing crow-specific. Neither installs the CLI.
- `Scaffolder` already has the right mechanism: it symlinks `{devRoot}/.claude/bin/<name>` for every `defaults.binaries.<name>`, and `{devRoot}/.claude/bin` is first on PATH. But `defaults.binaries` only lists `corveil`, so **no `.claude/bin/crow` symlink is ever created.**

## Fix
Have `Scaffolder` **always** materialize a `{devRoot}/.claude/bin/crow` symlink pointing at the running app's own `crow` binary (bundled CLI in release; `.build/debug/crow` in dev), independent of `defaults.binaries`. Reuse the existing symlink-management path in `Scaffolder` (create dir, reap stale links, only ever own symlinks — see `ScaffolderBinarySymlinkTests`). This is agent-agnostic and fixes discovery for Claude, Codex, Cursor, and OpenCode at once.

Consider also ensuring the OpenCode hook plugin's process env includes `.claude/bin` (its plugin may run with a more minimal env than the terminal), so `crow hook-event` resolves there too.

## Acceptance
- After scaffolding, `{devRoot}/.claude/bin/crow` exists as a symlink to the app's crow binary and resolves via PATH in a fresh agent shell (`command -v crow` succeeds) for all agent kinds.
- OpenCode sessions (TUI + the `crow hook-event` plugin) can invoke `crow` without any manual PATH tweaks.
- No dependence on a manual `.build/debug` PATH prepend, and no need for the interim `~/.local/bin/crow` symlink.
- Symlink is reaped/refreshed correctly on config or app changes (never clobbers a real file).

Related: #545 (OpenCode integration / Claude parity), #547 (OpenCode REPL persistence).

dgershman:
Closes #552

## Summary
- Scaffolder always materializes `{devRoot}/.claude/bin/crow` as a symlink to the running app's crow binary (dev `.build/{config}/crow` or release bundle `Contents/MacOS/crow`), independent of `defaults.binaries`.
- The crow symlink is managed separately from config-driven binary overrides — it survives re-scaffold passes, repoints when the app binary changes, and never clobbers a real file dropped by hand.
- `findCrowBinary(devRoot:)` prefers the scaffolded symlink so hook configs (Claude, Codex, Cursor, OpenCode) invoke crow through the stable PATH entry.

## Test plan
- [x] Added unit tests for crow symlink creation, persistence across empty overrides, repointing, and non-symlink safety
- [ ] Launch Crow, confirm `{devRoot}/.claude/bin/crow` exists and `command -v crow` succeeds in a fresh agent terminal
- [ ] OpenCode TUI + `crow hook-event` plugin resolves crow without manual PATH tweaks

### DIFF
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
@@ -196,14 +196,16 @@ public struct ClaudeHookConfigWriter: HookConfigWriter {
 
     // MARK: - Find crow Binary
 
-    /// Find the crow binary, checking common install locations.
-    public static func findCrowBinary() -> String? {
-        // Check same directory as running executable first (development builds)
+    /// Resolve the running app's own `crow` CLI — the bundled binary in a
+    /// release `.app` (`Contents/MacOS/crow`), or `.build/{config}/crow` in
+    /// dev. Does not consult `{devRoot}/.claude/bin/crow`; use that for
+    /// agent-facing resolution via `findCrowBinary(devRoot:)`.
+    public static func appCrowBinary() -> String? {
+        // Same directory as the running executable (dev + release bundles).
         let execURL = URL(fileURLWithPath: ProcessInfo.processInfo.arguments[0])
-        let buildDir = execURL.deletingLastPathComponent()
-        let devPath = buildDir.appendingPathComponent("crow").path
-        if FileManager.default.isExecutableFile(atPath: devPath) {
-            return devPath
+        let sibling = execURL.deletingLastPathComponent().appendingPathComponent("crow").path
+        if FileManager.default.isExecutableFile(atPath: sibling) {
+            return sibling
         }
 
         let candidates = [
@@ -217,5 +219,18 @@ public struct ClaudeHookConfigWriter: HookConfigWriter {
             }
         }
         return nil
+    }
+
+    /// Find the `crow` binary agents and hook configs should invoke. Prefers
+    /// `{devRoot}/.claude/bin/crow` when scaffolded and executable (CROW-552),
+    /// then falls back to `appCrowBinary()` and common install locations.
+    public static func findCrowBinary(devRoot: String? = nil) -> String? {
+        if let devRoot {
+            let symlink = (devRoot as NSString).appendingPathComponent(".claude/bin/crow")
+            if FileManager.default.isExecutableFile(atPath: symlink) {
+                return symlink
+            }
+        }
+        return appCrowBinary()
     }
 }
--- a/Sources/Crow/App/AppDelegate.swift
+++ b/Sources/Crow/App/AppDelegate.swift
@@ -493,7 +493,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             } catch {
                 NSLog("[Crow] Codex scaffold failed: %@", error.localizedDescription)
             }
-            if let crowPath = ClaudeHookConfigWriter.findCrowBinary() {
+            if let crowPath = ClaudeHookConfigWriter.findCrowBinary(devRoot: devRoot) {
                 let codexHome = ProcessInfo.processInfo.environment["CODEX_HOME"]
                     ?? NSString(string: "~/.codex").expandingTildeInPath
                 do {
@@ -516,7 +516,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             } catch {
                 NSLog("[Crow] Cursor scaffold failed: %@", error.localizedDescription)
             }
-            if let crowPath = ClaudeHookConfigWriter.findCrowBinary() {
+            if let crowPath = ClaudeHookConfigWriter.findCrowBinary(devRoot: devRoot) {
                 let cursorHome = ProcessInfo.processInfo.environment["CURSOR_CONFIG_DIR"]
                     ?? NSString(string: "~/.cursor").expandingTildeInPath
                 do {
@@ -539,7 +539,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             } catch {
                 NSLog("[Crow] OpenCode scaffold failed: %@", error.localizedDescription)
             }
-            if let crowPath = ClaudeHookConfigWriter.findCrowBinary() {
+            if let crowPath = ClaudeHookConfigWriter.findCrowBinary(devRoot: devRoot) {
                 // XDG spec: an empty `XDG_CONFIG_HOME` is treated as unset, so
                 // fall through to ~/.config/opencode rather than a relative path.
                 let xdgConfig = ProcessInfo.processInfo.environment["XDG_CONFIG_HOME"]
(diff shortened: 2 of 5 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Nice, focused change: the crow CLI symlink is cleanly separated from `defaults.binaries`, the reap-exclusion set keeps `installBinarySymlinks` from clobbering it, and the fallback ordering in `findCrowBinary(devRoot:)` is right (dangling/broken symlink → `isExecutableFile` fails → falls back to `appCrowBinary()`, so hook configs stay safe). One correctness bug blocks approval.

### Critical Issues

**🔴 `installCrowCLISymlink` never repairs a dangling `crow` symlink — defeats the "repoints when the app binary changes" guarantee** (`Sources/Crow/App/Scaffolder.swift:252`)

The repoint path gates removal on `fm.fileExists(atPath: link)`:

```swift
if fm.fileExists(atPath: link) {          // follows the symlink
    guard ... == .typeSymbolicLink else { ...return }
    try? fm.removeItem(atPath: link)
}
try fm.createSymbolicLink(atPath: link, withDestinationPath: target)
```

`FileManager.fileExists(atPath:)` **follows** symlinks, so for a *dangling* `crow` symlink (its previous target no longer exists) it returns `false`. The removal block is skipped, but the broken symlink inode is still present on disk, so `createSymbolicLink` then throws `EEXIST` ("a file with the same name already exists"), the error is logged + swallowed, and the symlink stays dangling forever.

Verified against Foundation directly:
```
fileExists(link) on dangling: false
attributesOfItem type is symlink: true
createSymbolicLink FAILED: ...a file with the same name already exists.
```

**Failure scenario:** the exact scenario this PR claims to handle. The old target vanishes — user moves `Crow.app` (run from `~/Downloads`, then drag to `/Applications`), a release↔debug rebuild, or `swift package clean` — so `.claude/bin/crow` now dangles. On next scaffold `appCrowBinary()` resolves the new, valid path, but the repoint fails and the symlink keeps pointing at the dead location. `command -v crow` in a fresh agent terminal resolves to the broken link and execution fails — precisely the OpenCode-plugin / PATH item in your own test plan. (Hook configs escape this because `findCrowBinary(devRoot:)` re-checks `isExecutableFile`; only PATH-based agent resolution breaks.)

The existing `installBinarySymlinks` avoids this by *unconditionally* `try? fm.removeItem(atPath: link)` before creating (Scaffolder.swift:221). Mirror that here — or drive the branch off `attributesOfItem` (which does **not** follow the link, as the guard-fail branch above already does) instead of `fileExists`:

```swift
if let attrs = try? fm.attributesOfItem(atPath: link) {
    guard (attrs[.type] as? FileAttributeType) == .typeSymbolicLink else {
        NSLog("[Scaffolder] crow exists at %@ but is not a symlink — leaving alone", link)
        return
    }
    try? fm.removeItem(atPath: link)   // handles live AND dangling symlinks
}
```

A regression test for the dangling case (create the link, delete its target, re-scaffold, assert it repoints) would lock this down — `crowCLISymlinkRepointsWhenAppBinaryChanges` misses it because both `crow-a` and `crow-b` stay on disk, so the old link is never dangling.

### Security Review
**Strengths:**
- No new external input, secrets, or network surface. Symlink target comes from the running app's own executable path, not user-controlled data.
- Non-symlink files dropped by hand are explicitly left untouched (`nonSymlinkCrowFileInBinDirIsLeftAlone` covers this) — no clobber of real files.

**Concerns:** none.

### Code Quality
- 🟢 (consider) `installCrowCLISymlink` runs after `installBinarySymlinks`, so a user-set `defaults.binaries["crow"]` is created and then silently overwritten by the app binary. This is intended per the PR description ("independent of `defaults.binaries`"), but a one-line note or an early skip when `overrides["crow"]` is present would make the precedence explicit.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 0 Yellow, 1 Green] findings.

---

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new external input, network, or credential surface — this is a local filesystem symlink materialization.
- Symlink creation is scoped to `{devRoot}/.claude/bin/crow` and only ever targets the running app's own resolved CLI (`appCrowBinary()`), so there's no injection of an arbitrary/attacker-controlled path.
- Correctly refuses to clobber a non-symlink `crow` file dropped by hand (`Scaffolder.swift:259-262`), avoiding accidental destruction of a user's real file.
- `findCrowBinary(devRoot:)` checks `isExecutableFile` (which follows the link), so a dangling symlink can never leak into hook configs — it falls back to `appCrowBinary()`.

**Concerns:**
- None.

### Code Quality
- **Core fix is correct.** `installCrowCLISymlink` (`Scaffolder.swift:255-264`) drives off `attributesOfItem` (lstat semantics — does not follow the link) instead of `fileExists` (follows the link). This is exactly right: a dangling `crow` symlink looks *absent* to `fileExists`, so the prior code skipped removal and then hit `EEXIST` on `createSymbolicLink`. The new code detects the dangling link, removes it, and repoints. The inline comment documents the rationale well.
- **Clean separation of concerns.** `appCrowBinary()` (app's own CLI) vs `findCrowBinary(devRoot:)` (agent-facing, prefers the scaffolded stable symlink) is a sensible split, and the doc comments make the precedence explicit.
- **Reap exclusion is correct.** `managedBinarySymlinks` (`Scaffolder.swift:182`, `:200`) keeps `installBinarySymlinks` from reaping the crow link, leaving ownership to its dedicated installer that runs immediately after.
- **Well tested.** Five new tests cover always-materialize, persistence across empty overrides, repoint-on-change, dangling-link repair, and non-symlink safety — mapping directly onto each behavior branch.
- **Consider (non-blocking):** In the narrow case where `appCrowBinary()` resolves to `nil` *and* a user has configured `defaults.binaries["crow"]`, `installBinarySymlinks` creates the user's crow link and then `installCrowCLISymlink`'s guard-else branch reaps it (`Scaffolder.swift:244-253`). In a real app bundle `appCrowBinary()` resolves the bundled sibling and effectively never returns nil, and `crow` as a user override isn't an expected config — so this is a theoretical corner, not a defect to fix here.

### Verification
`swift build` and the test suite could not be executed in this review environment — the checkout is missing the `GhosttyKit.xcframework` binary artifact (an environment limitation, unrelated to these changes). The review was performed by static analysis of the diff and surrounding code; the logic and test coverage are sound.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 1 Green] findings. Focused, correct, and well-tested fix for the dangling-symlink repoint edge case.

Issue #548, opened 2026-07-01
Does Crow run on Intel (x86_64) Macs? Audit build/compilation for architecture support


dgershman:
Question to resolve, then fix if needed: does Crow run on Intel (x86_64) Macs, and does the current compilation/build produce (or allow producing) an Intel-compatible binary?

Right now everything is being developed and run on Apple Silicon (arm64), and at least one referenced binary is arm64-only (e.g. the corveil default binary path is `corveil-darwin-arm64`). It's unclear whether the app builds and runs on Intel hardware at all.

Please investigate and report:
- Whether the current build targets arm64-only or produces a universal (arm64 + x86_64) binary.
- Whether any dependencies, bundled binaries, or hardcoded arch assumptions block Intel Macs.
- The minimum macOS version / hardware assumptions in the build config.

Then, if Intel support is desired and not currently working, make the build produce a universal binary (or otherwise support x86_64) so Crow runs on Intel Macs. If Intel support is intentionally out of scope, document that (README / build docs) and fail gracefully with a clear message on unsupported architectures.

Acceptance:
- A clear answer, documented in the issue/PR, on whether Crow currently runs on Intel Macs.
- If pursuing support: the app builds and launches on an x86_64 Mac (universal binary or x86_64 target), with any arch-specific binaries handled.
- If not: supported architectures are documented and unsupported ones are handled gracefully.

[🐦‍⬛ Created with Crow via Claude Code](https://github.com/radiusmethod/crow)

dgershman:
resolved

Issue #512, opened 2026-06-15
Settings text fields can't paste (⌘V / dictation) — app has no Edit menu


dhilgaertner:
## Summary

Text fields across the **Settings** screens can't accept a paste — `⌘V` does nothing, and voice-recognition/dictation insertion fails too. Copy something outside the app, click into any Settings text field, `⌘V` → nothing happens. This affects the whole Settings area.

## Root cause — the app has no Edit menu

`Sources/Crow/App/AppDelegate.swift` → `setupMenu()` (~lines 958–996) builds a custom `NSApp.mainMenu` that contains **only the App menu** (About / Settings / Restart / Reload / Hide / Quit). There is **no Edit menu** anywhere in the app.

On macOS, the standard text-editing actions (`cut:`, `copy:`, `paste:`, `selectAll:`, undo/redo) are routed through the **Edit menu's** items. With no Edit menu, the `paste:` selector isn't registered for validation/routing, so when `⌘V` is pressed in a SwiftUI `TextField` the system finds no menu item to handle it and silently drops the keystroke. **Dictation/voice paste uses the same responder-chain path, so it fails for the same reason.**

This is not a per-field problem: the Settings TextFields are all plain SwiftUI `TextField`/`TextEditor` (no custom `NSViewRepresentable`), and **no key monitor is swallowing ⌘V** — Ghostty's `.keyDown` monitor is gated by `guard NSApp.mainWindow == nil`, and the Settings window is key, so it doesn't interfere. The single missing piece is the Edit menu.

## Fix

Add a standard **Edit menu** in `setupMenu()` (before assigning `NSApp.mainMenu`) wiring the standard selectors so the responder chain can route them:

```swift
let editMenuItem = NSMenuItem()
let editMenu = NSMenu(title: "Edit")
editMenu.addItem(withTitle: "Undo",   action: Selector(("undo:")),  keyEquivalent: "z")
editMenu.addItem(withTitle: "Redo",   action: Selector(("redo:")),  keyEquivalent: "Z")
editMenu.addItem(.separator())
editMenu.addItem(withTitle: "Cut",    action: #selector(NSText.cut(_:)),       keyEquivalent: "x")
editMenu.addItem(withTitle: "Copy",   action: #selector(NSText.copy(_:)),      keyEquivalent: "c")
editMenu.addItem(withTitle: "Paste",  action: #selector(NSText.paste(_:)),     keyEquivalent: "v")
[...]

PR #556, opened 2026-07-02
Replace Ghostty with xterm.js terminal renderer for Intel Mac support (#548)

dgershman:
## Summary

- Replace GhosttyKit with an xterm.js terminal surface in `WKWebView`, backed by a native PTY running `tmux attach-session` — tmux remains the sole session backend (ADR 0001).
- Drop Zig/Ghostty from the build path; Crow now builds with SwiftPM only and ships universal macOS binaries (`arm64` + `x86_64`).
- Document the decision in ADR 0006 and update CI/Makefile/docs accordingly.

## Known limitations (v1 spike)

- Terminal UX regresses vs Ghostty: no GPU rendering, OSC 8 link hover, Quick Look, or rich context menu.
- WebKit may still show a native context menu on right-click (left as-is for now).
- Manager process-exit banner not wired for the shared attach client.

## Test plan

- [ ] `make build` succeeds on Apple Silicon
- [ ] `swift build --arch x86_64` succeeds (Intel compile check)
- [ ] Launch Crow, open a session terminal — tmux attach renders and accepts input
- [ ] Claude/Cursor TUI icons render correctly with a Nerd Font installed
- [ ] Cmd+F opens terminal search (tmux copy-mode path)
- [ ] Mouse selection/copy via tmux bindings (`crow-tmux.conf`) works


dhilgaertner (changes requested):
## Code & Security Review

Solid, well-scoped migration. I built and verified locally: **`swift build` succeeds on arm64 AND `swift build --arch x86_64` succeeds** — the central Intel-support claim is real (the old GhosttyKit xcframework only shipped a `macos-arm64` slice, so this genuinely unblocks Intel). No Ghostty *symbols* remain in Swift sources (only stale comments). The full CrowTerminal test suite passes except one failure, `retryReadinessEmitsTimedOutWhenSentinelMissing`, which reproduces identically and is **pre-existing / host-environment-dependent** — this PR touches neither the readiness/sentinel logic nor the test (the only `TmuxBackend.swift` edits are type renames, logging, and comments). Not a regression from this change.

### Critical Issues
None. No security concerns: the WKWebView loads only bundled local files via `loadFileURL`, the PTY command is internally constructed with `shellQuote` (not user input), and PTY output is shipped to JS as a JSON-encoded base64 string (`crowWrite`) so it can't break out of the eval context.

### Security Review
**Strengths:**
- No remote content loaded into the web view; `allowingReadAccessTo` is scoped to the resource folder.
- Native context menus / pasteboard requestors are suppressed on both the surface and the web view.
- `crowInput` only writes bytes the user is already typing into their own terminal — no privilege escalation surface.

**Concerns:** None blocking.

### Code Quality

**🟡 `scripts/build-ghostty.sh` is now orphaned dead code that contradicts the PR.** It was *expanded* (+300/−80) to assemble a **universal Ghostty** framework, yet this PR removes Ghostty from the build path entirely — the Makefile `ghostty` target and the mise `build:ghostty` task are both gone, and nothing else references the script (`rg build-ghostty` matches only its own usage comment). Combined with the retained `vendor/ghostty` submodule + `.gitmodules` entry and the `Frameworks/GhosttyKit*` `.gitignore` lines, this leaves a large chunk of dead Ghostty infrastructure that will mislead future maintainers. Delete the script (and ideally drop the submodule), or add a comment explaining why it's intentionally retained.

**🟡 Manager process-exit detection is silently lost.** Removing the `GhosttyApp.onChildExited` wiring means `appState.managerProcessExited` can never become true, so the "Manager process exited" restart banner no longer fires. This is documented under Known limitations, but it's a regression of shipped behavior — please file a follow-up issue and reference it in the code comment at `AppDelegate.swift` so it isn't lost.

**🟡 `PTYProcess.terminate()` never reaps the child → zombie process.** `terminate()` sends `SIGTERM` and immediately sets `childPID = -1`. The only `waitpid` is in `waitForExit()`, which is reached via the EOF path and guards on `pid > 0`; when `terminate()` runs first, that guard fails and the child is never reaped, leaving a zombie until the app exits. Low impact (one shared cockpit process), but the documented teardown path leaks. Consider reaping in `terminate()` (or `WNOHANG` + async reap).

**🟢 Duplicate `TERM`/`COLORTERM` env entries.** `PTYProcess.start` appends `TERM`/`COLORTERM` unconditionally, unlike `LANG`/`LC_CTYPE` which are guarded with `contains(where:)`. When Crow is launched from a shell that already exports `TERM`, the env array carries two `TERM=` entries and the inherited value may win. Guard these the same way for consistency.

**🟢 Stale comments referencing Ghostty.** `AppDelegate.swift` still claims `XTermSurfaceView.performKeyEquivalent` intercepts ⌘V/⌘C, but the new implementation only handles ⌘F (copy/paste is now WebKit's). Also `TmuxBackend.swift:55` (\"libghostty needs an NSWindow before ghostty_surface_new\"), `Enums.swift:63-64`, and `SessionService.swift:1505-1506` still describe Ghostty internals. Doc-only cleanup.

**🟢 `scheduleFit()` chattiness.** Called on every `layout()` plus a 50ms delayed re-fit, so live-resize produces a stream of `evaluateJavaScript` round-trips. Fine functionally; consider debouncing if resize feels janky.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 3 Yellow, 3 Green] findings. The migration itself is clean and the Intel build is verified; the Yellows are cleanup/tracking items (dead Ghostty script, dropped Manager-exit banner, PTY zombie) best folded into this round trip rather than a follow-up.

---

dhilgaertner (changes requested):
## Code & Security Review

Solid, well-scoped swap of GhosttyKit for an xterm.js-in-WKWebView surface backed by a native PTY. Build passes (`swift build` ✓), the Ghostty framework/Zig toolchain is cleanly excised from `Package.swift`, Makefile, CI, and mise, and ADR 0006 documents the decision. The security posture of the web layer is good: PTY bytes reach `term.write()` as raw `Uint8Array` (no `innerHTML`), and Swift→JS injection is prevented by base64 + `JSONEncoder` encoding in `sendToTerminal`. Native context menus and pasteboard `validRequestor` are correctly suppressed at both the view and WebView level.

### Security Review
**Strengths:**
- No dynamic HTML injection; terminal output is written as bytes to xterm.js, escape-sequence handling stays inside the sandboxed WebView.
- `evaluateJavaScript("window.crowWrite(\(jsonString))")` uses a JSON-encoded base64 payload — not string-interpolatable into JS.
- `terminal.html` is loaded via `loadFileURL(..., allowingReadAccessTo:)` scoped to the `xterm/` folder only.
- Vendored `xterm.js`/`xterm-addon-fit.js` are the standard upstream minified UMD bundles (MIT).

**Concerns:**
- No `WKNavigationDelegate.decidePolicyFor:` guard. Nothing currently drives navigation (WebLinksAddon isn't loaded), but adding a policy that blocks any non-`file://` navigation would harden the surface against terminal content ever redirecting the WebView. (Green)

### Code Quality

**Yellow — should fix:**
- `PTYProcess` is `@unchecked Sendable` but `masterFD`/`childPID` are shared across threads without synchronization. `terminate()` (main thread) does `readSource?.cancel(); ...; close(masterFD)` while the read-source event handler runs on `readQueue` and may be mid-`Darwin.read(self.masterFD, ...)`. `cancel()` does not synchronously stop an in-flight handler, so `close()` can race a concurrent `read()` on the same descriptor — a classic use-after-close / fd-reuse race at teardown. Serialize fd access (e.g. cancel + close on `readQueue`, or gate on a lock) — `PTYProcess.swift` `terminate()` / `startReading()`.

**Green — consider:**
- `PTYProcess.terminate()` sends `SIGTERM` and clears `childPID` but never `waitpid`s, so the `/bin/bash -c "tmux attach"` child becomes a zombie until Crow exits. Only the natural-EOF path (`count == 0` → `waitForExit`) reaps. Bounded to one process for the single long-lived cockpit surface, but a `waitpid` (or `SIGCHLD` reap) in `terminate()` would close it. `PTYProcess.swift`.
- `handleResize` does `UInt16(max(rows, 1))` on an `Int` from JS — values >65535 trap. Input is trusted/local and bounded by terminal geometry, but `UInt16(clamping:)` avoids any crash surface. `XTermSurfaceView.swift` `handleResize`.
- Stale `Ghostty*` references remain in comments (`Enums.swift:63-64`, `SessionService.swift:187/716/884/1505`, `TmuxBackendTests.swift:6-7`). Cosmetic; worth a sweep since the types are gone.
- Per-chunk `evaluateJavaScript` with base64 for every PTY read is inefficient for high-throughput output — acceptable for the v1 spike, but a batching/coalescing pass is a natural follow-up.

### Test note
- `retryReadinessEmitsTimedOutWhenSentinelMissing` (`TmuxBackendTests.swift:365`) fails deterministically on a fast host: the shell fires `precmd` inside the 150ms budget so `.shellReady` arrives before `.timedOut`. This test and the readiness code it exercises are **not** touched by this PR — it's a pre-existing timing-sensitivity, flagged for awareness, not attributed to these changes.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 4 Green] findings. The Yellow (PTYProcess fd close/read race) is the only blocker; addressing it — plus the cheap zombie-reap and resize-clamp Greens — in the same round trip lands a clean spike.

---

dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues

**🔴 `swift build` fails — the project does not compile.** `Packages/CrowTerminal/Sources/CrowTerminal/PTYProcess.swift:124`

```swift
readQueue.async { _ in
    var status: Int32 = 0
    waitpid(pid, &status, 0)
}
```

`DispatchQueue.async(execute:)` takes a `() -> Void` closure (zero arguments), but this closure declares a parameter (`_ in`). The compiler rejects it:

```
PTYProcess.swift:124:27: error: contextual closure type
'@Sendable @convention(block) () -> Void' expects 0 arguments,
but 1 was used in closure body
```

This is the sole compile error, and it fails the whole build (`swift build` exits non-zero). It was introduced by the "fix PTY reap" commit — the sibling reap path in `waitForExit()` (line 155) correctly uses `readQueue.async { [weak self] in … }` (a capture list, not a parameter). Fix is a one-liner: drop the `_ in`, or use `[pid] in` / a capture list if you want an explicit capture.

The PR's own test plan (`make build succeeds`, `swift build --arch x86_64 succeeds`) is unchecked, consistent with this. Since the stated goal of the PR is Intel-arch build support, a green build is table stakes.

### Security Review

**Strengths:**
- `sendToTerminal` base64-encodes PTY output and JSON-encodes it before injecting into `evaluateJavaScript`, so raw terminal bytes can't break out of the JS string context — good.
- `cockpitSurface()` builds the `tmux attach` command with `shellQuote()` (proper single-quote escaping) around every interpolated value; the only inputs are app-configured (`tmuxBinary`, `socketPath`) or a constant session name. No injection surface for untrusted input.
- WKWebView loads a local bundled `file://` page with read access scoped to the resource folder; no remote content, `allowTransparency: false`, native context menu suppressed. Reasonable posture.

**Concerns:**
- None beyond the code-quality items below.

### Code Quality

**🟡 PTY read handler ignores read errors (`count < 0`).** `PTYProcess.swift:130-149`

The `DispatchSourceRead` handler only branches on `count > 0` and `count == 0`. When `read()` returns `-1` (e.g. `EIO` on the pty master after the child/slave goes away — a documented outcome on Darwin, not just clean `EOF`=0), the handler falls through and does nothing. Because the source is level-triggered on readability and an errored fd stays "readable", the handler re-fires in a tight loop (CPU spin) and `onExit` is never invoked — so `onProcessExit` never fires. Handle `count <= 0` (or explicitly the `< 0` / non-`EINTR` case) by cancelling the source and reaping, same as the `count == 0` path.

**🟢 Exit status decoding drops signal info.** `PTYProcess.swift:158` — `(status >> 8) & 0xFF` is `WEXITSTATUS`, which is only meaningful when the child exited normally. A child killed by a signal reports code `0` here, masking abnormal termination. Consider `WIFSIGNALED`/`WTERMSIG` if the exit reason matters to the UI banner.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 1 Yellow, 1 Green] findings. The Red is a hard build break; nothing else can be validated until it compiles.

---

dhilgaertner (changes requested):
## Code & Security Review

Verified locally: `swift build` succeeds (after `scripts/generate-build-info.sh`) and all 66 CrowTerminal package tests pass. The WKWebView bridge design is sound and the Swift-side changes are clean and well-scoped. No blocking (Red) issues found; the Yellow items below should be fixed before merge.

### Critical Issues
None.

### Security Review

**Strengths:**
- The WKWebView loads only the bundled `terminal.html` with `loadFileURL(_:allowingReadAccessTo:)` scoped to the xterm resource folder — no remote content, no navigation surface (`XTermSurfaceView.swift:87`).
- PTY output is delivered to JS as base64 wrapped in `JSONEncoder`-escaped strings (`XTermSurfaceView.swift:210-215`), so terminal output can never break out of the `crowWrite` call into script injection.
- JS→native input goes through typed `WKScriptMessageHandler` messages with defensive casts; resize values are clamped (`XTermSurfaceView.swift:165-167`).
- `posix_spawn` with `POSIX_SPAWN_SETSID` and explicit fd hygiene (close master in child, dup2 slave, close slave when > stderr) is correct (`PTYProcess.swift:40-46`).

**Concerns:**
- 🟡 **Vendored xterm.js has no version, provenance, or license documentation.** `Resources/xterm/xterm.js` is 388 KB of minified JS with no version string, and the UMD wrapper / `xterm-addon-fit` naming indicates a pre-5.x (4.x-era, ~2021) build. There's no way to audit it against an upstream release, verify integrity, or know which upstream security fixes (there have been escape-sequence handling fixes since 4.x) it contains. Please pin and document the exact version + source URL (e.g. a `VERSION`/README next to the assets or a fetch-at-build script), include the xterm.js MIT license text, and preferably upgrade to a current `@xterm/xterm` 5.x release.

### Code Quality

- 🟡 **`PTYProcess.terminate()` closes `masterFD` while the read source may still be running** (`PTYProcess.swift:109-120`). `DispatchSourceRead.cancel()` is asynchronous; the event handler can be mid-`read(self.masterFD, …)` on `readQueue` when the main thread closes the fd — a read on a closed (and potentially reused) descriptor. The supported pattern is to close the fd in the source's `setCancelHandler` and let `cancel()` drive the close.
- 🟡 **Read handler ignores `count < 0`** (`PTYProcess.swift:127-137`). On a read error (e.g. `EIO` after the child exits, or `EINTR`) the handler does nothing, the source stays armed on a permanently-readable fd, and the handler re-fires in a busy loop — pinning a core and never firing `onExit`. Treat `count <= 0` (excluding `EINTR`/`EAGAIN`) as EOF: cancel the source and reap the child.
- 🟡 **`scripts/build-ghostty.sh` got a 300-line universal-build rewrite but is now dead code.** Nothing invokes it anymore — the `make ghostty` target, `mise build:ghostty` task, and all CI steps were removed — yet the `vendor/ghostty` submodule also remains in `.gitmodules`. This contradicts the PR's "drop Zig/Ghostty from the build path" and leaves a large untested script to rot. Either delete the script (and submodule), or if it's intentionally kept as the fallback documented in ADR 0006's alternatives, say so in the script header and leave it un-rewritten.
- 🟢 `terminate()` sends `SIGTERM` but never `waitpid`s the killed child (`PTYProcess.swift:112-115`), leaking a zombie per surface teardown (one per crash-watchdog "Restart tmux server" cycle in a long-running app). Reap after kill.
- 🟢 Exit-status decoding `(status >> 8) & 0xFF` (`PTYProcess.swift:149`) assumes `WIFEXITED`; a signal-terminated child reports exit code 0. Harmless today since `onProcessExit` isn't consumed, but worth fixing before the Manager exit banner gets wired to it.
- 🟢 Stale comment: `AppDelegate.swift:1114` now says "`XTermSurfaceView.performKeyEquivalent` intercepts ⌘V/⌘C" but the new implementation only intercepts ⌘F (`XTermSurfaceView.swift:131-139`). ⌘V/⌘C ride on WebKit's default responder behavior — update the comment and double-check the #512 paste path in the test plan.
- 🟢 `webView.setValue(false, forKey: "drawsBackground")` (`XTermSurfaceView.swift:41`) is a private-API KVC key; if WebKit ever drops it this raises `NSUnknownKeyException` at init. Consider `underPageBackgroundColor`/CSS instead, or wrap defensively.
- 🟢 No handler for `webViewWebContentProcessDidTerminate` — a WebContent process crash leaves a permanently blank cockpit with no recovery path. A reload-and-reattach handler would make the single shared surface more resilient.
- 🟢 `TmuxBackend.offscreenWindow` (`TmuxBackend.swift:86-95`) is now unused (the comment at `TmuxBackend.swift:639-641` says the WKWebView must not be parked offscreen) — dead code to remove.
- Tests: `PTYProcess` and the `XTermSurfaceView` message bridge have no unit tests. The pure parts (exit-status decode, pending-output buffering) are testable without a real PTY.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 4 Yellow, 6 Green] findings.

---

dhilgaertner (changes requested):
## Code & Security Review

Solid, well-documented migration off Ghostty. The xterm.js vendoring is clean (v6.0.0 with `LICENSE` + `VERSION` provenance, local `file://` load, no remote content, script-message bridge is a tight allowlist), the build/CI simplification is coherent, and the universal-binary plumbing in `bundle.sh`/`release.yml` looks correct. One confirmed blocker below.

### Critical Issues

**🔴 `TERM=xterm-kitty` breaks the terminal on stock macOS (no kitty installed)**
`Packages/CrowTerminal/Sources/CrowTerminal/PTYProcess.swift:62-64`

When `TERM` is absent from the environment, `PTYProcess.start` defaults it to `xterm-kitty`. The cockpit surface's PTY command is `tmux … attach-session …` (`TmuxBackend.cockpitSurface`, TmuxBackend.swift:613-620), so this `TERM` is what the tmux **client** uses to load terminfo.

`xterm-kitty` terminfo ships only with the kitty terminal — it is **not** present on a stock macOS install. I reproduced the failure in a real PTY on this machine (tmux 3.6a):

```
TERM=xterm-kitty:   attach_error=True  → 'missing or unsuitable terminal: xterm-kitty\r\nEXITCODE=1'
TERM=xterm-256color: attach_error=False → normal screen redraw
```

`tmux attach` exits 1 immediately, the PTY child dies, and the surface renders blank/broken with no recovery path (`onProcessExit` isn't wired for the cockpit surface).

Why this is easy to miss: a GUI `.app` launched from Finder/Dock inherits **no** `TERM`, so the `xterm-kitty` default kicks in for real users. A developer running from a terminal inherits `TERM=xterm-256color`, so the bug never appears in dev — and the test-plan item "Launch Crow, open a session terminal — tmux attach renders" is still unchecked.

Fix: default to `TERM=xterm-256color` (matches the conf's `default-terminal screen-256color` and is what xterm.js expects; the ImageAddon's sixel/iip/kitty-graphics support is independent of `TERM`). If kitty keyboard/graphics advertising is actually wanted, the app would have to bundle+`TERMINFO` the `xterm-kitty` entry, which this PR does not do.

### Security Review
**Strengths:**
- xterm.js loaded from a local `file://` URL with `allowingReadAccessTo` scoped to the resource folder; no remote origin, no JS eval of untrusted input.
- Native context menu / Services / Autofill suppressed on both the web view and host view; xterm textarea hardened (`autocomplete/autocorrect/spellcheck` off).
- `cockpitSurface` shell-quotes `tmuxBinary`/`socketPath`/`sessionName` before the `/bin/bash -c` attach command — no injection surface.
- PTY output is base64-framed into `window.crowWrite`, keeping raw bytes intact without string-escaping pitfalls.

**Concerns:** none.

### Code Quality
**🟡 Data races on `PTYProcess` FD/PID across the main thread and the read queue** — `PTYProcess.swift:101-107, 113-132, 134-167`
`masterFD` and `childPID` are plain `var`s (`@unchecked Sendable`) read/written from both the main thread (`write`/`resize`/`terminate`) and `readQueue` (event/cancel handlers, `reapChild`). A `write()` racing the cancel handler that `close()`s and `-1`s `masterFD` can write to a closed/reused descriptor. Teardown-scoped and rare, but it is a genuine race — consider serializing FD/PID access on `readQueue` (or an `os_unfair_lock`). Not a blocker on its own.

🟢 Manager process-exit banner intentionally dropped for the shared attach client — acknowledged and tracked in #558. Fine as a v1 limitation.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 1 Yellow, 1 Green] findings. The `TERM=xterm-kitty` default will break the terminal for the exact stock-macOS/Intel users this PR targets; switching to `xterm-256color` should be a one-line fix.

---

dhilgaertner (approved):
## Code & Security Review

Re-review after the three fix commits (PTY teardown race, xterm 6 + image addon, CI build, WKWebView hardening). The Ghostty→xterm.js migration is clean and the previously-raised issues are addressed. `swift build` succeeds locally.

### Critical Issues
None.

### Security Review
**Strengths:**
- PTY output is base64-encoded and JSON-escaped before `evaluateJavaScript("window.crowWrite(...)")` (`XTermSurfaceView.swift:222-227`) — untrusted terminal bytes cannot break out into the JS context.
- JS→native `WKScriptMessageHandler` messages are strictly type-checked (`crowInput`/`crowResize`/`crowError`) before use (`XTermSurfaceView.swift:270-299`).
- `WKWebView` loads the bundled page via `loadFileURL(_, allowingReadAccessTo:)` scoped to the `xterm` folder (`XTermSurfaceView.swift:88`); `NavigationHandler` cancels any non-`file://` navigation (`XTermSurfaceView.swift:307-317`). I empirically confirmed `responds(to:)` is `true` for the delegate selector despite the "nearly matches" compiler warning — WebKit does invoke it at runtime, so the gate is effective.
- Native context menus / Services / pasteboard requestors are suppressed on both the container and the `WKWebView` (`XTermSurfaceView.swift:116-125, 252-263`).
- The attach command is shell-quoted before reaching `PTYProcess` (`TmuxBackend.cockpitSurface`); PTY spawns via `posix_spawn` with an explicit argv (no shell interpolation of untrusted data beyond the quoted attach string).

**Concerns:** None blocking.

### Code Quality
- **PTY teardown race** looks correct now: `terminate()` nulls `childPID` before dispatching `waitpid`, and `reapChild` guards on `pid > 0`, so a concurrent read-EOF and explicit terminate can't double-notify exit (`PTYProcess.swift:113-181`).
- **WebContent crash recovery** reloads the UI while keeping the PTY alive (`startPTYIfNeeded` short-circuits on `isPTYStarted`), so a renderer crash doesn't orphan the tmux client (`XTermSurfaceView.swift:157-163, 181-196`).

### Notes (non-blocking, Green)
- **Test flake (environmental, not introduced here):** `TmuxBackendTests.retryReadinessEmitsTimedOutWhenSentinelMissing` fails locally 3/3 — it asserts an idle shell won't fire a prompt within a 150ms budget, but on a fast host the prompt arrives first (`.shellReady` instead of `.timedOut`). This PR does not touch `retryReadinessWatch`/sentinel logic (the TmuxBackend diff is renames + the cockpit-surface change), and the test exercises the tmux path, not the new PTY/WebView surface — so it's a pre-existing timing fragility, not a regression. Worth hardening separately (drive it off a deterministic signal rather than a wall-clock budget).
- **Navigation gate breadth:** `decidePolicyFor` allows *any* `file://` URL, not just the bundled `terminal.html`. No realistic trigger exists today (no weblinks addon; `term.write` doesn't navigate), but narrowing to the expected page would tighten defense-in-depth.
- **Delegate warning:** silence the `webView(_:decidePolicyFor:decisionHandler:)` "nearly matches" warning by matching the `@escaping @MainActor @Sendable` closure signature (or moving it to an extension) so the intent stays obvious to future readers.
- **`PTYProcess.write`** ignores partial-write return values (`PTYProcess.swift:101-107`). Fine for interactive keystrokes; if this path ever carries large pastes, loop until fully written.
- Manual test-plan items (Intel compile, TUI icon/Nerd Font rendering, Cmd+F copy-mode, mouse selection) are unchecked in the PR body — please confirm before merge since they can't be verified in CI.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 4 Green] findings. The one failing test is a pre-existing environment-sensitive flake unrelated to this change.

---

dhilgaertner (changes requested):
## Code & Security Review

Solid, well-scoped migration off GhosttyKit to an xterm.js/`WKWebView` surface backed by a native PTY. The build succeeds (`swift build`), the Ghostty toolchain (Zig, submodule, Metal, XCFramework) is fully excised from CI/Makefile/mise/Package.swift with no dangling references, and the teardown races called out in earlier rounds are addressed. The PR is honest about its v1-spike UX regressions.

I verified the key security control: although the compiler warns that `NavigationHandler.webView(_:decidePolicyFor:decisionHandler:)` "nearly matches" the protocol requirement, `responds(to:)` for the selector returns `true`, so the non-file-URL navigation guard **does** fire at runtime. Good.

### Critical Issues
None — no blocking Red findings.

### Security Review
**Strengths:**
- `terminal.html` is loaded via `loadFileURL(..., allowingReadAccessTo: folder)` scoped to the xterm resource dir, not the bundle/home — read access is sandboxed.
- Non-`file://` navigation is cancelled in `decidePolicyFor` (verified live).
- PTY output is delivered to JS as a JSON-encoded base64 string and decoded with `atob` into `term.write(bytes)` — terminal bytes are never interpreted as HTML, so no XSS path from shell output. Native context menus and pasteboard `validRequestor` are suppressed.
- PTY spawn uses `posix_spawn` with an explicit argv/env and `POSIX_SPAWN_SETSID`; the attach command is `shellQuote`d.

**Concerns:**
- **[Yellow] `XTermSurfaceView.swift:307` — nav-policy delegate signature mismatch on a security control.** The `decisionHandler` closure is `@escaping (WKNavigationActionPolicy) -> Void` but the SDK requirement is `@escaping @MainActor @Sendable (...)`, producing the "nearly matches optional requirement" warning. It works today (selector still bridges), but this is the classic footgun where a future toolchain silently stops treating it as the witness and the navigation guard goes dark with no error. Since this is the method blocking non-file navigation, match the signature exactly (or move it to an `extension NavigationHandler: WKNavigationDelegate` and annotate the closure) so the warning — and the latent silent-regression risk — is gone.

### Code Quality
- **[Green] `XTermSurfaceView.swift:157` — WebContent-crash reload has no cap/backoff.** `handleWebContentProcessTerminated` → `createSurface()` unconditionally; a reproducibly-crashing WebContent process would reload-crash-reload indefinitely. Load latency keeps it from being a tight CPU spin, but a bounded retry (matching the `failed`/Retry story used elsewhere) would be more robust.
- **[Green] `XTermSurfaceView.swift:314` — nav policy allows *all* file URLs, not just the resource dir.** The `allowingReadAccessTo` sandbox constrains actual reads and nothing can currently trigger such a navigation (no weblinks addon; terminal bytes aren't HTML), but tightening the allow to the bundled folder would make the guard self-contained.
- **[Green] `PTYProcess.swift` — no `deinit` safety net.** If a `PTYProcess` is deallocated without `terminate()`/`destroy()`, the child (tmux attach client) is orphaned and `masterFD` leaks. Only the app-lifetime shared cockpit surface exists today (and it calls `destroy()`), so impact is low, but a `deinit` that cancels the source / reaps would harden reuse.
- **[Green] `PTYProcess.swift:118` — `terminate()` does a blocking `waitpid(pid, &status, 0)` on the serial `readQueue`.** A child that ignores `SIGTERM` would wedge the queue. tmux's attach client exits on `SIGTERM` so this is fine in practice; a `WNOHANG` poll + short grace + `SIGKILL` would remove the theoretical hang.

### Test Note (informational, not attributable to this PR)
`retryReadinessEmitsTimedOutWhenSentinelMissing` fails on this host — it observes `.shellReady` within its 150 ms budget because the real shell prompt fires faster than expected. This PR only touched comments in that test file and did not change readiness logic, so it's pre-existing timing flakiness, not a regression. Worth a follow-up to make the budget deterministic, but it doesn't block this PR.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 4 Green] findings. The single Yellow (silence the nav-delegate signature warning on the security boundary) is a trivial fix; addressing it plus any of the Green items in the same round trip clears the path to approval.

---

dhilgaertner (approved):
## Code & Security Review

Re-review after the 05:35 UTC round. Commits `3d604e6` and `30386ec` land after that review and resolve **all five** of its findings — the single Yellow and all four Greens:

1. **Nav-delegate signature (Yellow)** — `NavigationHandler.webView(_:decidePolicyFor:decisionHandler:)` now declares `@escaping @MainActor @Sendable (WKNavigationActionPolicy) -> Void`, matching the SDK requirement exactly. A forced recompile of `XTermSurfaceView.swift` produces **zero** compiler warnings — the "nearly matches optional requirement" warning (and its silent-regression risk on the security boundary) is gone.
2. **File-URL scoping** — the navigation policy now allows only `file://` URLs inside `allowedResourceDirectory` (standardized-path prefix check, `XTermSurfaceView.swift:331-337`); everything else is cancelled.
3. **WebContent-crash reload cap** — `handleWebContentProcessTerminated` is bounded by `maxWebContentReloadAttempts = 3`, falling through to `onSurfaceCreationFailed` (the existing failed/Retry story).
4. **`PTYProcess` deinit safety net** — added: cancels the read source, SIGTERMs + reaps the child, closes `masterFD` under `readQueue.sync`.
5. **Blocking `waitpid` on terminate** — `forceReap` polls `WNOHANG` with a ~500ms grace then escalates to `SIGKILL`; the serial read queue can no longer wedge on a SIGTERM-ignoring child.

### Critical Issues
None.

### Security Review
**Strengths:**
- `terminal.html` loads via `loadFileURL(..., allowingReadAccessTo:)` scoped to the xterm resource directory, and the navigation delegate independently cancels any navigation outside that directory — two layers on the same boundary.
- PTY output crosses into JS as JSON-encoded base64 decoded straight into `term.write(bytes)`; terminal bytes are never interpreted as HTML, so no XSS path from shell output. No JS→native bridge beyond the four typed message handlers.
- `posix_spawn` with explicit argv/env and `POSIX_SPAWN_SETSID`; the tmux attach command is built from app-controlled values with correct POSIX single-quote escaping (`shellQuote`).
- Vendored xterm.js 6.0.0 / addon-fit 0.11.0 / addon-image 0.9.0 with LICENSE + VERSION provenance recorded; no CDN/network fetch at runtime.

**Concerns:** none at Red/Yellow.

### Code Quality
- **[Green] `PTYProcess.swift:204` — `forceReap` doesn't break on `waitpid` returning -1.** If the wait fails outright (e.g. ECHILD from an unexpected external reap), the loop burns the full grace period and then `SIGKILL`s a PID that may no longer belong to the child. Breaking out of the loop when `waitpid` returns -1 with `errno == ECHILD` would remove the theoretical stale-PID signal. Unreachable today (reaping is serialized on `readQueue` and guarded by `childPID`).
- **[Green] `terminal.html` — no CSP `<meta>` tag.** All scripts are local and navigation is delegate-blocked, so this is purely defense-in-depth, but a `default-src 'none'; script-src 'self'; style-src 'self' 'unsafe-inline'`-style policy would make the page self-defending if the surrounding controls ever loosen.
- **[Green] `XTermSurfaceView.swift:44` — `setValue(false, forKey: "drawsBackground")` uses a private WKWebView KVC key.** Standard community workaround, but it can throw or no-op on a future macOS; `underPageBackgroundColor`/`.clear` on macOS 12+ is the supported route.

### Verification
- `swift build` — succeeds; forced recompile of `XTermSurfaceView.swift` + `PTYProcess.swift` emits zero warnings.
- `swift test` (CrowTerminal) — **66/66 passed**, including the TmuxBackend integration suite and `retryReadinessEmitsTimedOutWhenSentinelMissing` (the test flagged as flaky last round passed on this run).

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 3 Green] findings. All prior-round findings are resolved; the remaining Greens are optional hardening.

---

PR #566, opened 2026-07-02
Auto-complete job sessions when the run finishes

### LINKED ISSUE #561: Auto-move jobs to completed status when they finish
Scheduled/auto jobs should automatically transition to `completed` in Crow once they finish. Today a job runs, does its work (e.g. files a ticket, opens a PR), but its Crow session/status doesn't advance to completed on its own — it lingers in an active/in-progress state and has to be moved by hand.

Expected: when a job's run reaches a terminal successful end (its prompt(s) completed and the agent has exited / gone idle with no further work), Crow sets the associated session status to `completed` automatically (the same `set-status … completed` transition, driven by the app rather than a manual step).

Consider:
- Define what "done" means for a job run — agent exit, idle-with-no-pending-work, or an explicit completion signal from the agent's state hooks — and use the most reliable signal available. This ties into the per-agent state-signal source (Claude hooks / OpenCode plugin `crow hook-event`).
- Only auto-complete on success; a job that errored or is awaiting input should not be silently marked completed. Failed/blocked runs should surface, not disappear.
- Applies to the `jobs` defined in config (Tanzaterra Tech Debt, Shell-crm renovates, etc.), and shouldn't affect manager sessions.

Acceptance:
- A job that finishes successfully has its Crow session status transition to `completed` with no manual action.
- Errored/awaiting-input job runs are NOT auto-completed.
- Behavior is consistent across agent kinds (Claude/OpenCode/Codex/Cursor) via their state-signal sources.

dgershman:
Closes #561

## Problem

Scheduled/automated jobs (e.g. "Tanzaterra Tech Debt", "Shell-crm renovates") spin up a fresh Crow session per run, do their work, then **linger in `.active` forever** — a human has to manually `set-status … completed`.

## Change

`JobScheduler` now watches each run it fires and, on its existing 30s tick, auto-completes the session once the run reaches a **successful** terminal end. A run completes when:

- every prompt has been delivered, **plus** a 20s settle window — guards against catching a stale top-level `.done` from an earlier prompt in a multi-prompt job (prompts are spaced 20s apart);
- the terminal has reached `.agentLaunched`; and
- the agent is at rest — `.done` (Claude) or `.idle` (TUI agents) — rather than still `.working` or `.waiting`.

`.waiting` covers both **awaiting-input** and **errored** (`StopFailure`), so those runs are left `.active` to surface rather than being silently completed. A 12h safety cap stops watching a blocked run so it doesn't linger in memory.

The check keys purely on `AgentActivityState`, so it behaves the same across **Claude / Codex / Cursor / OpenCode**. Completion goes through the existing `SessionService.completeSession`, which already guards manager sessions; jobs are `.job` kind and are never watched as managers regardless.

## Acceptance criteria

- [x] A job that finishes successfully transitions to `.completed` with no manual action.
- [x] Errored / awaiting-input runs (`.waiting`) are **not** auto-completed — they stay active and surface.
- [x] Consistent across agent kinds via the shared `AgentActivityState` hook pipeline.
- [x] Manager sessions unaffected.

## Implementation notes

- Entirely contained in `Sources/Crow/App/JobScheduler.swift` — no changes to the shared hook-event handler, `SessionService`, config, or CLI.
- The decision is factored into a pure `finishDecision(...)` function, unit-tested in `Tests/CrowTests/JobFinishDecisionTests.swift` (10 cases covering done/idle → complete, working/waiting → keep-waiting, undelivered/in-settle-window/pre-launch → keep-waiting, session-gone/inactive/timeout → stop-watching).

## Testing

- `make build` — clean.
- `swift test --filter JobFinishDecisionTests` — 10/10 pass.
- Existing `JobRepoResolutionTests` — still green.

https://claude.ai/code/session_01WCdsrShB5Uo18qoH8akVu2

### DIFF
--- a/Sources/Crow/App/JobScheduler.swift
+++ b/Sources/Crow/App/JobScheduler.swift
@@ -28,9 +28,30 @@ final class JobScheduler {
     /// follow-up prompts for a run.
     private let maxLaunchWaitPolls = 60
 
+    /// Grace period after the final prompt is delivered before a run is eligible
+    /// to auto-complete. Prevents catching a stale top-level `.done` from an
+    /// earlier prompt before the agent has picked up the last one (CROW-561).
+    private let finishSettleDelay: TimeInterval = 20
+    /// Safety cap: stop watching a run for completion after this long so a
+    /// blocked/erroring run doesn't linger in memory forever (CROW-561).
+    private let maxWatchDuration: TimeInterval = 12 * 3600
+
     /// Jobs currently being created — guards against a long worktree creation
     /// double-firing on the next tick before `lastRunAt` is persisted.
     private var inFlight: Set<UUID> = []
+
+    /// A job run being watched so its session auto-completes once the agent
+    /// finishes successfully (CROW-561).
+    private struct RunWatch {
+        let terminalID: UUID
+        let startedAt: Date
+        /// Set once the *last* prompt has been delivered; until then the run is
+        /// not yet eligible to be judged finished.
+        var promptsDeliveredAt: Date?
+    }
+
+    /// Active job runs, keyed by session id, awaiting successful-finish detection.
+    private var watchedRuns: [UUID: RunWatch] = [:]
 
     /// Reads the current job list (from AppDelegate's `appConfig`).
     var jobsProvider: () -> [JobConfig] = { [] }
@@ -61,8 +82,12 @@ final class JobScheduler {
     // MARK: - Tick
 
     private func tick() {
-        guard let devRoot = devRootProvider() else { return }
         let now = Date()
+        // Auto-complete any job runs that have finished successfully. Runs on
+        // every tick regardless of whether jobs are due (CROW-561).
+        checkFinishedRuns(now: now)
+
+        guard let devRoot = devRootProvider() else { return }
         for job in jobsProvider() where job.enabled {
             guard !inFlight.contains(job.id) else { continue }
             let baseline = job.lastRunAt ?? job.createdAt
@@ -92,7 +117,15 @@ final class JobScheduler {
             if let result = await self.sessionService.runJob(captured, devRoot: devRoot) {
                 // Persist run time first so the job isn't re-fired next tick.
                 self.onJobRan(captured.id, Date())
-                self.deliverRemainingPrompts(captured, terminalID: result.terminalID)
+                // Watch this run so its session auto-completes on success (CROW-561).
+                self.watchedRuns[result.sessionID] = RunWatch(
+                    terminalID: result.terminalID,
+                    startedAt: Date(),
+                    promptsDeliveredAt: nil
+                )
+                self.deliverRemainingPrompts(
+                    captured, sessionID: result.sessionID, terminalID: result.terminalID
+                )
             }
         }
     }
@@ -101,15 +134,22 @@ final class JobScheduler {
 
     /// Deliver every prompt after the first non-empty one, once Claude has
     /// launched, spaced by `promptGap`.
-    private func deliverRemainingPrompts(_ job: JobConfig, terminalID: UUID) {
+    private func deliverRemainingPrompts(_ job: JobConfig, sessionID: UUID, terminalID: UUID) {
         let remaining = job.prompts
             .map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
             .filter { !$0.isEmpty }
             .dropFirst()
-        guard !remaining.isEmpty else { return }
+        guard !remaining.isEmpty else {
+            // Single-prompt job: the sole prompt launches with the agent, so the
+            // run is fully delivered as soon as the terminal reports launched.
+            waitForAgentLaunched(terminalID: terminalID, polls: 0) { [weak self] in
+                self?.markPromptsDelivered(sessionID: sessionID)
+            }
+            return
+        }
 
         waitForAgentLaunched(terminalID: terminalID, polls: 0) { [weak self] in
-            self?.sendSequentially(Array(remaining), terminalID: terminalID)
+            self?.sendSequentially(Array(remaining), sessionID: sessionID, terminalID: terminalID)
         }
     }
 
@@ -132,8 +172,12 @@ final class JobScheduler {
     /// Send prompts one at a time, waiting `promptGap` before each so they don't
     /// collide with Claude still processing the previous one. Stops if the
     /// terminal disappears (e.g. the session was deleted).
-    private func sendSequentially(_ prompts: [String], terminalID: UUID) {
-        guard !prompts.isEmpty else { return }
+    private func sendSequentially(_ prompts: [String], sessionID: UUID, terminalID: UUID) {
+        guard !prompts.isEmpty else {
+            // Every prompt has been sent — the run is now eligible to finish.
+            markPromptsDelivered(sessionID: sessionID)
+            return
+        }
         DispatchQueue.main.asyncAfter(deadline: .now() + promptGap) { [weak self] in
             guard let self else { return }
             guard let terminal = self.appState.terminals.values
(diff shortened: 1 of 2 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Focused, well-factored change: the completion decision is isolated into a pure `finishDecision(...)` and unit-tested, the `Task`/`asyncAfter` callbacks all use `[weak self]`, `checkFinishedRuns` relies on Swift dictionary COW to safely delete keys mid-iteration, and everything stays `@MainActor`-serialized. Build is clean via `make build` and `JobFinishDecisionTests` passes 10/10 locally. No security surface (no auth, input parsing, or secrets touched).

One correctness issue blocks approval.

### Critical Issues

**🔴 `.idle` is treated as "finished," but it is also the never-started state — TUI agents (OpenCode/Cursor) can be auto-completed while still active.**
`Sources/Crow/App/JobScheduler.swift:264-271`

`finishDecision` returns `.complete` whenever the agent is `.idle` or `.done` after the settle window. For Claude this is safe because `UserPromptSubmit` flips the agent to `.working` the instant the first prompt is submitted, so a launched-but-idle Claude has effectively always begun. But for the TUI agents this PR explicitly claims parity with, that invariant does **not** hold:

- `Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeSignalSource.swift:11-12,52` — the OpenCode plugin emits **no** `UserPromptSubmit`; the state only leaves `.idle` on the first `PreToolUse` (`→ .working`).
- A fresh job session's `activityState` defaults to `.idle` (`Packages/CrowCore/.../AppState.swift`), and `activityState` is *only* ever set from hook events — there is no terminal-output fallback.

Failure scenario (single-prompt OpenCode job):
1. Session created, `activityState == .idle`. Terminal reaches `.agentLaunched`; the first prompt is delivered → `markPromptsDelivered` stamps `promptsDeliveredAt`.
2. OpenCode spends >20s reasoning before its first tool call — or the task is answer-only and makes **no** tool call at all — so no `PreToolUse` fires and the state is still the default `.idle`.
3. Next 30s tick: session active, past the 20s settle, readiness `.agentLaunched`, `activityState == .idle` → `.complete`. The session is silently marked `.completed` while the agent is still working (or before it has even begun).

This is exactly the "silently completed" outcome the PR says it avoids, and it undercuts the "Consistent across agent kinds" acceptance criterion. Suggested fix: require the run to have been *observed leaving idle at least once* before `.idle`/`.done` counts as finished — e.g. track a `sawActivity` flag on `RunWatch` set when `activityState` is ever `.working`/`.waiting`/`.done`, and only allow `.complete` once it's true. That distinguishes "at rest because finished" from "at rest because not started."

### Security Review
**Strengths:**
- No new external input, auth, crypto, or secret handling. Completion routes through the existing `SessionService.completeSession`, which already guards manager sessions; job kind is `.job` and never watched as a manager.

**Concerns:** none.

### Code Quality
- **🟢 Aborted multi-prompt delivery leaves a run watched until the 12h cap.** `sendSequentially` (`JobScheduler.swift:181-188`) returns without calling `markPromptsDelivered` if the terminal disappears mid-delivery, so `promptsDeliveredAt` stays `nil` and the watch lingers (harmlessly) for `maxWatchDuration`. Consider dropping the watch on that abort path. Consider-only.
- **🟢** A bare `swift build` fails until `scripts/generate-build-info.sh` runs (missing generated `BuildInfo`); this is pre-existing scaffolding, not from this PR — `make build` handles it. Noted only so future reviewers don't chase it.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 0 Yellow, 2 Green] findings.

---

dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues

**🔴 `.idle` is ambiguous — runs whose agent never ran (or never reported) are auto-completed as "success"**

`finishDecision` treats `.idle` as a successful finish (`Sources/Crow/App/JobScheduler.swift:269-270`), but `.idle` is also the **default** `activityState` of a fresh `SessionHookState` that has never received a hook event (`Packages/CrowCore/Sources/CrowCore/AppState.swift:664`). Combined with the fact that `readiness == .agentLaunched` means "launch command *sent*" — not "agent confirmed running" — this completes runs the feature explicitly promises to leave active:

1. **Deterministic — failed launches get buried as completed.** `launchAgent` has two failure paths that *park* readiness at `.agentLaunched` without launching any agent: agent can't build a launch command for the session kind (`Sources/Crow/App/SessionService.swift:606`, #424) and initial prompt file missing (`Sources/Crow/App/SessionService.swift:624`, CROW-439). For a job run, `waitForAgentLaunched` then fires `markPromptsDelivered`, no hook event ever arrives (state stays `.idle`), and the first tick ≥ 20 s later returns `.complete`. Before this PR those sessions stayed `.active` with the error echoed in the terminal; now they're moved to Completed within ~a minute, hiding the failure — the opposite of acceptance criterion 2.
2. **Race — agent still booting or unmonitored.** For a single-prompt job, `promptsDeliveredAt` is set the moment the launch command goes out (`Sources/Crow/App/SessionService.swift:639`). If the agent takes longer than the 20 s settle + tick to emit its first hook event (slow startup, folder-trust dialog, MCP servers), or the hook-config write failed (logged-and-continue at `Sources/Crow/App/SessionService.swift:554-557`, so **no** events ever arrive), the run is completed while the agent is booting or actively working.

**Suggested fix:** require evidence the agent actually ran before accepting `.idle` as "finished" — e.g. pass `hasHookEvents: !hookState.hookEvents.isEmpty` (or `lastTopLevelStopAt != nil`) into `finishDecision` and return `.keepWaiting` for `.idle` without it; alternatively track "observed `.working` at least once" on `RunWatch`. `.done` can stay as-is since only a real Stop event produces it. Please also add a test case for "agent never emitted a hook event."

### Security Review

**Strengths:**
- No new input surface, privileges, or credential handling; the change is scheduler-internal state observation.
- Completion goes through the existing `SessionService.completeSession` → `updateSessionStatus`, which keeps the manager-session guard (`Sources/Crow/App/SessionService.swift:2150`).
- Errored Claude runs (`StopFailure` → `.waiting`) are correctly left active rather than silently swept away, and `.waiting` also covers permission prompts — an unattended job blocked on a permission decision won't be falsely completed.

**Concerns:**
- 🟡 **Errored OpenCode/Cursor runs complete as success.** `OpenCodeSignalSource` maps `Notification` — which per its own comment includes `session.error` — to `.done` whenever no top-level Stop was recorded (`Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeSignalSource.swift:61-68`); `CursorSignalSource` has the same headless safety net. So the PR's "errored runs stay `.active` and surface" guarantee only holds for Claude's `StopFailure`. This mapping predates the PR, but auto-complete escalates its consequence from "wrong badge color" to "errored run silently marked completed." At minimum note the limitation; ideally distinguish error notifications from the headless-done safety net before treating `.done` as success for those agents.

### Code Quality

- Factoring the decision into pure, dependency-free `finishDecision(...)` is the right shape — the 10 unit tests cover the state matrix well, build is clean, and `JobRepoResolutionTests` still pass (verified locally).
- 🟢 `watchedRuns` is in-memory only: an app relaunch mid-run reverts those runs to the old linger-forever behavior. Acceptable degradation, but consider re-deriving watches for active `.job`-kind sessions at startup, or noting it in the doc comment.
- 🟢 When `waitForAgentLaunched` gives up after its 5-minute poll budget, the watch entry stays (with `promptsDeliveredAt == nil`) until the 12 h cap or a status change. Bounded, but you could remove the watch in the give-up branch for tidiness.
- 🟢 The "Snapshot so we can mutate `watchedRuns` while iterating" comment is slightly misleading — Swift's `for (k, v) in dict` already iterates a value copy; no explicit snapshot exists or is needed. Reword or drop.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 1 Yellow, 3 Green] findings.

---

dgershman:
Thanks @dhilgaertner — addressed in 80d7478.

**🔴 `.idle` treated as finished (blocking):** Fixed. Confirmed across all four signal sources (`ClaudeHookSignalSource`, `CodexSignalSource`, `OpenCodeSignalSource`, `CursorSignalSource`) that `.idle` is set *only* on a fresh non-resume `SessionStart` — it's the never-started default, never a resting-after-work state. My earlier assumption that TUI agents "rest at `.idle`" was wrong: OpenCode/Cursor finish via `Stop` or the `Notification` safety-net → `.done`. So `finishDecision` now returns `.complete` **only for `.done`**; `.idle` joins `.working`/`.waiting` as `.keepWaiting`. This directly fixes all three failure modes you raised (failed launches parked at `.agentLaunched` with no hook event, still-booting agents, and answer-only/slow-reasoning TUI runs) without regressing any agent kind, since every kind reaches `.done` on a real finish. Added a test asserting the "agent never emitted a hook event" (failed-launch) case stays `.keepWaiting`.

**🟡 OpenCode/Cursor error `Notification` → `.done`:** Noted as a documented limitation in the `finishDecision` doc comment. A proper fix (distinguishing `session.error` from the headless-done safety net) belongs in the signal sources and is out of scope for this PR; Claude/Codex errored runs (`StopFailure` → `.waiting`) remain correctly left active.

**🟢 Aborted multi-prompt delivery:** now drops the watch in the terminal-gone branch instead of lingering to the 12h cap.
**🟢 In-memory watch lost on relaunch:** documented on `watchedRuns`.
**🟢 Misleading "snapshot" comment:** reworded to note it's copy-on-write, not an explicit copy.
**🟢 `swift build` / BuildInfo:** pre-existing scaffolding — `make build` used throughout.

Tests: 11/11 in `JobFinishDecisionTests`, `JobRepoResolutionTests` still green, `make build` clean.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface: the change is confined to `JobScheduler`, keys purely on in-memory `AgentActivityState`/`TerminalReadiness`, and drives completion through the existing `SessionService.completeSession` → `updateSessionStatus`, which already guards manager sessions (`SessionService.swift:2150`). Verified.
- No credentials, external input, or injection paths touched. Prompt delivery to the terminal is unchanged in substance.

### Code Quality
Verified against the surrounding code and confirmed correct:
- **`.done`-only gate is right.** Claude/Codex map `Stop → .done` and `StopFailure → .waiting` (`ClaudeHookSignalSource.swift:92/97`, `CodexSignalSource.swift:61/75`), so errored runs correctly stay active. `.idle` is only ever the fresh-`SessionStart` default, so failed launches (readiness parks at `.agentLaunched`, no hook fires, state stays `.idle`) are no longer silently completed — the round-1 fix holds, and `doesNotCompleteWhenAgentNeverEmittedAHookEvent` locks it in.
- **`SessionStart` resume → `.done` cannot cause a false completion:** `watchedRuns` is in-memory and dropped on relaunch, so resumed/adopted sessions are never watched. Good.
- **Mutation-during-iteration is safe** — Dictionary value-copy + COW; the removals apply to the property while iteration proceeds over the original snapshot.
- **Bounded lifecycle:** undelivered prompts / stuck launches never set `promptsDeliveredAt`, so they can only `keepWaiting` until the 12h `maxWatchDuration` cap → `stopWatching`. No unbounded memory growth.
- **`finishDecision` is a clean pure function**; the 11-case suite (`JobFinishDecisionTests`) covers done/idle/working/waiting, settle window, launch gate, session-gone/inactive, and timeout.
- Moving `checkFinishedRuns` above the `devRoot` guard so completion runs even when `devRoot` is transiently nil is a nice touch.

Build clean (`swift build`), `JobFinishDecisionTests` 11/11 pass locally.

- The settle window is a fixed 20s timer from the last `TerminalRouter.send`. In the narrow case where the agent is still busy on the prior prompt when the window elapses, a stale `.done` is theoretically readable. It's an explicitly documented heuristic and `promptGap == finishSettleDelay` makes it generous; a future refinement could gate on observing a `.working` transition after the final prompt rather than a fixed timer.
- OpenCode/Cursor map an error `Notification` to `.done`, so an errored TUI run can complete as "success". Honestly documented in the `finishDecision` doc comment as a known limitation.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. The round-1 `.idle` concern is resolved, well-tested, and clearly documented.

PR #567, opened 2026-07-02
Wire Manager process-exit banner via tmux exit monitor (#558)

### LINKED ISSUE #558: Wire Manager process-exit banner for shared xterm.js attach client
## Context

PR #556 replaced GhosttyKit with a shared `XTermSurfaceView` that runs `tmux attach-session` as its PTY child. The old Ghostty path wired `GhosttyApp.onChildExited` to set `appState.managerProcessExited`, which drives the "Manager process exited — Restart" banner in `SessionDetailView`.

That callback is no longer connected, so `managerProcessExited` never becomes `true` and the restart banner never appears.

## Goal

Detect when the Manager's tmux pane/shell process exits (distinct from the shared attach client exiting) and surface the restart banner again.

## Acceptance

- [ ] When the Manager session's shell/agent process exits, `appState.managerProcessExited` becomes `true` and the restart banner appears.
- [ ] Restarting the Manager clears the flag (existing `restartManager` path).
- [ ] Attach-client teardown (`XTermSurfaceView.destroy`) does not false-positive the banner.

## References

- #556 / ADR 0006 (known limitation)
- `AppState.managerProcessExited`, `SessionDetailView`, `XTermSurfaceView.onProcessExit`, `TmuxBackend.cockpitSurface()`

dgershman:
Closes #558

## Problem

PR #556 replaced GhosttyKit with one shared `XTermSurfaceView` running a single `tmux attach-session` client. That dropped the old `GhosttyApp.onChildExited` seam that set `AppState.managerProcessExited`, so nothing ever set the flag `true` again — the **"Manager process exited — Restart" banner became dead code**.

The old callback can't be reconnected as-is: there's now one shared attach client with no per-window identity, and `XTermSurfaceView.onProcessExit` only fires when the *whole* attach client dies.

## Key insight

Under tmux, every terminal is a window inside one shared session, and the Manager's agent (`claude …`) runs **inside the window's shell wrapper**, not as the pane's direct child. So when the agent exits it does **not** kill the pane — the window persists and its foreground command falls back to a bare login shell (the exact state the orphan-window reaper already recognizes). `#{pane_dead}`/"window gone" therefore never fire on a normal agent exit.

## Approach

- **`TmuxBackend`** gains a poll monitor (`startManagerExitMonitor` / `stopManagerExitMonitor`) that reads the Manager window's `#{pane_current_command}` every 3s and fires once it observes the agent *run* and then return to a bare login shell. A pure, unit-tested classifier (`managerAgentDidExit`) encodes the transition policy and reuses `orphanLoginShells`. Inconclusive samples (missing binding on async launch, or a thrown `display-message` when the window is gone) are skipped, so attach-client teardown and restart/shutdown never false-positive. Stops after firing; re-armed on the next Manager launch.
- **`SessionService`** arms the monitor from `ensureManagerSession` (covers fresh launch, hydrate/adopt relaunch, and `restartManager`) and after `restartTmuxServer`; stops it at the start of `restartManager`. The existing restart path still clears the flag at `SessionService.swift:735`.
- **`AppDelegate`** stale `#558` breadcrumb comments replaced.

## Acceptance

- [x] Manager agent exits → `appState.managerProcessExited` becomes `true`, banner appears.
- [x] Restarting the Manager clears the flag (existing `restartManager` path) and re-arms the monitor.
- [x] Attach-client teardown (`XTermSurfaceView.destroy`) does not false-positive (monitor reads tmux window state, not the attach client; `nil`/window-gone samples never fire).

## Testing

- `make build` — clean.
- `swift test --package-path Packages/CrowTerminal --filter "ManagerExitMonitorTests|OrphanWindowReaperTests"` — 7/7 green (4 new `ManagerExitMonitorTests` covering: fires after run→shell; no fire before agent seen; no fire while running; no fire when window gone).

https://claude.ai/code/session_01WfYBT6dvEatoRus4xuNnhu

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -78,6 +78,12 @@ public final class TmuxBackend {
     /// `onReadinessChanged` for a tab the user just closed. Issue #282.
     private var readinessTasks: [UUID: [Task<Void, Never>]] = [:]
 
+    /// Poll loop watching the Manager window's foreground command so the
+    /// "Manager process exited" banner reappears under the shared xterm.js
+    /// attach client (#558). At most one runs at a time; re-armed by
+    /// `SessionService` on launch / restart. See `startManagerExitMonitor`.
+    private var managerExitMonitor: Task<Void, Never>?
+
     /// Public for test isolation. Production callers use `.shared`.
     public init() {}
 
@@ -143,6 +149,9 @@ public final class TmuxBackend {
         // cleanup (#282).
         for tasks in readinessTasks.values { tasks.forEach { $0.cancel() } }
         readinessTasks.removeAll()
+        // Stop the Manager exit poll too — its window is gone with the server
+        // (#558). `SessionService` re-arms it after `rebuildAllSurfaces`.
+        stopManagerExitMonitor()
         // Only unlink the per-terminal scratch files when we're actually
         // killing the server. On a clean quit that leaves the server running
         // they must survive so the next launch's `adoptTerminal` can detect
@@ -575,6 +584,20 @@ public final class TmuxBackend {
         return orphanLoginShells.contains(command)
     }
 
+    /// Decide whether the Manager agent has exited, from one poll sample of its
+    /// window's foreground command (#558). Because the agent (`claude …`) runs
+    /// *inside* the window's shell wrapper rather than as the pane's direct
+    /// child, its exit doesn't kill the pane — the foreground just falls back to
+    /// a bare login shell. So we report an exit only once we've seen the agent
+    /// actually running (`sawAgentRunning`, a non-shell foreground) and now see
+    /// a bare login shell. `nil` (window gone / read failed) is never an exit —
+    /// that path also covers teardown, keeping restart/shutdown false-positive
+    /// free. Pure so the transition policy is unit-testable without tmux.
+    nonisolated static func managerAgentDidExit(paneCommand: String?, sawAgentRunning: Bool) -> Bool {
+        guard sawAgentRunning, let command = paneCommand else { return false }
+        return orphanLoginShells.contains(command)
+    }
+
     /// Reap cockpit windows that no live terminal references AND that are
     /// sitting at a bare login shell — leaked windows from a timed-out
     /// `new-window` or a forgotten terminal (#408). `keepWindowIndices` is the
@@ -602,6 +625,66 @@ public final class TmuxBackend {
             NSLog("[Crow] Reaped \(reaped) orphaned bare-shell cockpit window(s) (#408)")
         }
         return reaped
+    }
+
+    // MARK: - Manager exit monitor (#558)
+
+    /// Watch the Manager terminal's tmux window and fire `onExit` the first time
+    /// its foreground command falls back to a bare login shell after the agent
+    /// was seen running — i.e. the Manager's `claude`/`codex`/… process exited.
+    ///
+    /// Under the shared xterm.js cockpit, `XTermSurfaceView.onProcessExit` only
+    /// fires when the whole `tmux attach-session` client dies, so it can't tell
+    /// a per-window agent exit apart (#558). We poll `#{pane_current_command}`
+    /// for the Manager window instead — the same signal the orphan reaper reads.
+    ///
+    /// At most one monitor runs; a second call cancels the first. The poll skips
+    /// samples where the binding is absent (async adopt on launch hasn't landed
+    /// yet) or the `display-message` throws (window gone / server down), so
+    /// attach-client teardown and restart/shutdown never false-positive. Stops
+    /// after firing once; `SessionService` re-arms it on the next Manager launch.
+    public func startManagerExitMonitor(
+        id: UUID,
+        pollInterval: TimeInterval = 3.0,
+        onExit: @escaping @MainActor () -> Void
+    ) {
+        stopManagerExitMonitor()
+        let nanos = UInt64(pollInterval * 1_000_000_000)
+        managerExitMonitor = Task { [weak self] in
+            var sawAgentRunning = false
+            while !Task.isCancelled {
+                try? await Task.sleep(nanoseconds: nanos)
+                if Task.isCancelled { return }
+                guard let self else { return }
+                // Read the Manager window's foreground command. A missing
+                // binding or a thrown read is inconclusive (not an exit): skip.
+                guard let windowIndex = self.bindings[id],
+                      let ctrl = self.controller,
+                      let raw = try? ctrl.displayMessage(
+                          target: "\(ctrl.sessionName):\(windowIndex)",
+                          format: "#{pane_current_command}"
+                      )
+                else { continue }
+                let command = raw.trimmingCharacters(in: .whitespacesAndNewlines)
+                if Self.orphanLoginShells.contains(command) {
+                    if Self.managerAgentDidExit(paneCommand: command, sawAgentRunning: sawAgentRunning) {
+                        NSLog("[CrowTelemetry manager:exit_detected terminal=\(id) command=\(command)]")
+                        onExit()
+                        return
+                    }
+                    // Bare shell before the agent ever launched (startup window);
+                    // keep watching until we see it running.
+                } else if !command.isEmpty {
+                    sawAgentRunning = true
+                }
+            }
+        }
+    }
+
+    /// Cancel the Manager exit monitor if one is running. Idempotent.
+    public func stopManagerExitMonitor() {
+        managerExitMonitor?.cancel()
+        managerExitMonitor = nil
     }
 
     /// Return the shared cockpit xterm surface, lazily creating it the
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -598,6 +598,24 @@ public final class TmuxBackend {
         return orphanLoginShells.contains(command)
     }
 
+    /// Advance the exit-monitor state machine by one poll sample, so the whole
+    /// transition — not just its terminal condition — is unit-testable without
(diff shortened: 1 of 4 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Solid, well-reasoned fix for the dead-banner regression from #556. The "poll the Manager window's foreground command and treat run→bare-shell as an exit" approach correctly matches tmux semantics (agent runs inside the window's shell wrapper, so its exit doesn't kill the pane), the pure `managerAgentDidExit` classifier is cleanly unit-tested, and the arm/stop lifecycle across `ensureManagerSession` / `restartManager` / `restartTmuxServer` / `shutdown` is carefully covered. Tests pass locally (7/7).

One should-fix concern.

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface: reads `#{pane_current_command}` for a fixed, app-owned window; no user-controlled input flows into the tmux target or format string.
- `nil`/thrown samples are treated as inconclusive, so teardown/shutdown never false-positive.

**Concerns:** None.

### Code Quality

**Yellow — the exit monitor polls tmux on the main thread every 3s for the app's whole lifetime.**
`TmuxBackend` is `@MainActor` (`TmuxBackend.swift:30`), so the `Task {}` in `startManagerExitMonitor` (`TmuxBackend.swift:635`) inherits MainActor isolation — confirmed by the synchronous `onExit()` call (a `@MainActor` closure) with no `await`. That means the per-tick `ctrl.displayMessage(...)` → `TmuxController.run` → `p.waitUntilExit()` (`TmuxController.swift:51`) blocks the **main thread** for the subprocess duration (~70ms p95 per the spike, up to the 2s watchdog timeout on a tmux hang), recurring every 3s indefinitely.

Existing tmux calls are on the main actor too, but they're event-driven / one-shot (the #408 reaper runs once at launch); a *perpetual background poll* on the UI thread is a new, avoidable cost — periodic frame hitches, and a worst-case ~2s UI freeze per tick if tmux wedges. `TmuxController` is `Sendable`, so the read can run off the main actor and hop back only to fire `onExit`, e.g.:
```swift
let sample = await Task.detached { try? ctrl.displayMessage(target: ..., format: "#{pane_current_command}") }.value
```
(capture `ctrl`/`windowIndex` on the main actor first). This keeps the poll from ever touching the UI thread. `TmuxBackend.swift:635`

**Green — the poll-loop state machine is untested.** Only the pure `managerAgentDidExit` classifier has coverage. The actual transition logic — `sawAgentRunning` being set by `else if !command.isEmpty` and the missing-binding/thrown-read skips (`TmuxBackend.swift:657-680`) — is where a regression would most likely hide and it's exercised by no test. Extracting the per-sample step into a pure `(prevSawAgent, sample) -> (nextSawAgent, fired)` reducer would make it as testable as the classifier already is.

**Green — sub-poll-interval crashes are missed.** An agent that launches and exits within one 3s window is never sampled in its running state, so `sawAgentRunning` stays false and the banner never fires. This is a missed-detection (not a false positive) and inherent to sampling; worth a one-line comment acknowledging the floor.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The Yellow (main-thread poll) is the only blocker; moving the subprocess off the MainActor resolves it.

---

dgershman:
Thanks for the review — addressed all three findings in 4668027:

- **Yellow (main-thread poll):** the `Task` still inherits MainActor isolation for the arm/fire lifecycle, but the blocking `display-message` now runs in a `Task.detached` — I capture `controller` + `windowIndex` on the main actor, read `#{pane_current_command}` off it (`TmuxController` is a `Sendable` struct), and hop back to MainActor only for the `onExit` call. The UI thread is never blocked by the subprocess, even on a tmux hang.
- **Green (untested state machine):** extracted the per-sample transition into a pure `advanceExitMonitor(sawAgentRunning:sample:)` reducer that the loop now calls, with 4 new tests covering the `sawAgentRunning` latch, the run→shell fire, the startup-shell no-op, and the `nil`/`""` inconclusive-sample skips.
- **Green (sub-poll-interval floor):** documented in the `startManagerExitMonitor` doc comment — an agent that launches and exits inside one poll interval is never observed running, so no banner fires (missed detection, not a false positive).

Tests: 8/8 green in `ManagerExitMonitorTests`; full `make build` clean.

dhilgaertner (changes requested):
## Code & Security Review

Focused, well-scoped fix that revives the dead "Manager process exited" banner (#556 regression) by polling the Manager window's `#{pane_current_command}` instead of relying on the now-shared attach client's process-exit signal. The core logic (`managerAgentDidExit` / `advanceExitMonitor`) is pure and unit-tested (8/8 green locally), and the concurrency is correct: `TmuxBackend` is `@MainActor`, state is captured on the main actor, and the blocking `displayMessage` runs off it via `Task.detached` on a `Sendable` `TmuxController`.

### Critical Issues
None.

### Security Review
**Strengths:**
- No new external input, network, auth, or credential surface. The poll reads a local tmux format string with a fixed, non-user-controlled format (`#{pane_current_command}`) against a target built from an internal session name + window index — no injection vector.
- Blocking subprocess kept off the UI thread, so a wedged tmux can't stall AppKit; the perpetual poll checks `Task.isCancelled` around every await and stops after firing once.

**Concerns:** None.

### Code Quality

**Yellow — `restartTmuxServer` leaves a stale "Manager process exited" banner** (`Sources/Crow/App/SessionService.swift:775`-`796`)
`restartManager` clears the flag (`appState.managerProcessExited = false`, line 760) before re-arming. `restartTmuxServer` kills the server and relaunches every terminal's agent — including a fresh Manager `claude` — via `rebuildAllSurfaces(forceRegister: true)`, then re-arms the monitor at line 789, but never clears `managerProcessExited`. So if the exit banner is already showing when the user picks the "Restart tmux Server" menu item, a healthy Manager agent comes back up while the "Manager process exited — Restart" banner stays on screen (the arm callback's `!managerProcessExited` guard only prevents re-firing, it doesn't clear). This stale-banner state is newly reachable because before this PR the flag was dead code. One-line fix: clear the flag in `restartTmuxServer` alongside the re-arm, mirroring `restartManager`.

**Green — consider:**
- The 3s poll spawns a `tmux display-message` subprocess for the app's whole lifetime once the Manager agent is running and never exits. It mirrors the existing orphan-reaper cadence, so likely acceptable, but worth confirming it's not stacking with other pollers.
- False-positive surface if the Manager agent ever hands the pane's tty foreground to a bare shell while still alive. The design reuses the proven `orphanLoginShells` set and the PR documents the accepted missed-detection floor; noting it only so the tradeoff is explicit.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The Yellow is a one-line clear in `restartTmuxServer`; addressing it lands in the same round trip.

---

dgershman:
Good catch — fixed in 4205b44. `restartTmuxServer` now clears `appState.managerProcessExited = false` alongside the re-arm, mirroring `restartManager`, so a banner showing when the user picks "Restart tmux Server" clears once the fresh Manager agent comes back up (the arm callback's `!managerProcessExited` guard only blocks re-firing, it can't clear a stale flag).

The two Greens are accepted, documented tradeoffs (poll cadence mirrors the orphan reaper; the bare-shell false-positive surface reuses the proven `orphanLoginShells` set with the missed-detection floor noted in the doc comment) — no code change.

`make build` clean.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new external input, network, or credential surface. The monitor reads a tmux format string (`#{pane_current_command}`) against a fixed target the app already controls; nothing user-controlled is interpolated into the tmux command.
- The `NSLog` telemetry line echoes the sampled command name only (`zsh`/`node`/…), not secrets or paths.

**Concerns:** None.

### Code Quality
Verified against the checkout (build clean, `ManagerExitMonitorTests|OrphanWindowReaperTests` 11/11 green):

- **Concurrency is handled carefully.** `TmuxBackend` is `@MainActor`; the poll `Task` inherits main-actor isolation but pushes the blocking `displayMessage` (`TmuxController.run` → `waitUntilExit()`) into a `Task.detached`, so a wedged tmux subprocess can't stall AppKit. `TmuxController` is a `Sendable` struct, so the value capture into the detached task is race-free. `Task.isCancelled` is re-checked *after* the detached read and *before* `onExit()` fires (`TmuxBackend.swift:701`), so a `restartManager` that cancels mid-poll can't produce a stale-window false positive.
- **Flag lifecycle is complete.** `managerProcessExited` lives only in in-memory `AppState` (default `false`, not in the persisted `store` model), so a hydrate/adopt relaunch starts clean — no stale banner. Both restart paths clear + re-arm: `restartManager` (stop → clear at `SessionService.swift:760` → `ensureManagerSession` re-arms) and `restartTmuxServer` (explicit clear + `armManagerExitMonitor`, with a comment correctly noting the `!managerProcessExited` guard can't clear a stale flag).
- **Fires once, re-armed on demand.** The monitor `return`s after firing and `startManagerExitMonitor` cancels any prior task, so at most one runs.
- **Pure, well-tested transition policy.** `managerAgentDidExit` / `advanceExitMonitor` encode the run→shell latch and are unit-tested including the inconclusive (`nil`/`""`) no-op cases that keep attach-client teardown false-positive free.

### Considerations (non-blocking, Green)
- **Suspend edge case:** if the Manager agent were job-control-suspended (Ctrl-Z), its foreground would transiently fall back to a login shell and the monitor would fire an "exited" banner. This is an inherent limitation of the polling heuristic (the same `orphanLoginShells` signal the reaper uses), unlikely in Manager usage, and recoverable via Restart — noting only for the record.
- **Minor redundancy:** in `advanceExitMonitor`, the login-shell branch calls `managerAgentDidExit(paneCommand: command, sawAgentRunning:)` where `orphanLoginShells.contains(command)` is already known true, so it reduces to `return (sawAgentRunning, sawAgentRunning)`. Harmless; current form keeps the single source of truth for the policy.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. Careful concurrency, complete flag lifecycle across all launch/restart paths, and green tests.

PR #568, opened 2026-07-02
Fix done-in-24h badge cap: use search issueCount instead of counting capped nodes (closes #562)

### LINKED ISSUE #562: Sidebar 'done in last 24h' ticket badge caps at 50 (closedIssues search hardcoded first: 50, counts paginated nodes)
## Summary

The **"Tickets"** header row in the right-hand sidebar (the row of status-count badges at the top) has a green ✓ badge showing *issues completed in the last 24 hours*. That number **saturates at 50** — it can never show more, even when I've actually closed 51+ tickets in the window. I've been closing well over 50/day, so the badge is under-reporting.

## Root cause (grounded)

The badge is a raw `.count` of a GitHub GraphQL `search` result that is hard-capped at `first: 50`, so the count physically can't exceed 50.

Data flow:

1. **Fetch is capped at 50** — `Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubTaskBackend.swift:323` (and the no-projects fallback at `:347`):
   ```graphql
   closedIssues: search(type: ISSUE, query: $closedQuery, first: 50) {
     nodes { ... on Issue { number title url state updatedAt ... } }
   }
   ```
   Note the asymmetry: `openIssues` immediately above uses `first: 100` (`:307`, `:338`); only `closedIssues` is pinned at 50.

2. **Badge = `.count` of that capped list** — `Sources/Crow/App/IssueTracker.swift:349`:
   ```swift
   appState.doneIssuesLast24h = ghResult.closedIssues.count
   ```
   Since `closedIssues` is at most 50 nodes, `doneIssuesLast24h` tops out at 50.

3. **Render** — `Packages/CrowUI/Sources/CrowUI/TicketBoardView.swift:620`:
   ```swift
   StatusCount(icon: "checkmark.circle.fill", color: .green, count: appState.doneIssuesLast24h)
   ```

(The hard ceiling in code is exactly **50**; a "never higher than ~51" observation is that same ceiling.)

## Recommended fix

GitHub's GraphQL `search` connection returns **`issueCount`** — the *total* number of matches, independent of the `first:` page size. So the count badge should read `issueCount` rather than counting the paginated `nodes`, which fixes it with **zero extra API cost / no pagination**:

1. In both queries (`GitHubTaskBackend.swift:323` and `:347`), add `issueCount` to the `closedIssues` search:
   ```graphql
   closedIssues: search(type: ISSUE, query: $closedQuery, first: 50) {
     issueCount
     nodes { ... }
   }
   ```
   (Optionally do the same for `openIssues` if any open badge counts nodes.)
2. Parse `closedIssues.issueCount` through `ConsolidatedGitHubResponse` and set `appState.doneIssuesLast24h` from it instead of `closedIssues.count` (`IssueTracker.swift:349`).

This keeps `nodes` capped (we only need up to 50 for any list rendering) while the **badge** reports the true 24h total.

### Notes / alternatives
- If any UI actually *renders the list* of done tickets (not just the count) and needs all >50 items, that path would need real pagination (`pageInfo { hasNextPage endCursor }`) — but for the header **count badge**, `issueCount` is sufficient and preferable.
- The same `first: 50` cap-then-`.count` pattern exists in the sibling backends (`GitHubCodeBackend.swift:238,258`, `GitLabTaskBackend.swift:78`, `GitLabCodeBackend.swift:60`); worth a follow-up pass if any of their counts are surfaced as badges, though this issue is scoped to the **done-in-24h** ticket badge.

## Where to look
`Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubTaskBackend.swift` (`consolidatedIssuesQuery` / `consolidatedIssuesQueryNoProjects`, lines 305–358; `parseIssuesResponse`) · `Sources/Crow/App/IssueTracker.swift:349` (+ the `ConsolidatedGitHubResponse` struct ~:537) · `Packages/CrowUI/Sources/CrowUI/TicketBoardView.swift:620`.

dhilgaertner:
Closes #562

The sidebar's green done-in-24h ticket badge saturated at 50: it counted the `closedIssues` GraphQL search nodes, which are hard-capped at `first: 50`. GitHub's `search` connection reports the total match count via `issueCount` independent of page size, so the badge now reads that — no pagination, no extra API cost.

- Added `issueCount` to the `closedIssues` search in both `consolidatedIssuesQuery` and `consolidatedIssuesQueryNoProjects`
- Threaded it through `parseIssuesResponse` as `AssignedListing.closedTotalCount`, defaulting to `closed.count` so the Jira/GitLab/Corveil backends and the SAML partial-recovery path are unaffected
- `IssueTracker` accumulates `closedTotalCount` into `doneIssuesLast24h`; the board's Done list still renders the (capped) nodes
- Tests: extended the parse fixture with `issueCount: 137` (> nodes.count) and added a fallback case for responses without `issueCount`

Verified with `swift build`, `swift test` (root, 264 tests), and `swift test --package-path Packages/CrowProvider`.

### DIFF
--- a/Packages/CrowProvider/Sources/CrowProvider/BackendTypes.swift
+++ b/Packages/CrowProvider/Sources/CrowProvider/BackendTypes.swift
@@ -213,6 +213,11 @@ public struct PRMetadata: Sendable {
 public struct AssignedListing: Sendable {
     public let open: [AssignedIssue]
     public let closed: [AssignedIssue]
+    /// Total recently-closed matches, independent of the page cap on `closed`
+    /// (GitHub's `search` fetches at most 50 nodes but reports the full
+    /// `issueCount`). Falls back to `closed.count` when the backend doesn't
+    /// report a total, so it's always safe to badge from.
+    public let closedTotalCount: Int
     /// GitHub-only; nil for GitLab.
     public let rateLimit: GitHubRateLimit?
     /// When non-nil, the backend completed the call but had to degrade the
@@ -230,12 +235,14 @@ public struct AssignedListing: Sendable {
     public init(
         open: [AssignedIssue],
         closed: [AssignedIssue],
+        closedTotalCount: Int? = nil,
         rateLimit: GitHubRateLimit? = nil,
         missingScope: String? = nil,
         samlRestricted: Bool = false
     ) {
         self.open = open
         self.closed = closed
+        self.closedTotalCount = closedTotalCount ?? closed.count
         self.rateLimit = rateLimit
         self.missingScope = missingScope
         self.samlRestricted = samlRestricted
--- a/Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubTaskBackend.swift
+++ b/Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubTaskBackend.swift
@@ -354,6 +354,7 @@ public struct GitHubTaskBackend: TaskBackend {
         }
       }
       closedIssues: search(type: ISSUE, query: $closedQuery, first: 50) {
+        issueCount
         nodes {
           ... on Issue {
             number title url state updatedAt
@@ -378,6 +379,7 @@ public struct GitHubTaskBackend: TaskBackend {
         }
       }
       closedIssues: search(type: ISSUE, query: $closedQuery, first: 50) {
+        issueCount
         nodes {
           ... on Issue {
             number title url state updatedAt
@@ -409,8 +411,12 @@ public struct GitHubTaskBackend: TaskBackend {
             dateFormatter: dateFmt,
             projectStatusOverride: .done
         )
+        // Total matches in the 24h window — `nodes` is capped at `first: 50`,
+        // so the badge count must come from the search connection's
+        // `issueCount` or it saturates at 50 (#562).
+        let closedTotal = (dataObj["closedIssues"] as? [String: Any])?["issueCount"] as? Int
         let rate = parseRateLimit(dataObj["rateLimit"] as? [String: Any])
-        return AssignedListing(open: open, closed: closed, rateLimit: rate, missingScope: missingScope)
+        return AssignedListing(open: open, closed: closed, closedTotalCount: closedTotal, rateLimit: rate, missingScope: missingScope)
     }
 
     /// Recover the accessible-org issues GitHub returned alongside a SAML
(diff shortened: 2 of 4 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Purely additive: the change requests one extra scalar (`issueCount`) from the GitHub `search` connection and reads it back. No new auth, network, or injection surface.
- Response parsing is defensive — `(dataObj["closedIssues"] as? [String: Any])?["issueCount"] as? Int` degrades to `nil` on a malformed/absent field rather than trapping, and `AssignedListing` falls back to `closed.count`.

**Concerns:** None.

### Code Quality
- Correct root-cause fix, matching the approach in #562: `nodes` stays capped at `first: 50` for list rendering while the badge count now reads the connection's true `issueCount`. Zero extra API cost / no pagination.
- `issueCount` is added to **both** query variants (`consolidatedIssuesQuery` and `consolidatedIssuesQueryNoProjects`), so the no-projects scope-retry path is covered too.
- The optional `closedTotalCount: Int? = nil` initializer arg defaulting to `closed.count` keeps the Jira/GitLab/Corveil backends and the SAML partial-recovery path (`recoverPartialIssues`) unaffected — they simply inherit the node count as before.
- No dedup regression: the previous badge counted the un-deduped `closedIssues.count`, and the new code counts the un-deduped `closedTotalCount`. The `uniqueDone` filter (`IssueTracker.swift:435`) only ever affected the rendered `allIssues` list, never the count, so semantics are preserved.
- Tests are on point: `testGitHubTaskBackendParsesConsolidatedIssues` now asserts `issueCount: 137` (> node count) flows to `closedTotalCount`, and `testGitHubTaskBackendClosedTotalFallsBackToNodeCount` covers the missing-`issueCount` fallback.

**Verification:**
- `swift build` — full app compiles clean (the `BuildInfo` symbol needs `scripts/generate-build-info.sh` via `make build`; a pre-existing harness requirement, unrelated to this PR).
- `swift test --package-path Packages/CrowProvider --filter BackendsTests` — 55 tests, 0 failures, including both new cases.

- `closedTotalCount` is the raw search total and is not deduped against the open set. In the unlikely event an issue matched both the assigned-open (`first: 100`) and closed-in-24h (`first: 50`) queries in the same refresh, the badge could over-count by one. This exactly matches prior behavior (the old count wasn't deduped either), so it's not a regression — just an observation if perfect accuracy is ever desired.
- The same `first: N`-then-`.count` pattern exists in the sibling backends (`GitHubCodeBackend`, `GitLab*`). The issue explicitly scopes this PR to the done-in-24h ticket badge, so leaving those for a follow-up is the right call.

### Summary Table

---

PR #580, opened 2026-07-03
Auto-complete: reconcile unwatched done job sessions on startup/tick

### LINKED ISSUE #579: Auto-complete: reconcile already-done job sessions not being watched (survive app restart)
Follow-up to #561/#566. Auto-completion works, but only for job runs the current `JobScheduler` process fired and is still watching in memory. In practice, job sessions that are visibly done (green, agent at rest) are left in `.active` and never transition — a human still has to complete them by hand, which is the exact problem #561 set out to fix.

## Root cause
`JobScheduler.watchedRuns` is **in-memory only** (`JobScheduler.swift:54-56`): "an app relaunch mid-run drops the watch, reverting that run to the pre-CROW-561 'linger until manually completed' behavior." The tick loop fires due jobs and drains `watchedRuns`, but never re-adopts existing job sessions. So a job is stranded green-but-`.active` whenever:
- the app was **relaunched** after the run finished (watch state lost), or
- the run **predates the #566 merge** (never watched).

There is no reconciliation pass over existing `.job` sessions.

## Fix
On startup and on the existing tick, **reconcile unwatched `.job` sessions**: for any `.active` job session that isn't in `watchedRuns`, evaluate it with the *same* at-rest criteria the watched path already uses (`RunWatch` fate logic in `JobScheduler`) and complete it when it qualifies:
- terminal `.agentLaunched`, agent at rest (`.done` for Claude, `.idle` for TUI agents), not `.working`/`.waiting`;
- `.waiting` (awaiting-input OR errored/`StopFailure`) stays `.active` — never silently complete those;
- reuse `SessionService.completeSession` (already guards manager sessions).

The reconciler must **rebuild enough state to apply the settle-window guard** (or treat a session that has been at rest across a full tick as settled) so it doesn't complete a multi-prompt job mid-gap. Keep the completion decision in the existing pure/unit-testable fate function so watched and reconciled paths can't diverge.

## Acceptance
- A job session that finished successfully is auto-completed even if `JobScheduler` never watched it (app restarted, or it predates the feature) — no manual action.
- Errored / awaiting-input job sessions are still left `.active`.
- Watched and reconciled paths share the same at-rest decision logic.
- Verified by completing the currently-stuck green/done job sessions after this lands.

Related: #561, #566.

dgershman:
Closes #579

Follow-up to #561/#566. Auto-completion of finished job sessions worked **only** for runs the current `JobScheduler` process fired and was still watching in memory (`watchedRuns` is in-memory only), so a job session left visibly done (green, agent at rest) stayed stuck in `.active` whenever:

- the app was **relaunched** after the run finished (watch state lost), or
- the run **predated** #566 (never watched).

## Change

Add a reconciliation pass — **on startup and on the existing 30s tick** — that adopts any active `.job` session not already in `watchedRuns`. Rather than duplicate the settle/decide machine, it inserts a `RunWatch` stamped with `startedAt`/`promptsDeliveredAt` at *now*, so the existing `checkFinishedRuns` loop judges it via the **same pure `finishDecision`** the fired path uses — the watched and reconciled paths can't diverge.

Adopting at *now* does two things for free:
- The run sits inside the `finishSettleDelay` window, so it can only complete on a **later** tick if it is *still* at rest (at rest across a full tick) — guarding against completing a multi-prompt job mid-gap.
- The 12h `maxWatchDuration` cap counts from adoption, not the session's real start, so a days-old predates-the-feature run is still eligible (using its real start would trip the cap and never complete it).

Errored / awaiting-input runs (`.waiting`) and the never-started default (`.idle`) stay `.active` unchanged — `finishDecision` only completes on `.done`. Runs we fired (already watched) are skipped so their real delivery timing is untouched; a session whose managed terminal is gone is left `.active` to surface.

## Acceptance criteria

- [x] A job session that finished successfully is auto-completed even if `JobScheduler` never watched it (app restarted, or predates the feature) — no manual action.
- [x] Errored / awaiting-input job sessions are still left `.active`.
- [x] Watched and reconciled paths share the same at-rest decision logic (`finishDecision`, reused verbatim — no change).

## Implementation notes

- Contained in `Sources/Crow/App/JobScheduler.swift`: a new pure `shouldReconcile(kind:status:alreadyWatched:)` eligibility gate, a `reconcileUnwatchedJobs(now:)` stateful adopter, and calls in `start()` and at the top of `tick()`. `finishDecision`, `checkFinishedRuns`, and `SessionService.completeSession` (which guards managers) are unchanged.
- Note: the ticket's "`.idle` for TUI agents" premise is stale — all agent kinds finish at `.done`; reusing `finishDecision` gets this right. Also corrected a stale `.done/.idle` comment in `JobFinishDecisionTests`.

## Testing

- `make build` — clean.
- `swift test --filter JobReconcileDecisionTests` — 6/6 (eligibility gate + adoption/settle contract).
- `swift test --filter JobFinishDecisionTests` — 11/11 still green.
- Post-merge (per acceptance): confirm the currently-stuck green/done job sessions auto-complete once this lands.

https://claude.ai/code/session_01JFKGGRAD3Dhw9iZmzZ8ycB

### DIFF
--- a/Sources/Crow/App/JobScheduler.swift
+++ b/Sources/Crow/App/JobScheduler.swift
@@ -68,6 +68,13 @@ final class JobScheduler {
     }
 
     func start() {
+        // Adopt any already-done job sessions we aren't watching so they still
+        // auto-complete after a relaunch, or if they predate this feature — the
+        // in-memory watch is otherwise lost and the run lingers in `.active`
+        // (CROW-579). Runs post-hydration (`start()` is wired after
+        // `hydrateState`), so restored sessions/terminals are present.
+        reconcileUnwatchedJobs(now: Date())
+
         // First tick after one interval — gives the app a grace period at
         // launch and lets overdue jobs fire shortly after.
         timer = Timer.scheduledTimer(withTimeInterval: tickInterval, repeats: true) { [weak self] _ in
@@ -85,6 +92,9 @@ final class JobScheduler {
 
     private func tick() {
         let now = Date()
+        // Adopt any active job sessions we aren't already watching (relaunched
+        // or predating the feature) so they can auto-complete too (CROW-579).
+        reconcileUnwatchedJobs(now: now)
         // Auto-complete any job runs that have finished successfully. Runs on
         // every tick regardless of whether jobs are due (CROW-561).
         checkFinishedRuns(now: now)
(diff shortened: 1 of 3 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface: the change reads in-memory `AppState` and mutates a private dictionary — no input parsing, no shell execution, no network, no persistence-format changes.
- Fails safe in every ambiguous case: the managed-terminal guard leaves torn-down sessions `.active` (`JobScheduler.swift:240-244`), and completion still flows through the unchanged `finishDecision`, which requires a real `.done` hook event — errored (`.waiting`) and never-launched (`.idle`) runs cannot be silently completed.

**Concerns:**
- None.

### Code Quality
Verified the load-bearing claims rather than taking the PR body's word:

- **Wiring order holds** — `jobScheduler.start()` (AppDelegate.swift:865) runs after `service.hydrateState()` (AppDelegate.swift:576), and `hydrateState` populates `appState.terminals` synchronously before the reconcile pass reads it. Readiness/hook state land asynchronously via the per-terminal rehydrate tasks, but that only delays completion to a later tick — it can't cause a wrong one.
- **Relaunch path actually works** — managed job terminals get a readiness slot in `hydrateState` (job sessions are not managers, so `SessionService.swift:191-196` applies), the tmux adopt path sets `.agentLaunched` and fills `.done` for `trackReadiness` terminals (`SessionService.swift:321-357`), and `trackReadiness` is true for managed non-manager terminals (`SessionService.swift:256`). So a relaunched done job satisfies all of `finishDecision`'s gates.
- **Race with `fire()` is safe** — a tick's reconcile can adopt a just-fired session during `runJob`'s suspension (the session is appended at `SessionService.swift:2008` before `fire`'s continuation registers the real watch), but the real `RunWatch` then overwrites the adopted one, and the 20s settle window makes a premature completion in that gap impossible.
- Tests: `make build` clean; full suite passes (274 tests / 32 suites), including the 6 new reconcile tests and the 11 existing finish-decision tests.

Green (consider, non-blocking):

1. **`maxWatchDuration` is now effectively void for job sessions** (`JobScheduler.swift:337`): when the 12h cap trips (`.stopWatching`), the session is still an active unwatched `.job`, so the next tick re-adopts it with a fresh `startedAt` — a blocked/errored run cycles adopt→cap→re-adopt forever. No wrong completion can result and memory cost is one dict entry, but the cap's stated purpose ("doesn't linger in memory forever", `JobScheduler.swift:35-37`) no longer holds for reconciled runs. Consider updating that comment, or tracking dropped session IDs to adopt only once.
2. **Partially-delivered multi-prompt jobs complete as success across a relaunch**: prompt-delivery progress is in-memory, so if the app quits after prompt 1 of 3, the remaining prompts are never re-delivered and reconcile completes the session once the agent is at rest. Pre-PR it lingered visibly `.active`. Not distinguishable without persisting delivery state — worth a sentence in the `reconcileUnwatchedJobs` doc comment as a known limitation.
3. **The stateful adopter is untested**: `shouldReconcile` and the settle contract are covered, but `reconcileUnwatchedJobs` itself (managed-terminal guard, `RunWatch` stamping) has no direct test. This follows the codebase's pure-decision test convention, so fine as-is — just noting the guard at `JobScheduler.swift:243` is only exercised manually.

### Summary Table

---

PR #583, opened 2026-07-03
Manager + button: pick coding agent before opening new session

### LINKED ISSUE #582: New-manager-session (+) button: pick the coding agent from a menu before opening
When starting a new (extra) **manager** session via the **+** button, let the user choose which coding agent to launch instead of silently using the configured default (`agentsByKind.manager` / `defaultAgentKind`).

## Behavior
- Clicking **+** for a new manager session shows a **context menu / picker** of coding agents: Claude Code, Codex, Cursor, OpenCode.
- **Only available agents are selectable** — an agent whose CLI can't be found (via the same `CodingAgent.findBinary` / `ShellEnvironment` resolution used at launch) is hidden or shown disabled, so you can't pick one that won't start.
- Selecting an agent opens the new manager session using **that** agent (its `managerLaunchCommand`), regardless of the config default.
- Each agent's existing icon/displayName (`sparkles`, `terminal.fill`, `cursorarrow.rays`, `chevron.left.forwardslash.chevron.right`) should appear in the menu for quick recognition.

## Defaults / details
- If exactly one agent is available, it's fine to skip the menu and open directly (no point prompting with a single option) — or still show it for consistency; implementer's call, note which was chosen.
- The configured `agentsByKind.manager` remains the pre-highlighted / default choice in the menu.
- This is a **per-new-session, one-shot** choice — it does not change the persisted config default.
- Scope is the manager-session + button specifically (the "extra manager sessions" entry point). Work-session agent selection is out of scope here unless it's the same code path.

## Acceptance
- The + (new manager session) action presents an agent picker listing Claude Code / Codex / Cursor / OpenCode.
- Unavailable agents (CLI not found) are not selectable.
- The new manager session launches with the selected agent, not just the config default.
- Choosing an agent does not mutate the persisted `agentsByKind` / `defaultAgentKind`.

Related: #545 (OpenCode agent integration / parity).

dgershman:
Closes #582

## What

The **+** button that spawns an extra **Manager** session (the pill next to "Manager" in the Review Board header) silently launched the configured default agent (`agentsByKind["manager"] ?? defaultAgentKind`). This turns it into an **agent-picker menu**: clicking **+** lists the available coding agents (with their icons), and picking one opens a new Manager session with **that** agent — for this session only, without mutating the persisted config.

## How

- **`AppState.onCreateManager`** now carries an optional `AgentKind` (`nil` = "use configured default").
- **`SessionService.createManagerSession`** gains an `agentKind:` override, resolved via a small `resolvedManagerAgentKind` helper: an explicit pick wins, else `agentsByKind["manager"] ?? defaultAgentKind`. No config write. The default-`nil` param keeps the existing create-session RPC path unchanged.
- **`ReviewBoardView`**'s `newManagerButton` becomes a `Menu` of `AgentRegistry.shared.allAgents()` when ≥2 agents are available; the configured Manager default is marked **(default)**. Each item uses the agent's existing `displayName` + `iconSystemName`.

## Availability

No new availability check was needed — `AgentRegistry` only holds agents whose CLI resolved via `findBinary()` at startup (Claude Code always registered; Codex/Cursor/OpenCode only when their binary is found). So the menu lists `allAgents()` directly, mirroring the existing Settings per-action picker. Uninstalled agents never appear.

## Single-agent case

Per the ticket's "implementer's call": with only one agent available there's nothing to pick, so **+** stays a plain button that opens with the default (matches how `CreateSessionView`/`SettingsView` disable their pickers below 2 agents).

## Tests

`Tests/CrowTests/ManagerMigrationTests.swift`:
- `resolvedManagerAgentKindPrefersExplicitOverride` — explicit pick wins over config default; config left untouched.
- `resolvedManagerAgentKindFallsBackToConfigDefault` — `nil` resolves `agentsByKind["manager"] ?? defaultAgentKind`.

`make build` clean; `swift test --filter ManagerMigrationTests` → 9/9 pass.

https://claude.ai/code/session_01KuRA2BAmpVAXDK4pXKPtzm

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -289,7 +289,9 @@ public final class AppState {
     }
 
     /// Called when the user clicks the sidebar "+" to spawn a new Manager session.
-    public var onCreateManager: (() -> Void)?
+    /// The optional `AgentKind` is a one-shot, per-session agent override chosen
+    /// from the picker menu; `nil` means "use the configured default" (#582).
+    public var onCreateManager: ((AgentKind?) -> Void)?
 
     /// Called when user clicks "Work on" for an assigned issue.
     public var onWorkOnIssue: ((String) -> Void)?  // receives issue URL
--- a/Packages/CrowUI/Sources/CrowUI/ReviewBoardView.swift
+++ b/Packages/CrowUI/Sources/CrowUI/ReviewBoardView.swift
@@ -412,27 +412,68 @@ public struct ManagerReviewsAllowListRow: View {
         }
     }
 
+    /// "+" for a new Manager session. When two or more coding agents are
+    /// available (registered in `AgentRegistry` — which only holds agents whose
+    /// CLI resolved via `findBinary()`), the "+" opens an agent-picker menu so
+    /// the user can launch this one session with a specific agent (#582). With a
+    /// single agent there's nothing to pick, so it stays a plain button that
+    /// opens with the configured default.
+    @ViewBuilder
     private var newManagerButton: some View {
-        Button {
-            appState.onCreateManager?()
-        } label: {
-            Image(systemName: "plus")
-                .font(.system(size: 12, weight: .bold))
-                .foregroundStyle(CorveilTheme.goldDark)
-                .frame(width: 28)
-                .padding(.vertical, 8)
-                .background(
-                    RoundedRectangle(cornerRadius: 6)
-                        .fill(CorveilTheme.bgSurface)
-                        .overlay(
-                            RoundedRectangle(cornerRadius: 6)
-                                .strokeBorder(CorveilTheme.goldDark.opacity(0.3), lineWidth: 1)
-                        )
-                )
+        let agents = AgentRegistry.shared.allAgents()
+        if agents.count >= 2 {
+            Menu {
+                let defaultKind = appState.agentKind(for: .manager)
+                ForEach(agents, id: \.kind) { agent in
+                    Button {
+                        appState.onCreateManager?(agent.kind)
+                    } label: {
+                        // The configured Manager default is pre-marked so it
+                        // reads as the default choice, without changing config.
+                        if agent.kind == defaultKind {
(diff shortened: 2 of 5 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface: the picker only offers agents already registered in `AgentRegistry`, which is populated at startup exclusively from binaries resolved via `findBinary()` — no user-supplied strings reach the launch-command path.
- The override is genuinely one-shot: `resolvedManagerAgentKind` reads config but never writes it, and the tests explicitly assert `agentsByKind` / `defaultAgentKind` are untouched (`Tests/CrowTests/ManagerMigrationTests.swift:106`).
- The RPC create-session path (`Sources/Crow/App/AppDelegate.swift:1375`) uses the default-`nil` parameter, so socket-driven session creation behavior is unchanged.

**Concerns:**
- None.

### Code Quality
- Resolution flow is coherent end-to-end: picker → `onCreateManager(AgentKind?)` → `createManagerSession(agentKind:)` → `Session.agentKind` → `managerCommand(for:)`, which already falls back to `defaultAgent` if a persisted kind is no longer registered (e.g. agent uninstalled before rehydrate) — so the override degrades safely.
- All `onCreateManager` call sites were updated for the signature change; the closure-type change makes any missed site a compile error.
- Single-agent fallback to a plain button matches the existing <2-agents pattern in `CreateSessionView`/`SettingsView`, and the shared `newManagerLabel` avoids duplicating the pill styling.
- Green (consider, pre-existing): `AgentRegistry.allAgents()` returns `Array(agents.values)` (`Packages/CrowCore/Sources/CrowCore/Agent/AgentRegistry.swift:47`), so menu item order is dictionary-order and can differ between app launches. SettingsView/CreateSessionView share this behavior; a stable sort inside `allAgents()` would fix all three at once. Non-blocking.
- Verified locally: `swift build` clean; `swift test --filter ManagerMigrationTests` → 9/9 pass, including both new resolution tests.

### Summary Table

**Recommendation:** Approve — driven by 0 Red, 0 Yellow, 1 Green findings.

---

PR #584, opened 2026-07-03
M0+M1: crowd daemon — Hummingbird HTTP/WS + core JSON-RPC handlers + Linux terminal

### LINKED ISSUE #581: Cross-platform Crow: web-app-first UI (browser + Electron/Tauri) over a headless Swift daemon
## Goal

Make Crow truly cross-platform (Linux + Windows, in addition to macOS), and enable connecting to **crow-ui from a web browser** in the future.

## Architecture decisions

- **Reuse the proven Swift logic**, don't rewrite in TS. Crow is already client/server (a JSON-RPC 2.0 Unix-socket server + the `crow` CLI). ~12.5K LOC of domain logic (CrowCore, CrowProvider's TaskBackend/CodeBackend split, CrowGit rebase-safety invariants, agent `StateSignalSource` state machines) is Foundation-only and near-Linux-clean — reuse it verbatim behind a headless `crowd` daemon.
- **Web-app-first.** The daemon serves the web UI over HTTP + one WebSocket (JSON-RPC + terminal byte-streaming). The **browser is a first-class client**; the UI is a pure client with no business logic — all state/actions go over the API. This is what enables "connect just the UI, everything else over API calls," and makes Crow remote-capable nearly for free.
- **All-Swift server layer.** `crowd` serves HTTP/WS itself (Hummingbird) and owns terminal PTYs. One runtime, one binary per platform — no Node sidecar.
- **Desktop shell is thin, deferred, and A/B-able.** Because the UI is a browser-reachable web app, the shell just loads `http://localhost:PORT`. We want to **A/B test Electron vs Tauri** (uniform Chromium vs system webview — the interesting comparison is xterm.js rendering/perf across engines). Keep them swappable via a small `HostBridge` interface (~a dozen methods) with `electron` / `tauri` / `browser` implementations; forbid direct shell API calls in the UI.

```
Browser tab  ·  Tauri shell  ·  Electron shell  ·  crow CLI     ← interchangeable thin clients
        │  HTTP (UI assets) + WebSocket (JSON-RPC + terminal bytes)
        ▼
Swift `crowd` daemon   — Hummingbird HTTP/WS + Swift PTY + reused CrowCore/Provider/Git/Persistence
        │  subprocess
        ▼
tmux · git · gh/glab/acli · claude/codex/…
```

## Reusable vs throwaway (≈32K LOC surveyed)

- **Reuse (~12.5K LOC + assets):** CrowCore domain brain, CrowProvider/Git/Persistence/IPC, the `crow` CLI (already Linux-buildable), and the terminal stack — xterm.js 6.0.0 assets, `crow-tmux.conf`, `crow-shell-wrapper.sh` (near-verbatim).
- **Rewrite (~18.7K LOC, macOS-locked):** CrowUI (SwiftUI), CrowTerminal Swift (WKWebView/Darwin PTY), Sources/Crow app shell (AppDelegate/NSMenu/notifications). Rebuilt as the web UI + Swift daemon.
- **Replace/drop:** CrowTelemetry's Network.framework OTLP receiver (~150 lines → tiny HTTP handler, or cut for MVP).

## Terminal

Browser access forces the PTY server-side (no node-pty in a browser tab). The daemon owns a PTY doing `tmux attach` and streams bytes to xterm.js over the WebSocket (like VS Code Server / ttyd). Port `CrowTerminal/PTYProcess.swift` (Darwin openpty/posix_spawn) to Linux (`openpty` in `<pty.h>`, link `-lutil`). tmux stays a separate server.

## Spike plan

- [ ] **M0 — `crowd` daemon skeleton (Linux):** new `Sources/CrowDaemon` executable + `crowd` product; embeds Hummingbird; serves a WebSocket speaking the existing JSON-RPC protocol; wires 3 real handlers (`new-session`, `list-sessions`, `add-worktree`) on `AppState` + CrowPersistence + CrowGit. Proves *Swift web server on Linux* + *handler extraction out of AppDelegate*.
- [ ] **M1 — Terminal over WebSocket:** port PTYProcess to Linux, `tmux attach` streamed to a served xterm.js page, launching a real `claude`. Reuse `crow-tmux.conf` + `crow-shell-wrapper.sh` + xterm assets. Watch tmux OSC readiness markers + extended-keys (Shift+Enter).
- [ ] **M2 — Minimal web UI:** session list (JSON-RPC over WS) + terminal, reachable in a browser.
- [ ] **Shells (parallel, after M2):** `desktop/electron/` + `desktop/tauri/` thin wrappers sharing the `HostBridge` seam; A/B compare.

## Already done (uncommitted)

The `crow` CLI now compiles and runs natively on Linux (Swift 6.3.3) via 4 guarded, macOS-safe `#if canImport` fixes: `crowSockStream` for Glibc's `__socket_type` (CrowIPC), `import Observation` (AppState), and `FoundationNetworking` imports in the three Jira REST clients. Not yet committed; macOS build not re-verified.

## Repo layout (all in-repo)

```
web/               ← shared crow-ui (client only)
desktop/electron/  ← thin shell + ElectronBridge
desktop/tauri/     ← thin shell + TauriBridge
Sources/CrowDaemon/← the crowd daemon
```

dgershman:
Implements **M0 + M1** of the cross-platform architecture spike (#581): a headless Swift `crowd` daemon that serves Crow's proven, Foundation-only domain logic over HTTP + WebSocket, so the UI can be a browser / Electron / Tauri thin client and Crow can run on Linux — reusing the Swift, not rewriting in TS.

## M0 — daemon skeleton + core JSON-RPC
- New `Packages/CrowDaemon` library + **`crowd`** executable product. Embeds **Hummingbird 2.x**; serves a WebSocket at **`/rpc`** speaking the existing JSON-RPC 2.0 protocol, reusing `CrowIPC.CommandRouter` verbatim (no semaphore bridge — we're already async).
- Wires 3 real handlers — **`new-session`**, **`list-sessions`**, **`add-worktree`** — on `AppState` + `CrowPersistence` (+ `CrowGit`), mirroring the app's `AppDelegate` closures but with **no AppKit / SessionService**. `add-worktree` materializes the worktree via `GitManager.createWorktree` to exercise the reused git layer on Linux.
- Also binds the existing **Unix socket** (`SocketServer`), so the current **`crow` CLI drives the daemon unchanged**. Both transports share one router.
- Hydrates `AppState` (sessions + worktrees) from `store.json` at boot. `import Observation` added to `AppState` so `@Observable` resolves on Linux (harmless on macOS).

## M1 — terminal over WebSocket
- **Port `PTYProcess` to Linux**: `#if canImport(Glibc)` imports, per-platform `read`/`write`/`ioctl`, and a header-only **`CPty`** C shim exposing `openpty(3)` (`-lutil`, Linux-only target dep). Adds opt-in `deliverOnMainQueue:` (default `true` keeps app behavior) so the daemon gets PTY output on the read queue and bridges it into an `AsyncStream`.
- **`/terminal`** WebSocket streams a PTY running `tmux attach-session` to xterm.js: raw bytes out (binary), keystrokes in (binary), `{type:"resize"}` control frame — the same four message types as the macOS WKWebView bridge. Reuses `TmuxController` + bundled `crow-tmux.conf`.
- Serves a WebSocket-transport `terminal.html` and streams the **xterm.js 6.0.0** assets straight from `CrowTerminal`'s bundle (no duplication).
- Makes **`CrowTerminal` compile on Linux**: the WKWebView/AppKit files and `TmuxBackend.cockpitSurface()` are gated behind `#if canImport(WebKit/AppKit)`; `Package.swift` links WebKit on macOS only, util on Linux only.

## Verification
- **macOS build unaffected**: `make build` (CrowApp + crow) passes; gated code takes its existing macOS branch.
- **Smoke-tested on macOS** (daemon on a test socket/port): `crow list-sessions` over the daemon's Unix socket; `/rpc` list-sessions + method-not-found over WS; `/health`; xterm assets + path-traversal guard (400); and a live **`/terminal` round-trip** — an `echo` keystroke streamed to the PTY and echoed back through PTY→tmux→WS.
- **Linux**: `swift build --product crowd` in a `swift:6` container is the recommended pre-merge check — the Glibc PTY branch + `CPty` shim can't be exercised on macOS.

## Not in this PR (follow-ups)
M2 web UI (session list + terminal), multiplexing `/rpc` + `/terminal` into one socket, the `crow-shell-wrapper.sh` OSC-readiness + agent auto-launch on the daemon terminal, Linux clipboard (`crow-tmux.conf` hardcodes `pbcopy`), and the `desktop/electron` + `desktop/tauri` thin shells.

Closes #581

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -7,6 +7,10 @@ let package = Package(
     products: [
         .executable(name: "CrowApp", targets: ["Crow"]),
         .executable(name: "crow", targets: ["CrowCLI"]),
+        // Headless cross-platform daemon (CROW-581). Build standalone on Linux
+        // with `swift build --product crowd` — it does not depend on the AppKit
+        // `Crow` target or its generated BuildInfo.
+        .executable(name: "crowd", targets: ["crowd"]),
     ],
     dependencies: [
         .package(path: "Packages/CrowCore"),
@@ -22,6 +26,7 @@ let package = Package(
         .package(path: "Packages/CrowIPC"),
         .package(path: "Packages/CrowTelemetry"),
         .package(path: "Packages/CrowCLI"),
+        .package(path: "Packages/CrowDaemon"),
         .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.5.0"),
     ],
     targets: [
@@ -56,6 +61,13 @@ let package = Package(
                 .product(name: "CrowCLILib", package: "CrowCLI"),
             ],
             path: "Sources/CrowCLI"
+        ),
+        .executableTarget(
+            name: "crowd",
+            dependencies: [
+                .product(name: "CrowDaemon", package: "CrowDaemon"),
+            ],
+            path: "Sources/crowd"
         ),
         .testTarget(
             name: "CrowTests",
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -1,4 +1,9 @@
 import Foundation
+// `@Observable` is re-exported by Foundation on Apple platforms, but on Linux
+// the Observation module must be imported explicitly for the macro to resolve.
+// Importing it unconditionally is harmless on macOS and unblocks the headless
+// `crowd` daemon reusing `AppState` on Linux (CROW-581).
+import Observation
 
 /// Check whether a repo name matches any of the given patterns.
 /// Supports exact matches and simple glob patterns with `*` (e.g., `org/*`, `*/repo`).
--- a/Packages/CrowDaemon/Package.swift
+++ b/Packages/CrowDaemon/Package.swift
@@ -1,0 +1,36 @@
+// swift-tools-version: 6.0
+import PackageDescription
+
+let package = Package(
+    name: "CrowDaemon",
+    platforms: [.macOS(.v14)],
+    products: [
+        .library(name: "CrowDaemon", targets: ["CrowDaemon"]),
+    ],
+    dependencies: [
+        .package(path: "../CrowCore"),
+        .package(path: "../CrowPersistence"),
+        .package(path: "../CrowGit"),
+        .package(path: "../CrowIPC"),
+        .package(path: "../CrowTerminal"),
+        .package(url: "https://github.com/hummingbird-project/hummingbird.git", from: "2.0.0"),
+        .package(url: "https://github.com/hummingbird-project/hummingbird-websocket.git", from: "2.0.0"),
+    ],
+    targets: [
+        .target(
+            name: "CrowDaemon",
+            dependencies: [
+                .product(name: "CrowCore", package: "CrowCore"),
+                .product(name: "CrowPersistence", package: "CrowPersistence"),
+                .product(name: "CrowGit", package: "CrowGit"),
+                .product(name: "CrowIPC", package: "CrowIPC"),
+                .product(name: "CrowTerminal", package: "CrowTerminal"),
+                .product(name: "Hummingbird", package: "hummingbird"),
+                .product(name: "HummingbirdWebSocket", package: "hummingbird-websocket"),
+            ],
+            resources: [
+                .copy("Resources/web"),
+            ]
+        ),
+    ]
+)
--- a/Packages/CrowDaemon/Package.swift
+++ b/Packages/CrowDaemon/Package.swift
@@ -32,5 +32,9 @@ let package = Package(
                 .copy("Resources/web"),
             ]
         ),
+        .testTarget(
+            name: "CrowDaemonTests",
+            dependencies: ["CrowDaemon"]
+        ),
     ]
 )
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
@@ -1,4 +1,6 @@
-#if canImport(Glibc)
+#if canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
 import Glibc
 #endif
 import CrowCore
@@ -34,13 +36,24 @@ public enum CrowDaemon {
         let commandRouter = makeCommandRouter(
             appState: appState, store: store, git: git, devRoot: options.devRoot)
 
-        // Unix socket — lets the existing `crow` CLI talk to the daemon.
-        let socketServer = SocketServer(socketPath: options.socketPath, router: commandRouter)
-        do {
-            try socketServer.start()
-            log("JSON-RPC Unix socket listening at \(options.socketPath)")
-        } catch {
-            log("WARNING: socket bind failed (\(error)); continuing with HTTP/WS only")
+        // Unix socket — lets the existing `crow` CLI talk to the daemon. Refuse
+        // to bind a socket another server already answers on (e.g. the desktop
+        // app's crow.sock): `SocketServer.start()` unlinks unconditionally, so
+        // binding it would hijack the app's CLI channel and dual-write the store
+        // (CROW-581 review). The daemon's default is a distinct `crowd.sock`;
+        // sharing the app's path is explicit via --socket, and even then we
+        // won't steal a live one.
+        if Self.socketInUse(options.socketPath) {
+            log("WARNING: \(options.socketPath) is already in use (another crowd or the Crow app). "
+                + "Not binding the Unix socket to avoid hijacking it — use a distinct --socket. Continuing with HTTP/WS only.")
+        } else {
+            let socketServer = SocketServer(socketPath: options.socketPath, router: commandRouter)
+            do {
+                try socketServer.start()
+                log("JSON-RPC Unix socket listening at \(options.socketPath)")
+            } catch {
+                log("WARNING: socket bind failed (\(error)); continuing with HTTP/WS only")
+            }
         }
 
         // Terminal cockpit (tmux). Optional — RPC still works without tmux.
@@ -96,6 +109,30 @@ public enum CrowDaemon {
     private static func log(_ message: String) {
         FileHandle.standardError.write(Data("[crowd] \(message)\n".utf8))
     }
+
+    /// Whether a live server is already accepting connections on the Unix socket
+    /// at `path` — used to avoid stealing another process's socket. A stale
+    /// socket file (nothing listening) returns false, so normal startup still
+    /// replaces it.
+    private static func socketInUse(_ path: String) -> Bool {
+        guard FileManager.default.fileExists(atPath: path) else { return false }
+        let fd = socket(AF_UNIX, SOCK_STREAM, 0)
+        guard fd >= 0 else { return false }
+        defer { close(fd) }
+        var addr = sockaddr_un()
+        addr.sun_family = sa_family_t(AF_UNIX)
+        _ = path.withCString { ptr in
+            withUnsafeMutablePointer(to: &addr.sun_path) { p in
+                p.withMemoryRebound(to: CChar.self, capacity: 104) { dest in strlcpy(dest, ptr, 104) }
+            }
+        }
+        let connected = withUnsafePointer(to: &addr) { p in
+            p.withMemoryRebound(to: sockaddr.self, capacity: 1) { sp in
+                connect(fd, sp, socklen_t(MemoryLayout<sockaddr_un>.size))
+            }
+        }
+        return connected == 0
+    }
 }
 
 /// Minimal CLI/env option parsing — kept dependency-free (no argument-parser,
@@ -104,8 +141,16 @@ public enum CrowDaemon {
 struct DaemonOptions {
     var httpPort: Int = 8787
     var host: String = "127.0.0.1"
-    var socketPath: String = SocketServer.defaultSocketPath()
+    var socketPath: String = DaemonOptions.defaultDaemonSocketPath()
     var devRoot: String = FileManager.default.currentDirectoryPath
+
+    /// A DISTINCT default from the app's `crow.sock`, in the same owner-only dir,
+    /// so running `crowd` with defaults never unlinks/steals the desktop app's
+    /// socket. Sharing the app's path is opt-in via `--socket` (CROW-581 review).
+    static func defaultDaemonSocketPath() -> String {
+        let dir = (SocketServer.defaultSocketPath() as NSString).deletingLastPathComponent
+        return (dir as NSString).appendingPathComponent("crowd.sock")
+    }
 
     static func parse(_ arguments: [String]) -> DaemonOptions {
         var options = DaemonOptions()
@@ -130,7 +175,10 @@ struct DaemonOptions {
             case "--host": if let value = next { options.host = value }; index += 1
             case "--socket", "--socket-path": if let value = next { options.socketPath = value }; index += 1
             case "--dev-root": if let value = next { options.devRoot = value }; index += 1
-            default: break
+            default:
+                if flag.hasPrefix("-") {
+                    FileHandle.standardError.write(Data("[crowd] WARNING: ignoring unknown flag '\(flag)'\n".utf8))
+                }
             }
             index += 1
         }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
@@ -18,6 +18,16 @@ public enum CrowDaemon {
     public static func run(arguments: [String] = CommandLine.arguments) async throws {
         let options = DaemonOptions.parse(arguments)
 
+        // The WS endpoints are unauthenticated; the Origin guard blocks
+        // cross-site browser hijacking but not same-origin/native clients on the
+        // network. Binding beyond loopback exposes git + a shell — warn loudly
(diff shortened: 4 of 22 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Solid, well-scoped M0+M1 spike. The platform gating (`#if canImport(AppKit/WebKit/Glibc)`), the `CPty` shim for `openpty(3)`, the `deliverOnMainQueue:` opt-in on `PTYProcess`, and the verbatim reuse of `CommandRouter`/`TmuxController`/`GitManager` are all clean and the macOS path is untouched. The blocker below is about the new network attack surface, not the architecture.

### Critical Issue

**No authentication or `Origin` check on the WebSocket endpoints → cross-site WebSocket hijacking (effective RCE).**
`RPCWebSocketHandler.mount` (`RPCWebSocketHandler.swift:14`) and `TerminalWebSocket.mount` (`TerminalWebSocket.swift:13`) both accept the upgrade unconditionally (`router.ws(\"/terminal\") { _, _ in .upgrade() }`), discarding the request/context without inspecting `Origin`.

Binding to `127.0.0.1` by default does **not** mitigate this: browsers do not apply the same-origin policy to *outbound* WebSocket connections, so any web page a user visits while `crowd` is running can `new WebSocket('ws://127.0.0.1:8787/terminal')` and receive an interactive shell (tmux attaches to the default shell — `TerminalCockpit.attachCommand()`). `/rpc` is similarly reachable, exposing `add-worktree`, which shells out to `git` via `GitManager.createWorktree`. This is arbitrary command execution from any origin.

The upgrade closure already receives the request — the fix is small: reject the upgrade unless the `Origin` header is absent (native clients) or on an allowlist (e.g. `http://127.0.0.1:<port>`/`localhost`). Please add that before this lands, since the `/terminal` byte path is fully functional in this PR rather than a follow-up.

### Security Review
**Strengths:**
- Path-traversal guards on `add-worktree` (`RPCHandlers.swift:95,99`) reuse the shared `Validation.isPathWithinRoot` for both `path` and `repo_path`.
- `StaticAssets` basename guard (`StaticAssets.swift:20-23`) blocks `/` and `..`; router-decoded params are checked post-decode, so `%2e%2e` is caught.
- Session-name validation and `work`-only restriction on `new-session` mirror the app's contract.
- `shellQuote` in `TerminalCockpit` correctly single-quotes the tmux binary/socket/session args.

**Concerns:**
- **(Red, above)** Missing `Origin`/auth check on `/rpc` and `/terminal`.
- **(Yellow)** `--host` accepts `0.0.0.0` (`DaemonOptions`, `CrowDaemon.swift:110-113`) with no warning. Combined with the unauthenticated endpoints, that silently exposes a shell to the whole network. At minimum log a loud warning when binding to a non-loopback address; ideally gate the terminal/RPC WS behind a token when host isn't loopback.

### Code Quality
- **(Yellow)** No automated tests for the daemon (RPC round-trip, path-traversal rejection, static-asset guard). The PR was smoke-tested manually, which is reasonable for a spike, but the path-traversal and `Origin` guards in particular are cheap to lock down with a test and easy to regress. A small `CrowDaemonTests` target would be worth it.
- **(Green)** `DaemonOptions.parse` silently ignores a malformed `--http-port` (falls back to 8787) instead of erroring — minor, but a bad flag failing quietly can be confusing to operate.
- **(Green)** `TerminalControl` resize accepts `rows`/`cols` of 0 (negative `Int` clamps to 0 via `UInt16(clamping:)`); harmless but a `max(1, …)` floor would avoid a degenerate tmux resize.
- **(Green)** No cap on concurrent `/terminal` connections; each spawns a PTY + `tmux attach`. Fine for a spike, worth bounding later.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 2 Yellow, 3 Green] findings. The unauthenticated, no-`Origin`-check WebSocket terminal is the blocker; the localhost default does not close the cross-site hijacking vector.

---

dgershman:
Thanks for the review — all findings addressed in e944879 (daemon-only; macOS/shared untouched).

**🔴 Cross-site WS hijacking** — new `WebSocketOriginGuard` rejects the `/rpc` and `/terminal` upgrade unless `Origin` is absent (native clients) or a loopback host, wired into both `ws(shouldUpgrade:)` closures. Live-verified: cross-site `Origin: https://evil.com` → **400** (not upgraded) on both endpoints; loopback + no-Origin → **101**; real WS RPC still works.

**🟡 Non-loopback bind** — binding a non-loopback `--host` (e.g. `0.0.0.0`) now logs a loud WARNING that the endpoints are unauthenticated (git + shell exposure). Token auth for genuine remote access remains a documented follow-up.

**🟡 No tests** — added `CrowDaemonTests` (swift-testing): Origin allowlist (loopback vs cross-site vs absent, incl. the `null` opaque origin and a `127.0.0.1.evil.com` lookalike), loopback-host detection, the static-asset traversal guard (extracted to a testable `StaticAssets.isSafeAssetName`), option parsing incl. malformed `--http-port`, and the connection limiter. 10 tests / 4 suites, all green.

**🟢 Malformed `--http-port`** — now logs a warning instead of silently defaulting.
**🟢 Resize floor** — rows/cols floored at 1×1 via `max(1, …)`.
**🟢 Connection cap** — `TerminalConnectionLimiter` bounds concurrent `/terminal` PTY+attach connections (16).

Note: the guard is loopback-only by design for this spike, so the web UI is a localhost tool until token auth lands. Re-requested your review.

dhilgaertner (changes requested):
## Code & Security Review

**Re-review after e944879.** All six findings from the previous round are addressed and verified: the `Origin` guard is applied at both WS upgrade points (`RPCWebSocketHandler.swift:18`, `TerminalWebSocket.swift:17`) with the right semantics (absent Origin = native client allowed, `null`/unparseable rejected, loopback allowlisted); the non-loopback bind warning is loud (`CrowDaemon.swift:25-29`); `CrowDaemonTests` locks down the origin guard, path-traversal guard, option parsing, and connection limiter — I ran the suite locally, **10/10 pass**, and `swift build --product crowd` succeeds on macOS. The malformed-port warning, 1×1 resize floor, and the 16-connection `TerminalConnectionLimiter` are all in. Nice round-trip.

Two new issues surfaced this round, both shared-resource collisions the daemon introduces rather than flaws in the guard work:

### Critical Issues
None.

### Security Review
**Strengths:**
- `WebSocketOriginGuard` is correct against the realistic bypasses: `http://127.0.0.1.evil.com` (host suffix trick), `null` origin (sandboxed iframes), and case tricks are all rejected — and covered by tests.
- Path-traversal guards on `add-worktree` (`RPCHandlers.swift:95,99`) and the `/xterm/:file` basename guard (`StaticAssets.swift:39-41`) hold up; params are checked post-percent-decode.
- `TerminalCockpit.shellQuote` correctly single-quotes everything interpolated into the PTY command.
- Worktree creation goes through `Process` argument arrays (no shell) in `GitManager`.

**Concerns:**
- **(Yellow)** `crowd`'s default `socketPath` is the desktop app's socket, and `SocketServer.start()` unlinks it unconditionally (`SocketServer.swift:51`). Start `crowd` with defaults on a macOS box where the Crow app is running and the daemon silently steals `~/.local/share/crow/crow.sock` — the app keeps serving a socket that no longer exists, and every `crow` CLI call is silently rerouted to the daemon (which supports only 3 methods). Worse, both processes now hold independent in-memory copies of `store.json` and rewrite the whole file on `mutate` (per-process cache + atomic whole-file write = last-writer-wins), so sessions/worktrees created in one process can be silently dropped by the other. Cheap fixes: probe-connect before unlinking and abort/warn if something answers, or give the daemon a distinct default (`crowd.sock`) and make sharing the app's path an explicit opt-in flag.
- **(Yellow)** The cockpit tmux socket is a fixed name in shared temp: `NSTemporaryDirectory() + "crowd-tmux.sock"` (`TerminalCockpit.swift:18`). On Linux — the entire point of this PR — that's `/tmp/crowd-tmux.sock`, world-writable directory, predictable name. Any other local user can squat the path (pre-create a file/socket/symlink) and permanently break `/terminal` for the victim (tmux's bad-owner check prevents outright session hijack, but the DoS stands), and two different users on one host can't both run `crowd` at all. The repo already has the right spot: `~/.local/share/crow/` is created `0700` by `SocketServer.defaultSocketPath()` — put the tmux socket there (or under `XDG_RUNTIME_DIR`).
- **(Green)** The origin allowlist admits *any* loopback origin on any port, so any other dev server on the user's machine (or an XSS in one) can drive `/rpc` and `/terminal`. The code comment acknowledges this tradeoff; pinning to the daemon's own `host:port` would tighten it when you do the auth follow-up.
- **(Green)** `add-worktree`'s `branch` reaches `git ls-remote --heads origin <branch>` as a positional (`GitManager.swift:74`); a leading-dash value like `--upload-pack=…` is option injection → command execution. Anything that passes the origin guard already has a full shell via `/terminal`, so this adds no marginal exposure today — but reject leading-`-` branch names in the handler as defense-in-depth before token-authed remote access lands.

### Code Quality
- **(Green)** `/rpc` silently drops undecodable frames (`RPCWebSocketHandler.swift:29-31`); JSON-RPC 2.0 calls for a `-32700` parse-error response. Encode failures are also swallowed. Fine for M0, worth fixing when a real client depends on it.
- **(Green)** `add-worktree` doesn't check that `session_id` names an existing session, so it persists an orphaned worktree row (`RPCHandlers.swift:110-115`).
- **(Green)** `DaemonOptions.parse` silently skips unknown flags (`default: break`), so a typo like `--http-prot 9001` is ignored without a peep — the malformed-value warning you added for `--http-port` sets the right precedent; a similar warning for unknown flags would round it out.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 5 Green] findings. The previous round's blocker is fully resolved and test-covered; what remains is the app-socket theft/dual-writer hazard and the world-shared `/tmp` tmux socket on Linux — both small, contained fixes.

---

dhilgaertner (approved):
## Code & Security Review

**M0+M1 `crowd` daemon** — reviewed the full diff, built `swift build --product crowd` (clean), and ran the daemon test suite (14/14 pass). This is a well-scoped, unusually security-conscious spike: the sensitive guards each ship with a comment explaining the threat and a locking-down test.

### Critical Issues
None.

### Security Review
**Strengths:**
- **Cross-site WebSocket hijacking** is correctly the headline threat and is handled at the upgrade for both `/rpc` and `/terminal` (`WebSocketOriginGuard`, `RPCWebSocketHandler.swift:453`, `TerminalWebSocket.swift:737`). `URL(string:)`-based host parsing rejects `127.0.0.1.evil.com`, `null`, and userinfo tricks — verified against the test matrix.
- **Path traversal** guarded twice: `StaticAssets.isSafeAssetName` (post-percent-decode basename check) for `/xterm/*`, and `Validation.isPathWithinRoot` (resolves `..`, prefixes with a trailing `/` so `/root-evil` can't match) for worktree + repo paths.
- **git option injection** closed by the leading-dash branch guard (`RPCHandlers.swift:392`); git is invoked with argument arrays, not a shell, so no metacharacter exposure.
- **Socket hijacking** avoided: distinct default `crowd.sock`, plus a live-connection probe (`socketInUse`) before binding so the desktop app's `crow.sock` is never unlinked/stolen.
- **DoS surface** bounded: `TerminalConnectionLimiter` (16) caps concurrent PTYs; tmux control socket lives in `$XDG_RUNTIME_DIR`/`~/.local/share/crow` (0700), never world-writable `/tmp`; resize floored at 1×1.
- Non-loopback bind emits a loud unauthenticated-exposure warning; default is `127.0.0.1`.

**Concerns (all within the documented spike threat model — non-blocking):**
- The trust model is "any loopback origin + any origin-less client is trusted," so any other localhost web app (any port) or native process on the host can drive `/terminal` (interactive shell) and `/rpc` (git). This is explicitly documented in `WebSocketOriginGuard` and the PR body, with token auth named as the follow-up. Acceptable for a loopback-only M0/M1; worth keeping front-of-mind before any remote bind ships.
- `isPathWithinRoot` resolves `..` but not symlinks — a pre-existing symlink under `devRoot` could point git outside it. Low risk for a spike; consider `resolvingSymlinksInPath()` when this hardens.

### Code Quality
- `PTYProcess.deinit` dropping the `readQueue.sync` wrapper is correct and arguably a latent-bug fix: at deinit refcount is 0, so no self-capturing `readQueue.async` block or in-flight `[weak self]` event handler can be running, and the old `dispatch_sync`-onto-current-queue path could trap on the daemon's short-lived PTYs. Reasoning is sound; macOS behavior unchanged (`deliverOnMainQueue: true` default preserved).
- Linux port is clean: `CPty` shim for `openpty(3)`, per-platform `read`/`write`/`ioctl` (the `UInt(TIOCSWINSZ)` glibc cast is the right call), and AppKit/WebKit surfaces gated behind `#if canImport` with matching `Package.swift` platform-conditional linking. macOS branch is untouched.
- Reuses `CommandRouter`, `TmuxController`, and the exact xterm 6.0.0 bundle (all four referenced assets confirmed present) rather than duplicating — the stated goal of the epic.
- Minor: `RPCWebSocketHandler` silently drops undecodable frames (no JSON-RPC parse-error response) — fine for the spike.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. Both considerations are the spike's explicitly-documented, in-scope follow-ups (auth + symlink hardening), not defects. Build and all 14 tests pass on macOS; the `swift:6` Linux build of the Glibc/`CPty` branch remains the correct pre-merge check as noted in the PR.

PR #590, opened 2026-07-04
Cross-platform web UI: M2 + M3 boards + desktop parity (consolidated)

dgershman:
Part of #581. Tracked by #591. Consolidates the cross-platform web UI work into a single PR now that the M0+M1 daemon (#584) has merged to `main`.

**Base re-targeted to `main`.** Absorbs the previously-stacked #585 (M2, closed) and #589 (Settings → Automation toggle, closed) — both are included here.

## What's in it
- **M2** — web UI: session sidebar + terminal tabs (over the `/rpc` + `/terminal` WebSockets)
- **M3** — boards: Ticket Board / Reviews / Allowlist, forward-to-app
- **Desktop parity** — brandmark + Tickets card + Reviews/Allowlist/Manager nav pills; status-hued session rows with a trailing glowing dot + remote-control glyph; PR badge on every session with a PR; right-click context menus (+ suppressed browser menu over the terminal); detail-header inline PR status + gold PR chip + right-aligned actions + right-aligned worktree path; mobile single-pane + floating hamburger + terminal touch-scroll
- **Live per-session endpoint** — batched `list-sessions-live` (remote-control active/available + PR status + memory-only PR links)
- **Dev ergonomics** — `--web-dir` live web assets + `make crowd-dev` watch/rebuild/restart
- **Settings → Automation** toggle for auto permission mode on new coder views — **Closes #586**

## Verification
`make build` (app + daemon) green; `swift test --package-path Packages/CrowDaemon` passes; web served/verified via `--web-dir`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01X7iKQZ8EGJswmgVCbd3sxc

Issue #15, opened 2026-04-03
Notification system with configurable sounds and macOS notifications


dhilgaertner:
## Summary

Add a notification system that alerts users to important events via sound effects and native macOS notifications.

## Features

- **Sound notifications** — play a configurable sound when notable events occur (e.g., task completion, errors, session state changes)
- **macOS native notifications** — integrate with the macOS notification center so users get alerts even when the app isn't focused
- **Configurable** — users should be able to:
  - Enable/disable sound notifications independently from system notifications
  - Choose from a set of built-in sounds or provide a custom sound file
  - Configure which event types trigger notifications
  - Set a "do not disturb" / mute mode

## Event Types to Consider

- Session task completed
- Build/test failures
- Agent waiting for input
- Session errors or crashes
- Worktree operations completed

## Implementation Notes

- Use the macOS `NSUserNotification` / `UNUserNotificationCenter` API (or Electron/Tauri equivalent)
- Sound playback should be non-blocking
- Respect system-level Do Not Disturb settings
- Store notification preferences in user settings

Issue #14, opened 2026-04-03
Animated fire effect behind branding logo driven by session activity


dhilgaertner:
## Summary

Add an animated fire/flame effect behind the branding logo in the right-hand column. The intensity of the effect should be driven by aggregate activity across all sessions.

## Design Concept

- **Idle state** — a subtle, warm glow behind the logo (embers / smoldering)
- **Low activity** — the glow intensifies, hints of small flickering flames appear
- **Medium activity** — visible flames licking up around/behind the logo
- **High activity (many active sessions)** — a roaring fire effect, fully animated flames

The idea: when you're doing a lot of work across many sessions, the fire really gets going. It's a fun, ambient indicator of how "hot" things are.

## Implementation Notes

- Activity signal should be aggregate across all sessions (terminal output, tool calls, etc.)
- Animation should be smooth and performant — consider CSS animations, canvas, or a lightweight WebGL approach
- Should degrade gracefully (no fire is fine, never broken UI)
- Intensity transitions should be gradual, not abrupt
- Configurable or disable-able for users who prefer a calmer UI

Issue #13, opened 2026-04-03
Security audit & codebase review for open-source readiness


dhilgaertner:
## Summary

Perform a thorough audit and review of the entire codebase in preparation for open-sourcing. This includes security review, refactoring opportunities, and general code quality improvements.

## Scope

- **Security audit** — scan for hardcoded secrets, credentials, API keys, or sensitive configuration that shouldn't be public. Review auth flows, input validation, and any network-facing code for OWASP top-10 vulnerabilities.
- **Dependency review** — audit dependencies for known CVEs, unnecessary packages, and license compatibility with our intended open-source license.
- **Refactoring opportunities** — identify code that could benefit from cleanup, better abstractions, or improved structure before external contributors see it.
- **Documentation** — ensure README, contributing guidelines, and code comments are appropriate for an open-source audience.
- **Git history** — check for any sensitive data that may have been committed historically and needs to be scrubbed before going public.

## Motivation

We're preparing to make this repository open source. Getting ahead of security issues and code quality concerns now avoids embarrassment and risk after the repo is public.

PR #594, opened 2026-07-05
Web UI parity: config fallback, remove-link, agent menu, settings modal, multi-select (CROW-593)

### LINKED ISSUE #593: Cross-platform web UI: continue parity + land pending fixes (follow-up to #590)
Follow-up to the cross-platform web UI epic #581 / tracking #591. #590 (M2 + M3 boards + desktop parity + web Settings modal) is merged. This ticket tracks the remaining web-UI parity work and post-merge verification.

## Already implemented (on branch `feature/crow-581-settings-app-fallback`, NOT yet in a PR)
Commits: `d942a1b`, `bbfe087`, `2e5ebf5`.
- **Daemon disk-fallback** for `get-config`/`set-config`: when the desktop app is down **or too old to know the method**, read/write `{devRoot}/.claude/config.json` directly. Fixes the web Settings modal showing `Unknown method: get-config` during a daemon-ahead rollout.
- **Detail-header parity**: issue/PR chips + inline PR status and the action buttons now share one line (chips left, buttons trailing); every action + PR quick-action button has a leading inline-SVG icon (monochrome, inherits color) matching the desktop SF Symbols.
- **Tickets card glyphs**: per-status icons (tray / flag / bolt / eye / check-circle) replacing plain dots; the 5 counts are clustered + centered.
- **Allowlist**: "Select All" + "Clear" for the Promote-to-Global flow (matches the desktop's multi-select).

**Next step:** land these via a PR (reopen #592 or open a fresh one) → review → merge.

## Post-merge verification (stale-app symptoms)
The running (pre-#590) app returns `Unknown method` for every forwarded call, so these need the **merged app relaunched** to confirm — none are web bugs:
- [ ] `create-manager` (the sidebar **+** button) works.
- [ ] Ticket / Reviews / Allowlist boards populate.
- [ ] Live PR surfaces on the linked session/ticket (e.g. issue #36 → PR #37).
- [ ] Settings modal shows `app_running: true` (banner gone) and writes flow through the app's `saveSettings`.

## Open follow-ups / polish
- [ ] Reword the Settings "Desktop app not running" banner to also cover "outdated / unreachable for config" (it fires whenever the daemon falls back to disk, not only when the app is down).
- [ ] Visually verify the new header icons + tickets glyphs in a browser and tune spacing/paths.
- [ ] Allowlist **delete**: does not exist anywhere today (AllowListService / app RPC / daemon / web) — the desktop only *promotes*. Add net-new (service method → app handler → daemon forward → web UI) only if we actually want it.

## Larger #591 items still open
- [ ] Token/auth gate for remote (non-loopback) access.
- [ ] WebSocket push instead of polling.
- [ ] Host-side actions (Open in VS Code / Open Terminal).
- [ ] Electron/Tauri desktop shells over a HostBridge.
- [ ] Terminal persistence + window adoption across restarts.

Refs: #581, #591, #590.

dgershman:
Follow-up to #590 (merged); continues the cross-platform web-UI parity work on a fresh branch off `main`. Supersedes closed #592. Kept as a **draft** — not for auto-merge, intentionally not labeled `crow:merge`.

## Landed the three pending #592 commits
- **Daemon config disk-fallback** (`RPCHandlers.swift`): `get-config`/`set-config` fall back to `{devRoot}/.claude/config.json` when the app is down **or** too old to know the method (`-32601`).
- **Detail-header parity** + **Tickets card glyphs** + **Allowlist select-all** (web assets).

## Parity fixes + new features (this PR)
- **remove-link command** — CLI `RemoveLink` + app `remove-link` RPC to detach a stale session link (links were only cleared on session-delete).
- **#6 set-locked** — added the missing daemon forwarder; the web Lock/Unlock no longer fails with "unknown method".
- **#4 allowlist** — auto-runs `refresh-allowlist` on open (populates without a manual Refresh) + sticky board header.
- **#8 cards** — full green/orange background tint by state (done/attention), matching the desktop rowBackgroundColor, not just the left edge.
- **#9 favicon** — added, reusing `brand.svg`.
- **#2 new-manager agent menu** — new `list-agents` RPC (app → daemon forward) backs a web context menu on the "+" button to pick which agent to launch (passes `agent_kind`, which create-manager already honors).
- **#7 settings modal** — job/workspace editors now (a) scroll within a consistent modal instead of clipping, (b) open as a **stacked child modal**, and (c) repo/label lists render as **chips**; plus the app-down banner reworded to cover the "outdated" fallback.
- **#5 multi-select** — sidebar checkmark toggle → per-row checkboxes + per-section All/Clear + a "N selected / Delete(N)" bulk bar, mirroring the desktop.

**#3 (rename):** already correct in source (`app.js` calls `rename-session`); the reported failure was the stale running daemon serving old web assets — no code change needed.

## Testing
- `swift test --package-path Packages/CrowDaemon --filter ConfigForwarderTests` — 4/4 pass.
- `make build` clean (`crow` + `CrowApp` + `crowd`) after each Swift change.
- `node --check` clean on `app.js` / `settings.js`.
- **Needs a `crowd` relaunch on this branch to exercise the web/daemon changes in a browser** (the running daemon predates them).

Closes #593

https://claude.ai/code/session_016awThNaRD1JV33NotPjs3V


dhilgaertner (changes requested):
## Code & Security Review

Reviewed the full diff vs `main` (201 files, +17.3k/−11.6k). Build is clean (`swift build` links `crow` + `crowd`) and every package test suite passes (CrowDaemon 74, CrowCore 343, CrowEngine 245, CrowIPC 35, CrowPersistence 31, CrowCLI 43, CrowTerminal 83). `node --check` clean on `app.js`/`settings.js`. The web UI is disciplined (`textContent`-only rendering, no HTML-injection XSS, auth token stays in an `HttpOnly; Secure; SameSite=Lax` cookie and never touches JS), password hashing is textbook (PBKDF2-HMAC-SHA256 @210k, per-hash CSPRNG salt, constant-time compare), and both WS upgrades enforce Origin + auth inline. Good work overall — but there are two blocking issues and a batch of should-fix items.

### Critical Issues

**Red 1 — `get-state` leaks daemon secrets to any authenticated `/rpc` client.**
`RPCHandlers.swift:773-787` builds `DaemonStateSnapshot(appState:config: ConfigStore.loadConfig(...))` and returns it verbatim; `DaemonStateSnapshot.config` (`DaemonStateSnapshot.swift:80`) is the **raw** `AppConfig`, so `jiraCredential.tokenRef`, per-workspace/manager gateway bearer `customHeaders`, and the `webAuth` PBKDF2 hash+salt all go over the wire. This bypasses `SettingsSecrets.strippedForTransport`, which the sibling `get-config` handler carefully applies (`RPCHandlers.swift:803`) and whose doc comment says these fields "must not be sent" to a browser. Failure scenario: a user logged in over the tailscale/ngrok proxy with the web password opens the browser console and calls `{"method":"get-state"}` on the already-open `/rpc` socket — they receive the Jira API token and AI-gateway auth headers in plaintext, plus an offline-crackable password hash. Fix: run the snapshot's config through `SettingsSecrets.strippedForTransport` (and drop `webAuth`) before encoding.

**Red 2 — curated terminal handlers shadow the engine implementations, regressing the primary workspace-creation flow.**
`CommandRouter.handle` (`CommandRouter.swift:27-32`) only falls through to the engine router when a method is **absent** from the curated map. The daemon curated map defines `new-terminal` (`RPCHandlers.swift:211`), `list-terminals` (`:189`), and `close-terminal` (`:268`), so the richer engine versions (`EngineRouter.swift:624-789`) are never reached. Since the macOS app is retired (ADR 0007/0008), this daemon router is the **only** server, so this is a regression from `main`:
- The curated `new-terminal` never calls `TmuxBackend.registerTerminal` (its own comment at `:208` says "readiness/sentinel wiring is app-only and skipped"). `crow send` to such a terminal then routes through `TmuxBackend.sendText`, which throws for an unregistered id, the error is swallowed, and the RPC still returns `sent: true`. Auto-respond and quick-actions (`canSend`) silently skip until a daemon restart re-adopts the row.
- Curated `list-terminals` (`:189-204`) omits the `readiness` field the engine version returns. `skills/crow-workspace/setup.sh` polls exactly that field, so every workspace creation stalls the full readiness timeout and logs "agent launch not confirmed" even on success.
- Curated `new-terminal` ignores the `--cwd` param, and skips `resolveClaudeInCommand`, RC injection, hook-config writing, and the `CROW_ARTIFACTS_DIR`/`CROW_SESSION_ID` env — so hooks/activity dots and the Images panel are dead for CLI-created work terminals.
Fix: delete the curated `new-terminal`/`list-terminals`/`close-terminal` so the engine fallback serves them, or port the missing behavior into them.

### Security Review

**Strengths:**
- PBKDF2-HMAC-SHA256 @210k iterations, fresh 16-byte CSPRNG salt per hash, genuine constant-time compare (`WebAuth.swift:31-48`); 256-bit CSPRNG session tokens, server-issued, in-memory (`WebAuth.swift:73-98`).
- Path-traversal guards with tests on both file routes (`StaticAssets.swift:50-52`, `Artifacts.swift:62-65` — basename-only + UUID-validated session segment).
- No shell injection in terminal attach — tmux binary/socket/group name are all shell-escaped; resize clamped ≥1×1 (`TerminalCockpit.swift:50-53, 85-87`).
- Config written `0o600` in a `0o700` dir; the web-password hash/salt are stripped on transport and restored on save so the browser can't read or blank them (`SettingsSecrets.swift`, `ConfigStore.swift:80-89`).

**Concerns (Yellow):**
- **Loopback fast-path trusts a proxy that omits `X-Forwarded-For`** (`WebAuth.swift:170-176`): a co-located TLS terminator/forwarder (`ssh -L`, `socat`, `stunnel`, nginx without `proxy_set_header`) makes every remote request look local and silently defeats the configured web password.
- **Cross-site WebSocket hijack via any private-network origin** (`WebSocketOriginGuard.swift:53-81`): on a `0.0.0.0` bind with no password, `isAllowedOrigin` accepts any private IP, so an attacker page on the same LAN/tailnet can open `ws://<victim>:8787/terminal` and get a shell. Documented as unauthenticated-by-choice, but the CSWSH angle is worth hardening.
- **Agent-planted SVG served inline** (`Artifacts.swift:15, 53-73`): `.svg` is served `image/svg+xml`, same-origin, no CSP, no `Content-Disposition: attachment`. A prompt-injected review agent can drop `pwn.svg` in its scratch dir; opening `/artifacts/<uuid>/pwn.svg` executes script in the daemon origin, which can drive same-origin `/rpc` (git). Force attachment or deny SVG.

### Code Quality (Yellow — should fix)

Daemon RPC / engine:
- `set-locked` bumps `updatedAt` (`RPCHandlers.swift:443, 447`); the engine `setLocked` deliberately leaves it untouched to preserve the retention clock — lock/unlock now resets the auto-cleanup age.
- `rename-session` shadows the engine handler (`RPCHandlers.swift:454-473`), so the CROW-593 remote-control rename sync this PR adds (`EngineRouter.swift:142-153`, routes through `renameSession` to push `/rename`) is unreachable — the exact bug the PR claims to fix.
- `mark-in-review` (`RPCHandlers.swift:884-891`) only flips local status and drops the provider ticket transition (`tracker.markInReview` → Jira/GitHub "In Review"), even though a live tracker is in scope — reintroduces the drift `resync-jira` exists to repair.
- Unsynchronized `config.json` read-modify-write across `set-config`, `set-web-password`, and `onJobRan` (`RPCHandlers.swift:810-854`, `CrowDaemon.swift:165-172`) with concurrent off-main-actor handlers → last-writer-wins; a scheduled job firing during a Settings save can clobber the edit or a just-set web password.
- `JSONStore.reload()` reads+decodes the file **outside** the lock then assigns `_data` (`JSONStore.swift:85-93`); a concurrent `mutate` can be rolled back in memory and then re-saved over disk, silently losing a session/terminal.

Web UI:
- Link `href` assigned from server data with no scheme allowlist (`app.js:1231, 1560`): a prompt-injected `crow add-link --url "javascript:…"` or a non-https provider URL renders a clickable chip; `rel="noopener"` blunts `javascript:` but `data:` still navigates. Add `if (!/^https?:/i.test(url)) skip`.
- `refreshTerminals` applies the response without checking `selectedId` still matches (`app.js:1438-1450`): fast A→B session switch shows session A's terminal (and switches the shared `/terminal` stream to A's window) under B's header.
- Pending RPCs are never rejected on socket close and a reconnect race can leak a second `/rpc` socket (`app.js:41-55`): in-flight actions freeze for the 10s timeout after a blip, and duplicate sockets double-fire `changed` refreshes.
- Settings Escape handler tears down the whole modal instead of the stacked sub-form overlay (`settings.js:153`) — conflates sub-form vs other-tab dirty state.
- Paste bypasses bracketed-paste mode (`app.js:1921-1928`, `terminal.html:111-113`): pasting copied text with hidden newlines into a shell auto-executes each line. Use `term.paste(text)`.

Desktop (Tauri):
- Sidecar reuse check hardcodes port 8787 and ignores `CROW_HTTP_PORT`/`CROW_SOCKET` (`crow-desktop/src-tauri/src/lib.rs:17, 114-122`): launching the desktop app while a custom-port crowd runs spawns a second crowd contending on the same `store.json`/socket/tmux — the exact double-daemon case the file header warns against.
- Remote content on `http://127.0.0.1:8787` is granted Tauri IPC (`opener`, `notification`) with `csp: null` and `withGlobalTauri: true` (`capabilities/default.json`, `tauri.conf.json:10, 21`), and `port_open()` trusts anything on 8787: any single XSS in the crowd web UI (which renders GitHub/GitLab-sourced strings) escalates to native IPC. Scope the capability and consider a handshake before `navigate`.
- crowd is torn down with `Child::kill()` / SIGKILL on exit (`lib.rs:170-172`): no chance to remove `crow.sock`/tmux windows/flush the store → stale socket + orphaned tmux on next launch; on app crash `RunEvent::Exit` never fires and crowd orphans. Prefer SIGTERM + bounded wait.

Scripts:
- `crowd-dev.sh:42-43` (watchexec branch) flattens the start command with `${START_CMD[*]}`, discarding array quoting; a repo path or `CROW_SOCKET` containing spaces/metacharacters word-splits or gets shell-evaluated. The poll branch (`:61`) correctly uses `"${START_CMD[@]}"`.

Also noted (Green — consider): `remove-link` with both `--id` and `--url` removes the union not the intersection (`EngineRouter.swift:901-903`); `work-on-issue`/tracker inject a URL into the Manager terminal without stripping control chars (`RPCHandlers.swift:671-686`, `CrowDaemon.swift:286`); `create-manager` skips the AgentRegistry gate the engine version enforces (`RPCHandlers.swift:867-877`); `autoReviewed`/session-token/`/rpc`-buffer sets are unbounded; CI never builds/clippy-checks the new Rust crate; the "forward to the app" comments throughout `RPCHandlers.swift` are stale under ADR 0007 and actively mislead about two-writer safety. `cpty.c` is a Linux-only header shim (no executable surface); the `PTYProcess` deinit `readQueue.sync` removal is sound; vendored xterm addons look like genuine official UMD builds.

### Summary Table

**Recommendation:** Request Changes — driven by 2 Red, 17 Yellow, ~12 Green findings. The two Reds are the priority: `get-state` must strip secrets, and the shadowed terminal handlers must delegate to (or port from) the engine so `crow send`, readiness, and hooks work for the sole server. The rename-session / mark-in-review / set-locked shadows are notable because they silently defeat parity fixes this same PR advertises.

---

dgershman:
Thanks for the thorough review. Addressed the two blocking issues and the clearly-correct fixes in `90f59b2`; below is what changed and what I'm deferring (with reasoning) so nothing is silently dropped.

## Fixed in `90f59b2`

**Red 1 — get-state secret leak.** `get-state` now runs the snapshot config through `SettingsSecrets.strippedForTransport` (the same path `get-config` uses), zeroing `jiraCredential.tokenRef`, gateway `customHeaders`, and the `webAuth` hash/salt before encoding. The existing `getStateReturnsDecodableSnapshotOfLocalAppState` test stays green.

**Red 2 — shadowed terminal handlers.** Deleted the curated `new-terminal` / `list-terminals` / `close-terminal` so `CommandRouter` delegates to the engine. Verified the engine `new-terminal` persists to the store (`capturedStore.mutate`, `EngineRouter.swift:740`) — so it's reseed-safe — and adds the readiness tracking, `registerTerminal`, RC injection, `--cwd`, and hooks/env the curated stubs skipped. `crow send` / readiness / auto-respond now work for CLI-created terminals.

**Shadowed parity fixes.**
- Deleted curated `rename-session` → the engine's `renameSession` runs, which pushes `/rename` (the CROW-593 sync this PR advertises). Confirmed `SessionService.renameSession` persists *and* calls `syncRemoteControlName`.
- `set-locked` no longer bumps `updatedAt`, preserving the auto-cleanup retention clock.

**Web / security.**
- Link chips gated to `http(s)` at both render sites — a `javascript:`/`data:` url can't become a clickable href.
- Terminal paste routes through `term.paste()` (bracketed-paste) in `app.js` and `terminal.html`.
- Artifact SVGs are served `Content-Disposition: attachment` so a planted `pwn.svg` can't render as a top-level document and script into the daemon origin.
- `crowd-dev.sh` watchexec branch preserves array quoting via `sh -c '… "$@"'`.

Verified: `swift build` links crow+crowd; all 74 CrowDaemon tests pass (incl. get-state snapshot decode + terminal-op validation); `node --check` + shellcheck clean.

## Deferred — tracked follow-ups

Valid findings I'm intentionally not rushing into a review-response commit, because they need dedicated design + testing:

- **`mark-in-review` / `create-manager` curated shadows** — same delegation pattern, but I want to verify engine parity (the provider ticket transition; the AgentRegistry gate + tmux spawn) with manager-create testing before deleting, rather than risk those flows here.
- **Concurrency** — the unsynchronized `config.json` read-modify-write (`set-config` / `set-web-password` / `onJobRan`) and `JSONStore.reload()` decoding outside the lock. Needs a config write-lock + reload-under-lock design with tests; it touches the persistence core.
- **Web resilience** — `refreshTerminals` selectedId guard, pending-RPC rejection on socket close + reconnect de-dupe, and the Settings Escape sub-form scoping. Grouping into one web-robustness follow-up.
- **Security hardening** — loopback `X-Forwarded-For` trust, CSWSH via private-network origin, and the Tauri IPC capability scoping / SIGTERM-on-exit / port-from-env. Several are documented unauthenticated-by-choice; the Tauri items need a libc/nix dep + a desktop rebuild/test pass. Worth a dedicated hardening change.

Re-requesting review on the fixed set.

dhilgaertner (changes requested):
## Code & Security Review

Reviewed the full 201-file diff vs `main` (crowd daemon + web UI cutover, CrowEngine extraction, Tauri wrapper, macOS app retirement) with a security-focused pass over the auth/WS surface, RPC handlers, web frontend, and the Tauri/CLI/terminal/build layer.

**Build & tests:** root `swift build` clean (`crow` + `crowd`); all package suites pass — CrowDaemon 74, CrowEngine 245, CrowCore 343, CrowTerminal 83, CrowIPC 35, CrowPersistence 31; `node --check` clean on `app.js`/`settings.js`.

**Prior review round (90f59b2) verified complete:** `get-state` now strips secrets via `SettingsSecrets.strippedForTransport` (`RPCHandlers.swift:659-661`); the shadowed curated handlers are gone (no duplicate registrations remain in either router); SVG artifacts are forced `Content-Disposition: attachment`, link chips are http(s)-gated at both render sites, and paste routes through bracketed-paste `term.paste()`.

### Critical Issues

None — no Red findings. No exploitable XSS, auth bypass in a correctly-deployed configuration, path traversal, or command injection was found.

### Security Review

**Strengths:**
- Web auth is well-built: PBKDF2-HMAC-SHA256 at 210k iterations with fresh CSPRNG salts, constant-time compare, 256-bit server-side session tokens, `HttpOnly; Secure; SameSite=Lax` cookies, rate-limited login, and `X-Forwarded-*` trusted only from loopback peers.
- Disciplined DOM construction in the web UI: every dynamic string (ticket/PR titles, authors, branches, session names, labels) flows through `textContent`; `innerHTML` is confined to clears and static SVG constants — no XSS sink found across ~3,000 lines.
- Secrets are centrally handled in `SettingsSecrets` and consistently applied on `get-config`/`set-config`/`get-state`; the browser round-trip can neither read nor blank stored credentials.
- Both `/rpc` and `/terminal` WS upgrades require Origin guard **and** auth; path-traversal guards on `/xterm/:file` and `/artifacts/:session/:file` (UUID segment + image-extension allowlist, checked after percent-decoding) hold.
- Tauri permission surface is minimal (no shell/fs/http plugins); sidecar spawn uses an absolute path with fixed args.

**Concerns (Yellow — should fix):**

1. **`WebAuth.swift:173-176` — a set web password is silently bypassed by any non-XFF loopback forwarder.** `ssh -L`, `socat`, or a cloud port-forward presents a loopback peer with no `X-Forwarded-For`, so every remote request is authorized as `"local"` — unauthenticated `/rpc` and `/terminal` despite the operator having set a password. The header comment acknowledges this; it deserves at least a loud startup warning when `webAuth` is configured, and a docs callout.
2. **`WebAuth.swift:118-122` — `SessionStore.prune()` has no caller.** Expired tokens are only deleted when individually re-checked, so a long-running daemon accrues abandoned login tokens indefinitely. Wire `prune()` into a periodic loop or remove it.
3. **`EngineRouter.swift:78-91` — engine `set-config` returns the saved config with secrets unstripped and applies the incoming config without `preservingSecrets`.** Today it's shadowed by the daemon's correct handler (`RPCHandlers.swift:696-712`), so it's unreachable from the web — but if this router is ever the web-facing surface, an authenticated `set-config` both leaks the Jira token/gateway headers in the response and wipes stored secrets (including `webAuth`) on save. Make it symmetric with the daemon handler.
4. **`RPCHandlers.swift:556-564` (also `EngineRouter.swift:321-327`) — `work-on-issue` sends an unvalidated `url` as terminal keystrokes.** A `url` containing newlines/control characters injects extra submitted lines into the Manager agent (`/crow-workspace \(url)\n`). Validate it as a well-formed http(s) URL with no control characters, mirroring the `branch` leading-dash guard.
5. **`app.js:1825-1831` — `terminalWebFontURL` is injected verbatim as a stylesheet `<link href>` with no scheme/host validation** (and no CSP to backstop it). A config-writing agent on the unix socket — the same threat model the link-chip gating addresses — can load an attacker-controlled stylesheet into the operator's browser (UI spoofing/overlay of the settings password field). Require `https:` at minimum.
6. **No Content-Security-Policy on any page** (`index.html`/`login.html`/`terminal.html`, no header server-side). The entire XSS posture rests on the `textContent` convention; one future `innerHTML` slip on a ticket title becomes drive-the-daemon (`new-terminal`/`send`). A `default-src 'self'`-style policy is cheap defense-in-depth.
7. **`crow-desktop/src-tauri/capabilities/default.json:6-8` + `lib.rs:114-115` — the privileged webview trusts whatever answers on 127.0.0.1:8787.** The app grants remote IPC (`core:default`, `opener:default`, `notification:default`, plus `withGlobalTauri: true`) to `http://127.0.0.1:8787` and reuses any pre-bound listener without verifying it's crowd. On a multi-user machine, another local user binding 8787 first gets `window.__TAURI__` in the Crow-branded window. Handshake (e.g. `GET /health`/`version.json` shape check) before `navigate()`, and drop `withGlobalTauri` if the web UI doesn't use IPC yet.

**Code quality (Yellow — should fix):**

8. **`app.js:24-54` — `rpc()` hangs forever when the daemon is unreachable.** `rpcConnect()`'s promise only settles in `onopen`; the 10s timeout is armed only *after* `await rpcState.ready`. A Settings save during a daemon restart leaves the button stuck at "Saving…" permanently. Reject the connect promise in `onclose` (or arm the timeout before awaiting).
9. **`app.js:41,47` — reconnect race leaks parallel `/rpc` sockets.** An `rpc()` call inside the 1s reconnect window creates a connection that the timer then overwrites; the orphan still processes `changed` notifications (duplicate refresh fan-out) and spawns its own reconnect chain. Guard the timer with `if (!rpcState.ready)` or close the superseded socket.
10. **Three unsynchronized read-modify-write writers of `config.json`:** `set-config` (`RPCHandlers.swift:705-708`), `set-web-password` (`:734-735`), and `JobScheduler.onJobRan` (`CrowDaemon.swift:168-171`) each do load→mutate→save with no shared lock and on different actors. A job finishing while the user saves Settings loses one side's write (e.g. a dropped `lastRunAt` re-fires the job).
11. **`lib.rs:17` — hardcoded `PORT: 8787` ignores `CROW_HTTP_PORT`,** so a user running crowd on a custom port (as `scripts/crowd-dev.sh` supports) gets a *second* crowd spawned against the same devRoot — the exact store/tmux contention the file's own header says must not happen. The spawned child also gets no `--socket`, contending on the default unix socket.
12. **`lib.rs:170-172` — sidecar teardown is SIGKILL with no graceful shutdown or reap.** crowd is a stateful daemon (store writes, tmux ownership, live PTYs); quitting mid-write kills it un-cleanly every time. SIGTERM → brief wait → kill, then `wait()`.
13. **No CI for `crow-desktop`:** `ci.yml` never builds the new Rust crate, so a PR breaking it merges green.
14. **Deleted root test suites whose logic still exists have no replacements:** `TmuxCrashRecoveryTests` (→ `SessionService.handleTmuxServerCrash`), `CommandLaunchesTokenTests` (→ `AgentLaunch.commandLaunchesToken`), `DeferredLaunchTests`, most of `ManagerHookConfigTests` (incl. the 0600 invariant), and `AttributionSkillTests` (the `${CROW_AGENT_DISPLAY_NAME}` regression guard) all covered code that moved to CrowEngine, where zero tests now reference it. The orphan-reaper got proper replacement tests, so this looks like oversight rather than policy.
15. **`scripts/crowd-dev.sh:31` — the watch list omits `Packages/CrowEngine`** (and the agent/provider packages) even though crowd depends on it; editing `SessionService.swift` in the hot-reload loop silently tests a stale binary.

### Green (consider — non-blocking)

- `Artifacts.swift`/`StaticAssets.swift`: add `X-Content-Type-Options: nosniff`; pin artifact `img.src` to `/artifacts/` client-side (`app.js:1141`).
- `WebSocketOriginGuard.swift:66-81`: IPv6 private origins (`fd00::/8`, link-local) are rejected on a wildcard bind — over-restrictive, fails safe, but breaks IPv6 LAN/tailnet browsers.
- Global login rate limiter lets one client lock out legitimate logins for up to a minute (acknowledged trade-off).
- `app.js:590-613`: failed bulk deletions leave stale ids in `selectedSessionIDs`, pre-checked next time selection mode opens; `app.js:211`: the fixed 2.5s sound-arm timer can chime for pre-existing PR states on slow first loads; `app.js:1317,1418,1489`: raw `e.message` written to xterm permits ANSI/OSC injection — strip ESC bytes.
- `settings.js:153`: Escape closes the whole Settings modal through a stacked job/workspace sub-form, discarding its edits.
- `terminal.html:80-90`: `swallowMouseMode` drops a combined DECSET (e.g. `CSI ?25;1000h`) wholesale, losing the non-mouse params.
- `TmuxBackend.swift:642`: the orphan-reaper's bare-shell check runs before the agent-name grace, contradicting the field doc's "never reaped mid-`new-terminal`" claim; `TmuxController.swift:127-129`: tab-containing window names shift the `listWindows` parse.
- `RPCHandlers.swift:319,323`: `set-status` uses two `Date()` calls for memory vs disk `updatedAt`; `:683` comment describes app-forwarding that no longer happens.
- Tauri: `"csp": null` in `tauri.conf.json`; `build.rs` only watches `.git/HEAD` so the About SHA goes stale on same-branch commits; the baked debug `crowd` path (`lib.rs:29`) must be resolved before any bundle ships (acknowledged TODO).

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 15 Yellow, 17 Green] findings.

---

dgershman:
Round 2 — fixed the concrete Yellows in `4c1e59c`; deferred the architectural ones with reasoning.

## Fixed in `4c1e59c`
- **#8/#9 rpc resilience** — `rpcConnect` rejects on a pre-open close (no more hang when the daemon is down) and rejects in-flight rpcs on close; the reconnect only fires when the closing socket is still active, so a call during the window can't leak a duplicate `/rpc` socket double-firing `changed`.
- **#4 work-on-issue url** — validated as a plain `http(s)` URL with no control chars in both the daemon and engine handlers, before it can reach the Manager as keystrokes.
- **#5 terminalWebFontURL** — now requires `https:` (this PR's web-font feature; an injected config can't load an attacker-controlled stylesheet).
- **#1 loopback bypass** — crowd logs a loud startup warning when a web password is set.
- **#2 prune** — `SessionStore.prune()` wired to a 5-min loop.
- **#11 Tauri port** — honors `CROW_HTTP_PORT` and passes `CROW_SOCKET` through, so it reuses a custom-port crowd instead of spawning a contending second daemon.
- **#13 / #15 tooling** — CI builds the `crow-desktop` crate; `crowd-dev.sh` watches CrowEngine/CrowProvider/CrowClaude so hot-reload can't test a stale binary.

Verified: swift build + cargo build clean; CrowDaemon 74 + CrowEngine 245 tests pass; node --check + shellcheck clean.

## Deferred (with reasoning)
- **#3 engine set-config symmetry** — real, but the engine `set-config` is shadowed by the daemon's correct handler (unreachable from the web today). I'll make it symmetric in the same pass that revisits the `mark-in-review`/`create-manager` delegation, so the engine surface is uniformly safe if it ever becomes web-facing.
- **#6 CSP** — worth doing, but a `default-src 'self'` policy now collides with the arbitrary-web-font feature this PR adds (a Google-Fonts `style-src`/`font-src`). It needs a policy that allows the operator's configured `terminalWebFontURL` origin; done carelessly it silently breaks the font setting. Tracking as its own change.
- **#7 Tauri IPC scoping** — the handshake-before-`navigate` is doable, but `withGlobalTauri` can't just be dropped: the notification shim uses `window.__TAURI__`. Wants a scoped capability + a `/version.json` shape check; grouping with the other Tauri hardening.
- **#10 config.json RMW lock** — real last-writer-wins race across `set-config` / `set-web-password` / `onJobRan`. Needs a shared config write-lock (or serializing through one actor) with tests; it touches the persistence core, so a dedicated change.
- **#12 SIGTERM teardown** — needs a `libc`/`nix` dep for `SIGTERM` + reap; folding into the Tauri hardening change.
- **#14 orphaned test coverage** — agreed, oversight not policy. The logic moved to CrowEngine; I'll port `TmuxCrashRecoveryTests`, `CommandLaunchesTokenTests`, `DeferredLaunchTests`, the `ManagerHookConfig` 0600 invariant, and the `AttributionSkill` regression guard into CrowEngine as a focused test-coverage commit.
- Greens noted (nosniff header, IPv6 origin allowance, ANSI-strip on error text, etc.) — will fold into the relevant follow-ups.

Re-requesting review.

dhilgaertner (changes requested):
## Code & Security Review

Round-3 pass: verified the round-2 fix commit (`4c1e59c`) finding-by-finding against the live source, plus a correctness re-check of the changed surfaces.

**Build & tests:** root `swift build` clean (`crow` + `crowd`); CrowDaemon and CrowEngine suites pass; `node --check` clean on `app.js`/`settings.js`.

### Verified fixed (9 of 15 round-2 Yellows)

- **#1 loopback-forwarder bypass warning** — crowd now logs a loud startup warning when `webAuth` is set (`CrowDaemon.swift:58-63`).
- **#2 `SessionStore.prune()`** — wired to a 5-minute loop (`CrowDaemon.swift:239-241`); the store is NSLock-protected, so the background task is safe.
- **#4 `work-on-issue` URL injection** — both the daemon (`RPCHandlers.swift:559-565`) and engine (`EngineRouter.swift:325-331`) handlers now require a plain `^https?://\S+$` URL and reject control characters before the value reaches the Manager as keystrokes.
- **#5 `terminalWebFontURL`** — `ensureWebFontLink` requires `https://` (`app.js:1843-1845`).
- **#8 `rpc()` hang** — `rpcConnect` rejects on a pre-open close and rejects all in-flight waiters on any close (`app.js:42-57`); a Settings save during a daemon restart now fails fast instead of sticking at "Saving…".
- **#9 reconnect race** — the reconnect timer only fires when the closing socket is still the active one and re-checks `rpcState.ready` before reconnecting; no duplicate `/rpc` sockets double-firing `changed`.
- **#11 hardcoded port** — `crow-desktop` honors `CROW_HTTP_PORT` and passes `CROW_SOCKET` through to the sidecar (`lib.rs:20-25,130-134`).
- **#13 desktop CI** — `ci.yml` now builds the Rust crate on macos-15.
- **#15 crowd-dev watch list** — now includes CrowEngine/CrowProvider/CrowClaude.

### Remaining from round 2 (Yellow — should fix)

1. **#3 — engine `set-config` still leaks/wipes secrets** (`EngineRouter.swift:78-92`). The daemon's `applyConfig` closure (`CrowDaemon.swift:203-206`) saves the incoming config verbatim — no `SettingsSecrets.preservingSecrets` — and the handler returns the saved config unstripped. It stays unreachable from the web only because the daemon router shadows the method; the first refactor that reorders that fallback turns an authenticated `set-config` into a Jira-token leak plus a stored-secret wipe (including `webAuth`). Make it symmetric with `RPCHandlers.swift`'s handler.
2. **#6 — still no Content-Security-Policy** on any served page or response header. The whole XSS posture rests on the `textContent` convention; a `default-src 'self'` policy is cheap insurance and also backstops the webfont/link gating added this round.
3. **#7 — the Tauri webview still trusts whatever answers on the port** (`capabilities/default.json`, `withGlobalTauri: true`, `tauri.conf.json` `"csp": null`). No handshake before `navigate()`; on a multi-user machine another local user binding the port first gets Tauri IPC in the Crow-branded window.
4. **#10 — three unsynchronized read-modify-write writers of `config.json`** (`set-config`, `set-web-password`, `JobScheduler.onJobRan`) can still lose a concurrent write (e.g. a dropped `lastRunAt` re-fires a job).
5. **#12 — sidecar teardown is still a bare `child.kill()`** (`lib.rs:184`) — SIGKILL on a stateful daemon mid-write on every app quit. SIGTERM → short wait → kill, then reap.
6. **#14 — the deleted root test suites still have no CrowEngine replacements** (tmux crash recovery, `commandLaunchesToken`, deferred launch, manager-hook 0600 invariant, attribution `${CROW_AGENT_DISPLAY_NAME}` guard).

### New (Green — consider, non-blocking)

- `app.js:55` — the auto-reconnect promise created by the timer has no awaiter until the next `rpc()` call, so each failed retry while the daemon is down surfaces as an unhandled-promise-rejection console warning (~1/s). Attach a no-op `.catch` when assigning from the timer.
- `app.js:1843` — when a previously valid webfont URL is edited to a non-https value, the early return leaves the stale `<link>` applied instead of removing it.
- Round-2 Green items stand as before.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 6 Yellow, 2 Green] findings (all six Yellows carried over unaddressed from round 2).

---

dgershman:
Round 3 — removed the web-font feature and worked through more of the yellows.

## Done
- **Web-font feature removed** (`revert(terminal)`): dropped the configurable terminal font + `@xterm/addon-web-fonts` + the Settings "Terminal" group; the terminal reverts to its built-in Nerd Font stack. Resolves **#5** and clears the way for a strict CSP (**#6**), which the external stylesheet load previously conflicted with.
- **#3 engine set-config symmetry** — preserves stored secrets (`preservingSecrets`) and strips them from the response (`strippedForTransport`), mirroring the daemon handler, so the surface is safe if it ever becomes web-facing.
- **#10 config.json write-lock** — `ConfigStore.withConfigLock` serializes load→mutate→save; `set-config`, `set-web-password`, and `JobScheduler.onJobRan` all go through it, so a job firing during a Settings save can't clobber the edit.
- **#12 SIGTERM teardown** — crow-desktop sends SIGTERM + waits 1.5s (letting crowd flush its store / release tmux + the socket) before SIGKILL, and reaps the child.

(Earlier rounds: #1/#2/#4/#8/#9/#11/#13/#15 in `4c1e59c`; the two Reds + rename/set-locked/link/paste/SVG/crowd-dev in `90f59b2`.)

Verified: swift build + cargo build clean; CrowDaemon 74 + CrowEngine 245 tests pass.

## Remaining (with reasoning)
- **#6 CSP** — now unblocked. I'll add `default-src 'self'` (`script-src 'self'`, `style-src 'unsafe-inline'` for xterm's injected styles, `img-/worker-src blob:`, `connect-src 'self'` for the WS), scoped to `index.html` (login.html has an inline script). Holding it for a commit I can browser-smoke-test: a wrong `connect-src` silently breaks the `/rpc` socket — worse than no CSP — and I can't verify that headlessly.
- **#7 Tauri IPC handshake** — verify the `:8787` listener is actually crowd (a `/version.json` shape check) before `navigate`, and scope the capability. Grouping with a Tauri-hardening pass (keeping `withGlobalTauri`, which the notification shim needs).
- **#14 orphaned test coverage** — porting `TmuxCrashRecoveryTests` / `CommandLaunchesTokenTests` / `DeferredLaunchTests` / the `ManagerHookConfig` 0600 invariant / the `AttributionSkill` guard into CrowEngine as a focused test commit.
- Greens fold into the above.

Re-requesting review.

dhilgaertner (changes requested):
## Code & Security Review

Round-4 pass: verified the five commits since round 3 (`42af53c`…`cb6d6ad`) finding-by-finding against the live source.

**Build & tests:** root `swift build` clean (`crow` + `crowd`); CrowDaemon 74/74 and CrowEngine 245/245 pass; `node --check` clean on `app.js`/`settings.js`. No Rust toolchain on the review machine and no CI checks reported on this branch, so the `lib.rs` changes are code-reviewed but not compile-verified this round (imports/deps check out; the commit states `cargo build` clean).

### Verified fixed (5 of 6 round-3 Yellows)

- **#3 engine `set-config` secrets** — now mirrors the daemon handler: `preservingSecrets` before apply, `strippedForTransport` on the response (`EngineRouter.swift:84-97`). No more token leak/wipe if the fallback ever becomes reachable.
- **#6 CSP** — `index.html` now ships a tight policy (`default-src 'self'`, `script-src 'self' 'wasm-unsafe-eval'`, `object-src 'none'`, `frame-ancestors 'none'`, …) and every asset gets `X-Content-Type-Options: nosniff` (`StaticAssets.swift:83-116`).
- **#7 Tauri webview trust** — `is_crowd()` verifies the `Server: crowd` response header before reusing an existing listener and again before `navigate()` (`lib.rs:49-79`, `lib.rs:205-218`); a foreign process on the port no longer gets the IPC-privileged webview.
- **#10 config write races** — `ConfigStore.withConfigLock` (`ConfigStore.swift:96-103`) now serializes the full load→mutate→save in `set-config` (`RPCHandlers.swift:715`), `set-web-password` (`RPCHandlers.swift:745`), and `JobScheduler.onJobRan` (`CrowDaemon.swift:176`).
- **#12 sidecar teardown** — SIGTERM, up to 1.5 s of `try_wait` polling, SIGKILL fallback, then reap (`lib.rs:239-258`). No more mid-write SIGKILL on every quit.

### Also verified this round

- **`list-terminals` window index restored** (`36d3cf8`, `EngineRouter.swift:783-787`) — fixes the blank web terminal pane regression from the round-1 handler consolidation.
- **Session-expiry detection** (`cb6d6ad`) is sound: `/auth/check` sits behind `WebAuthMiddleware` (not in the exempt list, `WebAuthRoutes.swift:62`), so it 204s only for an authorized caller; the client probe treats **only a definitive 401** as expiry (a thrown fetch = daemon down → keep reconnecting), fails fast on `rpc()` once dead, and stops the reconnect loop (`app.js:685-720`).

### Remaining (Yellow — should fix)

1. **#14 (carried from round 2) — deleted root test suites still have no replacements.** The behaviors survive the refactor without coverage: launch-token handling now lives in `CrowEngine/AgentLaunch.swift` + `EngineRouter.swift`, tmux crash recovery in `CrowTerminal/TmuxBackend.swift`, and the manager-hook 0600 / attribution `${CROW_AGENT_DISPLAY_NAME}` invariants are likewise untested. No test files changed since round 3, and the new security surface from this round (`/auth/check`, CSP headers, `withConfigLock`) also landed without tests — `DaemonSecurityTests` would be the natural home for the first two.

### New (Green — consider, non-blocking)

- `lib.rs:76-78` — `is_crowd()` matches `server: crowd` case-insensitively anywhere in the first ~4 KB of the response, **body included**, and there's a TOCTOU window between the check and `navigate()`. Fine against accidental port collisions (the stated threat), but a deliberate local squatter can trivially spoof it — worth a one-line comment so nobody mistakes it for authentication. Matching only within the header block would also cheaply harden it.
- `EngineRouter.swift:84-92` — the engine `set-config` load→merge→save is a fourth `config.json` writer that does *not* take `withConfigLock` (currently unreachable: the daemon router shadows the method — same caveat as the old #3).
- `app.js:62` — the reconnect-timer promise still has no `.catch` (round-3 Green): while crowd is down, each failed retry logs an unhandled-rejection warning ~1/s.
- `login.html` (inline script) and `terminal.html` are served without any CSP; a minimal policy with `script-src 'unsafe-inline'` would still pin `connect-src`/`object-src`.
- Round-3 webfont Green is moot — the feature was reverted in `601933f`.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 4 Green] findings. The single Yellow (#14, missing test replacements) has now been carried for three rounds; landing even the security-surface subset (`/auth/check` gating, CSP presence, config-lock concurrency) would clear the last blocker.

---

dgershman:
Round 5 — Yellow **#14** (orphaned test coverage) is cleared via #608, now on this branch as `dc07a39`.

## Done
- **DaemonSecurityTests** — `/auth/check` gating (not auth-exempt; allowlist exact; login-page vs 401 selector), CSP on `index.html` only with hardening directives, `ConfigStore.withConfigLock` concurrency.
- **AgentLaunchTests** — re-homes launch-token anchoring + deferred-launch hook/OTEL prep.
- **ClaudeHookConfigWriterTests** — manager-hook `0o600` + preserves unrelated settings.
- tmux crash recovery + attribution `${CROW_AGENT_DISPLAY_NAME}` already covered elsewhere (called out in #608).

Greens from round 4 left as-is (non-blocking).

Re-requesting review.

dhilgaertner (changes requested):
## Code & Security Review

Round-5 pass: verified the six commits since round 4 (`faf3893`…`cca4366`): Makefile/docs simplification (`998c313`), license docs (`a78a15b`), local-only web password + AI gateway management (`b545271`), the test re-home (`dc07a39`), tmux scrollback replay (`90f075b`), and the first-run setup wizard (`cca4366`).

**Build & tests:** CrowDaemon **92/92** (28 suites, up from 74 in round 4 — includes the new `DaemonSecurityTests`, `TerminalReplayTests`, and the `run-setup` forwarder tests), CrowEngine `AgentLaunch` 12/12, CrowClaude `ClaudeHookConfigWriter` 2/2 all pass locally; `node --check` clean on `app.js`/`settings.js`. CI (`.github/workflows/ci.yml`) now loops **every** package's test suite plus the Rust desktop build.

### Verified fixed

- **#14 (Yellow, carried since round 2) — missing test replacements: CLEARED.** `dc07a39` re-homes the dropped coverage exactly where round 4 asked: `WebAuthMiddlewareGatingTests` (auth-exempt allowlist is exactly `/login /logout /health /brand.svg`; `/auth/check` is gated; login-page-vs-401 selector), `ContentSecurityPolicyTests` (CSP scoped to `index.html`, hardening directives asserted, no `unsafe-inline`/wildcard), `ConfigLockConcurrencyTests` (`withConfigLock` loses no updates under concurrent RMW), plus `AgentLaunchTests` (launch-token anchoring, deferred hook-config write, OTEL env) and `ClaudeHookConfigWriterTests` (`settings.local.json` written `0o600`, only `ANTHROPIC_*` keys cleared). tmux crash recovery and attribution were confirmed already covered in `CrowTerminal`/`CrowCore`. The source changes are testability seams only — verified no behavior change.
- **`SecretRoutes` / `isLocalDirect` (`b545271`)** — the design is sound: secret writes are HTTP POSTs (so the peer address + `X-Forwarded-For` are in hand), gated to loopback-peer-with-no-XFF **and** Origin-checked via the same `WebSocketOriginGuard` trust rules; the `set-web-password` JSON-RPC method (which couldn't distinguish local from logged-in-remote) is removed, with a comment in `RPCHandlers.swift` warding off reintroduction. `buildGateway`'s both-or-neither invariant is enforced server-side and tested.
- **Round-4 Green (login/terminal CSP)** — resolved by documented decision: `StaticAssets.appliesCSP(to:)` now states why `login.html` (inline script) and `terminal.html` (debug-only) are excluded, and the scoping is pinned by a test.

### Security Review

**Strengths:**
- Secret management is now structurally local-only; a remote session can no longer rotate the password gating its own access, and gateway auth headers are stripped in transport both ways.
- The wizard's `run-setup` is one-shot (rejected once the devroot pointer exists), takes the config write lock, and re-exec is delayed so the RPC reply flushes first.
- Replay reuses the PTY's `AsyncStream` so replayed history serializes with live output — no concurrent writer on the WS.

**Concerns:** the two Yellows below.

### Findings

**Yellow 1 — updating an AI gateway silently wipes its stored auth-header secrets.** `get-config` ships gateways with header *values* blanked but keys kept (`SettingsSecrets.strippedForTransport`, `SettingsSecrets.swift:88-92`). The local editor prefills those stripped `Name: ` lines (`settings.js` `gatewayEditor`), so clicking **Update gateway** — e.g. to change only the base URL — POSTs `headers: {"X-Api-Key": ""}`; `SecretRoutes.buildGateway` (`SecretRoutes.swift:136-141`) filters only empty *keys* and persists the empty values verbatim, with no merge against the stored gateway. The stored secret is destroyed and the UI reports "saved". The help text ("re-enter values to change them") promises the opposite — blank should mean *keep*. Fix server-side: inside the `withConfigLock` block, merge empty-valued headers from the currently-stored gateway (blank value → keep stored value); that also matches the transport-stripping contract.

**Yellow 2 — `run-setup`'s re-exec dies when `crowd` was launched via `PATH`.** `CrowDaemon.reexec` (`CrowDaemon.swift:309-323`) calls `execv(argv[0])`, which does not search `PATH`. The README's install flow (`make install` symlinks `crowd` into `~/.local/bin`; user runs `crowd`) yields `argv[0] == "crowd"`, so after the first-run wizard the `execv` fails with ENOENT and the daemon **exits** (`exit(1)`) instead of restarting — the wizard polls for 30 s and lands on "did not come back online. Reload the page", but the page is dead until a manual relaunch. Setup state *is* persisted first, so recovery is manual-restart, not data loss — but this breaks the primary first-run flow the wizard exists for. Fix: resolve the real executable path (`_NSGetExecutablePath` on macOS / `/proc/self/exe` on Linux) or fall back to `execvp`.

### Code Quality (Green — consider, non-blocking)

- `run-setup` (`RPCHandlers.swift:748+`) decodes and saves a **raw** `AppConfig` with no `SettingsSecrets` treatment, and rides the shared `/rpc` router. During the unconfigured window auth is inert, so any client passing the Origin guard could seed `webAuth`/gateway secrets — bypassing the local-only gate `SecretRoutes` just built. One-shot + loopback-default bounds it; consider gating `run-setup` to local-direct like the secret POSTs (it's a first-run action on the machine anyway).
- Every tab click / reconnect re-captures and ships up to 50,000 lines (`TerminalCockpit.replayLines`) as one WS frame; with a full history that's a multi-MB burst per tab switch. Consider a smaller replay window for window-*switch* (vs. reconnect), or capping the frame.
- Carried round-4 Greens, still open: `app.js:62` reconnect-timer promise has no `.catch` (one unhandled-rejection warning ~1/s while crowd is down); `EngineRouter.swift` `set-config` remains a config writer that skips `withConfigLock` (still unreachable — daemon router shadows it); `crow-desktop/src-tauri/src/lib.rs` `is_crowd()` still matches `server: crowd` anywhere in the first 4 KB (body included) with no comment marking it as spoofable-by-design.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 5 Green] findings. The long-carried test-coverage Yellow is cleared and the security architecture of the secret routes is solid; the two new Yellows are small, well-localized fixes (a header-merge in `SecretRoutes`, an executable-path resolve in `reexec`) that should land in one round trip.

---

dgershman:
Round 5 Yellows addressed in https://github.com/radiusmethod/crow/pull/623 (into this branch):

1. **Gateway header merge** — blank values from the stripped local editor keep stored secrets; non-blank overrides; `clear` still clears.
2. **PATH-safe re-exec** — resolve real executable / `execvp` so first-run wizard restart works when `crowd` was launched from PATH.

Greens left as-is. Re-requesting review once #623 merges (or review that PR directly).

dhilgaertner (changes requested):
## Code & Security Review

Round-6 pass. **No new commits have been pushed since the round-5 review** — HEAD is still `cca4366`, the exact commit round 5 requested changes on. This round re-verified the outstanding findings against the current tree rather than re-litigating the full diff.

**Build & tests:** CrowDaemon **92/92** (28 suites) pass locally at HEAD, matching round 5.

### Critical Issues
None (no Red findings).

### Security Review
**Strengths:**
- The round-5 assessment stands: secret writes are structurally local-only (loopback peer + no `X-Forwarded-For` + Origin-checked), `get-config` strips gateway auth-header values in transport, and the first-run `run-setup` is one-shot under the config write lock.

**Concerns:**
- Both round-5 Yellows are unaddressed (below). Yellow 1 is secret-handling adjacent: the update path silently destroys stored gateway auth headers.

### Findings — carried from round 5, re-verified at HEAD

**Yellow 1 — updating an AI gateway silently wipes its stored auth-header secrets (unchanged).** `SecretRoutes.buildGateway` (`Packages/CrowDaemon/Sources/CrowDaemon/SecretRoutes.swift:136`) still filters only empty header *keys* and persists blanked *values* verbatim, with no merge against the stored gateway. Since `get-config` ships gateways with values stripped (`SettingsSecrets.swift`), clicking **Update gateway** to change only the base URL re-POSTs `"X-Api-Key": ""` and destroys the stored secret while the UI reports "saved" — the opposite of the "re-enter values to change them" help text. Fix inside the `withConfigLock` block: blank value → keep the stored value.

**Yellow 2 — `run-setup`'s re-exec dies when `crowd` was launched via `PATH` (unchanged).** `CrowDaemon.reexec` (`Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift:320`) still calls `execv(argv[0])`, which does not search `PATH`. With the README install flow (`crowd` on `PATH`), `argv[0] == "crowd"`, `execv` fails with ENOENT, and the daemon exits instead of restarting — the wizard's primary first-run flow ends at a dead page requiring a manual relaunch. Fix: resolve the real executable path (`_NSGetExecutablePath` on macOS / `/proc/self/exe` on Linux) or fall back to `execvp`.

### Code Quality (Green — consider, non-blocking; carried from round 5)
- Gate `run-setup` to local-direct like the secret POSTs — during the unconfigured window it can seed `webAuth`/gateway secrets over the shared `/rpc` router.
- `TerminalCockpit.replayLines` ships up to 50,000 lines as one WS frame on every tab switch; consider a smaller window for window-switch vs. reconnect.
- `app.js:62` reconnect-timer promise lacks a `.catch`; `EngineRouter.swift` `set-config` skips `withConfigLock` (currently unreachable); `crow-desktop` `is_crowd()` matches `server: crowd` anywhere in the first 4 KB with no spoofable-by-design comment.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 5 Green] findings, all carried from round 5 with no intervening commits. Both Yellows are small, well-localized fixes (a header-merge in `SecretRoutes.buildGateway`, an executable-path resolve in `reexec`) that should clear the PR for approval in one round trip.

---

dgershman:
Round 6: the round-5 Yellows are in https://github.com/radiusmethod/crow/pull/623 (into this branch). That PR also picked up a Red on the merge helper (URL + empty headers after blank-drop could make `config.json` undecodable and wipe config on next write) — fixed in the latest push on #623.

Once #623 merges, both Yellows land on this branch. Re-requesting review.

dhilgaertner (changes requested):
## Code & Security Review

Large PR (207 files, ~19k additions), but the net new surface is a Swift daemon (`crowd`) exposing an HTTP/WebSocket web UI plus web-access auth; most other files are package renames with trivial edits. I focused on the security-critical daemon/engine code and the web assets, and verified the build + test claims.

**Overall this is strong, security-conscious work.** No critical/blocking issues. The findings below are all "should fix" (Yellow), so per Crow's verdict rule this lands as Request Changes — the implementing agent can clear them in the same round trip.

### Critical Issues
None.

### Security Review

**Strengths (verified):**
- **No command injection.** Every shell-out goes through an argv array (`/usr/bin/env` + `GitManager`/`CrowGit`), never `/bin/sh -c` with interpolation. `add-worktree` additionally blocks leading-dash branch option-injection (`RPCHandlers.swift:262`) and `work-on-issue` validates the URL as plain `http(s)` before it reaches the Manager as keystrokes.
- **Auth/origin gating is consistent.** `/rpc` and `/terminal` both gate the WS upgrade on `originOK && auth.isAuthorized` (`RPCWebSocketHandler.swift:32-44`). The new `list-agents` / `remove-link` / `set-locked` methods and the whole `CommandRouter` fallback surface inherit that gate — no method escapes it.
- **Secret handling is careful.** `SettingsSecrets.strippedForTransport` blanks the Jira token, gateway header values, and web-password hash+salt on the way out; `preservingSecrets` restores stored secrets on the way in, so an authenticated remote `/rpc` client can neither read nor clear the password that gates remote access. Actual secret writes are HTTP POSTs gated by origin-check **and** `isLocalDirect` (loopback peer, no `X-Forwarded-For`) — CSRF-safe (`SecretRoutes.swift:154-161`).
- **PBKDF2-HMAC-SHA256** (210k iterations, per-salt), constant-time verify, HttpOnly+Secure+SameSite cookies, login rate-limiting, CSP on `index.html`, `nosniff`, SVG `Content-Disposition: attachment`, DNS-rebinding-resistant origin guard. All well done.

**Concerns:**

- 🟡 **Artifacts route follows symlinks → local file disclosure** — `Artifacts.swift:50` (`Data(contentsOf:)`) + `isSafeImageName` (`:69`). The name filter rejects `/`, `..`, and non-image extensions but never resolves the target, and `Data(contentsOf:)` follows symlinks. A `shot.png` symlink planted in `$TMPDIR/crow/artifacts/<uuid>/` pointing at `~/.claude/config.json` (or `~/.ssh/id_rsa`) is served verbatim to any authorized `GET /artifacts/<uuid>/shot.png` — leaking exactly the password hash/gateway headers that `strippedForTransport` keeps off every other transport, and contradicting the file's own "nothing outside this scratch tree is reachable" doc. Exploitation needs a local process (or a prompt-injected agent) to plant the symlink, so not critical, but the invariant is stated as a hard guarantee. Fix: resolve the final path and require it to remain inside `dir` (or open with `O_NOFOLLOW` / reject symlinks).

- 🟡 **`run-setup` is an unauthenticated write+re-exec primitive on a non-loopback bind** — `RPCHandlers.swift:748-779`. Pre-configuration there is no web password, so `authorize` returns "auth disabled" and only the Origin guard applies. `expandSetupDevRoot` accepts absolute paths, so on a `--host 0.0.0.0` bind any LAN/native client can make `Scaffolder` create directories at an attacker-chosen path, author `config.json`, and force a daemon restart. This lives inside the already-documented "non-loopback is unauthenticated" envelope, but it's a far more damaging member of that surface than the read methods. Consider gating it to a local-direct peer like the secret routes.

- 🟢 **`isPathWithinRoot` doesn't resolve symlinks** — `Validation.swift:9-13` uses `standardizedFileURL` (resolves `..` but not symlinks); callers are `add-worktree` / `new-terminal`. Requires a pre-existing symlink under devRoot to escape, so low. Consider comparing `resolvingSymlinksInPath()` of both sides.

### Code Quality

- 🟡 **`/terminal` WebSocket reconnects forever after the session dies** — `app.js:2027`. `termWs.onclose` unconditionally `setTimeout(connectTerminalWs, 1000)` with no `sessionDead` guard, unlike the `/rpc` loop (`app.js:62`). After a `crowd` restart invalidates a remote cookie, the `/rpc` bar flips to "Log in" but the terminal socket keeps handshaking (401 → close → retry) every second indefinitely. Add the same `sessionDead` guard.

- 🟡 **Allowlist "Clear" button gets stuck disabled** — `app.js:1841-1848` vs render at `:1816-1819`. The per-row checkbox `onchange` only patches the Promote button; the Clear button's `disabled` was fixed at render time from `allowlistSelection.size === 0`. Ticking a row from an empty selection enables Promote but leaves Clear disabled until an unrelated `renderBoard()` fires. Refresh the Clear button on toggle (or just `renderBoard()`, as Select-All/Clear do).

- 🟡 **Escape in a stacked settings sub-form tears down the whole modal** — `settings.js:164` + `:935-964`. The document `escHandler` always calls `closeSettings()`; the stacked job/workspace overlay only wires click-outside dismissal. Editing a sub-form marks the outer form dirty, so Escape triggers the outer "Discard unsaved changes?" and closes all of Settings instead of backing out of the sub-form — inconsistent with the backdrop-click behavior. Scope Escape to the topmost overlay.

- 🟢 **`get-config` comment/behavior divergence** — `RPCHandlers.swift:690-714`. The comment describes "forward to the app when reachable, else read disk," but it unconditionally reads disk and returns `app_running:false`. Behavior is intentional per ADR 0007 (daemon = authority); only the comment is stale. Several sibling handlers carry the same stale "forwarded to the app" comments — worth a sweep.

**No XSS or secret leakage in the web UI:** all provider-sourced strings render via a `textContent`-based `el()` helper; every `innerHTML` write is either clearing or static markup; untrusted URLs are gated by `/^https?:\/\//i` before becoming an `href`; gateway header values render only when `isLocal`. Confirmed clean.

### Testing (verified locally)
- `swift test` CrowDaemon — **92/92 pass** (incl. 50 `DaemonSecurityTests`: PBKDF2, origin guard, auth gating, secret stripping, branch option-injection).
- `swift test` CrowCore — **345/345 pass**.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 5 Yellow, 2 Green] findings. None are blockers; the security foundation is excellent. The Yellow items (artifacts symlink hardening, `run-setup` gating, terminal reconnect guard, and the two settings/allowlist UI bugs) are cheap fixes best landed in this same round trip.

---

dgershman:
Round 7 Yellows addressed in https://github.com/radiusmethod/crow/pull/623 (into this branch):

- Artifacts: resolve symlinks and keep the final path inside the scratch dir
- `run-setup`: local-direct only on `/rpc`
- `/terminal`: no reconnect loop after `sessionDead`
- Allowlist Clear: enables/disables with selection
- Settings Escape: backs out of stacked sub-form first

Greens left as-is. Re-requesting review once this lands on the parity tip (or as soon as #623 is reviewable on its own).

dhilgaertner (changes requested):
## Code & Security Review

Reviewed the security-critical `crowd` surface introduced/expanded here (web auth, WS origin guard, secret routes, static assets, RPC handlers) plus the new web assets. The core model holds up well: **`get-state`/`get-config`/`set-config` all strip credentials before they reach a client, the `/rpc` and `/terminal` WS upgrades are both gated on `originOK && WebAuthGuard.authorize`, secret writes are correctly confined to local-direct + Origin-checked POSTs, git shells out via argv arrays (no shell string), and the web UI renders all agent/user-controlled data through `textContent` with `^https?://`-anchored href guards — no DOM XSS found.** Nice work on the redaction and the constant-time PBKDF2 verify.

Two hardening items keep this from a clean approve.

### Security Concerns

**🟡 `run-setup` accepts an arbitrary absolute `dev_root` and re-execs — reachable pre-auth on an unconfigured non-loopback daemon.**
`RPCHandlers.swift:748-779` + `expandSetupDevRoot` (`:889-897`). On a not-yet-configured daemon, `WebAuthGuard.authorize` returns authorized for any peer (the password gate is opt-in/inert until a password exists, `WebAuth.swift:180-182`), and `run-setup` is only rejected *after* config exists. `expandSetupDevRoot` passes absolute paths through unchanged — no confinement. If the daemon is bound non-loopback (the loudly-warned unauthenticated mode) and unconfigured, a native client with no `Origin` header (bypasses the origin guard) can send `run-setup` with `dev_root:"/some/writable/abs/path"`, which scaffolds a tree there, writes `config.json` + the devRoot pointer, and `requestReexec()`s the daemon onto the attacker's devRoot — arbitrary-location write + persistent devRoot hijack + restart, in the pre-config window. Suggest gating `run-setup` to `isLocalDirect` (as `SecretRoutes` does) and/or confining `dev_root` under `$HOME`.

**🟡 `set-config` over `/rpc` persists `defaults.binaries` (agent executable-path overrides) and `jobs`, which execute at the next launch.**
`RPCHandlers.swift:715-742`. The handler doc at `:781-787` correctly argues secret *writes* must not be `/rpc` methods because the shared router "can't tell a local caller from a logged-in remote one" — but that same reasoning applies to `defaults.binaries` (absolute exec paths later run by agent launch) and `jobs` prompts, which `set-config` writes without the local-direct gate. For a logged-in remote operator this is not a new capability (they already have `new-terminal`/`send`), but in the unauthenticated non-loopback mode it's persistent RCE with no tmux required. Consider holding binary/jobs writes to the same local-direct bar as secret writes, consistent with the principle the code already states.

### Code Quality (non-blocking)

- 🟢 `get-config` comment (`RPCHandlers.swift:690-714`) describes forwarding to the app + a `-32601` fallback, but the handler always reads local disk and hard-codes `"app_running": false`. Stale comment / a client can't detect the app being up.
- 🟢 `isPathWithinRoot` (`Validation.swift:9-13`) collapses `..` lexically (`standardizedFileURL`) but doesn't `resolvingSymlinksInPath`, so a symlink under `devRoot` escapes the root. Pre-existing, needs a pre-planted link.
- 🟢 Login rate limiter is global (`WebAuth.swift:131-148`) — a lockout DoS (acknowledged in the comment).
- 🟢 `bulkDeleteSelected` (`app.js:625-640`) only removes ids from `selectedSessionIDs` on success and never `.clear()`s on exit — a failed delete leaves a stale pre-checked row for the next selection.
- 🟢 Stacked sub-form Escape (`settings.js`) closes the whole Settings modal rather than just the child overlay (backdrop-click is correctly per-layer; the keyboard path isn't).
- 🟢 External link chips use `rel="noopener"` but not `noreferrer` (`app.js:1307,1637`) — the full crowd path leaks as `Referer` to attacker-influenceable ticket/PR URLs.
- 🟢 Server error strings are `term.write`n without stripping ANSI/OSC (`app.js` ~1386) — marginal given the terminal renders untrusted PTY output anyway.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 7 Green] findings. Both Yellows are the same shape: powerful setup/config writes (`run-setup`, `set-config` binaries/jobs) not held to the local-direct bar the code sets for secret writes. Everything else is minor. Since this is a draft not marked for merge, these are cheap to fold into the next round.

---

dgershman:
Round 8 on the latest review:

1. **`run-setup` local-direct** — already landed via https://github.com/radiusmethod/crow/pull/623 (now merged into this branch). The review commit was on the pre-merge tip.
2. **`set-config` binaries/jobs** — gated to local-direct in https://github.com/radiusmethod/crow/pull/624 (into this branch). Unchanged binaries/jobs round-trips and other settings fields still work remotely.

Greens left as-is. Re-requesting review.

dgershman:
Correction: the set-config binaries/jobs gate is in https://github.com/radiusmethod/crow/pull/625 (not #624 — that number was a typo in the previous comment).

dhilgaertner (approved):
## Code & Security Review

Reviewed the security-critical web-daemon surface introduced/carried by this PR (auth, secret management, WebSocket gating, static-asset serving, config persistence) plus the current-round fixes (gateway-secret preservation, PATH-safe `crowd` re-exec). This is a mature, heavily-iterated PR (round 8+); the security design is exemplary and every prior review finding I traced has a documented fix in place.

### Critical Issues
None.

### Security Review

**Strengths:**
- **Password hashing** (`WebAuth.swift`): PBKDF2-HMAC-SHA256 at 210k iterations (OWASP 2023), per-hash random salt from a CSPRNG, and a constant-time compare on verify — no early-exit leak.
- **Session cookies**: `HttpOnly; Secure; SameSite=Lax`, 256-bit random tokens, in-memory TTL store; `prune()` is wired to a 300s timer (`CrowDaemon.swift:245`), so expired tokens don't accrue. `/login` is globally rate-limited (`LoginRateLimiter`).
- **Cross-site WS-hijacking defense** (`WebSocketOriginGuard`): rejects `/rpc` + `/terminal` upgrades whose `Origin` isn't loopback/allowed; `X-Forwarded-Host` is honored **only** from a loopback peer, and the opaque `"null"` origin is rejected. `/rpc` also enforces `WebAuthGuard.authorize` since it shells out to git.
- **CSRF on secret writes** (`SecretRoutes`): the web-password and AI-gateway writes are dedicated HTTP POSTs gated by both an Origin check *and* a local-direct (loopback, no `X-Forwarded-For`) check — a logged-in remote client and a malicious local page are both refused. Correctly kept off the transport-agnostic `/rpc` router (documented at `RPCHandlers.swift:785`).
- **Secret non-exfiltration** (`SettingsSecrets`): credential values are stripped on the way out to the browser and restored from disk on the way back — no web path can read, replace, or blank a stored token/header. The blank-value-means-keep merge (`mergingPreservedHeaders`) correctly rejects a URL-with-no-surviving-headers shape that would otherwise wipe the config on the next load (the #623 Red).
- **Path traversal**: `/xterm/*` is basename-guarded (`isSafeAssetName`) after the router's percent-decode.
- **CSP + `nosniff`** on the app page; `object-src 'none'`, `base-uri 'none'`, `frame-ancestors 'none'`, `connect-src 'self'`.
- **PATH-safe re-exec** (`CrowDaemon.reexec`): resolves the real executable via `_NSGetExecutablePath`/`realpath` (or `/proc/self/exe`) before `execv`, falling back to `execvp` for a bare `crowd` on PATH; single-instance lock fd is closed first.
- **Config concurrency**: read-modify-write is serialized through `ConfigStore.withConfigLock`, covered by `ConfigLockConcurrencyTests`. Config/pointer files written 0o600 / dirs 0o700.

**Concerns:** None blocking. The `X-Forwarded-*` trust model (a raw TCP forwarder that omits `X-Forwarded-For` would look local) is a known, documented deployment constraint, not a code defect.

### Code Quality
- Web `innerHTML` uses are all either `= ''` clears or static SVG literals — no provider/user-data interpolation — and CSP is a backstop.
- **[Green] Stale doc comment**: `SettingsSecrets.swift:60` and `AppConfig.swift:85` reference a "`set-web-password` RPC". No such RPC exists — the password is managed via the `/config/web-password` HTTP POST in `SecretRoutes` (as `RPCHandlers.swift:785` itself notes). The described invariant is correct; only the mechanism name is stale. Worth a one-line comment fix on a future pass.

### Verification
- `swift build --package-path Packages/CrowDaemon` — clean (deprecation warnings only).
- `swift test --package-path Packages/CrowDaemon` — **100 tests / 29 suites pass**, including `DaemonSecurityTests`, `PasswordHashTests`, `ConfigLockConcurrencyTests`, `ConfigForwarderTests`, `BoardForwarderTests`.

### Summary Table

---

dgershman:
Rebase onto \`main\` is done locally, but **force-push to \`feature/crow-593-web-ui-parity\` is blocked** by branch rules (no force-push; PR-only).

Conflict-resolved tip (including CROW-604 \`crow job\` RPC port into crowd after AppDelegate removal): https://github.com/radiusmethod/crow/pull/NEW

Please either:
1. Allow a one-time force-push to update this PR head, or
2. Close this PR in favor of the rebased PR above.

dgershman:
Correction: the conflict-resolved rebased tip is https://github.com/radiusmethod/crow/pull/634 (into `main`). Force-push to this PR's head branch remains blocked by branch rules.

dgershman:
Opened a sync PR **into this branch** (no force-push): will paste URL next.

That merges \`main\` + ports \`crow job\` RPCs into crowd so this PR can clear DIRTY without rewriting history.

dgershman:
Sync PR into this branch: https://github.com/radiusmethod/crow/pull/636

Merges \`main\` + ports \`crow job\` RPCs into crowd (no force-push). Once #636 merges, this PR should clear DIRTY.

dgershman:
Force-push to this head is still blocked by branch rules. Opened a **real merge** (not squash) of \`main\` into this branch: https://github.com/radiusmethod/crow/pull/NEW

Please merge that PR with a **merge commit** — squash would leave #594 DIRTY again.

dgershman:
Correction: the real-merge PR is https://github.com/radiusmethod/crow/pull/643

**Merge #643 with a merge commit (not squash).** After that, this PR should stop being DIRTY — local merge-tree against \`main\` reports no conflicts.

dgershman:
Sync PR into this branch (merge, not rebase): https://github.com/radiusmethod/crow/pull/see-next

Please merge it with a **merge commit** (not squash) so DIRTY clears.

dgershman:
Correction: sync PR is https://github.com/corveil/crow/pull/664

Please merge with a **merge commit** (not squash) so this PR clears DIRTY.

PR #596, opened 2026-07-06
Stop wiping user settings.json on every launch

### LINKED ISSUE #595: Scaffolder overwrites settings.json on every launch, silently discarding user customizations
## Environment

- **macOS version:** 26.5.1
- **Hardware:** Apple Silicon
- **Crow version/commit:** 9ce0d68 (main)

## Description

`Scaffolder.scaffold(...)` runs on every app launch (`AppDelegate.launchMainApp`, comment: "Update skills and CLAUDE.md on every launch") and unconditionally overwrites `{devRoot}/.claude/settings.json` with the bundled template:

```swift
// Always overwrite settings.json (permissions for crow, gh, git commands)
let settingsPath = (claudeDir as NSString).appendingPathComponent("settings.json")
let settingsTemplate = Self.bundledSettings()
try settingsTemplate.write(toFile: settingsPath, atomically: true, encoding: .utf8)
```

Any customization a user makes to that file — extra `permissions.allow` entries, `outputStyle`, `statusLine`, `hooks`, `env`, a tweaked `sandbox` block — is silently discarded the next time Crow starts, since the write never reads the existing file first.

## Steps to Reproduce

1. Launch Crow so it scaffolds `{devRoot}/.claude/settings.json`.
2. Hand-edit that file: add a custom `permissions.allow` entry (e.g. `Bash(npm test:*)`) and a top-level key Crow doesn't manage (e.g. `"outputStyle": "concise"`).
3. Quit and relaunch Crow.

## Expected Behavior

The user's custom permission entry and `outputStyle` key survive the relaunch.

## Actual Behavior

Both are gone — `settings.json` is back to exactly the bundled template, with no trace of the customization.

## Logs

Confirmed by reverting to this commit, rebuilding, and running the binary against a devRoot pre-seeded with a custom `settings.json` containing a marker key — the marker, the custom permission, and `outputStyle` were all wiped after a single launch, replaced entirely by the bundled template.

GalaxysHub:
## Summary

- `Scaffolder.scaffold(...)` runs on every app launch and used to unconditionally overwrite `{devRoot}/.claude/settings.json` with the bundled permissions template, silently discarding anything a user added by hand (extra `permissions.allow` entries, `outputStyle`, `statusLine`, `hooks`, `env`, a tweaked `sandbox` block).
- Crow now writes its required permissions to `{devRoot}/.claude/settings.local.json` instead, and never reads or writes `settings.json` at all — that file is now entirely the user's own.
- Claude Code merges `permissions.allow` across `settings.json` and `settings.local.json` (confirmed against Claude Code's docs), so `crow`/`gh`/`git` commands still run without a permission prompt.
- `settings.local.json` is itself guarded by a new `Scaffolder.mergeSettings(...)`, so hand-edits to *that* file survive too: it only fills in permissions/keys that are missing, and leaves the file byte-for-byte untouched when there's nothing to add.

Closes #595.

## Changes

- `Sources/Crow/App/Scaffolder.swift` — merge instead of overwrite; target `settings.local.json` instead of `settings.json`.
- `Tests/CrowTests/ScaffolderSettingsMergeTests.swift` — new suite: template fallback, user-added permissions survive, a missing bundled permission gets re-added, custom top-level keys survive, a user's `sandbox` override wins, a no-op write leaves the file untouched, a full `scaffold()` round-trip across a simulated relaunch, and two tests locking in that `settings.json` is never created or touched (even when one already exists).
- Doc/skill text updated for accuracy: `CONTRIBUTING.md`, `docs/architecture.md`, `docs/configuration.md`, `docs/getting-started.md`, `skills/crow-batch-workspace/SKILL.md`.

## Test plan

- [x] `swift test` — all 285 tests pass across 33 suites, including the 9 new ones.
- [x] Reproduced the original bug against the pre-fix binary: a hand-edited `settings.json` (custom permission + `outputStyle` + a marker key) was completely wiped after one launch, replaced by the bundled template.
- [x] Verified the fix against the built binary: same setup, but after the fix `settings.json` is left alone entirely and the custom permissions/keys land (and survive) in `settings.local.json` — including a deliberately-removed bundled permission getting silently re-added.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -123,6 +123,6 @@ Crow auto-generates certain files into `{devRoot}/.claude/` on launch (see `Scaf
 
 - `CLAUDE.md` — scaffolded from the root `CLAUDE.md`
 - `skills/crow-workspace/SKILL.md` — scaffolded from `Resources/crow-workspace-SKILL.md.template`
-- `settings.json` — always overwritten with pre-approved permissions
+- `settings.local.json` — merged with pre-approved permissions on every launch; `settings.json` is the user's own file and Crow never writes it
 
 **Do not edit the scaffolded copies.** Instead, modify the source files in the repository root or `Resources/` directory.
--- a/Sources/Crow/App/Scaffolder.swift
+++ b/Sources/Crow/App/Scaffolder.swift
@@ -139,10 +139,21 @@ struct Scaffolder {
         try CrowAttribution.expandSkillBody(attributionFooter, agentKind: managerAgentKind)
             .write(toFile: attributionFooterPath, atomically: true, encoding: .utf8)
 
-        // Always overwrite settings.json (permissions for crow, gh, git commands)
-        let settingsPath = (claudeDir as NSString).appendingPathComponent("settings.json")
+        // Write Crow's required permissions to settings.local.json, not
+        // settings.json. settings.json is the user's own file —
+        // Crow never writes it and never has to reconcile with whatever the
+        // user puts there. settings.local.json is Claude Code's local-
+        // override layer: its `permissions.allow` entries are merged with
+        // settings.json's at load time, so crow/gh/git commands still run
+        // without a prompt. `mergeSettings` still guards this file too — a
+        // full overwrite on every launch would just move the "nukes my
+        // customizations" bug from one filename to another. It fills in
+        // only what's missing and leaves the file untouched when there's
+        // nothing to add.
+        let settingsPath = (claudeDir as NSString).appendingPathComponent("settings.local.json")
         let settingsTemplate = Self.bundledSettings()
-        try settingsTemplate.write(toFile: settingsPath, atomically: true, encoding: .utf8)
+        let mergedSettings = Self.mergeSettings(existingPath: settingsPath, template: settingsTemplate)
+        try mergedSettings.write(toFile: settingsPath, atomically: true, encoding: .utf8)
 
         // Create prompts directory for crow-workspace prompt files
         let promptsDir = (claudeDir as NSString).appendingPathComponent("prompts")
@@ -500,7 +511,10 @@ struct Scaffolder {
         return CrowAttribution.sharedFooterInstructions
     }
 
-    /// The settings.json template with pre-approved permissions.
+    /// The pre-approved-permissions template, sourced from the repo's own
+    /// `settings.json` (used as a bundled resource name, not a hint about
+    /// the runtime destination — see `scaffold(...)`, which writes this
+    /// content to `{devRoot}/.claude/settings.local.json`).
     static func bundledSettings() -> String {
         if let content = loadFromRepo("settings.json") {
             return content
@@ -557,6 +571,64 @@ struct Scaffolder {
           }
         }
         """
+    }
+
+    /// Merges `template` into whatever's already on disk at `existingPath`,
+    /// instead of overwriting it. Guarantees:
+    ///
+    /// - `permissions.allow`: the union of the template's and the existing
+    ///   file's entries — any bundled entry the user's file is missing gets
+    ///   appended, but nothing already there is ever removed.
+    /// - Every other top-level key (`outputStyle`, `statusLine`, `hooks`,
+    ///   `env`, a hand-edited `sandbox` block, …): filled in from the
+    ///   template only if the user's file doesn't already have that key.
+    ///   An existing value always wins, even if it differs from the
+    ///   template.
+    ///
+    /// Returns the existing file's contents byte-for-byte, untouched, when
+    /// there's nothing to add — so a user's formatting isn't churned on
+    /// every launch. Falls back to `template` verbatim when there's no
+    /// existing file yet, or it isn't valid JSON.
+    static func mergeSettings(existingPath: String, template: String) -> String {
+        guard let existingData = FileManager.default.contents(atPath: existingPath),
+              let existingString = String(data: existingData, encoding: .utf8),
+              let existingObj = try? JSONSerialization.jsonObject(with: existingData) as? [String: Any],
+              let templateData = template.data(using: .utf8),
+              let templateObj = try? JSONSerialization.jsonObject(with: templateData) as? [String: Any]
+        else {
+            return template
+        }
+
+        var merged = existingObj
+        var changed = false
+
+        for (key, templateValue) in templateObj {
+            if key == "permissions", let templatePerms = templateValue as? [String: Any] {
+                var mergedPerms = merged["permissions"] as? [String: Any] ?? [:]
+                let templateAllow = templatePerms["allow"] as? [String] ?? []
+                let existingAllow = mergedPerms["allow"] as? [String] ?? []
+                let existingSet = Set(existingAllow)
+                let missing = templateAllow.filter { !existingSet.contains($0) }
+                if !missing.isEmpty {
+                    mergedPerms["allow"] = existingAllow + missing
+                    merged["permissions"] = mergedPerms
+                    changed = true
+                }
+            } else if merged[key] == nil {
+                merged[key] = templateValue
+                changed = true
+            }
+        }
+
+        guard changed else { return existingString }
+
+        guard let mergedData = try? JSONSerialization.data(
+            withJSONObject: merged,
+            options: [.prettyPrinted, .sortedKeys]
+        ), let mergedString = String(data: mergedData, encoding: .utf8) else {
+            return existingString
+        }
+        return mergedString
     }
 
     /// Try to load a file from the repo root (for development builds).
--- a/Sources/Crow/App/Scaffolder.swift
+++ b/Sources/Crow/App/Scaffolder.swift
@@ -148,12 +148,23 @@ struct Scaffolder {
         // without a prompt. `mergeSettings` still guards this file too — a
         // full overwrite on every launch would just move the "nukes my
         // customizations" bug from one filename to another. It fills in
-        // only what's missing and leaves the file untouched when there's
-        // nothing to add.
+        // only what's missing and reports `.upToDate` when there's nothing
+        // to add, so a steady-state launch skips the write entirely — the
+        // on-disk file (bytes, inode, mtime, mode) is left completely alone.
         let settingsPath = (claudeDir as NSString).appendingPathComponent("settings.local.json")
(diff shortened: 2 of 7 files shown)


dgershman (changes requested):
## Code & Security Review

Solid fix for #595. The move from "overwrite `settings.json`" to "merge into `settings.local.json`" is the right call, the merge semantics are correct (union `permissions.allow`, existing top-level keys always win, byte-for-byte return when there's nothing to add), and the test suite is genuinely thorough — it covers the fallback, survival, re-add, override-wins, no-op, and full-relaunch cases, plus the two "never touch `settings.json`" invariants. Docs are updated accurately.

There is one interaction issue worth fixing before merge, because it quietly undermines a security control the codebase goes out of its way to enforce elsewhere.

### Security Review

**Strengths:**
- `mergeSettings` only iterates *template* keys, so a user's `hooks`/`env`/`sandbox`/`outputStyle` and any non-`allow` permission sub-keys (`deny`, `ask`) are preserved — it can't clobber the co-writers (`writeHookConfig` / `writeGatewayEnv`) that share this file.
- No injection/eval surface; pure JSON read-modify-write with a safe template fallback on unparseable input.
- Writing Crow's machine-local permissions to `settings.local.json` (the local-override layer) is more correct than the shared `settings.json`, and the union-merge is consistent with Claude Code's documented precedence.

**Concerns:**

- **Yellow — `scaffold`'s write relaxes the `0o600` on the bearer-token file.** This PR makes `scaffold` co-own `{devRoot}/.claude/settings.local.json` — the *same* file whose `env` block can carry a resolved gateway bearer token, which `ClaudeHookConfigWriter.writeGatewayEnv` deliberately restricts to owner-only (`Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift:147-148`, with a dedicated regression test). But `scaffold` writes with `atomically: true` (`Sources/Crow/App/Scaffolder.swift:155-156`), which renames a fresh temp file over the target and resets its mode to the umask default (~`0o644`), dropping the `0o600`.
  - On the normal launch path this is transient — `scaffold` runs at startup (`AppDelegate.swift:469`) and `hydrateState()` (`:576`) later re-applies `0o600` via `writeGatewayEnv`. But the **Settings → "Re-scaffold"** action (`AppDelegate.swift:1197-1207`) runs `scaffold` with *no* following gateway-env write, leaving a token-bearing `settings.local.json` at `0o644` until the next full launch.
  - Related: `scaffold` writes the file on **every** launch even when `mergeSettings` reports no change, so the PR's "byte-for-byte untouched" guarantee holds for the returned *string* but not for the on-disk file (new inode/mtime/mode each launch).
  - Suggested fix: have `mergeSettings` signal "no change" so `scaffold` can skip the write entirely in the steady state, and when it does write, preserve the pre-existing file's mode (or re-apply `0o600` unconditionally, matching the sibling writer). Either removes both the perm-relaxation and the per-launch churn.

Low real-world severity (requires a configured gateway token + a local multi-user machine), but it defeats an explicitly-tested control, so it belongs in this round trip rather than a follow-up.

### Code Quality
- `Sources/Crow/App/SessionService.swift:1894-1898` still copies `bundledSettings()` into a review clone's `settings.json` — correct and intentional (throwaway clone, no user config to protect); no change needed, just confirming it's not an oversight.
- Merge logic, naming, and doc comments are clear and match the surrounding style.

### Build/Test Note
`swift build` fails in the review clone on `cannot find 'BuildInfo' in scope` — this is an **environment artifact**, not a PR defect: `BuildInfo.swift` is generated by `scripts/generate-build-info.sh` into a gitignored `Sources/Crow/Generated/`, and the clone hasn't run that prestep. None of the failing files are in this diff. Relaying the author's reported `swift test` (285 pass) on trust.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 0 Green] findings. Fix the `0o600`-relaxation / every-launch-write in `scaffold` and this is an easy approve.

---

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- The merge is union-only on `permissions.allow` sourced from the bundled template, so this code path can never *remove* a user permission, and it can only *add* entries Crow ships — no injection surface from user-controlled input into the merge.
- Verified empirically that `String.write(toFile:atomically:true)` preserves an existing file's `0600` mode on replace, so the scaffolder's per-launch rewrite of `settings.local.json` does not loosen the owner-only protection that `ClaudeHookConfigWriter.writeGatewayEnv` applies when the `env` block carries a resolved bearer token. Launch ordering also re-applies `0600` after scaffold (`scaffold` at `AppDelegate.swift:469` runs before `hydrateState` at `AppDelegate.swift:576`, whose path calls `writeManagerGatewayEnv()`).
- Confirmed the other writers of the same devRoot file (`ClaudeHookConfigWriter.writeHookConfig` / `writeGatewayEnv`) are read-modify-write on their own keys only, so the newly scaffolded `permissions` block survives hook/gateway updates, and vice versa.
- `scaffoldNeverTouchesAPreExistingSettingsJSON` locks in that a user's `settings.json` is never read or written — the original data-loss bug can't silently regress.

**Concerns:**
- None blocking. (Green) `mergeSettings` falls back to the template verbatim when the existing `settings.local.json` is unparseable JSON (`Scaffolder.swift:592-600`) — a user hand-edit with a stray comma gets silently replaced on next launch, losing the Crow-written hooks/gateway `env` along with it. This matches the existing convention in `writeGatewayEnv`/`writeHookConfig` (unparseable → treated as empty), and the file would be broken for Claude Code anyway, so it's a consider-only: leaving the file alone (or writing a `.bak`) would be gentler, and a test for this path would document the intended behavior either way.

### Code Quality
- (Green) `scaffold()` writes the merge result unconditionally (`Scaffolder.swift:156`), so even a no-op merge rewrites the file every launch — content is byte-identical (the test verifies the *returned string*), but the inode/mtime churn can trip file watchers and slightly undercuts the doc comment's "leaves the file untouched" claim. Consider skipping the write when `mergeSettings` returns the existing content unchanged.
- (Green) The `permissions` branch of the merge only unions `allow` (`Scaffolder.swift:606-616`). Today's template has no `deny`/`ask`, so this is fine now, but if the bundled template ever grows another `permissions` subkey it will silently never reach files that already have a `permissions` object.
- Test suite is thorough and well-targeted: user additions survive, missing bundled entries are re-added, custom top-level keys and a user `sandbox` override win, byte-for-byte no-op, a full `scaffold()` relaunch round-trip, and two tests pinning `settings.json` untouchability. Verified locally: all 285 tests pass across 33 suites, including the 9 new ones; `swift build` clean.
- Docs (`CONTRIBUTING.md`, `docs/architecture.md`, `docs/configuration.md`, `docs/getting-started.md`, batch-workspace skill) all updated consistently with the new file split.
- Note (no change needed): the review-clone path (`SessionService.swift:1894`) still writes the bundled template to the *clone's* `.claude/settings.json` — that's a Crow-owned ephemeral clone, not the devRoot file this PR protects, so it's consistent with the new contract.

### Summary Table

---

dgershman (approved):
## Code & Security Review

Focused fix for #595: `Scaffolder.scaffold(...)` no longer overwrites the user-owned `{devRoot}/.claude/settings.json` on every launch. Crow's required permissions now land in Claude Code's local-override layer, `settings.local.json`, via a guarded merge. Reviewed the diff (7 files), traced the interaction with the sibling `ClaudeHookConfigWriter.writeGatewayEnv`, built, and ran the new suite.

### Critical Issues
None.

### Security Review
**Strengths:**
- **0o600 re-assertion is a real catch.** `write(...atomically: true)` renames a fresh temp over the target and resets the mode to the umask default (~0o644). Since `settings.local.json`'s `env` block can carry a resolved gateway bearer token, `scaffold` re-applies `0o600` after every write (`Scaffolder.swift:166`), matching `writeGatewayEnv` (`ClaudeHookConfigWriter.swift:148`). The comment correctly identifies the Settings → "Re-scaffold" path (no trailing `writeGatewayEnv`) as the case that would otherwise leave a token-bearing file group/world-readable. `rescaffoldReAddsMissingPermissionAndRestoresOwnerOnly` locks this in.
- **No token clobber.** The bundled template has only `permissions` + `sandbox` (no `env`), and `mergeSettings` only fills *missing* top-level keys, so the gateway `env` written by `writeGatewayEnv` is never overwritten by a later scaffold. Steady-state `.upToDate` skips the write entirely, so the two writers compose cleanly (as the `SessionService.swift:1034` comment notes).
- **Merge is additive-only for permissions** — union of `allow`, nothing removed — so no permission is silently dropped on upgrade.

**Concerns:** None blocking.

### Code Quality
- Merge semantics are correct and precisely documented; `SettingsMergeOutcome` cleanly separates "write" from "leave the file untouched," and the inode/mtime/mode invariant is directly asserted (`steadyStateScaffoldSkipsRewriteAndLeavesFileAlone`).
- Migration is handled: an existing Crow-written `settings.json` is left in place and its `allow` entries still merge at load time, so no permission-prompt regression for upgraders. `scaffoldNeverTouchesAPreExistingSettingsJSON` verifies Crow won't touch it.
- Verified locally: `swift test --filter ScaffolderSettingsMergeTests` → 12/12 pass (build required the pre-existing `scripts/generate-build-info.sh` step; unrelated to this PR).

- **`sandbox` precedence nuance vs. the docs.** The real template writes a `sandbox` block into `settings.local.json`, which outranks `settings.json` in Claude Code's precedence. So a user who sets `sandbox` in their own `settings.json` will find it overridden by Crow's local file. The literal doc claim ("Crow never reads or writes `settings.json`") is accurate and test-verified, but "settings.json is entirely the user's own" slightly oversells it for the `sandbox` key specifically. Not a regression (pre-fix, Crow wiped the whole file), and `userSandboxOverrideWins` covers overriding via `settings.local.json` — just worth a doc footnote that sandbox is best overridden there.
- **`permissions` sub-keys beyond `allow`.** `mergeSettings` only unions `permissions.allow`; if the template later grows `permissions.deny`/`ask`, those wouldn't propagate. Fine today (template has only `allow`) — flagging as a latent limitation if the template expands.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. Correct, well-tested, and a genuine security improvement over the prior overwrite behavior.

---

PR #597, opened 2026-07-06
Auto-detect tmux server crash and reattach + claude --continue (closes #588)

### LINKED ISSUE #588: Auto-detect tmux server crash and reattach + claude --continue affected sessions (instead of re-running the initial plan)
## Summary

When the **tmux server crashes**, Crow does not automatically recover the affected sessions. Instead, each session — when I click it in the left sidebar — comes back up looking like it **re-ran its initial prompt/plan and failed**, and I'm left with a dead-ish terminal that suggests I run `claude --continue` / `claude --resume` by hand — but Crow never actually does it.

Crow should **detect a tmux-server crash and automatically re-attach + `claude --continue`** every affected session, instead of re-running the initial plan and leaving a manual suggestion.

## Repro / symptom

1. Have several active work sessions running Claude Code (each is a tmux window under the shared `crow-cockpit` session).
2. The tmux **server** crashes (whole server, not one window).
3. Click through the sessions in the left sidebar. Each one appears to **re-attempt its initial prompt (the big plan prompt) and fail**, then shows a "you can `claude --continue` / `--resume`" affordance.
4. That affordance does nothing automatically — I have to intervene per session, and even then it doesn't reliably continue the prior Claude conversation.

**Expected:** on a tmux-server crash, Crow detects it, re-attaches the windows, and relaunches each session with `claude --continue` (resume the existing conversation) — **not** a re-paste of the initial prompt — with a clear "tmux crashed, reconnecting…" status rather than a silent/failed re-run.

## Grounding (current behavior, from the code)

Crow runs every terminal as a **window inside one shared tmux session** named `crow-cockpit` (`Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:889`), attached via a single Ghostty cockpit surface (`TmuxBackend.cockpitSurface`, ~`:428-439`).

**Crash detection is only implicit / symptomatic.** There's no explicit "tmux server crashed" signal:
- On re-attach, `TmuxBackend.adoptTerminal` (`TmuxBackend.swift:254-277`) calls `TmuxController.listWindowIndices()`; if the window/server is gone it throws `TmuxBackendError.windowNotFound(...)` (`:264`) — it does **not** distinguish "server dead" from "one window gone".
- `TmuxController.hasSession()` (`Packages/CrowTerminal/Sources/CrowTerminal/TmuxController.swift:98-100`, `tmux has-session -t crow-cockpit`) is the signal that *could* tell "server is gone", but it isn't consulted on the crash path.
- The adoption failure is caught **silently with just a log** in `SessionService.rehydrateTerminalSurface` (`Sources/Crow/App/SessionService.swift:340-341`), which then falls through to registering a fresh window (`:350`).
- There *is* a "tmux is not responding → Restart tmux server" alert for a **hung** server (`Sources/Crow/App/AppDelegate.swift:202-224`, 2s watchdog), and a manual `restartTmuxServer()` path (`SessionService.swift:713-730`) — but a **crash** is discovered as an error on the next command, not by the watchdog, and recovery requires user action.

**The initial-prompt-vs-`--continue` decision** lives in `ClaudeCodeAgent.autoLaunchCommand` (`Packages/CrowClaude/Sources/CrowClaude/ClaudeCodeAgent.swift:44-91`): `.review`/`.job` sessions have an initial prompt file; `.work`/`.manager` fall through to `claude --continue` (`:91`). So the *restored*-launch path (`SessionService.launchAgent` → `resolveLaunch` → `.launchAgent`) resumes correctly.

**The likely reason it re-runs the initial plan instead of continuing:** the first-time launch path pastes the session's *initial prompt command* (`resolveLaunch` → `.pastePending(command)` → `pasteDeferredLaunch`), driven by `terminal.command`. On app-restart hydrate, Crow deliberately **clears** the stored `claude …` command from managed terminals (`SessionService.swift:98-106`) so the restore path takes `--continue`. But a **tmux-server crash while the app is still running** re-registers the window **without** going through that hydrate-clear — so `terminal.command` still holds the original prompt command, `resolveLaunch` returns `.pastePending`, and Crow **re-pastes the initial plan** (matching the symptom) instead of `--continue`. This is the piece to confirm and fix.

## Proposed change

1. **Detect a tmux-server crash explicitly.** In the adopt/re-attach path (`TmuxBackend.adoptTerminal` / `rehydrateTerminalSurface`), consult `hasSession()` and distinguish a dead **server** (all windows gone) from a single missing window — surface a new `TmuxError.serverCrashed` (add to `TmuxController.swift`) rather than a generic `windowNotFound`. Optionally add a lightweight periodic/`didBecomeActive` health check (`tmux has-session`) so a crash is noticed proactively, not only on the next click.
2. **Auto-recover on crash.** When a server crash is detected, run the existing recovery (`restartTmuxServer()` / `rebuildAllSurfaces(forceRegister: true)`, `SessionService.swift:713-730`) **automatically** for the affected sessions — re-register the windows and relaunch — without requiring the user to click through each session.
3. **Always resume, never re-run the initial plan, after a crash.** Ensure the post-crash relaunch takes the `claude --continue` branch, not the initial-prompt paste. Concretely: on a crash-driven re-register, clear/ignore the stored initial-prompt `terminal.command` for managed terminals the same way hydrate does (`SessionService.swift:98-106`), so `resolveLaunch` returns `.launchAgent` (→ `autoLaunchCommand` → `--continue`) instead of `.pastePending`. Guard against re-dispatching a session's initial plan once it has already been dispatched.
4. **Surface it clearly.** Replace the generic "Terminal didn't become ready in time / shell took too long" overlay (`Packages/CrowUI/Sources/CrowUI/SessionDetailView.swift:614-656`) — which is what the user hits today — with a crash-specific "tmux server crashed — reconnecting and resuming…" status while auto-recovery runs, so the failure is legible and the manual `claude --continue` suggestion isn't the only path.

## Acceptance criteria

- Killing the tmux server (`tmux kill-server`) while sessions are running is **detected** by Crow (not just discovered on the next sidebar click).
- Each affected session **auto-reattaches and relaunches with `claude --continue`**, resuming the prior Claude conversation — **without** re-running/re-pasting the initial prompt/plan, and without per-session manual clicks.
- A session that never got past its initial prompt is handled sanely (doesn't lose the prompt; doesn't double-dispatch it).
- The user sees a clear "tmux crashed → reconnecting/resuming" status instead of a generic readiness-timeout message with a dead `claude --continue` suggestion.
- Manager/jobs/review session kinds recover with their correct relaunch semantics.

## Notes / where to look
- tmux service + attach/adopt: `Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift` (`registerTerminal` ~`:175`, `adoptTerminal` `:254-277`, `ensureRunningServer` ~`:476`, cockpit session name `:889`); `TmuxController.swift` (`hasSession` `:98`, `listWindowIndices` `:102`).
- crash catch + relaunch: `Sources/Crow/App/SessionService.swift` — `rehydrateTerminalSurface` (`:275`, silent catch `:340-341`, re-register `:350`), hydrate command-clear (`:98-106`), readiness→`launchAgent` callback (`:374-402`), `launchAgent` (`:509-617`), `restartTmuxServer` (`:713-730`), `rebuildAllSurfaces(forceRegister:)` (`:226-232`).
- resume vs initial prompt: `Packages/CrowClaude/Sources/CrowClaude/ClaudeCodeAgent.swift:44-91` (`--continue` at `:91`).
- user-facing overlay: `Packages/CrowUI/Sources/CrowUI/SessionDetailView.swift:614-656`; hung-server alert `Sources/Crow/App/AppDelegate.swift:202-224`.
- **First step: confirm** the crash-path re-register keeps `terminal.command` populated (→ `.pastePending` re-paste of the initial plan) vs the hydrate path that clears it (→ `--continue`) — that's the crux of "it re-ran the plan."

dhilgaertner:
Closes #588

## What

When the tmux server crashes while Crow is running, Crow now detects it immediately, auto-reattaches every terminal, and relaunches each session's agent with `claude --continue` — with a crash-specific "tmux server crashed — reconnecting and resuming…" status instead of the generic readiness-timeout overlay. No per-session clicks.

## Root cause (corrected from the ticket)

- **No detection existed.** The cockpit attach client (`tmux attach-session`) dies with the server, but nothing subscribed to `XTermSurfaceView.onProcessExit`. Adopt failures were swallowed as generic `windowNotFound` (a dead server and a single closed window were indistinguishable — `ensureRunningServer` silently resurrects an empty cockpit first). The 2s watchdog only routes *hung*-server timeouts to an alert; a *crashed* server yields `cliFailed`, which was never surfaced.
- **The ticket's `terminal.command` hypothesis didn't hold**: `resolveLaunch` reads the in-memory `pendingLaunchCommands`, not `terminal.command`, and managed terminals are registered/persisted with `command: nil`. The existing rebuild machinery (`rebuildAllSurfaces(forceRegister: true)` + persisted `reviewPromptDispatched`) already yields `--continue` — the gaps were purely: no crash detection, no auto-trigger of that recovery, and no legible status.

## How

**Detection** (`TmuxBackend`)
- New `onCockpitExit`, wired to the cockpit surface's `onProcessExit` inside `cockpitSurface()` (recreated surfaces re-arm automatically). `shutdown()` reordered to destroy the surface *before* `kill-server` — its `waitUntilExit` pumps the run loop, so the old order would let a deliberate restart masquerade as a crash.
- New `onServerLost` fires when a cached controller's cockpit session has vanished mid-run (lazy secondary signal on the next tmux command).
- `adoptTerminal` throws a new `TmuxBackendError.serverCrashed` (whole server resurrected) distinct from `windowNotFound` (one window gone).

**Auto-recovery** (`SessionService`)
- `handleCockpitClientExit()`: server dead → `handleTmuxServerCrash()`; server alive (user detach) → `recycleCockpitSurface()` light reattach that leaves all windows running (behavior change: prefix-d previously left a dead view silently).
- `handleTmuxServerCrash()`: reentrancy-guarded + 10s crash-loop debounce + 120s fallback flag clear; runs `recycleTmuxServerAndRebuild()` (extracted from, and shared with, the manual Restart tmux Server menu path). Dispatched sessions resume via `autoLaunchCommand` → `claude --continue`; a never-dispatched initial prompt survives in `pendingLaunchCommands` and pastes exactly once; Manager/review/job semantics unchanged.
- Defensive guard: `rebuildAllSurfaces(forceRegister:)` nils a managed terminal's stored claude command (mirrors the hydrate-clear) so re-running the initial plan verbatim is structurally impossible.

**UX**
- New `AppState.tmuxCrashRecovering` drives the crash wording on both the loading and timed-out overlays (Retry + Copy diagnostics kept as the escape hatch); cleared when every tracked terminal settles. The hung-server alert is suppressed during recovery; a manual restart clears a stale flag.
- `TerminalSurfaceView.attach` now removes the previous destroyed cockpit view and re-points the search bar at the replacement (pre-existing leak on manual restart, user-visible with auto-recovery).

## Testing

- `Tests/CrowTests/TmuxCrashRecoveryTests.swift` (7): flag + re-arm, reentrancy, debounce, settle-based flag clear, pending-prompt survival/single paste, managed-only command clear, manual-restart interplay.
- `Packages/CrowClaude/.../ClaudeCodeAgentLaunchTests.swift` (3): `autoLaunchCommand` resume matrix (work/manager always `--continue`; review/job prompt-file first launch only).
- `Packages/CrowTerminal/.../TmuxBackendTests.swift` (+4 real-tmux integration): shutdown on an already-dead server recovers; external `kill-server` fires `onServerLost` exactly once and self-heals; `serverCrashed` vs `windowNotFound`; surface recycle no-op safety.
- Full gate: `make test` — all packages + root (283 root tests / 33 suites) green.
- Note: the full GUI flow (cockpit PTY exit → overlay) needs a visible window and wasn't driven headless — this branch's dev session runs *inside* the production Crow tmux server, so killing the live socket would sever active sessions. Manual check: run the patched app with 2–3 sessions, `tmux -S $TMPDIR/crow-tmux.sock kill-server`, expect immediate crash status + auto-resume via `--continue`; `tmux detach-client` should silently reattach.

### DIFF
--- a/Packages/CrowClaude/Tests/CrowClaudeTests/ClaudeCodeAgentLaunchTests.swift
+++ b/Packages/CrowClaude/Tests/CrowClaudeTests/ClaudeCodeAgentLaunchTests.swift
@@ -1,0 +1,54 @@
+import Foundation
+import Testing
+import CrowCore
+@testable import CrowClaude
+
+/// Locks in the resume-vs-initial-prompt decision in
+/// `ClaudeCodeAgent.autoLaunchCommand` (#588): work/manager sessions always
+/// resume with `--continue`; review/job sessions read their pre-written
+/// prompt file exactly once (`reviewPromptDispatched == false`) and resume
+/// with `--continue` on every relaunch after that — including the rebuild
+/// after a tmux server crash.
+@Suite("ClaudeCodeAgent.autoLaunchCommand resume semantics")
+struct ClaudeCodeAgentLaunchTests {
+
+    private let agent = ClaudeCodeAgent()
+
+    private func command(kind: SessionKind, dispatched: Bool) -> String? {
+        agent.autoLaunchCommand(
+            session: Session(name: "s", kind: kind, reviewPromptDispatched: dispatched),
+            worktreePath: "/tmp/wt",
+            remoteControlEnabled: false,
+            autoPermissionMode: false,
+            telemetryPort: nil
+        )
+    }
+
+    @Test func workAndManagerAlwaysResume() throws {
+        for kind in [SessionKind.work, .manager] {
+            for dispatched in [false, true] {
+                let cmd = try #require(command(kind: kind, dispatched: dispatched))
+                #expect(cmd.hasSuffix(" --continue\n"))
+                #expect(!cmd.contains("$(cat"))
+            }
+        }
+    }
+
+    @Test func reviewAndJobReadPromptFileOnFirstLaunchOnly() throws {
+        let review = try #require(command(kind: .review, dispatched: false))
+        #expect(review.contains("\"$(cat /tmp/wt/.crow-review-prompt.md)\""))
+        #expect(!review.contains("--continue"))
+
+        let job = try #require(command(kind: .job, dispatched: false))
+        #expect(job.contains("\"$(cat /tmp/wt/.crow-job-prompt.md)\""))
+        #expect(!job.contains("--continue"))
+    }
+
+    @Test func reviewAndJobResumeAfterPromptDispatched() throws {
+        for kind in [SessionKind.review, .job] {
+            let cmd = try #require(command(kind: kind, dispatched: true))
+            #expect(cmd.hasSuffix(" --continue\n"))
+            #expect(!cmd.contains("$(cat"))
+        }
+    }
+}
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -54,6 +54,12 @@ public final class AppState {
     /// and enables the "Restart Manager" action. Reset when the Manager relaunches.
     public var managerProcessExited: Bool = false
 
+    /// `true` while Crow auto-recovers from a tmux server crash (#588): set at
+    /// detection, cleared once every tracked terminal settles (`.shellReady`
+    /// or `.timedOut`) or by a fallback timeout. Drives the crash-specific
+    /// "tmux server crashed — reconnecting and resuming…" terminal overlay.
+    public var tmuxCrashRecovering: Bool = false
+
     /// The agent seeded into new sessions when the caller doesn't pick one.
     /// Mirrors `AppConfig.defaultAgentKind` so creation flows can read the
     /// current default without a config round-trip.
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TerminalSurfaceView.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TerminalSurfaceView.swift
@@ -82,7 +82,19 @@ public struct TerminalSurfaceView: NSViewRepresentable {
     /// `setFrameSize` here races with autolayout and is unnecessary.
     @MainActor
     private static func attach(surface: XTermSurfaceView, to container: NSView) {
-        container.addSubview(surface)
+        // A previous (destroyed) surface may still be parented here — a server
+        // restart/crash-recovery replaces the shared cockpit surface (#588).
+        // Drop the dead view so it doesn't sit over the live one.
+        for stale in container.subviews where stale is XTermSurfaceView && stale !== surface {
+            stale.removeFromSuperview()
+        }
+        // Keep the surface below any existing search bar — a plain addSubview
+        // would stack it on top and hide the bar.
+        if let existingBar = container.subviews.first(where: { $0 is TerminalSearchBar }) {
+            container.addSubview(surface, positioned: .below, relativeTo: existingBar)
+        } else {
+            container.addSubview(surface)
+        }
         surface.translatesAutoresizingMaskIntoConstraints = false
         NSLayoutConstraint.activate([
             surface.leadingAnchor.constraint(equalTo: container.leadingAnchor),
@@ -92,8 +104,12 @@ public struct TerminalSurfaceView: NSViewRepresentable {
         ])
 
         // Float the Cmd+F search bar over the surface (#471 gap 2). Only
-        // install one per container; later re-parents keep the same bar.
-        if container.subviews.contains(where: { $0 is TerminalSearchBar }) == false {
+        // install one per container; later re-parents keep the same bar —
+        // but re-point it at the current surface in case a restart swapped
+        // the cockpit view out from under it.
+        if let bar = container.subviews.first(where: { $0 is TerminalSearchBar }) as? TerminalSearchBar {
+            bar.setHostSurface(surface)
+        } else {
             let bar = TerminalSearchBar(frame: .zero)
             bar.translatesAutoresizingMaskIntoConstraints = false
             bar.setHostSurface(surface)
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -44,6 +44,19 @@ public final class TmuxBackend {
     /// the caller for normal handling.
     public var onUnresponsive: ((TmuxError) -> Void)?
 
+    /// Fired when the cockpit attach client's PTY exits outside a deliberate
+    /// `shutdown()`/`destroy()` — i.e. the tmux server crashed out from under
+    /// the app, or the user detached the client (#588). Deliberate teardown
+    /// never fires this: `shutdown()` destroys the surface (which nils the
+    /// underlying `onProcessExit`) before touching the server.
+    public var onCockpitExit: ((Int32) -> Void)?
+
+    /// Fired when a cached controller's cockpit session has vanished mid-run —
+    /// the server died while the app was live (#588). A fresh launch (no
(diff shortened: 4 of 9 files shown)


dgershman (approved):
## Code & Security Review

Well-scoped fix for #588 with strong test coverage. I built `CrowTerminal` clean and ran all 14 new tests — the 4 real-tmux integration tests, the 3 `autoLaunchCommand` resume tests, and the 7 root crash-recovery tests all pass. Traced the core control flow and found no blocking issues.

### Critical Issues
None.

### Security Review
**Strengths:**
- No new untrusted-input surface: the change is entirely internal process/UI lifecycle (tmux server lifecycle, surface re-parenting, readiness state).
- The `$(cat .crow-*-prompt.md)` prompt inlining is pre-existing and unchanged; the review/job first-launch gate (`reviewPromptDispatched`, persisted at `SessionService.swift:673-679`) correctly ensures a post-crash rebuild resumes with `--continue` rather than re-running the initial plan.
- No credentials, no network, no injection vectors introduced.

**Concerns:** None.

### Code Quality
Verified the subtle correctness claims hold:

- **`shutdown()` reorder is correct** (`TmuxBackend.swift:150-157`). Destroying the surface before `kill-server` matters because `XTermSurfaceView.destroy()` nils `onProcessExit` synchronously (`XTermSurfaceView.swift:97`) while `pty.onExit` dispatches an async `@MainActor` Task. Since `killServer`'s `waitUntilExit` pumps the run loop, the old order would deliver the attach client's death mid-shutdown and a deliberate restart would masquerade as a crash. New order closes that.
- **`serverCrashed` vs `windowNotFound`** — `rehydrateTerminalSurface`'s catch-all (`SessionService.swift:370`) handles both identically, so introducing the new error case does not regress the app-launch adopt path. On launch `controller` is nil during the first adopt, so `onServerLost` cannot spuriously fire crash recovery.
- **No re-entrant recovery storm** — the flag is raised before `recycleTmuxServerAndRebuild`, `shutdown` nils the controller (so `onServerLost` can't re-fire during rebuild) and destroys the surface first (so `onCockpitExit` can't re-fire), and the reentrancy guard + 10s debounce + 120s fallback are all exercised by tests.

Minor observations (non-blocking, no change requested):
- The 10s debounce (`handleTmuxServerCrash`, `SessionService.swift:872-877`) means a *genuine* second crash 5-9s after a recovery is silently dropped — no overlay, no auto-recovery until the user hits Restart tmux Server. This is the right tradeoff against a tight crash-loop, and the manual menu is the escape hatch; a future "tmux keeps crashing" surfaced state could close the gap.
- `crashRecoveryClearFallback` isn't cancelled in `restartTmuxServer`; a stale task from a prior crash survives a manual restart but no-ops at 120s (it re-checks the flag), so harmless.
- `clearCrashRecoveryFlagIfSettled` scans all `terminalReadiness.values`, so a terminal created concurrently during recovery could hold the flag up until it settles or the 120s fallback fires — self-healing edge case.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 3 Green] findings. Clean build, all 14 new tests green, core control flow verified.

PR #599, opened 2026-07-06
xterm.js: restore Shift+Enter → insert newline in Claude Code (closes #598)

### LINKED ISSUE #598: xterm.js: Shift+Enter no longer inserts a newline in Claude Code (regressed in Ghostty→xterm.js migration #548)
## Summary

After Crow moved its terminal renderer from **Ghostty** to **xterm.js** (commit `a364f8b` — "Replace Ghostty with xterm.js terminal renderer for Intel Mac support" #548), **Shift+Enter no longer inserts a newline in Claude Code** — it submits the prompt as if I'd pressed plain Enter. In the old Ghostty terminal, Shift+Enter inserted a literal newline (multi-line prompt) without submitting.

## Repro

1. Open a Crow session's Claude Code terminal (the xterm.js backend, post-#548).
2. In the Claude Code prompt, type some text, then press **Shift+Enter**.
3. **Expected:** a newline is inserted; the prompt stays open for more input (multi-line).
4. **Actual:** the prompt is submitted — Shift+Enter behaves identically to Enter.

## Root cause (investigation)

> Note: my local checkout is ~62 commits behind `origin/main`, so the file/line refs below are the **xterm.js files as they exist on `origin/main`** (they weren't in my local tree). Re-verify on current `main`.

The migration replaced the native Ghostty key path with xterm.js running in a `WKWebView`, and the modifier-aware Enter handling didn't carry over:

- **Old (Ghostty) path — distinguished Shift+Enter.** `Packages/CrowTerminal/Sources/CrowTerminal/GhosttySurfaceView.swift` captured `NSEvent` keyDown and preserved the Shift modifier into Ghostty's key struct (`translateMods()` ~`:490-503`: `if flags.contains(.shift) { mods |= GHOSTTY_MODS_SHIFT.rawValue }`; `ghostty_surface_key(surface, key)` ~`:298`). Combined with tmux `extended-keys on`, Claude Code received a distinct CSI-u / modifyOtherKeys sequence for Shift+Enter and treated it as "insert newline".
- **New (xterm.js) path — does NOT.** `Packages/CrowTerminal/Sources/CrowTerminal/Resources/xterm/terminal.html` constructs the `Terminal` (~`:64-71`) with no modified-key handling, and forwards raw data via `term.onData(d => handlers.crowInput.postMessage({data: d}))` (~`:95`). By default **xterm.js sends the same byte (`\r`) for both Enter and Shift+Enter** — it doesn't emit a distinct sequence for the Shift modifier on Return — so downstream (PTY → tmux → Claude Code) can't tell them apart, and Claude Code submits.
- **tmux side is already correct** — `Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf` (~`:69-80`) still sets `extended-keys on` + advertises `extkeys` terminal-features. The gap is purely that xterm.js isn't emitting the modified-key sequence into the PTY in the first place.

Net: tmux is ready to carry the modified-key sequence, but xterm.js never produces one for Shift+Enter, so the newline affordance is lost.

## Proposed fix

Make the xterm.js terminal emit a distinct sequence for Shift+Enter that Claude Code interprets as "insert newline", instead of a plain `\r`. Options (implementer to pick + verify the exact bytes against what Claude Code expects — likely the CSI-u form `\x1b[13;2u` that tmux `extended-keys` already carries, or the `\x1b\r` / `\n` form that iTerm2/VSCode use via `/terminal-setup`):

1. **Custom key handler in xterm.js (preferred).** In `terminal.html`, add `term.attachCustomKeyEventHandler(...)` (or a keydown listener) that detects `Enter` with `shiftKey === true`, writes the newline sequence to the PTY (via the same `crowInput` message path), and returns `false` to suppress xterm.js's default `\r`. This is the most targeted fix and keeps the binding regardless of xterm.js's default modifier handling.
2. **Enable xterm.js modified-key emission** so it produces CSI-u sequences for modified keys (leveraging the tmux `extended-keys` already configured), if xterm.js exposes a supported option/addon for that in the version vendored here — then Claude Code + tmux handle the rest as they did under Ghostty.

Either way: verify the byte sequence end-to-end (xterm.js → `PTYProcess.write` → tmux → Claude Code) and confirm plain Enter still submits.

## Acceptance criteria

- In a Crow xterm.js Claude Code terminal, **Shift+Enter inserts a newline** (multi-line prompt) and does **not** submit.
- Plain **Enter still submits** (no regression).
- Works inside the tmux window (the real runtime path), not just a bare terminal.
- Ideally note whether other modified keys that regressed in the Ghostty→xterm.js move (e.g. Option/Alt combos, Ctrl sequences Claude Code relies on) need the same treatment — or file a follow-up.

## Notes / where to look (on `origin/main`)
- xterm.js terminal init + data forward: `Packages/CrowTerminal/Sources/CrowTerminal/Resources/xterm/terminal.html` (`Terminal` ctor ~`:64-71`, `term.onData` ~`:95`).
- WKWebView ↔ PTY bridge: `Packages/CrowTerminal/Sources/CrowTerminal/XTermSurfaceView.swift` (`handleInput`/`crowInput` ~`:190`), `PTYProcess.swift` `write()` (~`:115-130`) — confirm nothing filters escape bytes.
- tmux (already correct, no change): `Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf` (`extended-keys on` ~`:72`).
- send path: `TmuxBackend.swift` `sendText()` (~`:322-353`).
- Migration commit: `a364f8b` "Replace Ghostty with xterm.js terminal renderer for Intel Mac support (#548)". Old reference implementation: `GhosttySurfaceView.swift` (`translateMods` ~`:490-503`).

dhilgaertner:
Closes #598

## Problem

The Ghostty→xterm.js migration (#548) lost modifier-aware Enter handling: xterm.js emits the same bare `\r` for Enter and Shift+Enter, so Claude Code submitted the prompt instead of inserting a newline.

## Fix

Add `term.attachCustomKeyEventHandler(...)` in `terminal.html` that intercepts modified Enter on keydown, sends a distinct sequence through the existing `crowInput` → PTY bridge, and returns `false` to suppress xterm.js's default `\r`:

- **Shift+Enter → CSI-u `\x1b[13;2u`** — the sequence `crow-tmux.conf`'s `extended-keys on` + `extkeys` terminal-features were already configured to carry. tmux forwards it to apps that negotiate extended keys (Claude Code → insert newline) and **downgrades it to plain `\r`** for apps that don't, so nothing leaks into vim/shells.
- **Option+Enter → `\x1b\r`** (ESC CR) — Claude Code's other native insert-newline chord, which also regressed in #548. `macOptionIsMeta` stays off so Option-key composed characters keep working for regular typing.

Plain Enter stays on the default path and still submits. No tmux config changes (already correct). Added a bundled-resource test pinning the handler's presence, since the xterm resources get re-vendored wholesale (`Resources/xterm/VERSION`).

## Verification

Ran end-to-end against a **real Claude Code prompt inside a tmux server started from the bundled `crow-tmux.conf`**, injecting the exact bytes through a PTY-attached client (`TERM=xterm-256color`, same as `PTYProcess`):

- `alpha` + `\x1b[13;2u` + `bravo` + `\x1b\r` + `charlie` → three lines in the prompt box, **no submission**.
- Plain `\r` in a shell pane still executes commands.
- Raw-mode pane byte capture: client `\x1b[13;2u` arrives as plain `\r` in a non-negotiating pane (clean degradation); `\x1b\r` passes through as Meta+Enter.
- `swift build` + `make test`: all suites pass. (CrowGit `RebaseTests` fail pre-existing and unrelated — the environment has no global git `user.email`/`user.name`, and those tests commit in temp repos outside any `includeIf` scope; they fail identically on `main`.)

## Follow-up notes (other Ghostty-era modified keys)

Ctrl combos (Ctrl+C/D/R/etc.) go through xterm.js's standard control-byte handling and were unaffected. General Option-as-Meta (e.g. Option+B/F word motion) remains off, matching the current ctor config — turning on `macOptionIsMeta` globally would break Option-composed characters, so that's left as a deliberate non-change; can file a follow-up if someone misses those bindings.

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/Resources/xterm/terminal.html
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/Resources/xterm/terminal.html
@@ -101,6 +101,38 @@
 
         window.crowFit = reportSize;
 
+        // xterm.js emits the same bare \r for Enter and Shift+Enter, so
+        // Claude Code can't tell them apart and submits on both (#598).
+        // Intercept modified Enter and send a distinct sequence through the
+        // same crowInput bridge:
+        //  - Shift+Enter → CSI-u \x1b[13;2u. crow-tmux.conf's extended-keys
+        //    setup carries it to apps that negotiate extended keys (Claude
+        //    Code → insert newline) and downgrades it to plain \r for apps
+        //    that don't, so nothing leaks into vim/shells.
+        //  - Option+Enter → \x1b\r (ESC CR), Claude Code's other native
+        //    insert-newline chord. macOptionIsMeta stays off so Option-key
+        //    composed characters keep working for regular typing.
+        // Returning false suppresses xterm.js's default \r for every event
+        // phase (keydown/keypress/keyup) of the combo.
+        term.attachCustomKeyEventHandler(function (e) {
+          if (e.key !== 'Enter' || e.ctrlKey || e.metaKey) {
+            return true;
+          }
+          if (e.shiftKey && !e.altKey) {
+            if (e.type === 'keydown') {
+              handlers.crowInput.postMessage({ data: '\x1b[13;2u' });
+            }
+            return false;
+          }
+          if (e.altKey && !e.shiftKey) {
+            if (e.type === 'keydown') {
+              handlers.crowInput.postMessage({ data: '\x1b\r' });
+            }
+            return false;
+          }
+          return true;
+        });
+
         term.onData(function (data) {
           handlers.crowInput.postMessage({ data: data });
         });
--- a/Packages/CrowTerminal/Tests/CrowTerminalTests/BundledResourcesTests.swift
+++ b/Packages/CrowTerminal/Tests/CrowTerminalTests/BundledResourcesTests.swift
@@ -63,6 +63,20 @@ struct BundledResourcesTests {
         #expect(body.contains(#"send-keys -X select-line ; run -d0.3 ; send-keys -X copy-pipe-no-clear "pbcopy""#))
     }
 
+    @Test func terminalHTMLHandlesModifiedEnter() throws {
+        // #598: xterm.js sends the same \r for Enter and Shift+Enter, so the
+        // host page must intercept modified Enter and emit distinct sequences
+        // (CSI-u \x1b[13;2u for Shift+Enter, ESC CR for Option+Enter) or
+        // Claude Code submits instead of inserting a newline. The xterm
+        // resources get re-vendored wholesale (see Resources/xterm/VERSION),
+        // so pin the handler's presence against an accidental overwrite.
+        let url = try #require(BundledResources.terminalHTMLURL)
+        let body = try String(contentsOf: url, encoding: .utf8)
+        #expect(body.contains("attachCustomKeyEventHandler"))
+        #expect(body.contains(#"\x1b[13;2u"#))
+        #expect(body.contains(#"\x1b\r"#))
+    }
+
     @Test func tmuxConfHasNoBarePrefixUnbind() throws {
         // #473: a bare `unbind-key -a` (no `-T`) defaults to the prefix
         // table, which is empty/non-existent after the first clear on


dgershman (approved):
## Code & Security Review

Small, well-scoped fix restoring modifier-aware Enter handling lost in the Ghostty→xterm.js migration (#548). Verified the diff, the tmux config dependency, and ran the tests.

### Critical Issues
None.

### Security Review
**Strengths:**
- The custom key handler posts only **hard-coded constant byte sequences** (`\x1b[13;2u`, `\x1b\r`) for two specific key combos — no user-controlled data flows into the `crowInput` → PTY bridge, so no injection surface is introduced.
- Reuses the existing `webkit.messageHandlers.crowInput` channel rather than adding new bridge plumbing; attack surface is unchanged.

**Concerns:** None.

### Code Quality
- Handler logic is correct and defensive: `keydown`-guarded posting prevents double-emission across event phases, while `return false` suppresses xterm.js's default `\r` for all phases of the combo. Ctrl/Meta are explicitly excluded (`return true`), and `shift && !alt` / `alt && !shift` cleanly separate the two chords, leaving Shift+Alt+Enter on the default path.
- The dependency on `crow-tmux.conf` is real and present — confirmed `set -gs extended-keys on` (line 98) and `terminal-features ',screen*:extkeys,xterm*:extkeys'` (line 106), which is what carries the CSI-u sequence to negotiating apps and downgrades it to `\r` elsewhere. The graceful-degradation claim holds.
- The new `terminalHTMLHandlesModifiedEnter()` test follows the established string-pinning convention in `BundledResourcesTests` and is well-justified given the xterm resources are re-vendored wholesale. It passes locally along with the rest of the suite (8/8). Raw-string patterns (`#"\x1b[13;2u"#`, `#"\x1b\r"#`) match the HTML literals exactly.

- The pinning test asserts the handler's *presence* but not its *behavior* — a logic regression (e.g. an inverted `shiftKey` check) would slip through. Acknowledged as an accepted limitation matching the file's testing style; the PR's manual end-to-end PTY verification covers the behavioral gap. No action required.

### Summary Table

---

PR #611, opened 2026-07-08
fix(tmux): drain stdout while waiting so large capture-pane doesn't deadlock (CROW-606)

dgershman:
## Summary
- Web terminal scrollback replay (CROW-606) silently no-op'd on reconnect: `capture-pane -pe` of a rich TUI pane exceeds the ~64KB pipe buffer, `TmuxController.run()` waited for exit before reading stdout, the 2s watchdog fired, and `try?` in `replayData` swallowed it.
- Drain stdout/stderr concurrently while waiting; give `capturePane` a longer timeout; add a regression test; document that `history-limit` only applies to new windows.

## Test plan
- [x] `swift test --package-path Packages/CrowTerminal --filter capturePane` (2/2)
- [ ] Restart crowd on this branch, reload the browser, confirm prior scrollback appears after reconnect
- [ ] Confirm wheel-scroll of retained history still does not enter tmux copy-mode


dgershman:
Closing — fix should land on #594 / feature/crow-593-web-ui-parity, not a nested PR.

PR #612, opened 2026-07-08
fix(tmux): drain stdout while waiting so large capture-pane doesn't deadlock (CROW-606)

dgershman:
## Summary
- Web terminal scrollback replay (CROW-606) silently no-op'd on reconnect: `capture-pane -pe` of a rich TUI pane exceeds the ~64KB pipe buffer, `TmuxController.run()` waited for exit before reading stdout, the 2s watchdog fired, and `try?` in `replayData` swallowed it.
- Drain stdout/stderr concurrently while waiting; give `capturePane` a longer timeout; add a regression test; document that `history-limit` only applies to new windows.

## Test plan
- [x] `swift test --package-path Packages/CrowTerminal --filter capturePane` (2/2)
- [ ] Restart crowd on this branch, reload the browser, confirm prior scrollback appears after reconnect
- [ ] Confirm wheel-scroll of retained history still does not enter tmux copy-mode


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface: the change is confined to child-process pipe handling and a timeout bump; no untrusted input parsing, no auth paths, no secrets.
- The `crow-tmux.conf` change is documentation only (clarifies that `history-limit` applies only to newly created windows).
- Error paths still capture stdout/stderr into `TmuxError.cliFailed` without leaking anything new.

**Concerns:**
- None.

### Code Quality
The core fix is correct and idiomatic: draining stdout/stderr on background threads *while* waiting for the child is the standard resolution for the pipe-buffer deadlock, and the implementation is sound —
- `DispatchGroup.enter/leave` + `group.wait()` gives a proper happens-before edge, so the `@unchecked Sendable` `PipeBox` reads on the caller thread after `wait()` are race-free (`TmuxController.swift:60-81`).
- `watchdog.didFire` is still checked after the drain, so a SIGTERM'd child correctly surfaces as `.timedOut` rather than a partial-output success (`TmuxController.swift:83`).
- The regression test genuinely reproduces the failure mode (>64 KB through the pipe with the pre-fix code deadlocking), pins `history-limit` via a server conf so the pane can retain the blob, polls with a sentinel to avoid raciness, and cleans up server/socket/conf in `defer` (`TmuxControllerTests.swift:134-160`). It passes in ~0.3s locally; the full `TmuxControllerTests` suite (8 tests) and `swift build` are clean.

Consider-level (Green) notes, non-blocking:
- `capturePane` can now block its caller for up to 10s (was 2s) in the hung-server case. `TerminalCockpit.replayData` is invoked from the WebSocket inbound-message loop (`TerminalWebSocket.swift:103`), which runs on a Swift-concurrency cooperative thread — a wedged tmux would stall that connection's keystroke processing (and occupy a pool thread) for the full 10s. This blocking pattern is pre-existing and the timeout is only a worst-case cap, but if it ever bites, wrapping the capture in a detached/blocking-friendly context would isolate it.
- Each `run()` now parks two GCD global-queue threads for the child's lifetime. Calls are short and effectively serial today, so this is fine; just something to keep in mind if `run()` ever becomes highly concurrent.
- `group.wait()` is unbounded, so a grandchild that inherited the pipe write-end and outlived tmux would hang `run()` past the watchdog. In practice tmux daemonizes with fds redirected to `/dev/null` and pane processes hold PTY fds (not these pipes), and the old code had the identical hazard in `readDataToEndOfFile()` — not a regression.

### Summary Table

---

PR #616, opened 2026-07-08
feat(web): enable/disable toggle on Jobs list rows

### LINKED ISSUE #615: Web UI: enable/disable toggle on Jobs list rows

## Problem
Settings → Jobs lists each job with Run / Duplicate / Delete, but enabling or disabling a job requires opening the job editor (`enabled` lives on the draft form). Want a per-row toggle on the list itself.

## Direction
Add an enable/disable control on each job row in `renderJobs` (`settings.js`) that flips `job.enabled`, marks dirty, and persists via the existing save path (or an immediate `set-config` if that matches other list-row actions). Mirror the desktop if it already has an inline enable control.

## Acceptance
- Jobs list shows an on/off toggle per row without entering the job editor.
- Toggling updates `job.enabled` and survives save / reload; disabled jobs still show as disabled in the subtitle.

dgershman:
## Summary
- Adds a per-row enable/disable switch on Settings → Jobs so operators can flip `job.enabled` without opening the job editor
- Persists via the existing dirty/save path (same as Duplicate/Delete); subtitle still shows `· disabled` when off

Closes #615

## Test plan
- [ ] Open Settings → Jobs and confirm each job row shows an on/off toggle before Run / Duplicate / Edit / Delete
- [ ] Toggle a job off → subtitle shows `· disabled` and Save becomes enabled
- [ ] Save, reload Settings → Jobs → job remains disabled with toggle off
- [ ] Toggle back on, save, reload → enabled again and subtitle no longer says disabled
- [ ] Confirm Run / Duplicate / Delete still work as before

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/settings.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/settings.css
@@ -208,7 +208,8 @@
 .st-row-main { flex: 1; min-width: 0; }
 .st-row-title { font-size: 12px; font-weight: 600; color: var(--text-primary); }
 .st-row-sub { font-size: 11px; color: var(--text-muted); margin-top: 2px; }
-.st-row-actions { display: flex; gap: 6px; flex: none; }
+.st-row-actions { display: flex; gap: 6px; flex: none; align-items: center; }
+.st-row-switch { margin-right: 2px; }
 .st-icon-btn {
   display: inline-flex; align-items: center; justify-content: center;
   width: 28px; height: 28px; padding: 0;
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/settings.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/settings.js
@@ -847,6 +847,18 @@
         setTimeout(() => { run.disabled = false; run.title = 'Run now'; setRowIcon(run, 'play'); }, 1500);
       };
       row.querySelector('.st-row-actions').insertBefore(run, row.querySelector('.st-row-actions').firstChild);
+      // Inline enable/disable (CROW-615) — same dirty/save path as Duplicate/Delete.
+      const enable = el('input', 'st-switch st-row-switch');
+      enable.type = 'checkbox';
+      enable.checked = !!job.enabled;
+      enable.title = job.enabled ? 'Disable job' : 'Enable job';
+      enable.setAttribute('aria-label', enable.title);
+      enable.onchange = () => {
+        job.enabled = enable.checked;
+        markDirty();
+        render();
+      };
+      row.querySelector('.st-row-actions').insertBefore(enable, row.querySelector('.st-row-actions').firstChild);
       body.appendChild(row);
     }
     const add = el('button', 'st-add', '+ Add job');


dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- The new toggle is built with `el()` + property assignment (no `innerHTML` with dynamic data); `title`/`aria-label` are fixed strings, so no XSS surface is added (`settings.js:851-861`).
- Persistence rides the existing dirty/save RPC path — no new endpoint, no new input surface.
- The tmux drain change doesn't alter argument handling; `capture-pane` output is only ever parsed locally.

**Concerns:**
- None found.

### Code Quality

**🟡 Yellow — PR scope is undocumented (PR description)**
The diff bundles two unrelated changes: the CROW-615 jobs toggle (settings.js/css) and the CROW-606 tmux stdout-drain fix (`TmuxController.swift`, `crow-tmux.conf`, +34-line regression test) — commit `1b298d9`. The PR body and test plan describe only the toggle, so a reviewer, the squash-merge message, and release notes would all miss a real concurrency fix in the terminal path. Please update the PR description (summary + test plan) to cover the tmux fix, or split it into its own PR. The tmux code itself is good — this is purely a traceability fix.

**🟢 Green — 10s blocking call on the cooperative thread pool (`TmuxController.swift:312`)**
`capturePane` now blocks up to 10s (was 2s), and `TerminalCockpit.replayData` calls it synchronously from the async WebSocket handler (`TerminalWebSocket.swift:103`), so a wedged tmux server pins a Swift-concurrency cooperative-pool thread for 10s per reconnecting client. With the drain fix in place the capture normally completes in milliseconds, so this only bites when tmux is truly hung — but several simultaneous browser reconnects against a wedged server could starve the pool. Consider hopping the capture onto a GCD/blocking-friendly context, or capping replay's timeout lower than the general 10s.

**🟢 Green — two GCD worker threads parked per `run()` call (`TmuxController.swift:64-73`)**
Each invocation now blocks two global-queue threads in `readDataToEndOfFile`. Fine at Crow's tmux call rates, but worth knowing if `run()` ever gets called in wide fan-out — GCD's per-QoS thread pool is finite. A single reader draining stdout while the caller drains stderr inline would halve it if it ever matters.

**Verified strengths:**
- The drain fix is the textbook cure for the wait-then-read pipe deadlock, and the `DispatchGroup.wait()` after `waitUntilExit` gives a proper happens-before edge to the `PipeBox` data (the `@unchecked Sendable` box is sound here).
- Watchdog interplay is correct: on timeout, SIGTERM closes the child's pipe ends, so both readers hit EOF and `group.wait()` can't hang.
- The regression test genuinely reproduces the failure mode (>64 KB with `history-limit` baked into the server conf, sentinel-polled) — reverting the drain fix would fail it via the 2s watchdog.
- Toggle semantics are consistent end-to-end: `!!job.enabled` matches the existing subtitle check (`settings.js:824`), the daemon decodes missing `enabled` as `true` and always re-encodes it (`JobConfig.swift`), and `listRow` binds edit only to the pencil button, so the checkbox doesn't trigger the editor.
- Tests: all 85 CrowTerminal tests pass locally, including the new `capturePaneDrainsLargeStdoutWithoutDeadlock`; `node --check` passes on settings.js.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The lone Yellow is quick: document the bundled CROW-606 tmux fix in the PR description (or split it out), and this is good to go.

---

dgershman:
Thanks for the review — addressed the Yellow and left the Greens as-is:

**🟡 PR scope undocumented** — Updated the PR summary + test plan to cover the bundled CROW-606 tmux stdout-drain fix (`1b298d9`). That commit was on the local `feature/crow-593-web-ui-parity` tip when this branch was cut but not yet on `origin`, so it landed in the PR diff; documenting it here keeps squash-merge / release notes honest without splitting mid-flight.

**🟢 10s blocking capture on the cooperative pool / 🟢 two GCD readers per `run()`** — Agreed these are worth knowing, but leaving them for a follow-up. With the drain fix the capture normally finishes in ms; the longer timeout only matters when tmux is already wedged, and Crow's tmux call rate doesn't currently fan out wide enough for the dual-reader cost to matter. Happy to file a follow-up if you'd rather track the hop-to-GCD / single-reader cleanup explicitly.

Ready for another look.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- The new toggle introduces no new attack surface: it's a plain checkbox built via the `el()` helper (no `innerHTML`), with static `title`/`aria-label` strings, and persists through the existing config save RPC — no new endpoint, no new input parsing.
- `TmuxController.run()` changes don't alter argument handling or shell out differently; the drain threads only read the pipes the process already owned. The watchdog timeout/SIGTERM path is preserved, and on timeout the dead child closes its pipe ends so `group.wait()` cannot hang beyond the old post-exit `readDataToEndOfFile` behavior.
- `crow-tmux.conf` change is comment-only.

**Concerns:**
- None found.

### Code Quality
Verified locally: `swift build` clean, all 85 CrowTerminal tests pass (including the new `capturePaneDrainsLargeStdoutWithoutDeadlock`), and `node --check` passes on `settings.js`.

- **Green** — `settings.js:853`: `!!job.enabled` is safe against missing keys — `JobConfig` decodes absent `enabled` as `true` and always re-encodes it, so the web UI never sees `undefined`; semantics also match the existing `· disabled` subtitle on line 824. No action needed, just confirming the falsy-means-disabled convention holds end to end.
- **Green** — `TmuxController.swift:312`: `max(TmuxController.defaultTimeout, 10.0)` always evaluates to `10.0` today; the `max` only matters if `defaultTimeout` is ever raised above 10s. Harmless and arguably self-documenting — consider a named constant (e.g. `capturePaneTimeout`) if this grows more call sites.
- **Green** — `settings.js:856-860`: the `onchange` handler calls `render()`, which rebuilds the list and drops keyboard focus from the switch the user just toggled. This matches the existing Duplicate/Delete behavior, so it's a pre-existing pattern rather than a regression — worth a focused a11y pass someday.
- **Green** — the regression test is well-constructed: 72 KB output genuinely exceeds the ~64 KB pipe buffer, so the test fails under the old wait-then-read ordering (watchdog fires, `try?` yields empty) and passes with the concurrent drain. Bundling the CROW-606 fix in a CROW-615 PR is unusual but clearly disclosed in the description and targets the parity feature branch.

### Summary Table

---

PR #617, opened 2026-07-08
feat(web): circular checkboxes for sidebar multi-select

### LINKED ISSUE #614: Web UI: circular checkboxes in sidebar multi-select

## Problem
Sidebar multi-select (`selectionMode` / select-all bulk delete) uses square native checkboxes (`.row-check` in `app.js`). Prefer circular checkboxes to match a more polished control language.

## Direction
Style `.row-check` (and any matching allowlist multi-select boxes if they share the pattern) as circular — custom appearance or `border-radius: 50%` with a checked state that still reads clearly as selected. Keep hit target / accessibility (keyboard + click on row still toggles).

## Acceptance
- Multi-select checkboxes in the left pane are circular when selected and unselected.
- Bulk select-all / delete behavior is unchanged.

dgershman:
## Summary
- Style sidebar multi-select (`.row-check`) and allowlist (`.allow-check`) checkboxes as circular controls with a clear gold checked state
- Keep native `<input type="checkbox">` for keyboard/focus accessibility; bulk select-all / delete behavior unchanged

Closes #614

## Test plan
- [ ] Enter sidebar selection mode — checkboxes are circular when unchecked
- [ ] Select rows via checkbox and via row click — checked state shows gold circle with checkmark
- [ ] Section All/Clear and bulk delete still work as before
- [ ] Allowlist multi-select checkboxes are also circular
- [ ] Keyboard focus ring still visible on the checkbox

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,8 @@ name: CI
 
 on:
   pull_request:
-    branches: [main]
+    # Base-branch filter: stacked feature PRs (e.g. onto crow-593) need CI too.
+    branches: [main, 'feature/**']
 
 concurrency:
   group: ci-${{ github.ref }}
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -76,7 +76,39 @@ html, body {
 }
 .divider-selall:hover { color: var(--gold); border-color: var(--gold); }
 .session-row.selecting { display: flex; align-items: flex-start; gap: 8px; }
-.row-check { margin-top: 3px; flex: 0 0 auto; accent-color: var(--gold); cursor: pointer; }
+/* Circular multi-select checkboxes (sidebar + allowlist) — native input kept for
+   keyboard/a11y; appearance stripped so the control reads as a circle (CROW-614). */
+.row-check,
+.allow-check {
+  appearance: none;
+  -webkit-appearance: none;
+  width: 16px;
+  height: 16px;
+  margin: 3px 0 0;
+  flex: 0 0 auto;
+  border-radius: 50%;
+  border: 1.5px solid var(--border-strong);
+  background: transparent;
+  cursor: pointer;
+  vertical-align: top;
+  transition: background 0.12s ease, border-color 0.12s ease;
+}
+.row-check:hover,
+.allow-check:hover { border-color: var(--gold); }
+.row-check:checked,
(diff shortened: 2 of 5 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Tmux invocations continue to pass arguments as an array to `Process` (no shell interpolation), so the new drain logic introduces no injection surface.
- The checkmark SVG is an inline data URI with static content — no external fetch, no user-controlled input.
- `PipeBox`'s `@unchecked Sendable` is safe as used: writes happen on the reader threads and are published to the caller through the `DispatchGroup.wait()` happens-before edge, so there is no data race.

**Concerns:**
- None.

### Code Quality
Verified locally: `swift build` succeeds and all 85 CrowTerminal tests pass, including the new `capturePaneDrainsLargeStdoutWithoutDeadlock` regression test (0.3s).

The core fix is correct: draining stdout/stderr concurrently while waiting on the child removes the classic pipe-buffer deadlock, and the regression test genuinely exercises it (>64 KB of scrollback with a sentinel). The tmux.conf comment about `history-limit` applying only to new windows is accurate and worth having, and the test bakes the limit into the server conf accordingly.

Green (consider, non-blocking):
- `Packages/CrowTerminal/Sources/CrowTerminal/TmuxController.swift:78` — `group.wait()` is unbounded. If a grandchild process ever inherits the pipe write-ends and outlives the tmux client, the drain threads won't see EOF and the call blocks past the watchdog. The pre-fix code had the same exposure at `readDataToEndOfFile`, so this is not a regression — just worth remembering if a hang here ever shows up.
- `Packages/CrowTerminal/Sources/CrowTerminal/TmuxController.swift:312` — the 10s floor applies to *all* `capturePane` calls, including the small 200-line diagnostic capture in `TmuxBackend.swift:1242` that runs when the server is already wedged; that failure path can now stall up to 10s instead of 2s. A `timeout` parameter on `capturePane` (10s only for the CROW-606 replay caller) would scope it more tightly.
- PR scope mixes two concerns — the CROW-614 checkbox styling and the CROW-606 tmux drain fix + conf docs. Both are well-executed and individually reviewed here, but separate PRs would keep revert/bisect granularity cleaner.
- CSS: `appearance: none` checkboxes render invisible in Windows High Contrast (forced-colors) mode; a `@media (forced-colors: active)` fallback restoring borders would cover it. The `.allow-check-spacer` width was correctly bumped 13→16px to match the new control size, and `:focus-visible` keeps keyboard accessibility intact.

### Summary Table

---

PR #618, opened 2026-07-08
feat(web): skeleton-load the left pane before list-sessions

### LINKED ISSUE #613: Web UI: skeleton loading for the left pane on first paint

## Problem
On first load / entering the app, the left pane waits on `list-sessions` (and related RPCs) before painting anything useful — it feels blank until responses return.

## Direction
Paint what we can immediately (chrome, section headers, placeholder/skeleton rows) and fill in real session cards as `list-sessions` / `list-sessions-live` land. Prefer skeleton placeholders over a spinner-only empty state.

## Acceptance
- Opening the app shows a structured left-pane skeleton (or cached last-known layout) before the first RPC round-trip completes.
- Real session rows replace skeletons without a jarring full re-layout when data arrives.

dgershman:
## Summary
- Paint left-pane chrome (brand, tickets card, nav pills) immediately on boot, before the first `/rpc` round-trip.
- Restore last-known sessions + ticket/review badge counts from `localStorage` when available; otherwise show shimmer skeleton rows matching session-card geometry.
- Swap in real session rows once `list-sessions` lands, without flashing a blank/"No sessions" empty state.

Closes #613

## Test plan
- [ ] Cold load (cleared site data): left pane shows brand + tickets chrome + Active section with skeleton rows before sessions appear
- [ ] Warm reload: cached session cards paint immediately, then update when `list-sessions` returns (no blank flash)
- [ ] Empty workspace after load: "No sessions" only after the first successful `list-sessions`, not during the wait
- [ ] Real rows replace skeletons without a jarring full re-layout
- [ ] Ticket counts / Reviews badge restore from cache on warm load

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 Backfill of merged PRs since the 0.1.0 release, grouped by theme.
 
+### Web UI
+
+- #613 — Left pane paints chrome + skeleton (or cached last-known) session rows on first load before `list-sessions` returns, then swaps in real rows without a blank flash.
+
 ### Automation
 
 - #137 — Session analytics emitted via Claude Code's OpenTelemetry exporter (configure via `CLAUDE_CODE_ENABLE_TELEMETRY` and `OTEL_EXPORTER_OTLP_ENDPOINT`).
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -166,6 +166,39 @@ html, body {
 .activity-badge { font-size: 10px; font-weight: 700; text-transform: uppercase; letter-spacing: 0.4px; }
 .empty { color: var(--text-muted); padding: 20px 6px; }
 
+/* ---- Sidebar skeleton (first paint before list-sessions, CROW-613) ---- */
+.session-row.skeleton-row {
+  cursor: default;
+  pointer-events: none;
+  border-left-color: var(--border-subtle);
+}
+.session-row.skeleton-row:hover { border-color: var(--border-subtle); }
+.skel {
+  display: block;
+  border-radius: 4px;
+  background: linear-gradient(
+    90deg,
+    rgba(255, 255, 255, 0.04) 0%,
+    rgba(255, 255, 255, 0.09) 45%,
+    rgba(255, 255, 255, 0.04) 90%
+  );
+  background-size: 200% 100%;
+  animation: crow-skel-shimmer 1.35s ease-in-out infinite;
+  animation-delay: var(--skel-delay, 0s);
+}
+.skel-agent { width: 12px; height: 12px; flex: 0 0 auto; border-radius: 3px; }
+.skel-name { flex: 1 1 auto; height: 12px; max-width: 58%; }
+.skel-dot { width: 8px; height: 8px; border-radius: 50%; flex: 0 0 auto; margin-left: auto; }
+.skel-subtle { height: 10px; width: 72%; margin-top: 8px; }
+.skel-meta { height: 10px; width: 48%; margin-top: 6px; }
+@keyframes crow-skel-shimmer {
+  0% { background-position: 100% 0; }
+  100% { background-position: -100% 0; }
+}
+@media (prefers-reduced-motion: reduce) {
+  .skel { animation: none; background: rgba(255, 255, 255, 0.06); }
+}
+
 /* ---- Detail ---- */
 #detail {
   display: flex;
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -462,6 +462,13 @@ const boardData = { tickets: null, reviews: null, allowlist: null };
 // Last-known sidebar layout (sessions + ticket/review badge counts) so first
 // paint isn't blank while /rpc connects (CROW-613).
 const SIDEBAR_CACHE_KEY = 'crow.sidebar.cache';
+// True when boot restored a cache entry (including an empty sessions list) —
+// distinguishes "remembered empty" from a cold start that should show skeletons.
+let sidebarCacheHit = false;
+function clearSidebarCache() {
+  try { localStorage.removeItem(SIDEBAR_CACHE_KEY); } catch (_) {}
+  sidebarCacheHit = false;
+}
 function restoreSidebarCache() {
   try {
     const raw = localStorage.getItem(SIDEBAR_CACHE_KEY);
@@ -470,6 +477,7 @@ function restoreSidebarCache() {
     if (Array.isArray(data.sessions)) sessions = data.sessions;
     if (data.tickets) boardData.tickets = data.tickets;
     if (data.reviews) boardData.reviews = data.reviews;
+    sidebarCacheHit = true;
   } catch (_) { /* corrupt cache — start empty */ }
 }
 function persistSidebarCache() {
@@ -479,6 +487,7 @@ function persistSidebarCache() {
       tickets: boardData.tickets,
       reviews: boardData.reviews,
     }));
+    sidebarCacheHit = true;
   } catch (_) { /* quota / private mode */ }
 }
 let ticketFilter = 'All'; // pipeline segment ('All' or a status rawValue); default to All so the board isn't misread as empty when work moves to Done
@@ -559,7 +568,7 @@ function el(tag, className, text) {
 let lastSidebarSig = null;
 function sidebarSignature() {
   return JSON.stringify([
-    sessionsLoaded, sessions, liveById, selectedId, selectedBoard,
+    sessionsLoaded, sidebarCacheHit, sessions, liveById, selectedId, selectedBoard,
     selectionMode, [...selectedSessionIDs],
     uiConfig.hideSessionDetails,
     boardData.tickets && boardData.tickets.counts,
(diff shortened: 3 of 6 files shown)


dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues

None.

### Security Review

**Strengths:**
- All cached data is rendered through the `el()` helper (`app.js:548`), which uses `textContent` — no XSS path from `localStorage`-sourced session names or ticket titles.
- Both cache functions are wrapped in `try`/`catch`, so a corrupt cache or a quota/private-mode failure degrades gracefully instead of breaking boot.
- Sound/notification detection is not fooled by the cache: `_prevSessionSnap` / `_prevReviewIDs` start `null` and the restore path never calls the detect functions, so the first real refresh only baselines — no spurious chimes from state that changed while the page was closed.
- Swift side: `tmux` is invoked with an argument array (no shell), the `PipeBox` writes are properly ordered before the reads via the `DispatchGroup` (`TmuxController.swift:60-78`), and watchdog/timeout semantics are preserved.

**Concerns:**
- 🟡 **Logout does not clear the sidebar cache** (`app.js:805-809` vs. new `persistSidebarCache`, `app.js:475`). The cache stores session names plus the *full* tickets and reviews board payloads (ticket titles/URLs, PR authors, etc.) under `crow.sidebar.cache`. After an explicit logout of a password-protected remote web session, that data remains readable in `localStorage` by anyone with access to the browser — exactly the scenario the web password / logout flow exists to protect. Fix is one line in the logout handler before `location.reload()`:
  ```js
  try { localStorage.removeItem(SIDEBAR_CACHE_KEY); } catch (_) {}
  ```

### Code Quality

- 🟢 `sidebarSignature()` (`app.js:558`) still omits `uiConfig.hideSessionDetails`. This gap pre-exists, but the new pre-config first paint widens it: cached rows (and skeletons, `app.js:623`) render with the default `hideSessionDetails: false`, and when `loadUIConfig()` later calls `renderSidebar()`, the unchanged signature short-circuits the repaint until unrelated state changes. Consider including `uiConfig.hideSessionDetails` in the signature.
- 🟢 `persistSidebarCache()` is called on every successful `refreshSessions` poll (`app.js:525`, every 10s) even when nothing changed, while `refreshBoard` correctly gates its persist on `changed` (`app.js:1654`). Consider gating the sessions persist the same way to avoid redundant `JSON.stringify` + `localStorage` writes.
- 🟢 `restoreSidebarCache` (`app.js:462`) validates only `Array.isArray(data.sessions)`, not element shape, and the boot-time `renderSidebar()` (`app.js:2383`) runs at top level. A stale-schema cache entry that throws during render would abort the remaining boot statements (`refreshSessions`, the polling `setInterval`s never start) until the user clears site data. Consider wrapping the boot paint in a `try`/`catch` that drops the cache and re-renders empty.
- 🟢 `group.wait()` (`TmuxController.swift:78`) is unbounded: if a grandchild ever inherits the pipe's write end and outlives the SIGTERM'd tmux client, `readDataToEndOfFile` never sees EOF and `run()` blocks despite the watchdog. The same exposure existed before this PR (the old post-exit read had it too), so this is not a regression — but `group.wait(timeout:)` with a small grace period would make the watchdog a true wall-clock bound.

**What's good:**
- The CROW-606 drain fix is correct and the regression test is excellent — it reproduces the >64 KB pipe-buffer deadlock with a real tmux server, bakes `history-limit` into a per-test conf (correctly noting the option is frozen at window birth), and asserts both sentinel and size. All 85 CrowTerminal tests pass locally, including the new one.
- The skeleton CSS respects `prefers-reduced-motion`, skeleton rows are `aria-hidden` with `pointer-events: none`, and the "No sessions" empty state is correctly gated on `sessionsLoaded` so it can't flash during the first RPC.
- The tmux.conf comment documenting that `history-limit` doesn't resize existing windows is a genuinely useful caveat for future readers.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 4 Green] findings. The Yellow (clear `crow.sidebar.cache` on logout) is a one-line fix; everything else is take-it-or-leave-it.

---

dgershman:
Addressed in the latest push (`d142ddc`):

- **Yellow — clear cache on logout:** `localStorage.removeItem(SIDEBAR_CACHE_KEY)` before reload.
- **Green — `hideSessionDetails` in signature:** included so `loadUIConfig()` can repaint after the pre-config first paint.
- **Green — gate session persist on change:** matches the `refreshBoard` pattern.
- **Green — boot paint try/catch:** drops a bad cache and continues so polls still start.

**Green — `group.wait()` timeout (TmuxController):** leaving as-is. That path came in with the CROW-606 drain fix on the parity base and isn’t part of this skeleton-loading change; happy to track a follow-up if you want a wall-clock bound there.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- All sidebar DOM built from cached `localStorage` data goes through `el()` which uses `textContent` (app.js:550-555) — a tampered or stale cache entry cannot inject HTML/script. Style values derived from cached data (`ind.color`, `prBadgeColor`) come from fixed lookup tables, not raw strings.
- Logout explicitly removes the sidebar cache (`localStorage.removeItem(SIDEBAR_CACHE_KEY)`, app.js:813) so session names/ticket titles don't linger in a shared browser after a password-protected remote session ends.
- The boot-time restore is wrapped in try/catch that clears a corrupt/stale-schema cache and still proceeds to the RPC refresh (app.js:2389-2397) — a poisoned cache can't brick boot.

**Concerns (Green — consider):**
- The cache is cleared only on *explicit* logout. If the cookie dies server-side (crowd restart → "Session expired"), cached session/ticket payloads remain in `localStorage` on the remote browser. It's the same user's browser and the page itself is auth-gated, so this is data-at-rest only — worth clearing when `sessionDead` flips, if you want parity with logout.

### Code Quality
- `TmuxController.run()` drain fix (TmuxController.swift:56-75) is correct: stdout/stderr are read concurrently while waiting, so a >64 KB `capture-pane` can no longer fill the pipe buffer and wedge the child into the watchdog. `PipeBox` + `DispatchGroup` gives proper happens-before for the cross-thread `Data` handoff.
- Green: `group.wait()` after `waitUntilExit` is unbounded. If a descendant ever inherited and held the pipe write-end past client exit, `run()` would hang with no watchdog. tmux daemonizes and redirects its fds so this doesn't bite in practice (and the old post-exit blocking read had the same exposure) — a `group.wait(timeout:)` fallback would make it airtight.
- Green: the 10s `capturePane` budget blocks the calling task synchronously; it runs in the per-connection WebSocket handler (TerminalWebSocket.swift:103), not the main thread, and only stalls when tmux itself does. Fine as-is.
- Green: with a cached "empty workspace" (`sessions: []`), boot shows shimmer skeletons rather than the remembered "No sessions" state until the first RPC lands. Cosmetic.
- Nice touches: `sessionsLoaded` + `uiConfig.hideSessionDetails` added to `sidebarSignature()` so the skeleton→real swap and details-visibility change actually invalidate the render cache; `lastSidebarSig = null` reset before the recovery re-render after a mid-render throw; `prefers-reduced-motion` handled in the shimmer CSS; skeletons are `aria-hidden` and `pointer-events: none`.
- Tests: `capturePaneDrainsLargeStdoutWithoutDeadlock` reproduces the exact failure (>64 KB capture) and passes in 0.3s. Full suite: 85/85 pass. `swift build` clean; `app.js` syntax-checked.

### Summary Table

---

dgershman:
Follow-up on the green notes from the latest review (`d142ddc` was already APPROVED):

- **Clear cache when `sessionDead` flips:** `clearSidebarCache()` now runs in `handleAuthOnDisconnect` when the remote cookie dies, matching explicit logout.
- **Cached empty workspace:** a restored `sessions: []` cache keeps "No sessions" instead of shimmering until the first RPC.
- **Tmux `group.wait()`:** still leaving as-is (CROW-606 base, not this PR).

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Client-side `localStorage` sidebar cache (session names, ticket/review counts) is cleared on **both** exit paths: explicit logout (`renderStatusBar` handler) and detected cookie-death on disconnect (`handleAuthOnDisconnect` → `clearSidebarCache`). Good defense-in-depth for shared/remote password-protected deployments.
- Unauthenticated users are served the auth-gate login page, not `app.js`, so cached payloads are never painted pre-auth. Combined with Crow's single-shared-password model (one principal), there's no meaningful cross-user exposure introduced.
- Boot restore is wrapped in try/catch that clears a corrupt/stale-schema cache and keeps booting (`refreshSessions` repaints), so a bad cache entry can't wedge first paint.

**Concerns:**
- None blocking.

### Code Quality
- **app.js (CROW-613):** Skeleton gating (`!sessionsLoaded && !sessions.length && !sidebarCacheHit`) correctly separates cold-start shimmer from a remembered-empty "No sessions" state. `uiConfig` has safe defaults defined well before the boot-time `renderSidebar()`, so the `hideSessionDetails` read in `skeletonRow` is never undefined; `loadUIConfig()` re-renders afterward. `sidebarSignature` was correctly extended with the new state flags + `hideSessionDetails` so renders aren't wrongly skipped. Reduced-motion is honored in CSS.
- **TmuxController.swift (CROW-606):** Textbook fix — both pipes are drained on background threads *before* `waitUntilExit()`, eliminating the >64KB pipe-buffer deadlock that silently no-op'd web-terminal replay. On timeout the watchdog's `terminate()` closes the pipes so the readers return and `group.wait()` can't hang. `DispatchGroup` provides the happens-before barrier for the `PipeBox` reads, making `@unchecked Sendable` sound.
- **Tests:** New regression test emits ~72KB of scrollback (above the ~64KB buffer) and asserts full capture without timeout. Verified locally: builds clean and the full `TmuxControllerTests` suite (8 tests) passes.
- Minor (non-blocking): this PR bundles the CROW-606 tmux fix alongside the CROW-613 web feature — reads like branch stacking rather than scope creep; both are cohesive and well-tested.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 4 Green] findings. Clean build, all tests pass, thoughtful security handling of the client-side cache.

PR #622, opened 2026-07-08
Quiet permanent enableAutoMerge failures; surface gh stderr

### LINKED ISSUE #621: Quiet enableAutoMerge failures: log gh stderr, don't spam permanent repo errors

## Problem
crowd logs a useless, scary line on every auto-merge attempt:

```
[Crow] enableAutoMerge failed for https://github.com/radiusmethod/shell-crm/pull/299: The operation couldn’t be completed. (CrowCore.ShellRunnerError error 0.)
```

Reproduced with the same `gh` call Crow makes:

```
$ gh pr merge <url> --auto --squash --delete-branch
GraphQL: Auto merge is not allowed for this repository (enablePullRequestAutoMerge)
```

So for repos without GitHub **Allow auto-merge** enabled, this is a **permanent** condition — not a transient failure. Today:

1. `ShellRunnerError` has no `LocalizedError` / useful `localizedDescription`, so the log shows `error 0` instead of the `gh` stderr.
2. On failure, `attemptEnableAutoMerge` clears `autoMergeInFlight` and never sets `autoMergeEnabledAt`, so the next poll retries and re-logs forever for every `crow:merge`-labeled PR on that repo.

## Direction
- Make `ShellRunnerError` surface `exitCode` + `output` in `localizedDescription` (or log `output` explicitly in the catch).
- Classify permanent / expected `gh` failures (at least `Auto merge is not allowed for this repository` / `enablePullRequestAutoMerge`) as **non-retryable**: log once at info/debug (or a single warning), and stop re-attempting for that PR/session (e.g. set a sentinel / leave in-flight / persist a "skipped" marker — pick the lightest approach that matches existing auto-merge state).
- Keep real transient failures (network, auth blip) retryable with a useful message.

## Acceptance
- A repo without auto-merge enabled does not spam `[Crow] enableAutoMerge failed … ShellRunnerError error 0` every poll.
- When logging is warranted, the message includes the actual `gh` stderr (e.g. `enablePullRequestAutoMerge`).
- Transient failures still retry.

## Notes
Seen on shell-crm PR #299 (`crow:merge` present, `mergeStateStatus: BLOCKED`, `autoMergeRequest: null`).

dgershman:
## Summary
- Make `ShellRunnerError` conform to `LocalizedError` so logs include exit code + `gh` stderr instead of opaque `error 0`.
- Treat repo-policy auto-merge denials (`enablePullRequestAutoMerge` / “Auto merge is not allowed”) as permanent: log once and leave the PR in `autoMergeInFlight` so polls stop retrying.
- Keep transient `gh` failures retryable with a useful message.

Closes #621

## Test plan
- [x] `swift test` CrowCore `ShellRunnerErrorTests`
- [x] `swift test` CrowEngine `IssueTrackerAutoMergeTests` (permanent vs transient classification)
- [ ] With a `crow:merge`-labeled PR on a repo that has Allow auto-merge disabled: confirm one skip log with real stderr, no per-poll spam
- [ ] Transient failure (e.g. network) still retries on the next poll


dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- `enableAutoMerge` invokes `gh` with a direct argv (no `sh -c`), so surfacing stderr adds no shell-interpolation surface around `prURL` (Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubCodeBackend.swift:266).
- The new error text is only written to the local system log via `NSLog`; `gh` GraphQL errors don't carry credentials, so no secrets exposure.
- The classification is read-only string matching — no new state mutation or privilege path.

**Concerns:**
- None.

### Code Quality

**Yellow — `isPermanentAutoMergeFailure` over-matches on the bare mutation name** (Packages/CrowEngine/Sources/CrowEngine/IssueTracker.swift:1789)

`gh` embeds the GraphQL mutation name in *every* error from that mutation, not just the repo-policy denial. The well-known example is `GraphQL: Pull request is in clean status (enablePullRequestAutoMerge)` — GitHub's response when auto-merge is requested on a PR whose required checks have **already** passed (e.g. the `crow:merge` label lands after CI settles). That state is not permanent, but this classifier freezes the PR in `autoMergeInFlight` for the process lifetime and logs the misleading "auto-merge not allowed on repo" message. It's also a behavioral regression for that case: previously the marker was cleared, so a later poll (after a new push, review, etc.) could retry and succeed; now it never will until app restart. The test `permanentFailureMatchesCapabilityNameAlone` deliberately codifies this breadth, so it's a design choice worth revisiting rather than an accident. Suggested fix: drop the bare-field-name match and key on the policy phrase ("auto merge is not allowed for this repository") — or, if you keep the field-name match as a phrasing-drift hedge, explicitly classify "clean status" as transient and make the skip log wording cause-neutral.

**Green (consider):**
- The updated doc comment on `autoMergeInFlight` (IssueTracker.swift:123-126) says it is "Cleared on *transient* failure", but a transient error inside `prHasCrowAuthoredCommit` (network failure fetching commits, IssueTracker.swift:1949-1953) returns `false` and takes the missing-authorship early return, which leaves the marker set permanently. Pre-existing behavior, but the new comment now overstates the guarantee — worth either clearing the marker on fetch *errors* (vs. genuine no-trailer results) or softening the comment.
- On a permanent skip the only signal is one NSLog line; since the user's action item is a repo setting ("Allow auto-merge"), consider surfacing it via the existing banner-notification path like `onAutoMergeEnabled` does, so it's discoverable without log spelunking.

**Strengths:** the `LocalizedError` conformance is exactly the right fix for the opaque `error 0` logs, empty-output trimming is handled, tests cover both classification directions plus the non-`ShellRunnerError` path, and docs/automation.md was updated to match. `swift test` passes: CrowCore `ShellRunnerErrorTests` (3/3) and CrowEngine `IssueTrackerAutoMergeTests` (36/36 in file).

### Summary Table

---

dgershman (commented):
Addressed the yellow finding:

- Permanent classification now keys only on `Auto merge is not allowed for this repository` — dropped the bare `enablePullRequestAutoMerge` match so retryable cases like `Pull request is in clean status (enablePullRequestAutoMerge)` stay retryable.
- Softened the `autoMergeInFlight` doc comment to note the pre-existing authorship-fetch transient path.

Left the two greens as-is for now: banner notification for permanent skips is a nice follow-up; authorship-fetch error vs no-trailer distinction is pre-existing and out of scope for this quieting fix.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- `enableAutoMerge` still invokes `gh` via direct argv (no `sh -c`), so surfacing stderr adds no shell-interpolation surface around `prURL` (`GitHubCodeBackend.swift:266-275`).
- The permanent-failure classification only *reduces* retries; it cannot broaden auto-merge eligibility. Authorship gating (`Crow-Session` trailer vs. known sessions) is untouched.
- Classification keys on the specific policy phrase "Auto merge is not allowed for this repository" rather than the GraphQL mutation name, which `gh` embeds in *every* error from that mutation — the tests (`cleanStatusWithMutationNameIsRetryable`, `bareMutationNameAloneIsNotPermanent`) pin this down well.

**Concerns:**
- Green: `ShellRunnerError.errorDescription` now surfaces full merged subprocess output to every caller that logs `localizedDescription`, not just the auto-merge path. `gh` doesn't emit credentials in errors, but if ShellRunner ever runs commands whose stderr can carry secrets, those would now land in the system log. Worth keeping in mind; no action needed today.

### Code Quality
- Fail-safe design: matching on `gh`'s error text is inherently brittle, but the failure mode is benign — if GitHub rewords the message, behavior degrades to the previous retry-and-log-every-poll, never to a stuck or wrongly-merged PR (`IssueTracker.swift:1787-1795`).
- The classifier handles both the raw `ShellRunnerError.nonZeroExit` thrown by `GitHubCodeBackend` and wrapped errors via `localizedDescription` fallback, and the new `LocalizedError` conformance makes that fallback actually carry the stderr text — the two changes compose correctly.
- The updated `autoMergeInFlight` doc comment honestly flags the pre-existing gap where a transient `fetchCrowAuthoredCommits` failure also freezes the marker (`IssueTracker.swift:124-127`) — good candidate for a follow-up ticket, but out of scope here.
- Green: `localizedCaseInsensitiveContains` is locale-sensitive; `range(of:options:.caseInsensitive)` with a nil locale would be fully deterministic. In practice `gh` emits the exact needle casing, so this cannot misfire for this phrase.
- Tests are thorough: permanent vs. transient classification (network, auth, clean-status, bare mutation name), empty-output formatting, and a regression test for the opaque `error 0` description. Verified locally: CrowCore 348/348 and CrowEngine 262/262 tests pass.
- `docs/automation.md` updated to document the one-shot skip semantics.

### Summary Table

---

PR #623, opened 2026-07-08
fix(review): preserve gateway secrets on update; PATH-safe crowd re-exec (PR #594)

dgershman:
## Summary
Addresses round-5 Yellows on #594:

- **Yellow 1** — Updating an AI gateway no longer wipes stored auth-header secrets. Blank header values from the local editor (stripped transport) merge with the currently-stored gateway under `withConfigLock`.
- **Yellow 2** — `CrowDaemon.reexec` resolves the real executable path (`_NSGetExecutablePath` / `/proc/self/exe`) and falls back to `execvp`, so a PATH-launched `crowd` restarts after the first-run wizard instead of `exit(1)`.

Greens left as-is (non-blocking).

## Test plan
- [x] `swift test --package-path Packages/CrowDaemon --filter 'SecretGatewayValidationTests|ReexecPathResolutionTests'` (8/8)
- [ ] Local: edit only a gateway base URL → stored API key still present after save
- [ ] Local: `crowd` from PATH → complete first-run wizard → daemon comes back online without manual relaunch

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Artifacts.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Artifacts.swift
@@ -47,7 +47,15 @@ enum Artifacts {
                   isSafeImageName(file) else {
                 return Response(status: .badRequest)
             }
-            guard let data = try? Data(contentsOf: dir.appendingPathComponent(file)) else {
+            // Resolve symlinks and require the final path stay inside `dir` —
+            // otherwise a planted `shot.png` → `~/.claude/config.json` symlink
+            // would leak secrets that strippedForTransport keeps off every other
+            // transport (review Yellow / CROW-593).
+            let candidate = dir.appendingPathComponent(file)
+            guard let resolved = resolvedPathInside(dir: dir, file: candidate) else {
+                return Response(status: .notFound)
+            }
+            guard let data = try? Data(contentsOf: resolved) else {
                 return Response(status: .notFound)
             }
             var headers: HTTPFields = [.contentType: contentType(for: file), .cacheControl: "no-store"]
@@ -69,6 +77,21 @@ enum Artifacts {
     static func isSafeImageName(_ name: String) -> Bool {
         !name.isEmpty && !name.contains("/") && !name.contains("..")
             && imageExtensions.contains((name as NSString).pathExtension.lowercased())
+    }
+
+    /// Resolve `file` (following symlinks) and return it only when the final
+    /// path is still under `dir`. Rejects dangling / escaping symlinks.
+    static func resolvedPathInside(dir: URL, file: URL) -> URL? {
+        let root = dir.resolvingSymlinksInPath().standardizedFileURL.path
+        let resolved = file.resolvingSymlinksInPath().standardizedFileURL.path
+        // Exact match (file is the dir itself — shouldn't happen) or a child.
+        guard resolved == root || resolved.hasPrefix(root + "/") else { return nil }
+        // The resolved path must still exist as a regular file (not a dir).
+        var isDir: ObjCBool = false
+        guard FileManager.default.fileExists(atPath: resolved, isDirectory: &isDir), !isDir.boolValue else {
+            return nil
+        }
+        return URL(fileURLWithPath: resolved)
     }
 
     private static func contentType(for file: String) -> String {
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
@@ -1,5 +1,6 @@
 #if canImport(Darwin)
 import Darwin
+import MachO
 #elseif canImport(Glibc)
 import Glibc
 #endif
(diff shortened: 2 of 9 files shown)


dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues

**🔴 Red — `mergingPreservedHeaders` can persist an invariant-violating gateway that makes `config.json` undecodable, and the next write then wipes the entire config.**

`Packages/CrowDaemon/Sources/CrowDaemon/SecretRoutes.swift:173` — the final `headers.filter { !$0.value...isEmpty }` can leave `customHeaders` empty while `baseURL` is non-empty. `buildGateway` deliberately lets blank-*valued* headers through (both-or-neither check passes), so this shape is reachable whenever a blank value has no stored counterpart:

- No stored gateway + body `{baseURL: "https://gw", headers: {"X-Api-Key": ""}}` (a user typing `X-Api-Key:` with no value in the editor — `parseHeaderLines` produces `""`).
- Stored gateway exists, but the user renames the only header key (new key has a blank value, stored has no entry for it).

The result `WorkspaceGateway(baseURL: "https://gw", customHeaders: [:])` encodes fine, but `WorkspaceGateway.init(from:)` (`AppConfig.swift:286`) throws on decode. `ConfigStore.loadConfig` then returns `nil` for the **whole** config, and every write handler does `loadConfig(devRoot:) ?? AppConfig()` — so the next config write silently resets everything: workspaces, Jira credential, web password. I reproduced this with a scratch test: `buildGateway` → `mergingPreservedHeaders(incoming:, stored: nil)` → encode → decode fails.

**Fix:** after the final filter, if `headers.isEmpty` while `baseURL` is non-empty, reject the request (400, "header X has no value and no stored secret to keep") rather than persisting. Add a regression test for the all-blank/no-stored case. (Note: pre-PR behavior persisted the blank *value*, which still decoded — the corrupting shape is new in this PR.)

### Security Review

**Strengths:**
- The merge runs entirely under `withConfigLock` and only inside the already local-only + Origin-gated POST handlers — no change to the trust boundary; secrets still never travel to remote clients.
- Non-blank incoming values correctly override stored ones, and `nil` incoming still clears — the `strippedForTransport` round-trip contract is honored.
- `resolvedExecutablePath` asks the kernel (`_NSGetExecutablePath` / `/proc/self/exe`) before trusting `argv[0]`, which *reduces* the attack surface of re-exec versus the old `execv(argv[0])`; `execvp` is only used for a bare (path-less) argv[0], matching how the process was originally found.

**Concerns:**
- None beyond the Red above.

### Code Quality

- 🟢 Green — `CrowDaemon.swift:329-338`: if `execv(resolved)` fails (e.g. binary replaced/unlinked during an upgrade), consider falling through to the `argv0`/`execvp` attempts instead of exiting immediately — the branches are alternatives but could be a chain.
- 🟢 Green — `CrowDaemon.swift:357-362` (Linux): a `readlink` result of exactly `PATH_MAX-1` may be silently truncated; comparing `n < buf.count - 1` would detect that and fall back to `execvp`.
- 🟢 Green — `blankHeaderWithNoStoredValueIsDropped` pins the exact dropping behavior that creates the Red; when fixing, this test should assert rejection (or invariant preservation) instead.
- Tests: the four merge tests and the re-exec path test are good targeted coverage; build passes and all 8 pass locally (`swift test --filter 'SecretGatewayValidationTests|ReexecPathResolutionTests'`).

### Summary Table

**Recommendation:** Request Changes — driven by 1 Red, 0 Yellow, 3 Green findings.

---

dgershman:
Addressed the Red: `mergingPreservedHeaders` now returns `Result` and rejects (400) when a non-empty `baseURL` would be persisted with no remaining headers after blank-drop / restore. Added regression tests for all-blank (no stored) and renamed-blank-key cases. Greens left as-is.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Secret preservation is done correctly: `mergingPreservedHeaders` runs **inside** `ConfigStore.withConfigLock` (SecretRoutes.swift:72, :106), so the read-merge-write is atomic against concurrent config writers, and the restored secrets never leave the server — the HTTP responses still return only booleans (`saved`/`gateway_set`), never the merged gateway.
- The merge contract matches the transport-stripping contract end to end. `SettingsSecrets.strippedForTransport` blanks header *values* but keeps keys; the web editor (`settings.js` `gatewayEditor`/`parseHeaderLines`) round-trips those keys with empty values; the daemon now restores the stored secret for any blank-valued key. Verified the JS side actually keeps blank values (`out[k] = line.slice(i + 1).trim()`), so the fix is reachable from the real client.
- The new rejection path (URL + no keepable headers → 400 `GatewayValidationError`) closes a real config-wipe hazard: that shape encodes but fails `WorkspaceGateway.init(from:)`'s both-or-neither decode invariant (AppConfig.swift:286), which would make the next `loadConfig` return `nil` and a subsequent save start from a blank `AppConfig()`. Good catch folding the prior round's Red into this branch, with tests for both the all-blank and renamed-key variants.
- Error messages don't leak secret values; local-only + Origin gating (`gateOK`) on the secret routes is unchanged.
- Re-exec hardening is sound: `_NSGetExecutablePath` + `realpath` (Darwin) / `/proc/self/exe` (Linux) prefers the kernel's view of the binary over `argv[0]`, and the `execvp` PATH-search fallback is only reached when kernel resolution fails and `argv[0]` has no slash — same user, same inherited `PATH`, no privilege boundary crossed.

**Concerns:**
- None blocking.

### Code Quality
- `withConfigLock` is `rethrows` with a `defer`red unlock, so the new throwing merge inside the closure can't leak the lock; `GatewayValidationError` is caught and mapped to 400 rather than a misleading 500 in both routes.
- The workspace route's not-found path now returns before any save, and `gateway_set` semantics are preserved (`merged != nil` ⇔ incoming non-nil).
- Test coverage is solid: 6 merge tests covering preserve/override/drop/reject/clear plus the re-exec path-resolution test. Verified locally — targeted suites 10/10, full CrowDaemon suite **99/99 passing**, package builds clean.

Green (non-blocking) observations:
- `reexec` attempts exactly one strategy — if `execv(resolved)` fails (e.g., binary replaced-by-rename since launch), it exits rather than falling through to `argv0`/`execvp`. A chained fallback would be marginally more resilient (CrowDaemon.swift:330).
- `mergingPreservedHeaders` rejects the URL-without-headers undecodable shape but not the mirror (headers-without-URL). It's unreachable today because `buildGateway` enforces both-or-neither first, but a symmetric guard would future-proof the exposed-for-tests helper (SecretRoutes.swift:189).
- Linux `readlink` into a `PATH_MAX` buffer silently truncates ultra-long exe paths; vanishingly rare, and the failure mode is a logged exec error.

### Summary Table

---

dgershman:
Round-7 Yellows from #594 addressed on this branch:

1. **Artifacts symlink disclosure** — resolve final path and require it stay under the scratch dir (+ regression test).
2. **`run-setup` local-direct** — `/rpc` WS rejects non-local callers before the handler runs.
3. **`/terminal` reconnect** — skip reconnect when `sessionDead`.
4. **Allowlist Clear** — refresh `disabled` on checkbox toggle.
5. **Escape in settings sub-form** — Escape clears the stacked overlay first, then Settings.

Greens left as-is (symlink resolve on `isPathWithinRoot`, stale get-config comments).

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- **Gateway secret preservation** (`SecretRoutes.mergingPreservedHeaders`) correctly implements the blank-value-means-keep contract that `strippedForTransport` relies on: stripped keys arrive blank, get restored from the stored gateway under `withConfigLock`, and truly-empty keys are dropped. Verified the lock is released on throw (`ConfigStore.withConfigLock` uses `defer { unlock() }`), so the new `GatewayValidationError` path can't deadlock.
- **Config-corruption guard (round-6 Red)** is closed: a URL with no keepable headers is rejected before persistence. Confirmed this matches `WorkspaceGateway.init(from:)`'s both-or-neither invariant (`AppConfig.swift:286`) — persisting that shape would fail the next decode and `loadConfig` would silently wipe the config. Well covered by `allBlankHeadersWithNoStoredSecretAreRejected` and `renamedBlankHeaderWithNoStoredMatchIsRejected`.
- **Artifact symlink escape** (`Artifacts.resolvedPathInside`) canonicalizes via `resolvingSymlinksInPath` and enforces containment before `Data(contentsOf:)`, blocking a planted `shot.png → ~/.claude/config.json` symlink. Reads the fully-resolved real path, so the residual TOCTOU window is negligible for a loopback artifact server.
- **run-setup gating**: the `/rpc` WebSocket handler now captures local-direct at upgrade and rejects a non-local `run-setup` (a write+re-exec) — sound defense-in-depth beyond Origin/auth.

**Concerns:** None.

### Code Quality
- `CrowDaemon.resolvedExecutablePath` handles Darwin (`_NSGetExecutablePath`+`realpath`), Linux (`/proc/self/exe`), and an `execvp` fallback so a bare `crowd` on PATH restarts instead of `exit(1)`. Only nit: two `String(cString:)` deprecation warnings at CrowDaemon.swift:355-356 — non-blocking.
- Web fixes are correct: Clear-button enable-state refresh on first tick, `sessionDead` guards preventing terminal reconnect loops after an expired cookie, and Escape backing out the topmost overlay first to match backdrop-click.
- Strong test coverage: build clean, 61/61 tests pass across the touched suites (`SecretGatewayValidationTests`, `DaemonSecurityTests`, `ArtifactsTests`, `ReexecPathResolutionTests`).

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 1 Green] findings. The lone Green (Darwin `String(cString:)` deprecation warnings) is non-blocking.

---

PR #625, opened 2026-07-09
fix(review): gate set-config binaries/jobs to local-direct (#594)

dgershman:
## Summary
- Extends the `/rpc` local-direct gate (already used for `run-setup`) so `set-config` changes to `defaults.binaries` or `jobs` are rejected for non-local peers.
- Harmless `set-config` fields (and unchanged binaries/jobs round-trips) still work remotely.
- Adds `LocalOnlyRPCGateTests`.

Addresses the remaining Yellow on #594 after #623 merged (`run-setup` was already fixed there; this review still flagged it because it ran against the pre-merge tip).

## Test plan
- [x] `swift test --filter LocalOnlyRPCGateTests`
- [ ] Confirm local Settings can still edit Jobs / binary overrides
- [ ] Confirm a non-loopback `/rpc` client gets `set-config binaries/jobs is local-only` when changing those fields

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
@@ -712,6 +712,10 @@ func makeCommandRouter(
                 "default_dev_root": .string(defaultDevRoot),
             ]
         },
+        // Non-secret settings write. `defaults.binaries` and `jobs` are held to
+        // the same local-direct bar as secret writes — the `/rpc` WebSocket
+        // handler rejects those field changes from non-local peers before this
+        // runs (review Yellow). Unix-socket / CLI callers are always local.
         "set-config": { params in
             guard let json = params["config"]?.stringValue,
                   let data = json.data(using: .utf8),
@@ -788,7 +792,9 @@ func makeCommandRouter(
         // reachable over the possibly-remote `/rpc` WebSocket and can't tell a
         // local caller from a logged-in remote one, so it must not carry secret
         // writes — that would let a remote client change the password that gates
-        // remote access (CROW-593).
+        // remote access (CROW-593). The same local-direct bar applies to
+        // `set-config` changes of `defaults.binaries` / `jobs` (gated in
+        // `RPCWebSocketHandler`) — those execute at the next launch.
 
         // Session/board actions — forward-only (need the app's coordinators).
         // Spawning a Manager forwards to the app when it's running (its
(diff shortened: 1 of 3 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Correctly extends the existing local-direct gate to the two real code-execution vectors reachable via a non-local `/rpc` peer:
  - `defaults.binaries` — executed by `Scaffolder` (external-tool installers on launch) and by agent binary discovery (`CodingAgent.findBinary()`), i.e. persistent RCE.
  - `jobs` — scheduled agent prompts that fire unattended.
- The gate lives in `RPCWebSocketHandler` (the only path that can carry a remote caller); Unix-socket/CLI callers remain always-local, matching the `SecretRoutes` model. Good separation of concerns and clearly documented in both `RPCHandlers.swift` and `RPCWebSocketHandler.swift`.
- **No adjacent vector left open:** the AI-gateway `baseURL` (a plausible `ANTHROPIC_BASE_URL` redirect) is *not* remotely writable via `set-config` — `SettingsSecrets.preservingSecrets` overwrites the entire gateway block from stored config on every write. `cli`/`provider` are host-tool metadata, not launch paths.
- Change-based comparison is the right model: injecting a *new* malicious binary/job is always a change → denied; unchanged round-trips (a remote editor re-saving what it read) pass, so harmless remote edits keep working. `RPCHandlers.swift:715` correctly documents this.

**Concerns:**
- None blocking. Gate and handler both decode `AppConfig` via plain `JSONDecoder`, and `preservingSecrets` never touches `binaries`/`jobs`, so the gate's `incoming != current` check faithfully reflects what the handler would persist. Malformed payloads return `false` and fall through to the handler's `invalidParams` with no write.

### Code Quality
- `localOnlyDenial` cleanly replaces the inline `request.method == "run-setup"` check while preserving the exact `run-setup is local-only` message (verified by `runSetupAlwaysDeniedWhenNotLocal`).
- `jobs` array equality is order-sensitive, so a pure reorder trips the gate — this is deny-safe (a remote reorder is simply rejected) and not a correctness problem.
- Tests cover the meaningful cases: binaries change, jobs change, harmless toggle allowed, and unchanged binaries/jobs round-trip allowed.

### Verification
- `swift build` — succeeds (pre-existing deprecation warnings only).
- `swift test --filter LocalOnlyRPCGateTests` — all 5 tests pass.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 0 Green] findings. A tightly-scoped, well-tested security hardening that closes the remaining #594 Yellow without breaking legitimate remote config edits.

PR #634, opened 2026-07-09
fix: rebase web UI parity onto main (resolves #594 conflicts)

dgershman:
## Summary
- Rebases `feature/crow-593-web-ui-parity` onto latest `main` to clear the merge conflicts on #594.
- Ports CROW-604 `crow job` RPCs into CrowEngine/CrowDaemon after AppDelegate removal (main landed the CLI while this branch still had app handlers).
- Mutating `job-*` verbs are local-direct on `/rpc`, matching the existing jobs/`set-config` gate.

Branch protection blocks `--force-with-lease` on `feature/crow-593-web-ui-parity`, so this is the conflict-resolved tip as a new PR. Once this lands (or replaces #594), the dirty merge state is gone.

## Test plan
- [x] `swift test --filter JobRPCSupportTests` (CrowEngine)
- [x] `swift test --filter JobCommandParsingTests` (CrowCLI)
- [x] `swift test --filter LocalOnlyRPCGateTests` (CrowDaemon)
- [ ] CI green
- [ ] Confirm `crow job list/add/run` against a local crowd


dhilgaertner (changes requested):
## Code & Security Review

Scope note: this PR rebases the `feature/crow-593-web-ui-parity` branch onto `main` and is presented as a fresh PR (branch protection blocks the force-push on #594). The bulk of the ~21k-line diff is the web-UI-parity feature already reviewed across the many #594 rounds. The **genuinely new** work here is the last commit — `0bf5985 fix(rebase): port crow job RPCs into crowd after AppDelegate removal` — so the review concentrates there, plus a security/integrity pass on the rebase result.

### Critical Issues
None.

### Security Review
**Strengths:**
- The mutating `job-*` verbs (`job-add`/`edit`/`enable`/`disable`/`delete`/`duplicate`) are correctly added to `RPCWebSocketHandler.localOnlyDenial` (`RPCWebSocketHandler.swift:124`). They write the same `AppConfig.jobs` array that `set-config jobs` guards, which is a persistent-RCE surface (jobs execute prompts/binaries at launch), so gating them local-direct on `/rpc` is the right call.
- These six verbs are consumed **only** by the `crow job` CLI over the Unix socket (local-direct), confirmed via `JobCommands.swift`. The web UI uses the pre-existing `run-job` + `set-config`, so the new gate adds defense-in-depth with **no functional regression** for remote web users.
- `job-run` remaining ungated on `/rpc` is consistent with the existing `run-job` handler: both only trigger an already-vetted job (its command surface was created under the local-only gate), so no new injection surface.
- Input validation is thorough: `validateRepoSlug` rejects path-like components (`.`/`..`/empty) before the slug becomes an on-disk folder; schedule/name/prompt decoders range-check and reject blanks. Config mutations run under `ConfigStore.withConfigLock` with disk-write-before-return, and the scheduler's `jobsProvider` reads config fresh from disk each tick, so there is no in-memory drift.
- No leftover merge-conflict markers anywhere in the diff; the rebase is clean.

**Concerns:**
- None blocking.

### Code Quality
- `JobRPCSupport.swift` is cleanly factored as pure, unit-testable helpers (mirrors the `JobScheduler.finishDecision` pattern) with 12 passing tests covering round-trips, out-of-range, and blank/path-like rejection.
- Verified locally: `swift build` succeeds; `JobRPCSupportTests` (12) and the full CrowDaemon security suite incl. `LocalOnlyRPCGateTests` (62) pass.
- **Yellow — missing regression test for the new security gate.** `LocalOnlyRPCGateTests` (`DaemonSecurityTests.swift:685`) tests `set-config` binaries/jobs gating (`setConfigJobsChangeIsLocalOnly`, `setConfigBinariesChangeIsLocalOnly`) but adds **no** assertion that the six new `job-*` verbs return a local-only denial from `localOnlyDenial(...)`. The gate is a static `switch`, so a future refactor could silently drop a case and reopen the RCE boundary with nothing to catch it. Add parallel cases (each verb → non-nil denial when not local-direct) beside the existing set-config tests.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 0 Green] findings. The port itself is correct, well-validated, and tested; the one gap is a cheap regression test guarding the newly-extended local-only RCE gate. Adding it lands in the same round trip.

dgershman:
Superseded by a merge-into-parity PR (no force-push). Use the sync PR targeting feature/crow-593-web-ui-parity instead.

PR #635, opened 2026-07-09
Add jump-to-bottom control to terminal surface

### LINKED ISSUE #633: Add jump-to-bottom control when scrolled up in chat/terminal
## Summary
When you scroll up in a Crow chat/terminal scroll view, there is no quick way to jump back to the bottom (latest output). Add a jump-to-bottom affordance tied to scrolling.

## Problem
Scrolling to the top (or anywhere above the live edge) leaves you without a fast path back to the bottom — you have to manually scroll all the way down.

## Proposed behavior
- When the user is not pinned/scrolled to the bottom, show a jump-to-bottom control (button/fab or similar).
- Activating it scrolls (or jumps) to the latest content at the bottom.
- Hide or disable the control when already at the bottom.
- Apply wherever Crow has long scrolling agent/chat/terminal output and this friction shows up.

## Acceptance criteria
- Scrolling away from the bottom reveals a clear jump-to-bottom control.
- Using it returns to the bottom quickly.
- Control is not shown (or is inactive) when already at the bottom.

dgershman:
Closes #633

## What & why
Scrolling up over long agent/chat/terminal output left no fast way back to the live edge — you had to wheel all the way down. This adds a floating **jump-to-bottom** pill to the terminal surface: it appears when the viewport is scrolled above the bottom and snaps back to the latest line on click, then hides when pinned.

## Approach
In Crow the "chat/terminal" is a single **xterm.js** instance in a WKWebView (`terminal.html`) — Claude Code and the agents run inside it, and there's no separate native chat view. The control is therefore implemented **entirely inside `terminal.html`**:

- A gold-accent pill (matching the Corveil theme) floats bottom-right, hidden by default (`pointer-events: none` when hidden so it never blocks selection/wheel at the bottom).
- Visibility is driven by xterm's `buffer.active.viewportY`/`baseY` (at bottom ⇒ hidden). `onScroll` is the primary trigger — leaving the bottom only ever happens via a user scroll, since new output while pinned auto-scrolls and stays pinned — with `onRender` as a safety-net re-check.
- Click calls `term.scrollToBottom()` and restores typing focus.

Because the cockpit surface shares one webview across all tabs, every session gets the control automatically. No native Swift, message-handler bridge, or SwiftUI overlay was needed.

## Scope
Terminal-only. The only other `ScrollView`s in the app are the short kanban columns in `TicketBoardView`, which aren't long agent output and are out of scope.

## Tests
- Added `terminalHTMLHasJumpToBottomControl` to `BundledResourcesTests` (the xterm resources get re-vendored wholesale, so the control's presence is pinned against accidental overwrite), alongside the existing modified-Enter guard.
- `swift test --package-path Packages/CrowTerminal --filter BundledResourcesTests` — all 9 pass.
- `make build` — clean.

## Manual verification
Open a session terminal, produce lots of output (e.g. `seq 1 500`), scroll up → the pill appears bottom-right → click → snaps to the latest line and the pill disappears; typing focus retained. Stays hidden when already pinned to the bottom.

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/Resources/xterm/terminal.html
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/Resources/xterm/terminal.html
@@ -17,10 +17,44 @@
       width: 100%;
       height: 100%;
     }
+    /* Jump-to-bottom control (#633): floats over the xterm viewport,
+       styled to the Corveil gold accent (see CorveilTheme.swift). */
+    #crow-jump-bottom {
+      position: absolute;
+      right: 16px;
+      bottom: 16px;
+      width: 34px;
+      height: 34px;
+      padding: 0;
+      display: flex;
+      align-items: center;
+      justify-content: center;
+      font-size: 20px;
+      line-height: 1;
+      color: #ddc482;
+      background: #22262a;
+      border: 1px solid rgba(221, 196, 130, 0.35);
+      border-radius: 17px;
+      box-shadow: 0 2px 8px rgba(0, 0, 0, 0.5);
+      cursor: pointer;
+      z-index: 10;
+      opacity: 1;
+      transition: opacity 0.12s ease;
+    }
+    #crow-jump-bottom:hover {
+      background: #2a2f34;
+      border-color: rgba(221, 196, 130, 0.6);
+    }
+    #crow-jump-bottom.hidden {
+      opacity: 0;
+      visibility: hidden;
+      pointer-events: none;
+    }
   </style>
 </head>
 <body>
   <div id="terminal"></div>
+  <button id="crow-jump-bottom" class="hidden" aria-label="Jump to latest output" title="Jump to bottom">&#9662;</button>
   <script src="xterm.js"></script>
   <script src="xterm-addon-fit.js"></script>
   <script src="xterm-addon-image.js"></script>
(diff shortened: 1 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Pure client-side UI change inside the sandboxed `terminal.html` WKWebView — no new native bridge, message handler, or IPC surface introduced.
- No user-controlled data flows into the DOM; the button label/attributes are static literals. No injection, XSS, or `innerHTML` sink added.
- No secrets, network calls, or auth-adjacent logic touched.

**Concerns:**
- None.

### Code Quality
- `atBottom()` correctly uses xterm's public `buffer.active.viewportY >= baseY` contract; `onScroll` as the primary trigger with `onRender` as a safety-net re-check is sound, and the reasoning is captured in a clear comment.
- `updateJumpButton()` is idempotent and cheap (a single `classList.toggle`), so wiring it to the frequently-firing `onRender` carries no meaningful perf cost during output streaming.
- `hidden` state sets `pointer-events: none`, so the pill never blocks selection/wheel at the live edge — a good touch. Click handler restores typing focus after `scrollToBottom()`.
- All `jumpBtn` accesses are null-guarded, and the initial `updateJumpButton()` after `reportSize()` ensures correct start state.
- Accessibility: `aria-label` and `title` are present.

### Test Coverage
- `terminalHTMLHasJumpToBottomControl` pins the control's presence against the wholesale re-vendoring of xterm resources — the right guard for this file. Verified locally: `swift test --filter BundledResourcesTests` → all 9 pass.

- When visible, the pill sits over the bottom-right terminal cells and could intercept a click/selection there — but it only appears while scrolled *above* the bottom, so those cells aren't the live edge. Fine as-is.

### Summary Table

---

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Pure static bundled HTML/CSS/JS — no new user input, network, or message-handler surface. The button dispatches only `term.scrollToBottom()` / `term.focus()`, no data crosses the WKWebView bridge.
- No injection vectors: no `innerHTML`, no string interpolation, no dynamic script.

**Concerns:**
- None.

### Code Quality
- `atBottom()` logic is correct: xterm sets `buffer.active.viewportY === baseY` exactly when pinned to the live edge and `viewportY < baseY` when scrolled up, so `viewportY >= baseY` is the right bottom check (`>=` safely covers equality).
- Visibility coverage is complete: `onScroll` handles user scroll-up, `onRender` is a correct safety net for output arriving while already scrolled up (that path doesn't emit `onScroll`), and the initial `updateJumpButton()` after `reportSize()` sets the correct start state.
- Hidden state uses `visibility: hidden` + `pointer-events: none`, so the pill leaves the tab order and never blocks selection/wheel when pinned to the bottom — matches the stated intent.
- Accessible: `aria-label` and `title` present; all handlers guard on `jumpBtn` existence.
- Good regression guard: `terminalHTMLHasJumpToBottomControl` pins the control against wholesale re-vendoring of the xterm resources. Verified locally — all 9 `BundledResourcesTests` pass (`swift test --filter BundledResourcesTests`).

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 0 Green] findings. Small, self-contained, well-tested, no security surface.

---

PR #641, opened 2026-07-10
fix(new-terminal): default cwd to primary worktree when omitted (#639)

### LINKED ISSUE #639: new-terminal fails with "session_id and cwd required" when adding a terminal in a session
## Summary
Adding a terminal from within a Crow session fails with:

```
[crow] new-terminal failed: session_id and cwd required
```

## Expected
`new-terminal` (UI / RPC) should supply the active session id and a sensible cwd (e.g. primary worktree or session cwd) so a terminal can be created.

## Actual
The call errors because `session_id` and/or `cwd` are missing or not forwarded from the add-terminal path.

## Notes
- Repro: open a session → add a terminal.
- Likely a regression in the CLI/RPC param wiring for `crow new-terminal` (session context not being passed).

## Acceptance criteria
- Adding a terminal from a session succeeds without that error.
- `session_id` and `cwd` are populated from the active session / worktree context.

dgershman:
Closes #639

## Problem
Adding a terminal from the web UI failed with:

```
[crow] new-terminal failed: session_id and cwd required
```

The web UI's "+" add-terminal button (`Resources/web/app.js`) calls `rpc('new-terminal', { session_id })` without a `cwd` — it can't reliably know the session's worktree path. The `new-terminal` RPC handler in `EngineRouter.swift` required **both** `session_id` and `cwd`, so the request was rejected. (The CLI is unaffected — `crow new-terminal` forces `--cwd` at parse time, so only the web UI hit this.)

## Fix
Make `cwd` optional on the `new-terminal` RPC. When it's absent, derive it server-side from the session's **primary worktree** (mirroring `SessionService.addTerminal`), falling back to `devRoot` so the derived path always satisfies the existing path-traversal guard. The server is the source of truth for worktree paths, so this fixes every client without the browser needing to know the path. No web UI change required — `app.js` already sends `session_id`.

## Test
Adds a regression test in `EngineRouterSmokeTests` proving a `new-terminal` request with only `session_id` is no longer rejected with "session_id and cwd required", and that `cwd` is derived from the primary worktree. It asserts at the guard boundary so the side-effect-free smoke suite stays independent of a live tmux server.

## Scope
Targets `feature/crow-593-web-ui-parity`, where the web UI and the headless `EngineRouter` handler live (`AppDelegate.swift` is deleted on that branch). Main's handler carries the same guard but its only client (the CLI) always sends `cwd`, so the reported symptom doesn't occur there — a matching relaxation on main is a possible follow-up, out of scope here.

https://claude.ai/code/session_01WniYXBrXjPWjgtK3iBaGSc

### DIFF
--- a/Packages/CrowEngine/Sources/CrowEngine/EngineRouter.swift
+++ b/Packages/CrowEngine/Sources/CrowEngine/EngineRouter.swift
@@ -656,9 +656,21 @@ public func makeEngineRouter(_ ctx: EngineContext) -> CommandRouter {
                 return ["worktrees": .array(items)]
             },
             "new-terminal": { @Sendable params in
-                guard let idStr = params["session_id"]?.stringValue, let sessionID = UUID(uuidString: idStr),
-                      let cwd = params["cwd"]?.stringValue else {
-                    throw RPCError.invalidParams("session_id and cwd required")
+                guard let idStr = params["session_id"]?.stringValue, let sessionID = UUID(uuidString: idStr) else {
+                    throw RPCError.invalidParams("session_id required")
+                }
+                // cwd is optional (#639): the web UI's "+" add-terminal button
+                // sends only session_id — it can't reliably know the worktree
+                // path. Default to the session's primary worktree — mirroring
+                // SessionService.addTerminal — then devRoot, so the derived
+                // path always satisfies the traversal guard below.
+                let cwd: String
+                if let explicit = params["cwd"]?.stringValue, !explicit.isEmpty {
+                    cwd = explicit
+                } else {
+                    cwd = await MainActor.run {
+                        capturedAppState.primaryWorktree(for: sessionID)?.worktreePath ?? devRoot
+                    }
                 }
                 // Validate cwd is within devRoot to prevent path traversal
                 guard Validation.isPathWithinRoot(cwd, root: devRoot) else {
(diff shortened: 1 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- The path-traversal guard (`Validation.isPathWithinRoot`) still runs on the *derived* `cwd`, not just client-supplied values — so the new server-side derivation cannot bypass the devRoot containment check. This is actually stricter than the native `SessionService.addTerminal` path, which does not validate at all.
- No new input is trusted: `session_id` is still required and UUID-validated; the derived worktree path originates from server state, not the client.
- `devRoot` fallback is chosen precisely because it always satisfies the traversal guard — a deliberate, documented divergence from `addTerminal`'s home-directory fallback.

**Concerns:**
- None.

### Code Quality
- `EngineRouter.swift:667-674` — clean optional handling; `!explicit.isEmpty` correctly treats an empty `cwd` string as absent and falls through to derivation.
- Behavior mirrors `SessionService.addTerminal` (`primaryWorktree(for:)?.worktreePath`), keeping the web-UI and native code paths consistent. The comment explains the intentional `devRoot`-vs-home fallback difference.
- Edge case worth noting (non-blocking): if a session's primary worktree legitimately lives outside `devRoot`, the web "+" button will now fail the guard rather than succeed as the native path would. In practice worktrees live under `devRoot`, and this matches the guard's existing intent — no regression introduced here.

### Verification
- `swift build` — clean.
- `swift test --filter EngineRouterSmokeTests` — both tests pass, including the new `newTerminalDerivesCwdFromPrimaryWorktree`, which asserts at the guard boundary so the smoke suite stays free of a live tmux dependency.

### Summary Table

---

PR #646, opened 2026-07-10
feat(terminal): drag-and-drop images into agent chats (closes #644)

### LINKED ISSUE #644: Support drag-and-drop images into agent chats (parity with Cursor/Claude Code TUI)
## Summary
Crow does not support dragging and dropping images into agent chats. The native TUI versions of Cursor and Claude Code accept image drops into the composer; Crow should match that.

## Problem
When using Cursor or Claude Code inside Crow, dropping an image onto the chat/terminal does nothing useful (or is blocked). Users have to use whatever workaround the agent TUI exposes instead of the familiar drag-and-drop path.

## Expected
- Drag an image file (png/jpg/webp/gif, etc.) onto the Crow agent chat/terminal surface.
- The image is delivered to the agent the same way the standalone Cursor / Claude Code TUI would accept it (path insert, paste, or multimodal attachment — whichever those agents already support).
- Works for managed agent sessions (at least Cursor and Claude Code); extend to other agents if they support the same input path.

## Actual
Image drag-and-drop into Crow chats is missing / non-functional compared to the standalone TUIs.

## Notes / implementation hints
- Crow hosts agents via xterm.js + tmux/PTY. Native AppKit/WebKit drag may currently be suppressed or only path-forwarded for text/files in limited cases.
- Check `XTermSurfaceView` / WKWebView drag handlers and any existing file-drop → path paste path; extend (or add) an image-drop path that the agent TUI understands.
- Prefer parity with how Cursor and Claude Code TUIs already consume dropped images rather than inventing a Crow-only attachment UI.

## Acceptance criteria
- Dropping an image onto a Cursor or Claude Code session in Crow attaches/inserts it so the agent can see/use the image.
- Non-image drops keep existing behavior (if any).
- No regression for normal text typing/paste in the agent composer.

dgershman:
Closes #644

## What

Drag-and-drop images into the agent terminal, with parity to the standalone Cursor / Claude Code TUIs (which accept a Finder image drop by inserting the file path into the composer).

Because the browser can't read a dropped file's filesystem path — and may even be a remote client — the image **bytes** are uploaded to crowd, which writes them into the session's artifacts scratch dir on the host and returns an absolute path. The client then pastes that (backslash-escaped) path into the composer via bracketed paste, **without submitting**, so the user can add a prompt before pressing Enter.

Reusing the existing per-session artifacts dir (`$TMPDIR/crow/artifacts/<uuid>/`) means the agent reads the file at a known absolute path **and** the drop shows up in the client's Artifacts "Images" panel.

## Changes

- **`Artifacts.swift`** — `POST /artifacts/:session`, the write side of the existing `GET /artifacts/:session/:file`. Behind `WebAuthMiddleware` (valid session required) plus a same-origin check (anti-CSRF, matching the secret-POST pattern). 10 MB cap, raster-only (`png/jpg/jpeg/gif/webp` — SVG deliberately excluded as script-bearing), sanitized collision-free filename (`drop-<8hex>-<stem>.<ext>`). Uploads take a dedicated HTTP route because both the `/rpc` and `/terminal` WebSockets cap frames at 1 MB.
- **`app.js`** — `dragover`/`drop` on the terminal; image files are uploaded and the returned path pasted into the composer via `term.paste` (bracketed). Non-image drops are ignored and never navigate the app away.
- **`tauri.conf.json`** — `dragDropEnabled: false` so the desktop webview forwards OS file drops to the DOM (Tauri v2 intercepts them by default), unifying browser + desktop on one code path.
- **Tests** — `uploadExtension` / `sanitizedUploadName` helpers in `ArtifactsTests`.

## Notes

- Based on `feature/crow-593-web-ui-parity` (the web-UI architecture), not `main`.
- `make daemon` builds clean; `ArtifactsTests` (7 tests) pass.

## Verification

1. Run crowd + open the web UI; select a Claude Code / Cursor session.
2. Drag a `.png` from Finder onto the terminal → escaped host path appears in the composer (not submitted); the image also appears in the Artifacts panel; the agent attaches it on send.
3. Path with spaces / multiple images → handled. Drag a `.txt`/`.pdf` → ignored (no navigation).
4. Typing + clipboard paste in the composer → unaffected.
5. Desktop: rebuild the Tauri app and repeat the Finder-drop check inside the window.

https://claude.ai/code/session_01H1NvQbNrsEekQymspbKJ4J

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Artifacts.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Artifacts.swift
@@ -1,5 +1,6 @@
 import Foundation
 import CrowCore
+import HTTPTypes
 import Hummingbird
 import NIOCore
 
@@ -13,6 +14,16 @@ import NIOCore
 /// outside this scratch tree is reachable.
 enum Artifacts {
     static let imageExtensions: Set<String> = ["png", "jpg", "jpeg", "gif", "webp", "svg"]
+
+    /// Extensions accepted for *dropped* images (drag-and-drop into the composer,
+    /// #644). Deliberately a subset of `imageExtensions`: SVG is script-bearing
+    /// and not a raster the agents consume, so it's kept out of the input path.
+    static let uploadableImageExtensions: Set<String> = ["png", "jpg", "jpeg", "gif", "webp"]
+
+    /// Cap for a single dropped image (10 MB). Screenshots blow past the 1 MB
+    /// WebSocket frame limit, which is why uploads take this dedicated HTTP route
+    /// rather than the `/rpc` or `/terminal` sockets.
+    static let maxUploadBytes = 10 * 1024 * 1024
 
     /// A session's dir, or nil if `sessionID` isn't a valid UUID (traversal guard).
     /// Shares `CrowCore.ArtifactPaths` with the terminal-env injection so the
@@ -38,8 +49,10 @@ enum Artifacts {
         }.sorted { $0.2 > $1.2 }
     }
 
-    /// Mount `GET /artifacts/:session/:file`.
-    static func mount(on router: Router<CrowHTTPContext>) {
+    /// Mount `GET /artifacts/:session/:file` (view) and `POST /artifacts/:session`
+    /// (drop an image into the composer, #644). Both sit behind `WebAuthMiddleware`;
+    /// the write side additionally requires a same-origin request (anti-CSRF).
+    static func mount(on router: Router<CrowHTTPContext>, boundHost: String) {
         router.get("/artifacts/:session/:file") { _, context -> Response in
             guard let session = context.parameters.get("session"),
                   let dir = dir(sessionID: session),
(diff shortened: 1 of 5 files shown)


dhilgaertner (approved):
## Code & Security Review

Reviewed the actual PR diff (5 files, base `feature/crow-593-web-ui-parity`). `swift build` for `CrowDaemon` is clean (only pre-existing deprecation warnings) and all 7 `ArtifactsTests` pass locally.

This adds a write side (`POST /artifacts/:session`) to the existing read-only artifacts route so drag-dropped images land in the session scratch dir and their host path is pasted (unsubmitted) into the composer. The security posture is careful and matches the established `SecretRoutes` pattern.

### Critical Issues
None.

### Security Review
**Strengths:**
- **Layered gate on the write route** — `WebAuthMiddleware` (added to `httpRouter` before `Artifacts.mount`, so it wraps the POST) requires a valid session / local-trust, *plus* a same-origin check via `WebSocketOriginGuard.isAllowedOrigin` (`Artifacts.swift:93`). That function is exhaustively unit-tested in `DaemonSecurityTests` (rejects `evil.com`, `127.0.0.1.evil.com`, opaque `null` origin, public IPs), so a cross-site page cannot drive an upload+inject via the ambient cookie even though the loopback peer is auto-trusted by `WebAuthGuard`.
- **Traversal-proof filenames** — server-side `sanitizedUploadName` reduces any input to `drop-<8hex>-<[A-Za-z0-9_-] stem>.<ext>` and is re-checked with `isSafeImageName`; the `../../etc/passwd` case is covered by a test. The random token also prevents same-name overwrites.
- **SVG deliberately excluded** from the upload path (`uploadableImageExtensions` is a strict subset), keeping script-bearing content out of the input path — consistent with the GET side forcing `Content-Disposition: attachment` for SVG.
- **10 MB cap** enforced via `request.body.collect(upTo:)` → 413; content-type params (`; charset=…`) stripped before media-type matching.
- **No shell-injection surface** — the pasted path is routed through xterm bracketed paste (no trailing newline → inserted, not submitted) and additionally client-side `shellEscapePath`d; since the server-sanitized basename has no metacharacters, this is belt-and-suspenders.

**Concerns:** none blocking.

### Code Quality (all Green / optional)
- **Session-switch race (`app.js` `uploadDroppedImages`)** — `sid` is captured at call time, but `term` is a singleton whose `termWs` is repointed on session switch. If the user switches sessions while a large upload is in flight, `term.paste(...)` lands the path in the *new* session's composer. Impact is low (path is not submitted and still points to the correct file), but a `if (sid === selectedId && term)` guard before pasting would close it.
- **No content/magic-byte validation** — arbitrary bytes are accepted under an image extension. Acceptable given the local-user trust boundary and that GET serves rasters with a correct `image/*` content-type (SVG excluded), but worth a note.
- **Route-level test gap** — only the pure helpers (`uploadExtension`, `sanitizedUploadName`) are unit-tested; the POST wiring (403 origin reject, 413/415, empty-body 400) has no integration test. The origin function itself is well covered elsewhere, so this is minor.
- **Client MIME filter** — `drop` filters strictly on `file.type.startsWith("image/")`; an image dropped with an empty MIME (happens on some platforms) is silently ignored even though the server could derive the ext from the filename. UX-only.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 4 Green] findings. Security-critical logic is sound, tested, and consistent with the codebase's established patterns; the remaining items are optional polish.

---

PR #658, opened 2026-07-10
fix(tmux): stop main-thread run-loop re-entrancy SIGSEGV in tab switching

### LINKED ISSUE #653: Fix main-thread run-loop re-entrancy crash in tmux tab switching (SIGSEGV in _NSWindowTransformAnimation dealloc)
A user on the latest `main` (installed `/Applications/Crow.app`) hit a `signal=11` (SIGSEGV) crash shortly after window creation. Crash is a main-thread run-loop re-entrancy fault in tmux tab/surface switching.

## Symptom

Crash log stack (main thread), bottom-up:

```
TerminalSurfaceView.syncSurface(into:) closure   (DispatchQueue.main.async block)
 -> TmuxBackend.makeActive(id:)                  TmuxBackend.swift:302
 -> TmuxController.selectWindow(index:)          TmuxController.swift:159
 -> TmuxController.run(_:timeout:)               TmuxController.swift:51
 -> Process.waitUntilExit()                      <- spins a NESTED run loop
 -> __CFRunLoopRun -> CA::Transaction::commit -> objc_autoreleasePoolPop
 -> -[_NSWindowTransformAnimation dealloc]       <- re-entrant teardown -> SIGSEGV
```

It crashed ~0.4s after `Main app launch complete — creating window`, i.e. as the cockpit surface first attached and the window-open animation was mid-commit.

## Root cause

`TmuxController.run()` waits for the `tmux select-window` subprocess with `Process.waitUntilExit()` **on the main thread** (`TmuxController.swift:51`). `waitUntilExit()` pumps a *nested run loop*; that nested loop services a pending CoreAnimation commit, which drains an autorelease pool and deallocates the in-flight `_NSWindowTransformAnimation` re-entrantly — segfaulting.

It is a **race** (needs a `select-window` wait to land during a live window animation), which is why it reproduces intermittently and only bit one of two builds in the reported logs.

Contributing: `syncSurface` already defers `makeActive` via `DispatchQueue.main.async` (`TerminalSurfaceView.swift:64`), but that deferral only avoids **layout** re-entrancy (per the comment at `:46-55`) — it does nothing about the **run-loop pump**, since the blocking subprocess still runs on the main thread.

## Fix (pattern already exists in the same file)

`startManagerExitMonitor` (`TmuxBackend.swift:687-699`) already does the right thing and says why — it runs the blocking `run`/`displayMessage` inside `await Task.detached { … }.value` to "keep that subprocess off the UI thread." The tab-switch hot path never got the same treatment.

Ship both tiers:

- **Tier 1 (surgical, targets the crash):** move the `select-window` subprocess off the main thread. `TmuxBackend` is `@MainActor` (`TmuxBackend.swift:30`) but `TmuxController` is a `Sendable` value struct — mirror `startManagerExitMonitor`: capture `controller` + window index on main, run `selectWindow` inside `await Task.detached { … }.value`, then record `activeTerminalID` back on main. `syncSurface` calls it as `Task { await … }`. Keep the AppKit re-parent / `makeFirstResponder` on main; only the blocking shell-out moves off.
- **Tier 2 (defense-in-depth, protects every caller):** stop `TmuxController.run()` from pumping the run loop at all — replace `p.waitUntilExit()` with a wait on a `DispatchSemaphore` signaled from `p.terminationHandler`. That truly blocks the calling thread (~70ms p95, still bounded by the existing 2s `ProcessWatchdog`) instead of re-entering the loop, so no future caller can reintroduce this by shelling out on main. `loadBufferFromStdin` (`TmuxController.swift:206`) and the other `waitUntilExit()` sites should get the same treatment.

## Acceptance criteria

- Tab switching (`makeActive` / `selectWindow`) no longer runs a run-loop-pumping subprocess on the main thread.
- `TmuxController.run()` waits for the child without pumping the main run loop; timeout/watchdog behavior is preserved.
- No `_NSWindowTransformAnimation`-dealloc SIGSEGV under repeated launch + tab-switch during the window-open animation.

dgershman:
Closes #653

## Problem

A user on installed `Crow.app` hit a `signal=11` SIGSEGV ~0.4s after window creation, during the window-open animation. Root cause: `TmuxController.run()` waited for its `tmux` subprocess with `Process.waitUntilExit()`, which spins a **nested run loop**. When a `select-window` wait landed on the main thread mid-animation, that nested loop serviced the pending CoreAnimation commit and re-entrantly deallocated the in-flight `_NSWindowTransformAnimation` → segfault. It's a race, so it reproduced intermittently.

`syncSurface` already deferred `makeActive` via `DispatchQueue.main.async`, but that only avoids *layout* re-entrancy — the blocking subprocess still ran on the main thread and still pumped the run loop.

## Fix (two tiers, mirroring existing patterns)

**Tier 2 — crash-critical: `TmuxController` no longer pumps the run loop.**
Every `waitUntilExit()` site (`run`, `loadBufferFromStdin`, `versionString`) now arms a `terminationHandler`-signaled `DispatchSemaphore` **before** `run()` and blocks on it. `terminationHandler` fires on Foundation's own background queue, independent of any thread's run loop, so the wait never pumps. This is deliberately **not** the "offload `waitUntilExit()` to a background thread" approach — `Process` delivers termination via the *launching* thread's run loop, so offloading deadlocks a process launched on the (now non-pumping) main thread. The 2s `ProcessWatchdog` still bounds the wait (its `terminate()` fires the handler). Same shape as `SessionService`'s existing `terminationHandler` continuation.

**Tier 1 — defense + responsiveness: move the tab-switch shell-out off the main thread.**
`TmuxBackend.makeActive(id:)` is now `async` and runs only the blocking `select-window` inside `Task.detached`, mirroring `startManagerExitMonitor`. `ensureRunningServer()` and the `activeTerminalID` record stay on the main actor. `TerminalSurfaceView.syncSurface` fires it as a detached `Task` inside its existing deferred block.

## Acceptance criteria

- [x] Tab switching (`makeActive`/`selectWindow`) no longer runs a run-loop-pumping subprocess on the main thread.
- [x] `TmuxController.run()` waits for the child without pumping the main run loop; timeout/watchdog behavior preserved.
- [x] No `_NSWindowTransformAnimation`-dealloc SIGSEGV under repeated launch + tab-switch during the window-open animation.

## Testing

Full `CrowTerminal` suite passes serialized — **80 tests in 11 suites, ~20s**, no deadlock (an intermediate offload-to-background approach *did* deadlock, which is what surfaced the launching-thread-run-loop subtlety documented above). `makeActive` tests updated for the async signature.

https://claude.ai/code/session_01QZRCQZBXhQUDCShc3Qiv4M

### DIFF
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TerminalSurfaceView.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TerminalSurfaceView.swift
@@ -48,9 +48,10 @@ public struct TerminalSurfaceView: NSViewRepresentable {
     /// responder. Same deferral rationale as `makeNSView`: attaching the shared
     /// surface (`cockpitSurface()` + WKWebView setup) can pump the main run loop,
     /// and updateNSView is itself invoked during layout, so doing this work
-    /// synchronously here re-enters layout. (`makeActive` is now `async` and runs
-    /// its `select-window` shell-out off the main actor — #653 — so it is fired
-    /// as a detached `Task` inside the deferred block.)
+    /// synchronously here re-enters layout. (`makeActive`'s `select-window`
+    /// shell-out runs synchronously on the main actor but no longer pumps a run
+    /// loop while it waits — #653 — so it no longer re-enters an in-flight
+    /// window animation.)
     @MainActor
     public func updateNSView(_ nsView: NSView, context: Context) {
         syncSurface(into: nsView)
@@ -69,7 +70,7 @@ public struct TerminalSurfaceView: NSViewRepresentable {
                 Self.attach(surface: surface, to: container)
             }
             surface.createSurface()
-            Task { try? await TmuxBackend.shared.makeActive(id: id) }
+            try? TmuxBackend.shared.makeActive(id: id)
             if let window = container.window,
                surface.window === window,
                window.firstResponder !== surface {
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TerminalSurfaceView.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TerminalSurfaceView.swift
@@ -45,10 +45,12 @@ public struct TerminalSurfaceView: NSViewRepresentable {
 
     /// Re-parent the surface if SwiftUI replaced the container, switch the
     /// attached tmux client to this terminal's window, and take first
-    /// responder. Same deferral rationale as `makeNSView`: `makeActive` shells
-    /// out to `tmux select-window` (which pumps the run loop), and updateNSView
-    /// is itself invoked during layout, so doing this work synchronously here
-    /// re-enters layout.
+    /// responder. Same deferral rationale as `makeNSView`: attaching the shared
+    /// surface (`cockpitSurface()` + WKWebView setup) can pump the main run loop,
+    /// and updateNSView is itself invoked during layout, so doing this work
+    /// synchronously here re-enters layout. (`makeActive` is now `async` and runs
+    /// its `select-window` shell-out off the main actor — #653 — so it is fired
+    /// as a detached `Task` inside the deferred block.)
     @MainActor
     public func updateNSView(_ nsView: NSView, context: Context) {
         syncSurface(into: nsView)
@@ -67,7 +69,7 @@ public struct TerminalSurfaceView: NSViewRepresentable {
                 Self.attach(surface: surface, to: container)
             }
             surface.createSurface()
-            try? TmuxBackend.shared.makeActive(id: id)
+            Task { try? await TmuxBackend.shared.makeActive(id: id) }
             if let window = container.window,
                surface.window === window,
                window.firstResponder !== surface {
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -315,14 +315,17 @@ public final class TmuxBackend {
     /// Bring `id`'s window into focus. Called by the UI when the user
     /// switches tabs.
     ///
-    /// `async` because the blocking `select-window` shell-out runs off the main
-    /// actor (`Task.detached`) — same rationale as `startManagerExitMonitor`.
-    /// `TmuxController.run` no longer pumps the run loop (#653), but shelling out
-    /// on the UI thread would still stall it ~70ms per switch; keeping it off the
-    /// main actor avoids that too. Only the pure value-struct `selectWindow` moves
-    /// off; `ensureRunningServer()` and the `activeTerminalID` record stay on the
-    /// main actor (they touch actor state).
-    public func makeActive(id: UUID) async throws {
+    /// The `select-window` shell-out runs synchronously on the main actor so tab
+    /// switches stay strictly serialized — the dedup guard, the `select-window`,
+    /// and the `activeTerminalID` update are atomic w.r.t. other switches, giving
+    /// last-switch-wins. This is safe even mid-window-open-animation because
+    /// `TmuxController.run` no longer pumps a nested run loop while waiting
+    /// (#653): the main thread blocks on the child without re-entering the
+    /// in-flight CoreAnimation commit that used to SIGSEGV. Deliberately NOT moved
+    /// off the main actor — doing so let two in-flight switches race and could
+    /// leave tmux focused on the wrong window until the next `updateNSView`
+    /// (PR #658 review).
+    public func makeActive(id: UUID) throws {
         guard let windowIndex = bindings[id] else {
             throw TmuxBackendError.unknownTerminal(id)
         }
@@ -330,18 +333,8 @@ public final class TmuxBackend {
         // subprocess (see `activeTerminalID`).
         if id == activeTerminalID { return }
         let start = Date()
-        let ctrl: TmuxController
         do {
-            ctrl = try ensureRunningServer()
-        } catch {
-            reportIfTimeout(error)
-            throw error
-        }
-        do {
-            // `ctrl` is a Sendable value struct + `windowIndex` an Int — both
-            // safe to hand to a detached task. Blocks that task's thread on the
-            // tmux subprocess instead of the UI thread.
-            try await Task.detached { try ctrl.selectWindow(index: windowIndex) }.value
+            try ensureRunningServer().selectWindow(index: windowIndex)
         } catch {
             reportIfTimeout(error)
             throw error
--- a/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift
@@ -314,7 +314,15 @@ public final class TmuxBackend {
 
     /// Bring `id`'s window into focus. Called by the UI when the user
     /// switches tabs.
-    public func makeActive(id: UUID) throws {
+    ///
+    /// `async` because the blocking `select-window` shell-out runs off the main
+    /// actor (`Task.detached`) — same rationale as `startManagerExitMonitor`.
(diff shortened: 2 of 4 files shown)


dhilgaertner (dismissed):
## Code & Security Review

Focused review of the tmux tab-switch SIGSEGV fix (#653). I read all four changed files, built the `CrowTerminal` package (clean), and ran the tmux suite.

### Critical Issues
None. The crash-critical Tier-2 change is correct.

### Security Review
**Strengths:**
- No new external input, no secrets, no injection surface. `select-window`/`load-buffer`/`-V` invocations continue to pass structured `[String]` argument arrays to `Process` (no shell), so the `Task.detached` move introduces no new command-injection risk.
- `ctrl` (Sendable value struct) + `windowIndex` (`Int`) are the only values crossing into `Task.detached` — no shared mutable state escapes the main actor.

**Concerns:** None.

### Code Quality

**Tier 2 — `TmuxController` termination signal: correct.** `makeTerminationSignal(for:)` is armed *before* `p.run()` at every site (`run` L49, `loadBufferFromStdin` L194, `versionString` L286), closing the lost-wakeup race. Waiting on the semaphore blocks without pumping a nested run loop, which is exactly what kills the `_NSWindowTransformAnimation` re-entrancy. Watchdog interplay is preserved (`terminate()` → `terminationHandler` → `done.signal()`), and the `loadBufferFromStdin` write-error path (L202) correctly waits on the same semaphore. The rationale comment (L317-336) is accurate and valuable.

- 🟡 **Async `makeActive` drops the ordering guarantee the old synchronous code had** — `TmuxBackend.swift:331` / `TmuxBackend.swift:344` / `TerminalSurfaceView.swift:72`. Previously `makeActive` ran synchronously on the main thread from the deferred block, so tab switches were strictly serialized and last-switch-wins. Now it is `async`, awaits a fire-and-forget `Task.detached` shell-out, and is itself launched from an unstructured `Task { try? await … }`. Two consequences under rapid switching (two switches inside the ~70ms `select-window` window):
  - The `if id == activeTerminalID { return }` guard (L331) can read a **stale** `activeTerminalID`. Sequence A(active)→B→A: `makeActive(B)` suspends on its detached `select-window(B)` before recording B; `makeActive(A)` then sees `activeTerminalID` still == A and returns early, issuing *no* `select-window(A)`; `makeActive(B)` resumes and records B. Final state: tmux is focused on B and the record says B, but the user's last action selected A — so the shared cockpit surface shows/routes keystrokes to the wrong window.
  - Even absent the guard, two `Task.detached` `select-window` subprocesses race with no ordering guarantee; the tmux server's final focus is whichever subprocess happens to finish last.
  
  This self-heals only on the *next* SwiftUI `updateNSView` for the visible tab (not on internal xterm re-renders), so the desync can persist visibly rather than for a few ms. Consider recording intended id / a latest-wins guard / debounce, or awaiting on the main actor so switches serialize. Narrow (sub-70ms double-switch) and not a crash, but it is a behavioral regression from the pre-PR serialized path and worth folding into this round.

### Testing
- `swift build` — clean.
- `makeActive` tests correctly updated for the async signature (`await #expect(throws:)`, `try await`) and pass.
- One suite test, `retryReadinessEmitsTimedOutWhenSentinelMissing`, fails on this machine — but it fails **identically on `main`** and exercises readiness-sentinel timing, which this PR does not touch. Pre-existing/environmental, not a regression from this change.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 0 Green] findings. The Tier-2 crash fix is solid; the single Yellow is the ordering regression introduced by the Tier-1 async conversion.

---

dgershman:
Addressed the Yellow ordering-regression finding in `bfa39a4` ("keep makeActive synchronous to preserve tab-switch ordering (#658 review)").

**What changed:** `TmuxBackend.makeActive(id:)` is reverted from `async` back to a synchronous `throws` on the `@MainActor` — one of the remedies you suggested ("awaiting on the main actor so switches serialize"). The dedup guard → `select-window` → `activeTerminalID` update are now atomic with no suspension point, so switches are strictly serialized / last-switch-wins. This removes both failure modes you called out: the stale `activeTerminalID` read (A→B→A) and the two racing `Task.detached` `select-window` subprocesses. `TerminalSurfaceView.syncSurface` calls it synchronously again inside its existing deferred block, and the `makeActive` tests are back to the sync signature.

**Still crash-safe:** the actual SIGSEGV fix is Tier-2, which is untouched — `TmuxController` no longer pumps a nested run loop (`makeTerminationSignal` + `DispatchSemaphore` at all three `run`/`loadBufferFromStdin`/`versionString` sites). So the synchronous main-thread wait can't re-enter the in-flight `_NSWindowTransformAnimation` commit. Tier-1's off-main-thread move (the source of the ordering race) is deliberately dropped; it was defense/responsiveness only, not the crash fix.

**Verification:** `swift build` clean; `TmuxBackendTests` — 41 tests in 4 suites pass (`makeActiveTracksActiveTerminal` covers the serialization). Note the review was submitted ~7 min before this commit landed, so it evaluated the earlier async version — re-review requested against current HEAD.

🐦‍⬛

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No new external input, network, or credential surface. `tmux` args are constructed from internal, typed state (window indices, session name), not user-controlled strings — no injection vector introduced.
- The `terminationHandler`-signaled `DispatchSemaphore` replaces `waitUntilExit()` without weakening the watchdog: `ProcessWatchdog.terminate()` still bounds every wait at `timeout` (2s), so a wedged tmux server cannot indefinitely freeze the caller.

### Code Quality
The core change — `makeTerminationSignal(for:)` arming a `terminationHandler` → `DispatchSemaphore` before `run()`, then `done.wait()` in place of `p.waitUntilExit()` — is correct and carefully reasoned:

- **No lost-wakeup race:** the handler is armed before `run()`, and each `Process` gets a fresh semaphore (init 0, one `signal()` balanced by one `wait()`), so there's no over-wait/dealloc-precondition crash.
- **Never-launched processes can't hang the caller:** every early-out path (`versionString`'s `guard (try? p.run()) != nil`, and the `try p.run()` throw sites in `run`/`loadBufferFromStdin`) returns/throws before reaching `done.wait()`.
- **Watchdog interplay preserved:** `terminate()` → SIGTERM → child exit → handler fires → `signal()`, so `didFire`/`.timedOut` semantics are unchanged. In `loadBufferFromStdin`'s EPIPE catch path, the child is already dead so `done.wait()` returns immediately — no hang.
- **Re-entrancy root cause addressed:** `terminationHandler` fires on Foundation's own queue, so the blocked thread no longer pumps a nested run loop — which is exactly what removes the `_NSWindowTransformAnimation`-dealloc re-entrancy (#653). The doc comment explaining why offloading `waitUntilExit()` to a background thread would deadlock (termination delivered via the launching thread's run loop) is accurate and valuable.
- **The review-driven revert (bfa39a4) is sound:** keeping `makeActive` synchronous on the main actor restores strict last-switch-wins tab ordering; the crash fix depends only on the Tier-2 no-pump wait, not on moving the shell-out off the main actor. Blocking-without-pumping is safe mid-animation.

**Verification:** `swift build` clean; `CrowTerminal` suite runs. The one failing test — `retryReadinessEmitsTimedOutWhenSentinelMissing` — fails **identically on `main`**; it's a pre-existing host-timing sensitivity (an idle-shell precmd fires within the 150ms budget), unrelated to this change. `makeActive`/backend tests pass.

**Minor (Green — consider, non-blocking):**
- A couple of comments outside the diff still describe the old primitive — `TmuxBackend.swift:734` and `:955` say `TmuxController.run` "blocks on `waitUntilExit()`", which is now the semaphore wait. The *reasoning* (keep the blocking subprocess off the UI thread) still holds, but the wording is now stale.
- The PR description still narrates the Tier-1 async `makeActive` conversion that commit bfa39a4 reverted; the net change is the controller termination-signal rewrite plus doc comments. The inline code comments are accurate — only the PR body drifted.

### Summary Table

---

PR #638, opened 2026-07-10
fix(terminal): stop resize clobber by coalescing + deduping fits

dgershman:
Closes #637

## Problem
The embedded terminal intermittently painted garbage — double lines, staggered/overlapping rows, misaligned frames — after resizing the window/sidebar or switching tabs. The stack is a single shared xterm.js cockpit (WKWebView) → native PTY (`TIOCSWINSZ`) → tmux → agent TUI. The agent clobbers when SIGWINCH'd faster than it can redraw.

## Root cause (resize spam)
1. `terminal.html` `reportSize()` posted `crowResize` on **every** fit, even when cols/rows were unchanged → a no-op fit still SIGWINCH'd the agent.
2. The DOM `resize` listener (`window.addEventListener('resize', reportSize)`) was **un-debounced** → a live drag fired dozens of fits/sec, each posting a resize mid-frame.
3. `XTermSurfaceView.layout()` called `scheduleFit()` on **every** layout pass, including tab-switch reparents and no-op passes that didn't change size.

## Fix
- **`terminal.html`** — `applyFit()` posts `crowResize` only when `cols`/`rows` actually change since the last post (same-size cycles → zero SIGWINCH). `scheduleFit()` coalesces drag bursts to one fit per `requestAnimationFrame`; both `window.crowFit` (Swift) and the DOM `resize` listener route through it. Added a one-shot re-fit after `document.fonts.ready` so the initial grid corrects once the `MesloLGS NF` cell metrics settle (first-paint off-by-a-few-columns).
- **`XTermSurfaceView.swift`** — `layout()` only `scheduleFit()`s on a real `bounds.size` change (tracked via `lastFitSize`, reset in `destroy()`), so reparents/no-op passes stop re-posting the size.

No tmux/backend changes needed: with a single attached client and no `window-size`/`aggressive-resize` overrides, one clean `TIOCSWINSZ` makes tmux resize the pane and force the agent redraw on its own. Eliminating the storm is the whole fix.

## Acceptance criteria
- ✅ Resize/sidebar/tab-switch no longer leaves permanently clobbered content — drags coalesce to one settled resize tmux redraws cleanly.
- ✅ Same-size fit cycles don't spam SIGWINCH — JS dedupe + `layout()` size-guard.
- ✅ Agent TUIs realign cleanly after real size changes (incl. the post-font-load correction), no manual refocus.

## Testing
- `make build` — clean.
- `swift test --package-path Packages/CrowTerminal` — 80/80 pass.
[...]

PR #663, opened 2026-07-10
fix(web-terminal): debounce reflow thrash + right-click "Reload terminal" (#661)

### LINKED ISSUE #661: xterm.js terminal reflow thrash on feature/crow-593-* web UI — display corrupts, needs full page reload; add right-click 'Reload terminal'
On the `feature/crow-593-*` web UI, the xterm.js terminal frequently jumps around and corrupts its own display (overlapping/misaligned lines, garbled box-drawing — see screenshot). The only recovery is refreshing the entire browser page, which is disruptive. Two asks: (1) fix the underlying reflow thrash, and (2) at minimum, add a right-click action to reload just the terminal surface without a full page refresh.

Screenshot (local): `Screenshot 2026-07-09 at 9.46.11 PM.png` — terminal text overwriting itself, rows misaligned, box-drawing broken. Classic xterm.js fit/reflow corruption.

## Root cause (shared terminal UI)

`Packages/CrowTerminal/Sources/CrowTerminal/Resources/xterm/terminal.html`:

- `reportSize()` (line 87-90) calls `fitAddon.fit()` then posts `{cols, rows}` to native — with **no debounce and no guard** against redundant/degenerate fits.
- It is wired to fire on **every** `window` `resize` event: `window.addEventListener("resize", reportSize)` (line 108).
- In the desktop WKWebView the surface is a fixed size, so `resize` rarely fires. In a **browser** (the crow-593 web build) `resize` fires constantly — layout reflow, flexbox, scrollbar appearance, devicePixelRatio changes — so `fit()` runs in a tight loop. Combined with the tmux client re-reflowing on each `cols/rows` change, the grid thrashes and the display corrupts. A full page reload fixes it only because it disposes and re-creates the `Terminal` and fits once against a now-stable layout.
- The `contextmenu` handler (line 82-84) currently only calls `event.preventDefault()` — it actively **swallows** the right-click, so there is no way to reload the terminal short of reloading the page.

## Fix

**1. Stop the reflow thrash (fundamental):**
- Debounce/RAF-batch `reportSize` so `fit()` runs once after resize activity settles, not on every event.
- Observe the terminal **container** via `ResizeObserver` instead of (or in addition to) the global `window` `resize`, so we react to the surface actually changing, not unrelated page layout.
- Only `postMessage({cols, rows})` when the fitted dimensions actually change from the last reported values — suppress no-op resizes that trigger needless tmux reflow.
- Guard against fitting while the element is 0×0 / detached (degenerate `proposeDimensions`), which produces junk grids.

**2. Right-click "Reload terminal" (minimum viable):**
- Replace the swallow-only `contextmenu` handler with a small menu (or a plain right-click action) that re-initializes just the xterm surface: dispose + recreate the `Terminal` (or reset + re-fit), re-attach the fit addon, and re-request a full repaint/scrollback from the backend — without a browser page reload.
- Keep paste behavior intact.

## Target branch

The web build lives on `feature/crow-593-*`, and that is where the corruption is reported, so land + verify the fix there. Note `terminal.html` is a **shared** asset (desktop uses it too), so the debounced/observed fit is a net win for both surfaces — but `crow:auto` branches off `main` by default; base this work on the `feature/crow-593-*` branch / its PR so the web build can be validated, rather than a fresh branch off `main`.

## Acceptance criteria

- Resizing the browser window (and normal page-layout churn) no longer corrupts the xterm.js display; `fit()` is debounced/observer-driven and only reports real dimension changes.
- The garbled/overlapping rendering in the screenshot no longer reproduces during resize on the web UI.
- Right-clicking the terminal offers a "Reload terminal" action that repaints/reinitializes only the xterm surface (no full page refresh) and recovers a corrupted display.
- No regression to the desktop WKWebView terminal (fit still works on tab switch / window resize) or to paste.

dgershman:
Closes #661

## Problem
On the crow-593 web UI the xterm.js terminal jumps around and corrupts its own display (overlapping rows, garbled box-drawing); the only recovery was a full browser refresh.

## Root cause
The web UI's terminal surface is **`app.js`'s embedded xterm** — not the shared `CrowTerminal/.../terminal.html` the ticket assumed. That desktop copy was already debounced in #638, and the daemon's `/terminal.html` is a debug-only page. `app.js`'s `fitTerminal()` called `fitAddon.fit()` + a resize control frame on **every** browser `resize` event with **no debounce, no dedup, and no degenerate-size guard**, so normal page churn (flexbox, scrollbars, devicePixelRatio) storms tmux with SIGWINCH and the grid thrashes.

## Fix — reflow (mirrors the desktop `applyFit`/`scheduleFit` dedup)
- Split `fitTerminal` into `applyTermFit()` (fit + dedup on last cols/rows, skip 0×0 / detached containers) and a `requestAnimationFrame`-coalesced `fitTerminal()`.
- Observe the `#terminal` container via `ResizeObserver` in addition to window `resize`, so splitter drags / panel collapses that don't fire a window resize still refit — both routes funnel through the coalesced, deduped path.
- `onopen` now fits **synchronously** (and resets the dedup baseline) so the fresh PTY's winsize is set before the scrollback replay (`select-window`), and re-fits once `fonts.ready` settles the Nerd Font cell metrics.

## Fix — Reload terminal (no full page refresh)
- Right-click menu gains **"Reload terminal"**: `term.reset()` to drop the mangled grid, then a clean WebSocket reconnect. The fresh attach re-fits and re-selects the window on open, so crowd replays the pane's tmux scrollback and repaints against a stable layout — the same recovery a page reload gives.

## Scope / safety
- Single file: `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js`. No Swift changes.
- Desktop WKWebView terminal (already fixed in #638) and paste are untouched → no desktop regression.

## Verification
- `node --check app.js` passes.
- Manual: resize the browser window / drag panel splitters — grid no longer thrashes and only real dimension changes reach the PTY; right-click → **Reload terminal** recovers a corrupted surface and replays scrollback without a page refresh.

https://claude.ai/code/session_01UcxzQDxJGCnGbMWBiKexvt

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -2078,6 +2078,15 @@ function ensureTerminal() {
   enableWheelScroll(document.getElementById('terminal'));
   enableFileDrop(document.getElementById('terminal'));
   window.addEventListener('resize', fitTerminal);
+  // Observe the container itself, not just the window: splitter drags / panel
+  // collapses resize the surface without firing a window `resize` (#661). Both
+  // routes funnel through the coalesced, deduped fitTerminal.
+  if (window.ResizeObserver) {
+    new ResizeObserver(fitTerminal).observe(document.getElementById('terminal'));
+  }
+  // The first fit can run with the Menlo fallback before the Nerd Font loads;
+  // once its cell metrics settle the grid recomputes, so re-fit (deduped).
+  if (document.fonts && document.fonts.ready) document.fonts.ready.then(fitTerminal);
   connectTerminalWs();
 }
 
@@ -2201,6 +2210,7 @@ function showTerminalMenu(e) {
   items.push({ label: 'Paste', action: pasteIntoTerminal });
   items.push({ label: 'Select all', action: () => term.selectAll() });
   items.push({ label: 'Clear', action: () => term.clear() });
+  items.push({ label: 'Reload terminal', action: reloadTerminal });
   const menu = el('div', 'ctx-menu');
   for (const it of items) {
     const item = el('div', 'ctx-item', it.label);
@@ -2220,7 +2230,12 @@ function connectTerminalWs() {
   termWs = new WebSocket(wsURL('/terminal'));
   termWs.binaryType = 'arraybuffer';
   termWs.onopen = () => {
-    fitTerminal();
+    // A fresh PTY starts at a default winsize, so force the real size through
+    // even if cols/rows match the previous socket, and fit synchronously so the
+    // resize reaches the PTY before the scrollback replay (select-window) below.
+    lastTermCols = 0;
+    lastTermRows = 0;
+    applyTermFit();
     if (activeTerminal) selectWindow(activeTerminal.window);
   };
   termWs.onmessage = (event) => {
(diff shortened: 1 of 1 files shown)


dhilgaertner (approved):
## Code & Security Review

Single-file, tightly-scoped fix for xterm reflow thrash + a right-click "Reload terminal" recovery path. The reasoning mirrors the already-shipped desktop dedup (#638) and the code holds up under scrutiny.

### Critical Issues
None.

### Security Review
**Strengths:**
- No new network surface or input handling — reuses the existing authenticated `/terminal` WebSocket and the established resize/select-window message shapes.
- `term.reset()` on reload drops the in-memory buffer only; scrollback is re-fetched from the server (tmux), so no client-side trust boundary shifts.
- No secrets, no injection surface, no auth/authz changes.

**Concerns:**
- None.

### Code Quality
- `ensureTerminal` is idempotent (`if (term) return`), so the added `ResizeObserver`, window `resize` listener, and `fonts.ready` handler each register exactly once — no observer/listener leak.
- The `fitTerminal`/`applyTermFit` split is correct: the WS `onopen` path calls `applyTermFit()` synchronously (with `lastTermCols/Rows` reset to 0) so the real winsize reaches the PTY before the `select-window` scrollback replay, while window/observer bursts are coalesced to one fit per animation frame.
- Same-size dedup plus the 0×0/detached-container guard correctly suppress the needless SIGWINCH storm that caused the grid corruption.
- `reloadTerminal` detaches the old socket's `onopen/onmessage/onclose/onerror` before reassigning `termWs`, so the stale `onclose` can't schedule a competing reconnect and `onerror` can't close the new socket — the one real footgun in a reconnect-on-close design, and it's handled. `connectTerminalWs`'s `sessionDead` guard is also respected on the reload path.
- Function declarations are hoisted and the module-level `let` state is initialized well before any `onopen` runs at user-action time — no TDZ risk. `node --check` passes.

Nits (non-blocking, no action required): the `lastTermCols/Rows = 0` reset inside `reloadTerminal` is redundant with the same reset in `onopen`, and the `ResizeObserver` firing on initial observe may emit a benign "ResizeObserver loop" console warning in some browsers — the rAF coalescing prevents any actual loop.

### Summary Table

---

PR #666, opened 2026-07-11
fix(daemon): un-gate scheduled jobs from set-config local-only (#665)

### LINKED ISSUE #665: Web app: saving Jobs config fails with "set-config binaries/jobs is local-only" (feature/crow-593-*)
On the `feature/crow-593-*` web UI, saving the **Jobs** configuration from the web Settings modal fails. The toast shows:

```
Save failed: set-config binaries/jobs is local-only
```

(The message renders twice, concatenated without a separator — `…local-onlySave failed: set-config binaries/jobs is local-only` — likely the same error toast emitted/rendered twice; minor secondary bug, see below.)

## What is happening

This is part of the #593 cross-platform web UI / daemon `get-config`/`set-config` work (disk-fallback for config, originally on `feature/crow-581-settings-app-fallback`). The web Jobs editor persists via the daemon `set-config` RPC, but the `binaries` and `jobs` keys are gated as **local-only** — the RPC refuses to set them and returns `set-config <key> is local-only`. Result: the web Jobs form can’t save at all.

`jobs` (and to a lesser extent `binaries`) are legitimately user-editable config that the Jobs UI exists to manage, so flagging them local-only makes the web Jobs editor non-functional. The desktop app writes them fine because it mutates `{devRoot}/.claude/config.json` directly rather than going through the local-only-gated RPC.

## Likely fix (pick per the intended design)

Two directions — the crow-593 author will know which matches the intended model:

1. **Route local-only keys through the disk-fallback path #593 already added.** When the web app saves `jobs`/`binaries`, persist via the direct `config.json` write (the same disk-fallback used for `get-config`/`set-config` when the desktop app is down/too old) instead of the RPC path that rejects local-only keys. i.e. local-only should mean "written to disk directly," not "un-writable from the web UI."
2. **Loosen / correct the local-only gate.** If `jobs` was swept into the local-only set unintentionally, allow it (and any other genuinely user-editable keys) to be set from the web UI via daemon→app or disk fallback, reserving `local-only` for keys that truly cannot be set remotely (e.g. absolute local `binaries` paths). At minimum `jobs` must be saveable from the web Jobs editor.

Also fix the **duplicated error toast** so a single failure surfaces one message, not two concatenated strings.

## Where to look (on the crow-593 branch — not in `main`)

The `set-config` RPC handler and its local-only key list, the web Settings/Jobs save action, and the daemon disk-fallback for `get-config`/`set-config` all live on `feature/crow-593-*` (none of this is merged to `main`, so it cannot be referenced from `main`). Grep the branch for `local-only`, `set-config`, and the local-only key list containing `jobs`/`binaries`.

## Target branch

The web UI + config RPC live only on `feature/crow-593-*`; base the fix on that branch / its PR and verify from the web Jobs editor. `crow:auto` branches off `main` by default, where this code does not exist — so retarget to the crow-593 branch or route to a human/worktree already on it; a fresh branch off `main` will have nothing to fix.

## Acceptance criteria

- Editing and saving a Job in the web Settings modal persists successfully with no `set-config … is local-only` error.
- The saved `jobs` config round-trips (reload the web app / restart shows the change), consistent with how the desktop app persists it.
- Saving other config from the web UI is unaffected; genuinely local-only keys still behave correctly.
- A single save failure shows exactly one error toast, not a doubled/concatenated message.

dgershman:
Closes #665

## Problem
On the crow-593 web UI, saving a Job from the Settings modal failed with `set-config binaries/jobs is local-only`. The web Jobs editor persists the whole `AppConfig` via the `set-config` RPC, and `RPCWebSocketHandler.localOnlyDenial` rejected that call whenever `jobs` changed from a peer that isn't **local-direct** (loopback + no `X-Forwarded-For`) — i.e. a proxied / LAN-IP session. A localhost browser already worked (local-direct peers skip the gate); a job edit blocked the save everywhere else.

## Fix (team decision: allow jobs from any authenticated session)
Jobs are a core, web-managed surface, so an authenticated remote session may edit them — matching how the desktop app manages jobs. Only `defaults.binaries` (absolute local paths executed at agent launch) remains local-only.

- **`RPCWebSocketHandler`** — narrow the gate to `binaries` only: drop the `jobs` comparison in `setConfigTouchesPrivilegedFields` and remove the `job-*` method cases (those are reached only by the CLI, which is always local over the Unix socket). Denial message is now `set-config binaries is local-only`.
- **`settings.js`** — render the corveil-binary field read-only when the session isn't local-direct (mirrors the web password & AI gateways), so remote users don't trip the still-active binaries gate.
- **`app.js`** — `modalDialog` drops any stray `.text-prompt-backdrop` before showing, so a save failure surfaces exactly one dialog (fixes the doubled/concatenated error toast).
- **`DaemonSecurityTests`** — a jobs change is now allowed remotely; a binaries change stays local-only.

## Security note
This intentionally widens the jobs write surface to any **authenticated** (`/rpc` still requires Origin + web-password auth) remote session, per the team decision — reversing part of CROW-604. The `job-*` RPCs are un-gated for coherence (same jobs-array surface). `binaries` (persistent-RCE via absolute launch paths) stays local-only.

## Verification
- `make daemon` ✓
- `swift test --package-path Packages/CrowDaemon --filter DaemonSecurityTests` → 62 tests pass, incl. `setConfigJobsChangeIsAllowedRemotely` and `setConfigBinariesChangeIsLocalOnly`.
- `node --check` on both edited web scripts ✓

## Acceptance
- [x] Editing & saving a Job in the web Settings modal persists with no `set-config … is local-only` error.
- [x] Saved `jobs` round-trips (set-config returns the merged config).
- [x] Other config saves unaffected; `binaries` stays genuinely local-only.
- [x] A single save failure shows exactly one dialog.

https://claude.ai/code/session_011PemiJvxc8QSWDVsdTrdMz

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
@@ -737,10 +737,11 @@ func makeCommandRouter(
                 "default_dev_root": .string(defaultDevRoot),
             ]
         },
-        // Non-secret settings write. `defaults.binaries` and `jobs` are held to
-        // the same local-direct bar as secret writes — the `/rpc` WebSocket
-        // handler rejects those field changes from non-local peers before this
-        // runs (review Yellow). Unix-socket / CLI callers are always local.
+        // Non-secret settings write. `defaults.binaries` is held to the same
+        // local-direct bar as secret writes — the `/rpc` WebSocket handler rejects
+        // that field change from non-local peers before this runs (review Yellow).
+        // Scheduled `jobs` are NOT gated (CROW-665): an authenticated remote
+        // session may edit them. Unix-socket / CLI callers are always local.
         "set-config": { params in
             guard let json = params["config"]?.stringValue,
                   let data = json.data(using: .utf8),
@@ -818,8 +819,9 @@ func makeCommandRouter(
         // local caller from a logged-in remote one, so it must not carry secret
         // writes — that would let a remote client change the password that gates
         // remote access (CROW-593). The same local-direct bar applies to
-        // `set-config` changes of `defaults.binaries` / `jobs` (gated in
-        // `RPCWebSocketHandler`) — those execute at the next launch.
+        // `set-config` changes of `defaults.binaries` (gated in
+        // `RPCWebSocketHandler`) — those absolute binary paths execute at the next
+        // launch. Scheduled `jobs` are no longer gated (CROW-665).
 
         // Session/board actions — forward-only (need the app's coordinators).
         // Spawning a Manager forwards to the app when it's running (its
--- a/Packages/CrowDaemon/Sources/CrowDaemon/RPCWebSocketHandler.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/RPCWebSocketHandler.swift
@@ -109,27 +109,30 @@ enum RPCWebSocketHandler {
     /// caller from a remote `/rpc` peer (review Yellow / CROW-593).
     ///
     /// - `run-setup`: write+re-exec of an arbitrary `dev_root`.
-    /// - `set-config` when `defaults.binaries` or `jobs` change: those execute at
-    ///   the next agent/job launch (persistent RCE on an unauthenticated
-    ///   non-loopback bind). Other `set-config` fields still flow through.
-    /// - `job-add` / `edit` / `enable` / `disable` / `delete` / `duplicate`: same
-    ///   jobs-array write surface as `set-config` jobs (CROW-604).
+    /// - `set-config` when `defaults.binaries` change: absolute local binary paths
+    ///   that execute at the next agent launch (persistent RCE on an
+    ///   unauthenticated non-loopback bind). Other `set-config` fields flow through.
+    ///
+    /// Scheduled `jobs` are intentionally NOT gated here (CROW-665): the Jobs
+    /// editor is a core web-Settings surface, so an authenticated remote session
+    /// may edit them, matching how the desktop app manages jobs. The `job-*` RPCs
+    /// are likewise un-gated (same jobs-array surface; today only the CLI, which is
+    /// always local, uses them).
     static func localOnlyDenial(for request: JSONRPCRequest, devRoot: String) -> String? {
         switch request.method {
         case "run-setup":
             return "run-setup is local-only"
         case "set-config":
             guard setConfigTouchesPrivilegedFields(request, devRoot: devRoot) else { return nil }
-            return "set-config binaries/jobs is local-only"
-        case "job-add", "job-edit", "job-enable", "job-disable", "job-delete", "job-duplicate":
-            return "\(request.method) is local-only"
+            return "set-config binaries is local-only"
         default:
             return nil
         }
     }
 
     /// True when the incoming `set-config` payload would change agent binary
-    /// overrides or scheduled jobs relative to what's on disk.
+    /// overrides relative to what's on disk. Scheduled `jobs` are no longer
+    /// privileged — an authenticated remote session may edit them (CROW-665).
     static func setConfigTouchesPrivilegedFields(_ request: JSONRPCRequest, devRoot: String) -> Bool {
         guard let json = request.params?["config"]?.stringValue,
               let data = json.data(using: .utf8),
@@ -139,6 +142,5 @@ enum RPCWebSocketHandler {
         }
         let current = ConfigStore.loadConfig(devRoot: devRoot) ?? AppConfig()
         return incoming.defaults.binaries != current.defaults.binaries
-            || incoming.jobs != current.jobs
     }
 }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1535,6 +1535,11 @@ async function quickAction(id, action, label) {
 function modalDialog({ title, body, okLabel = 'OK', cancelLabel = 'Cancel', danger = false } = {}) {
   return new Promise((resolve) => {
     let done = false;
+    // One dialog at a time: drop any stray backdrop still on screen so a
+    // double-fired error can't stack two overlapping cards whose text abuts into
+    // one concatenated message (CROW-665). These dialogs are always awaited
+    // sequentially, so a leftover here is a duplicate, never a live prompt.
+    document.querySelectorAll('.text-prompt-backdrop').forEach((b) => b.remove());
     const backdrop = el('div', 'text-prompt-backdrop');
     const card = el('div', 'text-prompt-card');
     if (title) card.appendChild(el('div', 'text-prompt-title', title));
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1535,12 +1535,6 @@ async function quickAction(id, action, label) {
 function modalDialog({ title, body, okLabel = 'OK', cancelLabel = 'Cancel', danger = false } = {}) {
   return new Promise((resolve) => {
     let done = false;
-    // One dialog at a time: drop any stray *modalDialog* backdrop still on screen
-    // so a double-fired error can't stack two overlapping cards whose text abuts
-    // into one concatenated message (CROW-665). Scoped to modalDialog's own marker
-    // class — `textPrompt` shares `.text-prompt-backdrop` but resolves only via its
-    // own handlers, so removing it here would orphan a live prompt (review Yellow).
-    document.querySelectorAll('.modal-dialog-backdrop').forEach((b) => b.remove());
     const backdrop = el('div', 'text-prompt-backdrop modal-dialog-backdrop');
     const card = el('div', 'text-prompt-card');
     if (title) card.appendChild(el('div', 'text-prompt-title', title));
@@ -1554,6 +1548,7 @@ function modalDialog({ title, body, okLabel = 'OK', cancelLabel = 'Cancel', dang
       backdrop.remove();
       resolve(v);
     }
+    backdrop.__finish = finish;
     function onKey(e) {
       if (e.key === 'Escape') { e.preventDefault(); e.stopPropagation(); finish(false); }
       else if (e.key === 'Enter') { e.preventDefault(); e.stopPropagation(); finish(true); }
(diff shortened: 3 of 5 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Focused change: `set-config` no longer holds scheduled `jobs` to the local-only bar — only `defaults.binaries` remains gated. The `job-*` RPCs are dropped from the gate, the web Settings corveil-binary field goes read-only when not local-direct, and `modalDialog` de-dupes stray backdrops. Reviewed against `feature/crow-593-web-ui-parity`.

### Critical Issues
None.

### Security Review
**Strengths:**
- The genuine persistent-RCE surface (`defaults.binaries` — absolute local paths launched at agent start) **stays** local-only, both server-side (`setConfigTouchesPrivilegedFields`, `RPCWebSocketHandler.swift:145`) and in the UI (read-only binary field, `settings.js:471`). Defense-in-depth is preserved; the JS read-only is cosmetic and the server remains the real gate.
- `/rpc` upgrade still requires Origin + web-password auth (`RPCWebSocketHandler.swift:44`), and `job-*` were CLI/Unix-socket paths.
- Test coverage tracks the intent well: `setConfigJobsChangeIsAllowedRemotely` and `setConfigBinariesChangeIsLocalOnly`. All 62 `DaemonSecurityTests` pass locally; `node --check` clean on both scripts.

**Concerns (Green — consider / confirm, not blocking):**
- A job runs arbitrary prompts through a freshly-launched Claude Code terminal on the host, so job-write is effectively "queue an agent on the host." This is now reachable by a remote peer. That's the *explicit, documented* team decision, and it is no worse than the already-ungated `add-worktree` (which shells out to git). One nuance worth an explicit ack: when **no web password is set**, `WebAuthGuard.authorize` returns authorized for non-loopback peers ("auth disabled", `WebAuth.swift:180`) and `localDirect` is false — so on a passwordless non-loopback bind the removed jobs gate was the last barrier. The PR's "any *authenticated* remote session" framing slightly understates this; `binaries` deliberately keeps protecting that exact config while `jobs` no longer does. Consistent with the existing model, but confirm it's intended for `jobs`.

### Code Quality
**🟡 `app.js:1542` — `modalDialog` cleanup selector is broader than its stated intent and can destroy a live `textPrompt`.**

`document.querySelectorAll('.text-prompt-backdrop').forEach((b) => b.remove())` matches **every** element with that class — but `textPrompt` uses the same `.text-prompt-backdrop` class (`app.js:1585`). The comment scopes the intent to stray *modalDialog* backdrops ("always awaited sequentially, so a leftover is a duplicate"), but `textPrompt` is not part of that sequential contract:

- Reachable path: a user opens a `textPrompt` while an async action is in flight — e.g. Cmd+F "Find in terminal" (fire-and-forget, `app.js:2159`) or `renameSession` (`app.js:1621`) — and an in-flight action then rejects into a `catch` that calls `alertModal` (e.g. `app.js:1243` switch-agent, `1648` delete, `2097` promote). `modalDialog` then removes the live prompt's backdrop from the DOM.
- Because `textPrompt` only resolves via its button/keydown handlers (`app.js:1599`), external `.remove()` does **not** call `finish()`: the prompt silently vanishes mid-type and its Promise is orphaned (a `renameSession` `await` hangs indefinitely).

Suggested fix: scope the cleanup to modalDialog's own backdrops (give `modalDialog` a distinct marker class and query that), rather than the shared `.text-prompt-backdrop`.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 1 Green] findings. The jobs un-gating itself is sound and well-tested; the blocker is the over-broad `modalDialog` backdrop cleanup that can silently remove a live text prompt.

---

dgershman:
Thanks for the review. Both findings addressed:

**🟡 modalDialog backdrop cleanup (fixed — 57476a4)**
Correct catch — the dedup queried the shared `.text-prompt-backdrop`, so an `alertModal` firing while a `textPrompt` (rename / find-in-terminal) was open would remove the live prompt's backdrop and orphan its Promise (the `await` hangs). `modalDialog`'s own backdrop now carries a distinct `modal-dialog-backdrop` marker and the cleanup targets only that class, so `textPrompt` is never touched.

**🟢 Passwordless non-loopback bind (confirmed intended, no code change)**
Acknowledged, and the framing correction is fair: with no web password set, `WebAuthGuard.authorize` returns authorized for non-loopback peers ("auth disabled") and `localDirect` is false, so on a passwordless non-loopback bind the removed jobs gate was the last barrier. Un-gating `jobs` there is the explicit team decision for this ticket (jobs are a first-class web-managed surface, on par with the already-ungated `add-worktree` git shell-out). `binaries` deliberately keeps protecting that exact passwordless config since absolute launch paths are the persistent-RCE surface; `jobs` no longer does, by intent.

dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- The privileged-field gate is correctly narrowed: `setConfigTouchesPrivilegedFields` now compares only `defaults.binaries` (RPCWebSocketHandler.swift:144), and `binaries` — absolute local paths executed directly at agent launch — stays local-only. `/rpc` still enforces Origin + web-password auth before any of this runs, so the widened jobs surface is authenticated-only.
- The `job-*` un-gating is coherent with the decision and I verified the comment's claim: no web script calls `job-add/edit/enable/disable/delete/duplicate` — the web Jobs editor persists via `set-config` (settings.js:143), and `job-*` is reached only by the CLI over the Unix socket (always local). So removing those cases changes no remote surface today.
- `settings.js` renders the corveil-binary field read-only when `!isLocal` (settings.js:471-473); with `readonly` no `oninput` handler is wired (settings.js:256-261), so `cfg.defaults.binaries.corveil` can't be mutated remotely and a remote save can't trip the still-active binaries gate. Mirrors the existing web-password / AI-gateway pattern.
- `DaemonSecurityTests` updated to assert the new invariant (jobs allowed remotely, binaries still local-only); full suite passes locally: **62 tests, 19 suites**.

**Consideration (Green):**
- The PR frames `binaries` as the RCE surface and jobs as lower-risk, but a job's `prompts` launch Claude Code, and with `jobsAutoPermissionMode` defaulting to `true` those prompts can run commands without confirmation — so remote job editing is also an authenticated-RCE path, not merely a config edit. This is the documented team decision (auth-gated, desktop parity), so not a blocker; flagging only so the risk framing in the security note isn't read as "jobs are harmless."

### Code Quality

**Yellow — modalDialog supersede leaks the old dialog's promise + keydown listener (app.js:1543)**
The dedup removes the stray backdrop's DOM node but never calls that dialog's `finish()`. Two consequences for the superseded dialog:
1. Its `Promise` never resolves — the awaiting `await alertModal(...)` / `confirmModal(...)` continuation hangs (or later resolves at a surprising time).
2. Its capture-phase `document` `keydown` listener (added app.js:1571, removed only inside `finish()` app.js:1553) stays attached. The next global Enter/Escape after the doubled dialog is dismissed by mouse fires the dead dialog's `onKey`, which calls `preventDefault()` + `stopPropagation()` — silently swallowing that one keystroke — before removing itself.

This isn't jobs-specific: any two near-simultaneous modal calls (two errors, or an error while a confirm is open) hit it. Suggested fix — supersede by finishing, not just removing:

```js
const backdrop = el('div', 'text-prompt-backdrop modal-dialog-backdrop');
// …
backdrop.__finish = finish;               // after finish is defined
// dedup:
document.querySelectorAll('.modal-dialog-backdrop')
  .forEach((b) => (b.__finish ? b.__finish(false) : b.remove()));
```

This resolves the old promise (as a cancel) and detaches its listener, keeping the one-dialog-at-a-time behavior without the leak.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 1 Green] findings. The daemon-side gate change is correct and well-tested; the one Yellow is a contained leak in the `modalDialog` dedup added here.

---

dgershman:
Thanks — good catch on the leak. Addressed in 175d58e.

**🟡 modalDialog supersede leaks promise + keydown listener (fixed)**
Went with your suggested shape: `finish` is attached to the backdrop (`backdrop.__finish = finish`), and the dedup calls the superseded dialog's `__finish(false)` (resolve as cancel) instead of a bare `.remove()`. That settles the old Promise and detaches its capture-phase `keydown` listener, so no more hung `await` and no swallowed Enter/Escape. The dedup runs *before* the new backdrop is appended, so it only matches prior dialogs, and it stays scoped to the `modal-dialog-backdrop` marker so a live `textPrompt` is never touched. `node --check` clean.

**🟢 jobs are also an authenticated-RCE path (agreed, framing note)**
Fair — with `jobsAutoPermissionMode` defaulting to `true`, a job's prompts launch Claude Code that can run commands without confirmation, so remote job-write is an authenticated-RCE path, not merely a config edit. That's the documented team decision (auth-gated, desktop parity, on par with the already-ungated `add-worktree`); I shouldn't have framed jobs as "harmless." `binaries` stays local-only precisely because absolute launch paths are the un-auth'd-passwordless-bind RCE surface, and `jobs` no longer does, by intent.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Context:** This PR narrows the `/rpc` local-only gate from `binaries` + `jobs` to `binaries` only, so an authenticated remote session may now edit scheduled `jobs` (reversing part of CROW-604). This is an explicit, documented team decision, gated behind `/rpc` auth (Origin + web-password), with `defaults.binaries` (absolute launch paths → persistent RCE) correctly kept local-only.

**Strengths:**
- The privileged-field check is correctly narrowed: `setConfigTouchesPrivilegedFields` now compares only `defaults.binaries`, and the `job-*` RPC cases were removed from `localOnlyDenial`. Verified those RPCs are CLI-only (`JobCommands.swift`) and travel over the always-local Unix socket — the web UI persists jobs via `set-config`, so removing the WebSocket cases changes no real exposure.
- `settings.js` renders the corveil-binary field read-only when `!isLocal`, mirroring the web-password/AI-gateway pattern, so remote users don't trip the still-active binaries gate. `isLocal` is sourced from `/auth/context` and used consistently.
- Denial message, tests, and inline comments were all updated coherently; `DaemonSecurityTests` now asserts jobs-remote-allowed and binaries-local-only.

**Consideration (Green — awareness, not blocking):**
- Scheduled jobs fire Claude Code with arbitrary `prompts`, and `jobsAutoPermissionMode` defaults to `true` (`AppConfig.swift:103`). The net effect is that an authenticated remote session can schedule prompts that execute on the host in auto-permission mode. This is the acknowledged intent of the PR and is behind auth, but worth keeping on record — if the threat model ever tightens, revisit whether the auto-permission default should apply to remotely-authored jobs.

### Code Quality
- `app.js` `modalDialog` supersede logic is sound: it resolves prior dialogs via `__finish(false)` (settling the Promise and detaching the capture-phase keydown listener) rather than bare-removing, and is scoped to the `.modal-dialog-backdrop` marker class so a live `textPrompt` (which shares `.text-prompt-backdrop`) is untouched. The query runs before the new backdrop is appended, so it matches only prior dialogs. Both prior review-round fixes are correctly applied.
- No logic errors, resource leaks, or race conditions found.

### Verification
- `node --check` on `app.js` and `settings.js` ✓
- `swift test --filter DaemonSecurityTests` → **62 tests pass**, including `setConfigJobsChangeIsAllowedRemotely` and `setConfigBinariesChangeIsLocalOnly` ✓

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 1 Green] findings. The diff faithfully and correctly implements a documented team decision, is well-tested, and preserves the genuine RCE surface (`binaries`) as local-only.

---

PR #669, opened 2026-07-12
refactor(terminal): share jump-to-bottom as an xterm.js addon (desktop + web)

### LINKED ISSUE #668: Jump-to-bottom: make it a shared xterm.js addon for desktop + web (replace desktop-only #635)
The jump-to-bottom control (#635, on `feature/crow-593-web-ui-parity` as commit `e902911`) only exists in the **desktop** terminal page, so it is missing from the **web** UI terminal. Rather than copy it into the web front-end, implement it once as a **custom xterm.js addon** loaded by both surfaces, and use it to replace the current desktop-only implementation.

## Why it is missing on web
- #635 added the button + scroll wiring to `Packages/CrowTerminal/Sources/CrowTerminal/Resources/xterm/terminal.html` (`<button id="crow-jump-bottom">` at :55, wired at :120-143). That page is what the macOS WKWebView loads.
- The web UI builds its **own** xterm terminal in `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` (`new Terminal(...)` in `ensureTerminal`, ~:2131) whose config is "mirrored by hand" from `terminal.html` (comment ~:2130), plus its own `web/terminal.html` (the "WebSocket transport variant"). The #635 control was never ported there — `crow-jump-bottom`/`scrollToBottom` appear **zero** times anywhere under `web/`.

This is a symptom of three hand-mirrored terminal front-ends (`CrowTerminal/terminal.html`, `web/terminal.html`, `web/app.js`) with duplicated Terminal config — the same reason reflow-debounce (#661) and ticket multi-select (#662) each had to be redone for web.

## Recommended design: a custom xterm.js addon
Both surfaces already use the addon pattern (`term.loadAddon(...)`) and both load vendored addons via `<script src>` — desktop from the bundle, web from the daemon `/xterm/...` route (`web/index.html:36-41`). Jump-to-bottom is pure terminal + DOM with **no transport coupling** (only needs the `Terminal`, its element, `onScroll`, an at-bottom check via `buffer.active.viewportY`/`baseY`, and `scrollToBottom()`), so it fits `ITerminalAddon` cleanly.

1. Add one file, e.g. `xterm-addon-crow-jumpbottom.js`, to the shared vendored xterm dir (`CrowTerminal/.../Resources/xterm/`), which the daemon already serves to web at `/xterm/…`. Follow the existing UMD/global shape (like `FitAddon` exposing `window.FitAddon`) since it loads via `<script src>`, not ES modules.
2. Implement as `ITerminalAddon`:
   - `activate(term)` — inject the button into `term.element`s container (position relative to the viewport, **not** `document.body`, since desktop is full-page and web is a tabbed container), wire `term.onScroll`, toggle visibility on at-bottom, click → `term.scrollToBottom()`.
   - `dispose()` — remove the button + listeners. Lifecycle matters: the web UI creates/switches/closes terminals (`ensureTerminal`/`switchTerminal`/`closeTerminal`), so a plain shared function would leak DOM/handlers.
3. `<script src=".../xterm-addon-crow-jumpbottom.js">` in all three pages and `term.loadAddon(new CrowJumpBottom())` at each `new Terminal(...)`.
4. **Delete** the inline #635 button/CSS/wiring from `CrowTerminal/.../terminal.html` and load the addon instead — one source of truth, so it can not drift again.

Explicitly NOT: forking vendored `xterm.js`, or unifying the whole page (transport genuinely differs — WKWebView `postMessage` vs WebSocket; `web/terminal.html` calls itself "the WebSocket transport variant"). The transport split is the correct seam; everything above it should be shared.

## Follow-on (note in ticket, do not implement here)
Make this the first of a shared "Crow terminal behaviors" addon set; subsequently migrate fit-coalescing (#638), reflow-debounce (#661), and paste-through into the same shared layer so per-surface parity drift stops. Track separately.

## Target branch
The web terminal + `#635` live only on `feature/crow-593-web-ui-parity`; base this on that branch / its PR and verify in **both** the desktop app and the web UI. `crow:auto` branches off `main` by default (where the web UI does not exist), so retarget to the crow-593 branch or route to a worktree already on it.

## Acceptance criteria
- A single addon file implements jump-to-bottom; the inline #635 implementation in `CrowTerminal/.../terminal.html` is removed in favor of loading the addon.
- The control appears and works identically in the **desktop** app and the **web** UI: hidden while at the bottom, shown when scrolled up, click snaps to latest output.
- On web, switching/closing terminals does not leak duplicate buttons or scroll listeners (addon `dispose()` runs).
- No regression to fit/resize, search, paste, or reflow behavior on either surface.

dgershman:
Closes #668

## What

Jump-to-bottom (#635) lived only in the **desktop** terminal page, so it was missing from the **web** UI. Rather than copy it, implement it once as a custom **xterm.js addon** and load it from both surfaces — deleting the desktop-only #635 implementation so there's one source of truth that can't drift again.

## Changes

- **New addon** `Packages/CrowTerminal/.../Resources/xterm/xterm-addon-crow-jumpbottom.js` (`ITerminalAddon`):
  - `activate(term)` injects a `.crow-jump-bottom` button into the terminal's **container** (anchored to the viewport, not `document.body`, since web is a tabbed container), wires `onScroll`/`onRender` to toggle visibility via the `buffer.active.viewportY >= baseY` at-bottom check, and snaps to latest on click.
  - `dispose()` removes the button + listeners so the web UI's create/switch/close lifecycle never leaks duplicate buttons or handlers.
  - Namespaced UMD global matching the vendored addons (`window.CrowJumpBottomAddon.CrowJumpBottomAddon`).
- **Loaded from all three front-ends** (script tag + `loadAddon` after `term.open`): desktop `terminal.html`, `web/index.html` + `app.js`, and the `web/terminal.html` debug page. The daemon already serves the file to web at `/xterm/…` (dynamic `/xterm/:file` route, `.copy("Resources/xterm")` bundling — no route/Swift change needed).
- **Deleted** the inline #635 button/CSS/wiring from desktop `terminal.html`.
- **Retargeted** `BundledResourcesTests`: pin that `terminal.html` loads the addon and the inline copy is gone, and that the addon file is bundled with its API surface.

## Not in scope (follow-on, tracked separately)
Forking vendored xterm or unifying the whole page across transports (WKWebView `postMessage` vs WebSocket — the correct seam). Making this the first of a shared "Crow terminal behaviors" addon set and migrating fit-coalescing (#638), reflow-debounce (#661), paste-through into it.

## Base branch
Targets `feature/crow-593-web-ui-parity` — the web UI + #635 only exist there, not on `main`.

## Testing
- `make build` — daemon + app compile.
- CrowTerminal `BundledResourcesTests` pass, including the two retargeted assertions (addon bundled + terminal.html loads it, inline copy removed).
- Manual verification of the pill (appears when scrolled up, hides at bottom, click snaps to latest; no fit/search/paste regressions; no leaked buttons on tab switch/close) should be done in both the desktop app and web UI.

https://claude.ai/code/session_01LwDCNHfWc7TVmS58y2Vmn4

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -2141,6 +2141,9 @@ function ensureTerminal() {
   term.loadAddon(searchAddon);
   term.loadAddon(webLinksAddon);
   term.open(document.getElementById('terminal'));
+  // Jump-to-bottom pill (#668), shared with the desktop surface. Must load after
+  // open() so the addon can anchor its button to the terminal's container.
+  term.loadAddon(new CrowJumpBottomAddon.CrowJumpBottomAddon());
   // WebGL renderer for throughput; must load after open(). Falls back to the
   // default renderer if the GL context is unavailable or gets lost.
   try {
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/index.html
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/index.html
@@ -39,6 +39,7 @@
   <script src="/xterm/xterm-addon-search.js"></script>
   <script src="/xterm/xterm-addon-web-links.js"></script>
   <script src="/xterm/xterm-addon-webgl.js"></script>
+  <script src="/xterm/xterm-addon-crow-jumpbottom.js"></script>
   <script src="/app.js"></script>
   <script src="/settings.js"></script>
 </body>
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/terminal.html
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/terminal.html
@@ -39,6 +39,7 @@
   <script src="/xterm/xterm.js"></script>
   <script src="/xterm/xterm-addon-fit.js"></script>
   <script src="/xterm/xterm-addon-image.js"></script>
+  <script src="/xterm/xterm-addon-crow-jumpbottom.js"></script>
   <script>
     // WebSocket transport variant of CrowTerminal's terminal.html. The xterm.js
     // config block is copied verbatim so font/scrollback/inline-image behavior
@@ -90,6 +91,9 @@
       term.parser.registerCsiHandler({ prefix: '?', final: 'l' }, swallowMouseMode);
 
       term.open(document.getElementById('terminal'));
+      // Jump-to-bottom pill (#668) — load after open() so it can anchor its
+      // button to the terminal container.
+      term.loadAddon(new CrowJumpBottomAddon.CrowJumpBottomAddon());
 
       const proto = window.location.protocol === 'https:' ? 'wss' : 'ws';
       const ws = new WebSocket(proto + '://' + window.location.host + '/terminal');
--- a/Packages/CrowTerminal/Sources/CrowTerminal/Resources/xterm/terminal.html
+++ b/Packages/CrowTerminal/Sources/CrowTerminal/Resources/xterm/terminal.html
@@ -17,45 +17,14 @@
       width: 100%;
       height: 100%;
     }
-    #crow-jump-bottom {
-      position: absolute;
-      right: 16px;
-      bottom: 16px;
-      width: 34px;
-      height: 34px;
-      padding: 0;
-      display: flex;
-      align-items: center;
-      justify-content: center;
-      font-size: 20px;
-      line-height: 1;
-      color: #ddc482;
-      background: #22262a;
-      border: 1px solid rgba(221, 196, 130, 0.35);
-      border-radius: 17px;
-      box-shadow: 0 2px 8px rgba(0, 0, 0, 0.5);
-      cursor: pointer;
-      z-index: 10;
-      opacity: 1;
-      transition: opacity 0.12s ease;
-    }
-    #crow-jump-bottom:hover {
-      background: #2a2f34;
-      border-color: rgba(221, 196, 130, 0.6);
-    }
-    #crow-jump-bottom.hidden {
-      opacity: 0;
-      visibility: hidden;
-      pointer-events: none;
-    }
   </style>
 </head>
 <body>
   <div id="terminal"></div>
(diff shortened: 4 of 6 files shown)


dhilgaertner (approved):
## Code & Security Review

Clean, well-scoped refactor: the desktop-only #635 jump-to-bottom control becomes a single shared xterm.js addon loaded by both surfaces, deleting the inline copy so it can't drift. Verified end-to-end, not just read.

### Critical Issues
None.

### Security Review
**Strengths:**
- New asset is served by the existing `/xterm/:file` route (`StaticAssets.swift:38`) out of CrowTerminal's bundle — no new route, no new attack surface, path-traversal guard (`isSafeAssetName`) unchanged.
- CSP compatibility holds: the same-origin `<script src="/xterm/…">` is covered by `script-src 'self'`, and the addon's programmatically-injected `<style>` is covered by `style-src 'self' 'unsafe-inline'` (`StaticAssets.swift:113`). No inline `<script>`, no eval, no external fetch.
- Button content is a static glyph via a class (`.crow-jump-bottom`); no untrusted/provider strings flow into the injected DOM.

**Concerns:**
- None.

### Code Quality
- `activate()` correctly anchors to `term.element.parentElement` (the container passed to `open()`), not `document.body` — right call for the web's tabbed container; guards the null-container case.
- `dispose()` is complete: tears down click listener, `onScroll`/`onRender` disposables, the button, and reverts the `position: relative` mutation it made (only when it set it). No leaked handlers/buttons across the web create/switch/close lifecycle.
- Old inline #635 control fully removed from desktop `terminal.html`; no dangling `#crow-jump-bottom` / `updateJumpButton` references remain anywhere.
- Tests retargeted sensibly and both pass — one pins the host page loads+references the addon and the inline copy is gone, one pins the addon file is bundled with its load-bearing API surface. Ran `swift test --filter BundledResourcesTests`: 10/10 green.

### Consider (Green, non-blocking)
- `app.js` / `index.html` call `new CrowJumpBottomAddon.CrowJumpBottomAddon()` without the `constructor missing` guard that the desktop `terminal.html` uses. Consistent with the sibling addons (Fit/Image/Search) loaded the same way, so acceptable — just noting the asymmetry.
- The web `ensureTerminal()` is a singleton (`if (term) return;`), so `dispose()`'s leak-prevention is largely defensive today rather than exercised — good to keep for the eventual multi-terminal path the comments anticipate.

### Summary Table

---

PR #670, opened 2026-07-12
fix(web-terminal): focus-gated tmux sizing across surfaces (#667)

### LINKED ISSUE #667: web terminal: multi-client tmux size tug-of-war corrupts the display (focus-gate sizing)
> **⚠️ Base branch: `feature/crow-593-web-ui-parity`** — do NOT branch off `main`.
> `crow:auto` branches off `main` by default; this work must be based on the
> crow-593 parity branch and the PR must target it (`--base feature/crow-593-web-ui-parity`),
> so the web build can be validated. The affected file (`app.js`) lives on the parity line.

## Summary
Follow-up to #661 / #663. #663 debounced each web client's **own** resize storm and added a right-click "Reload terminal" — necessary but not sufficient. When more than one surface is attached (Tauri desktop, browser, phone), they **fight over the shared tmux window size**, and the non-winning surface renders mismatched-width content → the overlapping/garbled rows from #661. A full page reload only "fixes" it by momentarily being the sole/latest client.

## Root cause (confirmed on the live cockpit, tmux 3.6a)
- All surfaces share **one** cockpit (`crow-cockpit`). Every `/terminal` connection opens a **grouped** session (`crowd-web-*`) that shares the same windows — so window **size is shared**; only the current-window pointer is independent.
- The server runs tmux defaults: **`window-size latest`** + **`aggressive-resize off`**. A shared window can only be one size, and "latest" means *the most recent client to touch it wins* — **with no notion of focus**. An idle phone or a background tab grabs the size the moment it attaches or relayouts, reflowing the window for the surface you're actually using.
- Evidence from the running server: two surfaces attached at **different** sizes (179×54 and 126×43), and windows frozen at a patchwork of sizes (179×54 / 194×50 / 126×43) left by clients that have since detached.

All three of a user's surfaces run the **web UI `app.js`** (Tauri desktop uses `frontendDist: ../ui`; browser + phone are plain web clients). The native WKWebView `CrowTerminal/.../terminal.html` is orphaned post-ADR-0008, and the daemon `/terminal.html` is debug-only.

## Proposed fix — focus-gated sizing ("only the focused surface owns the size")
tmux has no focus signal, so gate at the app layer.

**1. `app.js` (primary)** — builds on the `applyTermFit`/`fitTerminal`/`lastTermCols` helpers from #663:
- In `applyTermFit()`, early-return without fitting or sending a resize frame when `document.hidden || !document.hasFocus()` — a backgrounded surface never becomes tmux's "latest" and can't steal the shared size.
- Add `takeTerminalOwnership()` (reset `lastTermCols/lastTermRows`, then `applyTermFit()`) wired to `window` `focus` and `document` `visibilitychange` → on regaining focus the surface re-asserts its own size even if unchanged.
- With `window-size latest` unchanged, "latest" now converges to "the surface you most recently focused."

**2. `crow-tmux.conf` (complement)** — add `set -gw aggressive-resize on` so a window is sized only from clients for which it is the *current* window (a background surface parked on a different window can't drag another window's size). Takes effect after the server re-sources the conf / on reconnect.

**3. (optional) Reap leaked grouped sessions** — `TerminalCockpit.swift`: the live server has ~18 stale detached `crowd-web-*` sessions from crowd restarts / abnormal PTY exits that bypass the `defer closeViewSession`. Conservatively kill detached `crowd-web-*` groups at startup.

## Acceptance criteria
- With two surfaces on the **same** terminal at different sizes, only the **focused** surface drives the shared window size; switching focus transfers ownership cleanly; the background surface shows a stable (possibly clipped) view with no overlapping/garbled rows.
- `tmux show-options -gw aggressive-resize` → `on`; clients on different windows don't drag each other's size.
- No regression to #663's debounce or "Reload terminal".

## Notes
- Scope is `app.js` + `crow-tmux.conf` (+ optional `TerminalCockpit.swift`).
- Inherent limit: one shared tmux window has exactly one size, so a background device can't be pixel-perfect while a differently-sized surface is focused — but it will be stable and non-corrupting, and correct once focused.

Related: #661, #663.

dgershman:
Closes #667

Follow-up to #661 / #663. #663 debounced each web client's *own* resize storm; this addresses the *cross-surface* fight over the shared tmux window size.

## Root cause
Every web surface (Tauri desktop, browser, phone) opens a **grouped** tmux session (`crowd-web-*`) off the one `crow-cockpit`. Grouped sessions share the window list — so a window's **size** is shared; only the current-window *pointer* is independent. With tmux defaults `window-size latest` + `aggressive-resize off`, the **most recent client to touch a window wins its size, with no notion of focus**. An idle phone or background tab grabs the size the moment it attaches/relayouts → the focused surface renders at the wrong width → the overlapping/garbled rows from #661. A page reload only "fixes" it by momentarily being the sole/latest client.

## Fix (tmux has no focus signal → gate at the app layer)
- **`app.js`** — `applyTermFit()` early-returns when `document.hidden || !document.hasFocus()`, so a backgrounded surface never becomes tmux's "latest" and can't steal the size. New `takeTerminalOwnership()` (wired to `window` `focus` + document `visibilitychange`) re-asserts this surface's size on refocus even if unchanged, so "latest" converges to *"the surface you most recently focused."* Background surfaces show a stable, clipped, non-corrupting view.
- **`crow-tmux.conf`** — `set -gw aggressive-resize on` so a surface parked on a *different* window can't drag this window's size.
- **`TerminalCockpit.swift`** — reap **detached** `crowd-web-*` groups at crowd startup (leaked when a prior crowd died before its `defer closeViewSession` ran). `session_attached == 0` filter never kills a live browser's view.

No regression to #663's debounce or right-click "Reload terminal".

## Acceptance
- Two surfaces on the same terminal at different sizes → only the **focused** surface drives the shared window size; switching focus transfers ownership cleanly; the background surface is stable (possibly clipped) with no garbled rows.
- `tmux show-options -gw aggressive-resize` → `on`.

## Notes
Inherent limit (unchanged): one shared tmux window has exactly one size, so a background device can't be pixel-perfect while a differently-sized surface is focused — but it's stable, non-corrupting, and correct once focused.

https://claude.ai/code/session_019vtUHdzP8ZE1Nbi46JdC8m

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -2174,6 +2174,14 @@ function ensureTerminal() {
   enableWheelScroll(document.getElementById('terminal'));
   enableFileDrop(document.getElementById('terminal'));
   window.addEventListener('resize', fitTerminal);
+  // #667: on regaining focus/visibility, this surface reclaims ownership of the
+  // shared tmux window size (re-asserts even if unchanged) — so "window-size
+  // latest" converges to "the surface you most recently focused." Registered
+  // once (ensureTerminal is `if (term) return` guarded).
+  window.addEventListener('focus', takeTerminalOwnership);
+  document.addEventListener('visibilitychange', () => {
+    if (!document.hidden) takeTerminalOwnership();
+  });
   // Observe the container itself, not just the window: splitter drags / panel
   // collapses resize the surface without firing a window `resize` (#661). Both
   // routes funnel through the coalesced, deduped fitTerminal.
@@ -2358,6 +2366,14 @@ let fitScheduled = false;
 
 function applyTermFit() {
   if (!term || !fitAddon) return;
+  // #667: a backgrounded surface must not become tmux's "latest" client and
+  // steal the shared window size. Multiple surfaces (Tauri desktop, browser,
+  // phone) share one grouped-session window whose size is `window-size latest`,
+  // so an idle tab that fit()s + resizes on relayout would reflow the window for
+  // the surface actually in use → the garbled rows from #661. Gate here: only
+  // the focused/visible surface owns the size. It re-asserts on regaining focus
+  // via takeTerminalOwnership() (window focus / visibilitychange).
+  if (document.hidden || !document.hasFocus()) return;
   const node = document.getElementById('terminal');
   if (!node || !node.isConnected || node.clientWidth < 1 || node.clientHeight < 1) return;
   try { fitAddon.fit(); } catch (_) { return; }
@@ -2379,6 +2395,18 @@ function fitTerminal() {
   if (fitScheduled) return;
   fitScheduled = true;
   requestAnimationFrame(() => { fitScheduled = false; applyTermFit(); });
+}
+
+// #667: on regaining focus, re-assert this surface's size so it becomes tmux's
+// "latest" client and reclaims the shared window size from whatever background
+// surface last touched it — even if this surface's own grid didn't change (which
+// applyTermFit's lastTermCols/lastTermRows dedup would otherwise swallow). Reset
+// the dedup, then fit; the reset + the visible/focused state let applyTermFit
+// send the resize frame. Wired to window `focus` and document `visibilitychange`.
+function takeTerminalOwnership() {
+  lastTermCols = 0;
+  lastTermRows = 0;
+  applyTermFit();
 }
 
 function selectWindow(win) {
--- a/Packages/CrowDaemon/Sources/CrowDaemon/TerminalCockpit.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/TerminalCockpit.swift
@@ -28,9 +28,40 @@ struct TerminalCockpit: Sendable {
     /// one (default shell anchor) so the daemon works standalone too. A failed
     /// create surfaces as an attach error in the browser rather than aborting.
     private func ensureSession() {
-        guard !controller.hasSession() else { return }
-        let conf = BundledResources.tmuxConfURL?.path
-        try? controller.newSessionDetached(configPath: conf, env: [:], command: nil)
+        if !controller.hasSession() {
+            let conf = BundledResources.tmuxConfURL?.path
+            try? controller.newSessionDetached(configPath: conf, env: [:], command: nil)
+        }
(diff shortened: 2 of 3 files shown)


dhilgaertner (approved):
## Code & Security Review

Focused, well-reasoned fix for the cross-surface tmux size fight. The three layers (app-layer focus gate, `aggressive-resize on`, orphaned-view reaping) are coherent and the inline comments are unusually thorough. `swift build` on CrowDaemon passes (only a pre-existing unrelated warning in RPCWebSocketHandler.swift).

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface. `reapOrphanedViewSessions` only ever issues `kill-session` against names it verified carry the `crowd-web-` prefix, so it can never take down `crow-cockpit` or the desktop app's windows.
- No untrusted input flows into the tmux args; the format string is fixed and the session names are self-generated UUIDs.

**Concerns:**
- None.

### Code Quality
Solid overall. A few Green (consider) notes — none blocking:

- **Concurrent-crowd reap race (Green).** `TerminalCockpit.swift:54` — the safety comment claims an in-use view owned by a *concurrent* crowd is always `attached >= 1` and skipped. That holds once attached, but there is a brief window between `openViewSession`'s `new-session -d` (`TerminalCockpit.swift:71`) and the browser PTY running `attach-session`, during which the session is `attached == 0`. If a second crowd starts up in exactly that window it could reap the other's in-flight view, causing that browser's attach to fail. This is startup-only and extremely narrow, so low priority — but the comment slightly overstates the guarantee for the multi-crowd case it explicitly calls out.

- **First-size / reconnect while unfocused (Green).** With the `applyTermFit` gate, a surface that opens or reconnects its WebSocket while backgrounded won't send an initial resize until it regains focus, leaving the PTY at the default winsize meanwhile. This is by design and self-heals via `takeTerminalOwnership` on `focus`/`visibilitychange` — noting it only so the clipped-until-focus behavior is expected.

- **`document.hasFocus()` on the Tauri desktop wrapper (Green).** The gate relies on `document.hasFocus()` reporting correctly for whichever surfaces load `app.js`. Standard/reliable in browsers; worth a quick manual confirm on the Tauri desktop surface that a focused window drives the size as expected (it should recover on the `focus` event regardless).

- **No automated coverage (Green).** The reap parsing is inline around a live tmux call, so it's hard to unit-test as written; acceptable for the scope, but extracting the line-parse into a pure helper (like `replayFrame`) would make it testable.

### Summary Table

---

PR #672, opened 2026-07-12
fix(web-terminal): tab switch self-heals display like Reload (#671)

### LINKED ISSUE #671: fix(web-terminal): tab switch should recover the display like Reload, snappily (#667 follow-up)
Follow-up to #667 / #670. Right-click "Reload terminal" recovers a mismatched/corrupted surface, but switching session tabs does not — and reload is heavy. Make a tab switch self-heal the display the cheap way.

## Base branch
`feature/crow-593-web-ui-parity` — `app.js` lives on the parity line; branch off it and target it with the PR (`--base feature/crow-593-web-ui-parity`), same as #670.

## Root cause
- `switchTerminal()` (`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js`) → `selectWindow()` sends only `{type:'select-window'}`. `TerminalWebSocket.swift` selects the window and replays the pane **captured at the window's current size**, with no `term.reset()` and no size re-assert. `term.focus()` does not fire the `window` `focus` event, so `takeTerminalOwnership()` never runs → content laid out for the wrong width is painted into this grid → still garbled; a stuck-small window never grows back.
- `reloadTerminal()` → fresh socket → `onopen` does `term.reset()` + forced `applyTermFit()` (reclaims this surface's size) + `selectWindow()`, in that order (**resize THEN replay**). That forced resize is what fixes it; the full socket + PTY + `tmux attach` + 50k scrollback replay teardown is why it isn't snappy.

## Proposed fix (app.js only)
In `switchTerminal()`, call the existing `takeTerminalOwnership()` (resets `lastTermCols`/`lastTermRows`, refits, sends the resize frame) **before** `selectWindow(t.window)`, mirroring reload's onopen order — reclaim the shared window size for this surface, then let the replay come back at the corrected size. No socket teardown. Optionally `term.reset()` first for a fully clean slate (the replay already prepends `ESC[H ESC[2J ESC[3J`, so likely unnecessary).

## Acceptance
- After another surface has resized the shared window, switching tabs on this surface recovers the display (correct size, no overlapping/garbled rows) **without** a full reload — matching what right-click "Reload terminal" does, but instant.
- Also fixes "stays small after I close the phone tab": switching re-asserts this surface's size.
- No regression to right-click "Reload terminal" or #663's resize debounce.

## Out of scope
Stopping a concurrently-foreground small surface (phone) from re-stealing the size while both are visible — that's `window-size latest`→`largest`, a separate follow-up. This issue is recovery-on-switch only.

Related: #661, #663, #667, #670.

dgershman:
Closes #671. Follow-up to #667/#670.

## Problem
The web terminal shares one grouped-session tmux window (`window-size latest`) across surfaces (Tauri desktop, browser, phone). When another surface resizes that window, this surface's pane is laid out for the wrong width → garbled/overlapping rows (the #661 corruption), or stays stuck-small after the small surface goes away.

Right-click **Reload terminal** recovers it, but it's heavy (tears down the WebSocket + PTY + `tmux attach` + replays ~50k lines of scrollback). Switching session tabs recovered *nothing* — `switchTerminal()` only sent `select-window`, replaying the pane at the window's current (wrong) size with no size re-assert. `term.focus()` there doesn't fire the `window` `focus` event, so `takeTerminalOwnership()` never ran.

## Fix (app.js only)
Reorder `switchTerminal()` to mirror `reloadTerminal`'s `onopen` ordering, reusing the existing `takeTerminalOwnership()` primitive (`lastTermCols=0; lastTermRows=0; applyTermFit()`):

- **resize THEN replay** — call `takeTerminalOwnership()` (forces this surface's real size onto the shared window) **before** `selectWindow(t.window)`, so the replay comes back at the corrected size.
- No socket/PTY/scrollback teardown → instant.

`takeTerminalOwnership()` already exists (used by the window `focus` / `visibilitychange` listeners); its `applyTermFit()` guard checks *document* focus, which is true when clicking a tab in a visible browser, so the forced `{type:'resize'}` frame is sent synchronously before the `{type:'select-window'}` replay — the exact ordering that makes Reload work. No `term.reset()` needed: the backend replay prepends `ESC[H ESC[2J ESC[3J`.

Does not touch `reloadTerminal`, `connectTerminalWs`, or #663's resize debounce.

## Test plan
1. Open a session terminal in the browser; resize the same window from another surface (desktop/phone) so this surface goes garbled / stuck-small.
2. Switch session tabs (or terminal tabs) and back — display snaps to correct size, no garbled rows, **instantly** (no scrollback stall), matching right-click Reload.
3. "Stays small after closing the phone tab": switching tabs re-grows this surface.
4. Regression: right-click Reload still works; normal resize/reflow (#663 debounce) unchanged.

## Out of scope
Stopping a concurrently-foreground small surface (phone) from re-stealing the size while both are visible (`window-size latest`→`largest`) — separate follow-up. This is recovery-on-switch only.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1692,8 +1692,15 @@ function renderTabs() {
 function switchTerminal(t) {
   activeTerminal = t;
   renderTabs();
-  selectWindow(t.window);
   if (term) term.focus();
+  // #671: mirror reloadTerminal's onopen order — reclaim this surface's size for
+  // the shared tmux window (forced resize), THEN replay. Recovers a mismatched/
+  // stuck-small grid the cheap way, with no socket/PTY teardown. takeTerminalOwnership
+  // resets the applyTermFit dedup so the resize frame is sent even when this surface's
+  // CSS grid is unchanged, re-asserting size on the shared window before select-window
+  // replays the pane at the corrected size.
+  takeTerminalOwnership();
+  selectWindow(t.window);
 }
 
 async function addTerminal() {


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Pure client-side display-recovery change. No new auth, input-validation, injection, secrets, or crypto surface.
- `selectWindow`'s WebSocket payload is unchanged (`{type:'select-window', window: t.window}`) — no added attack surface. The forced `resize` frame carries only `term.rows`/`term.cols`.

**Concerns:**
- None.

### Code Quality
- The reorder in `switchTerminal()` (app.js:1692-1704) correctly mirrors the proven `connectTerminalWs.onopen` ordering (resize → select-window, app.js:2350-2353) and `reloadTerminal`'s recovery flow, reusing the existing `takeTerminalOwnership()` primitive instead of duplicating the dedup-reset logic. Good reuse.
- **Null/socket safety intact:** `switchTerminal` guards `if (term)`; `takeTerminalOwnership → applyTermFit` bails on `!term || !fitAddon` and on a 0×0/detached container; `selectWindow` guards `readyState === OPEN`. The programmatic call from `addTerminal` (app.js:1712) is safe.
- **No reentrancy:** `term.focus()` focuses the xterm textarea within an already-focused document and does not fire the window `focus` event, so the `window.focus` listener's `takeTerminalOwnership` is not re-triggered — which is exactly why the explicit call is required here.
- **Backgrounded-surface safety preserved:** if the document isn't focused/visible, `applyTermFit`'s #667 guard (app.js:2386) no-ops the resize, so this change never causes a backgrounded surface to steal the shared window size. Even the programmatic (`addTerminal`) path degrades gracefully.
- Scope is tight and matches the description: does not touch `reloadTerminal`, `connectTerminalWs`, or the #663 resize debounce.
- `node --check` passes.
- No automated test added, but this is a display self-heal in a static resource JS file with no unit-testable seam; the manual test plan is thorough and appropriate. Not a blocker.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 0 Green] findings. Minimal, well-scoped fix that reuses existing primitives and preserves the #667/#661/#663 invariants.

PR #674, opened 2026-07-12
fix(web-terminal): switching active terminal does a full Reload (#673)

### LINKED ISSUE #673: fix(web-terminal): switching active terminal should do a full Reload (cursor/text still misaligned after #672)
Follow-up to #671 / PR #672.

## Problem
The lightweight fix in #672 (call `takeTerminalOwnership()` — forced resize — before `selectWindow()` on tab switch, mirroring reload's onopen "resize THEN replay") **did not resolve** the corruption. Switching the active terminal still shows the text-jumping / misaligned state: **the cursor is not where the line actually is** — the rendered grid and the true pane content are out of sync.

The resize-then-replay-on-the-same-socket path is not enough to recover the surface. Only the heavyweight right-click **"Reload terminal"** actually fixes it.

## Requested fix
Make switching the active terminal do exactly what right-click **"Reload terminal"** does — i.e. `switchTerminal()` should invoke the full `reloadTerminal()` recovery (`term.reset()` + fresh WebSocket reconnect; `onopen` re-fits and re-selects the window, so crowd replays the pane against a stable layout) instead of the in-place `takeTerminalOwnership()` + `selectWindow()`.

`switchTerminal()` and `reloadTerminal()` both live in `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js`.

### Notes / considerations
- Accept the reload cost (socket + PTY + `tmux attach` + scrollback replay) on tab switch — correctness over snappiness, since the cheap path proved insufficient.
- `reloadTerminal()` currently replays `activeTerminal`'s window via `selectWindow(activeTerminal.window)` in `onopen`. `switchTerminal(t)` sets `activeTerminal = t` first, so calling the reload path after assigning `activeTerminal` should reconnect and select the newly-selected terminal's window. Verify the ordering (set `activeTerminal`, `renderTabs()`, then reload) so the fresh socket selects the right window.
- Only reload when the target actually differs from the current attach, if that avoids a needless teardown when re-clicking the already-active tab (optional).
- Don't regress right-click "Reload terminal" itself or #663's resize debounce.

## Acceptance
- Switching session/terminal tabs recovers the display to a clean, correctly-sized grid with the cursor aligned to the actual line — identical to right-click "Reload terminal".
- No leftover garbled/overlapping rows or cursor-vs-line mismatch after a switch.

## Base branch
`app.js` is on the parity line — branch off `feature/crow-593-web-ui-parity` and target it with the PR (`--base feature/crow-593-web-ui-parity`), same as #670 / #672.

Related: #661, #663, #667, #670, #671, PR #672.

dgershman:
Closes #673

## Problem
The lightweight fix in #672 (issue #671) — `takeTerminalOwnership()` (forced resize) then `selectWindow()` on the **live** socket — did not resolve the corruption. Switching the active terminal still left the rendered xterm grid out of sync with the true tmux pane: the cursor was not where the line actually was. Multiple surfaces share one grouped-session tmux window (`window-size latest`), so a resize-then-replay on the same socket can't recover the mismatch. Only right-click **"Reload terminal"** (`term.reset()` + fresh WebSocket reconnect) fixed it.

## Fix
Make `switchTerminal()` run the full `reloadTerminal()` recovery instead of the in-place path:

- Set `activeTerminal = t` (and `renderTabs()`) **first**, then call `reloadTerminal()`. `connectTerminalWs`'s `onopen` re-fits (`applyTermFit()`) then `selectWindow(activeTerminal.window)`, so the fresh socket selects the newly-chosen window and crowd replays the pane against a stable layout — identical to right-click "Reload terminal".
- **Guard:** re-clicking the already-active tab just re-focuses (no needless socket/PTY teardown).

## Non-regressions
- `takeTerminalOwnership()` stays wired to `window focus` / `visibilitychange` — focus-gated sizing (#667) unaffected.
- Right-click "Reload terminal" and #663's resize debounce are untouched.
- `addTerminal()` → `switchTerminal(created)`: `created` is non-active, so it gets the full reload + correct window selection.

Out of scope: session-switch (sidebar) flows through `refreshTerminals()`, not `switchTerminal()`; the ticket names only `switchTerminal()` / `reloadTerminal()`.

Related: #661, #663, #667, #670, #671, PR #672.

https://claude.ai/code/session_01VyS8LgTE55m2zfce8Sjh5f

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1690,17 +1690,21 @@ function renderTabs() {
 }
 
 function switchTerminal(t) {
+  // #673: re-clicking the active tab shouldn't tear down the socket/PTY.
+  if (activeTerminal && activeTerminal.id === t.id) { if (term) term.focus(); return; }
   activeTerminal = t;
   renderTabs();
   if (term) term.focus();
-  // #671: mirror reloadTerminal's onopen order — reclaim this surface's size for
-  // the shared tmux window (forced resize), THEN replay. Recovers a mismatched/
-  // stuck-small grid the cheap way, with no socket/PTY teardown. takeTerminalOwnership
-  // resets the applyTermFit dedup so the resize frame is sent even when this surface's
-  // CSS grid is unchanged, re-asserting size on the shared window before select-window
-  // replays the pane at the corrected size.
-  takeTerminalOwnership();
-  selectWindow(t.window);
+  // #673: the cheap resize-then-replay path (#671/#672) — takeTerminalOwnership()
+  // (forced resize) then selectWindow() on the live socket — couldn't recover the
+  // cursor/grid mismatch on switch; the shared grouped-session window stays desynced
+  // and only the heavyweight reload fixes it. Do the full "Reload terminal" recovery
+  // instead: term.reset() + fresh WebSocket reconnect. connectTerminalWs's onopen
+  // re-fits then selectWindow(activeTerminal.window) — set above, so the fresh socket
+  // selects the newly-chosen window — and crowd replays the pane against a stable
+  // layout, identical to right-click "Reload terminal". Correctness over the snappier
+  // in-place path, which proved insufficient.
+  reloadTerminal();
 }
 
 async function addTerminal() {


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Purely client-side terminal UI logic; no new input parsing, no injection surface. `selectWindow`/`resize` frames carry opaque tmux window ids over the pre-existing WebSocket control channel.
- `reloadTerminal()` detaches the old socket's `onopen/onmessage/onclose/onerror` before `close()`, so the stale socket can't reconnect or tear down the freshly-created `termWs` — no reconnect race introduced by routing switch through the reload path.

**Concerns:** None.

### Code Quality
- Correct ordering: `activeTerminal = t` is set **before** `reloadTerminal()`, so `connectTerminalWs`'s `onopen` runs `applyTermFit()` then `selectWindow(activeTerminal.window)` against the newly-chosen window — matching right-click "Reload terminal" behavior.
- The re-click guard (`app.js:1694`) correctly avoids needless socket/PTY teardown when the active tab is clicked again, and does not block the separate right-click "Reload terminal" action (which calls `reloadTerminal()` directly).
- Both callers verified safe: the tab-click path and `addTerminal()`. When `refreshTerminals()` makes the created terminal already-active (first terminal case), the guard short-circuits after the window was already selected in `refreshTerminals` — no double reload; otherwise the full reload selects `created.window`.
- The `!term` early-return in `reloadTerminal()` is benign: `term` is lazily created and never reset to null, so it's null only before the terminal view exists (no clickable tabs).
- Comments accurately document the correctness-over-snappiness trade-off and the #671/#672 history.
- `node --check` passes.

### Summary Table

No Red, Yellow, or Green findings — the change is a tight, well-reasoned fix that trades a snappier in-place path (proven insufficient) for the reliable full-reload recovery.

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 0 Green] findings.

---

PR #675, opened 2026-07-12
fix(web-terminal): full Reload recovery on tab/session switch (#673)

### LINKED ISSUE #673: fix(web-terminal): switching active terminal should do a full Reload (cursor/text still misaligned after #672)

dgershman:
Closes #673. Follow-up to #671 / #672.

## Problem
#672 made a tab switch call `takeTerminalOwnership()` (forced resize) before `selectWindow()` on the live socket — an in-place "resize THEN replay." In practice this **did not recover** a mismatched grid: after another surface reshaped the shared tmux window, switching still left the **cursor misaligned from the actual line** (text-jumping). Only the heavyweight right-click **Reload terminal** — `term.reset()` + a fresh WebSocket whose `onopen` re-fits and re-selects the window against a stable layout — repaints cleanly.

The same defect lives on the **session-switch** path: `selectSession → refreshTerminals` also does a plain `selectWindow` on the same persistent, shared socket.

## Fix (app.js only)
Route switching through the full reload path via a new `attachWindow(win)`:

- If `win` differs from the currently attached window → `reloadTerminal()` (the exact right-click recovery). Otherwise no-op.
- Called from **both** places that change which window this shared socket shows:
  - `switchTerminal` — terminal-tab switch (pulls back #672's `takeTerminalOwnership()` + `selectWindow()`).
  - `refreshTerminals` — session switch (was a plain `selectWindow`).
- The **change-guard** means a same-session background refresh (a "terminals changed" event) or re-clicking the already-active tab stays a no-op — no socket teardown mid-work.
- `connectTerminalWs`'s `onopen` keeps `attachedWindow` truthful across initial connect, `reloadTerminal`, and dropped-socket auto-reconnect.

Untouched: right-click **Reload terminal**, #663's resize debounce. `takeTerminalOwnership()` is retained for #667's focus/visibility size re-assert.

## Test plan
1. Open a session terminal in the browser; from another surface (desktop/phone) resize the same window so this surface's grid goes mismatched (cursor off the actual line).
2. Switch to another **session** tab and back → clean, correctly-sized grid, cursor aligned — identical to right-click Reload.
3. Switch **terminal** tabs within a session → same clean recovery.
4. Re-click the already-active tab, or let a background "terminals changed" refresh fire for the current session → **no** reload/flash (change-guard no-op).
5. Regression: right-click Reload still works; normal resize/reflow (#663 debounce) unchanged; #667 focus-based size reclaim unchanged.

## Note
Accepts the reload cost (socket + PTY + `tmux attach` + scrollback replay) on a real switch — correctness over snappiness, since #672's cheap path proved insufficient. Bounded to actual window changes by the guard.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1667,7 +1667,12 @@ async function refreshTerminals() {
     activeTerminal = terminals[0] || null;
   }
   renderTabs();
-  if (activeTerminal) selectWindow(activeTerminal.window);
+  // #673: session switches funnel through here too (selectSession → refreshTerminals),
+  // changing which window this shared socket shows — same corruption as a terminal-tab
+  // switch. attachWindow reloads only when the window actually changes, so a same-session
+  // background refresh (a "terminals changed" event) stays a no-op and won't tear down
+  // the socket mid-work.
+  if (activeTerminal) attachWindow(activeTerminal.window);
 }
 
 function renderTabs() {
@@ -1690,21 +1695,17 @@ function renderTabs() {
 }
 
 function switchTerminal(t) {
-  // #673: re-clicking the active tab shouldn't tear down the socket/PTY.
-  if (activeTerminal && activeTerminal.id === t.id) { if (term) term.focus(); return; }
   activeTerminal = t;
   renderTabs();
+  // #673: the in-place resize+replay from #672 didn't recover a mismatched grid —
+  // the cursor stayed misaligned from the actual line. Attaching to a different
+  // window now does the full "Reload terminal" recovery (reset + fresh socket)
+  // instead, which is the only path that reliably repaints against a stable
+  // layout. attachWindow only pays that cost when the window actually changes.
+  // (Re-clicking the active tab is a no-op: attachWindow guards on win ===
+  // attachedWindow, superseding #674's id-based early-return guard.)
+  attachWindow(t.window);
   if (term) term.focus();
-  // #673: the cheap resize-then-replay path (#671/#672) — takeTerminalOwnership()
-  // (forced resize) then selectWindow() on the live socket — couldn't recover the
-  // cursor/grid mismatch on switch; the shared grouped-session window stays desynced
-  // and only the heavyweight reload fixes it. Do the full "Reload terminal" recovery
-  // instead: term.reset() + fresh WebSocket reconnect. connectTerminalWs's onopen
-  // re-fits then selectWindow(activeTerminal.window) — set above, so the fresh socket
-  // selects the newly-chosen window — and crowd replays the pane against a stable
-  // layout, identical to right-click "Reload terminal". Correctness over the snappier
-  // in-place path, which proved insufficient.
-  reloadTerminal();
 }
 
 async function addTerminal() {
@@ -2354,7 +2355,12 @@ function connectTerminalWs() {
     lastTermCols = 0;
     lastTermRows = 0;
     applyTermFit();
-    if (activeTerminal) selectWindow(activeTerminal.window);
+    if (activeTerminal) {
+      selectWindow(activeTerminal.window);
+      // Keep the attach bookkeeping truthful after any (re)connect — initial
+      // connect, reloadTerminal, or a dropped-socket auto-reconnect (#673).
+      attachedWindow = activeTerminal.window;
+    }
   };
   termWs.onmessage = (event) => {
     if (event.data instanceof ArrayBuffer) term.write(new Uint8Array(event.data));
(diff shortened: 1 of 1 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Purely client-side UI wiring in `app.js`; no change to the WebSocket origin/auth guards, RPC surface, or the `select-window` control frame. No new input handling, secrets, or injection surface.
- `attachWindow` only ever forwards an already-validated `window` index through the existing `selectWindow`/`reloadTerminal` paths; server still clamps/validates (`TerminalWebSocket`).

### Code Quality
Verified the logic across every path:

- **Change-guard correctness.** `window` is `tmuxBinding.windowIndex` in the single shared `crow-cockpit` tmux session (one window per terminal across *all* sessions — `TerminalCockpit`, `EngineRouter:820`), so indices are globally unique. `win === attachedWindow` therefore no-ops only a genuine same-window re-attach, never a real cross-session switch. Sound.
- **Initial connect** avoids a redundant reload: on first `selectSession`, the socket is `CONNECTING` when `refreshTerminals → attachWindow` runs, so it only records `attachedWindow`; `onopen` performs the `selectWindow`.
- **Reconnect / mid-connect switch races** resolve correctly because `connectTerminalWs.onopen` re-reads the live `activeTerminal.window` and re-truths `attachedWindow`, rather than a captured value.
- **All `refreshTerminals` callers** (`selectSession`, `addTerminal`, `closeTerminal`, `handoffAgent`) are covered by the guard; no periodic/SSE trigger fires it, so no reload churn.
- Syntax check passes; no JS unit-test harness exists in this repo, so the absence of tests is consistent with existing browser-UI code.

**Green considerations (non-blocking):**
- Re-clicking the *already-active* tab (or returning focus to a same-window surface) now does nothing, dropping #672's in-place resize+replay. This is intentional and documented, and #667's focus/`visibilitychange` → `takeTerminalOwnership()` plus right-click **Reload** cover recovery. Worth a sanity check that a same-window surface whose grid went cursor-misaligned (the exact class the PR argues resize-alone can't fix) still self-heals on focus return, since that path is resize-only, not a full reload.
- Very narrow stale-guard edge: a single-terminal session whose active terminal is closed *externally* and a new terminal reuses the same freed tmux window index — `attachWindow(sameIndex)` would no-op, so the grouped view might not re-select. Extremely rare and right-click Reload recovers; noting only for completeness.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. The fix is a well-reasoned, correctly-guarded, single-file change that routes both switch paths through the proven full-reload recovery while bounding the cost to real window changes.

---

PR #684, opened 2026-07-13
fix(skill): inject matched workspace's customInstructions into worker prompt

### LINKED ISSUE #683: New workspace creation doesn't inject the workspace's customInstructions into the worker prompt
Creating a new workspace does not pull the `customInstructions` configured for that **specific workspace** — the launched worker's prompt is missing its `## Custom Instructions` section, so per-workspace guidance (e.g. the `rm` workspace's "assign the PR to dhilgaertner + label crow:merge…", tanzaterra's reviewer/Render steps) never reaches the agent.

## The value is stored correctly (gap is on consumption)
Each workspace in `{devRoot}/.claude/config.json` has a `customInstructions` string, edited via Settings → Workspace → Custom Instructions:
- Written per workspace: `Packages/CrowUI/Sources/CrowUI/WorkspaceFormView.swift:291-292, 352, 369`.
- Decoded per workspace: `Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift:384, 478`.

So the config round-trips fine. The bug is that new-workspace creation doesn't **resolve + inject** the matched workspace's value into the prompt.

## Two consumers — new-workspace creation is dropping it at one of them
1. **`/crow-workspace` skill prompt-build** — `skills/crow-workspace/SKILL.md:461-466` (and `Resources/crow-workspace-SKILL.md.template`) says to append a `## Custom Instructions` section from the workspace config verbatim when non-empty. If creation goes through the skill, ensure it resolves the **matched** workspace's `customInstructions` — not `defaults`, and not the template's own example value `"customInstructions": null` (SKILL.md:35) — and writes it into the prompt file that `setup.sh` launches.
2. **App-side launcher** — `Packages/CrowClaude/Sources/CrowClaude/ClaudeLauncher.swift:87-90` appends `## Custom Instructions` only when passed a non-nil `customInstructions` (surfaced via `ClaudeCodeAgent.launchCommand`, :105). If creation launches through this path, the **caller** must resolve `config.workspaces.first(where: { $0.name == <workspace-for-this-repo> })?.customInstructions` and pass it — verify it isn't passing `nil` or the wrong workspace. (Per-workspace lookup pattern already used at `SessionService.swift:892 / 1449 / 2065`.)

## Ask
Confirm which path new-workspace creation actually uses, then fix the per-workspace `customInstructions` resolution there so the section reliably lands in the worker prompt. Add a regression test/check.

## Repro
1. Set custom instructions on a workspace that has them (e.g. `rm` already does).
2. Create a new workspace/session for a repo in that workspace.
3. Inspect the generated prompt at `{devRoot}/.claude/prompts/crow-prompt-<session>.md` (or the launched agent's prompt) — the `## Custom Instructions` section is absent.

## Acceptance criteria
- New workspace creation for a workspace with non-empty `customInstructions` yields a worker prompt containing a `## Custom Instructions` section with that workspace's text **verbatim**.
- The **matched** workspace's instructions are used — not `defaults`, not another workspace's.
- Workspaces with empty/absent `customInstructions` still produce no section (unchanged).

## Note
This is core app / skill code on `main` (not the crow-593 web UI), so `crow:auto` branching off `main` is correct here — no branch redirect needed.

dgershman:
Closes #683

## Problem
New workspace/session creation didn't reliably add a `## Custom Instructions` section to the worker prompt, so per-workspace guidance (e.g. `rm`'s PR-assignment rules, tanzaterra's reviewer/Render steps) never reached the agent.

## Which path is actually used
Creation is driven **entirely by the `/crow-workspace` (and `/crow-batch-workspace`) skill** — the Manager Claude Code writes the prompt to `{devRoot}/.claude/prompts/crow-prompt-<session>.md` and hands it to `setup.sh --prompt-content`.

The app-side `ClaudeLauncher.generatePrompt` (ticket's consumer #2) is **not** on this path: `generatePrompt` is never called anywhere in `Sources/` — only in unit tests. It already handles `customInstructions` correctly, so the bug is in the skill instructions, not Swift launch code.

**Why the skill dropped it:** the resolution note lived at the very end of the "First Prompt Template" section, disconnected from the "Write the prompt file" step, and was vague on *which* workspace to read — never saying to use the matched `workspaces["{workspace}"].customInstructions` and not `defaults`/another workspace/the doc's own examples. The batch skill never mentioned `customInstructions` at all.

## Fix
- **`/crow-workspace`** (`SKILL.md` + `Resources/*.template`): added an explicit **"Resolve custom instructions"** step *before* writing the prompt that reads the **matched** workspace's field (explicitly not `defaults`, not another workspace, not the example strings), and rewrote the template's `## Custom Instructions` block to thread `{custom_instructions}` verbatim / omit when empty.
- **`/crow-batch-workspace`** (`SKILL.md` + `Resources/*.template`): resolve the matched workspace's `customInstructions` per session in Phase 2.
- Edited **both** live skills and their bundled `Resources/*.template` (dual-sourced — release builds scaffold from the templates).
- **Regression guard:** new `Tests/CrowTests/CustomInstructionsSkillTests.swift` asserts each skill + template carries the resolution instruction (matched-workspace lookup, `## Custom Instructions` heading, verbatim, no `defaults` fallback).

## Out of scope
No change to `ClaudeLauncher`/`ClaudeCodeAgent`/`CodingAgent.generatePrompt` — that path is unused by creation and already correct+tested; threading a param no caller passes would be speculative.

## Acceptance criteria
- ✅ Non-empty `customInstructions` → prompt contains `## Custom Instructions` with that workspace's text verbatim.
- ✅ The **matched** workspace's instructions are used — not `defaults`, not another workspace.
- ✅ Empty/absent `customInstructions` → no section.

## Test
`swift test --filter CustomInstructionsSkillTests` (after `bash scripts/generate-build-info.sh`) — both tests pass.

https://claude.ai/code/session_01MZWp7QiSX7U57YyHFbyc6L

### DIFF
--- a/Resources/crow-batch-workspace-SKILL.md.template
+++ b/Resources/crow-batch-workspace-SKILL.md.template
@@ -82,7 +82,7 @@ For each workspace spec, perform the same resolution as `/crow-workspace`:
    - Jira (task-only): fetch via the `jira` MCP server (`jira_get_issue {key}`); title is the work item `summary`.
 6. **Check for existing PR**: `gh pr list --repo {owner}/{repo} --search "{issue_number}" --state open --json number,title,headRefName,url --limit 5` (with `dangerouslyDisableSandbox: true`). For a Jira-task session this runs against the workspace's configured GitHub/GitLab code repo, not Jira.
 7. **Generate names**: slug, branch, worktree path, session name (following `/crow-workspace` naming conventions). **Jira:** resolve `{ticket_number}` as the **numeric suffix** of the key (`MAXX-6859` → `6859`); `{ticket_url}` is the full `…/browse/{key}` URL; the slug uses the lowercased full key (`{repo}-maxx-6859-{slug}`).
-8. **Compose prompt**: Use the First Prompt Template from `/crow-workspace`
+8. **Compose prompt**: Use the First Prompt Template from `/crow-workspace`. Resolve `{custom_instructions}` from the **matched** workspace's `workspaces["{workspace}"].customInstructions` in the config read in step 1 (not `defaults`, not another workspace) and append a verbatim `## Custom Instructions` section when non-empty — exactly per `/crow-workspace`'s **Resolve custom instructions** step and **First Prompt Template**.
 9. **Write prompt file**: `cat > {devRoot}/.claude/prompts/crow-prompt-{session_name}.md`
 
 This phase is sequential because each resolution involves LLM reasoning (scoring repos, generating slugs). It's fast (~2-3 seconds per workspace).
--- a/Resources/crow-workspace-SKILL.md.template
+++ b/Resources/crow-workspace-SKILL.md.template
@@ -307,6 +307,17 @@ Store the result as `{base_branch}` for substitution into Step 2 and into the pr
 
 Render the fetched content into the prompt template per the formatting rules in **First Prompt Template** below.
 
+**Resolve custom instructions (required before writing the prompt).** Read the
+**matched** workspace's entry in `{devRoot}/.claude/config.json` —
+`workspaces["{workspace}"].customInstructions`, where `{workspace}` is the same
+value passed to `setup.sh --workspace` — and store it as `{custom_instructions}`.
+This must be the matched workspace's own field: **do not** read `defaults`, **do
+not** use another workspace's value, and **do not** copy the example strings shown
+in the config sample above (`"Always run …"` / `null`). If the field is `null`,
+absent, or empty after trimming, treat `{custom_instructions}` as empty (no
+section). This value feeds the `## Custom Instructions` block of the prompt
+template (see **First Prompt Template** below).
+
 #### Step 1: Write the prompt file
 
 The LLM writes the prompt content (see template below — with the pre-fetched ticket/PR content embedded) to a file:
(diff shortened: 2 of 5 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Change is confined to skill-instruction Markdown and a Swift regression test — no runtime/auth/IO surface, no new external calls, no secrets.
- The resolution step is explicitly scoped to the *matched* workspace's `customInstructions` and forbids reading `defaults` or another workspace's value, which prevents cross-workspace instruction bleed (one workspace's guidance leaking into another's agent prompt).

**Concerns:**
- None.

### Code Quality
- **Correct bug surface.** Verified the PR's central claim: workspace/session creation is driven by the `/crow-workspace` skill, and the Swift `ClaudeLauncher` prompt path already handles `customInstructions` correctly and is covered by existing tests. So the real gap was the skill instructions, and that's exactly what this fixes. (Minor nit, non-blocking: the PR description says `generatePrompt` is "never called in Sources/" — the agent wrappers do call it — but the substantive point that the Swift path is already correct+tested holds, so scope is right.)
- **Dual-source coverage.** Both the live skill (`skills/crow-workspace/SKILL.md`, `skills/crow-batch-workspace/SKILL.md`) and the bundled `Resources/*.template` copies (which release builds scaffold from) are updated together — the exact failure mode where scaffolded installs silently regress is guarded.
- **Clear instruction rewrite.** The new "Resolve custom instructions" step is placed *before* the write-prompt step, names `workspaces["{workspace}"].customInstructions`, references the real `--workspace` flag and config sample, and specifies the empty/null/whitespace → omit-section behavior. Template block threads `{custom_instructions}` verbatim.
- **Regression guard verified.** Ran `swift test --filter CustomInstructionsSkillTests` after `scripts/generate-build-info.sh`: build succeeds, both tests pass. All asserted substrings were confirmed present in the target files. Assertions are intentionally exact (a reword breaks them) — appropriate for a doc-drift guard.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 0 Green] findings. Correct, minimal, well-targeted fix with a passing regression guard covering both live skills and their templates.

---

PR #686, opened 2026-07-13
fix(web-ui): harden mobile hamburger + edge chrome against notch/safe-area (#685)

### LINKED ISSUE #685: Mobile: keep the sidebar hamburger (☰) pinned/reachable while scrolling
On mobile, keep the sidebar **hamburger (☰)** pinned while scrolling so you can always get back to it easily.

## Current state (may already be partly done)
`#back-to-sidebar` (`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/index.html:14`) is already styled `position: fixed; top: 12px; left: 12px; z-index: 50` and shown on mobile when a session is selected (`web/app.css:386-404`, `:418`). No transformed/filtered/`will-change`/`contain` ancestor was found on `#app` / `#detail` / `#board` that would defeat `position: fixed`. So in the current crow-593 branch the ☰ *should* already stay pinned to the viewport top-left during scroll.

This means it's likely one of:
- **(a)** the running build predates that CSS → already fixed in-branch; verify on device and **close as already-addressed**, or
- **(b)** it's defeated/covered in a specific scroll context → fix that view.

## If (b): concrete things to check on a real mobile browser
- A containing-block breaker on an ancestor (`transform` / `filter` / `will-change` / `perspective` / `contain`) that turns `fixed` into scroll-with-ancestor. None found today, but re-verify with mobile chrome (address-bar show/hide, dynamic viewport) in play.
- ☰ getting **covered/overlapped** by scrolled content — z-index/padding. The board already clears it (`#board padding-top: 56px`, `app.css:425`; sticky `.board-head` at `:447`); check the **terminal** view and **long detail** views the same way so ☰ is never hidden behind content.
- On the sidebar/list view ☰ is `display: none` by design (`app.css:423`) because the list *is* the sidebar — confirm the user doesn't also want a fixed control there (probably not).

## Ask
Verify on a real mobile browser, across every has-selection scroll view (terminal, board/reviews/allowlist, long detail), that ☰ stays pinned top-left and tappable while scrolling; fix any view where it scrolls out of reach or is covered. If it already holds everywhere, close as already-addressed.

## Acceptance criteria
- On mobile, with a session selected, scrolling any pane keeps ☰ pinned at the top-left and tappable (opens the sidebar).
- No overlap that hides or blocks ☰ in the board, terminal, or long detail views.

## Target branch
Web UI (mobile) — base on `feature/crow-593-web-ui-parity`, not `main` (`crow:auto` defaults to `main`, where the web UI doesn't exist).

dgershman:
Closes #685

## Finding

The mobile sidebar hamburger (☰, `#back-to-sidebar`) **already stays pinned** while scrolling: `#app` is `height: 100dvh` with `body { overflow: hidden }`, so the page itself never scrolls — only `#board`, `#sidebar` (`overflow-y:auto`) and xterm scroll *internally*. A `position: fixed` element therefore can't scroll away, and there's no transformed/`contain` ancestor to defeat it. Coverage is handled too (☰ `z-index:50`; board `padding-top:56px` + sticky head z‑2; terminal skeleton z‑3; header `padding-left:60px`).

The one real-device gap the ticket flags ("re-verify with mobile chrome / dynamic viewport") is **safe areas**: `top:12px; left:12px` ignored notches/rounded corners, and `viewport-fit=cover` wasn't set (so `env(safe-area-inset-*)` was 0).

## Change

- `index.html`: add `viewport-fit=cover` so iOS reports non-zero safe-area insets.
- `app.css`: offset the ☰ by `env(safe-area-inset-top/left)`, and keep the clearances that stop the ☰ from covering/being-covered consistent with its new position — `#detail-header` padding-left, board `padding-top`, `#sidebar` padding-bottom, and the bottom-left `#statusbar` (clears the home indicator).

`env()` resolves to 0 on desktop / non-notched devices, so those layouts are unchanged. Pure HTML/CSS — no JS/Swift.

## Verify
Run the daemon, open the web UI in Chrome DevTools device mode (notched device, e.g. iPhone 14 Pro) or a real iOS device. With a session selected, scroll each has-selection view (terminal, board/reviews/allowlist, long detail header) and confirm the ☰ stays pinned top-left, isn't clipped by the notch/status bar, and taps through to reveal the sidebar. Confirm the statusbar sits above the home indicator.

🐦‍⬛ Created with Crow

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -386,8 +386,10 @@ html, body {
 #back-to-sidebar {
   display: none;
   position: fixed;
-  top: 12px;
-  left: 12px;
+  /* Offset by the device safe area so the ☰ clears notches / rounded corners and
+     the status bar on notched phones (viewport-fit=cover). env() → 0 elsewhere. */
+  top: calc(12px + env(safe-area-inset-top));
+  left: calc(12px + env(safe-area-inset-left));
   z-index: 50;
   width: 40px;
   height: 40px;
@@ -416,15 +418,15 @@ html, body {
 
   #app.has-selection #sidebar { display: none; }
   #app.has-selection #back-to-sidebar { display: flex; }
-  #app.has-selection #detail-header { padding-left: 60px; }
+  #app.has-selection #detail-header { padding-left: calc(60px + env(safe-area-inset-left)); }
 
   #app.has-selection.mobile-show-sidebar #sidebar { display: block; }
   #app.has-selection.mobile-show-sidebar #detail { display: none; }
   #app.has-selection.mobile-show-sidebar #back-to-sidebar { display: none; }
-  /* Board content clears the floating hamburger on mobile. */
-  #app.has-selection.board-active #board { padding-top: 56px; }
-  /* Extra room under the last card so it's fully reachable above mobile chrome. */
-  #sidebar { padding-bottom: 64px; }
+  /* Board content clears the floating hamburger on mobile (incl. its safe-area top). */
+  #app.has-selection.board-active #board { padding-top: calc(56px + env(safe-area-inset-top)); }
+  /* Extra room under the last card so it's fully reachable above mobile chrome / home indicator. */
+  #sidebar { padding-bottom: calc(64px + env(safe-area-inset-bottom)); }
   /* The status bar would overlap the fullscreen terminal on mobile — show it only
      while the sidebar is the visible view. */
   #statusbar { display: none; }
@@ -559,8 +561,9 @@ html, body {
 /* ---- Fixed connection/status bar (bottom-left, CROW-593) ---- */
 #statusbar {
   position: fixed;
-  left: 0;
-  bottom: 0;
+  /* Clear the home indicator / rounded corners under viewport-fit=cover. */
+  left: env(safe-area-inset-left);
+  bottom: env(safe-area-inset-bottom);
   z-index: 60;
   display: flex;
   align-items: center;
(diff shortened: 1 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Pure presentational HTML/CSS change — no JS, no Swift, no new data flow. Zero injection/auth/secret surface. `env(safe-area-inset-*)` is a UA-computed value, not attacker-controllable.
- `viewport-fit=cover` is the correct and only mechanism to make iOS report non-zero safe-area insets; scoped to layout only.

**Concerns:**
- None.

### Code Quality
Verified the geometry is internally consistent, so the ☰ neither clips nor is clipped after the safe-area shift:
- `#back-to-sidebar`: `top: calc(12px + inset-top)` + 40px height → bottom edge `52px + inset-top`; `#board padding-top: calc(56px + inset-top)` clears it. ✓ (app.css:391, 427)
- `left: calc(12px + inset-left)` + 40px width → right edge `52px + inset-left`; `#detail-header padding-left: calc(60px + inset-left)` clears it. ✓ (app.css:392, 421)
- `#sidebar padding-bottom` and `#statusbar` bottom/left insets correctly clear the home indicator; env() → 0 on desktop keeps all non-notched layouts byte-identical. ✓ (app.css:429, 565-566)
- Comments accurately document intent and the env()→0 fallback behavior.

### Consider (non-blocking)
- `env()` used without a fallback arg (e.g. `env(safe-area-inset-top, 0px)`). Any browser that supports `env()` at all returns 0 when there's no inset, so the target audience (modern iOS/Chrome mobile) is fully covered. A fallback would only harden against pre-2018 browsers that don't parse `env()` (where the whole `calc()` declaration would drop and the ☰ would lose its fixed offset). Given the modern-mobile target this is optional defensiveness, not a defect.
- On a notched phone the offset `#statusbar` now exposes its (intentionally border-less) bottom/left edges slightly inset from the corner; its background fill keeps it reading cleanly, so purely cosmetic.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. Correct, minimal, well-reasoned safe-area hardening with no regression risk on non-notched layouts.

Issue #690, opened 2026-07-13
Persist a SessionAnalytics snapshot at session end


dhilgaertner:
ADR 0008 follow-up **2 of 11** ([ADR](https://github.com/corveil/crow/blob/main/docs/adr/0008-ai-usage-efficiency-scorecard.md), refs #648, design PR #657). **v1-blocking.**

Today the `SessionAnalytics` aggregate lives in memory on `SessionHookState.analytics` and is blank after relaunch until new telemetry arrives; only raw datapoints persist (`telemetry.db`, 180-day retention). The scorecard's weekly rollups and trailing-4-week baseline need a durable per-session record.

**Scope:** persist a `SessionAnalytics` snapshot when a session ends, so weekly rollups and self-comparison baselines survive relaunch. This snapshot is also the persistence home for the compaction counter (see follow-up 3) and the source for the "sessions shipped" weekly count.

**Where:** `Packages/CrowCore/Sources/CrowCore/Models/SessionAnalytics.swift`, `AppState.swift` (`SessionHookState`), persistence via `CrowPersistence`.

PR #657, opened 2026-07-10
RFC/design: AI-usage rubric — efficiency+outcome scoring (design doc + feasibility) (refs #648)

dhilgaertner:
Design doc + data-feasibility for #648 (refs #648). **Not a feature build** — this PR adds ADR 0008 (`docs/adr/0008-ai-usage-efficiency-scorecard.md`, status Proposed; renumbered from 0007 after the Linux-CI ADR took that slot on main) and proposes the v1 scope + follow-up implementation tickets to split off the RFC.

## Recommendation (TL;DR)

**v1 = two separate surfaces, no combined number**: a per-session **A–F efficiency grade** (labeled penalty-point deductions from 100 — every deduction is a coachable sentence like "3 compactions: −15") plus a **plain throughput count**, rolled up **weekly** on a **private self-comparison scorecard** (baseline = your own trailing 4-week median). The #648 strawman `throughput × efficiency × alignment` is adopted as the **v2** shape with named trigger conditions (PR→session attribution + rework tracking + alignment tagging must exist first) — multiplicative is right then because it makes waste un-buyable with volume, but degenerate now because most sessions have zero countable outcomes at session grain.

## Data feasibility (code-verified)

| Metric | Data source | Status | Gap |
|---|---|---|---|
| Session cost + tokens | `SessionAnalytics` ← OTLP telemetry (`CrowTelemetry`) | ✅ Computable today | Aggregate is in-memory only; raw rows persist in `telemetry.db` (180d). Telemetry off by default; Claude-Code-only. |
| Avg input tokens/turn | `inputTokens / promptCount` | ✅ Computable today (proxy) | "Turn" proxied by `user_prompt` count; true per-turn reconstructable from DB rows, not modeled. |
| Cache ratio, API error rate, active time, lines, commits | `SessionAnalytics` fields | ✅ Computable today | Same persistence caveat. |
| Compaction count / session | `PreCompact`/`PostCompact` hooks → `hook-event` handler (session-resolved) | 🔧 Light plumbing | Events hit a 50-entry in-memory ring buffer only; needs a counter in the existing handler, persisted with the session analytics snapshot (ticket 2). |
| Session wall-clock duration | `SessionStart`/`SessionEnd` hooks | 🔧 Light plumbing | No timestamps derived today; `Session` persists only `createdAt`/`updatedAt`. |
[...]

Issue #648, opened 2026-07-10
RFC: rubric for rating AI usage in Crow — efficiency- and outcome-adjusted, not raw spend


dhilgaertner:
> **Type: design / brainstorm (RFC).** This is to align on the *shape* of an AI-usage rubric before building anything — capturing ideas, candidate metrics, and open questions. Not an implementation spec yet.

## Problem

Most AI-usage leaderboards rank by **raw spend / token count** — "you spent the most tokens, you're #1." That's a perverse incentive: it rewards *volume*, not *value or efficiency*. A long, sloppy, context-bloated session that had to compact five times and shipped nothing would "win" a spend leaderboard. We want a rubric that rates AI usage on **efficiency + outcomes + alignment**, so Crow can surface a healthier signal (a coaching scorecard and/or a leaderboard that rewards the right behavior).

## Goal

Design a rubric that adjusts (or replaces) raw spend with signals for:
1. **Efficiency / session hygiene** — penalize wasteful usage (bloated context, excessive compaction, runaway sessions).
2. **Outcomes / throughput** — reward actual delivery (tickets done, PRs shipped).
3. **Alignment** — reward work that maps to a high-level org KPI/goal; flag/deprioritize unrelated work.

### A. Efficiency / hygiene (mostly *penalties* / normalizers)
- **Compaction count** — number of context compactions in a session. High = the session ran too long / accumulated too much context. *(Directly derivable: Crow already captures `PreCompact`/`PostCompact` hook events.)*
- **Average input tokens per turn** — a proxy for how much context is being dragged along. Climbing avg-input over a session = context accumulation.
- **Input + cache tokens (accumulated context)** — cache-read + input token volume as a proxy for "how much you're carrying behind you." A high cache-read:output ratio can indicate a session that should have been split/reset.
- **Session length / duration** and **turns-to-outcome**.
- **Cost-per-outcome** — `$ per shipped PR` / `$ per ticket done` rather than raw `$`. This reframes spend as efficiency: spending a lot is fine *if* it ships a lot.
- **Rework rate** — reverted/reopened PRs, re-review rounds, churn — high rework = the tokens didn't stick.

### B. Outcomes / throughput (rewards)
- **Tickets completed** (Jira/GitHub done) and **PRs shipped/merged** per session/user/window. *(Crow already has `doneIssuesLast24h` + closed-PR/issue counts, and Crow-Session commit trailers link commits/PRs → sessions.)*
- **Review sessions completed** / review throughput.
[...]

PR #703, opened 2026-07-13
Handle OTLP delta-vs-cumulative temporality at ingest (stop token/cost double-count) (closes #689)

### LINKED ISSUE #689: Fix telemetry aggregation-temporality handling at ingest
ADR 0008 follow-up **1 of 11** ([ADR](https://github.com/corveil/crow/blob/main/docs/adr/0008-ai-usage-efficiency-scorecard.md), refs #648, design PR #657). **v1-blocking — scorecard grades ship behind this fix.**

`OTLPSum.aggregationTemporality` is parsed but dropped at ingest — the metrics table has no temporality column, so a query-time check in `sumMetric` is impossible with the current schema. Cumulative datapoints summed as if they were deltas can double-count token/cost totals.

**Scope:** handle delta vs. cumulative **at ingest** — either normalize cumulative datapoints to deltas on insert, or persist a temporality column and account for it in aggregation.

**Where:** `Packages/CrowTelemetry/Sources/CrowTelemetry/` (`OTLPReceiver.swift`, `Storage/TelemetryDatabase.swift`).

**Why it matters:** prerequisite for trusting any token/cost metric the efficiency scorecard grades on.

dhilgaertner:
Closes #689. ADR 0008 follow-up 1/11 (refs #648, design PR #657) — v1-blocking; makes the token/cost metrics the scorecard grades on trustworthy. Unblocks trustworthy values in #690's SessionAnalytics snapshot.

## Approach: (a) normalize cumulative → deltas at ingest

Chosen over (b) persisting a temporality column because it needs **no schema change or migration machinery** (the repo has none — `createTables()` is pure `CREATE TABLE IF NOT EXISTS`, and user DBs at `~/Library/Application Support/crow/telemetry.db` would need an upgrade path) and keeps every existing `SUM()` reader correct unchanged. For pure counters the deltas are the raw fidelity that matters.

## What changed

- **`OTLPModels.swift`**: typed `OTLPAggregationTemporality` enum (0/1/2 per metrics.proto) + `OTLPSum.temporality` accessor.
- **`TelemetryDatabase.swift`**: `insertMetric` gains defaulted `temporality`/`isMonotonic` params. For CUMULATIVE datapoints it stores the **delta** against the last cumulative value seen for that series (session + metric + sorted-keys attributes JSON), tracked in-memory inside the actor (race-free — read-prev/insert/update-prev is one actor hop). After an app restart the baseline is recovered from the stored series sum (stored deltas sum to the last cumulative value seen). Monotonic resets (value decreases, e.g. Claude Code restarted mid-session) store the new value instead of a negative delta; non-monotonic (UpDownCounter) series keep legitimate negative deltas. Zero-delta rows (exporter heartbeats/retries) are skipped.
- **`OTLPReceiver.swift`**: `processMetrics` splits the previously-flattened sum/gauge loop and threads `temporality`/`isMonotonic` through for sums; gauges unchanged.
- **Tests**: new `CrowTelemetryTests` suite (10 tests — the target was declared in Package.swift but had no sources): cumulative no-double-count, delta regression, gauge/unspecified passthrough, reset, first datapoint, restart baseline recovery, per-attribute-set series separation, zero-delta skip, non-monotonic negative delta, OTLP JSON temporality decode.

## Why this matters

Claude Code is launched with no `OTEL_..._TEMPORALITY_PREFERENCE`, so its SDK emits **CUMULATIVE** counters — the double-count was the live behavior for every token/cost number, not an edge case.

## Verified end-to-end

Booted the real `TelemetryService` in a harness and drove `/v1/metrics` over HTTP with curl: cumulative cost stream 1.0→2.5→4.0 + duplicate export + reset to 0.5 stored as rows `1.0, 1.5, 1.5, 0.5` (total **4.5**; old behavior would have reported **12.0**); delta/gauge/no-temporality payloads stored raw; process restart on the same db recovered the baseline exactly (cumulative 5.0 → delta 0.5). `swift test --package-path Packages/CrowTelemetry`: 10/10 pass; full `make test` green except pre-existing unrelated CrowGit sandbox failures (missing git identity).

## Accepted limitations (documented in code comments)

- Counter-reset immediately followed by an app restart can over-count once, bounded by the counter value at restart — avoiding it needs a persisted per-series state table, not worth the schema surface.
- Rows already written by old builds are inflated and can't be retroactively distinguished (telemetry is off by default, pre-v1); an upgrade mid-live-session self-corrects after one clamped reset.
- `isMonotonic == nil` is treated as monotonic (all claude_code counters are).

### DIFF
--- a/Packages/CrowTelemetry/.claude/skills/verify/SKILL.md
+++ b/Packages/CrowTelemetry/.claude/skills/verify/SKILL.md
@@ -1,0 +1,50 @@
+---
+name: verify
+description: Drive CrowTelemetry's OTLP ingest end-to-end — boot the real receiver, POST OTLP JSON with curl, inspect the SQLite db.
+---
+
+# Verifying CrowTelemetry changes
+
+The package's runtime surface is `TelemetryService` (`OTLPReceiver` HTTP
+listener on localhost + `TelemetryDatabase` SQLite file). There is no
+standalone binary — build a throwaway harness executable that depends on this
+package and boots the service:
+
+```swift
+// Sources/otlp-harness/main.swift — usage: otlp-harness <port> <data-dir>
+import Foundation
+import CrowTelemetry
+let service = try TelemetryService(port: UInt16(CommandLine.arguments[1])!,
+                                   dataDirectory: CommandLine.arguments[2]) { id in
+    print("[harness] data received for session \(id)")
+}
+Task { try await service.start() }
+dispatchMain()
+```
+
+Harness `Package.swift`: swift-tools-version 6.0, `platforms: [.macOS(.v14)]`,
+`.package(path: "<repo>/Packages/CrowTelemetry")`. Build with `swift build`,
+run in the background, then drive it.
+
+Gotchas:
+- Ingest is OTLP **HTTP/JSON only** (`OTEL_EXPORTER_OTLP_PROTOCOL=http/json`),
+  paths `/v1/metrics` and `/v1/logs`, POST only. One request per connection
+  (the receiver closes it after responding).
+- The resource must carry `crow.session.id` (a UUID string) in
+  `resource.attributes`, or the payload is silently skipped.
+- Datapoint values: `asDouble` (number) or `asInt` (**string**, e.g. `"100"`).
+- Sum metrics take `aggregationTemporality` (1=delta, 2=cumulative) and
+  `isMonotonic`; cumulative sums are normalized to deltas at insert.
+
+Minimal payload:
+
+```json
+{"resourceMetrics":[{"resource":{"attributes":[{"key":"crow.session.id","value":{"stringValue":"<UUID>"}}]},
+ "scopeMetrics":[{"metrics":[{"name":"claude_code.cost.usage",
+   "sum":{"aggregationTemporality":2,"isMonotonic":true,"dataPoints":[{"asDouble":1.0}]}}]}]}]}
+```
+
+Observe results with the sqlite3 CLI against `<data-dir>/telemetry.db`:
+`SELECT metric_name, value, attributes_json FROM metrics ORDER BY id` and
+compare `SUM(value)` per metric to the expected total. Events land in the
+`events` table via `/v1/logs`.
--- a/Packages/CrowTelemetry/Sources/CrowTelemetry/Models/OTLPModels.swift
+++ b/Packages/CrowTelemetry/Sources/CrowTelemetry/Models/OTLPModels.swift
@@ -89,11 +89,22 @@ public struct OTLPMetric: Codable, Sendable {
     public let gauge: OTLPGauge?
 }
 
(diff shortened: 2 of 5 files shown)


dgershman (approved):
## Code & Security Review

Verified end-to-end: `swift test --package-path Packages/CrowTelemetry` → **10/10 pass**, package builds clean. Traced both mutation call sites in `AppDelegate.swift` to reason about the new in-memory cache. The CUMULATIVE→delta normalization is correct, well-reasoned, and the accepted limitations are honestly documented in-code.

### Critical Issues
None.

### Security Review
**Strengths:**
- No new attack surface. Ingest remains loopback-only (`acceptLocalOnly`, `.loopback`) and unchanged by this PR.
- All new SQL (`storedSeriesSum`) is fully parameterized with `sqlite3_bind_*` — no interpolation. The null-safe `attributes_json IS ?` match is correct.
- Temporality is decoded via a bounded typed enum with a safe `.unspecified` fallback for unknown raw values.

**Concerns:** None.

### Code Quality
**Strengths:**
- Race-free by construction: `insertMetric` has no `await` between read-prev / compute-delta / update-prev, so the series-state mutation is atomic within a single actor hop. The PR's claim holds.
- Reset clamp (`delta < 0 && isMonotonic != false`) correctly distinguishes counter resets from legitimate UpDownCounter decreases; `nil` treated as monotonic is the right default for claude_code counters.
- Restart baseline recovery via `storedSeriesSum` is exact for the normal (non-reset) path, and elegantly works for non-monotonic series too since stored deltas always sum to the last cumulative value.
- `guard delta != 0` correctly drops exporter heartbeats — a re-exported identical cumulative double yields exactly `0.0`, and genuine small deltas are preserved. Verified by `zeroDeltaRowsAreSkipped`.
- Test suite covers the meaningful cases: double-count, delta passthrough, gauge/unspecified, reset, first-datapoint, cross-restart recovery, per-attribute series separation, zero-delta skip, non-monotonic negative delta, JSON decode.

**Consider (non-blocking):**
- `pruneOldData` (called once at startup) erodes the `storedSeriesSum` baseline. If a series ever outlived the retention window (default 180d) across an app restart, the recovered baseline would be too low and inflate the next delta. Practically impossible for Claude Code session lifetimes, and the same bounded class as the documented reset+restart case — a one-line comment noting prune as a second baseline-erosion vector would round out the in-code limitations list. (`TelemetryDatabase.swift:139`)
- `deleteSessionData` doesn't evict that session's entries from `lastCumulativeValues`. Harmless (UUIDs are never reused, and no further telemetry arrives for a deleted session) but a `lastCumulativeValues` cleanup on delete would keep the cache tidy over long app lifetimes. (`TelemetryDatabase.swift:260`)

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. Correct for all realistic inputs, thoroughly tested, limitations honestly documented.

---

PR #704, opened 2026-07-13
Persist SessionAnalytics snapshot at session end (v1-blocking scorecard foundation) (closes #690)

### LINKED ISSUE #690: Persist a SessionAnalytics snapshot at session end

dhilgaertner:
Closes #690. ADR 0008 follow-up **2 of 11** (refs #648, design PR #657). **v1-blocking.**

Persists a durable `SessionAnalyticsSnapshot` when a session reaches a terminal status, so the scorecard's weekly rollups and trailing-4-week self-comparison baseline survive relaunch — and survive the retention reaper, which routinely deletes completed/archived sessions along with their telemetry rows.

## Key decisions

- **Trigger**: status transition to `.completed` or `.archived` (both are treated as terminal everywhere else in the codebase). Two paths are covered, because they don't share a funnel:
  - `SessionService.updateSessionStatus` (used by `completeSession` → IssueTracker `markIssueDone`/auto-merge, JobScheduler);
  - the `set-status` RPC handler in `AppDelegate`, which mutates status directly. It deliberately keeps its direct mutation (routing it through `updateSessionStatus` would silently add a manager-session guard the RPC never had) and triggers the snapshot itself.
- **Data source**: prefers a fresh aggregate from `telemetry.db` via an injected async provider — this covers the relaunch-then-complete gap where `SessionHookState.analytics` is still nil — and falls back to the in-memory aggregate. When both are nil/empty it skips entirely, preserving any existing snapshot. `TelemetryDatabase.sessionAnalytics` returns **all-zeros** for a session with no rows (`COALESCE(SUM,0)`), so a new `SessionAnalytics.isEmpty` guard prevents bogus zero snapshots (the telemetry-off case ends gracefully: no crash, no entry).
- **Storage shape**: a new optional `analyticsSnapshots: [String: SessionAnalyticsSnapshot]?` field on `StoreData` in the existing `JSONStore` (`store.json`), keyed by session UUID string — exactly the `hookStates` backward-compat pattern (#367). No new store. Old `store.json` files decode with the field nil.
- **Idempotency**: dict keying means a session that resumes and ends again *updates* its snapshot (newest status + aggregate), never duplicates.
- **Lifecycle**: snapshots are **not** cascade-deleted with sessions — the auto-cleanup reaper deletes completed/archived sessions after `retentionHours`, and cascading would destroy the 4-week baseline. Unbounded growth is a known trade-off (records are tiny); pruning by `endedAt` can ride a follow-up.
- **Quit race**: the terminal transition fires the snapshot write asynchronously; `persistState()` (clean quit) backfills synchronously from the in-memory aggregate, never overwriting an existing (DB-derived, at-least-as-fresh) snapshot.
- **Terminal status recorded on the snapshot** (`completed` vs `archived`): the ADR derives "sessions shipped" from these snapshots *after* the session record is gone, so the outcome flag has to live here.
- **Not on `PersistedHookState`** — per the ADR, its contract stays the color-driving UI subset.

## Extension point for follow-up 3 (compaction counter)

Future fields ride `SessionAnalyticsSnapshot` as **optionals** (e.g. `compactionCount: Int?`) so previously persisted snapshots keep decoding — documented on the model and locked in by a decode-unknown-keys test. They should *not* go on `SessionAnalytics` (13 non-optional stored fields; a required addition would break old snapshots).

## Sibling dependency

#689 (telemetry temporality-at-ingest, follow-up 1) makes the token/cost numbers in these snapshots trustworthy — today cumulative datapoints are over-summed. The snapshot *mechanism* here is valid regardless; the *values* firm up once #689 lands. Not blocked on it (different package, low collision).

## Tests

- CrowCore: `isEmpty` semantics; snapshot ISO8601 JSON round-trip; unknown-future-key decode (extension-point contract).
- CrowPersistence: snapshot round-trip across a simulated relaunch (new `JSONStore` over the same directory); legacy `store.json` without the field decodes; a `persistState`-style live-state rewrite doesn't clobber snapshots; `SessionRepository.delete` cascades everything *except* snapshots.
- App target: provider-sourced snapshot on completion (+ survives reload); all-zeros aggregate writes nothing and preserves existing; nil provider falls back to in-memory; provider wins when in-memory is nil (relaunch gap); re-end updates instead of duplicating; non-terminal statuses and manager sessions never write; `persistState` backfills missing snapshots only.

`make test`: all packages + root suite pass (`CrowCore` 338, `CrowPersistence` 34, root 326). The only failures are pre-existing/environmental: `CrowGit` `RebaseTests` fail on machines with no global git identity ("Committer identity unknown" in the tests' temp repos) — untouched by this PR.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -285,6 +285,14 @@ public final class AppState {
         return new
     }
 
+    /// Look up a session's hook state without creating one. Use this for
+    /// read-only paths (e.g. snapshotting analytics at session end) so idle
+    /// sessions don't get empty `SessionHookState` objects instantiated —
+    /// which would also leak idle entries into `allHookStateSnapshots()`.
+    public func existingHookState(for sessionID: UUID) -> SessionHookState? {
+        _sessionState[sessionID]
+    }
+
     /// Remove hook state for a deleted session.
     public func removeHookState(for sessionID: UUID) {
         _sessionState.removeValue(forKey: sessionID)
--- a/Packages/CrowCore/Sources/CrowCore/Models/SessionAnalytics.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/SessionAnalytics.swift
@@ -1,7 +1,7 @@
 import Foundation
 
 /// Aggregated analytics data for a single Crow session, computed from OTLP telemetry.
-public struct SessionAnalytics: Sendable {
+public struct SessionAnalytics: Codable, Equatable, Sendable {
     /// Total cost in USD (from `claude_code.cost.usage`).
     public var totalCost: Double
     /// Input tokens (from `claude_code.token.usage` where type=input).
@@ -33,6 +33,12 @@ public struct SessionAnalytics: Sendable {
     public var totalTokens: Int {
         inputTokens + outputTokens + cacheReadTokens + cacheCreationTokens
     }
+
+    /// True when no telemetry has been recorded for the session. The SQL
+    /// aggregation (`COALESCE(SUM(value), 0)`) returns an all-zeros aggregate
+    /// for a session with no rows, so callers persisting snapshots must treat
+    /// an empty aggregate as "no data", not as a real measurement.
+    public var isEmpty: Bool { self == SessionAnalytics() }
 
     public init(
         totalCost: Double = 0,
--- a/Packages/CrowCore/Sources/CrowCore/Models/SessionAnalyticsSnapshot.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/SessionAnalyticsSnapshot.swift
@@ -1,0 +1,41 @@
+import Foundation
+
+/// Durable per-session analytics record, persisted when a session reaches a
+/// terminal status (`.completed` / `.archived`).
+///
+/// This is the persistence backbone for ADR 0008's scorecard: weekly rollups
+/// re-aggregate these snapshots, and the trailing-4-week self-comparison
+/// baseline is derived from them. Snapshots deliberately outlive session
+/// deletion — the retention reaper deletes completed/archived sessions (and
+/// their telemetry rows) after a configurable window, and the baseline must
+/// survive that cleanup. That is also why the terminal `status` is recorded
+/// here rather than looked up on the session: "sessions shipped" is counted
+/// from snapshots after the session record is gone.
+///
+/// Extension point (ADR 0008 follow-up 3): add future fields — e.g.
+/// `compactionCount: Int?` — as OPTIONALS on this struct, not on
+/// `SessionAnalytics`, so previously persisted snapshots keep decoding.
+public struct SessionAnalyticsSnapshot: Codable, Equatable, Sendable {
+    public var sessionID: UUID
+    /// When the session reached its terminal status. Weekly rollups bucket by
+    /// this date.
+    public var endedAt: Date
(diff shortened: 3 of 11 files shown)


dgershman (approved):
## Code & Security Review

Persisting a durable `SessionAnalyticsSnapshot` at terminal status so the ADR-0008 scorecard baseline survives the retention reaper. Focused, well-documented, thoroughly tested. Verified locally: `CrowCore` + `CrowPersistence` build clean; the three new `SessionAnalyticsSnapshot` tests pass.

### Critical Issues
None.

### Security Review
**Strengths:**
- Purely internal persistence to the local `store.json` — no untrusted input, no auth/authz surface, no injection or secrets exposure. The telemetry aggregate is read via the existing parameterized `COALESCE(SUM(value),0)` query.
- Backward-compat is handled the proven way: `analyticsSnapshots` is an optional field (mirrors the `hookStates` #367 pattern), so legacy `store.json` decodes with it `nil` and never trips the corrupt-store backup path (test-covered).

**Concerns:** None.

### Code Quality
- **Trigger coverage is complete.** Both terminal-transition paths are covered — `updateSessionStatus` (→ `completeSession`, JobScheduler) and the direct-mutation `set-status` RPC. Confirmed via grep that no other production path assigns `.completed`/`.archived` directly, so nothing slips the net.
- **Concurrency is sound.** `SessionService` is `@MainActor`, so the fire-and-forget `Task {}` writes and the `persistState` backfill serialize on the main actor; `store.mutate` is NSLock-guarded with a synchronous disk write. The quit-race is correctly closed: `persistState()` backfills synchronously from the in-memory aggregate and never overwrites an existing (DB-derived, at-least-as-fresh) snapshot.
- **The `isEmpty` guard is the right fix** for the all-zeros aggregate that `sessionAnalytics(for:)` returns for unknown/no-row sessions — prevents bogus zero snapshots and preserves any prior real snapshot. Provider-preferred / in-memory-fallback selection is correct, including the relaunch-then-complete gap.
- **Extension-point contract is locked in** by the decode-unknown-future-keys test, and the "optionals on the snapshot, not on `SessionAnalytics`" rationale is documented on the model.
- Non-cascade of snapshots on `SessionRepository.delete` is intentional and test-asserted.

- `writeAnalyticsSnapshot` stamps `endedAt = Date()` at async-write time, while the `persistState` backfill uses `session.updatedAt` (the transition time). Harmless for weekly bucketing, but using the transition time in both paths would make `endedAt` consistent regardless of which path wins.
- Unbounded `analyticsSnapshots` growth is acknowledged and deferred to a follow-up (records are tiny); noting it only so the prune-by-`endedAt` follow-up isn't forgotten.

### Summary Table

---

Issue #692, opened 2026-07-13
Persist session wall-clock duration


dhilgaertner:
ADR 0008 follow-up **4 of 11** ([ADR](https://github.com/corveil/crow/blob/main/docs/adr/0008-ai-usage-efficiency-scorecard.md), refs #648, design PR #657). **v1-blocking.**

`SessionStart`/`SessionEnd` hooks arrive but no timestamps are derived; `Session` persists only `createdAt`/`updatedAt`.

**Scope:** capture `SessionStart`/`SessionEnd` timestamps on the session model (or persist telemetry's `activeTimeSeconds` alongside).

**Note from the ADR:** wall-clock duration is **display-only context** — telemetry's `activeTimeSeconds` is the authoritative clock for all penalty normalization (an idle-overnight session shouldn't launder its compactions).

PR #705, opened 2026-07-13
Persist per-session compaction counter (PostCompact) on the analytics snapshot (closes #691)

### LINKED ISSUE #691: Persist a per-session compaction counter
ADR 0008 follow-up **3 of 11** ([ADR](https://github.com/corveil/crow/blob/main/docs/adr/0008-ai-usage-efficiency-scorecard.md), refs #648, design PR #657). **v1-blocking — depends on #690.**

`PreCompact`/`PostCompact` hooks are already registered (`Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift`) and arrive session-resolved at the hook-event handler (`Sources/Crow/App/AppDelegate.swift`), but events land only in a 50-entry in-memory ring buffer — nothing counts or persists them. Compaction count is the scorecard's heaviest-weighted penalty.

**Scope:**
- Increment a per-session counter on **`PostCompact` only** (a completed compaction — a failed/aborted compaction isn't graded waste) in the existing hook handler. No new wiring.
- Persist it with the session analytics snapshot from #690 — **not** on `PersistedHookState`, whose documented contract is the color-driving UI subset and should stay that way.

dhilgaertner:
Closes #691. ADR 0008 follow-up 3/11 (refs #648, design PR #657). Depends on #690 (SessionAnalytics snapshot — the counter's persistence home).

## #690 relationship

**#690 is merged (PR #704, commit cc027a7) and this builds directly on it** — the counter is added as `compactionCount: Int?` on `SessionAnalyticsSnapshot`, exactly per that struct's documented extension point (optionals on the snapshot, not `SessionAnalytics`, so pre-#691 snapshots keep decoding; absent = 0).

## What

- `SessionHookState` gains an in-memory `compactionCount` and a `noteCompactionEvent(_:)` seam that increments on **`PostCompact` only** — a completed compaction. `PreCompact` (and a failed/aborted compaction, which never emits `PostCompact`) is not graded waste.
- The existing `hook-event` RPC handler calls the seam next to the ring-buffer append — no new hook wiring (`PreCompact`/`PostCompact` were already registered in `ClaudeHookConfigWriter`).
- Both snapshot writers carry the count: `writeAnalyticsSnapshot` and the `persistState` quit-race backfill. The count always comes from in-memory hook state — telemetry.db has no compaction rows — regardless of whether the analytics aggregate came from the DB provider or the in-memory fallback.
- **`PersistedHookState` untouched** — its documented contract stays the color-driving UI subset (ADR 0008 mandate).

## Tests

- `noteCompactionEvent`: two `PostCompact` → 2; `PreCompact`/`Stop`/`StopFailure` don't increment; fresh state is 0.
- Snapshot JSON round-trip with the count; legacy snapshot without the key decodes with `compactionCount == nil` (treated as 0); unknown-future-key forward-compat test updated (it had used `compactionCount` as its example — now asserted to decode, with a new fake key covering forward-compat).
- Service level: 2 compactions persist as 2 and **survive a simulated relaunch** (fresh `JSONStore` over the same dir); zero compactions persist an explicit 0; the `persistState` backfill carries the count.
- Full gate: `make test` (all package suites + root `CrowTests`, 328 tests) passes.

## Known limitation

The live counter is in-memory only; an app relaunch mid-session resets it (analytics has a telemetry.db fallback for that gap; compaction events have no DB row). The ADR only mandates persistence with the end-of-session snapshot.

## Coordination

Shares the `CrowCore` analytics-model surface with #692 (session duration); this change touches only the new `compactionCount` field on the snapshot to minimize rebase friction.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -693,6 +693,9 @@ public final class SessionHookState {
     public var lastToolActivity: ToolActivity?
     public var hookEvents: [HookEvent] = []
     public var analytics: SessionAnalytics?
+    /// Completed compactions this app run (ADR 0008 follow-up 3). Persisted at
+    /// session end on `SessionAnalyticsSnapshot`, NOT on `PersistedHookState`.
+    public var compactionCount: Int = 0
     /// Timestamp of the most recent top-level Stop / StopFailure for this session.
     /// Used to suppress state elevation from background activity (e.g. the
     /// `awaySummaryEnabled` recap subagent in Claude Code ≥ 2.1.108) that
@@ -701,6 +704,13 @@ public final class SessionHookState {
     public var lastTopLevelStopAt: Date?
 
     public init() {}
+
+    /// ADR 0008: count completed compactions only. `PreCompact` (and any
+    /// failed/aborted compaction, which never emits `PostCompact`) is not
+    /// graded waste.
+    public func noteCompactionEvent(_ eventName: String) {
+        if eventName == "PostCompact" { compactionCount += 1 }
+    }
 }
 
 /// Codable, value-type snapshot of the *color-driving* subset of
--- a/Packages/CrowCore/Sources/CrowCore/Models/SessionAnalyticsSnapshot.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/SessionAnalyticsSnapshot.swift
@@ -12,8 +12,8 @@ import Foundation
 /// here rather than looked up on the session: "sessions shipped" is counted
 /// from snapshots after the session record is gone.
 ///
-/// Extension point (ADR 0008 follow-up 3): add future fields — e.g.
-/// `compactionCount: Int?` — as OPTIONALS on this struct, not on
+/// Extension point (ADR 0008 follow-ups): add future fields — like
+/// `compactionCount` below — as OPTIONALS on this struct, not on
 /// `SessionAnalytics`, so previously persisted snapshots keep decoding.
 public struct SessionAnalyticsSnapshot: Codable, Equatable, Sendable {
     public var sessionID: UUID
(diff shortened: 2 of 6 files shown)


dgershman (approved):
## Code & Security Review

Focused, well-tested follow-up that adds a per-session `compactionCount` to the analytics snapshot. I verified every structural claim in the PR body against the code and ran the CrowCore suite.

### Critical Issues
None.

### Security Review
**Strengths:**
- No security surface. Purely additive integer counter driven by a string equality check (`eventName == "PostCompact"`) — no input validation, injection, auth, crypto, or secrets concerns.
- No new hook wiring; reuses the already-registered `PreCompact`/`PostCompact` events and the existing `hook-event` RPC path.

### Code Quality
- **Counting semantics correct** (`AppState.swift:707`): `noteCompactionEvent` increments only on `PostCompact`, so `PreCompact` and failed/aborted compactions (which never emit `PostCompact`) aren't graded as waste — matches the ADR intent and the tests.
- **Placement is safe** (`AppDelegate.swift:2189`): the increment sits outside `persistedSnapshot`. I confirmed `PersistedHookState` (AppState.swift:724) excludes `compactionCount`, so bumping the counter does not trigger a spurious color-persist write — the count rides only the analytics snapshot, as intended.
- **Both writers covered**: `writeAnalyticsSnapshot` (SessionService.swift:2396) and the `persistState` quit-race backfill (SessionService.swift:2465) both read the count from in-memory hook state with a `?? 0` fallback — correct, since telemetry.db has no compaction rows even when the aggregate is DB-derived.
- **Backward compatible**: `compactionCount: Int?` is optional on the snapshot; pre-#691 JSON decodes as `nil` (treated as 0), and the forward-compat test was correctly updated to use a distinct fake key rather than the now-real `compactionCount`.
- **Concurrency**: `SessionHookState` is `@MainActor`-isolated, so the counter mutation (hook handler) and reads (async snapshot writers, via `existingHookState`) are serialized — consistent with the pre-existing `analytics` field pattern; no new data race.

**Consider (non-blocking):**
- `writeAnalyticsSnapshot` calls `appState.existingHookState(for: id)` twice (analytics fallback + compaction count). Negligible — two cheap MainActor dictionary lookups — not worth restructuring.
- The documented in-memory-only limitation (mid-session relaunch resets the live counter) is within ADR scope and clearly disclosed.

### Verification
- `swift test` (CrowCore, `SessionAnalyticsSnapshotTests`): all 6 tests pass, including `noteCompactionEventCountsPostCompactOnly`, `snapshotCompactionCountRoundTrips`, and `snapshotWithoutCompactionCountDecodesAsNil`. Clean compile.

### Summary Table

---

PR #707, opened 2026-07-13
Persist PR→session attribution store (Crow-Session trailer) + merged-per-window counts (closes #693)

### LINKED ISSUE #693: PR→session attribution store (persist Crow-Session trailer parse)
ADR 0008 follow-up **5 of 11** ([ADR](https://github.com/corveil/crow/blob/main/docs/adr/0008-ai-usage-efficiency-scorecard.md), refs #648, design PR #657). **Blocks v2 (combined score) and follow-up 6 (rework/merge-rate).**

The `Crow-Session:` commit trailer is already written *and* parsed (`IssueTracker.extractCrowSessionUUIDs` in `Sources/Crow/App/IssueTracker.swift`), but the parse result is used only as a boolean auto-merge eligibility gate over locally-known sessions and then discarded. There is no persisted PR→session mapping.

**Scope:**
- Persist `extractCrowSessionUUIDs` results as a durable PR→session attribution store.
- Derive merged-PR-per-window counts per session from it.

**Why it matters:** unlocks the scorecard's PR-based throughput metrics (PRs merged, merge rate) and is a named trigger condition for adopting the v2 multiplicative score from #648.

dhilgaertner:
Closes #693. ADR 0008 follow-up 5/11 (refs #648, design PR #657). Unblocks PR-based throughput metrics + is a v2 trigger condition; blocks follow-up 6 (rework/merge-rate).

## What

The `Crow-Session:` trailer parse (`IssueTracker.extractCrowSessionUUIDs`) was a discard-after-use auto-merge eligibility gate. This PR retains the mapping durably and adds the rollup the scorecard needs:

- **`PRSessionAttribution`** (CrowCore) — PR url/number/repo, all trailer session UUIDs (deduped, monotonic), state, `mergedAt`, first-seen/updated timestamps.
- **`StoreData.prAttributions`** — optional dict keyed by PR URL, exactly the #690 `analyticsSnapshots` backward-compat pattern; deliberately excluded from the `SessionRepository.delete` cascade so window counts survive the retention reaper.
- **Capture** rides the existing parse in `prHasCrowAuthoredCommit` (reuses `extractCrowSessionUUIDs`, no fork). ALL parsed UUIDs persist, known session or not — attribution is ground truth; the gate's known-session filter is a separate concern and its boolean is unchanged by construction (the `return Self.crowAuthored(...)` line is untouched).
- **State updates** hook the refresh flow right after `applyPRStatuses`: once merged, a PR leaves the gate path, so OPEN→MERGED/CLOSED transitions are recorded there. `mergedAt` is stamped once at first MERGED observation.
- **Read API**: `PRAttributionRepository` with `attribution(prURL:)`, `attributions(for:)`, and `mergedPRCount(for:in:)` — the merged-PR-per-window rollup follow-up 6 consumes.
- **Wiring**: `IssueTracker` now takes the shared `JSONStore`. This is load-bearing — `prAttributions` has no `appState` mirror, so a write through an ad-hoc `JSONStore()` instance would be silently clobbered by the next mutation of the shared instance's older in-memory snapshot.

## Notes / accepted limitations

- **`mergedAt` is Crow-observed** (no backend surfaces the real merge timestamp today; `PRRecord` has no such field). Error is bounded by the poll interval or app downtime — negligible for weekly windows. A future API-sourced value drops in without a schema change; documented on the field.
- **Coverage**: capture fires where commits are already fetched (the auto-merge/rebase flow). Manually merged PRs that never enter that flow aren't attributed — a cheap later extension when follow-up 6 needs fuller merge-rate denominators.
- Pre-existing ad-hoc `JSONStore()` writes in IssueTracker (fields with appState mirrors) were left as-is; possible follow-up cleanup.

## Tests

- `PRAttributionRepositoryTests` (CrowPersistence): relaunch round-trip, legacy-store forward compat, delete-cascade exclusion, window rollup filters (state/window/session), multi-session PR counts for each session.
- `IssueTrackerPRAttributionTests` (CrowTests): multi-trailer → one entry both UUIDs; no-trailer → no entry; unknown-session trailer persisted while `crowAuthored` stays false; monotonic re-capture; no-change re-capture writes nothing; OPEN→MERGED stamps `mergedAt` once; OPEN→CLOSED leaves it nil; state updates never create entries; end-to-end record → merge → `mergedPRCount` across relaunch.
- `IssueTrackerAutoMergeTests` untouched and passing (gate regression).
- `swift build` + `make test`: 335 root tests + all package tests green.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Models/PRSessionAttribution.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/PRSessionAttribution.swift
@@ -1,0 +1,54 @@
+import Foundation
+
+/// Durable PR → session attribution record, persisted whenever a PR's
+/// commits are fetched and their `Crow-Session:` trailers parsed.
+///
+/// This is ADR 0008 follow-up 5: the trailer parse used to be a
+/// discard-after-use auto-merge gate; this record retains the mapping so
+/// the scorecard can count PRs merged per session per window (and, in
+/// follow-up 6, merge rate). Records deliberately outlive session
+/// deletion — window counts must survive the retention reaper — and
+/// include unknown-session UUIDs: attribution is ground truth even when
+/// the session isn't (or is no longer) known locally.
+///
+/// Extension point: add future fields — e.g. `authorLogin: String?` —
+/// as OPTIONALS on this struct so previously persisted records keep
+/// decoding.
+public struct PRSessionAttribution: Codable, Equatable, Sendable {
+    public var prURL: String
+    public var repoNameWithOwner: String
+    public var prNumber: Int
+    /// Every UUID parsed from `Crow-Session:` trailers on the PR's commits,
+    /// deduped, in first-seen order. Monotonic: once observed an ID is never
+    /// removed, even if a later rebase drops the commit that carried it.
+    public var sessionIDs: [UUID]
+    /// Last observed PR state: "OPEN" / "MERGED" / "CLOSED".
+    public var state: String
+    /// When Crow first observed the PR in MERGED state; never overwritten.
+    /// Crow-observed (no backend surfaces the real merge timestamp yet), so
+    /// it can lag the actual merge by up to a poll interval or app downtime.
+    /// A future API-sourced value can replace it without a schema change.
+    public var mergedAt: Date?
+    public var firstSeenAt: Date
+    public var updatedAt: Date
+
+    public init(
+        prURL: String,
+        repoNameWithOwner: String,
+        prNumber: Int,
+        sessionIDs: [UUID],
+        state: String,
+        mergedAt: Date? = nil,
+        firstSeenAt: Date,
+        updatedAt: Date
+    ) {
+        self.prURL = prURL
+        self.repoNameWithOwner = repoNameWithOwner
+        self.prNumber = prNumber
+        self.sessionIDs = sessionIDs
+        self.state = state
+        self.mergedAt = mergedAt
+        self.firstSeenAt = firstSeenAt
+        self.updatedAt = updatedAt
+    }
+}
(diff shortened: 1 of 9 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No attack surface: session UUIDs are parsed from commit trailers and used only as dictionary keys/values, not as file paths (no traversal via `prURL`-as-key).
- No user input, injection sink, auth path, or secret handling introduced. Store file continues to be written owner-only (`0o600`).

**Concerns:**
- None.

### Code Quality
Strong. The implementation is correct, well-documented, and faithfully mirrors the established `analyticsSnapshots` (#690) pattern.

Verified against the tree:
- **Concurrency is sound** (the PR's "load-bearing" claim checks out): `IssueTracker` is `@MainActor` and the sole writer of `prAttributions`. `JSONStore.mutate` re-reads fresh `data` inside the closure (partial-field write) and flushes to disk *synchronously* before returning, so injecting the shared store is the right fix and the three remaining ad-hoc `JSONStore()` writes (`data.links` / `data.sessions` only) cannot clobber `prAttributions`.
- **State filter is correct across providers**: both GitHub and GitLab normalize `state` to `OPEN`/`MERGED`/`CLOSED` (`GitLabCodeBackend.normalizeState`), so `mergedPRCount`'s `state == "MERGED"` literal comparison is safe — not a GitHub-only assumption.
- **Merge logic**: `capturedAttribution` correctly diffs the mutated copy against the original `existing?` (stamps `mergedAt` once, monotonic session-ID union, no-op writes suppressed); `attributionStateUpdates` never creates entries (empty `parsedSessionIDs` + nil existing → nil).
- **Backward-compat & delete-cascade exclusion** are both covered by tests and behave as documented.

**Build & tests:** root + `CrowPersistence` targets compile; all 14 new tests pass (9 `IssueTracker` + 5 repository). (Plain `swift build` trips only on the codegen'd `BuildInfo`, which is a pre-existing make-only concern, unrelated to this PR.)

### Consider (non-blocking)
- **`state: String` vs. a typed enum** — the literal `"MERGED"` filter is correct today because both providers normalize, but an enum would make that invariant compiler-enforced. Kept as `String` for consistency with `ViewerPR.state` / #690 — a reasonable, deliberate choice.
- **Merge-observation coverage** — `mergedAt` is only stamped when the MERGED PR resurfaces via the stale-PR path (`collectStalePRURLs`, i.e. a session `.pr` link at merge time). A PR whose session is reaped before the merge is polled could miss the stamp. This is consistent with the PR's documented coverage limitations and is a fine follow-up-6 extension, not a defect in what's shipped.

### Summary Table

---

Issue #710, opened 2026-07-13
v1 read-only efficiency scorecard surface (A–F grade + sessions shipped, weekly)


dhilgaertner:
The v1 deliverable of ADR 0008 ([ADR](https://github.com/corveil/crow/blob/main/docs/adr/0008-ai-usage-efficiency-scorecard.md), refs #648). All data prerequisites are merged: #689 (temporality fix), #690 (persisted `SessionAnalytics` snapshots), #691 (compaction counter), #692 (wall-clock duration). This ticket is read-only compute + presentation over that persisted data — no new capture.

**Scope (all semantics fixed by the ADR):**
- **Per-session A–F grade:** labeled penalty-point deductions from 100 — every deduction is a coachable sentence ("3 compactions: −15"). Graded: compaction count, context pressure (`inputTokens/max(1,promptCount)`), cache hit ratio (`cacheRead/(input+cacheCreation)`, higher = better), API error rate, cost per shipped session (weekly grain only). Displayed, not graded: cost, active time, commits, churn hint.
- **Penalty normalization per active hour only** (`activeTimeSeconds` is the authoritative clock; wall-clock is display-only).
- **Weekly rollup re-aggregates raw numerators/denominators** across the week's sessions (Σ compactions / Σ active hours) — never an average of per-session grades.
- **Throughput headline: sessions shipped** (count of outcome-flagged sessions from persisted snapshots). `doneIssuesLast24h` stays a board badge, not a scorecard input.
- **Private self-comparison surface:** baseline = user's own trailing 4-week median. No leaderboard, no combined number in v1.
- **Guardrails:** <~5 prompts → "insufficient data"; `sessionsShipped == 0` week → cost metric shown "insufficient outcomes", never divided by a fallback.
- **Grade bands as named constants** with a binding 4-week calibration period (starting priors in the ADR).

**Where:** reads persisted snapshots from #690/#691/#692; new scorecard view in the app.

PR #718, opened 2026-07-13
Per-turn analytics reader over telemetry.db raw rows (enables context-trend detection) (closes #695)

### LINKED ISSUE #695: Per-turn analytics reader over telemetry.db raw rows
ADR 0008 follow-up **7 of 11** ([ADR](https://github.com/corveil/crow/blob/main/docs/adr/0008-ai-usage-efficiency-scorecard.md), refs #648, design PR #657). Not v1- or v2-blocking.

Today "turn" is proxied by `promptCount` (count of `user_prompt` events), so context pressure is a per-session average (`inputTokens / promptCount`). True per-turn records are reconstructable from raw rows in `~/Library/Application Support/crow/telemetry.db` (180-day retention) but not modeled.

**Scope:** model turns from `telemetry.db` raw rows instead of the `promptCount` proxy.

**Why it matters:** enables within-session context-*trend* detection — the ADR prefers flagging a rising trend over an absolute per-session average threshold for the context-pressure grade.

**Where:** `Packages/CrowTelemetry/Sources/CrowTelemetry/Storage/TelemetryDatabase.swift`.

dhilgaertner:
Closes #695. ADR 0008 follow-up 7/11 (refs #648, design PR #657).

Read-only per-turn reader replacing the `promptCount` proxy; enables within-session context-pressure trend detection for the scorecard grade.

## What
- **`TurnAnalytics`** (CrowCore): per-turn sums of input/output/cacheRead/cacheCreation tokens + cost, plus a documented `contextTokenEstimate` (input + cacheRead — upper-bounds true context when a turn spans multiple API requests).
- **`TelemetryDatabase.turnAnalytics(for:)`** (+ `TelemetryService.turnAnalytics(for:)` passthrough): reconstructs one record per `claude_code.user_prompt` event by segmenting stored token/cost **delta** rows on their timestamps (`timeUnixNano`, `received_at` fallback for NULL/unparseable). Rows stamped before the first prompt fold into turn 0, so per-field sums across turns match the `sessionAnalytics` aggregate (tested invariant).
- **`EfficiencyGrading.perTurnInputTrendSlope`**: pure least-squares slope of per-turn input tokens — makes rising-vs-flat detectable now; nil below 3 turns.
- ADR 0008 follow-up table row 7 annotated as implemented.

## What it does NOT do
- No new capture, no ingest/schema change — pure reader (return-empty-on-failure, matching the existing `sumMetric`-style helpers).
- **Not wired into `grade()`**: the average-based context-pressure deduction shipped by #710 stays. Handoff: a follow-up adopts `turnAnalytics` + `perTurnInputTrendSlope` for the trend-based grade and picks its threshold.

## Degradation
A session whose raw rows aged out of the 180-day retention (or telemetry off) returns `[]` — callers fall back to the `promptCount` average (tested).

## Coordination
#689 (ingest temporality, PR #703) already merged — built on top; reader-only additions to `TelemetryDatabase.swift`, no conflict.

## Tests
- CrowTelemetry: segmentation across 3 turns, pre-first-prompt fold + sum invariant, `received_at` fallback, empty/aged-out session, zero-traffic trailing prompt, existing aggregates regression-guarded.
- CrowCore: rising/flat/falling/noisy slope, <3 turns → nil.
- Gate: `make test` (all packages + root, 405 + 16 + 369 tests) and root `swift build` green.


dgershman (approved):
## Code & Security Review

Read-only per-turn reader over `telemetry.db`. Reviewed all 7 changed files, built + ran the CrowTelemetry (16) and CrowCore `TurnTrend` (5) suites locally — all green.

### Critical Issues
None.

### Security Review
**Strengths:**
- `turnAnalytics(for:)` and `storedSeriesSum` use parameterized `?` bindings throughout — no new injection surface. The pre-existing string-interpolated `DELETE` statements in `deleteSessionData`/`pruneOldData` are untouched by this PR.
- Pure reader: no new capture, no ingest/schema change, no secrets. Return-empty-on-failure matches the existing `sumMetric` helpers.
- All access stays serialized through the `TelemetryDatabase` actor.

**Concerns:** none.

### Code Quality
- `turnIndex` binary search returns `max(0, low-1)` ∈ `[0, boundaries.count-1]` — safe against the `sums` array; rows before the first prompt fold into turn 0, rows at a boundary timestamp join the turn that prompt opens (correct semantics). `TelemetryDatabase.swift:376`
- Statement lifecycle is sound: `eventStmt` is finalized before the early `return []` on empty boundaries; the metric `stmt` uses `defer`; a failed `prepare` returns `[]` with a nil handle. No leaks.
- Sum-invariant holds by construction: `turnAnalytics` and `sessionAnalytics` read the same stored delta rows with matching type-filter semantics (`json_extract` NULL ↔ switch `default`), and it's guarded by a regression test (`tokensBeforeFirstPromptFoldIntoTurnZero`).
- `perTurnInputTrendSlope` denominator `n·Σx² − (Σx)²` is nonzero for the distinct integer x's at n ≥ 3 — no divide-by-zero; nil below 3 turns is tested.
- Degradation to `[]` (aged-out / telemetry-off) is documented on both the DB method and the `TelemetryService` passthrough, and tested; callers fall back to the `promptCount` average.
- Minor (non-blocking): the comment at `TelemetryDatabase.swift:301-302` says "truncate once at the end," but `Int()` is applied per-turn (`:332`). Harmless — token deltas are integral, so Σ of the per-turn truncations equals `Int(Σ)` — but the comment slightly overstates. Worth a one-word tweak next time this file is touched; not worth a round trip on its own.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 1 Green] findings. Well-scoped, well-tested, cleanly degrading reader with no behavior change to the shipped grade.

---

PR #715, opened 2026-07-13
feat(web-ui): search/filter bar on Tickets & Reviews boards (#714)

dgershman:
Closes #714

Brings the #701 Allowlist search/filter bar to the **Tickets** and **Reviews** web boards so all three boards filter consistently.

## What changed
- **Shared helper `boardFilterInput(cls, value, placeholder, onValue)`** — extracts #701's filter input (with its focus + caret-restoration across the per-keystroke re-render) so the three boards share the code, not just the look. `renderAllowlist` now calls it too (behavior identical).
- **Tickets** (`renderTicketBoard`): case-insensitive search across **title / repo / #number / labels**, **composed with** the existing status pipeline filter. Filtering + selection pruning moved up front so the Select button, pruned `selectedIssueIDs`, and Start Working only ever act on the **visible** set — no hidden/stale ticket can be started. Placeholder "Filter tickets…".
- **Reviews** (`renderReviewBoard`): case-insensitive search across **title / repo / @author / #pr_number**. Placeholder "Filter reviews…".
- Clearing either input restores the full (status-filtered) list; empty states read "No matching …" while searching.
- CSS: `.allow-filter` → `.board-filter` (single source of truth for all three inputs).

## Acceptance criteria
- [x] Tickets search filters by title/repo/number/labels, on top of the status pipeline filter; clearing restores.
- [x] Reviews search filters by title/repo/author/number; clearing restores.
- [x] Selection + Start Working (tickets) and review actions operate on the filtered view — no stale/hidden items.
- [x] Look & behavior consistent with #701 (shared input styling/UX).

## Verification
- `node --check app.js` passes.
- Manual: filter each board; toggle Tickets Select mode, tick rows, filter some out → count/Start Working act only on visible; clear restores. Allowlist unchanged (regression).

[🐦‍⬛ Generated with Crow]

Issue #714, opened 2026-07-13
Web boards: add the #701 search/filter bar to the Tickets and Reviews boards


dgershman:
Build the same **search / filter bar** added to the Allowlist board in #701 for the **Tickets** board and the **Reviews** board in the web UI, so all three boards filter consistently.

Reference: #701 (already implemented/closed) added a case-insensitive filter input to the Allowlist board (`renderAllowlist`), mirroring the desktop `SearchField("Filter patterns…")`. Reuse that same input styling/UX so the three boards match.

## Tickets board
`renderTicketBoard(root)` — `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js:1858`.
- Add a module-level `ticketSearch` string (alongside `ticketFilter`, `:493`).
- Render a filter text input in/next to the `board-head` (`:1867`), placeholder e.g. "Filter tickets…".
- Filter the `issues` list by a case-insensitive substring across the ticket's text (title, repo, number, and ideally labels) — **composed with** the existing status pipeline filter (`ticketFilter`, `:1914-1915`), i.e. status filter AND search both apply.
- Keep selection working on the filtered set: stale selections are already pruned (`:1864-1865`); ensure Select / Start Working operate on what's visible.
- Clearing the input restores the full (status-filtered) list.

## Reviews board
`renderReviewBoard(root)` — dispatched at `app.js:1795`; reviews come from `boardData.reviews.reviews` (each has `repo`, etc. — see `:343`).
- Add a `reviewSearch` string and a filter input in the reviews `board-head`, placeholder e.g. "Filter reviews…".
- Filter the reviews list by case-insensitive substring across the review's text (title, repo, author, PR number).
- Clearing restores the full list; review actions (Start Review / open) still work on the filtered view.

## Acceptance criteria
- The **Tickets** board has a search bar that filters visible tickets by case-insensitive substring (title / repo / number / labels), applied **on top of** the existing status pipeline filter; clearing it restores the list.
- The **Reviews** board has the same search bar filtering reviews (title / repo / author / number); clearing restores the list.
- Selection + Start Working (tickets) and review actions operate correctly on the filtered view — no stale/hidden items acted on.
- Look and behavior are consistent with the Allowlist search from #701 (shared input styling/UX).

## Target branch
Web UI — base on `feature/crow-593-web-ui-parity` (builds on #701's search), not `main`.

[🐦‍⬛ Created with Crow via Cursor](https://github.com/radiusmethod/crow)

Issue #701, opened 2026-07-13
Web Allowlist: fade already-global entries (parity with main) + add a filter/search bar


dgershman:
Two improvements to the **Allowlist** board in the web UI, bringing it to parity with the desktop `AllowListView` on `main`.

## 1. Fade entries already in the global allowlist
Entries that are already enabled in the global list should render **slightly faded / muted** (rather than only being hide-able), so the ones **not** yet promoted stand out and are obviously the actionable ones.

- Web today: `renderAllowlist(root)` (`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js:2053`) knows each entry's `is_global` flag (used at `:2056`/`:2059`) but only offers a **"Hide Global"** toggle (`:2063-2064`) — there's no faded treatment when globals are shown.
- Mirror `main`'s desktop `AllowListView` (`Packages/CrowUI/Sources/CrowUI/AllowListView.swift`): global rows use `.opacity(entry.isInGlobal ? 0.6 : 1.0)` (`:152`) and `.foregroundStyle(entry.isInGlobal ? textMuted : textPrimary)` (`:140`).
- Fix: in `renderAllowlist`, add a faded/muted class to each `.allow-row` when `e.is_global` (e.g. `.allow-row.is-global { opacity: .6; }` + muted `.allow-pattern` color; CSS around `app.css:523-530`). Keep globals **shown** by default (faded); the existing "Hide Global" toggle can remain as an option. Non-global (promotable) rows stay at full emphasis so they're visually obvious.

## 2. Add a search/filter bar
Add a filter input to the allowlist board that filters entries by pattern.

- Mirror desktop's `SearchField("Filter patterns…")` (`AllowListView.swift:83`) with case-insensitive substring matching on `pattern` (`AllowListView.swift:22-24`: `pattern.localizedCaseInsensitiveContains(searchText)`).
- Fix: render a text input above `.allow-list` in `renderAllowlist`; filter the `entries` list by a case-insensitive substring on `e.pattern` before rendering. Keep it working alongside the existing "Hide Global" toggle and the Promote-to-Global / Select-all / Clear actions — prune any now-hidden selections the same way the board already prunes stale selections.

## Acceptance criteria
- Allowlist entries already in the global list render faded (~0.6 opacity + muted text), matching `main`/desktop; non-global (promotable) entries render at full emphasis and clearly stand out.
- A search/filter bar above the list filters entries by pattern (case-insensitive substring); clearing it restores the full list.
[...]

PR #720, opened 2026-07-14
Restore Tickets/Reviews board search filter (#715)

dgershman:
## Summary
- Re-applies the `#715` / `#714` board search/filter work that was dropped when `feature/crow-593-web-ui-parity` was reset to the rebased backup and force-pushed.
- Cherry-picks `e8a7038` onto current parity: shared `boardFilterInput` helper + Tickets/Reviews filter bars (Allowlist keeps using the same helper).

## Test plan
- [ ] Open Tickets board → “Filter tickets…” filters by title/repo/#/labels and composes with status filter
- [ ] Multi-select / Start Working only acts on the visible filtered set
- [ ] Open Reviews board → “Filter reviews…” filters by title/repo/@author/#pr
- [ ] Allowlist still filters via the shared helper (“Filter patterns…”)
- [ ] Clearing a filter restores the full (status-filtered) list

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -521,13 +521,13 @@ html, body {
 }
 
 .allow-list { display: flex; flex-direction: column; gap: 4px; }
-.allow-filter {
+.board-filter {
   width: 100%; box-sizing: border-box; margin-bottom: 8px;
   padding: 7px 10px; border: 1px solid var(--border-subtle); border-radius: 6px;
   background: var(--bg-card); color: var(--text-primary); font-size: 12px;
 }
-.allow-filter::placeholder { color: var(--text-muted); }
-.allow-filter:focus { outline: none; border-color: var(--border-strong); }
+.board-filter::placeholder { color: var(--text-muted); }
+.board-filter:focus { outline: none; border-color: var(--border-strong); }
 .allow-row {
   display: flex; align-items: flex-start; gap: 10px; padding: 8px 10px;
   border: 1px solid var(--border-subtle); border-radius: 6px; background: var(--bg-card);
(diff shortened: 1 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Pure client-side filtering over already-fetched board data; no new network, storage, or auth surface.
- Search matching is done with `String.includes` on lowercased text — no `innerHTML`/regex construction, so no injection or ReDoS risk. All rendering continues through the existing `el()`/`ticketCard()`/`reviewCard()` DOM builders (text nodes, not raw HTML).

**Concerns:**
- None.

### Code Quality
- Clean generalization: `boardFilterInput(cls, value, placeholder, onValue)` (app.js:1863) factors #701's allowlist filter into a shared helper reused by Tickets, Reviews, and Allowlist. CSS `.allow-filter` → `.board-filter` rename is complete with no dangling references.
- Focus/caret restore after the per-keystroke `renderBoard()` is correct — each board passes a distinct `cls`, and only one board is mounted at a time, so `document.querySelector('.' + cls)` resolves unambiguously.
- Ticket board filtering (status + search) is hoisted above the header (app.js:1892–1901) so `selectableUrls`, the Select button, selection pruning, and the list all operate on the same visible set — upholding the "no hidden ticket can be started" invariant. Selection pruning against the visible set is intentional and matches the PR's test plan.
- Filter inputs are appended before the empty-state early return on both boards, so filtering to zero results still leaves the input mounted to be cleared.
- `ticketHaystack`/`reviewHaystack` are hoisted function declarations, safe despite forward use.

Green-level consideration (non-blocking): the haystacks stringify absent fields as `"undefined"` (e.g. `'#' + i.number`), so a search for `undef` could match malformed rows. In practice title/repo/number/author are always populated and `labels` defaults to `[]`, so this never surfaces on real data — noting only for awareness.

### Summary Table

---

Issue #696, opened 2026-07-13
Alignment / KPI mapping: Jira priority + epic-parent, org-goal tagging, alignment weight


dhilgaertner:
ADR 0008 follow-up **8 of 11** ([ADR](https://github.com/corveil/crow/blob/main/docs/adr/0008-ai-usage-efficiency-scorecard.md), refs #648, design PR #657). **Blocks v2 — greenfield; the main new capture in the rubric.**

No epic/parent/priority/goal concept exists anywhere in the codebase. Sessions store `ticketURL`/`ticketTitle`/`ticketNumber`; `JiraTaskBackend` reads key/summary/status/assignee/labels but not priority or epic-link.

**Scope:**
- Extend `JiraTaskBackend` to read priority + epic/parent.
- Org-goal tagging on sessions.
- An alignment weight derived from those (category C of #648's rubric: does the work ladder up to an org KPI/goal).

**Why it matters:** named trigger condition for the v2 multiplicative score, and the fix for the v1 consequence that "done ≠ value" — closing low-value tickets counts the same as closing important ones until this exists. Builds on the backend protocols from ADR 0005.

PR #725, opened 2026-07-14
feat(web): org-goal tagging on sessions (#723)

### LINKED ISSUE #723: Web UI: org-goal tagging on sessions (ADR 0008 alignment)
## Summary

Expose **org-goal tagging** in the web UI so sessions can be marked against an org KPI/goal (ADR 0008 follow-up 8 / #696).

Backend already exists on the parity branch after rebase onto main:

- `Session.orgGoal`, `ticketPriority`, `alignmentWeight`
- `SessionService.setOrgGoal`
- CLI: `crow set-goal`

There is **no web control** to view or edit the goal, and goal/alignment are not clearly surfaced in session chrome.

## Why

Alignment weighting feeds the scorecard’s combined score path. Web-only users currently can’t tag sessions without the CLI.

## Work

- [ ] RPC (or extend an existing mutator) for set/clear org-goal if not already reachable from web auth paths
- [ ] Include `org_goal` (and optionally computed alignment weight / priority) in session payloads to the web client
- [ ] UI: set/clear goal on session (detail menu or header affordance); show current tag when set
- [ ] Optional: light indication on sidebar cards when a goal is tagged
- [ ] Wire to existing `crow set-goal` semantics (clear flag, persistence, `updatedAt`)

## Refs

- ADR 0008 follow-up 8 (alignment / org-goal)
- #696 and related main PRs
- CLI: `Packages/CrowCLI/.../MetadataCommands.swift` (`set-goal`)
- Parent parity: #593 / #581

## Test plan

- [ ] Set goal in web → persists across reload / get-state
- [ ] Clear goal in web → tag removed
- [ ] CLI `crow set-goal` and web stay consistent
- [ ] Session with goal shows the tag in the UI

dgershman:
Closes #723

## What

Exposes **org-goal tagging** in the web UI (ADR 0008 follow-up 8 / #696) and fixes the missing RPC that made `crow set-goal` a dead call.

The org-goal data model (`Session.orgGoal` / `ticketPriority` / `alignmentWeight`), `SessionService.setOrgGoal`, and the `crow set-goal` CLI all shipped with #696 — but **no RPC ever routed the `set-goal` method**, so both the CLI and the web had no reachable path to it.

## Changes

- **`EngineRouter.swift`** — register the missing `set-goal` handler. The daemon's Unix socket (CLI) and web WebSocket share one `commandRouter`, so this single handler wires both. Validation mirrors the CLI's `validateSetGoal` so the two surfaces enforce one contract — reject **both** / **neither** / a **blank** goal (a missing or typo'd param errors instead of silently wiping the tag), and hold the goal to the same `isValidSessionName` bound (≤256 chars, no control chars) the sibling free-text handlers keep. The manager session is excluded server-side. Delegates to the existing `SessionService.setOrgGoal`.
- **`EngineRouter.swift` (`get-session`)** — include `org_goal`, `ticket_priority`, and the computed `alignment_weight`, so a goal set from the web reads back via `crow get-session` (matches `docs/cli-reference.md`).
- **`RPCHandlers.swift`** — include **`org_goal`** in the web `list-sessions` payload (the source the web renders sessions from). `ticket_priority` / `alignment_weight` are intentionally **not** sent here — nothing on the web renders them, so shipping them in every poll to every client would be dead payload; they're surfaced via `get-session` instead.
- **`app.js` / `app.css`** — set / edit / clear the org goal from the session context menu (non-manager sessions), a clickable 🎯 goal row in the detail header, and a light 🎯 badge on sidebar cards when tagged. Reuses the existing `textPrompt`, `rpc`, and re-render helpers (modeled on `renameSession`, including its no-op guard).
- **Docs** — document `crow set-goal` in `docs/cli-reference.md` and `CLAUDE.md`'s Metadata Commands.

## Tests

- **`EngineRouterSmokeTests.swift`** — a headless router test locks in the newly-wired `set-goal` route *and* its full contract: route reaches the mutator, `clear`, and every rejection shape (both / neither / blank / over-length / control-char / unknown-session / manager) asserted to leave the tag **unchanged**.

## Test plan (from #723)

- [x] `make build` — `crow` + `crowd` compile/link cleanly.
- [x] `set-goal` RPC route + validation contract covered by `EngineRouterSmokeTests` (3/3 green).
- [ ] Set goal in web → 🎯 shows in header + sidebar; persists across reload / get-state. *(manual browser round-trip)*
- [ ] Clear goal in web → tag removed. *(manual)*
- [ ] `crow set-goal --session <id> --goal "X"` then reload web → shows "X" (CLI/web parity, now that the RPC exists). *(manual)*
- [ ] `crow set-goal --session <id> --clear` → tag gone. *(manual)*

https://claude.ai/code/session_01Gci8z5vL1jBv3YNn8eUD5f

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -28,6 +28,7 @@ crow delete-session --session <uuid>            → {"deleted":true}
 ### Metadata Commands
 ```
 crow set-ticket --session <uuid> --url "..." [--title "..."] [--number N]
+crow set-goal --session <uuid> --goal "..." | --clear                  → tag the session's org goal/KPI (feeds alignment weight; exactly one of --goal/--clear)
 crow add-link --session <uuid> --label "Issue" --url "..." --type ticket|pr|repo|custom
 crow list-links --session <uuid>
 crow transition-ticket --session <uuid> --to inProgress|inReview|done   → moves the linked ticket to a pipeline status (Jira honors jiraStatusMap)
--- a/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
@@ -157,6 +157,14 @@ func makeCommandRouter(
                         "ticket_badge": session.ticketBadgeLabel.map { .string($0) } ?? .null,
                         "provider": session.provider.map { .string($0.rawValue) } ?? .null,
                         "review_author": session.reviewAuthor.map { .string($0) } ?? .null,
+                        // Org-goal tag + alignment inputs (#723; ADR 0008
+                        // follow-up 8). `org_goal` drives the web tag display;
+                        // `ticket_priority` and the computed `alignment_weight`
+                        // are surfaced for parity with the model (untagged
+                        // sessions weight exactly 1.0).
+                        "org_goal": session.orgGoal.map { .string($0) } ?? .null,
+                        "ticket_priority": session.ticketPriority.map { .string($0.rawValue) } ?? .null,
+                        "alignment_weight": .double(session.alignmentWeight),
                     ]
                     if let worktree = appState.primaryWorktree(for: session.id) {
                         object["repo"] = .string(worktree.repoName)
--- a/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
@@ -157,14 +157,14 @@ func makeCommandRouter(
                         "ticket_badge": session.ticketBadgeLabel.map { .string($0) } ?? .null,
                         "provider": session.provider.map { .string($0.rawValue) } ?? .null,
                         "review_author": session.reviewAuthor.map { .string($0) } ?? .null,
-                        // Org-goal tag + alignment inputs (#723; ADR 0008
-                        // follow-up 8). `org_goal` drives the web tag display;
-                        // `ticket_priority` and the computed `alignment_weight`
-                        // are surfaced for parity with the model (untagged
-                        // sessions weight exactly 1.0).
+                        // Org-goal tag (#723; ADR 0008 follow-up 8) — drives the
+                        // web tag display (sidebar badge + detail header). The
+                        // computed `alignment_weight` / `ticket_priority` are
+                        // intentionally NOT sent: nothing on the web renders
+                        // them today, so shipping them in every poll to every
+                        // client was dead payload. A future consumer (scorecard
+                        // / session strip) can add them back alongside its use.
                         "org_goal": session.orgGoal.map { .string($0) } ?? .null,
-                        "ticket_priority": session.ticketPriority.map { .string($0.rawValue) } ?? .null,
-                        "alignment_weight": .double(session.alignmentWeight),
                     ]
                     if let worktree = appState.primaryWorktree(for: session.id) {
                         object["repo"] = .string(worktree.repoName)
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -199,8 +199,9 @@ html, body {
 .activity-badge { font-size: 10px; font-weight: 700; text-transform: uppercase; letter-spacing: 0.4px; }
 /* Org-goal tag (#723): the sidebar glyph reuses .badge sizing (non-clickable);
    the detail-header .meta-goal row shows the full text and turns gold on hover
-   to read as clickable. `cursor` inherits, so the clickable header pill needs
-   its own `pointer` to beat the sidebar glyph's `default` below. */
+   to read as clickable. The header pill needs its own `pointer` because the
+   base `.goal-badge { cursor: default }` (below) would otherwise apply to it —
+   `.meta-goal .goal-badge` (0,2,0) wins on specificity over `.goal-badge` (0,1,0). */
 .goal-badge { cursor: default; }
 .meta-goal { display: flex; align-items: center; }
 .meta-goal .goal-badge { color: var(--gold); background: rgba(221, 196, 130, 0.14); border-radius: 10px; padding: 2px 8px; font-size: 11px; cursor: pointer; }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -197,11 +197,13 @@ html, body {
   font-size: 10px; font-weight: 600; border: 1px solid; background: var(--bg-surface);
 }
 .activity-badge { font-size: 10px; font-weight: 700; text-transform: uppercase; letter-spacing: 0.4px; }
-/* Org-goal tag (#723): the sidebar glyph reuses .badge sizing; the detail-header
-   .meta-goal row shows the full text and turns gold on hover to read as clickable. */
+/* Org-goal tag (#723): the sidebar glyph reuses .badge sizing (non-clickable);
+   the detail-header .meta-goal row shows the full text and turns gold on hover
+   to read as clickable. `cursor` inherits, so the clickable header pill needs
+   its own `pointer` to beat the sidebar glyph's `default` below. */
 .goal-badge { cursor: default; }
 .meta-goal { display: flex; align-items: center; }
-.meta-goal .goal-badge { color: var(--gold); background: rgba(221, 196, 130, 0.14); border-radius: 10px; padding: 2px 8px; font-size: 11px; }
+.meta-goal .goal-badge { color: var(--gold); background: rgba(221, 196, 130, 0.14); border-radius: 10px; padding: 2px 8px; font-size: 11px; cursor: pointer; }
 .meta-goal:hover .goal-badge { background: rgba(221, 196, 130, 0.24); }
 .empty { color: var(--text-muted); padding: 20px 6px; }
 
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -197,6 +197,12 @@ html, body {
   font-size: 10px; font-weight: 600; border: 1px solid; background: var(--bg-surface);
 }
 .activity-badge { font-size: 10px; font-weight: 700; text-transform: uppercase; letter-spacing: 0.4px; }
+/* Org-goal tag (#723): the sidebar glyph reuses .badge sizing; the detail-header
+   .meta-goal row shows the full text and turns gold on hover to read as clickable. */
+.goal-badge { cursor: default; }
+.meta-goal { display: flex; align-items: center; }
+.meta-goal .goal-badge { color: var(--gold); background: rgba(221, 196, 130, 0.14); border-radius: 10px; padding: 2px 8px; font-size: 11px; }
+.meta-goal:hover .goal-badge { background: rgba(221, 196, 130, 0.24); }
 .empty { color: var(--text-muted); padding: 20px 6px; }
 
 /* ---- Sidebar skeleton (first paint before list-sessions, CROW-613) ---- */
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1682,11 +1682,8 @@ async function setSessionGoal(id, current) {
   const raw = await textPrompt(current ? 'Edit org goal' : 'Set org goal', current || '',
     { placeholder: 'e.g. Q3 latency KPI', okLabel: 'Save' });
   if (raw == null) return; // cancelled
-  const goal = raw.trim();
   try {
-    await rpc('set-goal', goal ? { session_id: id, goal } : { session_id: id, clear: true });
-    const s = sessions.find((x) => x.id === id);
-    if (s) { s.org_goal = goal || null; renderSidebar(); if (id === selectedId) renderHeader(s); }
+    await applyOrgGoal(id, raw.trim());
   } catch (e) {
     alertModal('Set goal failed: ' + (e.message || e));
   }
@@ -1694,12 +1691,21 @@ async function setSessionGoal(id, current) {
 
 async function clearSessionGoal(id) {
   try {
-    await rpc('set-goal', { session_id: id, clear: true });
-    const s = sessions.find((x) => x.id === id);
-    if (s) { s.org_goal = null; renderSidebar(); if (id === selectedId) renderHeader(s); }
+    await applyOrgGoal(id, '');
   } catch (e) {
     alertModal('Clear goal failed: ' + (e.message || e));
   }
+}
+
+// Shared org-goal mutation: a non-empty `goal` sets it, an empty string clears
+// it (RPC gets `{goal}` or `{clear:true}` — never a blank goal, which the
+// handler rejects). Reflects the change locally so the sidebar + header update
+// without a refetch; `alignment_weight` is server-computed and refreshes on the
+// next `list-sessions` poll (nothing renders it client-side).
+async function applyOrgGoal(id, goal) {
+  await rpc('set-goal', goal ? { session_id: id, goal } : { session_id: id, clear: true });
+  const s = sessions.find((x) => x.id === id);
+  if (s) { s.org_goal = goal || null; renderSidebar(); if (id === selectedId) renderHeader(s); }
 }
 
 async function deleteSession(id, name) {
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1047,6 +1047,13 @@ function sessionRow(s) {
 
   const badges = el('div', 'row-badges');
   if (s.ticket_badge) badges.appendChild(el('span', 'badge', s.ticket_badge));
+  // Light org-goal indicator (#723) — glyph only to keep the card compact; the
+  // full goal text lives in the tooltip and the detail header.
+  if (s.org_goal) {
+    const g = el('span', 'badge goal-badge', '🎯');
+    g.title = 'Goal: ' + s.org_goal;
+    badges.appendChild(g);
+  }
   // PR badge — shown whenever a PR link exists (stored, or live from the app
   // when it's only in memory); colored by live status when available.
   const prLink = (s.links || []).find((l) => l.type === 'pr') || liveFor(s.id).pr_link;
@@ -1205,6 +1212,13 @@ function sessionMenuItems(s) {
   if (s.ticket_url) items.push({ label: 'Copy issue link', action: () => copyToClipboard(s.ticket_url) });
   if (prUrl) items.push({ label: 'Copy PR link', action: () => copyToClipboard(prUrl) });
   if (s.ticket_url || prUrl) items.push({ sep: true });
+  // Org-goal tagging (#723) — any non-manager session can ladder its work up to
+  // an org KPI/goal. Managers are excluded from PR/issue tracking, so no goal.
+  if (s.kind !== 'manager') {
+    items.push({ label: s.org_goal ? 'Edit org goal…' : 'Set org goal…', action: () => setSessionGoal(s.id, s.org_goal) });
+    if (s.org_goal) items.push({ label: 'Clear org goal', action: () => clearSessionGoal(s.id) });
+    items.push({ sep: true });
+  }
   const hasPR = (s.links || []).some((l) => l.type === 'pr');
   if (s.kind === 'manager') {
     // Maintenance actions (restart manager / reload tmux) live in Settings → About;
@@ -1404,6 +1418,16 @@ function renderHeader(s) {
   root.appendChild(top);
 
   if (s.ticket_title) root.appendChild(el('div', 'subtle', s.ticket_title));
+  // Org-goal tag (#723) — click to edit/clear. Only rendered when tagged; the
+  // menu's "Set org goal…" is the entry point for untagged sessions.
+  if (s.org_goal) {
+    const goalRow = el('div', 'meta meta-goal');
+    goalRow.appendChild(el('span', 'goal-badge', '🎯 ' + s.org_goal));
+    goalRow.title = 'Org goal — click to edit';
+    goalRow.style.cursor = 'pointer';
+    goalRow.onclick = (ev) => { ev.stopPropagation(); setSessionGoal(s.id, s.org_goal); };
+    root.appendChild(goalRow);
+  }
   // Review sessions: surface the PR author. Prefer the live review request
   // (Reviews board), but fall back to the author persisted on the session at
   // review-creation so it still shows when the board is empty (CROW-593).
@@ -1648,6 +1672,33 @@ async function renameSession(id, current) {
     if (s) { s.name = name; renderSidebar(); if (id === selectedId) renderHeader(s); }
   } catch (e) {
     if (term) term.write('\r\n\x1b[31m[crow] rename failed: ' + (e.message || e) + '\x1b[0m\r\n');
+  }
+}
+
+// Set or update a session's org-goal tag (#723). Prompts free-text; an empty
+// value clears the tag (parity with `crow set-goal --clear`). Updates the local
+// session so the sidebar badge + detail header reflect it without a refetch.
+async function setSessionGoal(id, current) {
+  const raw = await textPrompt(current ? 'Edit org goal' : 'Set org goal', current || '',
+    { placeholder: 'e.g. Q3 latency KPI', okLabel: 'Save' });
+  if (raw == null) return; // cancelled
+  const goal = raw.trim();
+  try {
+    await rpc('set-goal', goal ? { session_id: id, goal } : { session_id: id, clear: true });
+    const s = sessions.find((x) => x.id === id);
+    if (s) { s.org_goal = goal || null; renderSidebar(); if (id === selectedId) renderHeader(s); }
+  } catch (e) {
+    alertModal('Set goal failed: ' + (e.message || e));
+  }
+}
+
+async function clearSessionGoal(id) {
+  try {
+    await rpc('set-goal', { session_id: id, clear: true });
+    const s = sessions.find((x) => x.id === id);
+    if (s) { s.org_goal = null; renderSidebar(); if (id === selectedId) renderHeader(s); }
+  } catch (e) {
+    alertModal('Clear goal failed: ' + (e.message || e));
   }
 }
 
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1700,8 +1700,7 @@ async function clearSessionGoal(id) {
 // Shared org-goal mutation: a non-empty `goal` sets it, an empty string clears
 // it (RPC gets `{goal}` or `{clear:true}` — never a blank goal, which the
 // handler rejects). Reflects the change locally so the sidebar + header update
-// without a refetch; `alignment_weight` is server-computed and refreshes on the
-// next `list-sessions` poll (nothing renders it client-side).
+// without a refetch.
 async function applyOrgGoal(id, goal) {
   await rpc('set-goal', goal ? { session_id: id, goal } : { session_id: id, clear: true });
   const s = sessions.find((x) => x.id === id);
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1682,8 +1682,10 @@ async function setSessionGoal(id, current) {
   const raw = await textPrompt(current ? 'Edit org goal' : 'Set org goal', current || '',
     { placeholder: 'e.g. Q3 latency KPI', okLabel: 'Save' });
   if (raw == null) return; // cancelled
+  const goal = raw.trim();
+  if (goal === (current || '')) return; // unchanged (or empty on an untagged session) — skip the write
   try {
-    await applyOrgGoal(id, raw.trim());
+    await applyOrgGoal(id, goal);
   } catch (e) {
     alertModal('Set goal failed: ' + (e.message || e));
   }
--- a/Packages/CrowEngine/Sources/CrowEngine/EngineRouter.swift
+++ b/Packages/CrowEngine/Sources/CrowEngine/EngineRouter.swift
@@ -592,6 +592,31 @@ public func makeEngineRouter(_ ctx: EngineContext) -> CommandRouter {
                     return ["session_id": .string(idStr)]
                 }
             },
+            // Set or clear the org-goal tag on a session (#723; ADR 0008
+            // follow-up 8). The data model + `SessionService.setOrgGoal` and the
+            // `crow set-goal` CLI shipped with #696, but no RPC ever routed the
+            // method — this wires the CLI socket *and* the web WebSocket (both
+            // share this router) to the mutator. `clear` wins over `goal`; a
+            // blank/whitespace goal is treated as a clear so an empty tag can't
+            // buy the on-goal alignment multiplier.
+            "set-goal": { @Sendable params in
+                guard let idStr = params["session_id"]?.stringValue, let id = UUID(uuidString: idStr) else {
+                    throw RPCError.invalidParams("session_id required")
+                }
+                let clear = params["clear"]?.boolValue ?? false
+                let goal: String? = {
+                    if clear { return nil }
+                    let trimmed = (params["goal"]?.stringValue ?? "").trimmingCharacters(in: .whitespacesAndNewlines)
+                    return trimmed.isEmpty ? nil : trimmed
+                }()
+                return try await MainActor.run {
+                    guard capturedAppState.sessions.contains(where: { $0.id == id }) else {
+                        throw RPCError.applicationError("Session not found")
+                    }
+                    capturedService.setOrgGoal(id: id, goal: goal)
+                    return ["session_id": .string(idStr), "org_goal": goal.map { .string($0) } ?? .null]
+                }
+            },
             "transition-ticket": { @Sendable params in
                 // CROW-529: transition a session's linked ticket to a pipeline
                 // status (honoring jiraStatusMap for Jira). `setup.sh` calls this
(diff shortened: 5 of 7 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Focused, well-commented PR that fixes a real dead-call bug: the `set-goal` method had a data model, `SessionService.setOrgGoal`, and a CLI command (all from #696) but no RPC ever routed it. Wiring one handler into the shared `commandRouter` correctly fixes both the CLI socket and the web WebSocket. Verified `swift build` compiles cleanly (the one warning is pre-existing, in an untouched file).

### Critical Issues
None.

### Security Review
**Strengths:**
- No XSS. All goal text reaches the DOM via `el()`'s `textContent` and `.title`/`.style` properties (`app.js:1050`, `1424`) — never `innerHTML`.
- Server-side validation is sound: `set-goal` guards `session_id` shape and existence before mutating (`EngineRouter.swift:603-610`), returning `invalidParams`/`applicationError` on bad input.
- Normalization is defensive and matches the model: `clear` wins over `goal`, and a blank/whitespace goal is trimmed to a clear, so an empty tag can't buy the on-goal alignment multiplier (`EngineRouter.swift:611-615`) — consistent with `Session.alignmentWeight`.

**Concerns:**
- None.

### Code Quality
- 🟡 **No test for the new `set-goal` RPC route.** `EngineRouterSmokeTests.swift` already provides a lightweight harness (`AppState` + `JSONStore` + `NoopHostBridge` → `makeEngineRouter` → `router.handle`). This PR's whole premise is that a missing route silently no-op'd with nothing to catch it — a smoke test dispatching `set-goal` and asserting `orgGoal` changed (plus the `clear`-precedence and blank→clear normalization) would lock in exactly that regression class. `EngineRouter.swift:602`
- 🟢 `alignment_weight` and `ticket_priority` are added to the `list-sessions` payload (`RPCHandlers.swift:166-167`) but not consumed anywhere in `app.js`. Intentional per the PR ("parity with the model") and harmless — noting only so it's a deliberate choice, not an oversight.
- 🟢 On web set/clear, the local session's `alignment_weight` isn't recomputed client-side; it refreshes on the next `list-sessions` poll. Not user-visible since the web doesn't render it. Fine as-is.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The Yellow (regression test for the newly-wired route) is a same-round-trip add; the implementation itself is correct and ships-ready otherwise.

---

dgershman:
Addressed the Yellow finding (regression test for the newly-wired `set-goal` route). Pushed `f970b98`.

Added one `@Test` to `EngineRouterSmokeTests.swift` using the existing `AppState`/`JSONStore`/`NoopHostBridge` → `makeEngineRouter` → `router.handle` harness. It seeds a session and dispatches `set-goal`, asserting on the resulting `orgGoal` mutation to prove the route reaches `SessionService.setOrgGoal`, and locks in exactly the regression class this PR fixes:

1. **Route reaches the mutator** — `goal: "Q3 revenue"` → `orgGoal == "Q3 revenue"`, no error.
2. **`clear` wins over `goal`** — `clear: true` + `goal: "ignored"` → `orgGoal == nil`.
3. **Blank/whitespace → clear** — re-set a goal, then `goal: "   "` → `orgGoal == nil` (empty tag can't buy the on-goal multiplier).
4. **Unknown session** — random UUID → `applicationError("Session not found")`, not a silent no-op.

`swift test --package-path Packages/CrowEngine --filter EngineRouterSmokeTests` → 3/3 green. Test-only change; production binaries unaffected and CI builds them.

Left the two Green non-blocking items as-is per your review (`alignment_weight`/`ticket_priority` in the payload are deliberate model parity; client-side `alignment_weight` refreshes on the next `list-sessions` poll).

Re-requesting review.

dhilgaertner (changes requested):
## Code & Security Review

The core premise checks out. I confirmed the missing-RPC diagnosis is real and that the fix reaches both surfaces: `commandRouter` (`RPCHandlers.makeCommandRouter`) backs *both* the CLI Unix socket (`CrowDaemon.swift:231`) and the web WebSocket (`CrowDaemon.swift:250`), and delegates unowned methods to `engineFallback` = `makeEngineRouter` (`CrowDaemon.swift:196-218`). Since `set-goal` isn't owned by the curated router, both paths land on the new handler. Param names match the CLI's wire format (`session_id` / `goal` / `clear`, `MetadataCommands.swift:56-59`). Persistence goes through `store.mutate`, and the store file-watcher broadcasts `changed` (`CrowDaemon.swift:620`), so other clients re-fetch — no staleness gap.

**Verified locally:** `swift build` succeeds, and `swift test --package-path Packages/CrowEngine --filter EngineRouter` passes all 3 tests including the new `set-goal` case.

### Security Review

**Strengths:**
- **No XSS.** `el()` assigns via `textContent` (`app.js:570`), and both goal render paths go through it. The sidebar tooltip uses `g.title = 'Goal: ' + s.org_goal` — attribute assignment, not markup. Free-text goals are inert.
- **No injection.** The goal string never reaches a shell, path, or query; it terminates at `Session.orgGoal` in `state.json`.
- **Guarded before mutating.** The handler checks session existence and throws `applicationError` rather than silently no-op'ing — and the test locks that path in.
- **Concurrency model respected.** The mutation runs inside `MainActor.run`, matching every sibling handler; `store.mutate` serializes the write under `NSLock`.

**Concerns:**
- Nothing blocking. Note that `goal` is unbounded free text persisted verbatim to `state.json` — worth a length cap eventually, but the existing `set-ticket` fields have the same property, so it's not a regression this PR introduces.

### Code Quality

**Yellow — `set-goal` silently clears the tag when neither `goal` nor `clear` is supplied** (`EngineRouter.swift:602-610`)

The handler derives `goal` as `params["goal"]?.stringValue ?? ""`, then trims to `nil`. So a call with *neither* param normalizes to a clear, wipes an existing tag, and returns `{"session_id": ..., "org_goal": null}` as success. That diverges from two contracts already established in this codebase:

1. **The CLI's own validation** (`Validation.swift:89-102`) explicitly rejects exactly these shapes: `(nil, false)` → *"Exactly one of --goal or --clear is required"*; blank goal → *"--goal must not be blank"*; both supplied → *"--goal and --clear are mutually exclusive"*. The RPC accepts all three silently. The CLI's `validate()` guards only the CLI path, so for the *web* — the surface this PR exists to enable — the RPC is the sole enforcement point, and it enforces the opposite of what the CLI documents.
2. **The neighboring `set-ticket` handler** (`EngineRouter.swift:571-588`), which the PR says it modeled placement on, uses `if let url = params["url"]?.stringValue` — an absent param means *leave unchanged*, never *destroy*.

The realistic failure: an RPC caller typos the param name (`org_goal` instead of `goal`, easy given `list-sessions` now ships `org_goal`) and silently wipes the session's tag with a success response. The new test locks this behavior in rather than catching it.

Suggested fix — mirror `validateSetGoal` so the two surfaces agree:
```swift
let hasGoal = params["goal"]?.stringValue != nil
guard clear || hasGoal else {
    throw RPCError.invalidParams("Exactly one of `goal` or `clear` is required")
}
```
If the lenient normalization is deliberate, it's worth saying so in the comment and reconciling `validateSetGoal` — right now a reader can't tell which contract is authoritative.

**Green — manager exclusion is only enforced in the menu.** `sessionMenuItems` gates on `s.kind !== 'manager'` (`app.js:1217`), but `sessionRow` (`app.js:1052`) and `renderHeader` (`app.js:1423`) render the badge for any tagged session, and the header row's `onclick` calls `setSessionGoal` for managers too — as does the RPC. A manager tagged via `crow set-goal` shows a 🎯 with an editable header row but no "Clear org goal" menu item. Recoverable (clear via the edit dialog's empty value), so cosmetic — but the policy either belongs in the handler or nowhere.

**Green — `clearSessionGoal` duplicates `setSessionGoal`'s clear branch** (`app.js:1694-1702`). Same RPC, same local update, same re-render; only the error string differs. Could collapse into a shared helper.

**Green — local update skips `alignment_weight`.** Both handlers set `s.org_goal` but leave `s.alignment_weight` at its pre-edit value (`app.js:1689`, `1699`), so it's stale until the next `list-sessions`. Nothing renders it today — latent only.

**Green — test plan.** 4 of 5 boxes are unchecked; only `make build` is confirmed. The new router test covers the RPC well, but the web round-trip (the PR's actual deliverable) is still unverified.

### Summary Table

The RPC wiring is correct and the diagnosis behind this PR is sound; the one Yellow is the handler's param contract contradicting the CLI validation that already exists a few files over. Worth settling in this round since the Greens are cheap to fold in alongside it.

---

dgershman:
Addressed round 2. Pushed `cef9569`. `make build` clean; `swift test --package-path Packages/CrowEngine --filter EngineRouterSmokeTests` → 3/3 green.

**🟡 `set-goal` param contract diverges from `validateSetGoal` — FIXED.** You're right: `params["goal"]?.stringValue ?? ""` → trim → nil meant a missing/typo'd `goal` (no `clear`) silently wiped the tag and returned success. The handler now mirrors the CLI's `validateSetGoal` (`Validation.swift:89`) exactly — one authoritative contract for both surfaces:
- both `goal` + `clear` → `invalidParams` (mutually exclusive)
- neither → `invalidParams` (exactly one required) ← the silent-wipe
- blank/whitespace `goal` → `invalidParams` (must not be blank)

Confirmed this rejects nothing the web emits: `setSessionGoal` only ever sends `{goal}` (non-blank, post-`trim()`) or `{clear:true}` (`app.js:1706`), never a bare `{session_id}` or `{goal:""}`. The smoke test now asserts all three rejection shapes leave `orgGoal` **unchanged** (not cleared), plus route-works / clear / unknown-session — mirroring `ValidationTests`. Reconciled the handler comment to point at `validateSetGoal` as the shared contract.

**🟢 Manager exclusion only half-enforced — FIXED in the handler.** Moved the policy to the single authoritative point you flagged: the RPC now rejects `.manager` sessions (`applicationError`), so `crow set-goal` and the web agree, and a manager can no longer be tagged via the CLI while the web renders a half-supported badge. The menu gate (`app.js:1217`) stays as affordance-hiding; the badge/header (`1052`/`1423`) render only `if (s.org_goal)`, which can no longer become true for a manager.

**🟢 `clearSessionGoal` duplicates `setSessionGoal`'s clear branch — FIXED.** Collapsed both into a shared `applyOrgGoal(id, goal)` (empty ⇒ `{clear:true}`); each caller keeps only its own error string.

**🟢 Local update skips `alignment_weight` — leaving as-is (reply).** `alignment_weight` is server-computed from `orgGoal` + `ticketPriority` (`Session.alignmentWeight`/`CombinedScore`); recomputing it in JS would duplicate that formula client-side, which the web-parity convention explicitly avoids ("never port logic to JS"). Nothing renders it today, and it refreshes on the next `list-sessions` poll — so it's latent-only, as you noted. Documented that in the `applyOrgGoal` comment.

**🟢 Test plan / web round-trip — reply.** The router test now covers the RPC route **and** its full contract headlessly. The browser round-trip (set/clear/edit → badge in sidebar + header) is inherently manual — it needs a live `crowd` + web session — so I'd rather not tick those boxes without actually exercising them. The unbounded-`goal` length note is a good follow-up but, as you said, matches the existing `set-ticket` fields, so it's not a regression this PR introduces; happy to file it as a separate issue.

Re-requesting review.

dhilgaertner (changes requested):
## Code & Security Review

Verified the core claim end-to-end: `set-goal` really was unrouted, and this handler genuinely fixes it for **both** surfaces. Traced it rather than taking the PR body's word for it — `CrowDaemon.swift:214` builds the curated router with `fallback: engineFallback`, and that same `commandRouter` instance backs the Unix socket (`:231`) and the `/rpc` WebSocket (`:250`); `CommandRouter.handle` (CrowIPC) forwards unowned methods to the fallback, and `set-goal` is not in the daemon's handler map. So one handler does wire CLI + web. `RPCWebSocketHandler.localOnlyDenial` is a denylist, not an allowlist, so the web path isn't silently gated.

`make build` / `swift build` clean; `swift test --package-path Packages/CrowEngine --filter EngineRouterSmoke` → **3/3 pass**, including the new test. (Root `swift test` finds no target — engine tests must be run via `--package-path`.)

The validation switch is exhaustive and correct, and the test earns its keep: re-setting `"keep me"` before each malformed shape proves rejection *without* mutation, which is exactly the silent-wipe bug that made this worth fixing.

### Security Review

**Strengths:**
- **No XSS.** `el()` (app.js:567) assigns via `textContent`, and `title` is set as a property — a goal of `` renders as literal text. The one place free-text meets the DOM is safe.
- `orgGoal`'s blast radius is contained: store, analytics snapshots, and the web. It's never interpolated into a shell, tmux `send`, or terminal write, so control characters can't become a command-injection or escape-sequence vector.
- Server-side enforcement is the right call — the RPC rejecting bad shapes rather than trusting `app.js` means a hand-rolled WS frame can't wipe a tag either.

**Concerns:**

**Yellow — `goal` has no length or control-character bound.** `EngineRouter.swift:611-629` only rejects blank. Its sibling handlers in the same router hold free text to `Validation.isValidSessionName` (256 chars, no control chars) — `new-session` at `EngineRouter.swift:102`, and `rename-session` via `SessionService`. `set-goal` accepts an unbounded string with embedded control characters.

This matters *because of* this PR, not despite it: before it, no reachable path existed. It now newly exposes an unvalidated free-text field to an **authenticated remote** web client (the CLI is always local; the WS is not). A multi-megabyte goal is persisted to the store, copied into every analytics snapshot, and re-broadcast in **every** `list-sessions` payload to **every** connected client on each poll. Not privilege escalation, so not Red — but it's a cheap fix that matches the contract the neighboring handlers already keep, and it fits this PR's own stated framing that the RPC is the sole enforcement point for the web.

### Code Quality

**Yellow — `.goal-badge { cursor: default; }` defeats the header row's own click affordance.** `app.css:202` sets `cursor: default` directly on `.goal-badge`. `cursor` is an inherited property, so the explicit rule on the child beats the `cursor: pointer` that `renderHeader` sets inline on the parent `.meta-goal` row (`app.js:1427`). The gold pill *is* the visible row content, so hovering the thing that says `title="Org goal — click to edit"` shows an arrow, not a hand.

The click still works (the handler is on the row and the badge click bubbles), and `.meta-goal:hover .goal-badge` still turns gold — so this is affordance-only, but it directly contradicts the CSS comment's stated intent ("turns gold on hover to read as clickable"). The class is doing double duty: `cursor: default` is right for the sidebar glyph (`badge goal-badge`, non-clickable) and wrong for the header pill. One line:

```css
.meta-goal .goal-badge { cursor: pointer; }
```

**Green — `ticket_priority` / `alignment_weight` are dead payload.** Added to `list-sessions` (`RPCHandlers.swift:166-167`) "for parity," but `applyOrgGoal`'s own comment concedes "nothing renders it client-side." Every session in every poll carries two fields no client reads. Fine to keep if a follow-up consumes them; worth dropping otherwise.

**Green — CLI read parity is still one-sided.** The write path now reaches both surfaces, but `EngineRouter`'s `get-session` (`:184`) and its minimal `list-sessions` (`:168`, id/name/status only) still omit `org_goal` — so a goal set from the web won't show in `crow get-session`. `get-state` is unaffected (it encodes `Session` via `Codable`, so `orgGoal` rides along, and the test plan's reload/get-state check should pass).

**Green — the manager exclusion is a new rule, not a mirror.** `validateSetGoal` has no kind check, so the "mirrors the CLI's contract" comment slightly oversells it. The rule is defensible and server-side is the right home; the comment just claims more symmetry than exists.

### Summary Table

The routing fix is correct and well-tested; both Yellows are small and localized. Bound the `goal` input and free the header cursor and this is good to land.

dgershman:
Addressed round 3. Pushed `ef465d5`. `make build` clean; `swift test --package-path Packages/CrowEngine --filter EngineRouterSmokeTests` → 3/3.

**🟡 `goal` unbounded / control chars — FIXED.** Held to the same `Validation.isValidSessionName` bound (≤256 chars, no control chars) as the sibling free-text handlers you cited (`new-session` `EngineRouter.swift:101`, `rename-session` via `SessionService`), applied after the blank check:
```swift
if let goal, !Validation.isValidSessionName(goal) {
    throw RPCError.invalidParams("Invalid goal (max \(Validation.maxSessionNameLength) chars, no control characters)")
}
```
Smoke test adds an over-length (257-char) and a control-char (`\u{07}`) case, both asserting the tag is left **unchanged**. Since the RPC backs the CLI socket too, this bounds `crow set-goal` as well — one contract, both surfaces.

**🟡 `.goal-badge { cursor: default }` defeats the header affordance — FIXED.** Scoped `cursor: pointer` onto `.meta-goal .goal-badge` (specificity 0,2,0 beats the bare `.goal-badge` 0,1,0), so the header pill reads as clickable while the sidebar glyph keeps `default`. Updated the CSS comment to explain the inheritance/specificity interplay so it doesn't regress.

**🟢 Dead payload — DROPPED.** Removed `ticket_priority` and `alignment_weight` from the web `list-sessions` payload (`RPCHandlers.swift`). Confirmed nothing in `app.js` reads them; `org_goal` (the only consumed field) stays. A future scorecard/session-strip consumer can re-add them alongside its use.

**🟢 CLI read parity — FIXED.** `get-session` (`EngineRouter.swift`) now includes `org_goal`, so a goal set from the web shows in `crow get-session`. Left the minimal `list-sessions` (`:168`, id/name/status) as-is — it's deliberately minimal — and `get-state` already carries `orgGoal` via `Codable`.

**🟢 "mirrors the CLI" oversold — FIXED (comment).** Reworded: the goal/clear/blank rules mirror `validateSetGoal`; the `isValidSessionName` bound mirrors the sibling handlers; the manager exclusion is called out as a server-side-only rule with no CLI equivalent.

On the unbounded-goal note being *"because of this PR, not despite it"* — agreed, that's exactly why I bounded it here rather than deferring. Re-requesting review.

dhilgaertner (changes requested):
## Code & Security Review

Verified locally: `swift build` clean; `swift test --package-path Packages/CrowEngine` → **289 tests / 39 suites passed**, `Packages/CrowCLI` → **96 passed**, including the new `set-goal routes to the mutator and enforces the CLI's validation contract`.

I also confirmed the PR's central architectural claim rather than taking it on faith: `CrowDaemon.swift:214` builds `makeCommandRouter(..., fallback: engineFallback)` where `engineFallback = makeEngineRouter(ctx)` (`:211`), and `CommandRouter.handle` (`CrowIPC/CommandRouter.swift:28`) delegates unknown methods to the fallback. The daemon router defines neither `set-goal` nor `get-session`, so both correctly fall through to the engine router — one handler really does wire the CLI socket *and* the web WebSocket. Good design; the "missing handler is structurally impossible" property from ADR 0007 holds here.

### Critical Issues

None. No Red findings.

### Security Review

**Strengths:**
- **Validation is the right call, at the right layer.** Holding the goal to `Validation.isValidSessionName` (≤256 chars, no control characters) is well-reasoned: this newly exposes a free-text field to a *remote* authenticated web client whose value is re-broadcast in every `list-sessions` poll. Unbounded here would mean a multi-megabyte string amplified to every connected client on every poll.
- **No XSS.** `el()` (`app.js:567`) assigns via `textContent`, and the goal reaches the DOM only through `el(...)` and `.title` — never `innerHTML`. Untrusted goal text is inert.
- **Fail-closed contract.** Rejecting *both* / *neither* / *blank* means a missing or typo'd param errors instead of silently wiping an existing tag. The test asserts non-mutation on every rejected shape (`#expect(currentGoal() == "keep me")`), which is exactly the right assertion — it proves rejection, not just an error return.
- **Manager exclusion enforced server-side**, not merely hidden in the web menu — the UI rule can't be bypassed via the socket.
- The `.whitespacesAndNewlines` trim in the RPC is *stricter* than the CLI's `.whitespaces`; combined with `isValidSessionName` rejecting newlines, no whitespace-only or embedded-newline goal can reach the store. No gap.

**Concerns:** none found.

### Code Quality

**Yellow — `docs/cli-reference.md:146` documents a `get-session` contract this PR leaves two-thirds unmet.**
The doc states the org-goal tag is "read back via `crow get-session` (`org_goal`, `ticket_priority`, `alignment_weight`)". `EngineRouter.swift:203` adds only `org_goal`; neither `ticket_priority` nor `alignment_weight` appears in any `get-session` response. Concretely, this still fails:
```bash
crow set-ticket --session <id> --priority high   # documented, works
crow get-session --session <id>                  # no ticket_priority, no alignment_weight
```
`Session.alignmentWeight` (`CrowCore/Models/Session.swift:113`) is a computed property and `ticketPriority` (`:67`) is right there — this is two lines next to the one you added. The mismatch predates you (it shipped wrong with #696), but this PR edits exactly that dictionary and its own comment claims "read parity for the write path this PR wired", so parity is incomplete as stated. Either add both fields or correct the doc line.

**Yellow — the PR description contradicts the code it describes.**
The body says `RPCHandlers.swift` will "include `org_goal`, `ticket_priority`, and the computed `alignment_weight` in the web `list-sessions` payload". The code (`RPCHandlers.swift:158-167`) deliberately ships **only** `org_goal`, with a comment explaining the other two are intentionally omitted as dead payload. **The code's decision is the correct one** — don't change it; the description is stale and should be updated to match, since it lands in the merge commit and misleads the next reader.

**Green — force-unwrap in a remote-reachable handler** (`EngineRouter.swift:637`, `rawGoal!`). Provably safe today: the `(nil, false)` case throws first, so `clear == false` implies `rawGoal != nil`. But that safety lives in a switch several lines away — a future case edit turns this into a daemon crash from an RPC. Binding in the pattern (`case (.some(let g), false)`) makes it structurally safe at zero cost.

**Green — no no-op guard in `setSessionGoal`** (`app.js:1682`). `renameSession` guards with `if (!name || name === current) return;`; the goal path has no equivalent. Submitting the dialog unchanged — or submitting empty on an *untagged* session — fires an RPC that calls `setOrgGoal`, bumps `updatedAt`, and rewrites the store for zero actual change. Harmless, just avoidable churn.

**Green — the CSS comment's reasoning is wrong** (`app.css:198-201`). It attributes the override to `cursor` inheritance and rule order ("below"), but `.meta-goal .goal-badge` (specificity 0,2,0) beats `.goal-badge` (0,1,0) on **specificity**, and it already sits below. The rule is needed and correct — only the explanation misleads.

**Green — `crow set-goal` is absent from `CLAUDE.md`'s Metadata Commands** (which lists `set-ticket`, `add-link`, `list-links`, `transition-ticket`, `resync-jira`). Omitting it was defensible while the command was a dead call; this PR is what makes it real, so it now belongs in the list.

**Worth calling out:** the test's docstring — "the whole path silently no-op'd with nothing to catch it" — names the actual class of bug (a shipped CLI with no route, undetected because nothing asserted the wiring). Locking the route itself in, not just the validation, is the right fix shape.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 4 Green] findings.

Both Yellows are documentation/contract accuracy, not defects — the implementation, validation, and test coverage are genuinely good, and the security posture is the strongest part of the change. Reconcile `docs/cli-reference.md:146` with `get-session` (add the two fields, or fix the line) and refresh the PR body to match what `list-sessions` actually sends, and this is a ready approve.

dgershman:
Addressed round 4. Pushed `ad4fa0b`. `make build` clean; `swift test --package-path Packages/CrowEngine --filter EngineRouterSmokeTests` → 3/3.

**🟡 `get-session` doesn't match `docs/cli-reference.md:146` — FIXED (add the fields).** Added `ticket_priority` + `alignment_weight` to the `get-session` dict (`EngineRouter.swift`), so `crow set-ticket --priority high; crow get-session` now returns all three the doc promises. `GetSession.run()` prints the raw dict via `printJSON`, so they surface directly. This is the deliberate single-session read the doc covers — it does **not** re-add them to the web `list-sessions` poll (that stays `org_goal`-only per round 3; different path, different rationale).

**🟡 PR description contradicts the code — FIXED (update the description).** Kept the code as-is (only `org_goal` in `list-sessions`) and rewrote the PR body to match. While there I also corrected the now-stale `EngineRouter.swift` bullet, which still described the round-1 "clear wins / blank→clear" behavior — it now documents the strict validation contract. Refreshed the test-plan boxes too.

**🟢 Force-unwrap `rawGoal!` — FIXED.** Refactored the `set-goal` switch to bind the goal in-pattern; it's exhaustive over `(goal?, clear)` and the only value-yielding arm (`.some(let g), false`) validates blank + `isValidSessionName` inline before assigning. No `!`, structurally safe against a future case edit, zero behavior change (test still 3/3).

**🟢 No no-op guard in `setSessionGoal` — FIXED.** Added `if (goal === (current || '')) return;` after the trim, mirroring `renameSession` — an unchanged edit (or empty submit on an untagged session) no longer fires an RPC that bumps `updatedAt` and rewrites the store. The explicit "Clear org goal" menu item uses `clearSessionGoal`, so clearing a tagged session is unaffected.

**🟢 CSS comment reasoning wrong — FIXED.** Reworded: the header pill wins by **specificity** (`.meta-goal .goal-badge` 0,2,0 > `.goal-badge` 0,1,0), not inheritance/order. Rule unchanged.

**🟢 `crow set-goal` absent from `CLAUDE.md` — FIXED.** Added it to the Metadata Commands list, and to the `Resources/CLAUDE.md.template` scaffold (kept in sync per the dual-source convention). `docs/automation.md` is a Jira-automation walkthrough, not the command list, so left as-is.

Re-requesting review.

dhilgaertner (approved):
## Code & Security Review

Verified locally: `swift build` → **Build complete**; `swift test --package-path Packages/CrowEngine --filter EngineRouterSmokeTests` → **3/3 passed**, including the new `set-goal` contract test.

### Critical Issues

None.

### Security Review

**Strengths:**

- **Server-side enforcement is at the right layer.** `set-goal` is reachable from the web WebSocket, and `RPCWebSocketHandler.localOnlyDenial` is a *denylist* — so an authenticated remote client can call this method. Putting validation in the RPC handler (`EngineRouter.swift:643-655`) rather than only in the CLI's `validate()` is the correct choice: the CLI's `validateSetGoal` guards nothing on the web path.
- **Bounding the free-text field is the right call, and non-obvious.** The CLI's `validateSetGoal` (`Packages/CrowCLI/Sources/CrowCLILib/Validation.swift:84`) checks only shape/blankness — no length or control-char bound. Holding `goal` to `Validation.isValidSessionName` (≤256 chars, no control chars) matters precisely because this string is newly reachable from a *remote* client and is re-broadcast in every `list-sessions` poll to every connected client. Without it, one client could persist an unbounded tag that every other client then pays for. The 257-char and `\u{07}` test cases lock this in.
- **No XSS.** The goal string is rendered via `el()` (`app.js:567`), which assigns `textContent`, and the tooltip via the `.title` property — both are assignment, not parsing. A goal of `` is inert.
- **Rejections don't mutate.** Every malformed shape throws before `setOrgGoal` is reached, and the test asserts `currentGoal() == "keep me"` after each — so a missing or typo'd param errors instead of silently wiping an existing tag. That was the actual failure mode worth guarding.
- **No TOCTOU.** The session lookup, the manager check, and `setOrgGoal` all execute inside one `MainActor.run` block (`EngineRouter.swift:657-666`), so nothing interleaves between the existence check and the mutation.

**Concerns:** None found.

### Code Quality

I traced the PR's central architectural claim rather than taking it on faith, and it holds: `makeCommandRouter` does **not** own `set-goal` or `get-session`, and `CrowDaemon.swift:214-218` passes `makeEngineRouter(ctx)` as `fallback:`. Both the Unix socket (`SocketServer`, line 231) and the web (`RPCWebSocketHandler.mount`, line 250) are handed that same `commandRouter`, so the single engine handler genuinely wires both surfaces. The daemon *does* own `list-sessions`, which is why `org_goal` had to be added to `RPCHandlers.swift` separately — that split is correct, not redundant.

The client/server contracts line up on the details that usually drift:

- Menu gating on `s.kind !== 'manager'` (`app.js:1218`) matches the server's `session.kind != .manager` rejection, and `kind` is in the `list-sessions` payload (`RPCHandlers.swift:144`) — so the UI gate can't disagree with the server.
- `setSessionGoal` trims before sending and stores the same trimmed value locally (`app.js:1685`, `1709`), matching what the server persists — so the optimistic local update can't drift from the stored value.
- The empty-input-clears path is coherent: an emptied box on a tagged session sends `{clear:true}` (never a blank `goal`, which the handler rejects), while an empty box on an untagged session short-circuits on the no-op guard. The client never emits a shape the server rejects.
- `alertModal`, `textPrompt`, and `rpc` all exist with the signatures used; the optimistic-update shape mirrors `renameSession` (`app.js:1672`) exactly, including the `id === selectedId` guard.

The CSS specificity comment (`app.css:200-205`) is accurate — `.meta-goal .goal-badge` (0,2,0) does beat `.goal-badge` (0,1,0).

**Green (consider — non-blocking):**

- **PR description overstates the docs change.** The body says docs were updated in `docs/cli-reference.md`, but that file isn't in the diff — it already documented `set-goal` from #696. Worth correcting the description, though the underlying situation is a point in the PR's favor: `docs/cli-reference.md:146` *already claimed* `crow get-session` reads back `org_goal` / `ticket_priority` / `alignment_weight`, which was untrue until this PR. The `get-session` change closes real docs-drift rather than creating it.
- **Redundant cursor declaration.** `goalRow.style.cursor = 'pointer'` (`app.js:1427`) is an inline style doing what `.meta-goal .goal-badge { cursor: pointer }` already covers. Harmless; folding it into `.meta-goal` in `app.css` would keep presentation in one place.
- **`org_goal` in the daemon's `list-sessions` has no automated coverage.** Noting it only for completeness — there's no JS test infra and no existing test asserting `list-sessions` payload fields, so this is consistent with the repo rather than a gap this PR introduced. The line is a one-liner `.map`.

**On the unchecked test-plan items:** the four manual browser round-trips are still open, and I couldn't close them from a headless review — the RPC contract is well covered by `EngineRouterSmokeTests`, but nothing has exercised the actual set → badge → reload path in a browser. The reasoning above is a static trace, not a runtime observation. Worth completing before merge, but the code reads as correct and the RPC layer beneath it is tested.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 3 Green] findings. The wiring claim checks out end-to-end, the validation contract is enforced where it actually needs to be, and the test locks in the silent-wipe and unbounded-input cases that would have been the real risks here.

dhilgaertner (approved):
## Code & Security Review

Verified against a clean checkout of `fed71cc` (diff vs `origin/feature/crow-593-web-ui-parity` — 7 files, +251/-0). Build clean; **500 tests pass** (291 CrowEngine, 96 CrowCLI, 113 CrowDaemon), including the 3/3 `EngineRouterSmokeTests` the PR claims.

### Critical Issues

None.

### Security Review

**Strengths:**
- **No XSS via the goal string.** `app.js:567` `el()` assigns through `textContent`, and both goal render paths go through it (`sessionRow` badge `app.js:1054`, header pill `app.js:1426`). The tooltip uses the `.title` *property* (`app.js:1055`), not `innerHTML`. A goal of `` renders as inert text.
- **Input is bounded at the only enforcement point that matters.** `set-goal` is the first free-text field this feature exposes to a remote web client, and the value is re-broadcast to every client in each `list-sessions` poll. `EngineRouter.swift:648` holds it to `Validation.isValidSessionName` (≤256 chars, no control chars) — the same bound the sibling `new-session`/`rename-session` handlers keep. Confirmed `Validation` is `public` in CrowCore (`Validation.swift:4`), so the engine reaches it legitimately rather than re-implementing it.
- **The silent-wipe bug is closed by construction.** The `switch (rawGoal, clear)` at `EngineRouter.swift:638` is exhaustive, and the only arm yielding a value validates it — no force-unwrap. A missing or typo'd param errors instead of clearing an existing tag. The smoke test asserts each rejection shape leaves the tag **unchanged**, not merely that it errors, which is the assertion that actually pins the bug.
- **Type-confusion fails safe.** `params["goal"]?.stringValue` / `params["clear"]?.boolValue ?? false` mean a non-string `goal` or string `"true"` degrade to the *reject* path, never to a wipe.
- **Manager exclusion is server-side** (`EngineRouter.swift:659`), not just a menu-level check in `app.js:1220` — the CLI can't bypass it.
- Auth is unchanged: `set-goal` rides the existing authenticated WS/socket gate; no new surface.

**Concerns:** none found.

### Code Quality

I checked the two architectural claims in the description, since both are load-bearing and easy to get wrong:

- **Router placement is correct.** `set-goal` lands only in `makeEngineRouter`, but `CrowDaemon.swift:214` passes `fallback: engineFallback`, so any method the daemon's curated router doesn't own falls through — the CLI socket *and* the web WebSocket both reach it. One handler genuinely wires both.
- **The `org_goal` / `alignment_weight` split is right, not arbitrary.** `EngineRouter`'s own `list-sessions` (`:168`) returns only id/name/status, so it cannot be the web's source; the daemon's curated `list-sessions` is, which is exactly where `org_goal` was added. `get-session` is *not* daemon-owned, so it falls back to the engine — where the computed fields were added. The payloads line up with the routing.
- `alignmentWeight` (`Session.swift:113`) is a pure computed property, so `get-session` adds no cost or failure mode.
- CLI↔RPC wire contract matches: `MetadataCommands.swift:56-59` sends `session_id`/`goal`/`clear` — precisely what the handler reads.
- `setSessionGoal`'s no-op guard is sound across the cases that matter: retyping the same value skips the write; emptying a tagged session sends `clear:true`; an empty entry on an untagged session no-ops rather than firing a pointless clear.

### Findings — all Green (consider)

- **`docs/cli-reference.md` is not in this diff**, though the description lists it under "Docs." That section already existed on the base branch; only `CLAUDE.md` and `Resources/CLAUDE.md.template` changed here. The *code* now matches the documented read-back contract — it's the description that overstates. Worth correcting so the next reader doesn't go looking for a doc change.
- **The ≤256/control-char bound lives only in the RPC**, not in the CLI's `validateSetGoal` (`Validation.swift:89`). Over-length or control-char goals fail server-side with an RPC error rather than a local `ValidationError`. Correct and safe — the RPC is the real chokepoint — but adding the bound to `validateSetGoal` would give `crow` users a cleaner local error and keep the "one contract" story symmetric.
- **Trim divergence:** the CLI trims `.whitespaces` while the RPC trims `.whitespacesAndNewlines`. A `"\n"`-only goal passes CLI validation and is rejected by the RPC — same outcome, different error source. Harmless today; worth aligning if these ever need to agree exactly.
- **A non-string `goal`** (e.g. `{"goal": 123}`) reports `Exactly one of 'goal' or 'clear' is required`, which is safe but slightly misleading about the actual problem.
- **Minor UI redundancy:** `renderHeader` sets `goalRow.style.cursor = 'pointer'` inline while `.meta-goal .goal-badge` already sets it in CSS; the CSS comment's "(below)" points at `.goal-badge`, which is declared *above*. (Specificity makes the rule win regardless of order, so behavior is fine.)
- **Read paths are untested:** no coverage asserting `get-session` returns `org_goal`/`ticket_priority`/`alignment_weight`, or that the daemon's `list-sessions` includes `org_goal`. These are one-line field mappings and the risky write path is well covered, so this is a nice-to-have rather than a gap I'd hold the PR for.

The four unchecked manual browser round-trips in the test plan are the remaining verification; nothing in the code reading suggests they'd fail.

### Summary Table

This fixes a real dead-call bug (a shipped CLI command with no route behind it), and the fix is tighter than the bug required: the validation contract is mirrored rather than re-invented, the rejection tests assert non-mutation, and the payload split between the two routers is deliberate instead of copy-paste. The Green items are all polish and none need to land in this PR.

PR #731, opened 2026-07-14
fix(web): report PR quick-action skips instead of a false "dispatched" (#730)

### LINKED ISSUE #730: Web PR quick actions report "dispatched" but silently no-op (Fix Checks / Conflicts / Address Review / Merge)
Clicking a PR quick-action button in the web UI (e.g. **Fix Checks**) shows `[crow] dispatched: Fix Checks` in the terminal but the action often **doesn't actually run** — the echo is a false success.

## Root cause
1. Web button → `quick-action` RPC; the client writes `[crow] dispatched: <label>` whenever the RPC does **not** throw (`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js:1534-1537`).
2. Daemon handler (`Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift`, `"quick-action"`) calls `autoRespond.dispatchManual(action:sessionID:)` and then **unconditionally** returns `["dispatched": true, ...]`.
3. `dispatchManual` (`Sources/Crow/App/AutoRespondCoordinator.swift`) returns **Void** and **silently skips** (NSLog only, nothing sent to the agent) when any of these hold:
   - no **managed** terminal — `terminals.first(where: { $0.isManaged })` is nil,
   - the terminal surface isn't initialized — `TerminalRouter.canSend(terminal)` is false,
   - the session has **no PR link**.

In those cases no prompt reaches the agent, but the RPC still reports success, so the web UI shows a "dispatched" echo for a no-op. (Trigger: sessions whose agent terminal isn't a managed/initialized surface — e.g. a terminal created without `--managed` reports `managed:false`, which `dispatchManual` rejects.)

## Fix
- Make `dispatchManual` return a result — a `Bool` (sent) or an enum with the skip reason (`noManagedTerminal` / `surfaceNotReady` / `noPRLink`).
- Have the `quick-action` daemon handler return that faithfully — e.g. `{"dispatched": false, "reason": "..."}`, or throw `DaemonRPCError.applicationError(reason)` so the client's catch fires.
- Have web `quickAction` (`app.js:1534`) surface it: only echo "dispatched" when the prompt was actually sent; otherwise show an actionable message (e.g. "Fix Checks couldn't run — no active agent terminal for this session").
- Nice-to-have: when the session has no running managed agent, offer to launch/attach one instead of failing.

## Applies to all four PR quick actions
`fixChecks` (Fix Checks), `fixConflicts` (Rebase & Fix Conflicts), `addressChanges` (Address Review), `mergePR` (Merge PR) — all route through the same handler + `dispatchManual`, so all can show a false "dispatched" (app.js:1467-1470).

## Acceptance criteria
- A PR quick action echoes "[crow] dispatched: …" **only** when the prompt was actually sent to the agent terminal.
- When it can't run (no managed terminal / surface not ready / no PR link), the web UI shows a clear reason instead of a false success.
- Consistent behavior across all four quick actions.

## Target branch
Web UI + daemon quick-action bridge — base on `feature/crow-593-web-ui-parity`, not `main`.

dgershman:
Closes #730

## Problem
Clicking a PR quick-action button in the web UI (**Fix Checks**, **Rebase & Fix Conflicts**, **Address Review**, **Merge PR**) always echoed `[crow] dispatched: <label>` — even when nothing reached the agent. `dispatchManual` returns `Void` and *silently skips* (NSLog only) when there's no managed terminal, the terminal surface isn't ready, or the session has no PR link, while the daemon handler returned `{"dispatched": true}` unconditionally and `app.js` echoed success whenever the RPC didn't throw. Users saw a false "success" for a no-op.

## Fix
Thread a typed result end-to-end so the echo is honest:

- **CrowCore** — new `QuickActionDispatchResult` (`.sent` / `.noManagedTerminal` / `.surfaceNotReady` / `.noPRLink`) with a user-facing `skipReason` (nil only when sent).
- **CrowEngine** — `dispatchManual` returns it. Marked `@discardableResult`, so the fire-and-forget auto-rebase caller (`CrowDaemon.swift onAutoRebaseConflicts`) is unchanged.
- **CrowDaemon** — the `quick-action` handler returns `{"dispatched": false, "action": …, "reason": …}` on a skip. This is a legitimate outcome, kept distinct from genuine RPC errors (app/tmux unavailable → `applicationError`; bad params → `invalidParams`).
- **Web** — `quickAction` echoes "dispatched" only on a real send; on a skip it writes `[crow] <label> couldn't run — <reason>` (yellow, actionable); genuine RPC failures stay on the existing red `catch`.

All four actions route through the same handler + `dispatchManual`, so behavior is consistent by construction.

## Tests
- Updated the daemon skip test (`quickActionReportsSkipWhenNoManagedTerminal`) to assert `dispatched:false` + a non-empty `reason` instead of the old false-success.
- Added a CrowEngine test: `dispatchManual` with no terminals returns `.noManagedTerminal`, and only `.sent` has a nil `skipReason`.
- `swift test` green for CrowEngine and CrowDaemon quick-action suites.

## Notes / out of scope
- `EngineRouter.swift` has a parallel `quick-action` handler using the app's `onQuickAction` hook (fire-and-forget, returns `dispatched:true`). Per ADR 0008 the macOS app is removed and crowd + web is the only client, so that path is legacy/unused and left untouched — possible follow-up if it's ever revived.
- The "offer to launch/attach an agent when none is running" nice-to-have is deferred; this PR scopes to honest reporting only.

https://claude.ai/code/session_016psD8Wj2mLRcZnT56vn46B

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Models/QuickAction.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/QuickAction.swift
@@ -18,3 +18,31 @@ public enum QuickAction: String, Sendable, Equatable {
     /// `isReadyToMerge` — merge the PR.
     case mergePR
 }
+
+/// Outcome of a manual `dispatchManual(action:sessionID:)` attempt. Lets callers
+/// (the daemon `quick-action` handler, the web UI) report honestly instead of
+/// assuming success: `dispatchManual` silently skips — NSLog only — when there's
+/// no managed terminal, the terminal surface isn't ready, or the session has no
+/// PR link, which previously surfaced as a false "dispatched" echo (#730).
+public enum QuickActionDispatchResult: Sendable, Equatable {
+    /// The prompt was pasted into the session's managed agent terminal.
+    case sent
+    /// The session has no managed terminal to send the prompt to.
+    case noManagedTerminal
+    /// The managed terminal exists but its surface isn't initialized yet.
+    case surfaceNotReady
+    /// The session has no `.pr` link, so there's no PR to act on.
+    case noPRLink
+
+    public var isSent: Bool { self == .sent }
+
+    /// User-facing reason a manual dispatch was skipped; `nil` when actually sent.
+    public var skipReason: String? {
+        switch self {
+        case .sent:              return nil
+        case .noManagedTerminal: return "no active agent terminal for this session"
+        case .surfaceNotReady:   return "the agent terminal isn't ready yet"
+        case .noPRLink:          return "this session has no linked PR"
+        }
+    }
+}
--- a/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
@@ -427,7 +427,14 @@ func makeCommandRouter(
             guard let actionStr = params["action"]?.stringValue, let action = QuickAction(rawValue: actionStr) else {
                 throw DaemonRPCError.invalidParams("action required (fixConflicts, addressChanges, fixChecks, mergePR)")
             }
-            await MainActor.run { autoRespond.dispatchManual(action: action, sessionID: id) }
+            // `dispatchManual` silently skips (no managed terminal / surface not
+            // ready / no PR link); report that faithfully as `dispatched:false`
+            // + a reason instead of a false success, so the web UI can surface an
+            // actionable message rather than echoing "dispatched" (#730).
+            let result = await MainActor.run { autoRespond.dispatchManual(action: action, sessionID: id) }
+            if let reason = result.skipReason {
+                return ["dispatched": .bool(false), "action": .string(action.rawValue), "reason": .string(reason)]
+            }
             return ["dispatched": .bool(true), "action": .string(action.rawValue)]
         },
 
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1530,10 +1530,19 @@ function actionBtn(label, iconName, variant, onclick) {
   return btn;
 }
 
-// Dispatch a PR quick-action (forwarded to the app's agent terminal).
+// Dispatch a PR quick-action (forwarded to the app's agent terminal). Echo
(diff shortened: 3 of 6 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- The `reason`/`label` strings are rendered via `term.write` into an xterm surface (not `innerHTML`), and the skip reasons originate from a fixed server-side enum (`QuickActionDispatchResult.skipReason`), so there's no injection surface introduced.
- No new auth, secret, or input-validation surface — the change only threads a typed outcome through an existing, already-authorized RPC path.

**Concerns:**
- None.

### Code Quality
- The fix is exactly the right shape: a typed `QuickActionDispatchResult` (`.sent` / `.noManagedTerminal` / `.surfaceNotReady` / `.noPRLink`) threaded end-to-end so the echo reflects reality instead of assuming success. This is a clean replacement for the silent-NSLog-and-return no-op.
- `@discardableResult` on `dispatchManual` correctly preserves the fire-and-forget auto-rebase caller (`CrowDaemon.swift:408`) without churn — verified it's the only other caller.
- Daemon skip vs. genuine error separation is sound: `dispatched:false` + `reason` for legitimate no-ops, `applicationError`/`invalidParams` reserved for real failures.
- Web `quickAction` uses strict `res.dispatched === false`, so success (`true`) still falls through to the "dispatched" echo; the `res.reason || '…'` fallback is a reasonable defensive default even though the daemon always sets `reason` on a skip.
- `rpc()` resolves with `msg.result || {}` (app.js:44), so `res` is always an object and the `res && …` guard is safe.
- Tests are updated honestly: the old false-success assertion (`dispatched == true`) is replaced with `dispatched == false` + non-empty `reason`, and the new CrowEngine suite pins the `.sent`-only-nil-skipReason invariant. `swift build` + both targeted test suites pass locally (CrowEngine and CrowDaemon).
- The out-of-scope `EngineRouter.swift` parallel handler is correctly identified as legacy/unused per ADR 0008 and appropriately left untouched with a documented follow-up note.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 0 Green] findings. Tight, well-tested, honestly-reported behavior fix.

---

PR #735, opened 2026-07-15
Route job agent-kind config→state sync through one testable choke point (closes #733)

### LINKED ISSUE #733: Job agent-type change (e.g. Cursor → Claude Code) not applied to newly launched jobs
## Summary

Changing the **Agent for scheduled jobs** in Settings from Cursor to Claude Code does not take effect: launching a new job afterward still runs **Cursor**. The setting appears to change in the UI (and persists to `config.json`), but the next job ignores it.

## Steps to reproduce

1. Settings → Agents → set **Agent for scheduled jobs** to `Cursor`.
2. Change it to `Claude Code`.
3. Launch / trigger a new job.
4. **Expected:** the job runs with Claude Code.
   **Actual:** the job still runs with Cursor.

## Likely root cause — `AppConfig.agentsByKind` vs `AppState.agentsByKind` are two stores that fall out of sync

There are two parallel copies of the per-action agent map:

- **`AppConfig.agentsByKind`** — persisted config. The Settings picker `perActionAgentPicker(kind: .job)` writes here and calls `save()`:
  - `Packages/CrowUI/Sources/CrowUI/SettingsView.swift` ~L839–851 — `set:` does `config.agentsByKind[key] = newValue; save()`.
- **`AppState.agentsByKind`** — in-memory runtime mirror, documented as "Mirrors `AppConfig.agentsByKind`" (`Packages/CrowCore/Sources/CrowCore/AppState.swift` ~L62–72).

The job launch resolves its agent from **AppState**, not AppConfig:
- `Sources/Crow/App/SessionService.swift` — job session is created with `agentKind: appState.agentKind(for: .job)`.
- `AppState.agentKind(for:)` returns `agentsByKind[sessionKind.rawValue] ?? defaultAgentKind` (`AppState.swift` ~L70).

So the resolver itself is correct and reads live at launch time — but if `save()` persists the change to `AppConfig` (and disk) **without propagating it into `AppState.agentsByKind`**, then `appState.agentKind(for: .job)` keeps returning the stale value (Cursor) until the app restarts / reloads config into AppState. That matches the symptom exactly.

(Note: `defaultAgentKind` may work while per-kind overrides don't, or vice versa, depending on which field gets synced — worth checking both the `agentsByKind` overrides and `defaultAgentKind` paths.)

## Suggested fix

Make the Settings save path the single source of truth so a change flows to **both** the persisted config and the runtime state before the next job launch. Options (implementer's call):
- Have `save()` (or its completion) write the updated `agentsByKind` / `defaultAgentKind` back into `AppState`, or
- Have the job launch resolve from the persisted `AppConfig` instead of the `AppState` mirror, or
- Collapse the two mirrors so there is one store.

## Acceptance criteria

- [ ] Same holds for both a per-kind override (`agentsByKind["job"]`) and clearing the override back to **Use default** (falls through to `defaultAgentKind`).
- [ ] The change is reflected in both `config.json` and the running app's resolution (`appState.agentKind(for: .job)`).
- [ ] Regression test covering "update job agent → resolve `agentKind(for: .job)` returns the new value without reloading config from disk."

## Files

- `Packages/CrowUI/Sources/CrowUI/SettingsView.swift` — `perActionAgentPicker`, `save()`
- `Packages/CrowCore/Sources/CrowCore/AppState.swift` — `agentsByKind`, `agentKind(for:)`, config→state sync
- `Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift` — `agentsByKind`, `agentKind(for:)`
- `Sources/Crow/App/SessionService.swift` — job session creation (`appState.agentKind(for: .job)`)

## Note

Prime hypothesis above is the AppConfig↔AppState sync gap; if that turns out to be already-synced, the alternative to check is the job scheduler capturing an agent kind at schedule-creation time rather than resolving live at run time.

dgershman:
Closes #733

## Finding

Changing **Agent for scheduled jobs** (e.g. Cursor → Claude Code) and launching a job — the ticket's prime hypothesis was an `AppConfig.agentsByKind` ↔ `AppState.agentsByKind` sync gap. On current `main` that propagation **already works**:

- `saveSettings` mirrors `defaultAgentKind` + `agentsByKind` into `AppState` (`AppDelegate.swift`, added June 2026).
- Jobs resolve the agent **live** at run time: `SessionService.runJob` → `agentKind: appState.agentKind(for: .job)` → the terminal launches `AgentRegistry.shared.agent(for: session.agentKind)`.
- `JobScheduler` captures **no** agent (the alternative hypothesis is false), and the Settings picker is the only writer of agent config (no CLI/RPC path).

The real gap: the config→state mirroring was **duplicated inline** in two `AppDelegate` sites in the macOS `Crow` target (effectively un-unit-testable), with **no regression guard**. That "remember to mirror every field on each save path" fragility is what could silently reintroduce the bug.

## Change

The ticket's own suggested fix — one source of truth:

- **`AppState.applyAgentConfig(_:)`** (CrowCore): a single choke point every config→state sync site funnels through, so no save path can leave the runtime mirror stale.
- Route both sync sites (startup hydrate + `saveSettings`) through it.
- **`AppStateAgentConfigSyncTests`** (swift-testing): locks the invariant with no disk reload — per-kind override update applies immediately, clearing the override falls back to `defaultAgentKind`, and a default-agent change propagates.

## Acceptance criteria

- [x] Next launched job uses the new agent, no restart — live resolve + synced state, guaranteed by the single choke point.
- [x] Works for a per-kind override **and** clearing back to Use default — covered by tests 1 & 2.
- [x] Reflected in both `config.json` (disk write unchanged) and `appState.agentKind(for: .job)`.
- [x] Regression test: update job agent → `agentKind(for: .job)` returns the new value without reloading config from disk.

## Verification

- `make build` succeeds.
- `swift test --package-path Packages/CrowCore --filter AppStateAgentConfigSync` — 3/3 pass.
- Existing agent tests (`AppConfigAgentKindTests`, `SessionAgentKindTests`) still pass.

https://claude.ai/code/session_01Q4grWhwkj5ZgDsYJJfhb7w

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -84,6 +84,16 @@ public final class AppState {
         return agentsByKind[sessionKind.rawValue] ?? defaultAgentKind
     }
 
+    /// Mirror the agent-selection config (`defaultAgentKind` + `agentsByKind`)
+    /// into runtime state so `agentKind(for:)` resolves the just-saved value
+    /// without a config reload (CROW-733). The single choke point every
+    /// config→state sync site funnels through, so no save path can leave the
+    /// mirror stale.
+    public func applyAgentConfig(_ config: AppConfig) {
+        defaultAgentKind = config.defaultAgentKind
+        agentsByKind = config.agentsByKind
+    }
+
     /// Terminal IDs whose Claude Code was launched with `--rc` — drives the
     /// per-session indicator badge. Survives toggle changes so existing sessions
     /// keep showing the badge until they're restarted.
--- a/Packages/CrowCore/Tests/CrowCoreTests/AppStateAgentConfigSyncTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AppStateAgentConfigSyncTests.swift
@@ -1,0 +1,56 @@
+import Foundation
+import Testing
+@testable import CrowCore
+
+// CROW-733: changing the per-action agent (e.g. "Agent for scheduled jobs")
+// must apply to the very next launched session without an app restart or a
+// config reload from disk. Jobs resolve their agent live via
+// `AppState.agentKind(for:)`, so the invariant under test is that
+// `AppState.applyAgentConfig(_:)` — the single choke point every config→state
+// sync site funnels through — makes the just-saved value observable immediately.
+
+@MainActor @Test func applyAgentConfigPropagatesJobOverrideUpdateWithoutReload() {
+    let state = AppState()
+    var config = AppConfig()
+
+    // First selection: Cursor for jobs.
+    config.agentsByKind["job"] = .cursor
+    state.applyAgentConfig(config)
+    #expect(state.agentKind(for: .job) == .cursor)
+
+    // Re-select in the same running app (no disk reload): Claude Code.
+    config.agentsByKind["job"] = .claudeCode
+    state.applyAgentConfig(config)
+    #expect(state.agentKind(for: .job) == .claudeCode)
+}
+
+@MainActor @Test func applyAgentConfigClearingJobOverrideFallsBackToDefault() {
+    let state = AppState()
+    var config = AppConfig()
+
+    // Override in place → resolves to the override.
+    config.defaultAgentKind = .claudeCode
+    config.agentsByKind["job"] = .cursor
+    state.applyAgentConfig(config)
+    #expect(state.agentKind(for: .job) == .cursor)
+
+    // "Use default" removes the override → resolves to defaultAgentKind.
+    config.agentsByKind.removeValue(forKey: "job")
+    state.applyAgentConfig(config)
+    #expect(state.agentKind(for: .job) == .claudeCode)
+    #expect(state.agentsByKind["job"] == nil)
+}
+
+@MainActor @Test func applyAgentConfigPropagatesDefaultAgentChange() {
+    let state = AppState()
+    var config = AppConfig()
+
+    // No per-kind override → resolution tracks defaultAgentKind.
+    config.defaultAgentKind = .cursor
+    state.applyAgentConfig(config)
+    #expect(state.agentKind(for: .job) == .cursor)
+
+    config.defaultAgentKind = .claudeCode
+    state.applyAgentConfig(config)
+    #expect(state.agentKind(for: .job) == .claudeCode)
+}
--- a/Sources/Crow/App/AppDelegate.swift
+++ b/Sources/Crow/App/AppDelegate.swift
@@ -582,8 +582,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         appState.excludeReviewRepos = config.effectiveExcludeReviewRepos
         appState.excludeTicketRepos = config.defaults.excludeTicketRepos
         appState.ignoreReviewLabels = config.defaults.ignoreReviewLabels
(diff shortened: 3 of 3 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues

None.

### Security Review

**Strengths:**
- No attack surface introduced. `applyAgentConfig(_:)` is a pure in-memory field copy between two already-trusted types — no I/O, no parsing, no shell/process invocation, no external input.
- The agent value flows to a launcher (`AgentRegistry.shared.agent(for:)`), which would matter if `AgentKind` were attacker-controlled — it isn't. The only writer is the Settings picker (`SettingsView.swift:294`, `:843-848`), and the PR's claim of "no CLI/RPC path" for agent config matches what I found.
- Disk write path is untouched, so no change to config file permissions or serialization.

**Concerns:** None found.

### Code Quality

The refactor is correct and does exactly what it says. Verified rather than taken on faith:

- **Both sync sites routed.** `AppDelegate.swift:585` (startup hydrate) and `:1315` (`saveSettings`) both call the choke point; the inline duplicate assignments are gone.
- **No third gap.** I checked the other `ConfigStore.loadConfig` call sites (`SessionService.swift:1013/1025/1422/1589/2205`, `IssueTracker.swift:394/2940/2956`) — they read config directly for unrelated purposes and never mirror agent fields into `AppState`, so there is no fourth site left holding a stale mirror.
- **The live-resolve premise holds.** `SessionService.swift:2141` resolves `agentKind: appState.agentKind(for: .job)` at job-run time, confirming the "next launched job, no restart" acceptance criterion follows from the synced mirror. `JobScheduler` captures no agent, as stated.
- **Tests are meaningful.** The three cases cover the override update, the clear-to-default fallback (asserting both the resolution *and* `agentsByKind["job"] == nil`), and the default-agent change — all without a disk reload, which is the actual invariant.

- `AppState.swift:87-91` — the doc comment states \"The single choke point every config→state sync site funnels through, so no save path can leave the mirror stale.\" That reads as an enforced invariant, but nothing enforces it: `defaultAgentKind` and `agentsByKind` remain `public var`, so a future save path can still assign them directly and reintroduce exactly the drift this PR set out to prevent. The tests lock the helper's *behavior*, not that callers *use* it — the real regression risk is unguarded. Either soften the wording to describe intent ("the site every sync path *should* funnel through"), or make it structural with `public private(set) var` so the compiler enforces it. Note the latter isn't free: `ManagerMigrationTests.swift:112-113` and `:134-141` assign both properties directly and would need a companion `applyAgentConfig` call or a test-only setter. Your call on whether that's worth it — the current shape is a genuine improvement over the duplicated inline assignments either way.

### Verification

Reproduced every claim in the PR description:

- `swift test --package-path Packages/CrowCore --filter AppStateAgentConfigSync` — 3/3 pass.
- `swift test --package-path Packages/CrowCore` — **422 tests in 12 suites pass**, including `AppConfigAgentKindTests` and the `BinaryOverrides` agent suites.
- `make build` — succeeds.

(One note for the record: the filtered test run initially failed with `sandbox-exec: sandbox_apply: Operation not permitted`. That was my local sandbox blocking SwiftPM's manifest compile, not a defect in this branch — it passed on retry.)

### Summary Table

Well-scoped diff: it correctly rejects the ticket's prime hypothesis, states the real gap (untestable duplicated mirroring), and fixes it with the minimum change plus a regression guard. The one Green note is about comment accuracy, not behavior.

---

PR #724, opened 2026-07-14
feat(web): efficiency scorecard pane (ADR 0008 parity)

dgershman:
Closes #721

Ports main's private **efficiency scorecard** to the web UI / `crowd` client. On `main`, CrowUI ships `ScorecardView`; the parity branch already has the Core models (`ScorecardModel`, grading, combined score, snapshots) but no web pane, no RPC, and `get-state` never exposed scorecard data. This wires it end-to-end in the web assets under CrowDaemon — **CrowUI is not resurrected** as a product surface.

## Approach: build server-side, render on the web
The desktop `ScorecardView` reads `ScorecardModel` directly. The web has no Swift value types, so rather than re-implement grading in JavaScript (and risk drift), `crowd` builds the **one** Core `ScorecardModel.build(...)` and ships a flat, JS-friendly `ScorecardDTO` over a new `get-scorecard` RPC. The web just renders it — so web and desktop can never disagree on a number (directly satisfies the test-plan parity requirement).

## Changes
- **CrowCore** — `ScorecardDTO` (+ `WeeklyDTO`/`GradeDTO`/`CombinedDTO`/`BaselineDTO`/`SessionRowDTO`): a lossless Codable projection of `ScorecardModel`. Enums-with-payloads (`GradeResult`, `CombinedScore.WeeklyResult`, `CostPerShipped`) collapse to structs with a discriminator + optional fields; dates are epoch **millis** (JS-native, no `dateEncodingStrategy` coupling with the `JSONValue` transport). Current week carries its derived rates inline for the baseline comparison.
- **CrowDaemon** — `get-scorecard` RPC: builds `ScorecardModel` from `appState.analyticsSnapshots` + `appState.prAttributions` (both already hydrated by SessionService/IssueTracker), encodes the DTO via the same `JSONValue(encoding:)` path as `get-state`. Carries `telemetryEnabled` + `snapshotCount` so the empty state can distinguish "telemetry off" from "on but nothing shipped".
- **Web** — `Scorecard` nav pill + full pane mirroring the desktop cards: weekly A–F grade with coachable per-metric deductions, the separate sessions-shipped surface, the v2 combined-score decomposition, the trailing-4-week "vs. your normal" baseline, context-only stats, prior weeks, and per-session drill-down. Empty state links to Settings when telemetry is off. JS formatters mirror `AnalyticsFormatting`.
[...]

PR #736, opened 2026-07-15
feat(web): per-session analytics strip (CROW-722)

### LINKED ISSUE #722: Web UI: per-session analytics strip (token/cost parity with desktop)
## Summary

Add a **per-session analytics strip** to the web session view, matching main’s CrowUI `SessionAnalyticsStrip` (live / snapshot token, cost, and related session analytics).

On the parity branch, session headers show things like review author, but **no analytics strip**. Core still has `SessionAnalytics` / hook-state analytics and end-of-session snapshots from main.

## Why

Without this, web users can’t see session efficiency signals that desktop already shows next to the session.

## Already on the branch (reuse)

- Hook-state / analytics fields in engine + store
- `SessionAnalyticsSnapshot` persistence for completed sessions
- Telemetry toggle in web Settings

## Work

- [ ] Decide live vs snapshot source for open sessions (hook aggregate vs telemetry.db) and expose via list-sessions / get-state / dedicated RPC
- [ ] Web strip in session header/detail: cost, tokens (input/output/cache as available), optional active/wall-clock duration
- [ ] Graceful empty state when telemetry is off or analytics are empty
- [ ] Keep layout consistent with existing web header density (don’t recreate a desktop-only card forest)

## Refs

- Desktop: `SessionAnalyticsStrip` on main CrowUI
- ADR 0008 (related data; this ticket is the per-session surface, not the weekly scorecard)
- Parent parity: #593 / #581

## Test plan

- [ ] Active Claude session with telemetry on shows updating/non-empty strip
- [ ] Telemetry off → muted/empty, no errors
- [ ] Completed session can show last snapshot values if live aggregate is gone

dgershman:
Closes #722

Adds a **per-session analytics strip** to the web session header, matching main's desktop `SessionAnalyticsStrip` — the parity gap called out in the ticket (web headers show review author etc. but no analytics).

## What
- **New `SessionAnalyticsDTO`** (CrowCore): a flat, JS-friendly projection of a session's analytics, built from either the **live** in-memory hook aggregate (open sessions) or the **snapshot** (terminal sessions), with a `source` discriminator. Sibling to `ScorecardDTO`.
- **`list-sessions-live`** now attaches a per-session `analytics` object, resolved server-side (live → snapshot fallback, mirroring `writeAnalyticsSnapshot`'s preference). No new RPC: `refreshLive()` already re-renders the selected header each poll, so the strip updates on the same cadence as the PR/RC badges.
- **Web strip** (`app.js`/`app.css`): `renderSessionAnalyticsStrip` shows Cost · Tokens · Tools · Active, plus Duration / Lines / Errors chips when present, reusing the scorecard's `fmt*` + `statChipEl` helpers scoped to a compact, unboxed `.analytics-strip` (keeps header density; no card forest).

## Source decision (live vs snapshot)
Per session, in priority order: live hook aggregate (non-empty) → durable end-of-session snapshot → nothing.

## Empty state
**Chips-only** (per review preference): the strip renders only when analytics data exists — absent when telemetry is off or nothing was recorded, and never for the Manager. Absence *is* the empty state.

## Note — surface parity over dormant infra
The analytics ingestion pipeline is still dormant on the parity branch (`TelemetryService` is never instantiated; `hookState.analytics` is never written at runtime), so today the strip is absent by default. This is the same posture as the #724 scorecard: it's built to consume the data and lights up **unchanged** once telemetry ingestion is wired. Wiring that ingestion is intentionally **out of scope** here.

## Base
Stacked on **`feature/crow-593-web-ui-parity`** (sibling to #725), not main.

## Test plan
- [x] `make build` green (daemon + desktop)
- [x] `swift test` — new `SessionAnalyticsDTOTests` (5 cases: live/snapshot inits, totalTokens, wall-clock passthrough, flat JSON encode) pass
- [x] `node --check app.js`
- [x] Telemetry off (current default): strip absent, header density unchanged, no console errors
- [ ] With a snapshot present: `list-sessions-live` returns the `analytics` object and the strip renders chips (covers "completed session shows last snapshot")

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Models/SessionAnalyticsDTO.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/SessionAnalyticsDTO.swift
@@ -1,0 +1,66 @@
+import Foundation
+
+/// A flat, `Codable`, JS-friendly projection of a single session's analytics for
+/// the web per-session strip (CROW-722; ADR 0008 web parity, sibling to the
+/// `ScorecardDTO` surface). The desktop `SessionAnalyticsStrip` reads a
+/// `SessionAnalytics` value directly; the web has no Swift value types, so
+/// `crowd` resolves the best available source per session — the live in-memory
+/// hook aggregate for an open session, else the durable end-of-session snapshot —
+/// and ships this DTO inside `list-sessions-live`. The web renders it verbatim
+/// with the same `fmt*` helpers the scorecard chips use, so web and desktop can't
+/// drift on the numbers or their formatting.
+public struct SessionAnalyticsDTO: Codable, Sendable, Equatable {
+    /// Which source produced these numbers: `"live"` = the in-memory hook
+    /// aggregate for an open session; `"snapshot"` = the durable end-of-session
+    /// record. Lets the web distinguish live-updating vs. final without a second
+    /// lookup (currently informational — the strip renders both identically).
+    public let source: String
+    public let totalCost: Double
+    public let inputTokens: Int
+    public let outputTokens: Int
+    public let cacheReadTokens: Int
+    public let cacheCreationTokens: Int
+    public let totalTokens: Int
+    public let activeTimeSeconds: Double
+    public let toolCallCount: Int
+    public let linesAdded: Int
+    public let linesRemoved: Int
+    public let apiErrorCount: Int
+    /// Wall-clock span (first `SessionStart` → last `SessionEnd`). Display-only
+    /// context beside "Active", never a grading input. `nil` hides the chip
+    /// (open-ended session, or an agent that never sends `SessionEnd`).
+    public let wallClockDurationSeconds: Double?
+
+    private init(source: String, analytics: SessionAnalytics, wallClockDurationSeconds: Double?) {
+        self.source = source
+        self.totalCost = analytics.totalCost
+        self.inputTokens = analytics.inputTokens
+        self.outputTokens = analytics.outputTokens
+        self.cacheReadTokens = analytics.cacheReadTokens
+        self.cacheCreationTokens = analytics.cacheCreationTokens
+        self.totalTokens = analytics.totalTokens
+        self.activeTimeSeconds = analytics.activeTimeSeconds
+        self.toolCallCount = analytics.toolCallCount
+        self.linesAdded = analytics.linesAdded
+        self.linesRemoved = analytics.linesRemoved
+        self.apiErrorCount = analytics.apiErrorCount
+        self.wallClockDurationSeconds = wallClockDurationSeconds
+    }
+
+    /// Live in-memory hook aggregate for an open session. `wallClockDuration`
+    /// comes from the session's lifecycle stamps (`nil` while the session hasn't
+    /// ended, or for agents that never send `SessionEnd`).
+    public init(live analytics: SessionAnalytics, wallClockDuration: TimeInterval? = nil) {
+        self.init(source: "live", analytics: analytics, wallClockDurationSeconds: wallClockDuration)
+    }
+
+    /// Durable end-of-session snapshot (terminal sessions). Snapshots are only
+    /// written for non-empty aggregates, so a snapshot is always meaningful.
+    public init(snapshot: SessionAnalyticsSnapshot) {
+        self.init(
+            source: "snapshot",
+            analytics: snapshot.analytics,
+            wallClockDurationSeconds: snapshot.wallClockDurationSeconds
+        )
+    }
+}
--- a/Packages/CrowCore/Tests/CrowCoreTests/SessionAnalyticsDTOTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/SessionAnalyticsDTOTests.swift
@@ -1,0 +1,93 @@
+import Foundation
+import Testing
+@testable import CrowCore
+
+// CROW-722 (ADR 0008 web parity): `SessionAnalyticsDTO` is a lossless, JS-friendly
+// projection of a session's analytics — either the live in-memory hook aggregate
+// or the durable end-of-session snapshot. These pin that both inits carry exactly
+// the numbers the desktop `SessionAnalyticsStrip` reads, so the web strip can't
+// drift from Core.
+
+@Test func liveInitCarriesEveryFieldAndComputesTotalTokens() {
+    let analytics = SessionAnalytics(
+        totalCost: 1.23,
+        inputTokens: 100,
+        outputTokens: 200,
+        cacheReadTokens: 300,
+        cacheCreationTokens: 400,
+        activeTimeSeconds: 3600,
+        linesAdded: 10,
+        linesRemoved: 4,
+        commitCount: 2,
+        promptCount: 5,
+        toolCallCount: 42,
+        apiRequestCount: 7,
+        apiErrorCount: 3
+    )
+    let dto = SessionAnalyticsDTO(live: analytics, wallClockDuration: 5400)
+
+    #expect(dto.source == "live")
+    #expect(dto.totalCost == 1.23)
+    #expect(dto.inputTokens == 100)
+    #expect(dto.outputTokens == 200)
+    #expect(dto.cacheReadTokens == 300)
+    #expect(dto.cacheCreationTokens == 400)
+    #expect(dto.totalTokens == 1000) // 100 + 200 + 300 + 400
+    #expect(dto.activeTimeSeconds == 3600)
+    #expect(dto.toolCallCount == 42)
+    #expect(dto.linesAdded == 10)
+    #expect(dto.linesRemoved == 4)
+    #expect(dto.apiErrorCount == 3)
+    #expect(dto.wallClockDurationSeconds == 5400)
+}
+
+@Test func liveInitWithoutWallClockOmitsDuration() {
+    let dto = SessionAnalyticsDTO(live: SessionAnalytics(totalCost: 0.5))
+    #expect(dto.source == "live")
+    #expect(dto.wallClockDurationSeconds == nil)
+}
+
+@Test func snapshotInitCarriesAnalyticsAndWallClock() {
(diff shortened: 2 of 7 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Build green (`swift build`, only a pre-existing `RPCWebSocketHandler` warning), `swift test --package-path Packages/CrowCore` → **443 tests / 15 suites pass** including the 5 new `SessionAnalyticsDTOTests`, `node --check app.js` clean. The DTO is a clean sibling to `ScorecardDTO`, the server-side source resolution is the right call (no JS re-implementation to drift), and the chips-only empty state is a good fit for header density.

One functional gap in the snapshot half of the feature keeps this from an approve.

### Critical Issues

None — nothing here can crash or corrupt state.

### Code Quality

**🟡 The snapshot fallback is dead in `crowd` — `reseed` never seeds the mirror it reads**

`RPCHandlers.swift:690` falls back to `appState.analyticsSnapshots[id.uuidString]`, but nothing ever populates that dictionary in the daemon from disk:

- The only reader→writer path is `SessionService.hydrateState()` (`SessionService.swift:93`), which the daemon never calls — `CrowDaemon.swift:564` says so explicitly ("AppKit-bound and out of scope for the daemon"). Repo-wide, `hydrateState()` is called *only* from tests.
- `CrowDaemon.reseed` (`CrowDaemon.swift:567-587`) restores `sessions`, `worktrees`, `terminals`, `links`, and `hookStates` — but **not** `analyticsSnapshots`.
- `PersistedHookState` (`AppState.swift:750-763`) carries only `activityState` / `pendingNotification` / `lastTopLevelStopAt`, so the restored hook state has an empty `analytics` too — the live branch can't cover for it.

Net: `appState.analyticsSnapshots` is `[:]` on every fresh `crowd` process, and is only ever written by `SessionService.writeAnalyticsSnapshot` during that process's own lifetime. **A session that completed before the daemon started never renders the strip, even with a valid snapshot sitting in `store.json`.** That's precisely the unchecked box in your own test plan ("With a snapshot present … covers 'completed session shows last snapshot'") — it doesn't pass today.

This also makes the PR body's claim inaccurate: wiring telemetry ingestion lights up the *live* half unchanged, but the *snapshot* half stays dark until `reseed` is fixed. The fix looks like one line alongside the existing `hookStates` restore:

```swift
// Mirror persisted analytics snapshots so the per-session strip (CROW-722)
// and the scorecard (#724) survive a daemon restart.
appState.analyticsSnapshots = data.analyticsSnapshots ?? [:]
```

Worth noting `get-scorecard` (`RPCHandlers.swift:745`) reads the same empty mirror, so this one line likely fixes a latent #724 bug too. If you'd rather keep that out of scope, please drop the `snapshot:` init and the `"snapshot"` discriminator rather than shipping a documented path that can't fire.

**🟡 Zero coverage on the resolution logic — the one place the bug actually is**

All 5 new tests exercise `SessionAnalyticsDTO`'s initializers, which are straight field copies. Untested: the live→snapshot preference, the `!live.isEmpty` gate, and the Manager exclusion — the branching that carries the actual risk, and where the defect above lives. `BoardForwarderTests` already builds an offline `makeCommandRouter` off a plain `AppState`/`JSONStore`; a test that seeds a snapshot into the store, reseeds, and asserts `list-sessions-live` returns `analytics.source == "snapshot"` would have caught this and would pin the contract going forward.

**🟢 `source` ships but is never read** — the web renders live and snapshot identically, so `source` is currently unused payload. Fine as forward-looking, and the doc comment is honest about it; just flagging it as the kind of field that quietly rots.

**🟢 Strip pops in on the second paint** — `renderHeader` can run before the first `refreshLive` populates `liveById`, so `liveFor(s.id).analytics` is undefined on first render and the strip appears one poll later, nudging the links row down. Cosmetic, and self-correcting once data exists.

### Security Review

**Strengths:**
- **No injection surface.** Every chip goes through `el()` → `textContent`, never `innerHTML`, and all values are numbers routed through `fmtCost`/`fmtCount`/`fmtTime`. Nothing here is attacker-controlled anyway.
- **Manager exclusion is defense-in-depth** — enforced server-side (`isManagerSession`) *and* client-side (`s.kind === 'manager'`), so the data never leaves the daemon rather than merely being hidden in the DOM.
- **No new exposure.** Cost/token data rides the existing authenticated `list-sessions-live` channel behind `WebAuth`, same posture as `get-scorecard`. No new RPC, no new route, no secrets, no auth/crypto changes.
- `try? JSONValue(encoding:)` fails closed — an encode error omits the key and the strip is simply absent, matching the empty state.

**Concerns:** none.

### Summary Table

The shape of this is right and the live path is sound; it's the snapshot path that needs either the one-line `reseed` fix (plus a router-level test) or removal.

---

dgershman:
Thanks for the thorough review — both 🟡 findings are fixed in 6ce0d6a.

**🟡 Dead snapshot fallback** — you're exactly right: `CrowDaemon.reseed` restored sessions/worktrees/terminals/links/hookStates but never `appState.analyticsSnapshots`, so on a fresh `crowd` it was `[:]` and the snapshot half could never fire. Added the one-liner alongside the `hookStates` restore:

```swift
appState.analyticsSnapshots = data.analyticsSnapshots ?? [:]
```

As you noted, `get-scorecard` reads the same mirror, so this also lights up the #724 scorecard for sessions that completed before the daemon started — a latent bug it fixes too.

**🟡 Zero coverage on the resolution logic** — added a router-level suite (`SessionAnalyticsStripLiveTests`, mirroring `LocalLiveReadTests`): it seeds a snapshot into the store, runs the real `reseed`, and asserts `list-sessions-live` returns `analytics.source == "snapshot"` with the right fields — so it fails if the reseed line regresses — plus the empty-state omission and the Manager exclusion. `reseed` dropped `private` so the test drives the actual hydrate path. Full CrowDaemon suite: **113 tests / 31 suites green.**

**On the PR-body accuracy** — good catch. With reseed fixed, the **snapshot** half now works across restarts (completed sessions with a persisted snapshot render immediately). Only the **live** (open-session) half stays dormant until telemetry ingestion is wired — that remains the out-of-scope item.

**🟢 `source` unused** — now read by the new test's `source == "snapshot"` assertion, and kept as the forward-looking live/snapshot discriminator. **🟢 Second-paint pop-in** — acknowledged; cosmetic and self-correcting, leaving as-is.

dhilgaertner (approved):
## Code & Security Review

Verified the diff end-to-end: built the packages, ran both new test suites plus the full CrowCore (443) and CrowDaemon (113) suites, and checked each parity claim in the PR body against `origin/main` rather than taking it on trust.

### Critical Issues
None.

### Verification performed
- `swift build` — green.
- `swift test` in `Packages/CrowCore` — 443 tests pass, including the 5 new `SessionAnalyticsDTOTests`.
- `swift test` in `Packages/CrowDaemon` — 113 tests pass, including the 3 new `SessionAnalyticsStripLiveTests`. No regressions from opening up `reseed`.
- `node --check app.js` — OK.

**Parity claims checked against `origin/main` (all hold):**
- Chip set matches `SessionAnalyticsStrip.swift` exactly — Cost/Tokens/Tools/Active always, Duration when non-nil, Lines when either counter is non-zero, Errors in red when `> 0`.
- `fmtCost`/`fmtCount`/`fmtTime` match `AnalyticsFormatting.cost/count/time` on every threshold and format, so the DTO's "can't drift on the numbers or their formatting" claim is accurate.
- The web adds a snapshot fallback and Manager exclusion that the desktop's `SessionDetailView:234` lacks. Both are deliberate and documented, and the Manager exclusion is consistent with `writeAnalyticsSnapshot`'s own `isManagerSession` skip.

**The `reseed` fix is correct and well-targeted.** `PersistedHookState` carries no analytics, so `hookState.analytics` is always empty in `crowd` and the snapshot branch was genuinely the only live path — dead without this. `CrowDaemon.swift:593` mirrors `SessionService.swift:93` (`hydrateState`) exactly, which is the right precedent. I confirmed no daemon-side writer of `appState.analyticsSnapshots` exists that the store-poll reseed could clobber.

### Security Review
**Strengths:**
- No new RPC or auth surface: `list-sessions-live` already returns per-session PR links and remote-control state; this adds non-sensitive counters (cost, tokens, tool calls, line counts). No secrets, credentials, or PII.
- No XSS: the strip builds every node through `el()`/`statChipEl()`, which use `textContent`. No `innerHTML` on any interpolated value.
- `existingHookState(for:)` reads without allocating hook state per session — no side effects in a read path.

**Concerns:** None found.

### Code Quality
- Correct null handling: `wallClockDurationSeconds != null` catches both the omitted key (Swift's synthesized `encodeIfPresent` drops nil optionals) and an explicit null.
- No DOM accumulation: `renderHeader` does `root.innerHTML = ''` before re-rendering, so the per-poll re-render can't stack strips.
- `liveFor` returns `{}` for a miss, so `.analytics` degrades to the empty state rather than throwing.
- Tests pin the branches that matter — the empty-state omission, the Manager exclusion, and the reseed-after-restart path the DTO's own tests can't reach. The `num()` int-or-double helper correctly sidesteps `JSONValue`'s Int-first decoding.

**Consider (non-blocking, no action needed to merge):**
- `RPCHandlers.swift:695` — `try? JSONValue(encoding: dto)` silently drops the key on an encode failure, making the strip vanish with no diagnostic. Unreachable for a DTO this flat, but a `log` in the `catch` would age better.
- `RPCHandlers.swift:679` — the DTO is resolved and encoded for every session on every poll, though only the selected session's strip renders. Negligible against the existing per-session work in this loop; noting it only in case the session count grows.
- The desktop places the strip as Row 4 with a `Divider`; the web places it above the links row without one. Cosmetic, and the PR states the placement intent.

### Summary Table

The scoped-to-consumption posture (strip absent until telemetry ingestion is wired) matches the #724 scorecard precedent and is explicitly called out as out of scope. The one unchecked test-plan box — rendering with a snapshot present — is now covered by `reseedMirrorsSnapshotsSoStripRendersAfterRestart`, which asserts the `analytics` object and its values come back from `list-sessions-live`.

Issue #739, opened 2026-07-15
Reland on feature/crow-593-web-ui-parity: job agent-kind config→state sync (redo #733)


dgershman:
## Context — reland on the correct branch

The original work for #733 ("job agent-type change not applied to newly launched jobs" → routed config→state sync through one testable choke point) was implemented in **PR #735**, but that PR was **merged into `main` by mistake**. It should have targeted the web-UI-parity integration branch.

- **PR #735** (merged to `main`, squash commit `2581797`) is being **reverted on `main`** — see the revert PR linked below.
- This ticket re-lands the exact same change, this time on the correct base branch.

## ⚠️ Base branch (merge target)

**Open the PR against `feature/crow-593-web-ui-parity` — NOT `main`.**

Branch off `feature/crow-593-web-ui-parity` and set that as the PR base.

## What to do

Re-apply the #735 change onto `feature/crow-593-web-ui-parity`. The original implementation is good, so the fastest path is to cherry-pick the original squash commit onto the feature branch:

```bash
git switch -c feature/crow-<this-issue>-reland-job-agent-sync --no-track origin/feature/crow-593-web-ui-parity
git cherry-pick 2581797c42d0549aa45a6fc7bebf5f0e6f933fbd
# resolve any drift vs the feature branch, then build + test
```

If cherry-pick conflicts with the feature branch, re-implement per the original spec in #733 instead.

## Original acceptance criteria (from #733)

- [ ] After changing **Agent for scheduled jobs** in Settings, the **very next** launched job uses the newly selected agent — no app restart required.
- [ ] Same holds for a per-kind override (`agentsByKind["job"]`) and clearing back to **Use default** (falls through to `defaultAgentKind`).
- [ ] Reflected in both `config.json` and the running app's resolution (`appState.agentKind(for: .job)`).
- [ ] Regression test: update job agent → `agentKind(for: .job)` returns the new value without reloading config from disk (`AppState.applyAgentConfig(_:)` choke point + `AppStateAgentConfigSyncTests`).
- [ ] Build passes; `make build` / CrowCore tests green.

## References

- Original issue: #733
- Original (mis-targeted) PR: #735
- Revert of #735 on `main`: see linked revert PR

PR #741, opened 2026-07-15
Revert #735 from main (merged to wrong base; relands in #739)

### LINKED ISSUE #733: Job agent-type change (e.g. Cursor → Claude Code) not applied to newly launched jobs

dgershman:
## Why

**PR #735** (closes #733) was merged into **`main`** by mistake — it should have targeted the web-UI-parity integration branch `feature/crow-593-web-ui-parity`. This PR reverts it from `main`.

This is a clean revert of the squash commit `2581797`; no code is lost. The same change is being re-landed on the correct base branch in **#739**.

## What

- Reverts `2581797` ("Route job agent-kind config→state sync through one testable choke point (closes #733) (#735)").
- Restores `AppState.applyAgentConfig` choke point + `AppStateAgentConfigSyncTests` to their pre-#735 state.

## Follow-up

- Reland ticket (correct branch): #739
- Original issue: #733 · Mis-targeted PR: #735


dhilgaertner (approved):
## Code & Security Review

Clean revert. Verified mechanically rather than by reading the diff alone — every claim in the description holds up.

### Verification

**The revert is exact.** `git diff 90a3d54 HEAD` is **empty** — the tree is byte-identical to the commit immediately preceding #735. This is the strongest available evidence that the revert restores the pre-#735 state precisely, with no incidental drift smuggled in. The branch is main + the single revert commit `f6caff6`, with no unrelated changes.

**No dangling references.** `rg applyAgentConfig` across the repo returns nothing, so the removed choke point has no orphaned callers. Both sync sites (`Sources/Crow/App/AppDelegate.swift:585-586` and `:1316-1317`) are restored to the inline `defaultAgentKind` / `agentsByKind` assignment pair, which is semantically identical to the `applyAgentConfig` body being removed — no behavior change beyond the intended revert.

**Build and tests green.** `swift build` succeeds and `swift test` passes **378 tests in 45 suites**. Deleting `AppStateAgentConfigSyncTests.swift` leaves no broken references.

> Note for anyone reproducing locally: a fresh worktree fails to build with `cannot find 'BuildInfo' in scope`. That is **not** this PR — `BuildInfo.swift` is generated by `scripts/generate-build-info.sh` and is untracked. Run that script first.

**Follow-up references are real.** #735 is confirmed `MERGED` into `main` (squash `2581797`), corroborating the mis-targeting. #739 is an **open issue** — the reland ticket, correctly describing `feature/crow-593-web-ui-parity` as the base — and that branch exists on the remote. The "What" section is accurate.

### Security Review

**Strengths:**
- Change is removal-only within app-state plumbing: no authentication, authorization, input-validation, injection, cryptographic, or secrets surface is touched.
- No configuration or dependency changes; no new external inputs.
- Restoring the inline assignments does not widen any trust boundary — the values still flow from the same already-parsed `AppConfig`.

**Concerns:**
- None identified.

### Code Quality

- No logic errors, resource leaks, or race conditions introduced. The two restored assignment sites run on the same actor-isolated path as before.
- Test-coverage reduction (`AppStateAgentConfigSyncTests`, 56 lines) is inherent to restoring the prior state, not a defect in this PR. Coverage returns with the reland.

**Green (consider) — non-blocking, no action needed here:**
- This intentionally reinstates the CROW-733 bug on `main` (a job agent-kind change won't apply to the next launched job without a reload). Because the reland targets the `feature/crow-593-web-ui-parity` integration branch rather than `main`, `main` carries that bug until the parity branch merges. That is the deliberate, stated trade-off and #739 tracks it — flagging only so the exposure window is a conscious decision.
- Issue #733 is currently `CLOSED` while its fix is being removed from `main`. #739 covers the work, so nothing is lost; reopening #733 is optional and purely a bookkeeping preference.

### Summary Table

PR #742, opened 2026-07-15
Reland: route job agent-kind config→state sync through one choke point (CROW-739)

### LINKED ISSUE #739: Reland on feature/crow-593-web-ui-parity: job agent-kind config→state sync (redo #733)

dgershman:
Re-lands #735 (CROW-733) onto `feature/crow-593-web-ui-parity` after that PR was merged to `main` by mistake and reverted there. **Base is the parity branch, not `main`.**

## What
- **`AppState.applyAgentConfig(_:)`** (CrowCore) — the single choke point every config→state agent sync site funnels through, so no save path can leave the runtime mirror (`agentKind(for:)`) stale.
- Route the mirror site through it. The original #735 target was the macOS app's two `AppDelegate` sites; on this parity branch those are **deleted under ADR 0008** and collapsed into one static choke point, **`CrowDaemon.applyConfigToAppState`** — the sole writer of these fields, called at boot and on every board tick, so runtime Settings edits take effect within one poll. That is the site rerouted here.
- **`AppStateAgentConfigSyncTests`** — locks the invariant with no disk reload.

## Drift vs. #735
`Sources/Crow/App/AppDelegate.swift` no longer exists on this branch, so the cherry-pick's AppDelegate hunk was dropped and its intent relanded at the daemon choke point instead. CrowCore additions applied unchanged.

## Verification
- `make build` succeeds.
- `swift test --package-path Packages/CrowCore --filter 'AppStateAgentConfigSync|AppConfigAgentKind|SessionAgentKind'` — 17/17 pass (incl. the 3 new `applyAgentConfig*` regression tests).

## Acceptance criteria (CROW-733)
- [x] Next launched job uses the new agent, no restart — live `agentKind(for: .job)` resolve + daemon re-runs `applyConfigToAppState` → `applyAgentConfig` within one poll.
- [x] Per-kind override update **and** clearing back to Use default — tests 1 & 2.
- [x] Reflected in both `config.json` (disk write unchanged) and `appState.agentKind(for: .job)`.
- [x] Regression guard with no disk reload — `AppStateAgentConfigSyncTests`.
- [x] Build passes; CrowCore tests green.

Closes #739

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/AppState.swift
+++ b/Packages/CrowCore/Sources/CrowCore/AppState.swift
@@ -90,6 +90,16 @@ public final class AppState {
         return agentsByKind[sessionKind.rawValue] ?? defaultAgentKind
     }
 
+    /// Mirror the agent-selection config (`defaultAgentKind` + `agentsByKind`)
+    /// into runtime state so `agentKind(for:)` resolves the just-saved value
+    /// without a config reload (CROW-733). The single choke point every
+    /// config→state sync site funnels through, so no save path can leave the
+    /// mirror stale.
+    public func applyAgentConfig(_ config: AppConfig) {
+        defaultAgentKind = config.defaultAgentKind
+        agentsByKind = config.agentsByKind
+    }
+
     /// Terminal IDs whose Claude Code was launched with `--rc` — drives the
     /// per-session indicator badge. Survives toggle changes so existing sessions
     /// keep showing the badge until they're restarted.
--- a/Packages/CrowCore/Tests/CrowCoreTests/AppStateAgentConfigSyncTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AppStateAgentConfigSyncTests.swift
@@ -1,0 +1,56 @@
+import Foundation
+import Testing
+@testable import CrowCore
+
+// CROW-733: changing the per-action agent (e.g. "Agent for scheduled jobs")
+// must apply to the very next launched session without an app restart or a
+// config reload from disk. Jobs resolve their agent live via
+// `AppState.agentKind(for:)`, so the invariant under test is that
+// `AppState.applyAgentConfig(_:)` — the single choke point every config→state
+// sync site funnels through — makes the just-saved value observable immediately.
+
+@MainActor @Test func applyAgentConfigPropagatesJobOverrideUpdateWithoutReload() {
+    let state = AppState()
+    var config = AppConfig()
+
+    // First selection: Cursor for jobs.
+    config.agentsByKind["job"] = .cursor
+    state.applyAgentConfig(config)
+    #expect(state.agentKind(for: .job) == .cursor)
+
+    // Re-select in the same running app (no disk reload): Claude Code.
+    config.agentsByKind["job"] = .claudeCode
+    state.applyAgentConfig(config)
+    #expect(state.agentKind(for: .job) == .claudeCode)
+}
+
+@MainActor @Test func applyAgentConfigClearingJobOverrideFallsBackToDefault() {
+    let state = AppState()
+    var config = AppConfig()
+
+    // Override in place → resolves to the override.
+    config.defaultAgentKind = .claudeCode
+    config.agentsByKind["job"] = .cursor
+    state.applyAgentConfig(config)
+    #expect(state.agentKind(for: .job) == .cursor)
+
+    // "Use default" removes the override → resolves to defaultAgentKind.
+    config.agentsByKind.removeValue(forKey: "job")
+    state.applyAgentConfig(config)
+    #expect(state.agentKind(for: .job) == .claudeCode)
+    #expect(state.agentsByKind["job"] == nil)
+}
+
+@MainActor @Test func applyAgentConfigPropagatesDefaultAgentChange() {
+    let state = AppState()
+    var config = AppConfig()
+
+    // No per-kind override → resolution tracks defaultAgentKind.
+    config.defaultAgentKind = .cursor
+    state.applyAgentConfig(config)
+    #expect(state.agentKind(for: .job) == .cursor)
+
+    config.defaultAgentKind = .claudeCode
+    state.applyAgentConfig(config)
+    #expect(state.agentKind(for: .job) == .claudeCode)
+}
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
@@ -617,8 +617,10 @@ public enum CrowDaemon {
         // resolves the built-in default (.claudeCode) via appState.agentKind(for:),
         // so the Settings manager/coder agent pickers are ignored on
         // restart/respawn even though they persist to config (CROW-433 / CROW-581).
-        appState.defaultAgentKind = config.defaultAgentKind
-        appState.agentsByKind = config.agentsByKind
+        // Route through the single choke point so no field is silently missed on
+        // any sync path and the next launched job resolves the just-saved agent
+        // without a config reload (CROW-733).
+        appState.applyAgentConfig(config)
     }
 
     /// Poll `store.json`'s mtime and reload when the desktop app writes it, so
--- a/Packages/CrowDaemon/Tests/CrowDaemonTests/DaemonSecurityTests.swift
+++ b/Packages/CrowDaemon/Tests/CrowDaemonTests/DaemonSecurityTests.swift
@@ -247,7 +247,9 @@ import CrowPersistence
             remoteControlEnabled: true,          // AppState default: false
             managerAutoPermissionMode: false,    // AppState default: true
             jobsAutoPermissionMode: false,       // AppState default: true
-            coderViewAutoPermissionMode: true)   // AppState default: false
+            coderViewAutoPermissionMode: true,   // AppState default: false
+            defaultAgentKind: .cursor,           // AppState default: .claudeCode
+            agentsByKind: ["job": .codex])       // AppState default: [:]
         try ConfigStore.saveConfig(cfg, devRoot: devRoot)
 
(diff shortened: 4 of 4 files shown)


dhilgaertner (changes requested):
## Code & Security Review

The refactor itself is correct and the behavioral claim holds. I verified the mechanism end-to-end: `applyConfigToAppState` reloads config from disk on every call (`CrowDaemon.swift:607`) and runs at boot (`:57`) and each board tick (`:487`), so a runtime Settings edit does reach `agentKind(for: .job)` within one poll with no restart. A repo-wide grep confirms `applyAgentConfig` is now the only production writer of `defaultAgentKind` / `agentsByKind` — the "single choke point" claim is accurate. `swift test --filter 'AppStateAgentConfigSync|AppConfigAgentKind|SessionAgentKind'` → 17/17 pass, matching the PR body. Swift build is clean.

One issue holds this back.

### Yellow — the regression guard doesn't guard the regression

`AppStateAgentConfigSyncTests` only ever calls `state.applyAgentConfig(config)` directly, so all three tests exercise a two-line setter that cannot realistically break. The bug CROW-733 actually describes — a launched job resolving a stale agent — lives at the *daemon* call site, and nothing asserts that site stays wired up.

I confirmed this by mutation testing rather than inference. Deleting `appState.applyAgentConfig(config)` from `CrowDaemon.swift:623` reintroduces CROW-733 in full, and the suite stays green:

- `ConfigAppStateSyncTests` — 2/2 pass
- `AppStateAgentConfigSyncTests` — 3/3 pass (they never touch the daemon)

So the guard passes in exactly the world it exists to prevent. That matters more than usual here: this is a reland of a change that already got lost once, and the guard is the stated defense against losing it again.

The fix is small, because the right test already exists. `ConfigAppStateSyncTests.appliesExcludeReposAndPermissionGatesFromConfig` (`DaemonSecurityTests.swift:237`) already writes a config to disk, calls `applyConfigToAppState`, and asserts every other field lands. Add the agent fields to the fixture and assert them after the call — pick values differing from the `AppState` defaults, per that test's own convention, so a pass proves the copy happened:

```swift
let cfg = AppConfig(
    // …existing fields…
    defaultAgentKind: .cursor,              // AppState default: .claudeCode
    agentsByKind: ["job": .codex])
// …
#expect(appState.agentKind(for: .job) == .codex)
#expect(appState.defaultAgentKind == .cursor)
```

That kills the mutation above and covers the disk→config→state path the ticket is really about. Worth noting that suite's docstring already claims "these lock in that **every** field lands" (`DaemonSecurityTests.swift:229`) — that statement is currently false, and this change makes it true again.

### Security Review

No new surface. The change relocates two assignments into a method; the data, its source (`ConfigStore.loadConfig`), and the trust boundary are unchanged. No secrets, no injection vector, no auth/authz impact. The `guard let config … else { return }` early-return still leaves `AppState` at its defaults rather than clobbering it to zero-values when `config.json` is absent — covered by `missingConfigLeavesDefaultsUntouched`.

### Code Quality

**Green — the choke point is convention, not enforcement.** The docstring asserts an invariant ("every config→state sync site funnels through") that the type doesn't enforce: `defaultAgentKind` and `agentsByKind` remain `public var` (`AppState.swift:79,84`), so a future caller can assign them directly and silently reopen this class of bug. `private(set)` would enforce it, but it would break existing direct assignments in `ManagerMigrationTests.swift:112-113,134-141`, so this is a genuine judgment call rather than an obvious win — flagging it as a consideration, not a request.

Comments and rationale are unusually good; the CROW-733 references and the "why" on the daemon call site will age well.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 1 Green] findings.

The production change is sound and I'd approve it as-is on the code; the request is scoped to making the guard cover the daemon site, which is a few lines in a test that already does the setup work.

dgershman:
Addressed the Yellow finding in c545368.

**The guard now guards the daemon site.** Rather than add a redundant setter test, I extended the existing disk→config→state test `ConfigAppStateSyncTests.appliesExcludeReposAndPermissionGatesFromConfig` (the one you pointed at) — it already writes config to disk and calls `applyConfigToAppState`. Added the agent fields to the fixture (`defaultAgentKind: .cursor`, `agentsByKind: ["job": .codex]`, both differing from the `AppState` defaults per that suite's convention) and assert `defaultAgentKind`, `agentsByKind`, and `agentKind(for: .job)` after the call.

**Mutation-verified, both directions:** with `appState.applyAgentConfig(config)` removed from `CrowDaemon.swift`, the test now fails all three new assertions (`agentsByKind → [:]`, `agentKind(for: .job) → claude-code`); restored, `ConfigAppStateSyncTests` 2/2 and `AppStateAgentConfigSyncTests` 3/3 are green. So the exact mutation you demonstrated is now caught, and the suite docstring's "every field lands" is true again.

**Green (`private(set)`):** leaving as-is, per your read that it's a judgment call — enforcing it would break the direct assignments in `ManagerMigrationTests.swift:112-113,134-141`, and the disk→config→state guard above now catches a dropped choke-point call regardless. Happy to revisit if you'd prefer the type-level enforcement.

dhilgaertner (approved):
## Code & Security Review

Verified the reland against `feature/crow-593-web-ui-parity`. This is a **behavior-preserving refactor plus a real regression guard** — the two setters at the daemon site are replaced by one call that assigns the same two fields. The PR body is accurate about the drift from #735, including the fact that `AppDelegate.swift` no longer exists on this branch and the intent was relanded at `CrowDaemon.applyConfigToAppState` instead.

### Verification performed

**The choke-point claim holds.** `rg` over the tree confirms the only production writes to `defaultAgentKind` / `agentsByKind` are `AppState.swift:99-100`, inside `applyAgentConfig`. Every other write is in test code. The `AppState` class is `@MainActor`-isolated, so the new method inherits the same isolation as the `@MainActor static applyConfigToAppState` that calls it — no concurrency surface added.

**Tests pass as claimed.** `swift test --package-path Packages/CrowCore --filter 'AppStateAgentConfigSync|AppConfigAgentKind|SessionAgentKind'` → 17/17, including the 3 new `applyAgentConfig*` tests. `CrowDaemon --filter DaemonSecurity` → 62/62.

**The daemon guard actually guards.** The comment at `DaemonSecurityTests.swift:265-270` claims that deleting the daemon call site is caught only by the disk→config→state assertion. I mutation-tested this by removing `appState.applyAgentConfig(config)` and re-running: `appliesExcludeReposAndPermissionGatesFromConfig` fails with 3 issues (`defaultAgentKind`, `agentsByKind`, and the live `agentKind(for: .job)` resolve), while the CrowCore unit tests stay green. The claim is exact, and this is the right place to put the guard — the unit tests alone would not have caught a deleted call site.

**Build.** The Swift build completes. `make build` then fails locally on a missing `cargo` toolchain for the desktop app — an environment issue on my machine, unrelated to this diff.

### Security Review

**Strengths:**
- No security surface is touched. The change reads config already loaded under `ConfigStore.withConfigLock` and mirrors two enum-typed fields into runtime state — no new I/O, no deserialization, no privilege or gate change.
- `AgentKind` is a typed value decoded with a `.claudeCode` fallback (`AppConfig.swift:169-170`), so a malformed or absent `config.json` field degrades to the safe built-in default rather than propagating an arbitrary value.
- The `DaemonSecurityTests` change only extends an existing config-sync assertion; the local-only/remote RPC gates are untouched and still pass.

**Concerns:** None found.

### Code Quality

Comments are accurate and earn their place — each states a constraint the code can't show (why the sync exists, which ticket, why the assertion lives at the daemon level). The new tests cover the three meaningful transitions: override update, clearing back to Use default, and a bare default change.

**Green (consider, non-blocking):** `defaultAgentKind` and `agentsByKind` remain `public var` (`AppState.swift:79,84`), so the "single choke point" is enforced by convention rather than by the type system — a future save path can still assign them directly and bypass `applyAgentConfig`. Tightening to `private(set)` would make the invariant structural, but it would break direct assignment in `ManagerMigrationTests` (CrowEngine) and would need those updated, so it is reasonable to leave for a follow-up rather than widen this reland.

### Summary Table

---

PR #743, opened 2026-07-15
fix(crow-desktop): serve frozen bundle web assets under make run (CROW-737)

### LINKED ISSUE #737: fix(crow-desktop): stop the Tauri shell serving web assets live from source
## Problem

In a dev checkout, `make run`'s Tauri window spawns a sidecar `crowd` with `--web-dir` pointed at the source `Resources/web` whenever that folder exists (`crow-desktop/src-tauri/src/lib.rs:181-188`). The desktop app therefore serves the web UI **live from source** — editing `app.js`/CSS/HTML and reloading the window reflects the change with no rebuild.

We don't want `make run` doing live web reloads. It should serve the **frozen, bundle-baked** assets (the copy made into crowd's resource bundle at `make daemon` time) so nothing updates until an explicit rebuild.

## Fix

Delete the `--web-dir` block in `lib.rs` so the sidecar `crowd` is always spawned **without** `--web-dir`. crowd then falls back to its compiled resource bundle (`StaticAssets.webResponse` serves from the bundle when `webDir` is `nil`). Live-from-source editing remains available *explicitly* via `make crowd-dev` (which still passes `--web-dir`); it is simply no longer automatic under `make run`.

## Changes

- `crow-desktop/src-tauri/src/lib.rs` — remove the `let web = …Resources/web; if web.is_dir() { cmd.arg("--web-dir").arg(&web); }` block (~lines 181-188) and its comment.
- `scripts/crowd-dev.sh:14` — the comment *"make run already spawns crowd with live web assets as a sidecar, so for most UI work you don't need this script at all"* becomes inaccurate; update it (`make run` now serves frozen bundle assets; use `crowd-dev` for live web).

## Notes

- Release installs are unaffected — the block is already skipped when the source tree isn't present.
- **Base branch: `feature/crow-593-web-ui-parity`** (integration branch; PR targets it, not `main`).

## Verification

- `make app && make run` in a dev checkout; edit `app.js`, reload the window → change must **not** appear (frozen).
- Confirm the spawned crowd log line has no `--web-dir`.
- `make crowd-dev` still serves live web (unchanged).
- `make daemon` + relaunch → edit now appears.

dgershman:
## Problem

In a dev checkout, `make run`'s Tauri window spawned a sidecar `crowd` with `--web-dir` pointed at the source web folder whenever it existed. The desktop app therefore served the web UI **live from source** — editing `app.js`/CSS/HTML and reloading the window reflected the change with no rebuild.

We don't want `make run` doing live web reloads. It should serve the **frozen, bundle-baked** assets (the copy baked into crowd's resource bundle at `make daemon` time) so nothing updates until an explicit rebuild.

## Fix

- `crow-desktop/src-tauri/src/lib.rs` — remove the `--web-dir` block so the sidecar `crowd` is always spawned **without** `--web-dir`. crowd falls back to its compiled resource bundle (`StaticAssets.webResponse` serves from the bundle when `webDir` is `nil`).
- `scripts/crowd-dev.sh` — update the stale header comment that claimed `make run` already serves live web assets. `make crowd-dev` (still passing `--web-dir`) is now the explicit path for live-from-source web editing.

Release installs are unaffected — the block was already skipped when the source tree isn't present.

## Verification

- [x] `cargo check` on the Tauri crate compiles clean.
- `make app && make run`, edit `app.js`, reload window → change must **not** appear (frozen).
- Spawned crowd log line (`[crow-desktop] spawned crowd …`) has no `--web-dir`.
- `make crowd-dev` still serves live web.
- `make daemon` + relaunch → edit now appears.

Base: `feature/crow-593-web-ui-parity` (integration branch, not `main`).

Closes #737

https://claude.ai/code/session_011EEynTgL9b4vdFBrSwaWpR

### DIFF
--- a/crow-desktop/src-tauri/src/lib.rs
+++ b/crow-desktop/src-tauri/src/lib.rs
@@ -178,14 +178,10 @@ pub fn run() {
                         cmd.arg("--socket").arg(sock);
                     }
                 }
-                // Dev: serve web assets live from source so UI edits show on reload
-                // (matches `make crowd-dev`). Skipped when the source tree isn't
-                // present (e.g. a release install), where crowd uses its bundle.
-                let web = std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
-                    .join("../../Packages/CrowDaemon/Sources/CrowDaemon/Resources/web");
-                if web.is_dir() {
-                    cmd.arg("--web-dir").arg(&web);
-                }
+                // No --web-dir: the sidecar serves the frozen web assets baked
+                // into crowd's resource bundle at `make daemon` time, so UI edits
+                // don't show until an explicit rebuild. For live-from-source web
+                // editing, use `make crowd-dev` (which passes --web-dir).
                 match cmd.spawn() {
                     Ok(child) => {
                         eprintln!(
--- a/scripts/crowd-dev.sh
+++ b/scripts/crowd-dev.sh
@@ -10,9 +10,9 @@
 # Note: Swift can't be hot-swapped into a running process, so `--watch` tears
 # down and respawns the daemon on every change — the same "the server restarts"
 # tradeoff as the Tauri dev loop. Prefer the default (no --watch) when you want a
-# daemon that stays up across edits. And note `make run` already spawns `crowd`
-# with live web assets as a sidecar, so for most UI work you don't need this
-# script at all.
+# daemon that stays up across edits. Note that `make run` spawns its `crowd`
+# sidecar serving the *frozen* bundle-baked web assets (no live reload), so this
+# script is what you want when you need live-from-source web editing.
 #
 # Always binds 127.0.0.1 (loopback only) — front it with an HTTPS reverse proxy
 # for remote access. Env overrides: CROW_HTTP_PORT (8787),


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review

**Strengths:**
- Net reduction in surface area. The sidecar no longer serves the web UI from a path derived from `CARGO_MANIFEST_DIR`; it serves only the assets baked into its own resource bundle. Asset provenance is now build-time-fixed rather than resolved from a writable source tree at spawn.
- Traversal defenses in `StaticAssets` (`isSafe`, `Bundle.module.url(forResource:...)`) are untouched, and the bundle path is the more constrained of the two branches.
- CSP / `nosniff` handling in `fileResponse` is unaffected — headers are applied by filename, independent of which branch loaded the bytes.

**Concerns:**
- None. No auth, input-validation, injection, secrets, or crypto surface is touched.

### Code Quality

Verified the central claim rather than taking it on faith — the fix only works if crowd truly falls back to the bundle *and* the bundle is truly frozen:

- `StaticAssets.swift:66-80` — `webResponse` takes the `webDir` branch only when non-nil, else falls through to `Bundle.module.url(..., subdirectory: "web")`. Dropping `--web-dir` lands on the bundle, not a 404. ✅
- `Packages/CrowDaemon/Package.swift:51-53` — `.copy("Resources/web")` means SwiftPM copies assets into the bundle at build time. This is what makes "frozen until an explicit rebuild" actually true; a symlinked resource would have silently preserved live-reload and defeated the PR's purpose. ✅
- `lib.rs:34` — dev `crowd_bin()` resolves to `.build/debug/crowd`, alongside which SwiftPM emits `CrowDaemon_CrowDaemon.bundle`, so `Bundle.module` resolves on the dev path exactly as it already does for release installs. ✅
- Deletion leaves no orphaned import: the removed block used fully-qualified `std::path::Path::new`, and `env!("CARGO_MANIFEST_DIR")` is still live at `lib.rs:34`.

Two behaviors worth naming for the record — both pre-existing and intentional, neither a defect in this change:

- **Reuse path still serves live.** `lib.rs:157` reuses an already-listening crowd, so a `make crowd-dev` daemon on :8787 will still serve live assets under `make run`. The new comments are correctly scoped to "its `crowd` sidecar" and don't overclaim, and the Makefile already documents reuse — so no change requested.
- **`CROW_WEB_DIR` is inherited.** `CrowDaemon.swift:735` reads the env var, which the sidecar inherits, so an exported `CROW_WEB_DIR` re-enables live serving under `make run`. That reads as a deliberate opt-in escape hatch, not a leak.

Comment accuracy — the reason this PR exists — is good: `crowd-dev.sh`'s stale claim that `make run` already serves live assets is corrected, and the replacement points at `make crowd-dev` as the live-editing path. That was the actively misleading part and it's now right.

**Testing:** No automated coverage, which is appropriate — this is a dev-ergonomics change to process-spawn arguments with no unit-testable seam. `cargo check` could not be independently re-run in this review environment (cargo not installed); the compile-safety argument above is by inspection, and the behavioral claim was verified by tracing the daemon source.

### Summary Table

Small, well-scoped, and it does exactly what it says. The two Green notes are observations about intentional pre-existing behavior, not requests.

PR #746, opened 2026-07-16
Backfill efficiency-scorecard snapshots from telemetry.db + fix empty-state discoverability (closes #745)

### LINKED ISSUE #745: Efficiency scorecard stays empty despite telemetry capturing — snapshots only written one-shot at session completion, no backfill
## Summary

The AI usage efficiency scorecard (ADR-0008, now `Accepted` per #727) renders an **empty state indefinitely** even while Crow is actively running and Claude Code telemetry is being captured. Investigated on a live install — the root cause is **not** that telemetry is off; it's that the scorecard renders exclusively from persisted `SessionAnalyticsSnapshot`s, and **snapshots are only written as a one-shot at the moment a non-Manager session transitions to `.completed`/`.archived`, with no backfill** from the telemetry that's already on disk.

## Evidence (live install, 2026-07-16)

**Telemetry capture is ON and working:**
- `~/Library/Application Support/crow/telemetry.db` exists with the full schema (`metrics`, `events`, `session_map` tables) and an actively-growing WAL (`telemetry.db-wal` ~57 KB, modified today). The OTLP receiver only creates/opens this DB when `config.telemetry.enabled` is true (`AppDelegate.swift` ~L995), so telemetry is enabled and ingesting.

**Yet zero snapshots have ever been persisted:**
- `~/Library/Application Support/crow/store.json` has **no `analyticsSnapshots` key at all** (top-level keys: `hookStates, links, sessions, terminals, worktrees`). The scorecard view builds its model purely from `appState.analyticsSnapshots` (`ScorecardView.swift:23`) and shows the empty state when it's empty (`ScorecardView.swift:38`). Empty snapshots ⇒ permanently empty scorecard.

## Root cause (grounded on `origin/main`)

`SessionService.writeAnalyticsSnapshot(for:status:)` is the **only** writer of snapshots, and it is heavily gated:

- `SessionService.swift:2404` — `guard status == .completed || status == .archived else { return }` — a snapshot is written **only** at a terminal-status transition. A session that stays `active`/`inReview`/`paused` never contributes.
- `SessionService.swift:2405` — `guard !appState.isManagerSession(id) else { return }` — the **Manager session is excluded**. (Notably, a large amount of real usage — e.g. this whole orchestration workload — happens in the Manager tab and is therefore invisible to the scorecard.)
- `SessionService.swift:2419-2423` — `let fresh = await analyticsProvider?(id) … guard let analytics, !analytics.isEmpty else { return }` — if the telemetry aggregate reads empty at the instant of completion (telemetry query racing the final metric flush, session ended before telemetry was enabled, etc.), the snapshot is **silently skipped** with no diagnostic.
- It's triggered only from `scheduleAnalyticsSnapshot` (`SessionService.swift` ~L2402) on the status change / `set-status` RPC — **there is no backfill path** that rebuilds snapshots from `telemetry.db`, even though `telemetry.db` retains ~180 days of raw per-session metrics.

**Net:** to get a single data point, a **non-Manager** session must reach **`completed`/`archived`** *while* telemetry is enabled *and* its aggregate is non-empty at that exact moment. If none of that has happened during a given run (my case — and the likely case for anyone who mostly runs long-lived or Manager-driven sessions), the scorecard is empty forever, and nothing tells the user why.

## Why the current empty-state doesn't help

`ScorecardView.swift:98`: *"The scorecard is computed from analytics snapshots written when sessions complete or archive. Snapshots require Claude Code telemetry, which is off by default — enable it in Settings, then finish a session."*

This is accurate but incomplete: it doesn't convey that (a) the **Manager session is excluded**, (b) already-ended sessions and sessions completed with a momentarily-empty aggregate are lost with **no backfill**, or (c) the ≥5-prompt grading floor (`EfficiencyGrading.Tuning.minimumGradablePromptCount`). A user who has telemetry on and *has* finished sessions can still see nothing.

## Proposed work (prioritized)

1. **Backfill / rebuild snapshots from `telemetry.db`.** This is the highest-value fix. `telemetry.db` already holds the raw per-session metrics (180-day retention); the scorecard should be able to (re)construct `SessionAnalyticsSnapshot`s for any session with telemetry rows — not depend solely on the one-shot write at completion. Fixes "enabled telemetry after sessions already ran" and "completed while the aggregate was momentarily empty."
2. **Make Manager-session usage visible (or clearly documented).** Decide whether Manager-tab usage should be counted (e.g. its own bucket) or is intentionally excluded — and if excluded, say so in the empty-state/docs. Right now the single biggest source of usage is silently invisible.
3. **Stop silently dropping completions.** When `writeAnalyticsSnapshot` skips due to empty analytics (`SessionService.swift:2422`), emit a diagnostic (log + a telemetry-health signal) so "I never see data" is debuggable rather than silent.
4. **Improve discoverability / empty-state.** Explain the real preconditions (non-Manager session → reach completed/archived → telemetry enabled+restarted → ≥5 prompts for a *grade*), and ideally show live capture status ("telemetry capturing — N sessions recorded this week; complete one to populate the scorecard").
5. **End-to-end verification.** `ScorecardWiringTests` / `SessionServiceAnalyticsSnapshotTests` cover the unit happy-path; add a check for the real-world gaps above (backfill from telemetry.db; a completed session with telemetry-on actually surfaces in the scorecard).

## Acceptance criteria

- With telemetry enabled and at least one **already-recorded** (but not snapshotted) session in `telemetry.db`, the scorecard can populate via backfill without requiring that session to be re-run.
- A completion that skips snapshotting produces a visible diagnostic rather than silence.
- The empty-state accurately explains all preconditions (incl. Manager exclusion + the grading floor).
- Manager-session usage is either counted or its exclusion is documented in-app.

---
_Filed from a live diagnosis: telemetry.db actively capturing (WAL growing) but `store.json` has zero `analyticsSnapshots`. The pipeline captures telemetry fine; it just never turns it into scorecard data unless a non-Manager session crosses the completed/archived gate with a non-empty aggregate — and there's no way to backfill._

dhilgaertner:
Closes #745. The scorecard rendered empty despite telemetry capturing because SessionAnalyticsSnapshots are only written one-shot at a non-Manager session's completed/archived transition, with no backfill (store.json had zero analyticsSnapshots while telemetry.db was actively ingesting). Adds a backfill path that rebuilds snapshots from telemetry.db for un-snapshotted sessions (idempotent, retention-aware, preserves richer hook-state fields), a diagnostic when a completion skips on empty analytics, and an empty-state that names all preconditions (Manager exclusion + grading floor). Manager-visibility: Manager usage is now counted in its own ungraded weekly bucket, sourced from time-windowed telemetry queries and persisted merge-only so weeks aged out of retention survive (ADR 0008 addendum documents it as excluded from every graded surface and the baseline). Tests cover backfill + idempotency + the skip diagnostic.

## What changed

- **Backfill (core fix)**: `SessionService.backfillAnalyticsSnapshots()` runs at launch (after telemetry starts, *before* retention pruning) and via a new "Rebuild" button in the scorecard. It enumerates sessions with telemetry rows (new `TelemetryDatabase.sessionIDs()`), skips existing snapshots (never clobbers/downgrades), orphans (deleted session records — status/endedAt would be fabricated), non-terminal sessions, and the Manager session; `endedAt` comes from the session's `updatedAt` so historical sessions bucket into their true ISO week rather than the current one.
- **Manager bucket**: new `ManagerWeeklyUsage` rollups computed from a `received_at`-windowed `sessionAnalytics(for:receivedBetween:)` query (same ISO-week bucketing as `ScorecardModel`), persisted in store.json merge-only, rendered as a "Manager Usage (ungraded)" card.
- **Skip diagnostic**: the previously silent empty-analytics return in `writeAnalyticsSnapshot` now NSLogs and bumps `AppState.analyticsSnapshotSkipCount` (+ timestamp), surfaced in the empty state.
- **Empty state / capture status**: rewritten copy naming all preconditions (telemetry enabled + restart → non-Manager session → completed/archived → ≥5 prompts for a grade, Manager tracked separately); live "Telemetry capturing — N sessions recorded" line backed by new `captureStatus()` (count + newest row), kept fresh by the OTLP receive callback.
- **ADR 0008**: "Backfill and Manager-usage addendum (implemented in #745)".

## Tests

- `Packages/CrowTelemetry/.../BackfillQueriesTests.swift` (temp-file DB fixture): sessionIDs dedup/empty/mapping-only, half-open window boundaries, full-range window == un-windowed (locks the shared-helper refactor), pruned rows disappear, captureStatus.
- `SessionServiceAnalyticsSnapshotTests`: backfill writes with historical endedAt; idempotent + never downgrades; skips orphans/Manager/non-terminal; aged-out (all-zeros) candidate handled gracefully + counted as skip; skip diagnostics increment only for real empty-data completions; Manager refresh merge-only + baseline-window coverage.
- `ScorecardWiringTests`: backfill and Manager rollups keep the AppState mirrors in sync.

`swift build` clean; `make test` passes across all packages except 3 pre-existing environmental failures in `CrowGit` rebase integration tests ("Committer identity unknown" — this machine has no global git identity, only repo-local; untouched by this PR).


dgershman (approved):
## Code & Security Review

Backfill path that rebuilds scorecard snapshots from `telemetry.db`, a merge-only Manager weekly-usage bucket, a skip diagnostic, and a rewritten empty state. Reviewed all 12 files; built and ran the PR's test suites locally.

### Critical Issues
None.

### Security Review
**Strengths:**
- All new SQL is parameterized. The `received_at` window is a compile-time constant clause (`windowClause`) with values bound via `sqlite3_bind_double`; `sessionIDs()`/`captureStatus()` use fully static SQL. No injection surface introduced.
- `sessionIDs()` rejects non-UUID `crow_session_id` rows rather than trusting DB text (`TelemetryDatabase.swift:265`).
- Backfill is conservative: skips orphaned telemetry sessions (would fabricate status/`endedAt`), non-terminal sessions, and the Manager session; never overwrites an existing snapshot (`SessionService.swift:2476`).
- Diagnostics log only session UUID + status — no sensitive content.

**Concerns:** None.

### Code Quality
- **Windowed-aggregate refactor is correct.** Bind offsets line up in every path — `sumMetric`/`countEvents` bind the window at index 3/4, `sumMetricWithAttribute` at 5/6. The `window == nil` path appends no clause and binds nothing, so un-windowed behavior is preserved; `windowedAnalyticsOverFullRangeMatchesUnwindowed` locks this parity. (`TelemetryDatabase.swift:480-556`)
- **Empty-snapshots-with-Manager-usage UI path is crash-safe.** The content pane now renders when only `managerUsageWeekly` is populated; `ScorecardModel.build` degrades to an "insufficient data" `currentWeek` (non-optional, with a `DateInterval` fallback) so the header/grade cards render gracefully with zero snapshots. (`EfficiencyScorecard.swift:107-155`)
- **Backfill written-count is accurate.** `writeAnalyticsSnapshot` persists via `store.mutate` before the `appState` mirror, so the post-write `store.data.analyticsSnapshots?[key] != nil` check reflects reality; the empty-analytics guard correctly suppresses the increment.
- **Persistence is backward-compatible.** New `managerUsageWeekly` is an optional `StoreData` field; new `SessionService`/`insertMetric`/`insertEvent` params are default-valued — old `store.json` decodes and existing callers are unaffected.
- **Manager rollup bucketing matches the graded surface** (ISO-8601 weeks, same `baselineWeekCount` span, merge-only upsert that survives retention pruning). The two self-documented edge cases (partial mid-week prune dip; static `weekKeyFormatter` capturing `.current` TZ) are bounded, self-correcting, and acceptable.
- **Tests are thorough:** half-open window boundaries, full-range == un-windowed, pruned-row disappearance, dedup, idempotency/never-downgrade, orphan/Manager/non-terminal skips, aged-out handling, and mirror-sync coverage.

### Verification
- `swift build` clean (after generating the untracked `BuildInfo.swift`, whose absence is a pre-existing environmental gap unrelated to this PR).
- `SessionServiceAnalyticsSnapshotTests`, `ScorecardWiringTests` — 36 tests pass.
- `BackfillQueriesTests` — 9 tests pass.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 0 Green] findings. Focused, well-tested, backward-compatible, and cleanly documented (ADR 0008 addendum included).

---

PR #748, opened 2026-07-17
feat(crow-daemon): auto-resume sessions after reboot (recreate tmux windows + relaunch every agent) (CROW-747)

### LINKED ISSUE #747: feat(crow-daemon): auto-resume sessions after reboot (recreate tmux windows + relaunch every agent)
## Problem

After a full machine reboot (or any `tmux kill-server`), reopening the app shows all sessions but their terminals are dead — the coding agents (`claude`, `cursor`, `codex`, `opencode`) don't resume.

Root cause: tmux is a normal process with **no reboot persistence** here (no resurrect/continuum). A reboot kills the tmux server, destroying every window/pane *and the agent process inside each*. Crow persists session **metadata** to disk (JSONStore — sessions, worktrees, terminal records), so sessions still render, but the live windows are gone.

On daemon startup the restore path only **adopts** existing windows; it never recreates missing ones:

- `CrowDaemon.swift` (~L491) calls `sessionService.rebuildAllSurfaces()` with the default `forceRegister: false` → the **adopt** path.
- `SessionService.adoptExistingSurfaces()` / `rebuildAllSurfaces()` call `TmuxBackend.adoptTerminal`, which **throws for a missing window** (`adoptThrowsForMissingWindow`, `adoptDistinguishesServerCrashFromMissingWindow`). The catch just **logs and skips** (`SessionService.swift` ~L314) → the surface never renders.
- `reconcileTerminalSurfaces()` (takeover tick, `CrowDaemon.swift` ~L498) then **prunes** terminal records whose window is gone — so records can be dropped rather than resurrected.

So metadata survives; nothing recreates the tmux window or relaunches the agent.

## Key insight — the relaunch machinery already exists and is agent-agnostic

The per-agent resume path is already there and covers all four agents — this is **not** about adding per-agent commands:

- `SessionService.launchAgent` resolves the command via `agent.autoLaunchCommand(...)` (`SessionService.swift` ~L784), keyed on `session.agentKind` (`Session.swift:9`).
- All four agents implement it with resume semantics: `ClaudeCodeAgent.swift:39`, `CursorAgent.swift:58`, `OpenAICodexAgent.swift:41`, `OpenCodeAgent.swift:64` (Claude resumes via `claude --continue`; others their own equivalents; review/job/manager branches vary).
- `rebuildAllSurfaces(forceRegister: true)` already resets readiness so a fresh shell's `.shellReady` fires the relaunch (`SessionService.swift` ~L333-336). It's used on manual takeover (~L1117) but **not** on cold start.

## Ask

On daemon cold start, detect that the tmux server / a session's window is gone and **auto-resume across every supported agent**:

1. Recreate the missing tmux session/window (`TmuxController.newSession -d` / `newWindow`) for each persisted terminal whose window no longer exists.
2. Drive the relaunch through the existing `agent.autoLaunchCommand` path for the terminal's `session.agentKind` (claude / cursor / codex / opencode) — reuse the current resume semantics, don't redefine them.
3. Ensure `reconcileTerminalSurfaces()` does **not** prune those records before the resume runs (ordering).

Likely the cleanest shape is: on a genuine cold start where the tmux server is absent, route through the `forceRegister: true` recreate-and-relaunch path instead of the adopt-only path — while keeping the cheap adopt path for a warm crowd restart where windows still exist (that path is correct then and must not be disturbed). Distinguish the two using the existing server-crash-vs-missing-window signal (`adoptDistinguishesServerCrashFromMissingWindow`).

## Constraints / notes

- Must work for **all** `AgentKind`s (`AgentKind.swift`: claude-code, cursor, codex, opencode), not just Claude — the comments/log strings currently say "claude --continue" but the mechanism is per-agent.
- Don't disturb the warm-restart adopt path (crowd restart *without* reboot, windows still alive).
- Manager session must also come back (`ensureManagerSession`).
- Respect each agent's existing per-session-type resume rules (some review/job/manager branches are intentionally "unsupported").

## Verification

- Repro: start crowd, create sessions on different agents (claude, cursor, codex, opencode), then `tmux kill-server` (fast proxy for reboot) and restart crowd → each terminal should come back with its agent relaunched/resumed per `agentKind`.
- Full: actual reboot, reopen app, confirm the same.
- Regression: a plain crowd restart (no server kill) must still use the adopt path and NOT relaunch agents into live panes.
- Unit: extend `SessionService` / `TmuxBackend` tests for "server gone at startup → recreate window + relaunch" across all agent kinds.

## Base branch

`feature/crow-593-web-ui-parity` (integration branch; PR targets it, not `main`).

dgershman:
Closes #747

## Problem

After a machine reboot (or any `tmux kill-server`), the tmux server and every window — plus the coding-agent process inside each — are destroyed. Crow persists session **metadata** (JSONStore), so sessions still render, but on daemon cold start the takeover only **adopted** existing windows and never recreated the missing ones: terminals came back dead and no agent (`claude` / `cursor` / `codex` / `opencode`) resumed.

## Fix

Route the daemon's first-tick takeover through the **existing** recreate-and-relaunch machinery when the cockpit session is genuinely gone, while keeping the cheap **adopt** path for a warm `crowd` restart where windows are still alive. The two are distinguished by a new non-mutating socket probe — the same server-crashed-vs-missing-window distinction the crash-recovery path (#588) already relies on.

- **`TmuxBackend.cockpitSessionIsLive()`** — non-mutating probe (throwaway controller when none is cached, so it never spuriously trips `onServerLost`) that answers *"did the server survive?"* at boot. `isRunning` can't: its cached `controller` is `nil` at daemon boot regardless of the real server state.
- **`SessionService.takeOverTerminalSurfaces()`** — probe + branch:
  - cockpit **alive** → `rebuildAllSurfaces(forceRegister: false)` — adopt in place, nothing relaunched;
  - cockpit **gone** → `rebuildAllSurfaces(forceRegister: true)` — recreate each persisted terminal's window and relaunch its session's agent via `launchAgent → agent.autoLaunchCommand` keyed on `agentKind` (Claude `--continue`, Cursor/Codex/OpenCode equivalents; each agent's unsupported review/job/manager branches stay unlaunched).
  - The recreate is **gated on the server actually being gone**, so a live pane is never re-registered and no agent is double-launched — the warm-restart regression the ticket calls out. The decision is factored into the pure, unit-tested `shouldRecreateSurfacesOnTakeover(cockpitSessionIsLive:)`.
- **`CrowDaemon.startBoardPoll`** — call `takeOverTerminalSurfaces()` instead of `rebuildAllSurfaces()` on the takeover tick, and **defer** `reconcileTerminalSurfaces()` to subsequent ticks: the per-terminal recreate runs as async `@MainActor` tasks whose fresh bindings haven't landed yet, so pruning on the same tick would drop the very records being resurrected (constraint #3).

The Manager returns via the unchanged `ensureManagerSession`; its window is recreated from the persisted command on the `forceRegister` path. No agent files were touched — the per-agent resume rules already live in each `autoLaunchCommand`.

## Verification

- **Unit / integration (added):** `TmuxBackend.cockpitSessionIsLive()` — alive→gone transition, fresh-backend-reads-socket (warm restart) case, unconfigured; pure `SessionService` takeover-policy tests.
- **Regression:** full **CrowTerminal (91)**, **CrowDaemon (113)**, and CrowEngine takeover suites green; `swift build` clean.
- **Manual (per ticket test plan):** start crowd with sessions on different agents, `tmux kill-server`, restart crowd → each terminal comes back with its agent resumed per `agentKind`; a plain crowd restart (no kill) still adopts and does **not** relaunch into live panes.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
@@ -466,10 +466,13 @@ public enum CrowDaemon {
     /// Broadcasts a `changed` nudge after each poll so clients re-fetch the
     /// boards reactively (M-D).
     ///
-    /// Also performs the terminal TAKEOVER: on the first tick it adopts every
+    /// Also performs the terminal TAKEOVER: on the first tick it restores every
     /// persisted tmux window into this process (so `TerminalRouter.send` /
-    /// `isRegistered` works) and ensures the Manager. Runs before `refresh()` so
-    /// automation dispatches have a live, registered Manager to reach.
+    /// `isRegistered` works) and ensures the Manager. On a warm crowd restart
+    /// the windows are still alive and it adopts them; on a cold start (reboot /
+    /// `tmux kill-server`) it recreates them and relaunches each session's agent
+    /// (CROW-747). Runs before `refresh()` so automation dispatches have a live,
+    /// registered Manager to reach.
     private static func startBoardPoll(
         tracker: IssueTracker,
         eventHub: EventHub,
@@ -488,16 +491,31 @@ public enum CrowDaemon {
                 if let sessionService {
                     if !didTakeOver {
                         await MainActor.run {
-                            sessionService.rebuildAllSurfaces()
+                            // CROW-747: on a genuine cold start (machine reboot
+                            // or `tmux kill-server`) the cockpit session and
+                            // every agent are gone; recreate each persisted
+                            // terminal's window and relaunch its session's agent
+                            // (Claude `--continue`, Cursor/Codex/OpenCode
+                            // equivalents). On a warm crowd restart the windows
+                            // are still live and this adopts them in place —
+                            // gated on the cockpit-alive probe so a live pane is
+                            // never re-registered or double-launched.
+                            sessionService.takeOverTerminalSurfaces()
                             sessionService.ensureManagerSession(devRoot: devRoot)
                         }
                         didTakeOver = true
+                        // Skip reconcile on the takeover tick: the per-terminal
+                        // window recreate above is dispatched as async @MainActor
+                        // tasks whose fresh bindings haven't landed yet, so
+                        // pruning now would drop the very records we're about to
+                        // resurrect (CROW-747). Reconcile runs on every
+                        // subsequent tick, once those tasks have settled.
+                    } else {
+                        // Reconcile terminals ↔ tmux windows each tick: prune
+                        // terminal records whose window is gone and reap orphaned
+                        // windows (targeted-auto, Manager-safe).
+                        await MainActor.run { sessionService.reconcileTerminalSurfaces() }
                     }
-                    // Reconcile terminals ↔ tmux windows each tick: prune terminal
-                    // records whose window is gone and reap orphaned windows
-                    // (targeted-auto, Manager-safe). Runs after `rebuildAllSurfaces`
-                    // on the takeover tick so adopted windows are in the keep-set.
-                    await MainActor.run { sessionService.reconcileTerminalSurfaces() }
                 }
                 await tracker.refresh()
                 await eventHub.broadcast()
--- a/Packages/CrowEngine/Sources/CrowEngine/SessionService.swift
+++ b/Packages/CrowEngine/Sources/CrowEngine/SessionService.swift
@@ -380,6 +380,48 @@ public final class SessionService {
         }
     }
 
+    /// Cold-start terminal takeover for the headless daemon (CROW-747). Probes
+    /// whether the tmux cockpit session survived, then restores accordingly:
+    ///
+    ///   - **Cockpit alive** — a warm `crowd` restart: the daemon process
+    ///     bounced but tmux and its windows kept running, so each agent is still
+    ///     live in its pane. Adopt the surviving windows in place via
+    ///     `rebuildAllSurfaces(forceRegister: false)`; nothing is relaunched.
+    ///   - **Cockpit gone** — a machine reboot or `tmux kill-server`: the server
+    ///     and every window (and the agent process inside each) were destroyed,
+    ///     but session metadata persisted to disk. Recreate each persisted
+    ///     terminal's window and relaunch its session's agent via
+    ///     `rebuildAllSurfaces(forceRegister: true)` — the same
+    ///     recreate-and-relaunch machinery the mid-run crash auto-recovery uses
+    ///     (#588). The relaunch routes through `launchAgent →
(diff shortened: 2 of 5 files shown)


dhilgaertner (approved):
## Code & Security Review

Reviewed the CROW-747 auto-resume change end-to-end: read the full diff plus the surrounding `rebuildAllSurfaces`, `reconcileTerminalSurfaces`, `launchAgent`/readiness wiring, and the tmux `ensureRunningServer`/`hasSession` paths. Built `CrowTerminal`, `CrowEngine`, and `CrowDaemon` cleanly and ran the new tests (all 5 pass).

### Critical Issues
None.

### Security Review
**Strengths:**
- No new external input, network, or credential surface. The probe shells `tmux has-session` against the existing stable socket with args built from already-validated config (`tmuxBinary`/`socketPath`), consistent with every other `TmuxController.run` call.
- `cockpitSessionIsLive()` returns `false` when tmux was never configured, so the recreate branch no-ops safely rather than acting on a half-initialized backend.

### Code Quality
**Strengths:**
- The probe is genuinely non-mutating: `hasSession()` runs `has-session` (never resurrects the server), and the throwaway `TmuxController` is deliberately not cached — verified that caching a session-less handle would spuriously fire `onServerLost` on the next `ensureRunningServer`. Good, well-documented reasoning.
- Recreate is correctly gated on the server actually being gone, so the warm-restart double-launch regression the ticket calls out is avoided; the `.shellReady → launchAgent → autoLaunchCommand` wiring (SessionService.swift:578) confirms every `agentKind` resumes, not just Claude.
- Deferring `reconcileTerminalSurfaces` off the takeover tick is sound: the per-terminal recreate runs as async @MainActor tasks whose fresh bindings haven't landed, so pruning on the same tick could drop records mid-recreate. Reconcile is idempotent and self-heals on every subsequent tick.
- Decision factored into the pure, unit-tested `shouldRecreateSurfacesOnTakeover`; the new integration tests cover the alive→gone transition and the fresh-backend-reads-socket case that `isRunning` is blind to.

**Green considerations (non-blocking):**
- Reconcile is now deferred for *both* branches; on the adopt path it was already safe to run immediately (live bindings point at live windows), so this adds a ~60s delay before dead-record pruning on a warm restart. Harmless given idempotency, but worth a mental note.
- Pre-existing doc comment at `SessionService.swift:291` still says `.shellReady` "drives launchClaude" — the real path is now `launchAgent`. Out of scope here; flagging for a future cleanup.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. Clean, well-tested, well-reasoned reuse of the existing #588 recreate-and-relaunch machinery.

---

PR #755, opened 2026-07-18
fix(crow-desktop): expose WKWebView a11y tree so voice dictation pastes into sessions (CROW-750)

### LINKED ISSUE #750: Web UI: voice dictation (Wispr Flow) doesn't paste into focused session input — research + fix
## Summary

When the cursor is in a Claude Code session's input in the web UI and you dictate with a voice tool (Wispr Flow), releasing the mic button does **not** paste the transcript into the session input. Wispr Flow behaves as if no text field is focused — it asks you to click a target location on screen, and only pastes once you click. Something about our session input makes dictation/accessibility tooling fail to recognize that an editable field is already focused.

**Target branch:** base this work on `feature/crow-593-web-ui-parity` (not `main`). This is a Web UI issue.

## Likely root cause

The session "input box" is **not a normal DOM text field** — it's an **xterm.js terminal** streaming PTY bytes over a WebSocket:

- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/terminal.html` — constructs `new Terminal(...)` (L60), `term.open(...)` (L93), and pipes input to the PTY over a `/terminal` WebSocket (L99–103). The terminal is loaded as a separate document/iframe (`activeTerminal.window` in `app.js`).
- xterm.js takes keyboard/IME input through a **visually-hidden helper `<textarea>`** (`.xterm-helper-textarea`, positioned offscreen / near-zero opacity). It is not a conventional, visible editable field.

Dictation tools like Wispr Flow detect the active editable element and inject text there. Two things plausibly defeat that here:

1. The real editable element is xterm's hidden helper textarea, which such tools often don't treat as a valid focused input (offscreen, non-standard, IME-focused).
2. The terminal runs inside an **iframe** (separate document), so the app's top-level document reports no focused editable field — the tool falls back to its "click where to paste" mode.

## What we want

Research task: figure out why Wispr Flow (and, ideally, dictation/assistive input in general) doesn't recognize the session input as a focused editable field, and propose a solution so that dictated text lands in the session input on mic release without an extra click.

Directions to investigate (not prescriptive):
- Whether exposing a real, focusable editable surface (e.g. a proper `<textarea>` overlay / composer that forwards to the PTY) would let dictation tools target it, instead of relying on xterm's hidden helper textarea.
- Whether the iframe boundary is the blocker, and whether hosting the terminal in the top-level document (or a same-origin arrangement dictation tools can see) fixes it.
- xterm.js accessibility / screen-reader mode options and helper-textarea attributes (role, `aria-*`, positioning) that might make the helper textarea discoverable to dictation tools.
- How Wispr Flow decides where to paste (focused element detection vs. cursor-position fallback), to confirm which of the above is actually responsible.
- Whether a small "voice / paste into session" affordance in the composer is a pragmatic fallback if native detection can't be made reliable.

## Acceptance criteria

- [ ] Root cause documented: why dictation tools don't see the session input as a focused editable field
- [ ] A proposed solution (with trade-offs) that lets dictated text paste into the session input on mic release, no extra click
- [ ] If feasible within scope, implement the fix; otherwise leave a clear, actionable follow-up
- [ ] Verified against Wispr Flow specifically; note behavior for other dictation/assistive input where practical
- [ ] Work branched from `feature/crow-593-web-ui-parity`

## Key references

- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/terminal.html` L44–103 (xterm.js construction + `/terminal` WebSocket input path)
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` — terminal embedding (`activeTerminal`, `ensureTerminal`, L450–451, ~L1322–1335)

dhilgaertner:
Closes #750

## Symptom

In the **desktop app**, dictating with Wispr Flow into a focused session terminal does nothing on mic release — Wispr behaves as if no text field is focused and asks you to click a paste target (its "can't find a focused field" / Scratchpad fallback). The **same web UI in a plain browser works fine.**

## Root cause (the browser-vs-app clue cracked it)

It's the desktop shell's **webview**, not the web page or xterm.

- The desktop app (`crow-desktop`, Tauri v2 / wry 0.55.1) spawns `crowd` and navigates its single window to `http://127.0.0.1:8787` (`crow-desktop/src-tauri/src/lib.rs`), so the web UI runs inside a **macOS WKWebView**. Per ADR 0010 the AppKit app is retired and this Tauri shell is now *the* native surface.
- Wispr Flow finds the focused editable field via the macOS **Accessibility (AX) API**. WKWebView builds its web-content AX tree **out-of-process and lazily** — WebKit (like Chromium/Gecko) waits until the host app's `NSApplication.accessibilityEnhancedUserInterface` flag is set (the flag VoiceOver flips) before it exposes the DOM to the AX API. Safari sets it proactively; a bare wry WKWebView never does. So non-VoiceOver AX clients (dictation tools, window managers — cf. Mozilla bug 1664992, Electron #7206) see no focused text field and fall back to click-to-paste.
- In a real browser the tree is exposed, so dictation works — exactly matching the observed behavior.

Ruled out (the ticket's original guesses): there is **no iframe** (xterm mounts in the top-level, same-origin document; `activeTerminal.window` is a *tmux* window; `index.html` even sets `frame-ancestors 'none'`), and xterm's hidden `.xterm-helper-textarea` is not the blocker — it works fine in a browser. A composer-overlay/`<textarea>` fix would **not** have helped, since it'd live in the same un-exposed WKWebView.

## Fix

Set `accessibilityEnhancedUserInterface = true` on our own `NSApplication` at startup, and re-assert it after the runtime `navigate(...)` to the crowd UI (via `objc2`/AppKit). Self-targeting needs no accessibility permission.

Two mechanisms were tried (empirically):

- ❌ Driving the same attribute through the **AX API** against our own pid (`AXUIElementSetAttributeValue`) returns `kAXErrorNotImplemented (-25208)` — that path is only meant for a *second* process like VoiceOver.
- ✅ The **direct AppKit property setter** (`-[NSApplication setAccessibilityEnhancedUserInterface:]`) takes; a gated `CROW_AX_DEBUG=1` readback confirms `accessibilityEnhancedUserInterface = true` at runtime.

### Trade-offs

- Setting this flag can nudge some window-manager/zoom behavior once at launch (it disables a couple of window animations — Apple forum 659755, Rectangle #129). Benign for this single-window app; watch for any launch-time window reposition.
- It's the same flag VoiceOver sets, so the app now behaves (accessibility-wise) as if an assistive client is attached. That's the intent — it's what makes the web content reachable to dictation/assistive tools generally, not just Wispr.

### Scope

macOS-only, additive, fails safe (it only sets an accessibility flag; if the set were ever a no-op, existing behavior is unchanged). No web-UI changes. Adds `objc2` + `objc2-app-kit` (`NSApplication` feature only) as macOS-only deps — both already in the lockfile via wry, so lockfile churn is just two dep entries.

## Files

- `crow-desktop/src-tauri/src/lib.rs` — `enable_webkit_accessibility()`; called in `setup()` and re-asserted (on the main thread) after `navigate(...)`.
- `crow-desktop/src-tauri/Cargo.toml` / `Cargo.lock` — macOS-only `objc2`/`objc2-app-kit` deps.

## Verification

- ✅ **Runtime readback** (this branch): `CROW_AX_DEBUG=1 ./crow-desktop/src-tauri/target/debug/Crow` logs `accessibilityEnhancedUserInterface = true` at both call sites; no errors; app starts/stops cleanly.
- ⏳ **Wispr Flow (behavioral, reviewer):** on a build of this branch (`make run`), select a session, focus the terminal, hold the Wispr mic, dictate, release — text should paste into the session input **with no extra click**.
- Optional: **Accessibility Inspector** (Xcode) targeting the `Crow` app → the focused terminal element resolves to an `AXTextArea` inside the web area (vs. not reachable before the fix).
- Regression sanity: normal typing, `Cmd+V`, right-click paste, and drag-drop file paste into the terminal still work; window behaves normally at launch.

## Follow-up

If Wispr still can't reach the field on some setups, the next lever is a WKWebView-scoped enable via Tauri's `with_webview` + `objc2-web-kit`; noted here rather than pre-emptively added.

https://claude.ai/code/session_016WEErRk9rgsvpM5e37Tj2j

### DIFF
--- a/crow-desktop/src-tauri/Cargo.lock
+++ b/crow-desktop/src-tauri/Cargo.lock
@@ -572,6 +572,8 @@ name = "crow"
 version = "0.1.0"
 dependencies = [
  "libc",
+ "objc2",
+ "objc2-app-kit",
  "serde",
  "serde_json",
  "tauri",
--- a/crow-desktop/src-tauri/Cargo.toml
+++ b/crow-desktop/src-tauri/Cargo.toml
@@ -29,3 +29,12 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 libc = "0.2"
 
+# macOS only: force-enable the WKWebView's out-of-process accessibility tree so
+# AX-based dictation/assistive tools (Wispr Flow) can find the focused editable
+# field (CROW-750). Both are already in the lockfile via wry's deps.
+[target.'cfg(target_os = "macos")'.dependencies]
+objc2 = "0.6"
+# Only NSApplication is needed (we set one accessibility flag on it), so trim the
+# default AppKit feature surface to keep the build lean.
+objc2-app-kit = { version = "0.3", default-features = false, features = ["NSApplication", "NSResponder"] }
+
--- a/crow-desktop/src-tauri/src/lib.rs
+++ b/crow-desktop/src-tauri/src/lib.rs
@@ -90,6 +90,54 @@ fn wait_for_port(timeout: Duration) -> bool {
     false
 }
 
+/// Force-enable this app's accessibility so the embedded WKWebView exposes its
+/// web-content accessibility tree to the macOS Accessibility (AX) API (CROW-750).
+///
+/// The web UI runs in a WKWebView, which builds its AX tree out-of-process and
+/// only *lazily* — WebKit (like Chromium/Gecko) waits until the host app's
+/// `NSApplication.accessibilityEnhancedUserInterface` is set (the flag VoiceOver
+/// flips) before it exposes the DOM to the AX API. Safari sets this proactively;
+/// a bare wry WKWebView never does. So AX-based dictation tools (Wispr Flow) and
+/// other non-VoiceOver assistive clients see *no* focused text field and fall
+/// back to "click where to paste" instead of pasting on mic release. In a plain
+/// browser the tree is exposed, which is why dictation only breaks in the app.
+///
+/// We set the flag directly on our own `NSApplication` (the internal AppKit
+/// setter — self-targeting needs no accessibility permission). Note: driving the
+/// *same* attribute through the AX API against our own pid returns
+/// `kAXErrorNotImplemented` (that path is only meant for a second process like
+/// VoiceOver), so the direct property setter is the one that actually takes.
+///
+/// Must run on the main thread (it touches `NSApplication`). Known trade-off:
+/// this attribute also nudges some window-manager/zoom behavior once (it
+/// disables a couple of window animations) — benign for a single-window app.
+#[cfg(target_os = "macos")]
+fn enable_webkit_accessibility() {
+    use objc2::msg_send;
+    use objc2::MainThreadMarker;
+    use objc2_app_kit::NSApplication;
+
+    let Some(mtm) = MainThreadMarker::new() else {
+        eprintln!("[crow-desktop] a11y enable skipped: not on the main thread");
+        return;
+    };
+    let app = NSApplication::sharedApplication(mtm);
+    // Safety: `app` is the live shared NSApplication; the selector takes a BOOL.
+    unsafe {
+        let _: () = msg_send![&app, setAccessibilityEnhancedUserInterface: true];
+        // Opt-in readback for verifying the flag actually took at runtime
+        // (`CROW_AX_DEBUG=1`); silent otherwise.
+        if std::env::var_os("CROW_AX_DEBUG").is_some() {
+            let on: bool = msg_send![&app, isAccessibilityEnhancedUserInterface];
+            eprintln!("[crow-desktop][a11y] accessibilityEnhancedUserInterface = {on}");
+        }
+    }
+}
+
+/// No-op on non-macOS: the AX/WKWebView plumbing this addresses is macOS-only.
+#[cfg(not(target_os = "macos"))]
+fn enable_webkit_accessibility() {}
+
 /// Native menu: a Crow app menu, standard Edit (so copy/paste shortcuts work in
 /// the web UI), a View menu with Reload (handy for a web frontend), and Window.
 fn build_menu(app: &tauri::App) -> tauri::Result<tauri::menu::Menu<tauri::Wry>> {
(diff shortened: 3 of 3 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Self-targets `NSApplication` only — no Accessibility TCC grant, no AX calls against other processes (`lib.rs:105-110`).
- macOS-gated via `cfg(target_os = "macos")` with a safe no-op elsewhere; deps are target-specific in `Cargo.toml`.
- Main-thread gated with `MainThreadMarker` — skips instead of UB if called off-main (`lib.rs:120-123`).
- Exposing the WKWebView AX tree matches Safari/VoiceOver semantics; AX clients already require user-granted Accessibility permission. No new injection surface, secrets handling, or privilege escalation.
- Debug path (`CROW_AX_DEBUG`) only logs a bool; no sensitive data.

**Concerns:**
- None security-related. Enabling `accessibilityEnhancedUserInterface` can affect window-manager/zoom animations (documented in the PR); acceptable for this single-window shell, same class of trade-off Electron notes for `AXEnhancedUserInterface`.

### Code Quality
- Root-cause analysis and fix path are sound: direct AppKit setter (not AX API) is the correct WKWebView lever; belt-and-suspenders call after `navigate` via `run_on_main_thread` is appropriate (`lib.rs:207`, `lib.rs:282`).
- `msg_send!` is justified — `objc2-app-kit` 0.3.x has no typed `setAccessibilityEnhancedUserInterface` binding; Safety comment is accurate.
- `cargo check` and `cargo clippy` both clean for `crow-desktop/src-tauri`.
- Green: consider logging if `run_on_main_thread` returns `Err` instead of `let _ = …` (`lib.rs:282`) — primary `setup()` call already covers the common path.
- Green: no automated test (expected for AppKit glue); PR’s `CROW_AX_DEBUG` + manual Wispr verification plan is adequate.

### Summary Table

---

PR #760, opened 2026-07-19
fix(crowd): enforce single writer for store.json (per-store flock, not per-socket) (CROW-759)

### LINKED ISSUE #759: Enforce a single writer for store.json (per-store lock, not per-socket)
## Summary
Crow's single-writer invariant for `store.json` is **not enforced**. The daemon's single-instance guard is keyed to the socket, not the store, so two `crowd` processes launched with different `--socket` values both write the same `~/Library/Application Support/crow/store.json`. Because `JSONStore.mutate` rewrites the **entire** `StoreData` on every change, a stale full-file write from a second writer silently clobbers **every** session the other writer knew about — presenting as "all my sessions disappeared."

## Symptom
- Session records collapse from ~N to a small subset (observed: **2 sessions but 16 worktrees / 24 terminals / 19 links** in `store.json` — the children of wiped sessions linger as orphans).
- The git worktrees/branches themselves survive on disk (the loss is the Crow session records, not the code).

## Root cause
1. **Whole-file writes.** `JSONStore.mutate` serializes and writes the entire `StoreData` every time (`Packages/CrowPersistence/Sources/CrowPersistence/JSONStore.swift:134`). Any lost update = total loss of the losing writer's view, not one record.
2. **Per-socket lock, shared store.** `acquireSingleInstanceLock` (`Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift:707`) takes an `flock` on `<socketPath>.lock`. It's crash-safe, but keyed to the socket. `AppSupportDirectory.url` (the store dir) is fixed regardless of `--socket`, so two daemons with different sockets both pass the guard and write the same file. `DaemonSecurityTests.swift:190-191` asserts "isolated daemon coexists" for distinct sockets — but the **store is not isolated, only the socket is**.
3. **Fail-open footgun.** If the lock file can't be `open()`ed, the guard logs a warning and `return true`s, running unguarded (`CrowDaemon.swift:710-714`).

## Required change
- **Add a store-scoped `flock`** acquired at daemon startup, in addition to the socket lock: lock a file beside the store, e.g. `AppSupportDirectory.url` + `store.json.writer.lock`. If acquisition fails → another process already owns the store → **log LOUD and refuse to start (fail closed).** This enforces "one writer of `store.json`" regardless of `--socket`.
- **Fix the fail-open footgun** for the store lock: if the lock file can't be opened, **fail closed** (do not run unguarded). Keep the existing warning behavior for the socket lock.
- **External-writer tripwire (diagnostic safety net).** In `JSONStore`, remember the store file's mtime+size after each `save`; before the next `mutate` (and in `reload()`), stat the file and if it changed under us since our last write, log LOUD `external write detected — a second writer is touching store.json`, stamped with `pid`. This makes any future bypass immediately attributable in the console the user already tails.
- **Startup log line:** on successful lock acquisition, log the socket path + store path + `pid`, so a second instance is obvious.

## Acceptance criteria
- A second `crowd` pointed at the same store dir (any `--socket`, including a distinct one) **refuses to start and logs why**.
- New `DaemonSecurityTests` case: the store lock rejects a second acquirer for the same store directory, while two daemons with genuinely separate store directories still coexist.
- The socket lock's existing per-socket behavior is unchanged.
- Manual check: `crowd` running; launch `crowd --socket /tmp/crow2.sock` → it exits with a clear log; the tripwire fires in the first daemon's log if a second writer ever does slip through.

## Out of scope (file as separate follow-ups)
- Orphan GC/reaper for worktrees/terminals/links whose owning session is gone.
- `deleteSession` guard that refuses to force-destroy a worktree with uncommitted/unpushed work unless explicitly forced (this is what force-removed in-progress work during automated cleanup).
- Decode-failure recovery that restores the newest good timestamped backup instead of resetting `store.json` to empty.

---
Diagnosed on branch `feature/crow-593-web-ui-parity` (SHA `cb57036`).

dgershman:
Closes #759

## Problem
Crow's single-writer invariant for `store.json` was **not enforced**. `acquireSingleInstanceLock` takes an `flock` on `<socketPath>.lock` — keyed to the **socket**, not the store. `AppSupportDirectory.url` is fixed regardless of `--socket`, so two `crowd` processes on different sockets both pass the guard and write the same `store.json`. Because `JSONStore.mutate` rewrites the **entire** `StoreData`, a stale full-file write from a second writer silently clobbers every session the other writer knew about — the "all my sessions disappeared" report.

## Change
- **Store-scoped `flock`** (`CrowDaemon.acquireStoreWriterLock`) on `<storeDir>/store.json.writer.lock`, acquired at startup in addition to the socket lock. A second daemon pointed at the same store dir (any `--socket`) **refuses to start and logs why**.
- **Fails closed**: a lock-file `open()` error returns false so the daemon refuses to start rather than running unguarded. The socket lock's fail-*open* behavior is unchanged.
- **Startup log line**: `pid` + socket path + store path, so a second instance is obvious in the log the user already tails.
- **reexec** closes the store-writer fd (no `O_CLOEXEC`) so the re-exec'd image can reacquire — mirrors the socket lock.
- **External-writer tripwire** in `JSONStore`: remember `store.json` mtime+size after each `save`/`init`/`reload`; before the next `mutate` (and in `reload`), stat the file and log LOUD, pid-stamped, if it changed under us. Diagnostic safety net — makes any future bypass immediately attributable.

## Tests
`Packages/CrowDaemon` — new `StoreWriterLockTests`:
- `secondAcquireOnSameStoreDirIsRefused` — store lock rejects a second acquirer for the same store dir.
- `distinctStoreDirsCoexist` — two genuinely separate store dirs still coexist.

`SingleInstanceLockTests` (per-socket behavior) unchanged and still green. `CrowPersistence` full suite (45 tests, incl. concurrent-mutate) passes with the tripwire in place.

```
swift test --package-path Packages/CrowPersistence   # 45 passed
swift test --package-path Packages/CrowDaemon --filter 'StoreWriterLockTests|SingleInstanceLockTests'   # 4 passed
```

## Acceptance criteria
- ✅ A second `crowd` pointed at the same store dir (any `--socket`) refuses to start and logs why.
- ✅ New `DaemonSecurityTests` case: store lock rejects a second acquirer; separate store dirs coexist.
- ✅ Socket lock's per-socket behavior unchanged.

## Out of scope (filed as follow-ups per the ticket)
Orphan GC for children of wiped sessions; `deleteSession` force-destroy guard; decode-failure backup restore.


dhilgaertner (changes requested):
## Code & Security Review

Solid, well-scoped fix. The core enforcement — a store-scoped advisory `flock` on `store.json.writer.lock`, acquired in addition to the per-socket lock — correctly closes the gap where two `crowd` processes on distinct `--socket` values shared one fixed store dir and clobbered each other via whole-file `mutate` rewrites. Fail-closed semantics, the reexec fd handoff, and the new tests are all right. Build + all 4 lock tests pass locally; `CrowPersistence` suite green.

One Yellow in the diagnostic tripwire.

### Critical Issues
None.

### Security Review
**Strengths:**
- Lock file created `0o600` (owner-only), consistent with the store's own perms.
- Advisory `flock(LOCK_EX|LOCK_NB)` released by the OS on exit/crash (fd held for process lifetime) — no stale-lock wedge.
- Fail-**closed** on a lock-file `open()` error is the right call for a single-writer invariant: refusing to start beats running unguarded. The divergence from the socket lock's fail-open is deliberate and documented.
- reexec closes `storeWriterLockFD` (no `O_CLOEXEC`) so the re-exec'd image can reacquire — mirrors the socket lock correctly.

**Concerns:**
- None.

### Code Quality

**🟡 Yellow — external-writer tripwire can false-fire during normal concurrent `mutate` (`Packages/CrowPersistence/Sources/CrowPersistence/JSONStore.swift:198-209`)**

`checkForExternalWrite` reads `lastKnownStat` under `writeLock`, **releases** the lock (`:201`), then stats the live file (`:203`):

```swift
writeLock.lock()
let known = lastKnownStat
writeLock.unlock()                        // <-- released here
guard let known else { return }
guard let current = Self.statSignature(fileURL) else { return }  // <-- stat AFTER release
guard current != known else { return }
NSLog("... EXTERNAL WRITE DETECTED ...")
```

A sibling `mutate` on another thread holds `writeLock` across **both** its `save()` and its `lastKnownStat` update (`:188-191`). If that sibling write completes in the window between the `unlock()` and the `statSignature`, then `current != known` even though the **same sole-writer process** authored both writes — a false "EXTERNAL WRITE DETECTED — the single-writer invariant is violated and the next whole-file write will clobber it" alarm during exactly the concurrent-`mutate` path this class advertises and tests as safe (`concurrentMutatesAreConsistent`, 50–100 threads).

Impact is limited: the tripwire is diagnostic-only (`NSLog`, no data effect), and it did not reproduce in 7 runs of the concurrent tests — the window is a single `stat` syscall wide. But the alarm's wording is severe enough to trigger false incident response, and it undermines the stated goal ("if this ever logs, a bypass slipped through"). Fix is a trivial reorder — capture `current` inside the same critical section as `known`:

```swift
writeLock.lock()
let known = lastKnownStat
let current = known == nil ? nil : Self.statSignature(fileURL)
writeLock.unlock()
```

Since `mutate`'s write section holds `writeLock` across save+baseline-update, snapshotting both under the lock makes them consistent for a sole writer while still catching a genuine second *process*.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 0 Green] findings. The enforcement flock is correct as-is; the single Yellow is a same-round-trip fix to the diagnostic's locking so it stops crying wolf under the concurrency the store supports.

---

Issue #728, opened 2026-07-14
Race: IssueTracker's throwaway JSONStore() instances clobber concurrent session writes (active sessions lost from store.json)


dgershman:
Active work sessions can silently disappear from the sidebar — their records vanish from `store.json` while the worktree, branch, open PR, and prompt file all survive. Root cause: **`store.json` writes are not serialized across writers**, so a stale full-store overwrite clobbers a session another writer just added.

Real incident: two active sessions (open issue **and** open PR — so never auto-completed, and reaper-ineligible) vanished; worktrees/branches/PRs intact, records absent from `store.json`.

## Mechanism
- `JSONStore`'s `lock` and `writeLock` are **instance** properties (`Packages/CrowPersistence/Sources/CrowPersistence/JSONStore.swift:52, 60`). Each instance loads its **own** in-memory `_data` from disk at `init` (`:76-92`), and `mutate` writes the **entire** `StoreData` snapshot back (`:102-117`). The `writeSeq`/`lastWrittenSeq` coalescing only orders writes **within a single instance**.
- `SessionService` owns **one** shared instance: `private let store: JSONStore` (`Sources/Crow/App/SessionService.swift:13`, injected from `AppDelegate.swift:558`). All session create/update/delete go through it.
- `IssueTracker` does **not** use it — it constructs **fresh** `JSONStore()` instances at **three** sites: `Sources/Crow/App/IssueTracker.swift:1033, 1452, 1868`. Different instance ⇒ different lock ⇒ its own (possibly stale) disk snapshot.

Race:
1. IssueTracker's refresh constructs `JSONStore()` → reads a `store.json` snapshot predating a just-created/updated session.
2. Concurrently, `SessionService.store` writes that session to `store.json`.
3. IssueTracker's fresh instance finishes and writes its **stale full-store snapshot** → clobbers the session. Because every write overwrites the whole sessions array, the session is silently dropped.

This explains the incident exactly: reaper-ineligible (open issue+PR), no cleanup (worktree survived), yet gone from the persisted store.

## Fix
Make **all writers share the single injected `JSONStore` instance**. Inject `SessionService`'s `store` into `IssueTracker` and replace the three `JSONStore().mutate { … }` calls (`IssueTracker.swift:1033, 1452, 1868`) with it. Then the one instance's `lock` + in-memory `_data` + `writeSeq` actually serialize every mutation, and no writer ever works from a stale disk read.

[...]

PR #761, opened 2026-07-19
Enforce a single writer for store.json via a store-scoped flock

### LINKED ISSUE #759: Enforce a single writer for store.json (per-store lock, not per-socket)

dgershman:
Closes #759

## Problem
Crow's single-writer invariant for `store.json` isn't enforced. The daemon's single-instance guard flocks `<socketPath>.lock` — keyed to the **socket**, not the store. `AppSupportDirectory.url` is fixed regardless of `--socket`, so two `crowd` processes on different sockets both pass the guard and write the same `store.json`. Because `JSONStore.mutate` rewrites the **entire** `StoreData`, a stale full-file write from the second writer silently clobbers every session the first knew about → "all my sessions disappeared" (children linger as orphans).

## Change
- **Store-scoped flock** (`acquireStoreWriterLock(storeDirectory:)`): locks `store.json.writer.lock` beside the store, acquired at startup **in addition** to the socket lock. A second `crowd` for the same store dir — any `--socket` — refuses to start. **Fails closed**: if the lock file can't be opened we refuse to run rather than write unguarded (blast radius = the user's whole session history). The socket lock's per-socket fail-*open* behavior is untouched.
- **External-writer tripwire** in `JSONStore`: remembers `(mtime, size)` after each write/`reload()`; before the next write and in `reload()`, stats the file and — if it changed under us — logs LOUD, pid-stamped `EXTERNAL WRITE DETECTED`. Diagnostic only; the flock is the enforcement.
- **Startup log line**: `pid` + socket + store path, so a second instance is obvious.
- `reexec` closes the store lock fd (no `O_CLOEXEC`) so the re-exec'd image can reacquire it instead of tripping its own fail-closed guard.
- On fresh installs the app-support dir is created before locking (we lock before constructing `JSONStore`, which normally creates it), so a first-run daemon doesn't fail closed on ENOENT.

## Tests
New `StoreWriterLockTests`: a second acquirer for the same store dir is refused; genuinely separate store dirs coexist. Existing `SingleInstanceLockTests` (socket) unchanged. All CrowPersistence + these daemon tests pass locally.

## Out of scope (separate follow-ups, per ticket)
Orphan GC for children of wiped sessions; a `deleteSession` guard refusing to force-destroy uncommitted work; decode-failure recovery from the newest good timestamped backup.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
@@ -48,6 +48,23 @@ public enum CrowDaemon {
             exit(1)
         }
 
+        // Store-writer guard: the socket lock above is keyed to the socket, but
+        // `store.json` lives at a fixed app-support path regardless of --socket,
+        // so two daemons on DIFFERENT sockets would still both write it — and
+        // because every mutate rewrites the whole file, the loser's sessions
+        // vanish silently. Lock the store itself (fail closed) so exactly one
+        // crowd may write it, whatever --socket each was launched with (#759).
+        let storeDirectory = JSONStore.defaultDirectory
+        guard acquireStoreWriterLock(storeDirectory: storeDirectory) else {
+            log("FATAL: another process already holds the store writer lock for "
+                + "\(storeDirectory.path) — refusing to start. Only ONE crowd may write store.json, "
+                + "regardless of --socket. Stop the other daemon first (a second writer silently "
+                + "clobbers every session the first one knew about).")
+            exit(1)
+        }
+
+        log("started: pid \(getpid()); socket \(options.socketPath); store \(storeDirectory.appendingPathComponent("store.json").path)")
+
         let store = JSONStore()
         let git = GitManager()
         let appState = await seedAppState(from: store)
@@ -316,6 +333,13 @@ public enum CrowDaemon {
         if singleInstanceLockFD >= 0 {
             close(singleInstanceLockFD)
             singleInstanceLockFD = -1
+        }
+        // Same rationale for the store lock (#759): an inherited fd keeps the
+        // flock held, so the re-exec'd image's fail-closed store guard would
+        // refuse to start.
+        if storeWriterLockFD >= 0 {
+            close(storeWriterLockFD)
+            storeWriterLockFD = -1
         }
         guard let argv0 = argv.first, !argv0.isEmpty else {
             log("re-exec failed: empty argv")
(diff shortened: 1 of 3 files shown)


dhilgaertner (approved):
## Code & Security Review

Focused, well-scoped fix for #759. The socket lock guarded the wrong resource (socket, not store); a store-scoped `flock` on `store.json.writer.lock` is the correct primitive and closes the "two daemons on different `--socket`s both write the fixed app-support store" gap. Verified locally: `CrowDaemon` builds clean, `StoreWriterLockTests` (2) pass, and all 45 `CrowPersistence` tests (incl. concurrency) pass.

### Critical Issues
None.

### Security Review
**Strengths:**
- Lock file created `0o600` (owner-only), matching the store's own permissions.
- **Fail-closed** on the store lock (`CrowDaemon.swift:777`) is the right call — the blast radius of a lost race is the user's entire session history, and the asymmetry vs. the socket lock's fail-*open* is explicitly reasoned in the doc comment.
- `flock` is advisory but process-lifetime-scoped and auto-released by the kernel on exit/crash, so a killed daemon never wedges the store permanently.
- Re-exec correctly closes `storeWriterLockFD` before `execv` (`CrowDaemon.swift:340`) so the re-exec'd image can reacquire — without it the fail-closed guard would refuse to restart. Mirrors the existing socket-lock handling.
- No secrets, no injection surface, no new network exposure.

### Code Quality
- `defaultDirectory` (`JSONStore.swift:115`) resolves to the same `AppSupportDirectory.url` the no-arg `JSONStore()` writes to, so lock and store dirs can't drift.
- First-run `createDirectory` before locking (`CrowDaemon.swift:774`) correctly avoids a spurious ENOENT fail-closed on fresh installs.
- `reload()` re-baselines the signature after `detectExternalWrite` (`JSONStore.swift:135-138`), so the intended app→daemon follow pattern (reload before mutate) stays quiet; the tripwire is genuinely diagnostic-only as documented.

- **Tripwire vs. app+daemon coexistence** (`JSONStore.swift:212`): the `flock` guards daemon-vs-daemon only; the desktop app never acquires it. In a true app+daemon-both-mutating window the tripwire will fire (correctly) *and* the whole-file write still clobbers — but that's the #728/app-vs-daemon problem the PR explicitly scopes out, so this is a note, not a request.
- **Test hygiene** (`DaemonSecurityTests.swift:207-224`): the two new `@Test`s run in parallel and both write the shared `nonisolated(unsafe) static storeWriterLockFD`, and each acquired fd is leaked (overwritten without `close`). Harmless here — outcomes don't depend on the static, temp dirs are distinct, and leaks are bounded to the test process — but a `defer { close }` + serialized suite would silence any future TSan noise.
- **`FileSignature` granularity** (`JSONStore.swift:96`): `(mtime, size)` can miss a same-second, same-size external rewrite. Fine for a diagnostic; noting the known limit.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 3 Green] findings. Correct primitive, well-reasoned fail-closed semantics, clean build + green tests, and residual gaps are explicitly scoped out per the ticket.

---

PR #784, opened 2026-07-20
fix(web): restore the in-progress spinner on ticket refresh (CROW-771)

### LINKED ISSUE #771: Restore web parity: in-progress spinner on ticket refresh (manual + auto-refresh)
## Summary

Restore the **in-progress spinner** on ticket refresh, lost in the native→web move
(ADR-0010, CROW-593). In the native app, pressing **Refresh** in the tickets area — and the
automatic every-minute ticket refresh — showed a small spinner while it fetched. The web UI
gives **no in-progress indication**; the list just silently updates when data arrives.

## Current behavior (web)

`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js`:

- `refreshTickets()` (~L2713) is fire-and-forget with no in-flight state:
  ```js
  async function refreshTickets() {
    try { await rpc('refresh-tickets'); } catch (_) { /* app down — ignore */ }
    setTimeout(() => refreshBoard('tickets'), 1200);
  }
  ```
- The tickets-summary card refresh button `↻` (~L767) and the Ticket Board "Refresh" button
  (~L2458) both call `refreshTickets()` with no spinner / disabled / spinning state.
- The auto-refresh interval (~L3605) — `setInterval(() => { if (selectedBoard === 'tickets' ||
  selectedBoard === 'reviews') refreshBoard(selectedBoard); }, 20000)` — also shows no
  indicator while refreshing.

## Native behavior (what to replicate)

Native `TicketBoardView` (git history at `eb7a489^`) rendered a `ProgressView()` spinner
(lines ~45 and ~606) during refresh — both on manual refresh and while an automatic refresh
was in flight — so the user always saw that a fetch was happening.

## Scope

- [ ] Add an in-flight state to `refreshTickets()` and reflect it on the tickets-summary card
      `↻` button and the Ticket Board "Refresh" button (spinner / spinning glyph / disabled
      while pending).
- [ ] Show the same indicator during the **automatic** ticket refresh (interval + any
      daemon poll-nudge-driven refresh), matching the native every-minute spinner.
- [ ] Ensure the indicator clears on success and on error (the RPC is best-effort and
      swallows errors today).
- [ ] Apply the same treatment to the Reviews board refresh for consistency (native showed it
      there too — see the second `ProgressView` in the native board view).

## Acceptance criteria

- [ ] Pressing Refresh in the tickets area shows a spinner until the refreshed list is
      rendered, then clears.
- [ ] The automatic (~1 min) ticket refresh shows the same in-progress indicator.
- [ ] No indicator gets stuck on if the refresh RPC fails or the daemon is down.

## References

- Web: `.../Resources/web/app.js` — `refreshTickets()` (~L2713), tickets card `↻` (~L767),
  Ticket Board Refresh (~L2458), auto-refresh interval (~L3605).
- Native (history): `Packages/CrowUI/Sources/CrowUI/TicketBoardView.swift` @ `eb7a489^`
  (`ProgressView()` at ~L45 and ~L606).

dhilgaertner:
Closes #771

Restores the in-progress spinner on ticket refresh, lost in the native→web move (ADR-0010, CROW-593).

## The signal already existed

`list-tickets` has always returned `"loading": appState.isLoadingIssues` (`RPCHandlers.swift:606`, `EngineRouter.swift:297`) — the web simply ignored the field. The indicator now ORs two sources:

- **`boardData.tickets.loading`** — the daemon's flag. Covers the **automatic** poll (what native's every-minute spinner showed) and any manual refresh outliving the client's 10s rpc timeout.
- **`ticketRefreshPending`** — local, optimistic. Covers the gap between the click and the first board re-read so the button reacts instantly.

Both live in module state: the web re-renders by destroy-and-rebuild, so DOM-only state would be wiped by `renderBoard()`. `sidebarSignature()` gains the predicate, otherwise its diff guard stops the sidebar `↻` from ever repainting.

## Making the automatic refresh visible

`startBoardPoll` broadcast its `changed` nudge only *after* `tracker.refresh()`. Since `isLoadingIssues` is runtime-only (not store-backed), the in-flight window was invisible to clients — the spinner would appear only if a 20s board poll happened to land inside it. Added a broadcast *before* the fetch too. The rate-limit guard inside `refresh()` returns before the flag is set, so a skipped poll correctly shows nothing.

## Two notes on the ticket

- **Its Reviews premise was off.** Both native `ProgressView`s (`TicketBoardView.swift:45` and `:606` @ `eb7a489^`) are tickets — the board header and the sidebar Tickets card. Native never had a Reviews spinner. Reviews still get the indicator for consistency, on its existing daemon re-read; its Refresh has no provider-fetch RPC and that behavior is unchanged.
- **`isLoadingReviews` is unusable as a signal** — `IssueTracker.swift:542-572` sets and clears it inside one synchronous block, so it's never observable as `true`. Noted, left alone.

## Not stuck-on

`refreshTickets()` clears its flag in a `finally`, so a failed RPC, a >10s fetch, or a downed daemon can't strand the spinner (AC3).

## Styling

Reuses the CROW-749 `.action-spinner`. It nests *inside* `.board-title` because that carries `margin-right: auto` — a sibling would be shoved to the right edge with the buttons instead of sitting beside the title as native did. The sidebar glyph needs its own keyframe rather than `wizard-spin`: the button is absolutely positioned with `translateY(-50%)`, which a bare `rotate(360deg)` would clobber, so the keyframe composes both transforms.

## Verification

- `web-tests`: **43 pass, 0 fail** — 5 new cases covering each flag independently, the sidebar `↻` (spinning class + disabled + title), and Reviews.
- `swift build` clean (the two remaining warnings are pre-existing, in `RPCWebSocketHandler.swift`).
- `swift test` in `Packages/CrowDaemon`: **116 tests, 32 suites, all pass**.
- Not done: live in-browser check. The running daemon is built from the main checkout, and starting a second `crowd` from this worktree would violate the single-writer store rule and disturb live sessions. Reviewer running this locally is worth a look at the animation.

https://claude.ai/code/session_019EmyGe57KG6N26Hs222dFw

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
@@ -566,6 +566,13 @@ public enum CrowDaemon {
                         await MainActor.run { sessionService.reconcileTerminalSurfaces() }
                     }
                 }
+                // Nudge *before* the fetch as well as after: `isLoadingIssues`
+                // is runtime-only (not store-backed), so without this the
+                // in-flight window is invisible to clients and the web's
+                // every-minute refresh spinner never appears (CROW-771). The
+                // rate-limit guard inside `refresh()` returns before the flag
+                // is set, so a skipped poll correctly shows no spinner.
+                await eventHub.broadcast()
                 await tracker.refresh()
                 await eventHub.broadcast()
                 try? await Task.sleep(nanoseconds: 60 * 1_000_000_000)
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
@@ -155,6 +155,15 @@ public enum CrowDaemon {
         // Write-actions that mutate session state run locally on the daemon's
         // own SessionService / store — the sole authority (ADR 0007).
 
+        // Push a `changed` nudge whenever the tracker's in-flight state moves,
+        // so clients can render a refresh indicator for the *automatic* poll.
+        // `isLoadingIssues` is runtime-only (not store-backed), so nothing else
+        // announces it. Wired unconditionally — it needs no cockpit, unlike the
+        // automations below (CROW-771).
+        await MainActor.run {
+            tracker.onLoadingIssuesChanged = { Task { await eventHub.broadcast() } }
+        }
+
         // Drive the board poll. It also performs the terminal "takeover" (adopt
         // persisted tmux windows + ensure the Manager) on startup (CROW-581).
         startBoardPoll(
@@ -566,13 +575,11 @@ public enum CrowDaemon {
                         await MainActor.run { sessionService.reconcileTerminalSurfaces() }
                     }
                 }
-                // Nudge *before* the fetch as well as after: `isLoadingIssues`
-                // is runtime-only (not store-backed), so without this the
-                // in-flight window is invisible to clients and the web's
-                // every-minute refresh spinner never appears (CROW-771). The
-                // rate-limit guard inside `refresh()` returns before the flag
-                // is set, so a skipped poll correctly shows no spinner.
-                await eventHub.broadcast()
+                // The in-flight half of the refresh is announced by
+                // `tracker.onLoadingIssuesChanged` (wired at startup), which
+                // fires *after* `isLoadingIssues` is set — nudging from here
+                // beforehand would race the flag and fire on skipped polls
+                // (CROW-771).
                 await tracker.refresh()
                 await eventHub.broadcast()
                 try? await Task.sleep(nanoseconds: 60 * 1_000_000_000)
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -472,6 +472,8 @@ html, body {
 
 .board-head { position: sticky; top: 0; z-index: 2; background: var(--bg-deep); display: flex; align-items: center; flex-wrap: wrap; gap: 10px; margin-bottom: 14px; padding: 4px 0 8px; }
 .board-title { color: var(--gold); font-size: 18px; font-weight: 700; margin-right: auto; }
+/* In-flight refresh spinner beside the board title (CROW-771). */
+.board-title .action-spinner { margin-left: 10px; vertical-align: middle; }
 .done-chip {
   color: var(--gold); background: rgba(221, 196, 130, 0.14);
   padding: 2px 8px; border-radius: 10px; font-size: 11px; font-weight: 600;
@@ -625,6 +627,17 @@ html, body {
   border: 1px solid var(--border-subtle); background: var(--bg-card); color: var(--text-secondary); font-size: 12px;
 }
 .tickets-refresh:hover { color: var(--text-primary); border-color: var(--border-strong); }
+/* In-flight ticket refresh (CROW-771). Can't reuse `wizard-spin`: the button is
+   absolutely positioned with `translateY(-50%)`, which a bare rotate keyframe
+   would clobber — so the keyframe composes both transforms. */
+.tickets-refresh.spinning {
+  opacity: 0.6; cursor: default;
+  animation: tickets-refresh-spin 0.7s linear infinite;
+}
+@keyframes tickets-refresh-spin {
+  from { transform: translateY(-50%) rotate(0deg); }
+  to   { transform: translateY(-50%) rotate(360deg); }
+}
 .tickets-row { display: flex; align-items: stretch; gap: 6px; margin: 2px 2px 8px; }
 .tickets-row .tickets-card { flex: 1 1 auto; margin: 0; }
 .sidebar-tools { display: flex; flex-direction: column; justify-content: center; gap: 4px; flex: 0 0 auto; }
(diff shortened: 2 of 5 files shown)


dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- UI-only change; no new RPC surface, auth paths, or credential handling.
- Spinner markup uses `el()` / `textContent` (no HTML injection).
- Reuses the existing `changed` nudge + `loading` boolean already exposed by `list-tickets`.

**Concerns:**
- None security-related.

### Code Quality
- **Yellow — stale `loading` can strand the spinner (AC3):** `ticketsRefreshing()` ORs `boardData.tickets.loading` with no client-side escape hatch. Two concrete stuck-on paths:
  1. `persistSidebarCache()` / `restoreSidebarCache()` (`app.js` ~494–502, ~483–492) persist the full tickets payload, including runtime-only `loading`. A cache written mid-fetch (`loading: true`) + failed boot `refreshBoard('tickets')` (daemon down) leaves the ↻ / board spinner on indefinitely — `refreshBoard` returns early on RPC failure without clearing the flag (`app.js` ~1985).
  2. Mid-session: client observes `loading: true` via the pre-fetch nudge, then the daemon dies before the post-fetch broadcast. Periodic `list-tickets` keeps failing; unlike `ticketRefreshPending`'s `finally`, nothing clears the daemon half.
  
  Fix: strip / never persist `loading` in the sidebar cache (and/or force `boardData.tickets.loading = false` on `list-tickets` failure / RPC disconnect). The local pending flag is already AC3-safe.

- **Green — pre-broadcast vs flag timing:** `CrowDaemon.swift` ~575 broadcasts *before* `tracker.refresh()` raises `isLoadingIssues`. Visibility for the automatic spinner relies on the client's 50ms `onServerChanged` debounce winning the race. Works in practice for multi-second provider fetches; cleaner to nudge from inside `refresh()` immediately after the flag is set (or set the flag, broadcast, then fetch).

- **Green — double nudge on skipped polls:** the pre-broadcast still fires when `refresh()` rate-limit-returns early (no spinner, but clients still re-fetch). Harmless churn.

- Manual path (`ticketRefreshPending` + `finally` + coalesce) and Reviews local pending look solid; CSS transform composition for `.tickets-refresh.spinning` is correct.
- `web-tests`: 43 pass, 0 fail (incl. 5 new spinner cases). `swift build --package-path Packages/CrowDaemon` clean (pre-existing `RPCWebSocketHandler` warning only).

### Summary Table

---

dhilgaertner:
Thanks @dgershman — all three findings addressed in 9cd2597.

### 🟡 Stale `loading` can strand the spinner

Confirmed both paths, and you were right that the local `finally` only protects half the state. Fixed with a single `clearDaemonRefreshFlag()` invoked from **every** path that loses authority over the flag:

| Path | Hook |
|---|---|
| `list-tickets` read fails | `refreshBoard`'s `catch` |
| daemon dies / socket drops | `ws.onclose` |
| boot from a mid-fetch cache | `restoreSidebarCache` |

`persistSidebarCache` also strips `loading` on the way out (your suggested fix), **and** restore scrubs it independently — belt and braces, because a cache written by an older build can still carry `loading: true`.

The `ws.onclose` hook turned out to be the important one, beyond the two paths you listed: the 20s interval only polls an *open* board, so if the daemon died while the user sat on a session, a stuck sidebar `↻` would spin until they opened a board. `onclose` catches that regardless of view.

### 🟢 Pre-broadcast vs flag timing

Agreed, and I took the cleaner option you suggested rather than leaving it to the 50ms debounce winning a race. Added `onLoadingIssuesChanged` to `IssueTracker` — matching the existing `onNewReviewRequests` / `onPRStatusTransitions` callback pattern — fired immediately after `isLoadingIssues` flips in both directions, wired to `eventHub.broadcast()` at daemon startup. The nudge now provably cannot precede the flag.

### 🟢 Double nudge on skipped polls

Falls out of the same change: the callback only fires on a real transition, and the rate-limit guard returns before the flag is ever set, so a skipped poll is now completely silent. The `startBoardPoll` pre-broadcast is gone.

### Verification

- `web-tests`: **50 pass, 0 fail** — 7 new, covering the clear helper (including idempotency), the persist→restore round-trip, and a legacy cache carrying `loading: true`.
- `swift build` clean; `swift test` in `Packages/CrowDaemon`: **116 tests, 32 suites, all pass**.

Still not verified live in a browser, for the reason in the PR description — the running daemon is built from the main checkout and a second `crowd` from this worktree would violate the single-writer store rule.

dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- UI-only indicator work; no new RPC methods, auth paths, or credential handling.
- Spinner/DOM built via `el()` + `textContent` / className — no HTML injection surface.
- Reuses the existing `list-tickets` `loading` boolean and `changed` EventHub nudge rather than widening the API.

**Concerns:**
- None security-related.

### Code Quality
Prior request-changes items from the first Crow review are addressed in `9cd2597`:

- **Stale `loading` (was Yellow):** `clearDaemonRefreshFlag()` on `list-tickets` failure, RPC `onclose`, and cache restore; `persistSidebarCache` strips `loading` and restore scrubs legacy caches. AC3-safe.
- **Pre-broadcast race (was Green):** replaced with `IssueTracker.onLoadingIssuesChanged`, fired after the flag flips both ways; skipped polls stay silent.
- **Double nudge on rate-limit skip (was Green):** falls out of the same callback — no transition, no broadcast.

Remaining notes (non-blocking):

- **Green — early `refresh()` return after flag set:** if `ConfigStore` fails after `isLoadingIssues = true`, defer clears immediately → brief true/false broadcast pair. Harmless flash.
- **Green — end-of-poll double broadcast:** clear-callback + `startBoardPoll`'s post-`refresh()` broadcast; client's 50ms `onServerChanged` coalesce absorbs it.
- **Green — no Swift test for `onLoadingIssuesChanged`:** web coverage is strong (50 pass); an IssueTracker transition test would lock the daemon half.

Manual path (`ticketRefreshPending` + `finally` + coalesce), Reviews local pending, sidebar signature inclusion, and CSS transform composition for `.tickets-refresh.spinning` all look correct.

**Verification:** `web-tests` 50/50 pass; `swift build --package-path Packages/CrowDaemon` clean.

### Summary Table

---

PR #785, opened 2026-07-20
Run the OTLP telemetry receiver in the daemon (restores the per-session analytics strip)

### LINKED ISSUE #772: Restore web parity: per-session analytics/cost bar on session detail (telemetry pipeline dormant in daemon)
## Summary

Restore the **per-session analytics/cost bar** on the session detail screen, lost in the
native→web move (ADR-0010, CROW-593). With observability/telemetry (OTLP) enabled, the native
app showed a strip with **Cost, Tokens, Tools, Active time, Duration, Lines +/-, API errors**.
In the web UI it never appears — because the **telemetry ingestion pipeline was never migrated
to the daemon**, so the data the strip needs is never produced.

## Root cause (investigated)

The render code, the DTO, and the RPC field all still exist — only the data source is dead:

- **Web render exists, correctly gated:** `renderSessionAnalyticsStrip(s, root)`
  (`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` ~L1616) reads
  `liveFor(s.id).analytics` and returns early if it's absent. Absence = nothing rendered (by
  design).
- **Daemon sends it — when it has it:** the `list-sessions-live` handler
  (`Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift` ~L751–797) attaches an
  `analytics` `SessionAnalyticsDTO` built from `appState.existingHookState(for: id)?.analytics`
  (live) or `appState.analyticsSnapshots[id]` (durable). Both are **empty at runtime**.
- **Why they're empty:** `TelemetryService` (the OTLP HTTP/JSON receiver, CrowTelemetry) that
  used to run inside the retired macOS app and populate `SessionHookState.analytics` in real
  time is **never instantiated in the daemon**. In `CrowDaemon.swift`, `telemetryPort` is
  effectively hardcoded `nil`, so no receiver is bound, `hookState.analytics` is never filled,
  no end-of-session snapshot is written, and the web strip's gate is never satisfied.
- **Native did it:** native `SessionDetailView` (`eb7a489^`) rendered
  `SessionAnalyticsStrip(analytics:)` straight off `appState.hookState(id).analytics`, which
  the in-process `TelemetryService` kept current while telemetry was on.

Net: this is not a missing UI — it's a **dormant backend pipeline**. The strip will light up
once the daemon runs the telemetry receiver and feeds `hookState.analytics`.

## Scope

- [ ] Start `TelemetryService` in the daemon when `config.telemetry.enabled` is set; bind the
      OTLP receiver on the configured port (`CrowDaemon.swift`, near the TerminalCockpit
      conditional).
- [ ] Thread the real telemetry port into `EngineContext` instead of `telemetryPort: nil`.
- [ ] Wire the receiver's per-session metrics into `appState.hookState(id).analytics` in real
      time (serialize on `@MainActor`), and keep the end-of-session snapshot write so the
      durable fallback (`analyticsSnapshots`) works.
- [ ] Verify `renderSessionAnalyticsStrip` then renders Cost/Tokens/Tools/Active/Duration/
      Lines/Errors on the detail screen (no web change expected beyond confirmation).

## Acceptance criteria

- [ ] With telemetry enabled, an active session's detail screen shows the analytics strip with
      live values, updating as the session runs.
- [ ] After a session ends, the strip still shows values from the durable snapshot.
- [ ] With telemetry disabled, the strip stays hidden (unchanged behavior).

## References

- Web: `renderSessionAnalyticsStrip` + `liveFor` (`.../Resources/web/app.js` ~L1616, ~L602).
- Daemon: `RPCHandlers.swift` `list-sessions-live` (~L751–797); `CrowDaemon.swift`
  telemetry wiring (`telemetryPort` currently nil); `SessionAnalyticsDTO` /
  `SessionAnalytics` (CrowCore); `TelemetryService` (CrowTelemetry).
- Native (history): `SessionDetailView.swift` / `SessionAnalyticsStrip` @ `eb7a489^`.
- ADR-0010 (`docs/adr/0010-retire-the-macos-app.md`). Related: telemetry/OTLP settings already
  exist in Settings → General.

dhilgaertner:
Closes #772

## Root cause

Not a missing UI. The web render (`renderSessionAnalyticsStrip`), the `list-sessions-live` `analytics` field, and `writeAnalyticsSnapshot` all survived the native→web move intact. What never migrated was the **producer**: `rg TelemetryService --type swift` matched exactly one file — its own definition — and `CrowTelemetry` was not even a dependency of `CrowDaemon`.

Two things were dead at once:

1. No OTLP receiver was bound, so `hookState.analytics` was never populated.
2. `SessionService` and `EngineContext` were built with `telemetryPort: nil`, so `AgentLaunch` / `ClaudeCodeAgent.autoLaunchCommand` never exported the `CLAUDE_CODE_ENABLE_TELEMETRY` / `OTEL_*` prefix — Claude Code was not emitting anything to receive in the first place.

The strip treats absent analytics as its empty state (by design), so it silently rendered nothing. Confirmed live: nothing was listening on port 4318 despite `telemetry.enabled: true`.

## Changes

- **`CrowDaemon.swift`** — start `TelemetryService` when `config.telemetry.enabled`; its `onDataReceived` (already `@MainActor`) writes `appState.hookState(id).analytics` and refreshes `telemetryCaptureStatus`. Thread the real port into both `SessionService` and `EngineContext`. Restore the lifecycle the native AppDelegate owned: start → scorecard backfill + Manager rollups → retention prune (that order, so rows about to age out still snapshot), and `stop()` before exit/re-exec.
- **`SessionService.swift`** — `backfillAnalyticsSnapshots` / `refreshManagerUsage` made `public` for the daemon; new injected `telemetryDeleteProvider` awaited in `deleteSession`, so all three delete paths (RPC handler, engine-router fallback, auto-cleanup reaper) purge telemetry rows through one choke point. `SessionAnalyticsSnapshot`s are deliberately kept — the scorecard aggregates history.
- **`Packages/CrowDaemon/Package.swift`** — add `CrowTelemetry`. Darwin-only (Network.framework), which `CrowDaemon` already was via `CrowTerminal`; `ci.yml`'s Linux allow-list excludes both.

No web change — the render code was already correct.

## Tests

- `LocalLiveActionTests` — two new cases in `SessionAnalyticsStripLiveTests` pinning the branch this fix makes reachable: an open session with a live hook aggregate renders `source: "live"`, and a live aggregate wins over a stale snapshot. The suite previously covered only the snapshot and omission branches.
- `SessionServiceTelemetryDeleteTests` (new) — deleting a session invokes the provider once with its id; the Manager session is never purged.

## Verification

- `swift build --product crowd` ✅
- `swift test --package-path Packages/CrowDaemon` — 118 tests ✅
- `swift test --package-path Packages/CrowEngine` — 313 tests ✅
- `swift test --package-path Packages/CrowTelemetry` — 25 tests ✅

Not yet exercised against a live daemon (would require restarting the running `crowd`): with telemetry enabled, expect the `[OTLPReceiver] Listening on localhost:4318` log, the `export CLAUDE_CODE_ENABLE_TELEMETRY=1 OTEL_… &&` prefix on spawned Claude sessions, and the strip filling in within one 4s `refreshLive` tick.

https://claude.ai/code/session_01Si2urYjhSLHt2Y1S276wMg

### DIFF
--- a/Packages/CrowDaemon/Package.swift
+++ b/Packages/CrowDaemon/Package.swift
@@ -23,6 +23,11 @@ let package = Package(
         .package(path: "../CrowCodex"),
         .package(path: "../CrowCursor"),
         .package(path: "../CrowOpenCode"),
+        // OTLP receiver + telemetry.db, so the daemon produces the per-session
+        // analytics the web strip and the scorecard read (#772). Darwin-only
+        // (Network.framework) — CrowDaemon already is, via CrowTerminal, and
+        // ci.yml's Linux allow-list excludes both.
+        .package(path: "../CrowTelemetry"),
         .package(url: "https://github.com/hummingbird-project/hummingbird.git", from: "2.0.0"),
         .package(url: "https://github.com/hummingbird-project/hummingbird-websocket.git", from: "2.0.0"),
         // Already resolved transitively (via NIO/Hummingbird); declared directly
@@ -44,6 +49,7 @@ let package = Package(
                 .product(name: "CrowCodex", package: "CrowCodex"),
                 .product(name: "CrowCursor", package: "CrowCursor"),
                 .product(name: "CrowOpenCode", package: "CrowOpenCode"),
+                .product(name: "CrowTelemetry", package: "CrowTelemetry"),
                 .product(name: "Hummingbird", package: "hummingbird"),
                 .product(name: "HummingbirdWebSocket", package: "hummingbird-websocket"),
                 .product(name: "Crypto", package: "swift-crypto"),
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
@@ -15,10 +15,21 @@ import CrowGit
 import CrowTerminal
 import CrowIPC
 import CrowPersistence
+import CrowTelemetry
 import Foundation
 import Hummingbird
 import HummingbirdWebSocket
 import NIOCore
+
+/// One-slot box for the `TelemetryService`, so the receiver's `onDataReceived`
+/// callback and `SessionService`'s analytics providers — all built before (or
+/// alongside) the service itself — can reach it once it exists. MainActor-isolated
+/// because that is where every reader already runs, which also makes it `Sendable`
+/// without a lock (#772).
+@MainActor
+final class TelemetryHolder {
+    var service: TelemetryService?
+}
 
 /// The headless `crowd` daemon: serves Crow's JSON-RPC domain logic and the
 /// browser terminal over HTTP + WebSocket, reusing `CrowCore`/`CrowIPC`/
@@ -119,6 +130,52 @@ public enum CrowDaemon {
             log("WARNING: tmux not found; /terminal + terminal RPC disabled (set CROW_TMUX to override)")
         }
 
+        // Telemetry (#772). The OTLP receiver that used to run inside the retired
+        // macOS app never moved to the daemon, so `hookState.analytics` was always
+        // empty and the web's per-session analytics strip — which treats absence as
+        // its empty state — never rendered. Built BEFORE SessionService because the
+        // service captures `telemetryPort` (the `OTEL_*` env prefix Claude Code
+        // needs to emit anything at all) and the analytics providers at init.
+        // Enable/port are read once here: changing them needs a daemon restart,
+        // exactly as it did natively.
+        let telemetryConfig = ConfigStore.loadConfig(devRoot: options.devRoot)?.telemetry
+            ?? TelemetryConfig()
+        // The `onDataReceived` callback has to reach the service it is being passed
+        // to, so it resolves through a holder the init fills in afterwards (the app
+        // used `self.telemetryService` for the same reason).
+        let telemetryHolder = TelemetryHolder()
+        let telemetry: TelemetryService?
+        if telemetryConfig.enabled {
+            do {
+                telemetry = try TelemetryService(
+                    port: telemetryConfig.port,
+                    onDataReceived: { sessionID in
+                        // Already @MainActor — the analytics write is serialized
+                        // with every other AppState mutation.
+                        Task { @MainActor in
+                            guard let service = telemetryHolder.service else { return }
+                            let analytics = await service.analytics(for: sessionID)
+                            appState.hookState(for: sessionID).analytics = analytics
+                            // Keep the capture-status line live without re-querying
+                            // the DB; the exact session count refreshes at startup.
+                            appState.telemetryCaptureStatus = TelemetryCaptureStatus(
+                                sessionCount: max(appState.telemetryCaptureStatus?.sessionCount ?? 0, 1),
+                                lastReceivedAt: Date())
+                        }
+                    })
+            } catch {
+                log("WARNING: failed to create the telemetry receiver on port \(telemetryConfig.port) "
+                    + "(\(error)) — per-session analytics will be unavailable")
+                telemetry = nil
+            }
+        } else {
+            telemetry = nil
+        }
+        await MainActor.run { telemetryHolder.service = telemetry }
+        // nil when telemetry is off OR the receiver failed to bind: without a live
+        // receiver the `OTEL_*` exporter vars would point at a closed port.
+        let telemetryPort: UInt16? = telemetry.map(\.port)
+
         // Host the real SessionService so the daemon can spawn Manager (and
         // later review/job) workspaces headlessly with the app down (ADR 0007;
         // CROW-581, M-E2). It drives tmux through the process-global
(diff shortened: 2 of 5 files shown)


dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- OTLP receiver remains loopback-only (`requiredLocalEndpoint` + `acceptLocalOnly`) — no new network exposure.
- `telemetryPort` is withheld when create fails or telemetry is disabled, so agents are not pointed at a closed exporter in those paths.
- Manager session is guarded before `telemetryDeleteProvider` runs; snapshots are deliberately retained for scorecard history.
- Telemetry row purge is gated on successful disk cleanup, so a retryable delete failure does not drop metrics for a still-live session.

**Concerns:**
- No new auth/injection issues in this diff. Pre-existing localhost OTLP ingest (any local process can POST fake session UUIDs) is unchanged and acceptable for the loopback threat model.

### Code Quality
- **Yellow — `CrowDaemon.swift` (~234–249): `telemetry.start()` runs after `telemetryPort` is already captured into `SessionService` / `EngineContext`.** Create-failure correctly sets `telemetry = nil` so exporters are not configured, and the comment at ~175–177 states the port must be nil without a live receiver. But `start()` is where the DB opens and the listener actually begins, and on failure the Task logs + returns while:
  1. `telemetryPort` stays non-nil → Claude launches still get `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:<port>` against nothing listening.
  2. `telemetryHolder.service` stays set → delete/analytics hit an unopened DB (silent `SQLITE_MISUSE` / no-op).

  `start()` itself does not need `SessionService` — only `backfillAnalyticsSnapshots` / `refreshManagerUsage` do. Await `telemetry.start()` **before** computing `telemetryPort` and constructing `SessionService`; on failure clear the service (treat like create-failure) so the port stays nil. Then backfill → capture status → prune after the service exists. That preserves the intended lifecycle without the dead-exporter window.

- **Green — `CrowDaemon.swift` ~161–163:** `sessionCount: max(..., 1)` never reflects newly seen sessions after the one-shot startup `captureStatus()`. Fine for a status line, but consider calling `captureStatus()` (or tracking distinct IDs) on ingest if the count should stay accurate.

- **Green — tests:** New strip + delete-provider coverage is solid (`SessionAnalyticsStripLiveTests`, `SessionServiceTelemetryDeleteTests` — both passed locally). A case asserting failed disk cleanup does *not* invoke `telemetryDeleteProvider` would lock the ordering comment at `SessionService.swift` ~1547–1553.

### Static Analysis
- `SessionServiceTelemetryDeleteTests`: 2/2 passed.
- `SessionAnalyticsStripLiveTests`: 5/5 passed.
- (Local `swift test` exit code 1 is an x86_64/arm64 XCTest bundle loader quirk; Swift Testing suites themselves passed.)

### Summary Table

---

dhilgaertner:
Thanks — pushed 8e6e6ac.

**Yellow (`start()` ordering) — fixed.** You were right that the create/start split was the real boundary: `start()` is where the DB opens and the listener comes up, so "constructed" was never a safe proxy for "usable." Now it goes create → publish holder → `await start()` → derive `telemetryPort`, all before `SessionService` is built. A throw from either step clears the holder and leaves the port nil, collapsing start-failure into the same shape create-failure already had. Publishing the holder *before* `start()` rather than after also closes the (tiny) window where a first datapoint could land on an empty holder. Only the backfill genuinely needed `SessionService`, so that stays deferred — backfill → capture status → prune, order preserved.

**Green 2 (failed-cleanup ordering test) — added.** `failedDiskCleanupKeepsTelemetryRows` chmods a review clone's parent to `0555` so `removeItem` fails with EACCES, then asserts the session survives with `sessionDeletionError` set and the provider was never called. Confirmed it exercises the real branch (the `Failed to remove review clone` log fires during the run), rather than passing vacuously.

**Green 1 (`sessionCount: max(..., 1)`) — leaving as-is, deliberately.** Worth flagging that `appState.telemetryCaptureStatus` currently has **no reader**: its only consumer was `CrowUI/ScorecardView`, which died with the macOS app in ADR-0010. `get-scorecard` ships `telemetryEnabled` + `snapshotCount` instead and never touches it. So the field is written-but-unread today, and calling `captureStatus()` on every ingest would add a SQL `COUNT(DISTINCT …)` per OTLP POST purely to keep a value nobody displays accurate. I kept the cheap approximation for parity with the native behavior; if/when the web grows a capture-status line, that is the right moment to decide between per-ingest recount and tracking distinct IDs in memory. Happy to open a follow-up issue if you would rather it not stay implicit.

Full suites green: CrowDaemon 118, CrowEngine 314, CrowTelemetry 25, plus `swift build --product crowd`.

dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- OTLP remains loopback-only (`requiredLocalEndpoint` + `acceptLocalOnly`) — no new network exposure from wiring the receiver into the daemon.
- `telemetryPort` is derived only after a successful `start()`, and create/start failure clears the holder and leaves the port nil, so agents are not pointed at a dead exporter.
- Telemetry row purge is centralized in `deleteSession`, gated on successful disk cleanup, and skipped for the Manager session; scorecard snapshots are deliberately retained.
- `stop()` before exit/re-exec avoids port/db reuse races on restart.

**Concerns:**
- No new auth/injection issues in this diff. Pre-existing localhost OTLP ingest (any local process can POST) is unchanged and acceptable for the loopback threat model.

### Code Quality
- Prior Yellow (`start()` after `telemetryPort` was baked into `SessionService` / `EngineContext`) is fixed in `8e6e6ac` — create → publish holder → `await start()` → then derive port / build SessionService.
- Delete-ordering coverage added (`failedDiskCleanupKeepsTelemetryRows`) locks the “purge only after cleanup succeeds” contract.
- **Green — `CrowDaemon.swift` ~168–170:** `sessionCount: max(..., 1)` keeps the status line live without a DB round-trip, but will undercount newly seen sessions until the next startup `captureStatus()`. Fine for a status line; refresh via `captureStatus()` (or track distinct IDs) if the count should stay accurate at runtime.

### Static Analysis
- `SessionServiceTelemetryDeleteTests`: 3/3 passed.
- `SessionAnalyticsStripLiveTests`: 5/5 passed.
- (Local `swift test` exit code 1 is the known x86_64/arm64 XCTest bundle loader quirk; Swift Testing suites themselves passed.)

### Summary Table

---

PR #786, opened 2026-07-20
fix(web): terminal touch-scroll on iOS — own the gesture, scroll real history (#777)

### LINKED ISSUE #777: Regression (web parity): iPhone terminal touch-scroll rubber-bands / repeats the same frame instead of scrolling history
## Summary

**Regression (introduced by the web-UI-parity merge).** Opening the web UI terminal on an
iPhone, scrolling up doesn't move through history — it **rubber-bands and re-displays the same
last paragraph over and over**, seemingly forever. Touch scrolling in the terminal is net-new
code that arrived with parity (the native app had no mobile surface), and it fights iOS Safari's
native overscroll instead of scrolling the terminal buffer.

## Confirmed root cause (code + git provenance)

The touch-scroll shim in `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js`
(`enableTouchScroll`, ~L3006):

```js
node.addEventListener('touchmove', (e) => {
  if (lastY == null || e.touches.length !== 1 || !term) return;
  const y = e.touches[0].clientY;
  const cell = (node.clientHeight / (term.rows || 24)) || 18;
  const delta = Math.trunc((lastY - y) / cell);
  if (delta !== 0) { term.scrollLines(delta); lastY = y; }
}, { passive: true });
```

Problems:
- **Registered `{ passive: true }` and never calls `preventDefault()`** — so it cannot stop iOS
  Safari's native touch/overscroll on the fixed xterm canvas. The browser rubber-bands the frame
  while the shim tries to scroll the buffer; the native gesture wins.
- **No alternate-screen handling** — while Claude Code's TUI owns the screen (alt buffer),
  `term.scrollLines(delta)` is a **no-op** (no scrollback in the alt buffer), so the only visible
  motion is Safari's rubber-band → the same frame repeats.
- **Fragile cell metric** — `node.clientHeight / term.rows` is sensitive to a mis-fit on mobile
  (devicePixelRatio, viewport, keyboard) and can make `delta` 0 (no scroll) or erratic.

**Git provenance (parity-merge regression):** `app.js` did not exist on `main` before the parity
merge (`eb7a489^` → absent); `enableTouchScroll` was introduced by the CROW-593 parity workstream
(`c03daee` #590) and reached `main` via `eb7a489` (#594). There was **no mobile/touch terminal
before parity**, so this path was never exercised natively.

## Why the symptom looks like "infinite repeat"

In the alternate screen `scrollLines` does nothing, and the passive handler can't cancel the
native gesture — so a swipe just triggers iOS overscroll/rubber-band on a single rendered frame,
which snaps back and re-renders the same content. It reads as "scroll forever, always the last
paragraph."

## Proposed fix

- Register `touchmove` **non-passive** and `preventDefault()` it so the terminal owns the gesture
  (stop competing with Safari overscroll).
- **Handle the alternate-screen case** so a swipe through a Claude Code session reaches real
  history (tie into whatever the desktop ticket lands for alt-screen scrollback), instead of a
  silent no-op.
- **Fix cell metrics** — derive line height from xterm's actual row metrics after a settled fit;
  guard against 0/NaN.
- Verify on **iOS Safari** specifically (and note Android Chrome behavior).

## Acceptance criteria

- [ ] On iPhone, a one-finger swipe scrolls through the terminal's real history instead of
      rubber-banding the same frame.
- [ ] No infinite/repeating overscroll; the view lands where you scrolled.
- [ ] Scrolling works while a TUI (Claude Code) is active, not just at a shell prompt.
- [ ] No regression to desktop wheel scroll or to text selection.

## References

- Web (new, parity): `app.js` — `enableTouchScroll` (~L3006, `{ passive: true }`, no
  `preventDefault`, no alt-screen branch), `enableWheelScroll` (~L3025, alt-buffer early return),
  `ensureTerminal` (~L2925).
- Provenance: parity merge `eb7a489` (#594) via `c03daee` (#590).
- Related: the desktop scroll ticket (mouse-swallow + alt-screen scrollback) — this mobile fix
  should build on the same alternate-screen decision.

_Related follow-up worth considering: unify the two terminal implementations (`app.js` inline vs
`terminal.html`) so fixes can't drift between them again._

dhilgaertner:
Closes #777

## Problem

The touch-scroll shim that arrived with the web-UI-parity merge (`c03daee` #590 → `eb7a489` #594) registered `touchmove` as `{ passive: true }` and never called `preventDefault()`, so it could not stop iOS Safari's native overscroll. Safari rubber-banded the fixed xterm canvas while the shim tried to scroll the buffer, and the native gesture won.

With a TUI on the alternate screen there was no scrollback to scroll either, so `term.scrollLines` was a silent no-op and the bounce-back was the *only* motion — which is why it read as "the same paragraph, forever."

## Fix

**Own the gesture** — `#terminal` gets `touch-action: none` (and `#terminal-wrap` `overscroll-behavior: contain`) so Safari never claims the pan at all, and `touchmove` is now registered non-passive and `preventDefault()`ed. Pinch-zoom over the terminal is intentionally traded away (change the font size instead); long-press text selection is unaffected.

**Alternate screen forwards to the TUI** instead of no-oping, mirroring `enableWheelScroll`'s existing "a fullscreen TUI owns the wheel" policy. Per line of swipe it emits what a wheel tick would: SGR wheel reports (`\e[<64;1;1M` / `\e[<65;1;1M`) when the app has mouse tracking on — tmux's `mouse on` relays these to Claude Code — and cursor keys otherwise, honouring application-cursor-keys mode. Capped at 24 lines per event so a fling can't flood the PTY. No daemon or tmux changes.

**Fixed cell metrics** — line height now derives from `.xterm-screen` (exactly `rows × cellHeight` under both the WebGL and DOM renderers) with container and 18px fallbacks and NaN/zero guards, and sub-cell drags accumulate rather than being truncated to zero. The old `clientHeight / rows` could stall the scroll entirely on mobile.

Also extracted the one guarded PTY writer as `sendToPTY`, matching `terminal.html`'s helper of the same name.

## Tests

New `Packages/CrowDaemon/web-tests/touch-scroll.test.js` — 20 assertions driving the real `enableTouchScroll` from `app.js` in jsdom against a fake xterm + PTY socket (same harness shape as `board.test.js`): non-passive registration + `preventDefault` (the regression itself), local scrollback with sub-cell accumulation, all four alternate-screen encodings, the fling cap, multi-touch pass-through, and degenerate metrics.

- `npm test` in `web-tests` — 32 board + 20 touch-scroll assertions pass (the script now runs both suites)
- `swift build --package-path Packages/CrowDaemon` — clean
- `swift test --package-path Packages/CrowDaemon` — 116 tests, 32 suites, pass

## Verification checklist

Automated coverage above; the remaining items need a physical device:

- [ ] iOS Safari: swipe at a shell prompt walks through history and lands where released, no bounce-back repeat
- [ ] iOS Safari: swipe scrolls a live Claude Code TUI transcript
- [ ] Long-press selection still works; two-finger gestures don't scroll the buffer
- [ ] Desktop wheel scroll + drag-select unchanged

Android Chrome honours the same contract (`touch-action: none` + non-passive `preventDefault`), so it should behave identically.

## Follow-up

Unrelated to this fix but worth a ticket: `app.js` and `terminal.html` are two independent terminal implementations that have already drifted (`terminal.html` swallows the mouse-tracking DECSET modes, `app.js` does not). Worth unifying.

https://claude.ai/code/session_01WKQpzbFaMPipA4EhevSnpe

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -351,8 +351,12 @@ html, body {
 .tab.add { color: var(--gold); font-size: 16px; font-weight: 700; }
 .tab-close { color: var(--text-muted); font-size: 14px; }
 .tab-close:hover { color: var(--red); }
-#terminal-wrap { flex: 1; min-height: 0; background: #1e1e1e; position: relative; }
-#terminal { width: 100%; height: 100%; }
+#terminal-wrap { flex: 1; min-height: 0; background: #1e1e1e; position: relative; overscroll-behavior: contain; }
+/* #777: `touch-action: none` hands the one-finger gesture to enableTouchScroll
+   instead of iOS Safari's overscroll — without it Safari rubber-bands the fixed
+   xterm canvas and the swipe never reaches the buffer. Costs pinch-zoom over the
+   terminal (change the font size instead); long-press selection is unaffected. */
+#terminal { width: 100%; height: 100%; touch-action: none; }
 
 /* ---- Terminal skeleton loading overlay (#677) ----
    Masks the xterm attach/reload window (blank/garbled grid, reflow flash,
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -2955,9 +2955,7 @@ function ensureTerminal() {
     webglAddon.onContextLoss(() => webglAddon.dispose());
     term.loadAddon(webglAddon);
   } catch (_) { /* WebGL unavailable → canvas/DOM renderer */ }
-  term.onData((data) => {
-    if (termWs && termWs.readyState === WebSocket.OPEN) termWs.send(new TextEncoder().encode(data));
-  });
+  term.onData(sendToPTY);
   // Cmd/Ctrl+C copies the selection (falling through to SIGINT when nothing is
   // selected so Ctrl+C still interrupts); Cmd+V pastes. Lets the browser own
   // copy/paste instead of tmux's copy-mode.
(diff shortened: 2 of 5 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Alternate-screen PTY writes use only hardcoded CSI sequences (SGR wheel 64/65 or cursor keys) — no user-controlled strings reach `sendToPTY` via the touch path.
- `MAX_PTY_SCROLL_LINES = 24` bounds a single coalesced jump so a wild `touchmove` cannot dump an unbounded escape stream into the PTY.
- `sendToPTY` keeps the existing WebSocket `OPEN` guard; empty payloads are skipped.
- Multi-touch is explicitly declined (`lastY = null`, no `preventDefault`), so pinch/zoom ownership stays with the browser where `touch-action` still allows it outside `#terminal`.

**Concerns:**
- None material. This is client-side gesture plumbing over an already-authenticated terminal WebSocket; no new auth, injection, or secret-handling surface.

### Code Quality
- **Green** — `sendScrollToPTY` always emits SGR (`\x1b[<64;1;1M`) at cell `1;1`, regardless of active mouse encoding or finger position. Fine for tmux/`mouse on` and most TUIs; consider mapping touch → cell coords and honouring non-SGR encodings if older apps mis-behave.
- **Green** — Comment claims parity with `enableWheelScroll`, but wheel *returns* on the alt buffer (xterm synthesizes) while touch *injects* via `sendToPTY`. Same policy, different mechanism — worth a one-line clarification so future edits don’t “unify” them incorrectly.
- **Green** — `touch-action: none` + non-passive `preventDefault` is the right fix for iOS rubber-banding; pinch-zoom over the terminal is an intentional trade. Device checklist items (esp. long-press selection) are still unchecked — worth confirming on hardware before relying on that claim alone.
- Solid regression suite (`touch-scroll.test.js`, 20 assertions): non-passive registration, sub-cell accumulation, all four alt-screen encodings, fling cap, multi-touch pass-through, degenerate metrics. `npm test` → 32 board + 20 touch-scroll pass; `swift build --package-path Packages/CrowDaemon` clean.

### Summary Table

---

PR #789, opened 2026-07-20
fix(web): restore the mouse-mode swallow in the app terminal (#776)

### LINKED ISSUE #776: Regression (web parity): terminal scroll hits a wall + snaps to bottom on mouse-move (missing mouse-mode swallow in app.js)
## Summary

**Regression (introduced by the web-UI-parity merge).** In the web UI's Claude Code session
terminal, scrolling up **hits a wall** well before the history should end, and while parked at
that wall, **moving the mouse snaps the viewport back to the bottom**. This worked 100% in the
native macOS app before parity. Root cause is a concrete divergence: the new web terminal was
built without the **mouse-mode swallow** that the native terminal renderer has.

## Confirmed root cause (code + git provenance)

The web UI renders its own inline xterm.js in
`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` (`ensureTerminal`, ~L2925). The
native macOS terminal renders `Packages/CrowTerminal/Sources/CrowTerminal/Resources/xterm/terminal.html`.

- **`terminal.html` swallows mouse-tracking DECSET/DECRST** (`?1000/1001/1002/1003/1005/1006/1015/1016`)
  at the parser so the TUI never puts xterm into mouse-reporting mode — client-side selection,
  context menu, and wheel-scroll work instead (CROW-581, `terminal.html:80–91`).
- **`app.js` `ensureTerminal` never got that swallow.** Confirmed: `git log -S MOUSE_MODES -- app.js`
  and `-S swallowMouseMode -- app.js` → *never present*.
- The web wheel handler additionally **no-ops on the alternate-screen buffer**:
  `enableWheelScroll` returns early when `term.buffer.active.type === 'alternate'`
  (`app.js:3029`).

**Git provenance (this is a parity-merge regression, not pre-existing):**
- `app.js` did not exist on `main` before the parity merge (`git cat-file -e eb7a489^:…/app.js`
  → absent).
- `ensureTerminal` / `enableWheelScroll` / `enableTouchScroll` were introduced by the CROW-593
  parity workstream (`c03daee` #590, `7d8e6fb` CROW-593), landed on `main` via `eb7a489` (#594).
- `terminal.html` (with the swallow) predates all of this (`a364f8b`, #548) — it was the native
  app's renderer, which is why the native experience was correct.

## Why the two symptoms happen (mechanism — inferred, pending browser repro)

Claude Code's TUI runs in the **alternate-screen buffer** (no scrollback) and enables mouse
tracking:

- **The "wall":** in the alternate buffer, `enableWheelScroll` returns early → the wheel does
  nothing; the replayed history lives in the *normal* buffer, out of reach while the TUI owns
  the screen.
- **Snap-to-bottom on mouse-move:** because the swallow is missing, Claude Code's mouse tracking
  stays active → mouse-motion → SGR mouse reports to the PTY → TUI repaints → xterm follows the
  new output to the bottom. Secondary contributor: the parity-new fit/ownership churn
  (`window.focus`/`visibilitychange` → `takeTerminalOwnership`, #667; `ResizeObserver` →
  `fitTerminal`, #661) → tmux `SIGWINCH` → full repaint → bottom.

**To confirm at repro time:** check `term.buffer.active.type` while scrolled up (expect
`alternate`) and watch the `/terminal` WebSocket for `\x1b[<…` mouse reports on mouse-move.

## Proposed fix

- **Port the mouse-mode swallow into `app.js` `ensureTerminal`** — register the same
  `registerCsiHandler({prefix:'?', final:'h'/'l'})` swallow as `terminal.html:80–91`. This is
  the "restore parity with the native renderer" fix and should stop the mouse-move snap.
- **Decide alternate-screen scroll behavior** so "scroll up through a Claude Code session" works
  (e.g., surface the normal-buffer history / tmux copy-mode, rather than silently no-oping).
- Audit the fit/ownership repaint triggers (#661/#667) for their contribution to losing scroll
  position, and avoid yanking to bottom on a pure resize when the user is scrolled up.

## Acceptance criteria

- [ ] Moving the mouse while scrolled up no longer snaps the terminal to the bottom.
- [ ] Scrolling up reaches the session's real history (not a premature wall), including while a
      TUI (Claude Code) is active.
- [ ] Behavior matches the native `terminal.html` renderer (mouse suppression on).
- [ ] Inner apps that legitimately want the mouse degrade the same acceptable way `terminal.html`
      already documents.

## References

- Web (new, parity): `app.js` — `ensureTerminal` (~L2925), `enableWheelScroll` (~L3025, alt-buffer
  early return L3029), fit/ownership (`takeTerminalOwnership`, `fitTerminal`, ~L2983–2997).
- Native renderer (has the fix): `Packages/CrowTerminal/.../terminal.html:80–91` (mouse swallow).
- Daemon replay: `Packages/CrowDaemon/Sources/CrowDaemon/TerminalCockpit.swift:92–122`
  (`replayLines = 50000`, `capture-pane` replay).
- Provenance: parity merge `eb7a489` (#594) via `c03daee` (#590) + `7d8e6fb` (CROW-593).

_Related follow-up worth considering: unify the two terminal implementations (`app.js` inline vs
`terminal.html`) so fixes can't drift between them again._

dhilgaertner:
Closes #776

## Problem

In the web UI / Tauri desktop shell, scrolling up in a Claude Code terminal and then **moving the mouse snaps the viewport back to the bottom**. This worked in the native macOS terminal before the web-UI-parity merge (`eb7a489`, #594).

## Root cause

`Resources/web/terminal.html:80-91` has swallowed the mouse-tracking DECSET/DECRST toggles (`?1000/1001/1002/1003/1005/1006/1015/1016`) at the xterm parser since CROW-581. The app's inline terminal (`app.js` → `ensureTerminal`), added by the CROW-593 parity workstream, **never got that swallow** (`git log -S MOUSE_MODES -- app.js` → never present).

Confirmed against the live tmux server:

```
$ tmux -S $TMPDIR/crow-tmux.sock list-panes -a -F '… mouse_any=#{mouse_any_flag} alt=#{alternate_on}'
4 Claude Code mouse_any=1 alt=1 …
5 Claude Code mouse_any=1 alt=0 …
…
```

Every Claude Code pane reports `mouse_any_flag=1` — the agent TUI really does enable mouse tracking — and `crow-tmux.conf`'s `mouse off` forwards those DECSETs straight through to the client. So xterm.js entered mouse-reporting mode and emitted an SGR report on every mouse **move** → the TUI repainted → the scrolled-up viewport followed the new output to the bottom. Drag-select and the browser context menu were eaten for the same reason.

Two corrections to the issue's write-up, worth recording:

1. The renderer carrying the swallow is the **daemon's** `Resources/web/terminal.html`, not `CrowTerminal/.../xterm/terminal.html` (the native one has no swallow).
2. The alternate-buffer early return in `enableWheelScroll` is **not** the main cause of the "wall": `alternate_on=0` for nearly every pane (`alternate-screen off` + `terminal-overrides ,xterm*:smcup@:rmcup@`), so the client's xterm is normally in the main buffer with its 50k scrollback intact. The wall is dominated by the snap-to-bottom.

## Changes

- **Port the `MOUSE_MODES` swallow into `ensureTerminal`**, registered before `open()`/the first write. Same handler as `terminal.html`.
- **Stop `enableWheelScroll` bailing on the alternate-screen buffer.** With the mouse swallowed no inner app can receive the wheel, so bailing just handed it to xterm's alternate-scroll fallback, which emits **arrow keys** that Claude Code reads as input-history navigation — the exact failure `crow-tmux.conf` documents under `alternate-screen off`. The handler now owns the event in both buffers; an alternate buffer has no scrollback, so there the wheel is swallowed rather than scrolled.
- **`WebTerminalAssetTests`** — a drift guard pinning the swallow (all eight modes, both the `h` and `l` CSI handlers) into `app.js` *and* `terminal.html`, plus a check that the wheel handler no longer falls through. This is the guard that would have caught the original divergence.

## Testing

- `swift test --package-path Packages/CrowDaemon` → 118 tests / 33 suites pass, including the new guard.
- `node --check` on `app.js` clean.
- Interactive browser check not run from this environment (Chrome here can't reach loopback — `ERR_CONNECTION_REFUSED` where `curl` gets 200). To verify by hand in a session terminal: `term._core.coreMouseService.areMouseEventsActive` → `false`, `term.buffer.active.type` → `normal`, then scroll up and move the mouse — the viewport must stay put, with no `\x1b[<…` reports on the `/terminal` WebSocket.

## Follow-up

Unifying the two xterm setups (`app.js` inline vs `web/terminal.html`) remains the issue's suggested follow-up; the drift guard is the interim protection.

https://claude.ai/code/session_01Aqok8ShzCHzkBZERsUaVBj

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -2941,6 +2941,32 @@ function ensureTerminal() {
   term.loadAddon(imageAddon);
   term.loadAddon(searchAddon);
   term.loadAddon(webLinksAddon);
+
+  // Mouse-mode swallow — same handler as web/terminal.html (CROW-581). The
+  // agent TUIs (Claude Code, Cursor) turn the xterm mouse protocol on
+  // themselves, and crow-tmux.conf's `mouse off` means tmux forwards those
+  // DECSETs straight through to this client. Left alone, xterm.js enters
+  // mouse-reporting mode and every mouse MOVE emits an SGR report to the PTY →
+  // the TUI repaints → the viewport is yanked to the bottom while the user is
+  // scrolled up; drag-select and the browser context menu are eaten too.
+  // Dropping the mode toggles at the parser keeps selection, the context menu,
+  // and wheel-scroll client-side. Inner apps that genuinely want the mouse
+  // (vim/htop) lose it here — the same acceptable trade terminal.html makes.
+  // #776: this surface was built without the swallow, so it regressed against
+  // the standalone renderer. Must be registered before open()/first write.
+  const MOUSE_MODES = new Set([1000, 1001, 1002, 1003, 1005, 1006, 1015, 1016]);
+  function swallowMouseMode(params) {
+    const arr = params && params.params ? params.params : params;
+    const len = arr ? arr.length : 0;
+    for (let i = 0; i < len; i++) {
+      const v = arr[i];
+      if (MOUSE_MODES.has(Array.isArray(v) ? v[0] : v)) return true; // handled → drop it
+    }
+    return false; // not a mouse mode → let xterm apply it normally
+  }
+  term.parser.registerCsiHandler({ prefix: '?', final: 'h' }, swallowMouseMode);
+  term.parser.registerCsiHandler({ prefix: '?', final: 'l' }, swallowMouseMode);
+
   // #677: seed the skeleton before first open() so it covers the initial xterm
   // layout / first-fit reflow flash; connectTerminalWs() re-arms it below.
   showTerminalSkeleton();
@@ -3020,14 +3046,22 @@ function enableTouchScroll(node) {
 
 // Let xterm.js own wheel scrolling in the browser: scroll the local 50k-line
 // scrollback instead of forwarding the wheel to tmux (whose server-global
-// `mouse on` would otherwise drop into copy-mode — janky in a browser). A
-// fullscreen/alternate-screen app (vim, htop) still gets the wheel.
+// `mouse on` would otherwise drop into copy-mode — janky in a browser).
+//
+// #776: this used to bail on the alternate-screen buffer to "let TUIs handle
+// the wheel", but the mouse-mode swallow in ensureTerminal means no inner app
+// can receive mouse events here — so bailing just handed the wheel to xterm's
+// alternate-scroll fallback, which emits ARROW KEYS. Claude Code reads those as
+// input-history navigation: exactly the bug crow-tmux.conf documents under
+// `alternate-screen off`. Keep owning the event in both buffers; an alternate
+// buffer has no scrollback, so there we swallow the wheel rather than scroll.
+// (crow-tmux.conf also strips the client's smcup/rmcup, so in practice the
+// terminal stays in the main buffer and the scrollback path is the live one.)
 function enableWheelScroll(node) {
   node.addEventListener('wheel', (e) => {
     if (!term) return;
     const buf = term.buffer && term.buffer.active;
-    if (buf && buf.type === 'alternate') return; // let TUIs handle the wheel
-    term.scrollLines(e.deltaY > 0 ? 3 : -3);
+    if (!buf || buf.type !== 'alternate') term.scrollLines(e.deltaY > 0 ? 3 : -3);
     e.preventDefault();
     e.stopPropagation();
   }, { capture: true, passive: false });
--- a/Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift
+++ b/Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift
@@ -33,6 +33,24 @@ import Testing
         return try String(contentsOf: url, encoding: .utf8)
     }
 
+    /// The single line declaring the swallowed-mode set. Anchoring the per-mode
+    /// assertions to it keeps them meaningful: a bare `contains("1000")` over
+    /// all of `app.js` would match any unrelated timeout literal (review).
+    private static func mouseModesLine(_ source: String) throws -> Substring {
+        try #require(
+            source.split(separator: "\n").first { $0.contains("const MOUSE_MODES") },
+            "no `const MOUSE_MODES` declaration")
+    }
+
+    /// The body of a top-level `function <name>(` … `\n}` block, for asserting on
+    /// one handler rather than the whole file.
+    private static func functionBody(_ name: String, in source: String) throws -> Substring {
+        let start = try #require(source.range(of: "function \(name)("), "no \(name)")
+        let rest = source[start.upperBound...]
+        let end = try #require(rest.range(of: "\n}\n"), "unterminated \(name)")
+        return rest[..<end.lowerBound]
+    }
+
     /// Both surfaces must drop the mouse-tracking DECSET/DECRST toggles
     /// (`?1000/1001/1002/1003/1005/1006/1015/1016`) at the parser, for `h` (set)
     /// and `l` (reset) alike — a swallow that covers only one of the two leaves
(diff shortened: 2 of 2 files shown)


dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Ports the existing CROW-581 mouse-mode swallow (`?1000/1001/1002/1003/1005/1006/1015/1016` for both CSI `h`/`l`) into `ensureTerminal` before `open()`/first write — correct placement, matches `terminal.html`.
- Stopping mouse-reporting mode reduces unintended PTY input (SGR mouse reports on move), which is a net security/UX win for this surface.
- No new auth, secrets, injection surface, or privileged paths.

**Concerns:**
- None material. Swallowing mouse modes still means vim/htop lose mouse in the browser — same documented tradeoff as `terminal.html`.

### Code Quality
- **Yellow** — `WebTerminalAssetTests.wheelHandlerNeverFallsThroughOnAlternateBuffer` (`WebTerminalAssetTests.swift:62-65`) only asserts the absence of the exact string `return; // let TUIs handle the wheel`. Reintroducing an alternate-buffer early return with a different comment would pass while restoring the arrow-key / history-nav bug. Assert the positive ownership pattern instead, e.g. that `enableWheelScroll` contains `buf.type !== 'alternate'` (and still always `preventDefault`s), rather than a comment fingerprint.
- **Green** — Comment typo in `app.js:2952`: "Droping" → "Dropping".
- **Green** — Per-mode `source.contains("1000")` is weak for `app.js` (many unrelated `1000` timeouts). The `MOUSE_MODES` / `registerCsiHandler` checks already carry the real signal; consider anchoring mode checks to the `MOUSE_MODES` Set line if you tighten the suite.
- Logic of the wheel change itself looks correct: own the event in both buffers; scroll only when not alternate; always `preventDefault`/`stopPropagation` so xterm's alternate-scroll arrow-key fallback cannot fire.
- `node --check` on `app.js` clean; `WebTerminalAssetTests` passed locally (Swift Testing). Full `swift test` package load hit an unrelated arm64/x86_64 xctest bundle arch mismatch in this environment; the filtered suite still ran green.

### Summary Table

---

dhilgaertner:
Thanks — both test findings addressed in 1e05599.

**Yellow — comment-fingerprint assertion (`WebTerminalAssetTests.swift:62-65`).** Agreed, that was a weak guard. Replaced with a positive assertion over `enableWheelScroll`'s extracted body: the alternate check may only gate `scrollLines` (`buf.type !== 'alternate'`), both `preventDefault()` and `stopPropagation()` are always called, and the sole `return` is the `if (!term)` guard. Verified the way you'd want: I reintroduced the bail with fresh wording —

```js
if (buf && buf.type === 'alternate') return; // fullscreen apps own it
```

— and the test fails on two expectations (the alternate-check shape, and `returns → 2) == 1`), then reverted. So a reworded early return no longer slips through.

**Green — per-mode `contains("1000")`.** Agreed, and taken. The mode assertions are now anchored to the `const MOUSE_MODES` line rather than the whole file, so unrelated timeout literals can't satisfy them.

**Green — "Droping" typo at `app.js:2952`.** Not changing this one: the line already reads "Dropping" with two p's on the branch (`rg -i droping` over the tree returns nothing). I think this was a misread of the rendered diff — happy to be corrected if you're seeing it somewhere I'm not.

Full `swift test --package-path Packages/CrowDaemon` → 118 tests / 33 suites green locally.

dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Swallowing DECSET/DECRST mouse-tracking modes (`?1000`–`?1016`) at the xterm parser keeps SGR mouse reports off the PTY WebSocket — fewer unsolicited input sequences from move/drag events.
- No new auth, network, or credential surfaces; change is confined to client-side terminal input handling.
- Handler is registered before `open()` / first write, so a mode set on connect cannot race past the swallow.

**Concerns:**
- None. Intentional trade-off (vim/htop lose mouse in-browser) matches the existing `terminal.html` policy from CROW-581.

### Code Quality
- Root cause and fix are sound: port of the proven `MOUSE_MODES` / `swallowMouseMode` handler into `ensureTerminal`, plus correcting `enableWheelScroll` so the alternate-buffer path no longer falls through to xterm alternate-scroll (arrow keys → TUI history nav).
- `WebTerminalAssetTests` is a strong drift guard — pins all eight modes, both `h`/`l` CSI registrations on both assets, and the wheel ownership shape (scroll gated, event always consumed, single early return).
- New suite passed locally (`WebTerminalAssetTests`); `node --check` on `app.js` is clean. Desktop/Shell Lint CI green; Build & Test still pending at review time.

**Green (consider):**
- `functionBody` ends on the first `\n}\n` — fine for today’s flat `enableWheelScroll`, but brace-aware parsing would survive nested formatting changes.
- Duplicated swallow between `app.js` and `terminal.html` remains; the PR correctly defers unification and relies on the drift guard in the meantime.

### Summary Table

---

PR #796, opened 2026-07-21
fix(web): default Tickets pipeline tab to Backlog, not All

### LINKED ISSUE #795: Tickets view: default the pipeline tab to Backlog, not All
When opening the **Tickets** view, the default pipeline tab should be **Backlog** instead of **All**.

## Where
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js:595` — `const PIPELINE = ['All', 'Backlog', 'Ready', 'In Progress', 'In Review', 'Done'];`. The view currently initializes its selected pipeline filter to the first entry (`'All'`). The tab row is rendered around `app.js:846`.

## Change
- Set the **initial selected-tab state** to `'Backlog'`. Do **not** reorder the `PIPELINE` array or remove `All` — `All` stays available as a tab, it's just no longer the default on load.
- Scope the default so it applies whenever the Tickets view mounts (initial page load / navigating back to Tickets). If a user's last-selected tab is already persisted anywhere, preserve that behavior and only change the cold-start default.

## Verify
- Open the Tickets view from a fresh load → the **Backlog** tab is pre-selected and its filtered list is shown.
- Clicking **All** (and the other pipeline tabs) still works and filters correctly.

dgershman:
Closes #795. Sets the Tickets view's initial selected pipeline tab to Backlog; PIPELINE array unchanged and All remains selectable. Preserves any persisted last-selected tab — only the cold-start default changes (`ticketFilter` is session-only in-memory state, so nothing is persisted).

## Verify
- Fresh load → **Backlog** tab pre-selected, showing its filtered list.
- Clicking **All** and the other pipeline tabs still filters correctly.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -564,7 +564,7 @@ function persistSidebarCache() {
     sidebarCacheHit = true;
   } catch (_) { /* quota / private mode */ }
 }
-let ticketFilter = 'All'; // pipeline segment ('All' or a status rawValue); default to All so the board isn't misread as empty when work moves to Done
+let ticketFilter = 'Backlog'; // pipeline segment ('All' or a status rawValue); default to Backlog so the Tickets view opens on the intake queue (CROW-795). 'All' stays a selectable tab.
 let allowlistHideGlobal = false;
 let allowlistFilter = ''; // #701: case-insensitive substring filter on entry pattern
 let ticketSearch = ''; // #714: case-insensitive substring on ticket text, composed with ticketFilter


dhilgaertner (approved):
## Code & Security Review

One-line default change (`app.js:567`), `All` → `Backlog`. Verified against the payload contract rather than the diff alone.

### Correctness
- `'Backlog'` is exactly `TicketStatus.backlog.rawValue` (`Packages/CrowCore/Sources/CrowCore/Models/Enums.swift`) and a member of `PIPELINE` (`app.js:595`), so the tab renders with `.active` at `app.js:2724` and the equality filter at `app.js:2668` matches real `project_status` values. No silent "no tab selected / empty list" state.
- **The concern the replaced comment encoded is genuinely handled.** The old default existed "so the board isn't misread as empty when work moves to Done" — the risk being tickets with no project status vanishing behind a non-`All` default. `Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift:610` maps `issue.projectStatus == .unknown → .backlog` before serializing, so every issue always carries a real pipeline status and un-triaged tickets land *in* Backlog. Repos with no project board still show a populated default tab.
- Scope matches #795 exactly: `PIPELINE` is not reordered, `All` remains selectable, and `ticketFilter` is in-memory session state with no persistence and no other writer (only `app.js:2734` sets it), so only the cold-start default moves.
- Only one `app.js` exists in the repo — no bundled duplicate to drift out of sync.

### Security Review
**Strengths:**
- No auth, network, storage, or DOM-injection surface touched. The value is a compared-against-allowlist string, never interpolated into HTML — `app.js:2841` builds the status pill via `el()`/`textContent`, not `innerHTML`.
- No change to what data is fetched or who may see it; the filter is purely presentational over an already-authorized payload.

**Concerns:**
- None.

### Code Quality
- Comment updated alongside the value and cites CROW-795 — matches the file's existing convention of ticket-referenced inline rationale.
- `Packages/CrowDaemon/web-tests/board.test.js` assigns `T.ticketFilter` explicitly in every case (`:85`, `:130`, `:134`), so no test depended on the old default and none needed updating. **Note:** I could not execute the suite locally — `jsdom` is not installed in this review worktree and the sandbox blocks the install — so this is read-verified, not run-verified. Reading the assignments makes a regression here implausible for a constant this shape.

### Non-blocking observations (Green)
- No test pins the cold-start default, so a future refactor could flip it back unnoticed. A one-line assertion on the initial `ticketFilter` would lock in the #795 intent cheaply.
- Native and web now differ on this default: `AppState.selectedTicketStatus` initializes to `.inProgress`. #795 scopes itself to `app.js`, so this is correctly out of scope here — flagging only in case cross-surface parity is wanted later.

### Summary Table

---

PR #798, opened 2026-07-21
fix(web): spin an inner glyph in the Tickets refresh button, not the whole button

### LINKED ISSUE #797: Tickets refresh button spins the whole button instead of a spinner inside it
Clicking the **refresh** button on the Tickets summary card rotates the **entire button** while the refresh is in flight, instead of showing a spinner *inside* an otherwise-stationary button. It looks like the button itself is tumbling.

## Where
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css:640-653` — `.tickets-refresh` and its in-flight state `.tickets-refresh.spinning { animation: tickets-refresh-spin 0.7s linear infinite; }` (CROW-771). The animation is applied to the button element itself, so the whole control rotates. The CSS comment at `:646` already notes it couldn't reuse `wizard-spin`.
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js:686` `ticketsRefreshing()` toggles the `.spinning` class; the Tickets card is rendered around `app.js:825`.

## Expected
- The button chrome (border, padding, hit area) stays put; only a **spinner/glyph inside** the button animates during refresh. Match the in-flight pattern already used elsewhere (e.g. the board-title refresh spinner noted at `app.css:483`, and the detail-panel action-button spinner CROW-749).

## Fix direction
- Move the rotation off the button and onto an inner element: give `.tickets-refresh` an icon/`<span>` child and animate *that* (`.tickets-refresh.spinning .icon { animation: … }`), or swap the icon for a dedicated spinner node while `spinning`. Keep the 0.7s linear infinite timing.

## Verify
- Click refresh on the Tickets card → the button stays fixed in place; a spinner turns inside it; when the fetch completes the spinner stops and the normal icon returns. No layout shift.

dgershman:
Closes #797. Moves the in-flight rotation off `.tickets-refresh` onto an inner spinner node: while refreshing, the `↻` glyph is swapped for the shared `.action-spinner` ring (matching the board-title / CROW-749 spinners), keeping 0.7s linear infinite timing via the existing `wizard-spin` keyframe. The bespoke `tickets-refresh-spin` keyframe is deleted. The button is now flex-centered and fixed-size, so its chrome stays put with no layout shift.

## Verify
Click refresh on the Tickets card → the button stays fixed in place; a spinner turns inside it; when the fetch completes the spinner stops and the `↻` returns. No layout shift.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -641,21 +641,15 @@ html, body {
 .tickets-title { color: var(--gold); font-weight: 700; font-size: 13px; }
 .tickets-refresh {
   position: absolute; right: 0; top: 50%; transform: translateY(-50%);
+  display: inline-flex; align-items: center; justify-content: center;
   width: 22px; height: 22px; border-radius: 6px; cursor: pointer; line-height: 1;
   border: 1px solid var(--border-subtle); background: var(--bg-card); color: var(--text-secondary); font-size: 12px;
 }
 .tickets-refresh:hover { color: var(--text-primary); border-color: var(--border-strong); }
-/* In-flight ticket refresh (CROW-771). Can't reuse `wizard-spin`: the button is
-   absolutely positioned with `translateY(-50%)`, which a bare rotate keyframe
-   would clobber — so the keyframe composes both transforms. */
-.tickets-refresh.spinning {
-  opacity: 0.6; cursor: default;
-  animation: tickets-refresh-spin 0.7s linear infinite;
-}
-@keyframes tickets-refresh-spin {
-  from { transform: translateY(-50%) rotate(0deg); }
-  to   { transform: translateY(-50%) rotate(360deg); }
-}
+/* In-flight ticket refresh (CROW-771): the button chrome stays put while the
+   `.action-spinner` glyph swapped in for the `↻` turns inside it (CROW-797) —
+   matching the detail-header action-button spinner (CROW-749). */
+.tickets-refresh.spinning { opacity: 0.6; cursor: default; }
 .tickets-row { display: flex; align-items: stretch; gap: 6px; margin: 2px 2px 8px; }
 .tickets-row .tickets-card { flex: 1 1 auto; margin: 0; }
 .sidebar-tools { display: flex; flex-direction: column; justify-content: center; gap: 4px; flex: 0 0 auto; }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -642,7 +642,7 @@ html, body {
 .tickets-refresh {
   position: absolute; right: 0; top: 50%; transform: translateY(-50%);
   display: inline-flex; align-items: center; justify-content: center;
-  width: 22px; height: 22px; border-radius: 6px; cursor: pointer; line-height: 1;
+  width: 22px; height: 22px; padding: 0; border-radius: 6px; cursor: pointer; line-height: 1;
   border: 1px solid var(--border-subtle); background: var(--bg-card); color: var(--text-secondary); font-size: 12px;
 }
 .tickets-refresh:hover { color: var(--text-primary); border-color: var(--border-strong); }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -830,7 +830,10 @@ function ticketsCard() {
   const head = el('div', 'tickets-head');
   head.appendChild(el('span', 'tickets-title', 'Tickets'));
   const busy = ticketsRefreshing();
-  const refresh = el('button', 'tickets-refresh' + (busy ? ' spinning' : ''), '↻');
+  // While refreshing, swap the ↻ glyph for the shared `.action-spinner` ring so
+  // the spinner turns inside a stationary button, not the button itself (CROW-797).
+  const refresh = el('button', 'tickets-refresh' + (busy ? ' spinning' : ''), busy ? '' : '↻');
+  if (busy) refresh.appendChild(el('span', 'action-spinner'));
   refresh.title = busy ? 'Refreshing tickets…' : 'Refresh tickets';
   refresh.disabled = busy;
   refresh.onclick = (e) => { e.stopPropagation(); refreshTickets(); };
--- a/Packages/CrowDaemon/web-tests/board.test.js
+++ b/Packages/CrowDaemon/web-tests/board.test.js
@@ -165,14 +165,18 @@ check('spinner shown without the daemon flag', q('.board-title .action-spinner')
 check('Refresh button disabled', [...q('.action-btn')].find((b) => b.textContent === 'Refresh').disabled === true);
 
 console.log('\nRefresh spinner — sidebar Tickets card ↻:');
-check('↻ spins + disabled while refreshing', (() => {
+check('↻ swaps for inner .action-spinner + disabled while refreshing', (() => {
   const b = T.ticketsCard().querySelector('.tickets-refresh');
-  return b.className.includes('spinning') && b.disabled === true && /Refreshing/.test(b.title);
+  return b.className.includes('spinning') && b.disabled === true && /Refreshing/.test(b.title)
+    // The spinner turns on an inner node, not the button itself (CROW-797): the ↻
+    // glyph is gone and the shared `.action-spinner` ring is present.
+    && b.querySelector('.action-spinner') !== null && b.textContent === '';
 })());
 T.ticketRefreshPending = false;
-check('↻ idle when not refreshing', (() => {
+check('↻ glyph restored, no spinner node when idle', (() => {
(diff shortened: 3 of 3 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Small, well-scoped CSS/JS change: the in-flight rotation moves off the absolutely-positioned `.tickets-refresh` button onto an inner `.action-spinner` node, and the bespoke `tickets-refresh-spin` keyframe (which only existed to compose `translateY(-50%)` with the rotation) is deleted. The approach matches the three existing `.action-spinner` call sites (`app.js:1500`, `:2687`, `:3026`), so the UI now has one spinner idiom instead of two.

### Critical Issues
None.

### Security Review
**Strengths:**
- No network, auth, storage, or user-input surface touched — pure presentational change.
- No `innerHTML` introduced; the swap goes through the `el()` helper (`app.js:667`) which uses `textContent` + `createElement`, so no XSS vector. (Note the neighbouring `markInReviewAction` does use `innerHTML` save/restore — untouched here, and on a static string.)

**Concerns:** None.

### Code Quality

**Green — verified correct:**
- Sizing fits: `* { box-sizing: border-box }` (`app.css:24`) means the 12×12 `.action-spinner` sits inside the 22×22 button's 20px content box with room to spare; `display: inline-flex` + centering keeps both the glyph and the ring centred.
- `.board-title .action-spinner { margin-left: 10px }` (`app.css:486`) is scoped to `.board-title`, so it does not leak an offset into the button — the ring stays centred.
- The button keeps `transform: translateY(-50%)`, and the spinner's own `wizard-spin` rotate now applies to a separate element, so there is no transform clobbering — the exact bug the deleted comment described.
- `busy ? '' : '↻'` is safe: `''` is non-null, so `el()` sets `textContent = ''` rather than skipping it.
- Accessible name is preserved while the glyph is gone: `refresh.title` is set to "Refreshing tickets…" on the same render pass (`app.js:837`).
- Existing suite passes: `npm test` in `Packages/CrowDaemon/web-tests` → 50 passed / 0 failed (board), 34 passed / 0 failed (row), touch-scroll clean.

**Yellow — should fix (test coverage):**
- `Packages/CrowDaemon/web-tests/board.test.js:167-176` — the two "Refresh spinner — sidebar Tickets card ↻" assertions only check `className.includes('spinning')`, `disabled`, and `title`. All three were already true before this PR, so the suite passes identically on `main` and on this branch: **the behaviour this PR exists to change is unasserted**, and a future refactor could revert the inner-node swap without turning the suite red. The repo already maintains a jsdom harness for exactly this button, so extending it is cheap — e.g. inside the busy case add `b.querySelector('.action-spinner') !== null && b.textContent === ''`, and in the idle case `b.querySelector('.action-spinner') === null && b.textContent === '↻'`.

**Green — nit:**
- The PR description says the button "is now flex-centered and fixed-size" — it was already `width: 22px; height: 22px` before this change (`app.css:644`). Only the flex-centering is new. Description-only, no code impact.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 3 Green] findings. The implementation itself looks right; adding the two assertions that pin the inner-spinner behaviour is all that stands between this and an approve.

---

dhilgaertner (changes requested):
## Code & Security Review

Nice cleanup — deleting the bespoke `tickets-refresh-spin` keyframe in favour of the shared `.action-spinner` / `wizard-spin` pair is the right call, and it removes the `translateY(-50%)`-clobbering hazard the old comment had to warn about. One measured geometry bug keeps this from being an approve.

### Security Review
**Strengths:**
- No new input handling, network calls, or auth surface. The change is presentational.
- DOM is built with `el()` → `textContent` + `appendChild`, never `innerHTML` with interpolated data, so no XSS vector is introduced (`app.js:835-836`).
- `refresh.onclick` keeps `e.stopPropagation()` so the button can't accidentally trigger the card's `selectBoard('tickets')` navigation.

**Concerns:** none.

### Code Quality

**🟡 Yellow — the spinner renders as an 8×12 ellipse, not a circle** (`app.css:642-647`)

`.tickets-refresh` never zeroes the UA button padding, and `* { box-sizing: border-box }` (`app.css:24`) is in force. So the flex content box is `22 − 2px border − 12px padding = 8px` wide, and `.action-spinner` (a flex item, `flex-shrink: 1` by default, automatic minimum size ≈ its 4px borders) is squeezed to 8px wide while keeping its 12px height. `border-radius: 50%` on 8×12 is an oval, and it visibly wobbles as `wizard-spin` rotates it.

Measured in headless Chrome against a minimal repro using the exact rules from this diff:

```
buttonPadding: ["1px", "6px", "1px", "6px"]   // UA default, never reset
buttonRect:    22 × 22
spinnerRect:    8 × 12                        // <- squashed
```

Adding `padding: 0` to `.tickets-refresh` restores it (re-measured: `spinnerRect: 12 × 12`). Belt-and-braces, `flex: 0 0 auto` on the spinner inside this button would also stop the shrink:

```css
.tickets-refresh {
  position: absolute; right: 0; top: 50%; transform: translateY(-50%);
  display: inline-flex; align-items: center; justify-content: center;
  width: 22px; height: 22px; padding: 0; border-radius: 6px; cursor: pointer; line-height: 1;
  ...
}
```

This is why the CROW-749 precedent works and this one doesn't: `.action-btn` is auto-width, so its spinner is never compressed. Worth re-checking the idle `↻` too — it's been overflowing that same 8px content box all along; `padding: 0` centers it properly as a bonus.

**🟢 Green — test can't see this class of bug** (`web-tests/board.test.js:168-180`)

The new assertions are the right ones for jsdom (glyph swapped out, `.action-spinner` node present/absent, `disabled`, `title`) and both pass locally — full suite: **34 passed, 0 failed**. But jsdom has no layout engine, so no assertion here could have caught the ellipse. Not a request to add layout testing; just noting that the CSS side of this change is only verifiable by eye or by a real engine, so please confirm visually after the padding fix.

**🟢 Green — nit** (`app.js:835`)

`busy ? '' : '↻'` works (`el()` guards on `!= null`, so `''` sets an empty `textContent`), and the test asserts `textContent === ''`, so this is fine as written. `null` would express "no text" slightly more directly if you touch the line anyway.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The approach is right; it just needs `padding: 0` on `.tickets-refresh` so the ring stays round.

---

dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No user-controlled data flows into the DOM: the spinner node is built with `el('span', 'action-spinner')` (`document.createElement` + static class), so there is no XSS surface added.
- The `innerHTML` restore pattern used by the CROW-749 sibling (`markInReviewAction`, app.js:1497-1499) is *not* copied here — `ticketsCard()` re-renders from scratch each time, so no saved-markup round-trip.
- No auth, network, storage, or input-validation surface touched. Pure CSS/DOM presentation change.

**Concerns:**
- None.

### Code Quality

**Green — considered and verified, no action required:**

1. **Correctness of the swap** (`app.js:835-836`) — `el()` guards with `text != null`, so the `busy ? '' : '↻'` argument sets `textContent = ''` rather than skipping; the subsequent `appendChild` is therefore safe and the button contains exactly the spinner while busy. Verified against `el()` at app.js:667.

2. **No orphaned CSS** — `tickets-refresh-spin` is deleted and `rg` finds zero remaining references anywhere in the repo (including native/Swift). The `translateY(-50%)`-composing hack it existed to work around is gone with it, and `.tickets-refresh` keeps its static `transform: translateY(-50%)` for positioning only.

3. **Sizing math holds** (`app.css:642-646`) — the global `* { box-sizing: border-box }` (app.css:24) means the 22px button has a 20px content box and `.action-spinner` renders at 12px total; the ring fits with room to spare, and `padding: 0` (commit 491bbe8) correctly zeroes the UA button padding that would otherwise squash it out of round. The absolutely-positioned `inline-flex` blockifies to `flex`, which still centers — and with fixed width/height there is no layout shift, matching the PR's stated goal.

4. **Reuses the shared primitive** — `.action-spinner` + `wizard-spin` at 0.7s linear infinite, identical to the board-title (app.js:2687, 3026) and detail-header (app.js:1500) spinners. One fewer bespoke keyframe in the stylesheet.

5. **Test coverage is real, not tautological** (`board.test.js:168-180`) — the assertions pin both directions of the swap (`.action-spinner` present + `textContent === ''` while busy; spinner absent + `textContent === '↻'` when idle), which is exactly the regression CROW-797 describes. Verified passing locally: \`npm test\` → **34 passed, 0 failed**, with both Tickets-card checks green.

6. **Naming nit (non-blocking)** — `.tickets-refresh.spinning` no longer animates anything; it now only carries `opacity: 0.6; cursor: default`. The class still reads correctly as an in-flight state marker and the tests key off it, so renaming would be churn. Noting only for the record.

7. **Accessibility (non-blocking)** — while busy the button has no text content, so its accessible name falls back to `title=\"Refreshing tickets…\"`. `title` is a valid last-resort accname source and the button is `disabled` in that state, so this is acceptable; an explicit `aria-label` would be marginally more robust if this pattern spreads.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 7 Green] findings.

Tight, well-scoped fix that deletes more than it adds, converges on an existing shared primitive, and lands with a test that would actually catch the regression coming back.

---

Issue #732, opened 2026-07-15
Ticket Board: unify per-category icons & colors across pipeline headings and sidebar


dgershman:
## Summary

On the Ticket Board screen, each pipeline **category** (Backlog / Ready / In Progress / In Review / Done) is styled inconsistently between the two places it appears:

- **Main pipeline headings** (`PipelineSegment`) render a plain colored `Circle()` dot + label — **no icon**.
- **Sidebar** (`StatusCount`) renders an **SF Symbol icon + color** per status, but the icon and color are **hardcoded in the view** rather than derived from the model.
- The canonical color source (`TicketStatus.color`) and the sidebar disagree for **Backlog** (`.gray` vs `CorveilTheme.textMuted`).

There is currently **no canonical icon** for a `TicketStatus` — icons live only as literals inside `StatusCount`, so the main headings and sidebar can't share them.

Make each category use the **same icon and color** everywhere, and style the category headings similarly across the page.

## Proposed approach

1. Add a canonical `icon` (SF Symbol `systemName`) property to `TicketStatus`, alongside the existing `color`, in the UI extension at `Packages/CrowUI/Sources/CrowUI/CorveilTheme.swift` (`extension TicketStatus`, ~line 145). Seed it from the icons already used in the sidebar so nothing visually regresses:
   - Backlog → `tray`
   - Ready → `flag.fill`
   - In Progress → `bolt.fill`
   - In Review → `eye.fill`
   - Done → `checkmark.circle.fill`
   - Unknown → a sensible fallback (e.g. `questionmark.circle`)
2. In `PipelineSegment` (`Packages/CrowUI/Sources/CrowUI/TicketBoardView.swift`), replace the bare colored `Circle()` (~line 304) with `Image(systemName: status.icon)` tinted `status.color`, so each category heading shows the same icon+color as the sidebar.
3. In `StatusCount` (same file, ~lines 616–620), source the icon and color from `TicketStatus` (`status.icon` / `status.color`) instead of hardcoded literals, so there is a single source of truth. This also fixes the Backlog color mismatch (canonical `.gray`).
4. Keep `AllPipelineSegment` (the gold `square.grid.2x2` "All" segment) as-is — it is not a status category.

## Acceptance criteria

- [ ] `TicketStatus` exposes a canonical `icon` (SF Symbol) property mirroring `color`, as the single source of truth.
[...]

PR #803, opened 2026-07-22
fix(web): green PR status checkmarks (#802)

### LINKED ISSUE #802: PR/ticket status checkmarks stay black while labels are green
## Problem

When a PR is approved and checks are passing, the status **text** renders green but the **checkmark icons** stay black. The same happens in the left-hand sidebar under status (ticket status / Done marks): the checkmarks are black instead of green.

## Expected

Checkmarks should use the same green (`--green`) as their labels — both in the session detail PR status row and in the sidebar status/ticket count icons.

## Likely cause

Unicode `✔` / `✕` / `⚠` render as Apple emoji and ignore CSS `color`, so only the adjacent label text picks up the green tint. Monochrome SVG icons (`currentColor`) would inherit correctly.

## Notes

Seen in the Crow web UI (`CrowDaemon` Resources/web).

dgershman:
Closes #802

## Problem

When a PR is approved and checks pass, the status **text** rendered green but the **checkmark icons** stayed black — in both the sidebar session-row PR pill and the session-detail PR status row.

## Cause

The glyphs `✔` / `✕` / `⚠` are Unicode dingbats that macOS renders as **Apple color emoji**, which ignore CSS `color`. Only the adjacent label text picked up the green/red tint.

## Fix

Give the emoji glyph entries an `icon` name into the existing monochrome `ICONS` set (`check` / `close` / `warning`) and render them through `icon()`, which emits an SVG with `stroke="currentColor"`. The checkmark now inherits the same color as its label.

This is the exact path the sidebar ticket counts already use for the green **Done** mark (#732) — reused, not reinvented. The geometric glyphs (`◷` / `○` / `?`) and the intentional `🏷` `crow:merge` tag stay as text, since they already respect `color`.

- `sessionRow` PR pill (`pr-ico`) → SVG when the part has an `icon`
- `prStatusInline` / `prStatusPart` → refactored to take the glyph object and emit `icon() + label`, both tinted `part.color`
- `.pr-stat` becomes an inline-flex chip so the icon and label align

Labels, `title`, and `aria-label` are unchanged (still driven by `part.label`), so the accessibility description is unaffected.

## Test plan

- Open a session whose PR is **approved + checks passing** → detail PR status row shows a **green** ✔ next to "Checks pass" / "Approved" (not black); sidebar row pill checkmarks are green.
- Failing checks / changes requested → **red** ✕. Conflicts → **red** ⚠. Merged → **purple** ✔.
- Pending (`◷`), no-checks (`?`), no-reviews (`○`), and the `🏷` merge-label marker render unchanged.

🐦‍⬛ Generated with Crow

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -764,7 +764,8 @@ html, body {
 /* ---- Detail header: gold PR chip + inline PR status ---- */
 .link-chip.link-pr { color: var(--gold); border-color: var(--border-strong); }
 .pr-status-inline { display: inline-flex; flex-wrap: wrap; align-items: center; gap: 10px; margin-left: 4px; }
-.pr-stat { font-size: 11px; font-weight: 600; }
+.pr-stat { display: inline-flex; align-items: center; gap: 4px; font-size: 11px; font-weight: 600; }
+.pr-stat-glyph { line-height: 1; }
 
 /* Artifacts — per-session generated images (CROW-593). Hidden until the
    session has any; a compact horizontal strip under the detail header. */
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1200,7 +1200,7 @@ function sessionRow(s) {
     prb.style.borderColor = color;
     const parts = prBadgeParts(pr);
     for (const part of parts) {
-      const ico = el('span', 'pr-ico', part.glyph);
+      const ico = part.icon ? icon(part.icon, 10) : el('span', 'pr-ico', part.glyph);
       ico.style.color = part.color;
       prb.appendChild(ico);
     }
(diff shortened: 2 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Pure presentational change to already-trusted, static status vocabulary. No new user/network-controlled data flows into the DOM.
- The new SVG markup comes only from the fixed `ICONS` constant set (`check`/`close`/`warning`), passed through the existing `icon()` helper — no interpolation of dynamic values into `innerHTML`, so no XSS surface is introduced.
- Labels remain the sole source of the accessibility description; `aria-label`/`title` on the sidebar pill (`app.js:1209-1211`) are still driven by `part.label`, unchanged.

**Concerns:**
- None.

### Code Quality
- Correct reuse of the established monochrome `icon()` path (the same one the sidebar Done mark already uses), rather than a new mechanism — consistent with CROW-773/CROW-732.
- `icon` names all resolve to existing `ICONS` entries: `check` (app.js:1061), `close` (app.js:1072), `warning` (app.js:1066). Verified.
- `prStatusPart` signature changed from `(text, color)` to `(part)`; all four call sites (within `prStatusInline`) were updated. No stale callers remain.
- Fallback path is preserved: parts without an `icon` field (geometric `◷`/`○`/`?` and the `🏷` merge-label tag) still render as text via `pr-ico`/`pr-stat-glyph`, matching the stated intent.
- `.pr-stat` inline-flex + gap and the `.ico` flex wrapper align the SVG with its label correctly; sidebar icons size explicitly (10px) so they don't depend on the `.pr-ico` font-size rule.
- Minor (consider, non-blocking): the `glyph` field is now dead for the four icon-bearing entries (never rendered when `icon` is set). Harmless — it documents intent and remains the fallback shape — so no change needed.
- JS passes `node --check`.

### Summary Table

---

Issue #783, opened 2026-07-20
Replace tmux with an out-of-process PTY supervisor (crow-ptyd); supersede ADR 0001


dgershman:
**Epic — decompose per phase before implementing. Phase 0 is the safe self-contained starter; a single "rip out tmux + rebuild a multiplexer" pickup is too large for one autonomous PR.**

## Context
ADR 0001 (Accepted 2026-05-25) made **tmux the sole terminal backend**. It's load-bearing — an audit of the live tree found **14 capabilities** riding on it. Goal: build an in-house PTY supervisor, reach parity, delete tmux, and supersede 0001.

**Why:** shed the hard external dependency (`tmux ≥ 3.3` must be installed; `CrowEngine/TmuxDiscovery.swift` hard-errors on absence), stop version drift, stop fighting tmux copy-mode/mouse/alternate-screen semantics from the browser, and own the multiplexer.

**The constraint that dictates the design:** live agent PTYs must survive a `crowd` crash/restart. tmux delivers this only because its **server runs outside crowd's process tree**; a `forkpty` child of crowd dies with crowd. So the replacement is a **separate long-lived supervisor** (`crow-ptyd`), not in-process forkpty. `Packages/CrowTerminal/Sources/CrowTerminal/PTYProcess.swift` already provides the raw `openpty`+`posix_spawn`+resize+reap primitive (Linux-capable via its `CPty` shim) — reused *inside* the sidecar, not in crowd.

## Capability parity checklist (re-provide each; today's tmux mechanism in parens)
1. PTY-outlives-crowd persistence — sidecar owns PTYs; crowd persists only a `{socket, session, windowId}` binding (mirrors `TmuxBinding`, `CrowCore/Models/Terminal.swift:20`).
2. Warm-adopt vs cold-recreate — liveness probe replaces `TmuxBackend.cockpitSessionIsLive()`; warm restart re-binds the window (`SessionService.adoptTerminal`), cold start relaunches the agent `--continue`.
3. Multi-client attach, independent current-window — sidecar per-window subscriber fan-out with per-client view cursor (replaces grouped `new-session -t crow-cockpit`, `TerminalCockpit.swift`). This is ADR 0009's "N clients for free."
[...]

PR #821, opened 2026-07-22
Terminal scrollback: heal stale alt-screen / 5000-line windows

### LINKED ISSUE #804: Terminal scrollback: agent-session history incomplete on scroll-up (stale alt-screen / 5000-line windows + capture-pane replay)
## Summary

Scrolling up in the web xterm.js terminal does **not** reliably show the complete Claude Code session history. After auditing the whole pipeline (tmux capture → daemon → WebSocket → xterm.js) plus a **live diagnostic of the running tmux server**, this is **not a fundamental design flaw** — the architecture is sound and works for correctly-created windows. It's a small set of concrete bugs, dominated by one: **windows created before the current tmux config are permanently stuck in the alternate-screen buffer (no scrollback) with a 5000-line history-limit, and nothing heals them after a daemon/config upgrade.**

## Live evidence (running tmux server)

`tmux -S <crow-tmux.sock> list-windows -a -F '#{window_name} size=#{history_size} limit=#{history_limit} alt=#{alternate_on}'`:

| Window | history_size | **history_limit** | **alternate_on** |
|---|---|---|---|
| Claude Code | 1 | **5000** | **1** ⚠️ broken |
| Claude Code | 121 | 50000 | 0 ✓ |
| Claude Code | 327 | 50000 | 0 ✓ |
| Claude Code | 263 | 50000 | 0 ✓ |
| Manager | 1 | **5000** | **1** ⚠️ broken |
| Manager 2 / 3 | 500 / 210 | 50000 | 0 ✓ |

- `global history-limit = 50000`, `alternate-screen off` — both correct **now**.
- But two live windows are `alternate_on=1` **and** `history_limit=5000`. A pane in the alt buffer has **zero scrollback**; a 5000 cap truncates the rest. `alternate-screen off` only governs windows created while it is in effect, and tmux **cannot retroactively** pull a pane out of the alt screen or resize a window's history (documented caveat: "history-limit applies only to new windows"). So these predate the config and are permanently degraded.
- Widths differ across windows (154/170/218 cols) — the secondary reflow problem (below).

## How the pipeline works (for reference)

- **Live output**: native PTY runs `tmux attach-session` (`TmuxBackend.attachCommand`, `TmuxBackend.swift:79`); tmux's rendered bytes stream raw over the WS into `term.write()` — no transformation (`TerminalWebSocket.swift:45-70`).
- **Reconnect / tab-switch / select-window**: the daemon **rebuilds** the buffer from `tmux capture-pane -p -e -S -50000` (`TmuxController.capturePane`, `TmuxController.swift:310-317`; `TerminalCockpit.replayData`, `TerminalCockpit.swift:98-107`) and prefixes it with `ESC[H ESC[2J ESC[3J` (home + clear screen + **clear scrollback**) so repeated selects rebuild rather than stack (`TerminalCockpit.replayFrame`, `TerminalCockpit.swift:122-128`; `TerminalWebSocket.swift:92-106`; issue #606, tested in `TerminalReplayTests.swift`). **Consequence: after any reconnect, scroll-back == exactly what `capture-pane` returned, bounded by that window's `history-limit`.**
- **The alt-screen design bet** (`crow-tmux.conf:38-65`): Claude Code is a full-screen TUI that would normally use the alt buffer (no scrollback). Crow disables it two ways — `set -gs alternate-screen off` (inner apps) + `set -gas terminal-overrides ',xterm*:smcup@:rmcup@,screen*:smcup@:rmcup@'` (client) — so the transcript lands in the main buffer's scrollback. Mouse-mode reporting is swallowed so the wheel scrolls the window instead of being sent as arrow keys (`app.js:3238-3261`; issues #776/#777/#486). This is the correct approach and mirrors Claude Code in a plain terminal (iTerm), where scroll-up shows the whole conversation.
- **xterm.js**: `scrollback: 50000` in the web UI (`app.js:3225-3232`). Note the standalone/desktop `terminal.html:74` still uses `scrollback: 10000` (discrepancy). No auto-scroll-to-bottom on write; jump-to-bottom pill addon; wheel owns scroll in both buffers, swallowing in the (should-be-absent) alt buffer (`app.js:3374-3419`).

## Root causes, ranked (bug vs. design)

1. **Stale windows stuck in alt-screen / 5000-line history after upgrade — BUG, high impact, fixable.** The config fix shipped but nothing enforces it on, or heals, pre-existing windows. This is almost certainly the reported symptom.
2. **`capture-pane` + `ESC[3J` replay caps history to tmux's buffer and reflows at the wrong width — BUG-ish / known limitation.** Every reconnect wipes xterm's own (possibly richer) live scrollback and rebuilds from a width-sensitive, `history-limit`-bounded capture. Aligns with the risk already called out in the open epic **#783** ("a raw byte ring is insufficient; tmux does the VT emulation for free").
3. **50k ceiling — design limit, tunable.** Long, tool-heavy sessions can exceed 50k lines; older transcript is then gone. Raising trades memory.
4. **Ephemeral live-region content never commits to scrollback — inherent, minor.** Content Claude Code repaints in place (streaming frame, collapsed tool output) was never a scrolled-off line, so it can't be in scrollback. Identical to a plain terminal, so **not** the reported bug.

## Fix checklist

**A. Heal + enforce window config (the high-impact fix)**
- [ ] Guarantee every **managed** window is born with `history-limit 50000` **and** main buffer (no alt screen). Verify the option scope: `crow-tmux.conf` uses `set -gs alternate-screen off`, but `alternate-screen` is a **window** option — confirm `-gs` (server) actually applies as a per-window default vs. `setw -g` / `-w`; the client-level `smcup@` override may be what's actually saving new windows today. Add/extend a `CrowTerminalTests` assertion for the effective option.
- [ ] On daemon start / after a config or daemon upgrade, **detect degraded windows** (`alternate_on=1` OR `history_limit < 50000`) and heal them. tmux can't fix these in place, so: (a) surface them in the UI as "degraded scrollback — recreate to fix", and/or (b) offer a one-click **recreate window** that relaunches the agent (`claude --continue`) in a fresh correctly-configured window, preserving session linkage.
- [ ] Add a diagnostic (`crow doctor` or a daemon log line / UI affordance) that reports per-window `history_limit` + `alternate_on` so this condition is visible.

**B. Replay fidelity (cross-link #783, do not re-spec the VT-engine rework)**
- [ ] `capture-pane` renders at the *current* pane width — capture/replay at the window's stored width to avoid re-wrap corruption when a different-width client attaches.
- [ ] Reconsider the unconditional `ESC[3J` scrollback wipe on benign reconnects/tab-switches — it can discard live-accumulated scrollback richer than the capture. Only rebuild when the buffer is actually stale.
- [ ] Cross-link **#783** (Phase 2: per-window VT-state model + ≥50k scrollback ring) as the strategic home for faithful screen replay; this ticket stays within the current tmux architecture.

**C. Ceiling + consistency**
- [ ] Make the 50k scrollback/`history-limit` configurable; consider a higher default for agent sessions.
- [ ] Align `terminal.html:74` (`scrollback: 10000`) with the web UI's 50000 (or the configured value) if that surface is still shipped (post-ADR-0010).

**D. Tests**
- [ ] Extend `TerminalReplayTests` for width/reflow and the no-wipe-on-benign-reconnect behavior.
- [ ] Add a test asserting managed windows are created with `history-limit 50000` and effective main-buffer config.

## Reproduce / verify

```bash
SOCK=$(find "$TMPDIR" -maxdepth 1 -name 'crow-tmux.sock')
tmux -S "$SOCK" list-windows -a -F '#{window_name} size=#{history_size} limit=#{history_limit} alt=#{alternate_on}'
```
Windows created before the config bump show `limit=5000 alt=1` (broken); windows created after show `limit=50000 alt=0` (scroll-up works). Recreating a broken window fixes it — confirming it's the stale-window bug, not a rendering dead-end.

## References
- Prior art (closed): #606 (scrollback replay), #776/#777 (scroll/mouse-mode regressions), #486/#452 (copy-mode scroll trap), #637/#638 (resize clobber), #668/#686 (jump-to-bottom).
- Strategic (open): **#783** (replace tmux with `crow-ptyd`; VT-engine + scrollback ring is a GO/NO-GO gate) — the home for the deep replay rework.
- ADRs: 0001 (tmux-only backend), 0006 (xterm.js renderer), 0010 (web UI is the only client).
- Key code: `Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf` (:38-65, :126-142); `TmuxController.swift:310-317`; `TerminalCockpit.swift:96-128`; `TerminalWebSocket.swift:45-106`; `TmuxBackend.swift:79`; `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` (:3225-3419, :3587-3620).

dhilgaertner:
Closes #804

## Problem

Scrolling up in the web xterm.js terminal doesn't reliably show the complete Claude Code session history. The dominant, reproducible cause (verified against a live tmux read): **windows created before the current `crow-tmux.conf` are permanently stuck in the alternate-screen buffer (`alternate_on=1`) with `history-limit=5000`.** A pane in the alt buffer has zero scrollback; the 5000 cap truncates the rest. tmux can't resize history or leave the alt buffer in place, and nothing detected, surfaced, or healed these after the config bump.

```
Manager     limit=5000   alt=1   ⚠ broken (pre-config)
Claude Code limit=5000   alt=1   ⚠ broken (pre-config)
Claude Code limit=50000  alt=0   ✓ (all windows born after the bump)
```

## Fix (checklist item A: detect + surface + one-click recreate)

- **Detect**: `TmuxController.listWindowScrollback()` reads per-window `history_limit` + `alternate_on`; `TmuxBackend.isScrollbackDegraded()` is the pure policy (alt buffer OR below the 50000 floor). `TmuxBackend.scrollbackHistoryLimit` is the single source of truth shared with the conf, the daemon replay window, and xterm.js.
- **Diagnostic**: the daemon logs every degraded window on start (`tmux:scrollback_degraded index=… history_limit=… alternate_on=…`).
- **Surface**: `list-terminals` now returns `scrollback_degraded`; the web UI badges the affected tab with a ⚠ and offers a confirm-gated **Recreate**.
- **Heal**: `recreate-terminal` RPC → `SessionService.recreateTerminalSurface` kills the degraded window and rebuilds a correctly-configured one (50000-line main buffer), relaunching the agent via `claude --continue` — reusing the CROW-747 rehydrate/relaunch machinery. Manager terminals route to `restartManager`. A `crow recreate-terminal` CLI command mirrors it.

Healing is **manual** (no auto-heal): recreating interrupts the live agent, so the user decides when.

## Consistency

- Desktop `xterm/terminal.html` scrollback `10000 → 50000` (matches the web UI / tmux / replay).
- Stale "10k-line scrollback" comments in `crow-tmux.conf` corrected.

## Tests

- `ScrollbackHealthTests` — pure policy truth table (old limit, alt buffer, floor boundary, custom floor).
- `TmuxControllerTests.newWindowUnderBundledConfIsNotDegraded` — real-tmux enforcement that a window created under the **bundled** conf is born `50000`/main-buffer and not flagged degraded.
- `RecreateTerminalTests` — recreate not-found returns false + RPC param validation, headless.

All build; new tests pass. (One pre-existing readiness-timing flake, `retryReadinessEmitsTimedOutWhenSentinelMissing`, is unrelated — confirmed failing on the clean base.)

## Out of scope (follow-up)

Checklist B (capture-pane/`ESC[3J` replay width/reflow fidelity) is left to epic #783 rather than re-specifying the VT-engine rework here.

https://claude.ai/code/session_01GYMwoE9Wi8HMdteFE9NjHm

### DIFF
--- a/Packages/CrowCLI/Sources/CrowCLILib/Commands/TerminalCommands.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Commands/TerminalCommands.swift
@@ -71,6 +71,31 @@ public struct CloseTerminal: ParsableCommand {
     }
 }
 
+/// Recreate a terminal whose tmux window has degraded scrollback (CROW-804).
+/// Kills the stale window and rebuilds a fresh, correctly-configured one,
+/// relaunching the agent (`claude --continue`). Destructive to the running
+/// agent — the web UI confirms first; from the CLI the caller owns that choice.
+public struct RecreateTerminal: ParsableCommand {
+    public static let configuration = CommandConfiguration(commandName: "recreate-terminal", abstract: "Recreate a terminal to restore full scrollback (CROW-804)")
+    @Option(name: .long, help: "Session UUID") var session: String
+    @Option(name: .long, help: "Terminal UUID") var terminal: String
+
+    public init() {}
+
+    public func validate() throws {
+        try validateUUID(session, label: "session UUID")
+        try validateUUID(terminal, label: "terminal UUID")
+    }
+
+    public func run() throws {
+        let result = try rpc("recreate-terminal", params: [
+            "session_id": .string(session),
+            "terminal_id": .string(terminal),
+        ])
+        printJSON(result)
+    }
+}
+
 /// Rename a terminal tab.
 public struct RenameTerminal: ParsableCommand {
     public static let configuration = CommandConfiguration(commandName: "rename-terminal", abstract: "Rename a terminal tab")
--- a/Packages/CrowCLI/Sources/CrowCLILib/CrowCommand.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/CrowCommand.swift
@@ -30,6 +30,7 @@ public struct CrowCommand: ParsableCommand {
             NewTerminal.self,
             ListTerminals.self,
             CloseTerminal.self,
+            RecreateTerminal.self,
             RenameTerminal.self,
             Send.self,
             AddLink.self,
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -357,6 +357,9 @@ html, body {
 .tab.add { color: var(--gold); font-size: 16px; font-weight: 700; }
 .tab-close { color: var(--text-muted); font-size: 14px; }
 .tab-close:hover { color: var(--red); }
+/* CROW-804: degraded-scrollback warning badge on a terminal tab. */
+.tab-degraded { color: var(--gold); font-size: 12px; }
+.tab-degraded:hover { color: var(--text-primary); }
 #terminal-wrap { flex: 1; min-height: 0; background: #1e1e1e; position: relative; overscroll-behavior: contain; }
 /* #777: `touch-action: none` hands the one-finger gesture to enableTouchScroll
    instead of iOS Safari's overscroll — without it Safari rubber-bands the fixed
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -360,6 +360,12 @@ html, body {
 /* CROW-804: degraded-scrollback warning badge on a terminal tab. */
 .tab-degraded { color: var(--gold); font-size: 12px; }
 .tab-degraded:hover { color: var(--text-primary); }
+/* CROW-804: Manager surface has no tabs — a slim warning strip carries the
+   degraded-scrollback notice + Recreate action instead. */
+.degraded-strip { display: flex; align-items: center; gap: 8px; padding: 6px 12px; font-size: 12px; }
+.degraded-msg { color: var(--text-secondary); }
+.degraded-recreate { color: var(--gold); cursor: pointer; font-weight: 600; }
+.degraded-recreate:hover { color: var(--text-primary); text-decoration: underline; }
 #terminal-wrap { flex: 1; min-height: 0; background: #1e1e1e; position: relative; overscroll-behavior: contain; }
 /* #777: `touch-action: none` hands the one-finger gesture to enableTouchScroll
    instead of iOS Safari's overscroll — without it Safari rubber-bands the fixed
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -2053,7 +2053,25 @@ function renderTabs() {
   // moots the stale-tab naming bug: a renamed manager session has no tab to go
   // stale, since tabs are labeled from the terminal name, not the session name.)
   const sel = sessions.find((x) => x.id === selectedId);
-  if (sel && sel.kind === 'manager') return;
+  if (sel && sel.kind === 'manager') {
+    // #680: managers have no tabs. But the Manager window is a common CROW-804
+    // "stuck alt-screen / 5000-line" case, and with no tab there'd be no ⚠ or
+    // Recreate. Surface a slim warning strip with a Recreate action when the
+    // Manager terminal is degraded; otherwise leave #tabbar empty so it stays
+    // collapsed. Recreate routes through restartManager (SessionService).
+    const degraded = terminals.find((t) => t.scrollback_degraded);
+    if (degraded) {
+      const strip = el('div', 'degraded-strip');
+      strip.appendChild(el('span', 'tab-degraded', '⚠'));
+      strip.appendChild(el('span', 'degraded-msg', 'Scrollback degraded — this Manager window can\'t show full history.'));
+      const btn = el('span', 'degraded-recreate', 'Recreate');
+      btn.title = 'Rebuild this Manager terminal to restore full scroll-up history (restarts the agent).';
+      btn.onclick = () => recreateTerminal(degraded);
+      strip.appendChild(btn);
+      bar.appendChild(strip);
+    }
+    return;
+  }
   for (const t of terminals) {
     const tab = el('div', 'tab' + (activeTerminal && t.id === activeTerminal.id ? ' active' : ''));
     const label = el('span', null, t.name);
@@ -2128,8 +2146,13 @@ async function recreateTerminal(t) {
     return;
   }
   await refreshTerminals();
-  // Re-attach so the fresh window streams into the surface immediately.
-  if (activeTerminal && activeTerminal.id === t.id) attachWindow(activeTerminal.window);
+  // Recreate keeps the terminal id but binds a NEW tmux window index, and
+  // refreshTerminals() only swaps activeTerminal when the id disappears — so the
+  // active tab would keep the old (killed) window object and attachWindow's
+  // `win === attachedWindow` guard would skip the reload. Re-point at the
+  // refreshed row and switch onto its new window so the fresh pane streams in.
+  const refreshed = terminals.find((x) => x.id === t.id);
+  if (refreshed) switchTerminal(refreshed);
 }
 
 // ---------------------------------------------------------------------------
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -2059,6 +2059,15 @@ function renderTabs() {
     const label = el('span', null, t.name);
     label.onclick = () => switchTerminal(t);
     tab.appendChild(label);
+    // CROW-804: this terminal's tmux window is stuck with degraded scrollback
+    // (alternate-screen buffer and/or the old 5000-line history-limit) that
+    // tmux can't fix in place. Badge it and offer a one-click recreate.
+    if (t.scrollback_degraded) {
+      const warn = el('span', 'tab-degraded', '⚠');
+      warn.title = 'Scrollback degraded — this window can\'t show full history (created before the current config). Click to recreate it and restore scroll-up.';
+      warn.onclick = (e) => { e.stopPropagation(); recreateTerminal(t); };
+      tab.appendChild(warn);
+    }
     const close = el('span', 'tab-close', '×');
     close.onclick = (e) => { e.stopPropagation(); closeTerminal(t); };
     tab.appendChild(close);
@@ -2100,6 +2109,27 @@ async function closeTerminal(t) {
   try { await rpc('close-terminal', { session_id: selectedId, terminal_id: t.id }); } catch (_) {}
   if (activeTerminal && activeTerminal.id === t.id) activeTerminal = null;
   await refreshTerminals();
+}
+
+// CROW-804: heal a terminal whose tmux window has degraded scrollback. Recreate
+// kills the window and rebuilds it under the current config, relaunching the
+// agent (`claude --continue`) — so confirm first, since it interrupts whatever
+// is running in the pane.
+async function recreateTerminal(t) {
+  const ok = await confirmModal(
+    'This rebuilds “' + (t.name || 'terminal') + '” to restore full scroll-up history. '
+    + 'The agent running in it will be restarted (resumed via claude --continue).',
+    { title: 'Recreate terminal', okLabel: 'Recreate', danger: true });
+  if (!ok) return;
+  try {
+    await rpc('recreate-terminal', { session_id: selectedId, terminal_id: t.id });
+  } catch (e) {
+    if (term) term.write('\r\n\x1b[31m[crow] recreate-terminal failed: ' + (e.message || e) + '\x1b[0m\r\n');
+    return;
+  }
+  await refreshTerminals();
+  // Re-attach so the fresh window streams into the surface immediately.
+  if (activeTerminal && activeTerminal.id === t.id) attachWindow(activeTerminal.window);
 }
 
 // ---------------------------------------------------------------------------
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -2143,6 +2143,10 @@ async function recreateTerminal(t) {
     await rpc('recreate-terminal', { session_id: selectedId, terminal_id: t.id });
   } catch (e) {
     if (term) term.write('\r\n\x1b[31m[crow] recreate-terminal failed: ' + (e.message || e) + '\x1b[0m\r\n');
+    // Register-then-kill means a failed heal leaves the old window live and
+    // still degraded — refresh so the ⚠ / Recreate affordance re-renders for a
+    // retry instead of vanishing on a half-applied state.
+    await refreshTerminals();
     return;
   }
   await refreshTerminals();
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -2136,7 +2136,7 @@ async function closeTerminal(t) {
 async function recreateTerminal(t) {
   const ok = await confirmModal(
     'This rebuilds “' + (t.name || 'terminal') + '” to restore full scroll-up history. '
-    + 'The agent running in it will be restarted (resumed via claude --continue).',
+    + 'The agent running in it will be restarted (and resumed where the agent supports it).',
     { title: 'Recreate terminal', okLabel: 'Recreate', danger: true });
   if (!ok) return;
   try {
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -2146,12 +2146,16 @@ async function recreateTerminal(t) {
     return;
   }
   await refreshTerminals();
-  // Recreate keeps the terminal id but binds a NEW tmux window index, and
-  // refreshTerminals() only swaps activeTerminal when the id disappears — so the
-  // active tab would keep the old (killed) window object and attachWindow's
-  // `win === attachedWindow` guard would skip the reload. Re-point at the
-  // refreshed row and switch onto its new window so the fresh pane streams in.
-  const refreshed = terminals.find((x) => x.id === t.id);
+  // Recreate binds a FRESH tmux window, but `new-window` (no -a) reuses the
+  // index just freed by killWindow — so the new index usually EQUALS the old
+  // one. Even after re-pointing at the refreshed row, attachWindow's
+  // `win === attachedWindow` guard would then skip the reload and leave the
+  // surface on the dead pane (the stale-*index* case; the stale-*object* case
+  // was fixed earlier). Clear attachedWindow so the reattach can't short-circuit,
+  // then switch onto the refreshed row — for the primary Manager its id changes,
+  // so fall back to the activeTerminal refreshTerminals already swapped in.
+  const refreshed = terminals.find((x) => x.id === t.id) || activeTerminal;
+  attachedWindow = null;
   if (refreshed) switchTerminal(refreshed);
 }
 
--- a/Packages/CrowDaemon/Sources/CrowDaemon/TerminalCockpit.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/TerminalCockpit.swift
@@ -35,6 +35,20 @@ struct TerminalCockpit: Sendable {
         // Reap grouped sessions this or a prior crowd leaked (#667). Runs on
         // every startup, whether we adopted or created the cockpit.
         reapOrphanedViewSessions()
+        logDegradedScrollbackWindows()
+    }
+
+    /// Diagnostic: log every window whose scroll-up can't show the full
+    /// transcript — stuck in the alternate-screen buffer (`alternate_on=1`)
+    /// and/or capped below the current `history-limit` because it was created
+    /// before the config bump. tmux can't heal these in place; the web UI badges
+    /// them and offers a recreate (CROW-804). Best-effort; never throws.
+    private func logDegradedScrollbackWindows() {
+        guard let windows = try? controller.listWindowScrollback() else { return }
+        for w in windows where TmuxBackend.isScrollbackDegraded(
+            historyLimit: w.historyLimit, alternateOn: w.alternateOn) {
+            NSLog("[CrowTelemetry tmux:scrollback_degraded index=\(w.index) history_limit=\(w.historyLimit) alternate_on=\(w.alternateOn ? 1 : 0)]")
+        }
     }
 
     /// Kill `crowd-web-*` grouped sessions left detached by a prior crowd's
--- a/Packages/CrowDaemon/Sources/CrowDaemon/TerminalCockpit.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/TerminalCockpit.swift
@@ -103,11 +103,13 @@ struct TerminalCockpit: Sendable {
         _ = try? controller.run(["select-window", "-t", "\(group):\(index)"])
     }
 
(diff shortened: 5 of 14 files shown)


dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None (no Red).

### Security Review
**Strengths:**
- CLI validates session/terminal UUIDs before RPC; `recreate-terminal` rejects missing/invalid ids with `invalidParams`.
- Confirm modal + `el(..., text)` uses `textContent`, so terminal names in the recreate prompt are not HTML-injected.
- Destructive recreate is explicit (web confirm; CLI caller-owned) and reuses existing kill/rehydrate paths rather than inventing a new privilege surface.
- Degraded detection is best-effort (`try?` / empty set on failure) and does not fail startup or `list-terminals`.

**Concerns:**
- None material. `recreate-terminal` is as powerful as other local terminal RPCs under the existing socket/web-auth trust model.

### Code Quality
- **Yellow — stale window attach after recreate (`app.js`):** `refreshTerminals()` only replaces `activeTerminal` when the id is *missing*:
  ```js
  if (!activeTerminal || !terminals.find((t) => t.id === activeTerminal.id)) {
    activeTerminal = terminals[0] || null;
  }
  ```
  Recreate preserves terminal id but changes `window`. The UI keeps the old object, then `attachWindow(activeTerminal.window)` hits `if (win === attachedWindow) return` and never reloads onto the new index. The extra `attachWindow` at the end of `recreateTerminal` has the same stale value. Backend heal succeeds; the active tab stays on the killed window until the user switches tabs. Fix: when the id is still present, assign `activeTerminal` from the refreshed row (or `switchTerminal(refreshed)`), then attach.
- **Yellow — Manager heal not surfaced in web UI:** PR problem statement calls out Manager as a common `limit=5000 alt=1` case, and `recreateTerminalSurface` routes Manager → `restartManager`, but `renderTabs()` returns early for `kind === 'manager'` (no tabs → no ⚠ / Recreate). Users get logs + CLI/Settings restart only — no in-product signal for the primary broken surface.
- **Green — unused API:** `TmuxBackend.windowScrollbackHealth()` is unused; `TerminalCockpit` calls `controller.listWindowScrollback()` directly.
- **Green — “SSOT” still multiplies:** `scrollbackHistoryLimit = 50000` is documented as shared with conf / daemon replay / xterm, but conf + `app.js` / `terminal.html` still hardcode `50000`. Drift risk remains.
- **Green — `list-terminals` cost:** every list now shells `list-windows` via `degradedWindowIndices()`. Fine for push/refresh; watch if this becomes a hot poll path.

### Tests
- `ScrollbackHealthTests` (5) and `RecreateTerminalTests` (2) passed under package-local `swift test --filter …`.
- Happy-path recreate (kill + new window index + UI reattach) is still untested — the stale-attach bug above would have been caught by a thin integration/UI check.

### Summary Table

---

dhilgaertner:
Thanks for the review — addressed in `7087908`:

**Yellow — stale window attach after recreate (`app.js`):** Fixed. After `refreshTerminals()`, `recreateTerminal` now re-points at the refreshed row and `switchTerminal`s onto its new window index, so the active tab reloads onto the fresh window instead of short-circuiting on the killed one. (The Manager path already reattaches correctly because `restartManager` mints a new terminal id, which `refreshTerminals` swaps in.)

**Yellow — Manager heal not surfaced:** Fixed. `renderTabs()` no longer returns early unconditionally for managers — when the Manager terminal is degraded it renders a slim warning strip (⚠ + "Recreate") that routes through `restartManager`. Non-degraded managers keep the collapsed empty tabbar.

**Green — unused API:** Removed `TmuxBackend.windowScrollbackHealth()`; `TerminalCockpit` reads `controller.listWindowScrollback()` directly.

**Green — `list-terminals` cost:** Confirmed a non-issue and left as-is. `refreshTerminals()` is event-driven (session select / terminal ops) — the periodic polls (`refreshSessions` 10s, `refreshLive` 4s, `refreshArtifacts` 5s) never call it — so the added `list-windows` is not on a hot poll path.

**Green — SSOT still multiplies:** Left as-is. The `50000` literal is inherently duplicated across a Swift constant, the tmux conf text, and JS — there's no shared import across those three languages without a codegen step. `TmuxBackend.scrollbackHistoryLimit` is the documented authority and the enforcement test pins new windows to it; the conf/JS literals are annotated as mirroring it.

**Tests:** Happy-path recreate remains covered by the not-found + RPC-validation guards plus the real-tmux enforcement test; the JS reattach fix has no unit harness in this repo, so it's covered by the reasoning above rather than a new test.

dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Destructive recreate is confirm-gated in the web UI (`confirmModal` with `danger: true`); CLI documents that the caller owns the choice.
- RPC params are UUID-validated; terminal must belong to the named session before any kill/rehydrate.
- Window kill uses the terminal's own `tmuxBinding.windowIndex` (no caller-supplied index).
- Modal body uses `textContent` via `el()` — no HTML injection from terminal names.
- No secrets, auth bypasses, or command-injection surfaces introduced; tmux args stay structured arrays.

**Concerns:**
- None security-blocking. Recreate remains a privileged local/authenticated RPC (same trust model as `close-terminal` / `restartManager`).

### Code Quality
- **Yellow — wrong Manager target:** `SessionService.recreateTerminalSurface` (`SessionService.swift` ~490–492) branches on `session.isManager` and always calls `restartManager`, which hard-codes `AppState.managerSessionID`. Additional Manager sessions from `createManagerSession` / `create-manager` are also `kind == .manager`. Recreating a *secondary* Manager therefore restarts the **primary** Manager (destructive) and leaves the selected degraded window untouched. Gate on `sessionID == AppState.managerSessionID` (or equivalent) and fall through to the rehydrate path for non-primary Managers.
- **Green — confirm copy is Claude-specific:** `recreateTerminal` in `app.js` says the agent resumes via `claude --continue`, but the primary Manager path uses a fresh `managerCommand` (new terminal UUID), and non-Claude work agents use `agent.autoLaunchCommand`. Consider agent-/Manager-aware wording.
- **Green — “single source of truth” is still multi-literal:** `TmuxBackend.scrollbackHistoryLimit` is great for the policy floor, but `TerminalCockpit.replayLines`, web/desktop xterm `scrollback: 50000`, and `crow-tmux.conf` remain independent literals. Wiring daemon replay to the shared constant (where package deps allow) would harden the consistency claim.
- **Green — test coverage:** `ScrollbackHealthTests` + not-found/RPC validation are solid; a unit test that a non-primary Manager does *not* call `restartManager` would lock the fix above. Happy-path recreate remains intentionally untested (tmux).

### Static Analysis
- `ScrollbackHealthTests` (5) and `RecreateTerminalTests` (2) passed under Swift Testing.
- Package build for CrowTerminal/CrowEngine succeeded for the changed paths.

### Summary Table

---

dhilgaertner:
Thanks — second pass addressed in `9acbe8d`:

**Yellow — wrong Manager target:** Fixed. `recreateTerminalSurface` no longer branches on `session.isManager` (which matched secondary Managers too and would restart the *primary* via `restartManager`'s hard-coded `AppState.managerSessionID`). It now gates on the well-known primary id via a pure `SessionService.shouldRestartPrimaryManagerOnRecreate(sessionID:)`. Secondary Managers fall through to the rehydrate path, which re-runs their stored `managerCommand` on a fresh window and heals the *selected* window. Locked by the new `onlyPrimaryManagerRestartsManager` unit test.

**Green — Claude-specific confirm copy:** Generalized — the recreate confirm now says "resumed where the agent supports it" instead of naming `claude --continue`, so it fits the primary Manager (fresh `managerCommand`) and non-Claude agents.

**Green — SSOT still multi-literal:** `TerminalCockpit.replayLines` now derives from `TmuxBackend.scrollbackHistoryLimit` (CrowDaemon already depends on CrowTerminal), so the daemon replay window can't drift from the policy floor. The tmux conf and xterm HTML remain literals — no cross-language import exists — and stay annotated as mirroring the constant.

**Green — test coverage:** Added the non-primary-Manager policy test noted above.

(Also merged latest `main` into the branch.)

dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None (no Red).

### Security Review
**Strengths:**
- CLI validates session/terminal UUIDs; RPC rejects missing ids with `invalidParams`.
- `recreateTerminalSurface` requires the terminal to belong to the named session; kill uses the stored `tmuxBinding.windowIndex` (no caller-supplied index).
- Web path is confirm-gated (`confirmModal` + `danger: true`); modal body uses `el()` → `textContent` (no HTML injection from terminal names).
- Primary vs secondary Manager routing is correctly gated on `AppState.managerSessionID` (`shouldRestartPrimaryManagerOnRecreate`), with a unit test lock.
- No new secrets, auth bypasses, or command-injection surfaces; tmux args stay structured arrays. Trust model matches existing local/authenticated terminal RPCs.

**Concerns:**
- None security-blocking.

### Code Quality
- **Yellow — recreate reattach still no-ops when tmux reuses the window index (`app.js`):** `attachWindow` bails on `win === attachedWindow`, then only `reloadTerminal()`s on a real change. `TmuxController.newWindow` uses bare `new-window` (no `-a`), which creates at the **first free index** — almost always the index just freed by `killWindow`. So after a successful heal:
  1. Work terminals: `switchTerminal(refreshed)` passes the same numeric `window` → no reload / no `select-window` onto the fresh pane.
  2. Primary Manager: `restartManager` mints a new terminal id (so `refreshTerminals` swaps `activeTerminal`), but `attachWindow(newWindow)` still short-circuits when the index is reused.
  Backend heal succeeds; the active surface can stay on a dead/wrong pane until the user manually reloads. Prior review fixed the stale-*object* case; this is the remaining stale-*index* case. Fix: clear `attachedWindow` (or call `reloadTerminal()` after pointing at the refreshed row) at the end of `recreateTerminal`, e.g. `attachedWindow = null` then `switchTerminal(refreshed || activeTerminal)`.
- **Yellow — recreate reports success after a failed rehydrate (`SessionService.swift`):** `recreateTerminalSurface` kills the old window, then always `return true` after `rehydrateTerminalSurface`, even when registration fails and `tmuxBinding` stays `nil`. RPC returns `{recreated: true}`; the UI refreshes as if healed, but the terminal is now unbound (worse than the degraded-but-live prior state). Return `false` / throw when the post-rehydrate binding is missing so the web error path can surface it.
- **Green — enforcement test is weak on alt-screen:** `newWindowUnderBundledConfIsNotDegraded` only runs `/bin/sh -c 'sleep 60'`, which never requests the alternate buffer — so a regression that broke `alternate-screen off` / `smcup@` would still pass. Assert the effective window option (or drive a process that issues smcup) if you want the checklist item from #804 fully locked.
- **Green — conf still uses `set -gs alternate-screen off`:** Issue #804 notes this is a window option (`setw -g` / `set -gw` may be the correct scope). Empirically new windows are healthy (likely helped by `terminal-overrides` smcup@); worth confirming option scope so detection/`alternateOn` stays meaningful long-term.

### Tests
- `ScrollbackHealthTests` (5) and `RecreateTerminalTests` (3) passed under Swift Testing.
- Architecture warning from an x86_64 xctest bundle residual is environmental; the arm64 Swift Testing run is green.

### Summary Table

---

dhilgaertner:
Thanks — third pass addressed in `7c52f00`:

**Yellow — stale-*index* reattach (`app.js`):** Fixed. You're right that `new-window` (no `-a`) reuses the index just freed by `killWindow`, so the recreated window usually has the *same* numeric index and `attachWindow`'s `win === attachedWindow` guard skipped the reload. `recreateTerminal` now sets `attachedWindow = null` and re-switches (`switchTerminal(refreshed || activeTerminal)`), so the reattach always fires — for work terminals and for the primary Manager's swapped-in row alike.

**Yellow — false success on failed rehydrate (`SessionService.swift`):** Fixed. The old window is already killed, so `recreateTerminalSurface` now returns `false` when the post-rehydrate `tmuxBinding` is `nil` (terminal left unbound), and the `recreate-terminal` RPC throws "Could not recreate terminal (not found or tmux window failed to register)" — the web `recreateTerminal` catch writes that to the surface instead of refreshing as if healed.

**Green — alt-screen option scope:** Fixed. `alternate-screen` is a window option, so the conf now uses `set -gw alternate-screen off` (documented scope). I verified in tmux 3.6a that this is behavior-identical to the old `-gs` (both → `alternate_on=0`; a bare conf with no setting → `1`), so `alternateOn` detection stays meaningful.

**Green — weak enforcement test:** Fixed. Added `bundledConfKeepsInnerAppsOutOfAlternateBuffer`, which drives a pane that issues smcup (DECSET `1049h`) and asserts it stays in the main buffer (`alternate_on == 0`). Since a bare conf yields `1` for the same pane, this now catches a regression that broke `alternate-screen off` or its scope — the previous `sleep`-only pane never requested the alt buffer.

Verification: full build clean, `node --check` passes, `ScrollbackHealthTests` (5), `TmuxController` enforcement incl. the new alt-screen test, and `RecreateTerminalTests` (3) all green.

dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None (no Red).

### Security Review
**Strengths:**
- CLI validates session/terminal UUIDs; RPC rejects missing ids with `invalidParams`.
- `recreateTerminalSurface` requires the terminal to belong to the named session; kill uses the stored `tmuxBinding.windowIndex` (no caller-supplied index).
- Web path is confirm-gated (`confirmModal` + `danger: true`); modal body uses `el()` → `textContent` (no HTML injection from terminal names).
- Primary vs secondary Manager routing is correctly gated on `AppState.managerSessionID` (`shouldRestartPrimaryManagerOnRecreate`), with a unit test lock.
- Prior review gaps are addressed: force-reattach via `attachedWindow = null`, Manager degraded strip, fail-loud when rehydrate leaves `tmuxBinding == nil`, bundled-conf alt-screen enforcement test, `set -gw alternate-screen off`.
- No new secrets, auth bypasses, or command-injection surfaces; tmux args stay structured arrays. Trust model matches existing local/authenticated terminal RPCs.

**Concerns:**
- None security-blocking.

### Code Quality
- **Yellow — kill-then-register makes register failure worse than degraded-but-live (`SessionService.swift` ~505–536):** `recreateTerminalSurface` kills the old window *before* `rehydrateTerminalSurface` registers a replacement. The new fail-loud `return false` when `updated.tmuxBinding == nil` is correct for the RPC, but the pane is already gone — worse than the prior degraded-but-live state the comment calls out. Compounding effects:
  1. `applyRehydrationResult(..., original: seed, ...)` passes a seed whose binding was already cleared, so a failed register does **not** persist the unbound state (`nil != nil` is false) and disk keeps a dead `windowIndex`.
  2. After any `refreshTerminals()`, `scrollback_degraded` is false (no binding) so the ⚠ / Recreate affordance disappears — UI heal path is gone even though the terminal is broken.
  3. The web `catch` writes the error into `term` (likely still pointed at the dead pane) and returns without refreshing, so the user may see neither a clear error nor a reliable retry control.

  Prefer register-then-kill (keep the old index, bind the new window, then `killWindow` the saved old index), or on register failure avoid applying a nil binding / restore the previous row and surface a persistent retry strip. Brief duplicate-agent overlap is better than an unbound terminal with no badge.

- **Green — SSOT still multi-literal:** `TmuxBackend.scrollbackHistoryLimit` now feeds policy + `TerminalCockpit.replayLines`, but `crow-tmux.conf` and web/desktop xterm `scrollback: 50000` remain independent literals.
- **Green — `list-terminals` cost:** each list shells `list-windows` via `degradedWindowIndices()`. Fine for push/refresh; watch if this becomes a hot poll path.
- **Green — alt-screen test timing:** `bundledConfKeepsInnerAppsOutOfAlternateBuffer` uses `Thread.sleep(0.3)` before reading `alternate_on`; consider a short poll/retry if this flakes under load.

### Tests
- `ScrollbackHealthTests` (5) and `RecreateTerminalTests` (3) passed under Swift Testing.
- Architecture warning from an x86_64 xctest bundle residual is environmental; the arm64 Swift Testing run is green.

### Summary Table

---

dhilgaertner:
Thanks — fourth pass addressed in `1afe16b`:

**Yellow — kill-then-register worse than degraded-but-live:** Fixed with **register-then-kill**. `recreateTerminalSurface` now registers the fresh window first and only kills the saved old index once the new binding is confirmed. This resolves all three compounding effects you listed:
1. No stale-disk mismatch — on failure `applyRehydrationResult` is *not* called (the old binding stays in both memory and disk), so state is consistent; on success it persists the new binding before the kill.
2. Badge persists — on failure the old window is still live and degraded, so `scrollback_degraded` stays true and the ⚠ / Recreate affordance remains.
3. The re-armed readiness is snapshotted and restored on failure so a managed terminal isn't left half-armed, and the RPC returns false → the web error path surfaces it. Accepted cost on success is a brief duplicate-agent overlap, which (as you noted) beats an unbound terminal.

Also: the web `catch` now calls `refreshTerminals()` so the affordance re-renders for a retry after a failed heal.

**Green — alt-screen test timing:** Fixed. `bundledConfKeepsInnerAppsOutOfAlternateBuffer` now runs a shell that *continuously* re-requests the alt screen and polls `alternate_on` across a 1s window (asserting all-0), instead of a single 0.3s sleep — robust against an early read and load-induced timing.

**Green — SSOT conf/xterm literals** and **`list-terminals` cost:** Confirmed non-issues, left as-is. The conf and xterm HTML literals have no cross-language import path to `TmuxBackend.scrollbackHistoryLimit` (documented as mirroring it); `refreshTerminals` is event-driven (session-select / terminal ops — the `setInterval` polls never call it), so the added `list-windows` isn't a hot path.

Verification: full build clean, `node --check` passes, and `ScrollbackHealthTests` (5), the `TmuxController` enforcement + polling alt-screen test, and `RecreateTerminalTests` (3) are all green.

dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None (no Red).

### Security Review
**Strengths:**
- CLI validates session/terminal UUIDs; RPC rejects missing ids with `invalidParams`.
- `recreateTerminalSurface` requires the terminal to belong to the named session; kill uses the stored `tmuxBinding.windowIndex` (no caller-supplied index).
- Web path is confirm-gated (`confirmModal` + `danger: true`); modal body uses `el()` → `textContent` (no HTML injection from terminal names).
- Primary vs secondary Manager routing is correctly gated on `AppState.managerSessionID` (`shouldRestartPrimaryManagerOnRecreate`), with a unit test lock.
- Register-then-kill preserves degraded-but-live on register failure (readiness restored; ⚠/Recreate remains); RPC fails loud when binding stays nil.
- No new secrets, auth bypasses, or command-injection surfaces; tmux args stay structured arrays. Trust model matches existing local/authenticated terminal RPCs.

**Concerns:**
- None security-blocking.

### Code Quality
Prior Yellows from earlier review rounds are addressed: force-reattach via `attachedWindow = null`, Manager degraded strip, fail-loud unbound, primary-only `restartManager`, register-then-kill, robust alt-screen poll test, `set -gw alternate-screen off`.

- **Green — stale failure comments after register-then-kill:** `EngineRouter.swift` `recreate-terminal` handler still says a failed register leaves the terminal unbound because “the old window was killed”; `SessionService.recreateTerminalSurface`’s doc comment still reads kill-then-rebuild. Code is correct — worth aligning the comments so the next pass doesn’t reintroduce kill-first.
- **Green — `app.js` reattach comment:** still describes kill-then-register index reuse; with register-then-kill the work-terminal index usually differs, but clearing `attachedWindow` remains necessary for the primary Manager path (index reuse after `restartManager`). Comment could note that.
- **Green — SSOT still multi-literal:** `TmuxBackend.scrollbackHistoryLimit` feeds policy + `TerminalCockpit.replayLines`; conf / web / desktop xterm still hardcode `50000` (documented; prior review accepted).
- **Green — accepted overlap:** brief duplicate-agent window during register-then-kill is the right trade vs unbound-on-failure.

### Tests
- `ScrollbackHealthTests` (5) passed.
- `RecreateTerminalTests` (3) passed.
- `newWindowUnderBundledConfIsNotDegraded` and `bundledConfKeepsInnerAppsOutOfAlternateBuffer` passed (real tmux).
- Architecture warning from an x86_64 xctest bundle residual is environmental; the arm64 Swift Testing run is green.

### Summary Table

---

Issue #822, opened 2026-07-23
Spike: web terminal shows duplicate Claude Code TUI frames on scroll-up — Crow strips the app's own scroll ownership


dhilgaertner:
## Summary

Scrolling up in the web terminal shows **stacked duplicate copies of the Claude Code TUI** instead of a clean transcript. In a naked terminal (iTerm/Terminal.app), the same Claude Code session scrolls *within itself* — the input box stays pinned and the wheel scrolls the app's own transcript. The recent scrollback work (#804 / #821) healed stale alt-screen windows but did **not** fix this, because the duplicate-frame artifact is a consequence of Crow's core scroll model, not of degraded windows.

This is a spike to characterize the root cause against a live session and decide between two fixes.

## Root cause (analysis)

Claude Code is a **full-frame repainting TUI**. In a capable terminal it keeps itself in a fixed viewport (via the alternate screen and/or its own mouse-wheel reporting) and manages its own scrollback — that's why the input box never scrolls away and the wheel scrolls the app's transcript.

Crow deliberately does the opposite. To deliver a single unified xterm.js scrollback (the browser wheel scrolls a 50k-line buffer, `capture-pane` replays history on reconnect — CROW-606), it **strips Claude Code's scroll ownership at three layers**:

1. `Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf:56` — `set -gw alternate-screen off` keeps inner apps in the pane's **main** buffer (no alt buffer).
2. `crow-tmux.conf:70` — `terminal-overrides ',xterm*:smcup@:rmcup@,screen*:smcup@:rmcup@'` cancels the **client's** alt-screen capability too.
3. `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js:3337-3348` — `swallowMouseMode` drops the DECSET 1000-1016 mouse-tracking sequences Claude Code emits to claim the wheel.

**Why the duplicates appear:** denied a fixed screen, Claude Code's cursor-addressed full-frame repaints (which fire constantly — streaming tokens, spinner, every keystroke) land in a main buffer that is *itself scrolling*. Each repaint is deposited into the 50k scrollback as sediment. Scrolling up walks the fossil record of hundreds of past frames = "Claude Code multiple times."

`crow-tmux.conf:47-50` already acknowledges the milder version ("vim/htop/less leave their final frame in scrollback"), but underestimates a *continuously* repainting app: it's not one leftover frame, it's a duplicate per repaint.

[...]

PR #826, opened 2026-07-23
Per-surface hybrid terminal scroll model (ADR-0013)

### LINKED ISSUE #824: Implement per-surface hybrid terminal scroll model (ADR-0013, from spike #822)
Implements the decision from spike #822 (**Option B**, per-surface hybrid), recorded in **ADR-0013** (`docs/adr/0013-terminal-scroll-model.md`). Full evidence + prototype notes: `docs/spikes/822-terminal-scroll.md`.

## Goal

Agent-TUI surfaces (Claude Code, Cursor) own their own viewport + scrollback like a naked terminal (no duplicate-frame sediment); plain shell / review surfaces keep the unified xterm.js 50k scrollback + CROW-606 replay.

The spike proved the tmux side works and reproduced the sediment deterministically (`alternate-screen off` → 42 stacked frame copies, `history_size=1641`; `alternate-screen on` → 1 live frame, `history_size=0`). Prototype branches (reference only, **not for merge**): `spike/822-option-a`, `spike/822-option-b`.

## Scope

- [ ] **Daemon: classify agent-TUI windows** and set `alternate-screen on` per agent window at creation (Crow already names them, e.g. "Claude Code", and launches a known command). Plain shell / review windows keep the global default `off`.
- [ ] **`app.js`: conditional `swallowMouseMode`** — swallow DECSET 1000–1016 only on non-agent (main-buffer) surfaces; let mouse modes through on agent surfaces so the app claims the wheel.
- [ ] **`app.js`: conditional `enableWheelScroll`** — alt buffer / active mouse tracking → forward the wheel to the app (`sendScrollToPTY`); else scroll the unified xterm scrollback. (Mirror of what `enableTouchScroll` already does for touch.)
- [ ] **Client alt-buffer detection (the crux).** For `app.js` to route by `buffer.active.type === 'alternate'`, the client must also enter the alt buffer for agent windows — which the global `smcup@/rmcup@` strip in `crow-tmux.conf` currently prevents. Pick one: (a) scope that strip to non-agent surfaces, or (b) have crowd signal window-kind to `app.js` over the existing control channel and route on that instead of buffer type.
- [ ] **Alt-buffer pass over the surrounding UX:** CROW-606 replay for an agent window (restores only the current frame — confirm that's acceptable / handled), the jump-to-bottom pill, and copy/paste selection semantics on the alt buffer.
- [ ] **Tests:** extend the tmux/scrollback tests to cover per-window `alternate-screen` state and the conditional wheel/mouse routing; keep the #804/#821 `alternate_on` degraded-window detection meaningful.
- [ ] Flip **ADR-0013** from Proposed → Accepted when this lands.

## Out of scope
- Replay reflow fidelity (tracked by epic #783).

Refs: spike #822, ADR-0013, CROW-606 (#609/#612), #776/#789, #777/#786, #804/#821.

https://claude.ai/code/session_01F7kfrUnGCfrieiTGC4bQiR

dhilgaertner:
Closes #824

Implements the ADR-0013 decision from spike #822: agent-TUI surfaces own their own viewport and scrollback like a naked terminal, while plain shell / review surfaces keep the unified xterm.js 50k scrollback and CROW-606 replay.

## The bug

Scrolling up in an agent tab showed stacked duplicate copies of the Claude Code TUI. Denied a fixed viewport, an agent's full-frame repaints landed in a main buffer that was itself scrolling, so every clear-and-redraw deposited its predecessor into the 50k history as sediment.

## Measured result

Two windows side by side on one isolated server under the **bundled** conf, driven by a repainting TUI mimic:

| window | `alternate-screen` | `alternate_on` | `history_size` | stacked footer copies |
|---|:---:|:---:|:---:|:---:|
| win 0 — plain shell (global default) | 0 | 0 | **2183** | **60** |
| win 1 — agent surface | 1 | 1 | **0** | **1** |

Sediment gone on the agent surface; the shell keeps its full unified scrollback.

## The crux, resolved

ADR-0013 left one open question: how does `app.js` know a surface is an agent TUI? It offered **(a)** scope the `smcup@/rmcup@` strip to non-agent surfaces, or **(b)** signal window-kind.

**(a) is unimplementable**, verified against a live server:
- `terminal-overrides` is a **server** option matched on the **client's `TERM`**, not on a window.
- `tmux list-clients` shows **one client serving every tab** — each web surface opens a single grouped `crowd-web-*` session, and switching tabs is `select-window` within that one attach. smcup is emitted once for the whole attachment, so there is no per-window client buffer state to scope.

So `term.buffer.active.type` is **permanently `'normal'`** on the web client. This is the trap the spike's Option B prototype fell into: its `swallowMouseMode` early-return never fired, and its `mouseTrackingMode` fallback was suppressed by the very swallow it was meant to gate — a circular dependency that latched into the plain-shell path. Productionizing it as written would have shipped a no-op.

**Chosen: (b), on the existing `list-terminals` payload.** `crowd` emits a per-terminal `agent_surface` flag beside `scrollback_degraded`, sourced from the `alternate-screen` window option it actually set (read back via `#{alternate-screen}` in the same `list-windows` call), so daemon and client agree by construction with no name-matching. The `/terminal` socket was avoided deliberately: its server→client direction is binary-only through a single writer task.

## Changes

**Daemon**
- `TmuxController.setWindowOption` + `#{alternate-screen}` in `listWindowScrollback`.
- `registerTerminal(agentSurface:)` sets `alternate-screen on` per agent window *before* the launch command is pasted. Keyed on `isManaged` — **not** `agentKind` (always non-nil, falls back to a default) and **not** `trackReadiness` (false for Manager sessions, which are agent TUIs too). Re-applied on adopt so windows predating this heal when their agent restarts, without interrupting a live agent.
- `list-terminals` emits `agent_surface`.

**Client**
- `swallowMouseMode` hoisted to module scope and made conditional — agent surfaces let DECSET 1000–1016 through so the app claims the wheel.
- `enableWheelScroll` routes through a shared `appOwnsScroll()` predicate instead of swallowing; `enableTouchScroll` shares it so touch and wheel can't disagree. Forwarding reuses the existing `sendScrollToPTY`.
- `macOptionClickForcesSelection: true` — letting mouse modes through costs drag-select, and xterm.js gates the Mac force-selection path on this option, which **defaults to `false`**. Without it the ⌥-drag escape hatch silently does nothing. Surfaced as a context-menu hint.

## Regression guard (#804/#821)

`isScrollbackDegraded` treated `alternateOn` as degraded outright, so every agent window would have been badged ⚠ Recreate. It now takes `alternateScreenEnabled` — an agent surface in the alt buffer is the design working. The `history_limit` floor still catches the real casualties, which measured `5000`. Documented blind spot: an agent window genuinely wedged in the alt buffer at 50000 is no longer distinguishable from normal.

## Alt-buffer pass

- **CROW-606 replay** — correct by construction, no change. `capture-pane` on an alt pane returns the current frame, and `replayFrame` prepends `ESC[H ESC[2J ESC[3J` so reconnects rebuild rather than stack.
- **Jump-to-bottom pill** — correctly stays hidden (`viewportY == baseY == 0`); there is nothing to jump back to.
- **Copy/paste** — the one real regression, mitigated by ⌥-drag above. Plain shells unaffected.

## Drive-by

`BundledResourcesTests` asserted `set -gs alternate-screen off`, but #821 corrected the conf to `set -gw` — **that assertion has been failing on `main`** since. Fixed while editing the same line.

## Tests

- `ScrollbackHealthTests` — kind-aware truth table, including that a **non-agent** window in the alt buffer is still degraded.
- `TmuxControllerTests.agentWindowOptsIntoAlternateScreenWithoutAffectingSiblings` — real tmux: per-window `on` sticks, coexists with a sibling on the global `off`, and is not flagged degraded.
- `WebTerminalAssetTests` — client routing shape, the conditional swallow, and the selection escape hatch.
- New `web-tests/wheel-scroll.test.js` — 24 jsdom assertions over the wheel and the conditional swallow, incl. graceful degradation when `agent_surface` is absent.

`swift test` on CrowDaemon (149) and CrowEngine (325) fully green; CrowTerminal green except `retryReadinessEmitsTimedOutWhenSentinelMissing`, **confirmed failing on the clean base** with all changes stashed (the pre-existing flake #821 documented). Full JS suite green apart from a pre-existing `row.test.js` failure, likewise confirmed against an untouched `app.js`.

ADR-0013 flipped Proposed → Accepted, recording the crux resolution and why (a) was rejected.

## Out of scope

Replay reflow fidelity (epic #783).

https://claude.ai/code/session_01SEV3MkQnjbLKqr3qNzf1wZ

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Models/Terminal.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/Terminal.swift
@@ -90,3 +90,29 @@ public struct SessionTerminal: Identifiable, Codable, Sendable {
         tmuxBinding = try container.decodeIfPresent(TmuxBinding.self, forKey: .tmuxBinding)
     }
 }
+
+extension SessionTerminal {
+    /// Whether this terminal hosts a repainting agent TUI, and so takes the
+    /// alt-buffer scroll model instead of the unified scrollback (ADR-0013).
+    /// Single source of truth for that classification — the daemon uses it to
+    /// set `alternate-screen on` at window creation/adopt, and `list-terminals`
+    /// uses it for the `agent_surface` fallback, so the two can't disagree.
+    ///
+    /// TWO shapes qualify, and both are load-bearing:
+    ///   * a managed work terminal (`isManaged`), and
+    ///   * a Manager session's terminal, which always launches an agent but is
+    ///     built WITHOUT `isManaged` — `createManagerTerminal` relies on the
+    ///     memberwise default of `false`. Keying on `isManaged` alone silently
+    ///     excludes the Manager, which is itself one of the repainting agent
+    ///     windows #822 was reported against.
+    ///
+    /// Deliberately NOT `agentKind` (it always resolves to a configured
+    /// default, so it never discriminates) and NOT `trackReadiness` (false for
+    /// Manager sessions precisely because they launch the agent via `command`).
+    ///
+    /// `session` is this terminal's session; `nil` (not yet hydrated) falls
+    /// back to `isManaged` alone.
+    public func isAgentSurface(session: Session?) -> Bool {
+        isManaged || (session?.isManager ?? false)
+    }
+}
--- a/Packages/CrowCore/Sources/CrowCore/Models/Terminal.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/Terminal.swift
@@ -100,19 +100,36 @@ extension SessionTerminal {
     ///
     /// TWO shapes qualify, and both are load-bearing:
     ///   * a managed work terminal (`isManaged`), and
-    ///   * a Manager session's terminal, which always launches an agent but is
-    ///     built WITHOUT `isManaged` — `createManagerTerminal` relies on the
-    ///     memberwise default of `false`. Keying on `isManaged` alone silently
-    ///     excludes the Manager, which is itself one of the repainting agent
-    ///     windows #822 was reported against.
+    ///   * a Manager session's **agent** terminal, which launches an agent via
+    ///     `command` but is built WITHOUT `isManaged` — `createManagerTerminal`
+    ///     relies on the memberwise default of `false`. Keying on `isManaged`
+    ///     alone silently excludes the Manager, itself one of the repainting
+    ///     agent windows #822 was reported against.
+    ///
+    /// The `command != nil` half of the Manager test is what keeps this from
+    /// over-classifying. A Manager session can hold ADDITIONAL plain shells —
+    /// `new-terminal` with just a `session_id` (the `+` button, or
+    /// `crow new-terminal --session <manager-uuid>`) yields `isManaged: false`
+    /// with no command. Those are line-streaming surfaces and must keep the
+    /// unified 50k scrollback; only the terminal that actually launches the
+    /// agent gets the alt-buffer model. `createManagerTerminal` always passes a
+    /// non-nil `managerCommand(for:)`, so the two are cleanly separable.
     ///
     /// Deliberately NOT `agentKind` (it always resolves to a configured
     /// default, so it never discriminates) and NOT `trackReadiness` (false for
     /// Manager sessions precisely because they launch the agent via `command`).
     ///
+    /// Known boundary: an extra Manager-session terminal created with an
+    /// explicit `--command` is treated as an agent surface. That is a deliberate
+    /// trade — the alternative is persisting a new per-terminal field — and the
+    /// cost is only that a hand-launched full-screen program there gets the
+    /// naked-terminal scroll model instead of the unified one.
+    ///
     /// `session` is this terminal's session; `nil` (not yet hydrated) falls
     /// back to `isManaged` alone.
     public func isAgentSurface(session: Session?) -> Bool {
-        isManaged || (session?.isManager ?? false)
+        if isManaged { return true }
+        guard session?.isManager == true else { return false }
+        return command != nil
     }
 }
--- a/Packages/CrowCore/Tests/CrowCoreTests/AgentSurfaceClassificationTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AgentSurfaceClassificationTests.swift
@@ -1,0 +1,50 @@
+import Foundation
+import Testing
+@testable import CrowCore
+
+/// Which terminals take the alt-buffer scroll model (ADR-0013, #824).
+///
+/// The classification lives in one place because the daemon uses it twice — to
+/// set `alternate-screen on` at window creation/adopt, and as the
+/// `list-terminals` `agent_surface` fallback before a window exists. If those
+/// two ever disagree, the client routes the wheel one way while tmux is
+/// configured the other.
+@Suite("Agent-surface classification")
+struct AgentSurfaceClassificationTests {
+
+    private func terminal(isManaged: Bool, sessionID: UUID = UUID()) -> SessionTerminal {
+        SessionTerminal(sessionID: sessionID, name: "t", cwd: "/tmp", isManaged: isManaged)
+    }
+
+    @Test func managedWorkTerminalIsAnAgentSurface() {
+        let session = Session(name: "work", status: .active, kind: .work)
+        #expect(terminal(isManaged: true).isAgentSurface(session: session))
+    }
+
+    @Test func plainShellIsNotAnAgentSurface() {
+        // The unified 50k scrollback is the right model for line-streaming
+        // output; this is the case that must NOT regress.
+        let session = Session(name: "work", status: .active, kind: .work)
+        #expect(!terminal(isManaged: false).isAgentSurface(session: session))
+    }
+
+    /// The regression this suite exists for. `createManagerTerminal` builds its
+    /// row via `SessionTerminal(sessionID:name:cwd:command:)`, so `isManaged`
+    /// takes the memberwise default of `false` — yet the Manager runs a
+    /// full-frame repainting agent and was one of the windows #822 was reported
+    /// against. Keying the scroll model on `isManaged` alone silently left the
+    /// Manager on the sediment path.
+    @Test func managerTerminalIsAnAgentSurfaceDespiteNotBeingManaged() {
+        let session = Session(name: "Manager", status: .active, kind: .manager)
+        let managerTerminal = SessionTerminal(
+            sessionID: session.id, name: session.name, cwd: "/dev/root", command: "claude --rc")
+        #expect(!managerTerminal.isManaged, "precondition: the Manager row carries no isManaged flag")
+        #expect(managerTerminal.isAgentSurface(session: session))
+    }
+
+    @Test func unhydratedSessionFallsBackToIsManaged() {
+        // Before the session is known we can only trust the flag we have.
+        #expect(terminal(isManaged: true).isAgentSurface(session: nil))
+        #expect(!terminal(isManaged: false).isAgentSurface(session: nil))
+    }
+}
--- a/Packages/CrowCore/Tests/CrowCoreTests/AgentSurfaceClassificationTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AgentSurfaceClassificationTests.swift
@@ -42,9 +42,31 @@ struct AgentSurfaceClassificationTests {
         #expect(managerTerminal.isAgentSurface(session: session))
     }
 
+    /// The counterweight to the test above. A Manager session can hold extra
+    /// plain shells — `new-terminal` with only a `session_id` (the `+` button,
+    /// or `crow new-terminal --session <manager-uuid>`) produces
+    /// `isManaged: false` with no command. Classifying by session kind ALONE
+    /// swept those into the alt-buffer model, taking away the unified 50k
+    /// scrollback from an ordinary line-streaming shell. Only the terminal that
+    /// actually launches the agent qualifies.
+    @Test func plainShellInAManagerSessionIsNotAnAgentSurface() {
+        let session = Session(name: "Manager", status: .active, kind: .manager)
+        let shell = SessionTerminal(sessionID: session.id, name: "Shell", cwd: "/dev/root")
+        #expect(shell.command == nil, "precondition: an added shell carries no launch command")
+        #expect(!shell.isAgentSurface(session: session))
+    }
+
     @Test func unhydratedSessionFallsBackToIsManaged() {
         // Before the session is known we can only trust the flag we have.
         #expect(terminal(isManaged: true).isAgentSurface(session: nil))
         #expect(!terminal(isManaged: false).isAgentSurface(session: nil))
+    }
+
+    @Test func managerSessionKindDoesNotLeakAcrossSessions() {
+        // A work session's plain shell stays unified even if a Manager exists.
+        let work = Session(name: "work", status: .active, kind: .work)
+        let shell = SessionTerminal(sessionID: work.id, name: "Shell", cwd: "/repo", command: "vim")
+        #expect(!shell.isAgentSurface(session: work),
+                "a command alone must not promote a work-session shell")
     }
 }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -769,6 +769,12 @@ html, body {
 .ctx-item.ctx-danger { color: var(--red); }
 .ctx-item.ctx-danger:hover { background: rgba(255, 69, 58, 0.14); }
 .ctx-sep { height: 1px; background: var(--border-subtle); margin: 4px 6px; }
+/* Non-interactive footnote in the terminal context menu — tells the user how to
+   select text on an agent surface, where the app owns the mouse (ADR-0013). */
+.ctx-hint {
+  padding: 6px 10px 2px; font-size: 11px; color: var(--text-secondary);
+  border-top: 1px solid var(--border-subtle); margin-top: 4px; cursor: default;
+}
 
 /* ---- Detail header: gold PR chip + inline PR status ---- */
 .link-chip.link-pr { color: var(--gold); border-color: var(--border-strong); }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -3302,6 +3302,71 @@ function setTerminalReconnecting(on) {
 // Terminal font stack: Nerd Fonts → system monospace.
 const DEFAULT_TERM_FONT = '"MesloLGS NF", "MesloLGS Nerd Font", "JetBrainsMono Nerd Font", "Hack Nerd Font", "FiraCode Nerd Font", Menlo, Monaco, monospace';
 
+// --- Per-surface hybrid scroll model (ADR-0013) -----------------------------
+//
+// Crow runs two scroll models side by side, chosen per tmux window:
+//
+//   * PLAIN SHELL / REVIEW surfaces keep the unified model — output flows into
+//     xterm's 50k scrollback, the daemon replays tmux history on reconnect
+//     (CROW-606), and the wheel scrolls that local buffer.
+//   * AGENT-TUI surfaces (Claude Code, Cursor, Manager) own their own viewport
+//     like a naked terminal. crowd sets `alternate-screen on` for those windows
+//     so the agent's constant full-frame repaints stay in the alt buffer, which
+//     has no scrollback. Without that, every repaint deposited its predecessor
+//     into the shared history and scrolling up walked hundreds of stacked
+//     copies of the TUI (#822).
+//
+// `agent_surface` comes from `list-terminals`, sourced from the tmux window
+// option crowd actually set — NOT from `term.buffer.active.type`. The client
+// can't use the buffer type here: crow-tmux.conf strips the client's
+// smcup/rmcup, and `terminal-overrides` is keyed on the client's TERM while a
+// single tmux client serves every tab, so there is no per-window client buffer
+// state to read. The buffer/mouse-mode checks below are kept as an additional
+// signal for surfaces that do legitimately enter the alt buffer.
+function activeSurfaceIsAgent() {
+  return !!(activeTerminal && activeTerminal.agent_surface);
+}
+
+// Mouse-mode swallow — same handler as web/terminal.html (CROW-581). The agent
+// TUIs turn the xterm mouse protocol on themselves, and crow-tmux.conf's
+// `mouse off` means tmux forwards those DECSETs straight through to this
+// client.
+//
+// On a PLAIN SHELL we drop them: left alone, xterm.js enters mouse-reporting
+// mode and every mouse MOVE emits an SGR report to the PTY → the TUI repaints →
+// the viewport is yanked to the bottom while the user is scrolled up;
+// drag-select and the browser context menu are eaten too (#776).
+//
+// On an AGENT SURFACE we let them THROUGH, so the agent claims the wheel and
+// owns its own scroll exactly like a naked terminal (ADR-0013). The cost is the
+// one the swallow was avoiding: native drag-select and the right-click menu are
+// eaten inside agent windows. Hold ⌥ (Alt) to force selection anyway — that's
+// why `macOptionClickForcesSelection` is enabled in the Terminal config, since
+// xterm.js gates the Mac force-selection path on it and it defaults to false.
+const MOUSE_MODES = new Set([1000, 1001, 1002, 1003, 1005, 1006, 1015, 1016]);
+function swallowMouseMode(params) {
+  if (activeSurfaceIsAgent()) return false; // agent surface → let it own the mouse
+  const arr = params && params.params ? params.params : params;
+  const len = arr ? arr.length : 0;
+  for (let i = 0; i < len; i++) {
+    const v = arr[i];
+    if (MOUSE_MODES.has(Array.isArray(v) ? v[0] : v)) return true; // handled → drop it
+  }
+  return false; // not a mouse mode → let xterm apply it normally
+}
+
+// True when the app on the other end should receive the scroll instead of us
+// scrolling xterm's local buffer: an agent surface, an actual alternate buffer,
+// or an app that has mouse tracking on. Shared by the wheel and touch shims so
+// both agree on who owns the surface.
+function appOwnsScroll() {
+  if (activeSurfaceIsAgent()) return true;
+  const buf = term && term.buffer && term.buffer.active;
+  if (buf && buf.type === 'alternate') return true;
+  const modes = (term && term.modes) || {};
+  return !!(modes.mouseTrackingMode && modes.mouseTrackingMode !== 'none');
+}
+
 function ensureTerminal() {
   if (term) return;
   fitAddon = new FitAddon.FitAddon();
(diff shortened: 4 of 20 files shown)


dgershman (changes requested):
## Code & Security Review

### Critical Issues
None (no Red).

### Security Review
**Strengths:**
- `TmuxController.setWindowOption` / `run` use `Process` argument arrays (no shell interpolation), and call sites hardcode `"alternate-screen"` / `"on"` — no injection surface from the new API.
- `agent_surface` is a daemon-derived boolean on the existing `list-terminals` payload; clients cannot self-promote a shell tab into the agent mouse path.
- Graceful degradation when `agent_surface` is absent or `activeTerminal` is null falls back to the prior shell/swallow behavior (`wheel-scroll.test.js`).

**Concerns:**
- None security-blocking. The hybrid model intentionally widens mouse-protocol passthrough on agent surfaces (drag-select / context-menu trade documented in ADR-0013); mitigated with `macOptionClickForcesSelection` + the ⌥ hint.

### Code Quality
- **Yellow — Manager terminals are excluded from the agent scroll model despite ADR-0013.** Classification is keyed on `terminal.isManaged` in `SessionService.rehydrateTerminalSurface` / `prepareTerminal` and `EngineRouter` `new-terminal` (`agentSurface: isManaged` / `t.isManaged`). Manager terminals are created **without** `isManaged` (defaults `false`) in `createManagerTerminal` (`SessionTerminal(sessionID:name:cwd:command:)`), and existing tests document that fact (`RemoteControlRenameTargetTests`: “Manager terminals are RC-active without carrying the `isManaged` flag”). The new comments claim `isManaged` was chosen *because* Manager is an agent TUI (`trackReadiness` is false there) — but `isManaged` is also false for Manager, so:
  1. `enableAlternateScreen` is never set at register/adopt for Manager windows.
  2. `list-terminals` `agent_surface` falls back to `t.isManaged` → `false` when unbound, and to `agentSurfaces.contains(window)` (empty for Manager) when bound.
  3. Manager keeps the unified scrollback path and will still accumulate TUI sediment — the exact #822 failure mode the ADR marks Accepted to fix for “Claude Code, Cursor, **and the Manager**.”

  Fix: don’t overload `isManaged`. Use a predicate that covers managed work agents **and** Manager sessions (e.g. `isManaged || session.isManager`, or set agent surface from “has a launch command that is an agent TUI”), and mirror it in register, adopt, and the `list-terminals` fallback. Add a regression test that a Manager-shaped terminal (`isManaged: false`, manager session / non-nil agent command) gets `agentSurface: true`.

- **Green — Context-menu hint is Mac-only.** “Hold ⌥ to select text” is correct for macOS once `macOptionClickForcesSelection` is on; Win/Linux already force-select with Shift. Crow is Mac-first, so this is fine — consider mentioning Shift if the web UI is used from other platforms.
- **Green — Documented degraded-window blind spot** (agent wedged in alt buffer at history 50000) is an accepted ADR cost; `history_limit` floor still catches the real #804/#821 casualties.
- **Green — Strong test coverage** for the policy truth table, real-tmux sibling isolation, asset shape, and 24 jsdom wheel/swallow cases (verified locally: ScrollbackHealthTests 7/7; wheel-scroll 24/24).

### Summary Table

---

dhilgaertner:
Thanks @dgershman — the Yellow was correct, and it was a real bug rather than just a comment inaccuracy. Fixed in db869b4.

## Manager terminals excluded from the agent scroll model

Verified your claim directly: `createManagerTerminal` builds its row via `SessionTerminal(sessionID:name:cwd:command:)`, and `isManaged` has a memberwise default of `false`. So the Manager was excluded at all three points you listed — no `enableAlternateScreen` at register/adopt, `agent_surface: false` from the fallback, and the unified-scrollback path retained. That's the exact #822 sediment, on a window the live tmux read shows running Claude Code (`Manager`, `Manager 2`).

The comments made it worse by arguing the opposite: *"isManaged, not trackReadiness: the latter is false for Manager sessions."* True of `trackReadiness` — and equally true of `isManaged`, which is precisely what the rationale needed to rule out.

**Fix, as you suggested — stop overloading `isManaged`.** Added `SessionTerminal.isAgentSurface(session:)` in CrowCore (`isManaged || session.isManager`) and routed all four call sites through it: `registerTerminal`, the adopt path, `prepareTerminal`, and the `list-terminals` fallback. It lives in CrowCore specifically because the daemon needs the same predicate in two places — the option it sets, and the flag the client routes on — and those drifting apart is what produces a client routing the wheel one way while tmux is configured the other.

I used `session.isManager` rather than "has an agent-TUI launch command": a Manager session has exactly one terminal and it always launches an agent (`managerCommand(for:)`), so the session kind is the reliable signal, whereas sniffing the command string would be fragile against `--rc`/`--continue` variants and custom agents.

**Regression test** — `AgentSurfaceClassificationTests`, exactly the shape you asked for. `managerTerminalIsAnAgentSurfaceDespiteNotBeingManaged` builds a Manager-shaped terminal (`isManaged: false`, manager session, agent command), asserts the `!isManaged` precondition explicitly so the test can't silently stop covering the case, then asserts it classifies as an agent surface. It fails on the pre-fix predicate. Plus managed work terminals, plain shells (the must-not-regress case), and the unhydrated-session fallback.

## Green — Mac-only hint

Fair, and worth doing since the web UI is reachable from any browser (phone, Linux). The hint is now platform-aware, mirroring xterm.js's own `shouldForceSelection`: `⌥` on macOS, `Shift` elsewhere.

## Green — blind spot and coverage

No change; noted as accepted ADR costs.

Docs corrected to match in all three places that carried the wrong claim: the `registerTerminal` doc comment, the `crow-tmux.conf` scroll-model block, and ADR-0013 (which now spells out why `agentKind`, `trackReadiness`, and `isManaged`-alone each fail).

CrowCore 464, CrowDaemon 149, CrowEngine 325 green; `wheel-scroll.test.js` 24/24. CrowTerminal green apart from `retryReadinessEmitsTimedOutWhenSentinelMissing`, which I re-confirmed fails on a clean base with all changes stashed (the pre-existing flake #821 documented).

dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- `setWindowOption` goes through `Process` argv (`TmuxController.run`), not a shell string — no command-injection surface from the new API.
- `agent_surface` is a daemon-authored boolean on `list-terminals`; the context-menu hint is inserted via `textContent` (`el()`), so no XSS from the new UI.
- Classification and the tmux window option share one predicate (`SessionTerminal.isAgentSurface`), avoiding client-side name-matching or trusting UI-supplied kind.

**Concerns:**
- No security blockers. The new public `setWindowOption(name:value:)` is currently only called with hardcoded `"alternate-screen"` / `"on"`; consider an allowlist if other callers appear later (Green).

### Code Quality
- **Yellow — stale `activeTerminal` after `refreshTerminals` (`app.js` ~2035–2037).** When the same terminal id remains selected, the code keeps the old object instead of rebinding to the fresh `list-terminals` row. `agent_surface` (and `window`) are now load-bearing for wheel/mouse routing via `activeSurfaceIsAgent()` / `attachWindow(activeTerminal.window)`. A background refresh that learns the authoritative tmux flag (binding appears, adopt re-applies the option) will not update routing until a tab switch. `recreateTerminal` already works around this; `refreshTerminals` should do `activeTerminal = terminals.find(t => t.id === activeTerminal?.id) || terminals[0] || null`.
- **Yellow — selection escape-hatch discoverability (`app.js` ~3666–3675).** ADR-0013 and the PR correctly note that letting mouse modes through eats native drag-select and the right-click menu on agent surfaces. The only in-product teaching of ⌥/Shift-drag lives inside `showTerminalMenu`, gated on `activeSurfaceIsAgent()`. If DECSET mouse reporting also blocks or suppresses that `contextmenu` path, users lose both selection and the hint. Verify the custom menu still opens under live agent mouse tracking; if not, put the hint in always-visible chrome (or open the menu on a modifier-click).
- **Green — `list-terminals` shells out twice** (`degradedWindowIndices` + `agentSurfaceWindowIndices`), each calling `listWindowScrollback`. Consider one shared read.
- **Green — `web/terminal.html`** still unconditionally swallows mouse modes. Fine for the debug page (sediment is fixed server-side); worth a one-line comment so it does not drift back into “must match app.js” expectations.
- Tests look strong: kind-aware `ScrollbackHealthTests`, real-tmux coexistence, classification suite, and 24/24 `wheel-scroll.test.js` assertions. Locally verified `ScrollbackHealthTests`, `AgentSurfaceClassificationTests`, and the JS wheel suite green.

### Summary Table

---

dhilgaertner:
Thanks @dgershman — both Yellows addressed in a758e6d, plus two of the three Greens. One Green declined with reasoning below.

## Yellow 1 — stale `activeTerminal` after `refreshTerminals`

Confirmed and fixed. The guard only checked the old row was still *present*, keeping the stale object whenever the id matched:

```js
if (!activeTerminal || !terminals.find((t) => t.id === activeTerminal.id)) {
  activeTerminal = terminals[0] || null;
}
```

That was harmless while `activeTerminal` was only a selection marker. It stopped being harmless when `agent_surface` became load-bearing, and your point about it changing under a stable id is exactly right: `list-terminals` answers with the pre-binding fallback until the tmux window exists, and with the authoritative option read afterwards — and the adopt path re-applies the option on crowd restart, which is precisely a background refresh with an unchanged id. Routing stayed on the old value until a tab switch.

Now rebinds as you suggested, with a comment recording *why* the object identity matters so it doesn't get "simplified" back.

## Yellow 2 — is the hint reachable under live mouse tracking?

Good catch to question it — if the menu were suppressed, users would lose selection *and* the only teaching of the way back. **Verified it still opens**, on two independent grounds:

1. xterm.js's only `contextmenu` listener is `rightClickHandler`, and its body is `n(e,t,i), r&&s.rightClickSelect(e), t.value=s.selectionText, t.select()` — no `preventDefault()`, no `stopPropagation()`.
2. Our listener is on `#terminal-wrap`, an **ancestor** of xterm's element, so the event bubbles up regardless. And the mouse-tracking path's `preventDefault()` is on `mousedown`, which does not suppress the separately-dispatched `contextmenu` event.

So the escape hatch holds. I added jsdom coverage rather than leaving it resting on my reading of a minified bundle — hint present on an agent surface, absent on a plain shell, absent once a selection already exists (Copy is offered instead). A regression now shows up as a test failure instead of a silently undiscoverable feature.

I did *not* move the hint into always-visible chrome: it's only relevant on agent surfaces, only when the user is trying to select, and permanent chrome in a terminal pane is a real cost for a case the menu already covers. Happy to revisit if you'd rather have it unconditionally visible.

## Green — two tmux subprocesses per `list-terminals`

Fixed. Both sets derive from the same three fields of one `list-windows`, so they now come from a single `windowScrollbackClassification()` read. `degradedWindowIndices` / `agentSurfaceWindowIndices` remain as thin wrappers for single-set callers, so nothing else had to change.

## Green — `web/terminal.html` divergence

Documented in place. Worth being explicit that the divergence is *intended*: the debug page has no `agent_surface` or `activeTerminal` to route on, and the #822 sediment is fixed server-side by the per-window option, so that page benefits from the fix regardless — what it forgoes is only the wheel forwarding. Comment says "do NOT resync this with app.js" so it doesn't drift back.

## Green — allowlist on `setWindowOption` — declined

I'd rather not add one now. It takes `Process` argv (no shell), both call sites pass hardcoded `"alternate-screen"` / `"on"`, and the value never touches user input, so an allowlist today would be unreachable code guarding a threat that doesn't exist. It would also be the wrong shape if a second option ever lands, since the right constraint is per-option value validation rather than a name list. Happy to add one the moment a caller passes anything non-constant — flagging it here so the reasoning is on record.

CrowCore 464, CrowDaemon 150, CrowEngine 325 green; `wheel-scroll.test.js` now 30/30. CrowTerminal green apart from the pre-existing `retryReadinessEmitsTimedOutWhenSentinelMissing` flake.

dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- `TmuxController.setWindowOption` / `run` pass argv arrays to `Process` (no shell interpolation), and production callers only pass the fixed `alternate-screen` / `on` pair.
- `agent_surface` is a daemon-derived boolean on the existing `list-terminals` payload — not inferred from window names or client-controlled input.
- Context-menu hint uses `el(..., text)` → `textContent` (no HTML injection).
- Missing/`null` `agent_surface` correctly degrades to the plain-shell path (swallow + local scrollback).

**Concerns:**
- None material. `setWindowOption(name:value:)` is a general public API without an allowlist; fine today because callers are hardcoded — keep it that way.

### Code Quality
- **Yellow — `isAgentSurface` over-classifies every terminal in a Manager session.** `SessionTerminal.isAgentSurface` is `isManaged || (session?.isManager ?? false)` (`Packages/CrowCore/Sources/CrowCore/Models/Terminal.swift`). That correctly covers the unmanaged Manager agent, but the web `+` button calls `new-terminal` with only `session_id` (defaults to `managed: false` → a **Shell**). On a Manager session that Shell still gets `agentSurface: true` at `registerTerminal` (`EngineRouter` / `SessionService`), so it inherits `alternate-screen on`, wheel→PTY forwarding, and no unified 50k scrollback — the exact model the ADR reserves for repainting agent TUIs. The new suite asserts the Manager agent case but never asserts “Shell on Manager is not an agent surface.” Narrow the predicate (e.g. Manager **and** agent-launch terminal — `command != nil`, or a dedicated flag on `createManagerTerminal`) and add that regression test.
- **Green — Tab-switch stale mouse modes:** mitigated by existing `#673` `attachWindow` → `reloadTerminal`/`term.reset()` when the window index changes.
- **Green — Degraded-window blind spot** (agent wedged in alt buffer at history_limit=50000) is documented and accepted; history floor still catches the real #804/#821 casualties.
- **Green — Tests:** `AgentSurfaceClassificationTests`, `ScrollbackHealthTests`, and `wheel-scroll.test.js` (30) pass here. `row.test.js` still fails as the PR notes (pre-existing on `main`).

### Summary Table

---

dhilgaertner:
Thanks @dgershman — you caught an over-correction I introduced in db869b4 while fixing the opposite bug. Fixed in b6530a6.

## Yellow — `isAgentSurface` over-classifies Manager-session terminals

Confirmed and fixed. Round 1 fixed *under*-classification (the Manager's agent was excluded because its row carries no `isManaged`), and my fix over-shot by keying on session kind alone — sweeping in **every** terminal in a Manager session. Your scenario is real: `new-terminal` with just a `session_id` yields `isManaged: false` with no command, and that ordinary line-streaming shell was getting `alternate-screen on`, wheel→PTY forwarding, and no unified 50k scrollback. Precisely the model the ADR reserves for repainting agent TUIs, applied to a plain shell.

I went with your `command != nil` suggestion over a dedicated flag:

```swift
isManaged || (session.isManager && command != nil)
```

Verified the two are cleanly separable: `managerCommand(for:)` returns a non-optional `String` and `createManagerTerminal` always passes it, while an added shell reaches `SessionTerminal(command:)` as `nil` (`rawCommand = params["command"]?.stringValue`, absent → nil, and `deferLaunch` is false for an unmanaged terminal so it isn't nil'd for the other reason either).

I preferred this to a dedicated flag on `createManagerTerminal` because the classification has to be recomputable from the **persisted row** — the `list-terminals` fallback only has `SessionTerminal` + `Session`, so a construction-time-only flag wouldn't survive to where it's needed, and persisting a new field is a store migration for a case `command` already answers. I also avoided touching `isManaged` on the Manager row: several call sites key off it (`data.terminals.removeAll { $0.isManaged }`, and `RemoteControlRenameTargetTests` documents Managers as RC-active *without* the flag), so widening it would have blast radius well outside this PR.

**Accepted boundary, now documented at the predicate and in the ADR:** an extra Manager-session terminal created with an explicit `--command` is still treated as an agent surface. The cost is limited to a hand-launched full-screen program in that one spot getting the naked-terminal scroll model instead of the unified one — worth less than a persisted field, but flagging it rather than leaving it implicit.

**Tests**, both of which fail on the round-1 predicate:
- `plainShellInAManagerSessionIsNotAnAgentSurface` — the regression you asked for. Asserts the `command == nil` precondition explicitly so the test can't silently stop covering the case.
- `managerSessionKindDoesNotLeakAcrossSessions` — a command alone must not promote a *work*-session shell, pinning the other half of the conjunction.

## Greens

Nothing to change — tab-switch reset, the degraded blind spot, and the `setWindowOption` allowlist (keeping callers hardcoded, as you note) all stand as-is.

CrowCore 466, CrowDaemon 150, CrowEngine 325 green; `wheel-scroll.test.js` 30/30. CrowTerminal green apart from the pre-existing `retryReadinessEmitsTimedOutWhenSentinelMissing` flake, and `row.test.js` remains the pre-existing `main` failure.

dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- `agent_surface` is server-derived (tmux `#{alternate-screen}` or the same `isAgentSurface` predicate used at registration) — not client-supplied.
- `setWindowOption` / `enableAlternateScreen` pass fixed literals via `Process` argv (no shell interpolation).
- `sendScrollToPTY` emits fixed escape sequences capped by `MAX_PTY_SCROLL_LINES` (24); no attacker-controlled payload.
- Context-menu hint uses `textContent` with a static `⌥`/`Shift` label — no XSS surface.
- Plain shells keep the mouse-mode swallow (#776); agent surfaces are an explicit opt-in.

**Concerns:**
- No medium/high/critical security issues in the changed trust boundaries.

### Code Quality

**Yellow — `agent_surface` fallback does not run when tmux classification fails with a binding present** (`EngineRouter.swift` ~937–939)

The comment says that when tmux is unreachable, `list-terminals` falls back to `SessionTerminal.isAgentSurface(session:)`. The code only falls back when `tmuxBinding` is `nil`:

```swift
t.tmuxBinding.map { agentSurfaces.contains($0.windowIndex) }
    ?? t.isAgentSurface(session: session)
```

`windowScrollbackClassification()` returns `([], [])` on controller-nil or catch. With a binding present that becomes `agent_surface: false` for every bound agent tab (including Manager). The client then swallows mouse modes and scrolls xterm locally instead of forwarding the wheel. Empty `agentSurfaces` is also a valid success (all shells), so emptiness cannot signal failure — the read needs an explicit success/Optional so the predicate fallback can run on failure as documented.

**Yellow — sticky `mouseTrackingMode` can make a plain shell inherit agent wheel routing** (`app.js` `appOwnsScroll` ~3377–3382)

Before this PR, mouse modes were always swallowed, so `term.modes.mouseTrackingMode` stayed `'none'` on the shared xterm. Agent surfaces now let DECSET 1000–1016 through. `switchTerminal` → `attachWindow` → `reloadTerminal` → `term.reset()` clears that on the happy path (WS `OPEN`), but `attachWindow` skips reload when the socket is not open:

```javascript
if (termWs && termWs.readyState === WebSocket.OPEN) reloadTerminal();
```

After visiting an agent tab, a shell tab can still see `mouseTrackingMode !== 'none'`. Then `activeSurfaceIsAgent()` is false, yet `appOwnsScroll()` returns true via the legacy mouse-tracking branch and forwards wheel/touch to the PTY (arrow keys / SGR) instead of scrolling the unified 50k buffer — regressing the shell path this PR is careful to preserve. Under the smcup strip + conditional swallow, that branch is effectively vestigial for real shells and now harmful; gate it (or clear modes on surface change) so non-agent surfaces never inherit a prior agent tab's modes.

**Green — consider**
- Documented degraded-detection blind spot for a wedged agent window at `history_limit=50000` is an accepted ADR trade-off; floor still catches #804/#821 casualties.
- `terminal.html` deliberately keeps an unconditional swallow — comment is clear; do not “resync” with `app.js`.
- Classification predicate (`isManaged` ∨ Manager+`command`) is well tested, including the Manager under-classify and plain-shell-in-Manager over-classify traps.
- `refreshTerminals` rebinding `activeTerminal` to the fresh row is the right fix for stale `agent_surface`.
- Strong coverage: `AgentSurfaceClassificationTests`, `ScrollbackHealthTests`, real-tmux coexistence test, `WebTerminalAssetTests`, and 30/30 `wheel-scroll.test.js`. Pre-existing `row.test.js` failure confirmed unrelated.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 4 Green] findings.

---

dhilgaertner:
Thanks @dgershman — both Yellows were real fail-open holes, and both are fixed in 18ee03c. These are good catches: each one degrades *silently* into the exact behavior this PR exists to prevent.

## Yellow 1 — `agent_surface` fallback never ran on a failed tmux read

Confirmed. `windowScrollbackClassification()` returned `([], [])` for two different things — "read succeeded, no agent windows" and "read could not run" — and `list-terminals` only fell back to the predicate when `tmuxBinding` was `nil`. So with tmux down and a binding present, every bound agent tab reported `agent_surface: false`. The client would then swallow mouse modes and scroll xterm locally *while tmux has that window in the alternate buffer* — which has no scrollback, so the wheel does nothing at all. And my comment claimed the fallback covered it.

Your framing is the key point: **emptiness cannot signal failure**, because a server of nothing but plain shells is legitimately empty. The read now returns an Optional, and `list-terminals` falls back to `SessionTerminal.isAgentSurface` whenever tmux didn't answer:

```swift
"agent_surface": .bool(
    classification.flatMap { c in
        t.tmuxBinding.map { c.agentSurfaces.contains($0.windowIndex) }
    } ?? t.isAgentSurface(session: session)),
```

The two convenience accessors keep failing open with `?? []`. I kept that deliberately for `degraded` — not badging a window ⚠ on a failed read is the safe direction there, unlike `agent_surface` where `false` is an active instruction to the client.

## Yellow 2 — sticky `mouseTrackingMode` leaking onto a plain shell

Also confirmed, and the causal chain is exactly as you traced it. Before this PR the swallow was unconditional, so `mouseTrackingMode` was pinned at `'none'` on the shared xterm and that legacy branch was harmless. Letting DECSETs through on agent surfaces made it live state on a **single xterm instance shared across every tab**, and `attachWindow` only clears it when the socket is already `OPEN`:

```javascript
if (termWs && termWs.readyState === WebSocket.OPEN) reloadTerminal();
```

so a shell tab entered during a reconnect kept the previous agent tab's modes, and `appOwnsScroll()` returned true through the legacy branch despite `agent_surface: false` — wheel and touch to the PTY as arrow keys, instead of the unified 50k buffer.

Fixed by making the daemon flag authoritative **in both directions**, with the legacy signals demoted to the unclassified case only:

```javascript
if (activeTerminal && typeof activeTerminal.agent_surface === 'boolean') {
  return activeTerminal.agent_surface;
}
```

I took the gating option over clearing modes on surface change, because gating doesn't depend on reset timing — the reconnect window is precisely when `term.reset()` is skipped. And as you note the branch is vestigial for real shells anyway: under the smcup strip plus the conditional swallow, a genuine shell surface can never legitimately be in the alt buffer or be mouse-tracking, so nothing is lost.

## Tests

- `classificationReportsFailureDistinctlyFromAnEmptyResult` — a read that can't run is `nil`, while the fail-open accessors still yield `[]`. Uses its own `TmuxBackend` instance rather than `.shared` so it doesn't depend on whether the host has a live tmux server.
- Four jsdom cases splitting the two situations that used to be conflated: a **known** shell ignores a sticky `mouseTrackingMode` and a stale alt buffer, while an **unclassified** surface still honors both. The old "legacy signals route to the PTY" cases asserted the now-incorrect behavior, so they were rewritten rather than kept.
- The asset assertion now pins that the flag is read as a tri-state and can answer `false` — a future "simplification" back to a truthiness check would fail it.

CrowCore 466, CrowDaemon 150, CrowEngine 325 green; `wheel-scroll.test.js` 34/34. CrowTerminal green apart from the pre-existing `retryReadinessEmitsTimedOutWhenSentinelMissing` flake; `row.test.js` remains the pre-existing `main` failure.

dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- `agent_surface` is daemon-derived (`#{alternate-screen}` read-back, or the same `SessionTerminal.isAgentSurface` predicate used at registration) — not client-supplied or name-matched.
- `TmuxController.setWindowOption` / `run` use `Process` argv arrays (no shell interpolation); production callers only pass the fixed `alternate-screen` / `on` pair.
- `sendScrollToPTY` emits fixed escape sequences capped by `MAX_PTY_SCROLL_LINES` (24); no attacker-controlled payload.
- Context-menu hint uses `el(..., text)` → `textContent` (no XSS).
- Missing / non-boolean `agent_surface` degrades to the plain-shell path (swallow + local scrollback).

**Concerns:**
- None material. `setWindowOption(name:value:)` remains a general public API without an allowlist; acceptable while callers stay hardcoded.

### Code Quality
Prior Yellows from earlier Crow rounds are addressed on `18ee03c`:
- `windowScrollbackClassification()` returns `nil` on failure so `list-terminals` can fall back to `isAgentSurface` when tmux cannot answer (`EngineRouter.swift`).
- `appOwnsScroll()` treats a known `agent_surface` boolean as authoritative in both directions, so a plain shell cannot inherit a prior agent tab's sticky `mouseTrackingMode` (`app.js`).

**Green — consider (non-blocking):**
- Documented degraded-detection blind spot for a wedged agent window at `history_limit=50000` remains an accepted ADR trade-off; the history floor still catches #804/#821 casualties.
- Accepted classification boundary: an extra Manager-session terminal launched with an explicit `--command` is treated as an agent surface.
- `terminal.html` deliberately keeps an unconditional swallow — comment is clear; do not “resync” with `app.js`.
- Live agent windows keep their current buffer until restart after adopt re-applies the option — documented; ⚠ Recreate remains the immediate path.

**Tests / CI:** `AgentSurfaceClassificationTests` (6), `ScrollbackHealthTests` (8), and `wheel-scroll.test.js` (34/34) pass locally. CI Build & Test / Desktop / Shell Lint are green on this HEAD.

### Summary Table

---

PR #840, opened 2026-07-24
Fix OTLP /v1/logs decode (tolerant decoding + diagnostics + transport hardening)

### LINKED ISSUE #823: OTLP receiver: /v1/logs decode fails on every export ("isn't in the correct format") — Claude Code log telemetry dropped + log spam
## Symptom

`crowd` logs this continuously — roughly every 5–10 seconds, from multiple threads, for the lifetime of any active Claude Code session:

```
[OTLPReceiver] Logs decode error: The data couldn't be read because it isn't in the correct format.
```

Observed live on 2026-07-23 (`crowd[51789]`). It is **systematic** — every OTLP logs export fails, not intermittent.

## What's actually happening

Crow runs a localhost OTLP HTTP/JSON receiver (telemetry port, default 4318) that ingests Claude Code's OpenTelemetry export. Claude Code is launched with (`AgentLaunch.swift`, `ClaudeCodeAgent.swift`):

```
CLAUDE_CODE_ENABLE_TELEMETRY=1
OTEL_METRICS_EXPORTER=otlp
OTEL_LOGS_EXPORTER=otlp
OTEL_EXPORTER_OTLP_PROTOCOL=http/json
OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:<port>
```

The receiver decodes both signals with `JSONDecoder` (`OTLPReceiver.swift:117-146`):
- `/v1/metrics` → `OTLPMetricsPayload` — **works** (the analytics/cost path is fed).
- `/v1/logs` → `OTLPLogsPayload` — **throws on every request** → the event is dropped and the error above is logged (`OTLPReceiver.swift:136-145`).

The `.localizedDescription` "…isn't in the correct format." is the shared text for `DecodingError.dataCorrupted` / `.typeMismatch` / `.valueNotFound` — i.e. the body is either **not valid/complete JSON** or has a **field whose type doesn't match** `OTLPLogRecord`.

## Why it's logs-only (the diagnostic asymmetry)

Metrics and logs share the same decode path, `OTLPResource`, `OTLPAttribute`, and `OTLPAnyValue`, and metrics succeed — so the fault is specific to either the **logs payload shape** or the **logs transport**. Two concrete candidate root causes, both consistent with the exact error and the metrics-OK/logs-fail split:

1. **Transport — `HTTPParser` doesn't support chunked bodies.** `HTTPParser` (`HTTPParser.swift:59-142`, self-documented at :60-62 as "no chunked encoding") derives the body **solely from `Content-Length`**. If a logs POST arrives with `Transfer-Encoding: chunked` (or otherwise no `Content-Length`), `contentLength` defaults to `0` (:124) and the parser returns `.complete` with an **empty body** (:134) → `JSONDecoder` on empty data → *"isn't in the correct format."* Logs and metrics are exported by different exporter instances and can use different framing; logs bodies are also larger, making chunked framing more likely.
2. **Payload shape — an `OTLPLogRecord` field type mismatch.** Per the OTLP/JSON spec, enum fields (e.g. `severityNumber`) MAY be encoded as the **string enum name** while the model expects `Int?` (`OTLPModels.swift:155`); a wrong-typed value on an *optional* field still **throws** (`decodeIfPresent` only tolerates absent/null, not a type mismatch). `body`/attribute `AnyValue` may also arrive as `arrayValue`/`kvlistValue`/`bytesValue`, which `OTLPAnyValue` doesn't model (`OTLPModels.swift:17-51`).

**We can't currently tell which** — because the failure handler logs only `error.localizedDescription` and throws away the request body + headers (`OTLPReceiver.swift:143`). That undiagnosability is itself part of the bug.

## Impact

- **All Claude Code log/event telemetry is dropped.** `processLogs → database.insertEvent` (`OTLPReceiver.swift:215-254`) never runs, so the events table gets nothing — any event-based analytics/scorecard is incomplete. (Metrics/cost via `/v1/metrics` still work, which is why the analytics bar is only partly populated.)
- **Continuous log spam** (~every 5–10s per active session) drowns the daemon log — this is what surfaced it.
- Not a crash; the receiver keeps serving.

## Proposed fix (ordered)

1. **Capture the evidence first.** On decode failure, log the request `Content-Type`, `Content-Length`, `Transfer-Encoding`, `Content-Encoding`, body byte count, and a truncated/redacted prefix of the raw body. This pins cause #1 vs #2 definitively in one run. (Today only `localizedDescription` is logged.)
2. **Transport hardening in `HTTPParser`:** support `Transfer-Encoding: chunked`; when `Content-Length` is absent, do **not** silently return an empty body; branch on `Content-Type` (reject non-JSON with a clear message) and consider `Content-Encoding: gzip`.
3. **Tolerant OTLP/JSON decoding for logs:** accept `severityNumber` as number **or** string enum name; accept int64 fields as string **or** number; extend `OTLPAnyValue` for `arrayValue`/`kvlistValue`/`bytesValue`; skip an unparseable record instead of failing the whole payload.
4. **Regression test:** add a fixture captured from a real Claude Code `/v1/logs` export to `CrowTelemetryTests`. The scoped skill `Corveil/crow/Packages/CrowTelemetry:verify` can drive the receiver end-to-end (boot receiver → POST OTLP JSON via curl → inspect `telemetry.db`).

## Repro
Launch any Claude Code session under Crow with telemetry enabled (`AppConfig.telemetry.enabled`, port 4318) and watch `crowd` stderr / the `make run` terminal — the error repeats on every logs export.

## References
- Code: `Packages/CrowTelemetry/Sources/CrowTelemetry/Receiver/OTLPReceiver.swift:117-151, 215-254`; `Receiver/HTTPParser.swift:59-142`; `Models/OTLPModels.swift:17-51, 131-164`; `CrowEngine/AgentLaunch.swift` + `CrowClaude/ClaudeCodeAgent.swift` (OTEL env).
- Related (closed): #689 (metrics temporality at ingest), #745 (scorecard/telemetry). Not a duplicate — this is the logs signal.

dhilgaertner:
Closes #823

## Which candidate cause was it?

The ticket offered two: **(1) chunked transport** or **(2) payload shape**. It is **#2**, and #1 is ruled out by evidence rather than argument.

I pointed a throwaway listener at a spare port with the same `OTEL_*` env `AgentLaunch` sets and drove a real Claude Code session. The export arrives **`Content-Length` framed**, not chunked:

```
POST /v1/logs HTTP/1.1
content-type: application/json
user-agent: OTel-OTLP-Exporter-JavaScript/0.208.0
content-length: 21382
```

Corroborating: locally `telemetry.db` held **1794 metric rows and 0 event rows**. Metrics traverse the same `HTTPParser`, so bodies were always arriving intact.

## The actual cause

Claude Code stamps `event.sequence` onto **every** log record, and the OpenTelemetry JS exporter encodes JS integers as a JSON **number**:

```js
if (t === "number") {
  if (!Number.isInteger(e)) return { doubleValue: e };
  return { intValue: e };          // JSON number, not string
}
```

`OTLPAnyValue.intValue` was `String?`, so this threw `typeMismatch` and — because `JSONDecoder` is all-or-nothing — discarded the entire export. The real capture carries 31 numeric `intValue` attributes across 14 records (`input_tokens`, `duration_ms`, `cost_usd_micros`, …), so the failure was systematic, matching "every export, never intermittent".

Decoding the committed fixture with the **pre-fix** model reproduces the production string exactly:

```
localizedDescription: The data couldn't be read because it isn't in the correct format.
underlying: DecodingError.typeMismatch: expected value of type String.
  Path: resourceLogs[0].scopeLogs[0].logRecords[0].attributes[10].value.intValue
```

…and `attributes[10]` is `{"key": "event.sequence", "value": {"intValue": 0}}`.

Metrics survived because every Claude Code *metric* attribute is a string and its datapoints use `asDouble`.

## Second bug, found on the way

`event.name` carries the **bare** name on the wire (`user_prompt`); the qualified `claude_code.user_prompt` is in the body. Every reader in `TelemetryDatabase` matches the qualified form (`countEvents`, `turnAnalytics`). Fixing only the decode would have landed events that no reader matches — `promptCount`/`toolCallCount` would have stayed at 0 and `turnAnalytics` would still return `[]`. Names are now qualified at ingest, so the analytics SQL is untouched.

## What changed

**Tolerant OTLP/JSON decoding** (`OTLPModels.swift`) — `intValue` accepts number or string, as do `asInt`, `timeUnixNano`, `severityNumber` and `aggregationTemporality` (the last two also accepting their `SEVERITY_NUMBER_*` / `AGGREGATION_TEMPORALITY_*` names). `arrayValue`/`kvlistValue`/`bytesValue` flatten to JSON instead of silently becoming `""`. A malformed record is skipped rather than failing the export, with the skip counted into `OTLPDecodeDiagnostics` so the loss is logged rather than silent.

**Diagnostics** (`OTLPReceiver.swift`) — failures now log the `DecodingError` case, **coding path** and reason plus framing headers and body size. The old `localizedDescription` was the same sentence for every possible cause, which is what made this undiagnosable. A truncated raw body is gated behind `CROW_TELEMETRY_LOG_RAW_BODIES=1`, since bodies carry `user.email`, `user.id` and `organization.id`. Repeat failures are throttled to one report per minute per path — the spam half of the ticket.

**Transport hardening** (`HTTPParser.swift`) — not the cause, but the parser derived the body from `Content-Length` alone and returned a *complete* request with an **empty body** when absent, which would have surfaced as this same opaque error. It now supports `Transfer-Encoding: chunked`, rejects an unframed POST by name, and rejects a non-JSON `Content-Type` or compressed body with a specific message. Per discussion, gzip *decoding* is deliberately out of scope (Crow never sets `OTEL_EXPORTER_OTLP_COMPRESSION`); it is rejected clearly instead.

## Verification

- 66 tests in `CrowTelemetryTests` pass (33 new). The fixture is a real Claude Code 2.1.219 capture with identifiers replaced and **every JSON type preserved** — a test asserts the numeric `intValue` fields are still numeric, so a future re-capture can't silently stop covering the bug.
- End-to-end through the package's `verify` skill: 14 events land with `claude_code.*` names over **both** Content-Length and chunked framing; metrics still ingest; rejections return their specific messages.

```
claude_code.mcp_server_connection|7   claude_code.api_request|2
claude_code.assistant_response|2      claude_code.user_prompt|1
claude_code.plugin_loaded|1           claude_code.permission_mode_changed|1
```

- `make test` is green except `CrowTerminal`'s `retryReadinessEmitsTimedOutWhenSentinelMissing`, which fails identically on a clean tree — pre-existing and unrelated (CrowTerminal does not depend on CrowTelemetry).

https://claude.ai/code/session_01FEernNSx3DF5xb3a7jGfSd

### DIFF
--- a/Packages/CrowTelemetry/.claude/skills/verify/SKILL.md
+++ b/Packages/CrowTelemetry/.claude/skills/verify/SKILL.md
@@ -29,12 +29,25 @@ run in the background, then drive it.
 Gotchas:
 - Ingest is OTLP **HTTP/JSON only** (`OTEL_EXPORTER_OTLP_PROTOCOL=http/json`),
   paths `/v1/metrics` and `/v1/logs`, POST only. One request per connection
-  (the receiver closes it after responding).
+  (the receiver closes it after responding). `Content-Length` and
+  `Transfer-Encoding: chunked` both work; a compressed body or a non-JSON
+  `Content-Type` is rejected by name.
 - The resource must carry `crow.session.id` (a UUID string) in
   `resource.attributes`, or the payload is silently skipped.
-- Datapoint values: `asDouble` (number) or `asInt` (**string**, e.g. `"100"`).
-- Sum metrics take `aggregationTemporality` (1=delta, 2=cumulative) and
-  `isMonotonic`; cumulative sums are normalized to deltas at insert.
+- Datapoint values: `asDouble` (number) or `asInt` (string **or** number).
+- Sum metrics take `aggregationTemporality` (1=delta, 2=cumulative, or the
+  `AGGREGATION_TEMPORALITY_*` name) and `isMonotonic`; cumulative sums are
+  normalized to deltas at insert.
+- Decoding is deliberately tolerant (#823): scalars are accepted in either
+  their number or string form, and a malformed record is skipped with a
+  `skipped malformed N logRecords` log line rather than failing the export.
+- Event names arrive **bare** in the `event.name` attribute (`user_prompt`);
+  they are qualified to `claude_code.user_prompt` at ingest, which is the form
+  `countEvents` and `turnAnalytics` match on.
+- Decode failures log the `DecodingError` coding path, which names the
+  offending field. Set `CROW_TELEMETRY_LOG_RAW_BODIES=1` to also log a
+  truncated raw body — off by default because bodies carry account IDs and
+  emails. Repeat failures are throttled to one report per minute per path.
 
 Minimal payload:
 
--- a/Packages/CrowTelemetry/.claude/skills/verify/SKILL.md
+++ b/Packages/CrowTelemetry/.claude/skills/verify/SKILL.md
@@ -32,6 +32,14 @@ Gotchas:
   (the receiver closes it after responding). `Content-Length` and
   `Transfer-Encoding: chunked` both work; a compressed body or a non-JSON
   `Content-Type` is rejected by name.
+- `Content-Type: application/json` is **required** on POST — the receiver fails
+  closed. Pass it explicitly with curl, which otherwise sends
+  `application/x-www-form-urlencoded` for `-d` and gets a 400:
+
+  ```
+  curl -X POST http://localhost:<port>/v1/logs \
+       -H 'Content-Type: application/json' --data-binary @payload.json
+  ```
 - The resource must carry `crow.session.id` (a UUID string) in
   `resource.attributes`, or the payload is silently skipped.
 - Datapoint values: `asDouble` (number) or `asInt` (string **or** number).
--- a/Packages/CrowTelemetry/Package.swift
+++ b/Packages/CrowTelemetry/Package.swift
@@ -17,7 +17,8 @@ let package = Package(
         ),
         .testTarget(
             name: "CrowTelemetryTests",
-            dependencies: ["CrowTelemetry"]
+            dependencies: ["CrowTelemetry"],
+            resources: [.copy("Fixtures")]
         ),
     ]
 )
--- a/Packages/CrowTelemetry/Sources/CrowTelemetry/Models/OTLPModels.swift
+++ b/Packages/CrowTelemetry/Sources/CrowTelemetry/Models/OTLPModels.swift
@@ -515,6 +515,12 @@ public struct OTLPLogRecord: Codable, Sendable {
     /// (`user_prompt`) while the fully-qualified `claude_code.user_prompt` sits
     /// in the body. Readers in `TelemetryDatabase` match the qualified form, so
     /// unqualified names are prefixed here at ingest.
+    ///
+    /// The `claude_code.` prefix assumes a single producer, which holds because
+    /// `AgentLaunch` exports the `OTEL_*` env only for `agent.kind ==
+    /// .claudeCode` — no other harness reaches this receiver. If a second agent
+    /// ever exports here, this needs a source-aware prefix (derived from the
+    /// resource's `service.name`) rather than a constant.
     public var resolvedEventName: String? {
         let candidate = eventName
             ?? attributes.flatMap { extractAttribute("event.name", from: $0) }
--- a/Packages/CrowTelemetry/Sources/CrowTelemetry/Models/OTLPModels.swift
+++ b/Packages/CrowTelemetry/Sources/CrowTelemetry/Models/OTLPModels.swift
@@ -327,6 +327,17 @@ public struct OTLPMetricsPayload: Codable, Sendable {
 public struct ResourceMetrics: Codable, Sendable {
     public let resource: OTLPResource?
     public let scopeMetrics: [ScopeMetrics]?
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        resource = try? container.decodeIfPresent(OTLPResource.self, forKey: .resource)
+        scopeMetrics = container.otlpLenientArray(
+            ScopeMetrics.self,
+            forKey: .scopeMetrics,
+            label: "scopeMetrics",
+            diagnostics: decoder.otlpDiagnostics
+        )
+    }
 }
 
 /// Metrics from a single instrumentation scope.
@@ -456,6 +467,17 @@ public struct OTLPLogsPayload: Codable, Sendable {
 public struct ResourceLogs: Codable, Sendable {
     public let resource: OTLPResource?
     public let scopeLogs: [ScopeLogs]?
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        resource = try? container.decodeIfPresent(OTLPResource.self, forKey: .resource)
+        scopeLogs = container.otlpLenientArray(
+            ScopeLogs.self,
+            forKey: .scopeLogs,
+            label: "scopeLogs",
+            diagnostics: decoder.otlpDiagnostics
+        )
+    }
 }
 
 /// Log records from a single instrumentation scope.
--- a/Packages/CrowTelemetry/Sources/CrowTelemetry/Models/OTLPModels.swift
+++ b/Packages/CrowTelemetry/Sources/CrowTelemetry/Models/OTLPModels.swift
@@ -1,5 +1,178 @@
 import Foundation
 
+// MARK: - Decode Diagnostics
+
+/// Collects non-fatal decode losses (records/attributes skipped) for one decode.
+///
+/// Tolerant decoding deliberately drops malformed elements instead of failing
+/// the whole export — this box is how that silent loss gets reported. Pass one
+/// in via `JSONDecoder.userInfo[.otlpDiagnostics]`.
+public final class OTLPDecodeDiagnostics: @unchecked Sendable {
+    private let lock = NSLock()
+    private var counts: [String: Int] = [:]
+
+    public init() {}
+
+    /// Number of elements skipped under `label` (e.g. `"logRecords"`).
+    public func skipped(_ label: String) -> Int {
+        lock.lock()
+        defer { lock.unlock() }
+        return counts[label] ?? 0
+    }
+
+    /// Human-readable summary, or `nil` when nothing was skipped.
+    public var summary: String? {
+        lock.lock()
+        defer { lock.unlock() }
+        guard !counts.isEmpty else { return nil }
+        return counts.sorted { $0.key < $1.key }
+            .map { "\($0.value) \($0.key)" }
+            .joined(separator: ", ")
+    }
+
+    fileprivate func recordSkip(_ label: String) {
+        lock.lock()
+        defer { lock.unlock() }
+        counts[label, default: 0] += 1
+    }
+}
+
+public extension CodingUserInfoKey {
+    /// Key for an `OTLPDecodeDiagnostics` collecting skipped elements.
+    static let otlpDiagnostics = CodingUserInfoKey(rawValue: "com.corveil.crow.otlp.diagnostics")!
+}
+
+// MARK: - Tolerant Decoding Helpers
+
+/// Consumes any JSON value without inspecting it.
+///
+/// Used to advance an unkeyed container past an element that failed to decode —
+/// `JSONDecoder` does not advance `currentIndex` when `decode` throws.
+private struct SkippedValue: Decodable {
+    init(from decoder: Decoder) throws {}
+}
+
+/// Maps an OTLP/JSON enum name to its number, e.g. `SEVERITY_NUMBER_INFO` → 9.
+///
+/// The protobuf JSON mapping allows enums to travel as either the number or the
+/// symbolic name; Claude Code sends numbers, other SDKs send names.
+private func otlpEnumValue(_ name: String, prefix: String, cases: [String: Int]) -> Int? {
+    var bare = name.uppercased()
+    if bare.hasPrefix(prefix) { bare.removeFirst(prefix.count) }
+    if let exact = cases[bare] { return exact }
+    // Severity names carry a 1-4 suffix within each band: DEBUG2 = DEBUG + 1.
+    guard let last = bare.last, let offset = last.wholeNumberValue, (2...4).contains(offset) else {
+        return nil
+    }
+    guard let base = cases[String(bare.dropLast())] else { return nil }
+    return base + offset - 1
+}
+
+private let severityNumberCases: [String: Int] = [
+    "UNSPECIFIED": 0, "TRACE": 1, "DEBUG": 5, "INFO": 9,
+    "WARN": 13, "ERROR": 17, "FATAL": 21,
+]
+
+private let aggregationTemporalityCases: [String: Int] = [
+    "UNSPECIFIED": 0, "DELTA": 1, "CUMULATIVE": 2,
+]
+
+extension KeyedDecodingContainer {
+    /// Whether `key` is present and not JSON `null`.
+    private func hasValue(_ key: Key) -> Bool {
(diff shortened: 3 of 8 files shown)


dgershman (changes requested):
## Code & Security Review

Solid root-cause fix for #823: tolerant `intValue` decoding, per-record skip + diagnostics, and transport hardening that stops empty-body false completions. 66 `CrowTelemetryTests` passed locally. Fixture identifiers are sanitized (`telemetry-fixture@example.com`). Raw-body logging correctly gated behind `CROW_TELEMETRY_LOG_RAW_BODIES`.

### Critical Issues (if any)
None (no Red).

### Security Review
**Strengths:**
- Listener remains `acceptLocalOnly` / loopback-only (`OTLPReceiver.swift`)
- 64 MiB body cap + 8 KiB header cap; gzip/`Content-Encoding` and non-JSON types rejected by name
- Decode failure logs use coding path; PII-bearing body prefix is opt-in and throttled (`OTLPReceiver.swift:174-199`)

**Concerns:**
- **Yellow** — `HTTPResponse.badRequest` builds JSON via string interpolation (`HTTPParser.swift:32-34`). New parser errors embed request-controlled values (`Content-Length`, `Content-Type`, chunk size, method/path) into that message (`HTTPParser.swift:147`, `172`, `179`, `211`, `163`). A quote/backslash in those fields yields malformed or injected JSON. Escape with `JSONSerialization` (or equivalent) before embedding.
- **Yellow** — Chunked size check uses `body.count + size <= maxBodyBytes` (`HTTPParser.swift:228-230`). A huge hex chunk size can overflow `Int` addition and trap (process crash / DoS on the receiver queue). Check `size <= maxBodyBytes` and `body.count <= maxBodyBytes - size` (or saturating/checked arithmetic) before appending.

### Code Quality
- **Green** — `dechunk` advances past chunk data without verifying the trailing CRLF (`HTTPParser.swift:235-237`). Malformed framing can desync; validate `CRLF` before continuing.
- **Green** — `ResourceMetrics` / `ResourceLogs` still synthesize `scopeMetrics` / `scopeLogs` arrays; one bad scope drops the whole resource. Consider `otlpLenientArray` at that layer too for consistency with `logRecords` / `metrics`.
- Tests for numeric `intValue`, event-name qualification, framing rejections, and throttle behavior are thorough and match the claimed failure mode.

### Summary Table

---

dhilgaertner:
Thanks @dgershman — all four findings were real and are fixed in c31cb6e.

**Yellow — JSON injection in error bodies.** Fixed. `HTTPResponse.badRequest`/`notFound` now build the body with `JSONSerialization` instead of interpolation. I used `.withoutEscapingSlashes` so paths stay readable (`OTLP/JSON`, not `OTLP\/JSON`). Verified against a live receiver with a hostile header:

```
$ curl -X POST localhost:44320/v1/logs -H 'Content-Type: application/"evil\' -d '{}'
{"error":"Unsupported Content-Type: application/\"evil\\ (OTLP/JSON only)"}
```

…which round-trips through `json.loads` cleanly. I also added a 120-char cap on quoted request-controlled values — headers can be ~8 KiB under the existing header cap, and there was no reason to echo all of that into the response or the log line.

**Yellow — chunked size overflow.** Confirmed and fixed. Worth noting the trap needs a *non-empty* buffer: with nothing buffered, `0 + Int.max` doesn't overflow. Feeding `4\r\ntest\r\n7FFFFFFFFFFFFFFF\r\n` to the previous parser dies with SIGTRAP (exit 133) — a remotely-triggerable crash of the receiver queue, so your read was right. Now checked against the remaining budget (`size <= max`, `body.count <= max - size`); the same input returns a clean 400 and the receiver keeps serving. My first version of the regression test used an empty buffer and wouldn't have caught it — the committed one has the preceding chunk.

**Green — chunk terminator.** Fixed; `dechunk` validates the trailing CRLF before advancing, so a size that disagrees with the framing errors instead of silently desyncing.

**Green — lenient scope arrays.** Done for both `ResourceLogs` and `ResourceMetrics`, with skips counted into `OTLPDecodeDiagnostics` like the other layers.

75 tests pass (9 new). The overflow test is a real guard — on the previous code it crashes the test process rather than failing. Re-verified end-to-end: the fixture still ingests 14 events, and the receiver survives the overflow input.

The one red herring in `make test` is `CrowTerminal`'s `retryReadinessEmitsTimedOutWhenSentinelMissing`, which fails identically on a clean tree — pre-existing and unrelated to this package.

dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Listener stays loopback-only (`acceptLocalOnly = true`) — appropriate for local OTLP ingest.
- Raw body logging is opt-in via `CROW_TELEMETRY_LOG_RAW_BODIES=1`, with an explicit note that bodies can carry `user.email` / account IDs (`OTLPReceiver.swift:186-193`).
- Error responses use `JSONSerialization` so request-controlled header/path values cannot break out of the JSON string (`HTTPParser.swift:45-52`); values are also truncated (`quoted`, limit 120).
- Body size caps exist for declared `Content-Length` and reassembled chunked payloads (64 MiB), with an overflow-safe remaining-budget check for huge chunk sizes (`HTTPParser.swift:249-254`).
- Fixture identifiers/emails are sanitized (`telemetry-fixture@example.com`, zeroed UUIDs).

**Concerns:**
- **Yellow — Unbounded receive buffer on incomplete chunked bodies** (`HTTPParser.swift` / `OTLPReceiver.swift:95-101`): For `Transfer-Encoding: chunked`, `dechunk` returns `.needsMore` when no CRLF has arrived yet (`HTTPParser.swift:222-224`). `OTLPReceiver.receiveData` then appends indefinitely with no ceiling on `accumulated`. A local client can OOM Crow by sending chunked headers and then a CRLF-free body stream. Content-Length is capped; chunked is not until a size line parses. Cap `accumulated.count` (e.g. `maxBodyBytes` + header allowance) in the `.needsMore` path and/or reject in `dechunk` when remaining wire bytes already exceed the budget, and add a regression test.

### Code Quality
- Root-cause fix is sound: tolerant `intValue` (number|string), per-record skip with `OTLPDecodeDiagnostics`, and `resolvedEventName` qualification so analytics SQL keeps matching.
- Diagnostics upgrade (`DecodingError` case + coding path, failure throttle) directly addresses the undiagnosable log spam in #823.
- Transport hardening (reject unframed POST, non-JSON Content-Type, compressed bodies; validate chunk terminators) is well motivated and well tested.
- **Green — Absent `Content-Type`:** `rejectUnsupportedEncoding` only rejects when the header is present and wrong (`HTTPParser.swift:195-202`). Requiring `application/json` (or `*+json`) for body methods would fail closed more consistently.
- **Green — `claude_code.` prefix:** `resolvedEventName` always qualifies bare names (`OTLPModels.swift:518-524`). Fine for Claude Code today; may need a source-aware prefix if other agents share this receiver later.

### Testing
`swift test --package-path Packages/CrowTelemetry` — **75 tests passed** (including new HTTPParser + OTLP logs decoding coverage). Pre-existing XCTest bundle architecture warning on this host did not prevent Swift Testing from completing green.

### Summary Table

---

dhilgaertner:
Fixed in d06f483. Thanks — the Yellow was worse than the write-up suggested, and finding it turned up a sibling case.

**Yellow — unbounded receive buffer.** Confirmed, and I measured it rather than reasoning about it. Against the previous build, 250 MiB of CRLF-free chunked data took the receiver from 8 MB to **15.96 GB RSS** — the repeated `Data` append/copy in the recursive `.needsMore` path amplifies it well past linear. That's an OOM of the whole machine from an unprivileged local client, not just a stalled connection.

Three guards now, outermost first:

1. **`OTLPReceiver` refuses any connection that buffers past `maxRequestBytes`** (body cap + header allowance) while still incomplete. I put the primary stop here deliberately: the parser can only reject what it can already frame, so bounding it at the buffer's owner covers *every* `.needsMore` path rather than the one we happened to find.
2. `dechunk` errors once an unterminated chunk-size line passes 1 KiB — the specific case you identified.
3. `dechunk` errors once unterminated **trailers** pass the header cap. The zero-chunk branch had the same shape (`range(of: CRLFCRLF)` → `.needsMore` forever); it isn't in the finding but it's the same bug.

Same attack against the fixed build:

```
pid=64412  RSS before: 8304 KB
server closed the connection after 1.0 MiB: BrokenPipeError
response: {"error":"Malformed chunk size line"}
RSS after:  9888 KB
receiver still serving -> 200      (fixture still ingests 14 events)
```

**Green — absent `Content-Type`.** Adopted; it's now required on body methods, so a client not speaking OTLP/JSON fails closed. One consequence worth flagging: this rejects `curl -d '{...}'` without an explicit header, since curl defaults to `application/x-www-form-urlencoded`. That's a real ergonomic cost on a localhost debug receiver, so I've documented the flag in the `verify` skill rather than leaving people to hit a confusing 400.

**Green — `claude_code.` prefix.** Left as a constant, but I've written the assumption down instead of leaving it implicit: `AgentLaunch` exports the `OTEL_*` env only for `agent.kind == .claudeCode`, so this receiver has exactly one producer today. The comment names `service.name` as the discriminator to switch to if that changes. Building the source-aware version now would be untestable generality — there's no second producer to derive a prefix from.

81 tests (6 new). Note the branch picked up a `main` merge from your side mid-review; I rebased onto it and re-ran green.

dgershman (approved):
## Code & Security Review

Prior request-changes rounds (JSON error-body escaping, chunk-size overflow trap, unbounded chunked receive buffer, missing Content-Type fail-closed) are addressed in the follow-up commits. Re-reviewed the full CrowTelemetry diff on the current tip.

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Listener remains loopback-only (`acceptLocalOnly = true`, `OTLPReceiver.swift`).
- Framing defenses are solid: 64 MiB body / 8 KiB header caps, overflow-safe chunk budget (`HTTPParser.swift:275-280`), CRLF-free chunk-size and trailer caps, and a `maxRequestBytes` backstop in `receiveData` (`OTLPReceiver.swift:99-105`).
- Request-controlled values in 400 bodies go through `JSONSerialization` and `quoted` truncation (`HTTPParser.swift:45-52`, `197-199`).
- Raw body logging stays opt-in (`CROW_TELEMETRY_LOG_RAW_BODIES=1`) with throttle; fixture PII is sanitized (`telemetry-fixture@example.com`, zeroed IDs).
- Non-JSON `Content-Type` and compressed bodies are rejected by name; `Content-Type` is now required on body methods.

**Concerns:**
None that block. Residual notes are Green only (below).

### Code Quality
- Root cause fix is correct and well evidenced: numeric `intValue` tolerance, per-element `otlpLenientArray` + `OTLPDecodeDiagnostics`, and `resolvedEventName` qualification so analytics SQL keeps matching.
- Diagnostics (`DecodingError` path + `FailureThrottle`) directly fix the undiagnosable spam in #823.
- **Green** — `Transfer-Encoding` that is present but not `chunked`, paired with `Content-Length`, still accepts the length-framed body (`HTTPParser.swift:157-182`). Harmless for Claude Code’s exporter; slightly less fail-closed than the `Content-Encoding` path if an exotic TE ever appears.
- **Green** — `resolvedEventName` still hardcodes the `claude_code.` prefix (`OTLPModels.swift:519-530`). Documented and safe while only Claude Code reaches this receiver; revisit if a second harness shares the socket.
- **Green** — A near-cap chunked body can hit the wire `maxRequestBytes` backstop before reassembly finishes (chunk framing overhead). Irrelevant at real Claude Code export sizes (~tens of KB).

### Testing
- Local: `swift test --package-path Packages/CrowTelemetry` — **81 tests passed** (Swift Testing). Host still prints the pre-existing XCTest bundle arch warning; it did not affect results.
- CI: Build & Test, Desktop (Tauri) Build, and Shell Lint all pass on this PR.

### Summary Table

---

PR #832, opened 2026-07-23
docs: harness capability gap audit — re-verify #827 pins against current upstream (#828)

dgershman:
Closes #828. Baseline: #827 (capability matrix + capability-tiers ADR — still open; its matrix + version-pinned "why"s are the checklist).

## What
Adds `docs/agent-harness-gap-audit.md` and re-checks every version-pinned reason in the #827 tiers-ADR seed against the **currently installed** upstream CLIs — Cursor `2026.07.23`, Codex `0.141.0`, OpenCode `1.17.10` — via each binary's own `--help`/`--version` plus upstream docs/changelogs.

## Findings (of 7 genuine gaps)
**5 now available upstream:**
- **Resume/continue** — Cursor `--continue`/`--resume`, Codex `resume --last`, OpenCode `--continue`/`--session` (restores history)
- **Auto-permission** — Cursor `--force`/`--auto-review`/`--trust`, Codex `-a never` / `--dangerously-bypass-approvals-and-sandbox`; OpenCode's `--auto` probe is now stale (`run --dangerously-skip-permissions`)
- **MCP** — `cursor-agent mcp`, `codex mcp`, `opencode mcp` (was Claude-only)
- **Per-project hooks** — `.cursor/hooks.json`, `.codex/hooks.json`, `.opencode/plugins/` → closes the shared-`cwd` session-resolution collision on all three non-Claude harnesses
- **Codex review** — `codex review --base <branch>` (was `nil` / "Phase C, Claude-only")

**1 partial:** native remote control now exists (Codex experimental `remote-control` + `--remote`; OpenCode `serve`/`attach`/`acp`) but is heavier than the working `crow send` stdin paste — no user-facing capability gained today.

**1 still absent:** Codex async hooks — `async:true` is still parsed-but-skipped at 0.141.0 (pin holds; only the version reference `v0.139.0`→`v0.141.0` is stale).

## Spin-off closure tickets
- #829 — Cursor closures
- #830 — Codex closures
- #831 — OpenCode closures
- Deferred (no ticket): Codex async hooks; Cursor/OpenCode native RC.

## Notes
- Doc-only; **no runtime behavior change.**
- Linked from `docs/architecture.md`.
- §5 records the exact matrix + tiers-ADR corrections to apply once #827 merges (nothing in-repo to edit until then).

## Test plan
- [x] `docs/agent-harness-gap-audit.md` exists; every #827 version-pinned reason re-verified (§2).
[...]

Issue #828, opened 2026-07-23
Audit harness adapters for upstream features that can close matrix gaps


dgershman:
Follow-up to #827 (blocked by it — needs the capability matrix + capability-tiers ADR as the baseline).

## Summary
The Cursor/Codex/OpenCode adapters were written against specific upstream CLI versions — several gaps are pinned to a moment in time (e.g. "Codex hooks sync-only as of v0.139.0", "no `--continue` in MVP"). Upstream CLIs move fast. Using the capability matrix + capability-tiers ADR from #827 as the baseline, re-check whether those ❌ / ⚠️ cells can now be closed because the harness gained the feature after its adapter was written.

## Task — per harness (`agent`/Cursor, `codex`, `opencode`), re-check current upstream for:
- **Resume / continue** equivalent → flips the Resume cell (currently ❌ for Cursor/Codex, partial for OpenCode).
- **Non-interactive / auto-approve** flag → Auto-permission (would drop OpenCode's `--help` probe hack; currently ignored by Cursor/Codex).
- **Per-project / per-session hook config** (not global) → per-session scope + fixes the shared-`cwd` collision that affects all three non-Claude harnesses.
- **Async hooks** (Codex specifically) → hook-async cell (pinned "sync-only as of v0.139.0").
- **Review / print / headless mode** enabling `/crow-review-pr` on Codex → Review cell (currently returns `nil`, "Phase C").
- **Native remote-control mechanism** → RC cell (retire the `crow send` stdin fake; Codex is `supportsRemoteControl=false`).
- **MCP support** → MCP parity beyond Claude (currently Claude-only via `~/.claude.json`).
- **Proper prompt injection** so `CursorLauncher` can finally be auto-wired (today `.work` drops into a bare TUI with no generated prompt).

## Method
Inspect each harness's `--help`, changelog, and docs; note the version a feature landed in. Treat each version-pinned reason recorded in the capability-tiers ADR (#827) as a checklist item.

## Deliverable
`docs/agent-harness-gap-audit.md` (or issue comments) listing, per harness × dimension: **still absent** vs **now available** (upstream link/flag + min version) + recommended closing approach. Open spin-off implementation tickets for the worthwhile closures. Update the matrix doc + capability-tiers ADR wherever a documented "why" no longer holds.

## Acceptance criteria
- [ ] Every version-pinned reason in the capability-tiers ADR re-verified against current upstream.
[...]

PR #841, opened 2026-07-24
OpenCode adapter: Jira MCP, per-project plugins, version-aware auto-probe (#831)

### LINKED ISSUE #831: OpenCode adapter: resume-with-history, retire --help probe, MCP, per-project plugins
Spin-off from the harness gap audit (#828); baseline #827. See `docs/agent-harness-gap-audit.md` §3c.

`opencode` 1.17.10 closes the CROW-545/547 MVP caveats. Close in one adapter pass:

- **Resume with history** — replace the bare-`opencode` resume fallback in `OpenCodeLaunchArgs.resumeTUICommand` with `opencode --continue` (or `--session <id>`). Removes the "TUI re-entry, no history" caveat.
- **Retire the `--help` probe hack** — `opencode run --dangerously-skip-permissions` is stable in 1.17.x and the top-level TUI `--auto` flag is gone, so `tuiSupportsAuto`'s `--auto` probe never matches on ≥1.17. Hard-code `--dangerously-skip-permissions` on the headless `run` step (where auto-approve already applies) and drop the subprocess probe + cache in `OpenCodeLaunchArgs`.
- **MCP** — register Jira MCP via `opencode mcp`; parity with Claude.
- **Per-project plugins** — install `crow-hooks.js` into the worktree's `.opencode/plugin/` (project scope) with the session UUID baked in, instead of the global `~/.config/opencode/plugin/`. Closes the shared-`cwd` collision.

**Not in scope:** native RC via `opencode serve`/`attach`/`acp` — heavier than the working `crow send` paste; no user-facing capability gained today. Note ACP as the strategic option for structured driving later.

Upstream: https://opencode.ai/docs/plugins/ · https://opencode.ai/docs/config/ · https://opencode.ai/docs/cli/

dgershman:
Closes #831

Closes the CROW-545/547 OpenCode MVP caveats now that `opencode` 1.17.x is stable, per the #828 harness gap audit (§3c). Scope was trimmed by the ticket review — resume-with-history was **already shipped** (#547) — so this pass covers the three remaining closures.

## Per-project plugins (session UUID baked in)
`OpenCodeHookConfigWriter.writeHookConfig` (was a no-op) now installs `<worktree>/.opencode/plugins/crow-hooks.js` with the Crow **session UUID baked in**, so each worktree emits `crow hook-event --session <uuid> …` — exact resolution, no `cwd` matching. That **closes the shared-`cwd` collision** two sessions rooted at (or resolving to) the same path could otherwise hit.

The global `~/.config/opencode/plugins/` plugin stays as a fallback for hand-started sessions, but **self-suppresses** (returns no hooks) whenever a per-project plugin exists in the cwd — so the two never double-emit (OpenCode dedups plugins by file URL, `plugin.ts@v1.18.4`, so both files load). Verified under Bun: global defers when the per-project file is present, emits when not; per-project is always authoritative.

## Jira MCP registration (parity with Claude)
New `OpenCodeMCPConfigWriter` mirrors the user's global `jira` MCP from `~/.claude.json` (`mcpServers.jira` — the same server Claude Code sessions inherit post-CROW-528) into a Crow-owned `~/.config/opencode/opencode.json` under `mcp.jira`, translating Claude's local/remote shape into OpenCode's. Writing `opencode.json` (which OpenCode merges alongside the user's hand-edited `opencode.jsonc`, `config.ts@v1.18.4:258-260`) means we never strip the user's comments. Merge-safe, idempotent, and a **no-op when there's no Claude Jira MCP** to mirror. `OpenCodeLauncher`'s Jira prompt now prefers `jira_get_issue` via the MCP with an `acli` fallback.

## Version-aware auto-permission probe (narrow, don't delete)
Per the #832 round-1 correction: the TUI `--auto` flag was gone only in `v1.17.x` and re-added in `v1.18.0`, so `runAutoApproveSuffix`'s try-`--auto`/else-`--dangerously-skip-permissions` fallback **stays**. Only `tuiSupportsAuto`'s top-level `--auto --help` probe is dead weight — and only on `<1.18`. It's now version-gated: below 1.18 we answer "no auto" from the `opencode --version` string alone (skipping the subprocess); on `≥1.18` or an unparseable version we still run the `--help` probe, so a future upstream flip is caught without a code change.

## Not in scope (unchanged)
Native RC via `opencode serve`/`attach`/`acp`; `--session <id>` targeting and `.work` resume (both optional per the audit).

## Test plan
- `swift test --package-path Packages/CrowOpenCode` — 46 tests green (new coverage: per-project write/remove, self-suppress guard, MCP translation/merge/idempotency/unparseable-refusal, SemVer parse + `<1.18` narrowing).
- Bun-parsed both generated plugin variants and drove the self-suppress guard end-to-end (global defers with per-project present; emits without; per-project authoritative).
- Re-probed the installed `opencode` 1.17.10: top-level TUI has no `--auto`/`--yolo`/`--dangerously-skip-permissions`; `opencode run` has `--dangerously-skip-permissions`; `opencode mcp add` / config `mcp` block confirmed.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/LaunchScaffold.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/LaunchScaffold.swift
@@ -113,10 +113,15 @@ enum LaunchScaffold {
                 }
                 // Mirror the user's Claude `jira` MCP into OpenCode's global
                 // config so OpenCode sessions get the same `jira_*` tools
-                // (parity with Claude; CROW-831). No-op when the user has no
-                // Claude Jira MCP to mirror.
-                attempt("OpenCode Jira MCP registration") {
-                    OpenCodeMCPConfigWriter.installGlobalMCPConfig(configHome: configHome)
+                // (parity with Claude; CROW-831). Never throws — it returns an
+                // Outcome — so inspect and log non-success directly rather than
+                // through `attempt`, whose catch branch would never fire.
+                let mcpOutcome = OpenCodeMCPConfigWriter.installGlobalMCPConfig(configHome: configHome)
+                switch mcpOutcome {
+                case .registered, .removed, .unchanged, .noSource:
+                    CrowDaemon.log("OpenCode Jira MCP registration: \(mcpOutcome)")
+                case .skippedUnparseable, .failed:
+                    CrowDaemon.log("WARNING: OpenCode Jira MCP registration: \(mcpOutcome)")
                 }
             }
         }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/LaunchScaffold.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/LaunchScaffold.swift
@@ -111,6 +111,13 @@ enum LaunchScaffold {
                 attempt("OpenCode global config install") {
                     try OpenCodeHookConfigWriter.installGlobalConfig(configHome: configHome, crowPath: crowPath)
                 }
+                // Mirror the user's Claude `jira` MCP into OpenCode's global
+                // config so OpenCode sessions get the same `jira_*` tools
+                // (parity with Claude; CROW-831). No-op when the user has no
+                // Claude Jira MCP to mirror.
+                attempt("OpenCode Jira MCP registration") {
+                    OpenCodeMCPConfigWriter.installGlobalMCPConfig(configHome: configHome)
+                }
             }
         }
     }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/LaunchScaffold.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/LaunchScaffold.swift
@@ -102,27 +102,29 @@ enum LaunchScaffold {
 
         if AgentRegistry.shared.agent(for: .openCode) != nil {
             attempt("OpenCode scaffold") { try OpenCodeScaffolder.scaffold(devRoot: devRoot) }
+            // XDG spec: an empty `XDG_CONFIG_HOME` is treated as unset, so fall
+            // through to ~/.config/opencode rather than a relative path.
+            let configHome = nonEmptyEnv("XDG_CONFIG_HOME")
+                .map { ($0 as NSString).appendingPathComponent("opencode") }
+                ?? NSString(string: "~/.config/opencode").expandingTildeInPath
+            // The state-bridge plugin needs the `crow` binary; the MCP mirror
+            // does not — so gate only the former on `crowPath`.
             if let crowPath {
-                // XDG spec: an empty `XDG_CONFIG_HOME` is treated as unset, so
-                // fall through to ~/.config/opencode rather than a relative path.
-                let configHome = nonEmptyEnv("XDG_CONFIG_HOME")
-                    .map { ($0 as NSString).appendingPathComponent("opencode") }
-                    ?? NSString(string: "~/.config/opencode").expandingTildeInPath
                 attempt("OpenCode global config install") {
                     try OpenCodeHookConfigWriter.installGlobalConfig(configHome: configHome, crowPath: crowPath)
                 }
-                // Mirror the user's Claude `jira` MCP into OpenCode's global
-                // config so OpenCode sessions get the same `jira_*` tools
-                // (parity with Claude; CROW-831). Never throws — it returns an
-                // Outcome — so inspect and log non-success directly rather than
-                // through `attempt`, whose catch branch would never fire.
-                let mcpOutcome = OpenCodeMCPConfigWriter.installGlobalMCPConfig(configHome: configHome)
-                switch mcpOutcome {
-                case .registered, .removed, .unchanged, .noSource:
-                    CrowDaemon.log("OpenCode Jira MCP registration: \(mcpOutcome)")
-                case .skippedUnparseable, .failed:
-                    CrowDaemon.log("WARNING: OpenCode Jira MCP registration: \(mcpOutcome)")
-                }
+            }
+            // Mirror the user's Claude `jira` MCP into OpenCode's global config
+            // so OpenCode sessions get the same `jira_*` tools (parity with
+            // Claude; CROW-831). Never throws — it returns an Outcome — so
+            // inspect and log non-success directly rather than through
+            // `attempt`, whose catch branch would never fire.
+            let mcpOutcome = OpenCodeMCPConfigWriter.installGlobalMCPConfig(configHome: configHome)
+            switch mcpOutcome {
+            case .registered, .removed, .unchanged, .skippedUserOwned, .noSource:
+                CrowDaemon.log("OpenCode Jira MCP registration: \(mcpOutcome)")
+            case .skippedUnparseable, .failed:
+                CrowDaemon.log("WARNING: OpenCode Jira MCP registration: \(mcpOutcome)")
             }
         }
     }
--- a/Packages/CrowOpenCode/Package.swift
+++ b/Packages/CrowOpenCode/Package.swift
@@ -9,19 +9,9 @@ let package = Package(
     ],
     dependencies: [
         .package(path: "../CrowCore"),
-        // SHA-256 for the MCP-mirror provenance digest (so the sidecar records a
-        // hash, not a third plaintext copy of the Jira token). Already resolved
-        // repo-wide via CrowDaemon's web-auth PBKDF2 (CROW-593).
-        .package(url: "https://github.com/apple/swift-crypto.git", "3.0.0" ..< "5.0.0"),
     ],
     targets: [
-        .target(
-            name: "CrowOpenCode",
-            dependencies: [
-                "CrowCore",
-                .product(name: "Crypto", package: "swift-crypto"),
-            ]
-        ),
+        .target(name: "CrowOpenCode", dependencies: ["CrowCore"]),
         .testTarget(name: "CrowOpenCodeTests", dependencies: ["CrowOpenCode"]),
     ]
 )
--- a/Packages/CrowOpenCode/Package.swift
+++ b/Packages/CrowOpenCode/Package.swift
@@ -9,9 +9,19 @@ let package = Package(
     ],
     dependencies: [
         .package(path: "../CrowCore"),
+        // SHA-256 for the MCP-mirror provenance digest (so the sidecar records a
+        // hash, not a third plaintext copy of the Jira token). Already resolved
+        // repo-wide via CrowDaemon's web-auth PBKDF2 (CROW-593).
+        .package(url: "https://github.com/apple/swift-crypto.git", "3.0.0" ..< "5.0.0"),
     ],
     targets: [
-        .target(name: "CrowOpenCode", dependencies: ["CrowCore"]),
+        .target(
+            name: "CrowOpenCode",
+            dependencies: [
+                "CrowCore",
+                .product(name: "Crypto", package: "swift-crypto"),
+            ]
+        ),
         .testTarget(name: "CrowOpenCodeTests", dependencies: ["CrowOpenCode"]),
     ]
 )
--- a/Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeAgent.swift
+++ b/Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeAgent.swift
@@ -20,8 +20,10 @@ import CrowCore
 ///     pipes bind fd 0 and break keyboard input; `--prompt` may only pre-fill.
 ///  2. **State hooks are a JS plugin, not a `hooks.json`.** OpenCode has no
 ///     command-based hook file; `OpenCodeHookConfigWriter` installs a plugin
-///     into `~/.config/opencode/plugins/` that shells out to `crow
-///     hook-event`. See that type for details.
+///     into the worktree's `.opencode/plugins/` with the session UUID baked in
+///     (a global fallback in `~/.config/opencode/plugins/` covers hand-started
+///     sessions and self-suppresses when the per-project plugin is present).
+///     See that type for details.
 ///
 /// OpenCode has no `--rc`/`--name`/`--permission-mode` analog. The closest
 /// permission knob is `--auto` / `--dangerously-skip-permissions`, probed via
--- a/Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeHookConfigWriter.swift
+++ b/Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeHookConfigWriter.swift
@@ -5,16 +5,36 @@ import CrowCore
 /// command-based hook file like Claude Code's `settings.json` or Cursor's
 /// `hooks.json`; instead it auto-loads JS/TS **plugins** from
 /// `~/.config/opencode/plugins/` (global) and `<project>/.opencode/plugins/`
-/// (per-project). MVP is global-only, matching Codex/Cursor's scope.
+/// (per-project). Upstream globs `{plugin,plugins}` (`plugin.ts@v1.18.4:21`),
+/// so either spelling loads; Crow's writer uses `plugins/`.
 ///
-/// So `installGlobalConfig` writes a single plugin file,
-/// `crow-hooks.js`, that subscribes to OpenCode's `event` bus plus the
-/// `tool.execute.before/after` and `permission.ask` hooks and shells out
-/// (via Bun's `$`) to `crow hook-event --agent opencode --event <PascalName>`,
-/// piping a JSON payload with the worktree `cwd` on stdin — exactly the shape
-/// the crow server's `hook-event` RPC expects (it resolves the session by
-/// matching `cwd` against registered worktree paths, the same mechanism
-/// Codex's global hooks use).
+/// **Two scopes, one plugin body (CROW-831):**
+///
+///  - **Per-project (primary).** `writeHookConfig` writes
+///    `<worktree>/.opencode/plugins/crow-hooks.js` with the Crow **session
+///    UUID baked in**, so each worktree emits `crow hook-event --session
+///    <uuid> …`. Resolution is exact — no cwd matching — which closes the
+///    shared-`cwd` collision two sessions rooted at (or resolving to) the same
+///    path could otherwise hit. Every Crow-launched OpenCode session
+///    (`.work`/`.job`/`.review`) flows through `AgentLaunch` →
+///    `writeHookConfig`, so all of them get a session-scoped plugin.
+///
+///  - **Global (fallback).** `installGlobalConfig` still writes a single
+///    `<configHome>/plugins/crow-hooks.js` with **no** session UUID; it
+///    resolves the session by matching the worktree `cwd` against registered
+///    worktree paths (the same mechanism Codex's global hooks use). This
+///    covers a `opencode` a user starts *by hand* in a terminal Crow didn't
+///    auto-launch. To avoid **double emission** — OpenCode dedups plugins by
+///    file URL, so the global and per-project files are distinct and *both*
+///    load — the global plugin self-suppresses (returns no hooks) whenever a
+///    per-project `crow-hooks.js` exists in the cwd. So in a Crow worktree only
+///    the session-scoped plugin ever fires.
+///
+/// Both variants subscribe to OpenCode's `event` bus plus the
+/// `tool.execute.before/after` and `permission.ask` hooks and shell out (via
+/// Bun's `$`) to `crow hook-event --agent opencode --event <PascalName>`,
+/// piping a JSON payload (`{ cwd, … }`) on stdin — the shape the crow server's
+/// `hook-event` RPC expects.
 ///
 /// The plugin maps OpenCode's event/hook vocabulary onto Crow's canonical
 /// PascalCase names so `OpenCodeSignalSource` can share Claude/Codex/Cursor's
@@ -36,65 +56,115 @@ import CrowCore
 /// `permission.ask` hook fires exactly when OpenCode requests a decision; we
 /// only observe it (never set `output.status`), so the user's/agent's choice
 /// still stands.
-///
-/// CROW-545 gap: the event *names* are now verified, but the *timing/semantics*
-/// (esp. whether `session.idle` is the right "done" signal for interactive
-/// TUI sessions) should still be confirmed empirically at runtime — the same
-/// caveat Cursor's writer carries for `afterAgentResponse`.
-///
-/// Because `HookConfigWriter`'s per-session API doesn't fit OpenCode's
-/// global model, the per-session methods are intentionally no-ops — real
-/// work happens via the static `installGlobalConfig` call at app launch.
 public struct OpenCodeHookConfigWriter: HookConfigWriter {
 
     public init() {}
 
-    // MARK: - HookConfigWriter Conformance (no-ops)
+    // MARK: - HookConfigWriter Conformance (per-project plugin)
 
-    /// No-op. OpenCode plugins are global, not per-worktree — see
-    /// `installGlobalConfig`. Per-project `<worktree>/.opencode/plugins/` is
-    /// deferred to a follow-up; MVP stays global-only.
-    public func writeHookConfig(worktreePath: String, sessionID: UUID, crowPath: String) throws {}
+    /// Install `<worktreePath>/.opencode/plugins/crow-hooks.js` with
+    /// `sessionID` baked in, so this worktree's OpenCode emits
+    /// `crow hook-event --session <sessionID> …`. Idempotent — we own this
+    /// single-purpose file and overwrite it wholesale, so there's nothing to
+    /// merge. Called from the `AgentLaunch` path on every OpenCode launch.
+    public func writeHookConfig(worktreePath: String, sessionID: UUID, crowPath: String) throws {
+        let pluginsDir = Self.worktreePluginsDir(worktreePath)
+        try FileManager.default.createDirectory(atPath: pluginsDir, withIntermediateDirectories: true)
+        let pluginPath = (pluginsDir as NSString).appendingPathComponent(Self.pluginFileName)
+        let content = Self.pluginSource(crowPath: crowPath, sessionID: sessionID)
+        try content.write(to: URL(fileURLWithPath: pluginPath), atomically: true, encoding: .utf8)
+    }
 
-    /// No-op. The global plugin stays in place when individual sessions are
-    /// deleted; it serves all sessions.
-    public func removeHookConfig(worktreePath: String) {}
+    /// Remove the per-project plugin from a worktree's `.opencode/plugins/`,
+    /// leaving any user-authored plugins in that directory untouched. Best
+    /// effort: also prunes the `plugins`/`.opencode` dirs when they're left
+    /// empty (i.e. Crow created them), so tearing a session down leaves no
+    /// trace, but never touches a directory that still holds other files.
+    public func removeHookConfig(worktreePath: String) {
+        let pluginsDir = Self.worktreePluginsDir(worktreePath)
+        let pluginPath = (pluginsDir as NSString).appendingPathComponent(Self.pluginFileName)
+        let fm = FileManager.default
+        guard fm.fileExists(atPath: pluginPath) else { return }
+        try? fm.removeItem(atPath: pluginPath)
+        Self.removeIfEmpty(pluginsDir)
+        Self.removeIfEmpty((worktreePath as NSString).appendingPathComponent(".opencode"))
+    }
 
-    // MARK: - Global Configuration
+    // MARK: - Global Configuration (fallback)
 
     /// Install or refresh `<configHome>/plugins/crow-hooks.js`. `configHome`
     /// is OpenCode's config dir (default `~/.config/opencode`, honoring
     /// `XDG_CONFIG_HOME`). Idempotent — we own this single-purpose file and
-    /// overwrite it wholesale on every launch, so there's nothing to merge
-    /// (unlike Cursor's shared `hooks.json`).
+    /// overwrite it wholesale on every launch. The global plugin carries no
+    /// session UUID and self-suppresses when a per-project plugin is present
+    /// (see the type doc), so it only ever fires for hand-started sessions.
     public static func installGlobalConfig(configHome: String, crowPath: String) throws {
-        let pluginsDir = (configHome as NSString).appendingPathComponent("plugins")
+        let pluginsDir = globalPluginsDir(configHome)
         try FileManager.default.createDirectory(atPath: pluginsDir, withIntermediateDirectories: true)
-        let pluginPath = (pluginsDir as NSString).appendingPathComponent("crow-hooks.js")
-        let content = pluginSource(crowPath: crowPath)
+        let pluginPath = (pluginsDir as NSString).appendingPathComponent(pluginFileName)
+        let content = pluginSource(crowPath: crowPath, sessionID: nil)
         try content.write(to: URL(fileURLWithPath: pluginPath), atomically: true, encoding: .utf8)
+    }
+
+    // MARK: - Plugin Source
+
+    static let pluginFileName = "crow-hooks.js"
+
+    /// `<configHome>/plugins` — the global config home already ends at
+    /// `…/opencode`, so no `.opencode` segment is added.
+    private static func globalPluginsDir(_ configHome: String) -> String {
+        (configHome as NSString).appendingPathComponent("plugins")
+    }
+
+    /// `<worktree>/.opencode/plugins` — per-project scope adds the `.opencode`
+    /// segment OpenCode discovers project config/plugins under.
+    private static func worktreePluginsDir(_ worktree: String) -> String {
+        let opencodeDir = (worktree as NSString).appendingPathComponent(".opencode")
+        return (opencodeDir as NSString).appendingPathComponent("plugins")
+    }
+
+    private static func removeIfEmpty(_ dir: String) {
+        let fm = FileManager.default
+        guard let contents = try? fm.contentsOfDirectory(atPath: dir), contents.isEmpty else { return }
+        try? fm.removeItem(atPath: dir)
     }
 
     /// The JS plugin body, with `crowPath` baked in as a string literal
     /// (mirrors how Cursor bakes `crowPath` into its `hooks.json` commands).
-    /// Written as plain `.js` so it needs no `@opencode-ai/plugin` types or
-    /// a build step — OpenCode runs it directly on Bun.
-    static func pluginSource(crowPath: String) -> String {
+    /// Written as plain `.js` so it needs no `@opencode-ai/plugin` types or a
+    /// build step — OpenCode runs it directly on Bun.
+    ///
+    /// When `sessionID` is non-nil the emit carries `--session <uuid>` (exact
+    /// resolution, per-project scope). When nil the emit omits it and the
+    /// server resolves by cwd match (global fallback scope), and the plugin
+    /// self-suppresses if a per-project plugin exists in the cwd.
+    static func pluginSource(crowPath: String, sessionID: UUID? = nil) -> String {
         let crow = jsStringLiteral(crowPath)
+        let session = jsStringLiteral(sessionID?.uuidString ?? "")
+        let scopeNote = sessionID != nil
+            ? "session-scoped (UUID baked in below)"
+            : "global fallback (resolves by cwd; defers to a per-project plugin)"
         return """
         // Crow ↔ OpenCode hook bridge — auto-generated by Crow (safe to delete).
+        // Scope: \(scopeNote).
         // Forwards OpenCode lifecycle events to the running Crow app so session
         // cards reflect agent state. Regenerated on every Crow launch.
         //
-        // Each emit pipes a JSON payload ({ cwd, ... }) to `crow hook-event`;
-        // the crow server resolves the Crow session by matching `cwd` against
-        // registered worktree paths, so no session UUID is needed here.
+        // Each emit pipes a JSON payload ({ cwd, ... }) to `crow hook-event`.
+        // With a session UUID we pass `--session <uuid>` so the crow server
+        // resolves the session exactly; without one it matches `cwd` against
+        // registered worktree paths.
         const CROW = \(crow);
+        const SESSION = \(session);
 
         async function emit($, cwd, event, extra) {
           try {
             const payload = JSON.stringify(Object.assign({ cwd }, extra || {}));
-            await $`echo ${payload} | ${CROW} hook-event --agent opencode --event ${event}`.quiet();
+            if (SESSION) {
+              await $`echo ${payload} | ${CROW} hook-event --session ${SESSION} --agent opencode --event ${event}`.quiet();
+            } else {
+              await $`echo ${payload} | ${CROW} hook-event --agent opencode --event ${event}`.quiet();
+            }
           } catch (_) {
             // Fire-and-forget: a hook failure (e.g. Crow not running) must
             // never disrupt OpenCode.
@@ -105,6 +175,18 @@ public struct OpenCodeHookConfigWriter: HookConfigWriter {
           // Prefer the git worktree path — that's what Crow registers and
           // matches on. Fall back to the process cwd.
           const cwd = worktree || directory;
+          // Global fallback only: if this worktree has a session-scoped Crow
+          // plugin, defer to it. OpenCode loads both (they are distinct file
+          // URLs), so without this guard the same event would emit twice.
+          if (!SESSION) {
+            try {
+              if (await Bun.file(cwd + "/.opencode/plugins/crow-hooks.js").exists()) {
+                return {};
+              }
+            } catch (_) {
+              // Bun.file unavailable / unreadable — fall through and emit.
+            }
+          }
           return {
             event: async ({ event }) => {
               switch (event.type) {
--- a/Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeHookConfigWriter.swift
+++ b/Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeHookConfigWriter.swift
@@ -92,7 +92,10 @@ public struct OpenCodeHookConfigWriter: HookConfigWriter {
         let pluginsDir = Self.worktreePluginsDir(worktreePath)
         let pluginPath = (pluginsDir as NSString).appendingPathComponent(Self.pluginFileName)
         let fm = FileManager.default
-        guard fm.fileExists(atPath: pluginPath) else { return }
+        // Remove the plugin if it's still there — but don't early-return when it
+        // isn't: a user who took the "safe to delete" header at its word and
+        // removed `crow-hooks.js` by hand must not be left with an orphaned,
+        // self-ignoring `.gitignore` that `git status` can't even surface.
         try? fm.removeItem(atPath: pluginPath)
         // Drop our `.gitignore` only if it still holds exactly our content — a
         // user may have replaced it with their own rules (same provenance
--- a/Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeHookConfigWriter.swift
+++ b/Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeHookConfigWriter.swift
@@ -82,21 +82,23 @@ public struct OpenCodeHookConfigWriter: HookConfigWriter {
     }
 
     /// Remove the per-project plugin from a worktree's `.opencode/plugins/`,
-    /// leaving any user-authored plugins in that directory untouched. Best
-    /// effort: also removes our self-scoped `.gitignore` and prunes the
-    /// `plugins`/`.opencode` dirs when they're left empty (i.e. Crow created
-    /// them), so tearing a session down leaves no trace, but never touches a
-    /// directory that still holds other files.
+    /// leaving any user-authored files (plugins *and* a hand-written
+    /// `.gitignore`) untouched. Best effort: removes our own `.gitignore` when
+    /// it still carries exactly our content, and prunes the `plugins`/`.opencode`
+    /// dirs when they're left empty (i.e. Crow created them), so tearing a
+    /// session down leaves no trace — but never touches a file or directory that
+    /// isn't ours.
     public func removeHookConfig(worktreePath: String) {
         let pluginsDir = Self.worktreePluginsDir(worktreePath)
         let pluginPath = (pluginsDir as NSString).appendingPathComponent(Self.pluginFileName)
         let fm = FileManager.default
         guard fm.fileExists(atPath: pluginPath) else { return }
         try? fm.removeItem(atPath: pluginPath)
-        // Drop our `.gitignore` only if it's the sole remaining entry (i.e. Crow
-        // owns the dir); a user plugin alongside means leave everything.
+        // Drop our `.gitignore` only if it still holds exactly our content — a
+        // user may have replaced it with their own rules (same provenance
+        // discipline as writeGitignore).
         let gitignorePath = (pluginsDir as NSString).appendingPathComponent(".gitignore")
-        if (try? fm.contentsOfDirectory(atPath: pluginsDir)) == [".gitignore"] {
+        if (try? String(contentsOfFile: gitignorePath, encoding: .utf8)) == Self.gitignoreBody {
             try? fm.removeItem(atPath: gitignorePath)
         }
         Self.removeIfEmpty(pluginsDir)
@@ -142,18 +144,29 @@ public struct OpenCodeHookConfigWriter: HookConfigWriter {
         try? fm.removeItem(atPath: dir)
     }
 
+    /// The exact body Crow writes to `.opencode/plugins/.gitignore`. A single
+    /// source of truth so `writeGitignore` and `removeHookConfig` agree on what
+    /// "ours" means when deciding whether it's safe to write/delete.
+    static let gitignoreBody = """
+    # Crow-generated — safe to delete.
+    \(pluginFileName)
+    .gitignore
+    """
+
     /// Write a self-scoped `.gitignore` into `dir` that ignores Crow's generated
     /// plugin (and the `.gitignore` itself), so an agent's `git add -A` in the
-    /// worktree never stages them. Idempotent; best effort (a failure just means
-    /// the files remain stageable, the pre-existing behavior).
+    /// worktree never stages them. `.opencode/plugins/` is the *user's*
+    /// directory (their own plugins live there), so this never clobbers a
+    /// pre-existing `.gitignore` — it writes only into an empty slot or over our
+    /// own previous body. Idempotent; best effort.
     private static func writeGitignore(inDir dir: String) {
         let path = (dir as NSString).appendingPathComponent(".gitignore")
-        let body = """
-        # Crow-generated — safe to delete.
-        \(pluginFileName)
-        .gitignore
-        """
-        try? body.write(to: URL(fileURLWithPath: path), atomically: true, encoding: .utf8)
+        if let existing = try? String(contentsOfFile: path, encoding: .utf8),
+           existing != gitignoreBody {
+            // A `.gitignore` we didn't write (user's own rules) — leave it.
+            return
+        }
+        try? gitignoreBody.write(to: URL(fileURLWithPath: path), atomically: true, encoding: .utf8)
     }
 
     /// The JS plugin body, with `crowPath` baked in as a string literal
--- a/Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeHookConfigWriter.swift
+++ b/Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeHookConfigWriter.swift
@@ -73,19 +73,32 @@ public struct OpenCodeHookConfigWriter: HookConfigWriter {
         let pluginPath = (pluginsDir as NSString).appendingPathComponent(Self.pluginFileName)
         let content = Self.pluginSource(crowPath: crowPath, sessionID: sessionID)
         try content.write(to: URL(fileURLWithPath: pluginPath), atomically: true, encoding: .utf8)
+        // The plugin embeds the absolute `crow` path and the session UUID and
+        // lives inside the user's git worktree, where an agent's `git add -A`
+        // could stage it. A self-scoped `.gitignore` keeps our generated files
+        // out of the index without touching OpenCode's own `.opencode/.gitignore`
+        // (a level up, which upstream manages).
+        Self.writeGitignore(inDir: pluginsDir)
     }
 
     /// Remove the per-project plugin from a worktree's `.opencode/plugins/`,
     /// leaving any user-authored plugins in that directory untouched. Best
-    /// effort: also prunes the `plugins`/`.opencode` dirs when they're left
-    /// empty (i.e. Crow created them), so tearing a session down leaves no
-    /// trace, but never touches a directory that still holds other files.
+    /// effort: also removes our self-scoped `.gitignore` and prunes the
+    /// `plugins`/`.opencode` dirs when they're left empty (i.e. Crow created
+    /// them), so tearing a session down leaves no trace, but never touches a
+    /// directory that still holds other files.
     public func removeHookConfig(worktreePath: String) {
         let pluginsDir = Self.worktreePluginsDir(worktreePath)
         let pluginPath = (pluginsDir as NSString).appendingPathComponent(Self.pluginFileName)
         let fm = FileManager.default
         guard fm.fileExists(atPath: pluginPath) else { return }
         try? fm.removeItem(atPath: pluginPath)
+        // Drop our `.gitignore` only if it's the sole remaining entry (i.e. Crow
+        // owns the dir); a user plugin alongside means leave everything.
+        let gitignorePath = (pluginsDir as NSString).appendingPathComponent(".gitignore")
+        if (try? fm.contentsOfDirectory(atPath: pluginsDir)) == [".gitignore"] {
+            try? fm.removeItem(atPath: gitignorePath)
+        }
         Self.removeIfEmpty(pluginsDir)
         Self.removeIfEmpty((worktreePath as NSString).appendingPathComponent(".opencode"))
     }
@@ -127,6 +140,20 @@ public struct OpenCodeHookConfigWriter: HookConfigWriter {
         let fm = FileManager.default
         guard let contents = try? fm.contentsOfDirectory(atPath: dir), contents.isEmpty else { return }
         try? fm.removeItem(atPath: dir)
+    }
+
+    /// Write a self-scoped `.gitignore` into `dir` that ignores Crow's generated
+    /// plugin (and the `.gitignore` itself), so an agent's `git add -A` in the
+    /// worktree never stages them. Idempotent; best effort (a failure just means
+    /// the files remain stageable, the pre-existing behavior).
+    private static func writeGitignore(inDir dir: String) {
+        let path = (dir as NSString).appendingPathComponent(".gitignore")
+        let body = """
+        # Crow-generated — safe to delete.
+        \(pluginFileName)
+        .gitignore
+        """
+        try? body.write(to: URL(fileURLWithPath: path), atomically: true, encoding: .utf8)
     }
 
     /// The JS plugin body, with `crowPath` baked in as a string literal
--- a/Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeLaunchArgs.swift
+++ b/Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeLaunchArgs.swift
@@ -10,8 +10,48 @@ public enum OpenCodeLaunchArgs {
     /// omits an optional flag, never breaks launch.
     private nonisolated(unsafe) static var tuiAutoFlagCache: [String: Bool] = [:]
     private nonisolated(unsafe) static var runHelpCache: [String: String] = [:]
+    private nonisolated(unsafe) static var versionCache: [String: SemVer?] = [:]
 
     private static let helpProbeTimeoutSeconds: TimeInterval = 5
+
+    /// The first `opencode` release that re-added the top-level TUI `--auto`
+    /// flag. It was present pre-`1.17`, **removed across the `1.17.x` window**,
+    /// then re-added in `1.18.0` (2026-07-14; verified `sst/opencode`
+    /// `tui.ts@v1.18.4`). Below this, the TUI `--help` probe for `--auto` can
+    /// only ever answer "no" — so we short-circuit it (CROW-831). At/above it we
+    /// still probe, so a *future* upstream flip is caught without a code change.
+    static let tuiAutoReintroducedVersion = SemVer(1, 18, 0)
+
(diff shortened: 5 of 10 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Verified locally: `swift test --package-path Packages/CrowOpenCode` → **46 tests / 5 suites pass**; `swift build --package-path Packages/CrowDaemon` → **Build complete**. No failures.

The three pieces are well-scoped and the doc comments are unusually good — the `plugin.ts@v1.18.4` / `config.ts@v1.18.4:258-260` citations make the upstream assumptions auditable. Three things should be fixed before merge.

### Critical Issues

None. Nothing here can corrupt state or break launch.

---

### 🟡 1. `tuiAutoKnownAbsent` narrows further than the version history it documents

`Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeLaunchArgs.swift:51-54`

```swift
static func tuiAutoKnownAbsent(version: SemVer?) -> Bool {
    guard let version else { return false }
    return version < tuiAutoReintroducedVersion   // < 1.18.0
}
```

The doc comment eight lines above states the actual history: `--auto` was *"present pre-`1.17`, **removed across the `1.17.x` window**, then re-added in `1.18.0`"*. The predicate implements `< 1.18.0`, which also swallows every build **below 1.17** — where the flag *was* present. The PR description says the same thing ("gone only in `v1.17.x`"), so this is code disagreeing with its own stated premise, not a documentation gap.

**Failure scenario:** a user on `opencode` 1.16.x runs a `.job` session with `autoPermissionMode`. Pre-PR, `tuiSupportsAuto` probed `opencode --help`, found `--auto`, and `tuiAutoApproveSuffix` appended it. Post-PR, `installedVersion` returns `1.16.x` → `tuiAutoKnownAbsent` → `true` → `supports = false` → `opencode --continue` launches with **no** auto-approve, and the unattended session blocks on the first permission prompt of a follow-up turn. It fails closed rather than dangerously, which is why this is Yellow and not Red — but it's a live regression for pre-1.17 installs.

`OpenCodeLaunchArgsTests.swift:732` locks the wrong behavior in:

```swift
#expect(OpenCodeLaunchArgs.tuiAutoKnownAbsent(version: OpenCodeLaunchArgs.SemVer(1, 0, 0)))
```

Suggested fix — narrow to the window that is actually documented:

```swift
static let tuiAutoRemovedVersion = SemVer(1, 17, 0)

static func tuiAutoKnownAbsent(version: SemVer?) -> Bool {
    guard let version else { return false }
    return version >= tuiAutoRemovedVersion && version < tuiAutoReintroducedVersion
}
```

…and flip the `SemVer(1, 0, 0)` expectation to `false`, adding a `SemVer(1, 16, 9)` → `false` case.

---

### 🟡 2. Mirrored MCP credentials land in a world-readable file

`Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeMCPConfigWriter.swift:507-515`

```swift
let data = try JSONSerialization.data(withJSONObject: root, options: [.prettyPrinted, .sortedKeys])
try data.write(to: URL(fileURLWithPath: targetPath), options: .atomic)
```

`translateClaudeServer` copies the source server's `env` → `environment` and `headers` verbatim. For a Jira MCP those are exactly the fields that carry secrets — `JIRA_API_TOKEN`, `Authorization: Bearer …` (the PR's own test fixture uses `["Authorization": "Bearer x"]`). The **source** file is deliberately owner-only — on this machine `~/.claude.json` is `-rw-------`. The **target** is written with `.atomic` and no mode fix-up, so it lands at the umask default (~`0644`): group- and world-readable.

This repo already documents the exact hazard, in `Packages/CrowEngine/Sources/CrowEngine/Scaffolder.swift:177-185`:

> `atomically: true` renames a fresh temp file over the target, which resets its mode to the umask default (~0o644). This same file's `env` block can carry a resolved gateway bearer token … so re-apply 0o600 here.

Same convention at `ClaudeHookConfigWriter.writeGatewayEnv:147`, `ConfigStore.swift:89`, `JSONStore.swift:285`, `CrowLog.swift:106`. The new writer is the one credential-bearing writer that skips it.

**Failure scenario:** a user with a token-bearing Jira MCP in `~/.claude.json` (0600) launches `crowd`; their token is now readable by every local account and every unsandboxed process at `~/.config/opencode/opencode.json`.

Suggested fix, matching the sibling writers:

```swift
try fm.createDirectory(atPath: configHome, withIntermediateDirectories: true)
try data.write(to: URL(fileURLWithPath: targetPath), options: .atomic)
// `mcp.<name>.environment`/`headers` mirror Claude's tokens; ~/.claude.json is
// 0600, so don't widen them. `.atomic` resets mode to the umask default.
try? fm.setAttributes([.posixPermissions: 0o600], ofItemAtPath: targetPath)
```

Worth asserting in `OpenCodeMCPConfigWriterTests` too — `ConfigStoreTests.swift:129` and `ScaffolderSettingsMergeTests.swift:260` already have that pattern to copy.

---

### 🟡 3. The mirror is write-only — no un-mirror, no opt-out

`Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeMCPConfigWriter.swift:482-504`

Two halves of the same gap:

**(a) Deletion doesn't propagate.** When the user removes `jira` from `~/.claude.json`, `translateClaudeServer` returns `nil` → `.noSource` → early return. The previously-written `mcp.jira` entry stays in `opencode.json` indefinitely, still holding the copied credential. The stated goal is *"exactly as Claude would"* — but Claude loses the tool while OpenCode keeps launching a stale server. Parity holds only in the add direction. An explicit `.removed` outcome (drop `mcp[serverName]` when Crow previously wrote it and the source is gone) would close it.

**(b) `enabled: true` is unconditionally re-asserted.** A user who edits the Crow-written entry to `"enabled": false` — the natural way to turn an unwanted mirrored server off — fails the `NSDictionary(dictionary: existing).isEqual(to: openCodeServer)` check on the next daemon launch, so line 504 rewrites it back to `enabled: true`. There is currently no way to keep the mirror off. Preserving an existing explicit `enabled` value (or comparing on everything *except* `enabled`) would fix it.

**Failure scenario:** user disables the mirrored Jira MCP in `opencode.json`, restarts Crow, and it is silently re-enabled — every launch, with no log line saying so.

---

### Security Review

**Strengths:**

- **No shell injection in the generated plugin.** Every interpolation in `` $`echo ${payload} | ${CROW} hook-event --session ${SESSION} …` `` goes through Bun Shell's escaping, so a hostile `tool_name` or a path with spaces/quotes can't break out of the command. `jsStringLiteral` (`OpenCodeHookConfigWriter.swift:227-232`) escapes `\` and `"` before the path reaches the JS literal.
- **The `--session` shape is real, not assumed.** `HookEventCommand.swift:21-22` accepts `--session` and validates it as a UUID (`validate()`), so the per-project emit resolves as documented. Baking the UUID in genuinely closes the shared-`cwd` collision.
- **Refuses to clobber what it can't parse.** `readClaudeJiraServer` distinguishes absent from unparseable, and both the source and target unparseable paths return `.skippedUnparseable` instead of overwriting — covered by `refusesToTouchUnparseableOpenCodeConfig`.
- **Merge is genuinely non-destructive.** `$schema`, `theme`, and sibling `mcp.*` entries all survive (`mergePreservesExistingOpenCodeKeys`); writing `opencode.json` rather than the user's `opencode.jsonc` keeps their comments intact.
- **No secret ever reaches a log.** Both `NSLog` calls carry only paths and `error.localizedDescription`.
- **`permission.ask` stays observe-only** — the hook never sets `output.status`, so Crow cannot silently auto-approve on the user's behalf.

**Concerns:** the file-mode downgrade (🟡 2) and the stale-credential persistence half of (🟡 3).

---

### Code Quality

**🟢 `removeHookConfig` is never reached in production.** The only call site is `SessionService.swift:1707`, hardcoded to `ClaudeHookConfigWriter().removeHookConfig(...)`. The new OpenCode implementation and its two tests are exercised only by the test suite. In the delete path the worktree is removed wholesale so nothing leaks, but routing that line through `agent.hookConfigWriter` (as `writeHookConfig` already is at `:908` / `:1434`) would make the empty-dir pruning actually do something and stop the two paths drifting.

**🟢 `.opencode/plugins/crow-hooks.js` is untracked inside the user's repo.** It carries an absolute `crow` path (so, the home dir and username) plus the session UUID, and an agent running `git add -A` in its own worktree can commit it. Same exposure class as `.claude/settings.local.json`, so this is consistent with existing behavior rather than new — but a one-line append to `.git/info/exclude` alongside the write would remove the footgun for both.

**🟢 `installGlobalMCPConfig` failures are invisible to the scaffold.** It returns `.failed`/`.skippedUnparseable` rather than throwing, so `LaunchScaffold.attempt("OpenCode Jira MCP registration")` (`LaunchScaffold.swift:118-120`) always records success and the `WARNING: … failed` branch is unreachable for this step. Checking the returned `Outcome` and logging non-success through `CrowDaemon.log` would match what the neighbouring `attempt` blocks give you.

**🟢 The whole of `~/.claude.json` is parsed at every daemon launch** to read one key. That file accumulates project history and is 2.3 MB on this machine; multi-MB is normal for heavy users. It's a once-per-launch cost on a non-hot path, so this is a note rather than a request — but only `mcpServers` is needed.

**🟢 `$schema` is dropped for pre-existing files.** Line 496's `root = parsed` discards the `["$schema": …]` default, so an existing `opencode.json` without a schema key never gains one. Arguably the right call (don't inject into the user's file) — flagging only because the fresh-file path does add it, and `registersJiraMCPFromClaudeConfig` asserts it, so the asymmetry isn't obvious from the tests.

**Test coverage is solid.** The new suites cover both plugin variants, the self-suppress guard, per-project write/remove, user-plugin preservation, MCP translation (local/remote/unusable), merge, idempotency, unparseable refusal, and SemVer ordering. The gaps line up with the findings above: no file-mode assertion, no source-removal case, and `tuiAutoKnownAbsent` asserts the over-broad range.

---

### Summary Table

| # | Color | Finding | Location |
|---|-------|---------|----------|
| 1 | 🟡 | `tuiAutoKnownAbsent` also skips the probe below 1.17, where `--auto` exists | `OpenCodeLaunchArgs.swift:51-54` |
| 2 | 🟡 | Mirrored MCP tokens written at ~0644; source is 0600 | `OpenCodeMCPConfigWriter.swift:511` |
| 3 | 🟡 | No un-mirror on source removal; `enabled: false` re-flipped every launch | `OpenCodeMCPConfigWriter.swift:482-504` |
| 4 | 🟢 | OpenCode `removeHookConfig` never called in production | `SessionService.swift:1707` |
| 5 | 🟢 | Untracked `.opencode/` plugin committable from a worktree | `OpenCodeHookConfigWriter.swift:70-76` |
| 6 | 🟢 | Writer failures invisible to `LaunchScaffold.attempt` | `LaunchScaffold.swift:118-120` |
| 7 | 🟢 | Full multi-MB `~/.claude.json` parse per launch for one key | `OpenCodeMCPConfigWriter.swift:529-540` |
| 8 | 🟢 | `$schema` default dropped for pre-existing config files | `OpenCodeMCPConfigWriter.swift:496` |

**Recommendation:** Request Changes — driven by [0 Red, 3 Yellow, 5 Green] findings. #1 and #2 are one-liners; #3 needs a small lifecycle decision. Nice work on the version-gating rationale and the self-suppress guard — the double-emit analysis is exactly right, just applied over too wide a version range.

---

dgershman:
Thanks for the thorough review — all three Yellows fixed, plus the logging Green, in `1eb9634e`.

**🟡 1 — version window.** Sharp catch: `< 1.18.0` also swallowed pre-1.17 where `--auto` existed (fail-closed regression for 1.16.x). Narrowed to `[tuiAutoRemovedVersion=1.17.0, tuiAutoReintroducedVersion=1.18.0)`; flipped `SemVer(1,0,0)`→`false` and added `1.16.9`→`false` / `1.17.0`→`true`.

**🟡 2 — file mode.** Now `chmod 0600` after the atomic write, matching `Scaffolder.swift:177` / `writeGatewayEnv:147`. New `writesTargetOwnerOnly` test asserts it.

**🟡 3 — write-only mirror.** Both halves closed:
- **(a) un-mirror:** source removal now drops the stale `mcp.jira` (new `.removed` outcome) so a credential-bearing server never lingers.
- **(b) opt-out sticks:** an explicit `enabled: false` on the mirrored entry is preserved across launches instead of re-flipped to `true`.

New tests: `unMirrorsWhenSourceRemoved` (+ clean no-op after), `preservesUserDisabledEnabled`.

**🟢 6 — logging.** `installGlobalMCPConfig` returns an `Outcome` (never throws), so `attempt(...)` always logged success. `LaunchScaffold` now inspects it and logs — WARNING on `.failed`/`.skippedUnparseable`.

Greens deferred, with reasoning:
- **#4 (removeHookConfig routing):** the sole call site (`SessionService.swift:1707`) is a `nonisolated static performDiskCleanup(for:)` over worktree items with **no agent/agentKind in scope**, and the worktree is `git worktree remove --force`'d wholesale on the very next line — so the empty-dir pruning is moot in production either way. Plumbing agentKind through that static cleanup is a bigger, unrelated change for zero prod benefit; my `removeHookConfig` is correct + tested for any caller that does route through `agent.hookConfigWriter`. Happy to do it as a follow-up if you'd prefer.
- **#5 (`.opencode/` git-exclude):** same exposure class as the existing, un-excluded `.claude/settings.local.json`; making it a one-liner here means resolving the worktree's shared `$GIT_COMMON_DIR/info/exclude` (a subprocess in the launch path). Consistent-with-existing rather than new, so I've left it — worth a small dedicated PR that covers `.claude`/`.cursor`/`.codex`/`.opencode` uniformly.
- **#7 (whole-file parse) / #8 ($schema asymmetry):** noted. `JSONSerialization` must parse the whole file for one key (once/launch, non-hot path); `$schema` is deliberately not injected into a pre-existing user file. Left as-is per your "note rather than request."

49 tests / 5 suites green.

dhilgaertner (changes requested):
## Code & Security Review — round 2

Verified locally on `1eb9634`: `swift test --package-path Packages/CrowOpenCode` → **49 tests / 5 suites pass**; `swift build --package-path Packages/CrowDaemon` → **Build complete**. No failures.

**Round-1 findings: all four addressed, each with a test.**

| Round 1 | Status | Where |
|---|---|---|
| 🟡 1 version window too wide | ✅ Fixed — `tuiAutoRemovedVersion = 1.17.0`, predicate is now `[1.17.0, 1.18.0)` | `OpenCodeLaunchArgs.swift:30,60-63`; `tuiAutoKnownAbsentOnlyForRemovedWindow` covers `1.16.9`/`1.17.0`/`1.17.10`/`1.18.0`/`nil` |
| 🟡 2 credentials at ~0644 | ✅ Fixed — extracted `write(_:to:configHome:fm:)` re-applies `0o600` after `.atomic`, with the rationale in the doc comment | `OpenCodeMCPConfigWriter.swift:134-151`; `writesTargetOwnerOnly` asserts the mode |
| 🟡 3a no un-mirror | ✅ Implemented — new `.removed` outcome drops `mcp.jira` and prunes an empty `mcp` | `:97-107`; `unMirrorsWhenSourceRemoved` |
| 🟡 3b `enabled: false` re-flipped | ✅ Fixed — an existing explicit `enabled` is carried onto the translated entry before the equality check | `:112-116`; `preservesUserDisabledEnabled` |
| 🟢 6 scaffold swallowed outcomes | ✅ Fixed — `Outcome` inspected and non-success logged as `WARNING` | `LaunchScaffold.swift:114-125` |

The 0600 fix is the right shape (matches `Scaffolder.swift:177` / `ClaudeHookConfigWriter.writeGatewayEnv:147`), and factoring the write into one helper means the removal path can't drift from the register path. One issue falls out of the new un-mirror logic.

### Critical Issues

None.

---

### 🟡 1. Un-mirror deletes `mcp.jira` whether or not Crow wrote it

`Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeMCPConfigWriter.swift:97-107`

```swift
// 3a. Source gone → drop any stale mirror we wrote.
guard let sourceServer, let translated = translateClaudeServer(sourceServer) else {
    guard mcp[serverName] != nil else { return .noSource }
    mcp.removeValue(forKey: serverName)
```

The comment says *"drop any stale mirror **we wrote**"*, but the predicate is only `mcp["jira"] != nil` — there is no provenance record, so Crow cannot distinguish an entry it mirrored from one the user authored. The doc comment resolves this by declaring ownership (`:62-63`: *"Crow owns the `mcp.jira` key in this file; user-authored MCP servers belong in `opencode.jsonc`"*), but that convention is Crow's alone — `~/.config/opencode/opencode.json` is OpenCode's own documented global config path and where `opencode mcp add` lands, so it is exactly where a user's hand-configured Jira server already lives.

**Failure scenario.** An OpenCode-primary user who set up their own Jira MCP before this PR (the only way to get `jira_*` tools in OpenCode until now) has `mcp.jira` in `~/.config/opencode/opencode.json` and — not being a Claude Code user — no `mcpServers.jira` in `~/.claude.json`. On the next `crowd` boot, `readClaudeJiraServer` returns `.success(nil)` → the guard fires → their server, including whatever token sits in `environment`/`headers`, is deleted and the file rewritten. It recurs every launch, so re-adding it doesn't help; nothing in the UI says why, only a `CrowDaemon.log` line reading `removed`. That precondition pair isn't exotic — it's the default state of the users this feature targets.

The same missing provenance makes the register path lossy in the other direction: with a Claude `jira` present, line 121 overwrites a user's hand-authored `mcp.jira` with the translated one (the `enabled` carry-over at `:112-116` is the only field preserved). And because OpenCode merges `opencode.json` alongside `opencode.jsonc`, a user who *does* follow the doc comment and puts a server named `jira` in their `.jsonc` gets it silently shadowed by Crow's entry — with no un-mirror path, since Crow only ever edits the `.json`.

**Suggested fix** — persist what Crow wrote (Crow's own state dir, not the user's config, so nothing has to survive OpenCode's schema validation), and treat "no record" as "not ours" on **both** paths:

```swift
// ~/.local/share/crow/opencode-mcp-mirror.json — {"jira": {…last entry Crow wrote…}}
let mirrored = readMirrorRecord()[serverName]

// remove path: only reclaim what we actually wrote
guard let existing = mcp[serverName] as? [String: Any] else { return .noSource }
guard let mirrored, NSDictionary(dictionary: existing).isEqual(to: mirrored) else {
    NSLog("[OpenCodeMCPConfigWriter] mcp.%@ wasn't written by Crow; leaving it alone", serverName)
    return .noSource
}

// register path: same guard before overwriting a pre-existing entry
```

If a sidecar file isn't worth it, the cheaper version is to keep the entry and log once — a lingering mirror is a smaller harm than deleting a config Crow didn't create. Either way, worth a test alongside `unMirrorsWhenSourceRemoved`: *user-authored `mcp.jira` + no Claude source → entry survives*, which is the case the current suite doesn't cover.

---

### Security Review

**Strengths:**

- **The credential fix is complete and correctly reasoned.** `write` chmods `0600` on both the register and remove paths, and the doc comment explains *why* `.atomic` makes it mandatory rather than defensive — future readers won't strip it as redundant.
- **Refusal semantics preserved.** Source-unparseable and target-unparseable still return `.skippedUnparseable` before any write, so a malformed file is never clobbered (`refusesToTouchUnparseableOpenCodeConfig`).
- **No secret reaches a log.** `NSLog` calls carry paths and `error.localizedDescription`; `LaunchScaffold.swift:122-124` logs the `Outcome` case, whose only payload is a write-error string.
- **No shell injection in the generated plugin.** Every interpolation in `` $`echo ${payload} | ${CROW} hook-event --session ${SESSION} …` `` goes through Bun Shell's escaping; `jsStringLiteral` (`OpenCodeHookConfigWriter.swift:227-232`) escapes `\` and `"` before `crowPath` reaches the JS literal.
- **`--session` is a real, validated flag.** `HookEventCommand.swift:21-33` accepts it and validates it as a UUID, and `AgentKind.swift:24` confirms `opencode` is the right `--agent` value — the per-project emit resolves as documented.
- **`permission.ask` stays observe-only** — the hook never sets `output.status`.
- **The self-suppress guard is sound for the paths it covers.** Gated behind `if (!SESSION)`, so a session-scoped plugin can never suppress itself, and both variants are asserted (`globalPluginSourceSelfSuppressesWhenPerProjectExists`, `perProjectPluginSourceBakesInSessionUUID`).

**Concerns:** the credential-deletion half of 🟡 1 — the removal path takes the user's token with it.

---

### Code Quality

**🟢 The `<1.18` comments in `OpenCodeLaunchArgs` now contradict the fixed predicate.** `:92-96` still reads *"on builds older than `tuiAutoReintroducedVersion` the TUI `--auto` flag is known absent"*, and `:107-108` *"Known `<1.18`: the top-level `--auto` flag does not exist"* — both describe the pre-fix behavior. Round 1 flagged the comment/code disagreement in this exact function; it's now inverted rather than resolved. `:55-59` and `:17-30` are correct, so it's just these two.

**🟢 Step 1's comment promises a distinction the code doesn't make.** `:71-73` says untranslatable input means *"leave the mirror alone"*, but line 98 folds `translateClaudeServer(...) == nil` into the same branch as "absent", so a half-edited `mcpServers.jira` (say, momentarily carrying only `env`) deletes the mirror rather than leaving it. Low impact on its own — it mainly makes the deletion path in 🟡 1 easier to trigger.

**🟢 MCP registration is nested under `if let crowPath`.** `LaunchScaffold.swift:105,119` — the mirror needs the OpenCode config home, not the `crow` binary. When `findCrowBinary` returns nil, Jira registration silently never runs (and never un-mirrors), for a reason unrelated to MCP. Moving it outside the `if let` is a one-line change.

**🟢 `.opencode/plugins/crow-hooks.js` is untracked inside the user's repo** (carry-over from round 1, unaddressed and still Green). It now carries the session UUID *and* the absolute `crow` path, and an agent running `git add -A` in its own worktree can commit it. Same exposure class as `.claude/settings.local.json`; a one-line append to `.git/info/exclude` beside the write would close it for both.

**🟢 `removeHookConfig` is still unreachable in production** (carry-over). `SessionService.swift:1707` remains hardcoded to `ClaudeHookConfigWriter()`, so the new per-project remove and its two tests are exercised only by the suite. Harmless today (worktree teardown removes the tree wholesale), but the empty-dir pruning at `OpenCodeHookConfigWriter.swift:89-90` never runs, and `writeHookConfig` already routes through `agent.hookConfigWriter` at `:908`/`:1434` — the asymmetry is the thing that will drift.

**Test coverage is strong.** 49 tests, and the three new suites lock in exactly what round 1 asked for: the `[1.17.0, 1.18.0)` boundary from both sides, the `0600` mode, the un-mirror round trip including the follow-up `.noSource`, and the `enabled: false` opt-out surviving a launch. The one gap lines up with 🟡 1 — no case where `mcp.jira` exists without a Claude source behind it.

---

### Summary Table

| # | Color | Finding | Location |
|---|-------|---------|----------|
| 1 | 🟡 | Un-mirror deletes any `mcp.jira`, including a user-authored one, with no provenance record | `OpenCodeMCPConfigWriter.swift:97-107` |
| 2 | 🟢 | `<1.18` comments now contradict the corrected `[1.17.0, 1.18.0)` predicate | `OpenCodeLaunchArgs.swift:92-96,107-108` |
| 3 | 🟢 | "Untranslatable → leave alone" documented but not implemented | `OpenCodeMCPConfigWriter.swift:71-73` vs `:98` |
| 4 | 🟢 | MCP mirror gated on `crowPath`, which it doesn't use | `LaunchScaffold.swift:105,119` |
| 5 | 🟢 | Untracked `.opencode/` plugin committable from a worktree (carry-over) | `OpenCodeHookConfigWriter.swift:70-76` |
| 6 | 🟢 | OpenCode `removeHookConfig` never called in production (carry-over) | `SessionService.swift:1707` |

Round 1's three Yellows are properly closed — the version-window narrowing and the `0600` write are exactly right, and the `enabled` carry-over is a neater solution than the compare-on-everything-except approach I suggested. 🟡 1 is the "when Crow previously wrote it" half of finding 3 that didn't make it into the implementation; a small provenance record closes it and makes the overwrite path safe at the same time. The two comment fixes (#2, #3) are the kind that cost nothing now and mislead a reader later.

---

dgershman:
Round 2 addressed in `902877ab` — the 🟡 and four of five Greens.

**🟡 1 — un-mirror/overwrite provenance.** You're right that `~/.config/opencode/opencode.json` is exactly where an OpenCode-primary user's own `jira` lands, so "Crow owns this key" was an unsafe assumption. Implemented the sidecar you sketched: Crow records the exact entry it writes in `~/.local/share/crow/opencode-mcp-mirror.json` (`0600`, its own state dir) and only removes/overwrites `mcp.jira` when the on-disk entry still matches that record. A user-authored entry (no record) or a user-edited one (record mismatch) is left completely alone — new **`.skippedUserOwned`** outcome, on **both** the register and remove paths. That also subsumes the `enabled: false` opt-out (editing makes it "not ours"), so I dropped the special-case carry-over. New tests `doesNotOverwriteUserAuthoredEntry` (register path) and `preservesUserAuthoredEntryOnSourceRemoval` (your exact failure scenario); all lifecycle tests now thread an injected record path so they never touch real state.

**🟢 2 — stale `<1.18` comments.** Fixed both to describe the `[1.17.0, 1.18.0)` window.

**🟢 3 — untranslatable ≠ absent.** A present-but-untranslatable Claude `jira` now returns `.noSource` and leaves the mirror alone, instead of falling into the delete branch — matches the step-1 comment.

**🟢 4 — MCP gated on crowPath.** Moved out of the `if let crowPath` block; it uses the config home, not the binary, so it no longer silently no-ops (or skips un-mirror) when `findCrowBinary` returns nil.

**🟢 5 — `.opencode/` committable.** `writeHookConfig` now writes a self-scoped `.opencode/plugins/.gitignore` (ignores `crow-hooks.js` + itself), so `git add -A` in a worktree can't stage the generated plugin. Verified in a throwaway repo — `git add -A` stages nothing. `removeHookConfig` cleans the `.gitignore` up when Crow owns the dir. (Scoped to our own dir so it doesn't fight OpenCode's `.opencode/.gitignore` a level up.)

**🟢 6 — removeHookConfig routing.** Deferred again, same reasoning as round 1: the sole call site is a `nonisolated static performDiskCleanup(for:)` over worktree items with no `agentKind` in scope, and the worktree is `git worktree remove --force`'d on the next line, so the empty-dir pruning is moot in production. Routing it through `agent.hookConfigWriter` means threading agentKind through that static path — happy to do it as a focused follow-up if you'd like, but it's zero-behavior-change today.

50 tests / 5 suites green; `swift build --package-path Packages/CrowDaemon` clean. (The earlier Linux CI red — `NSDictionary.isEqual` not deep-equating `[String]` vs a JSON `NSArray` — was fixed in `cd4d8fe9` via canonical-JSON comparison, and that run went green before this push.)

dhilgaertner (changes requested):
## Code & Security Review — round 3

Verified locally on `902877a`: `swift test --package-path Packages/CrowOpenCode` → **50 tests / 5 suites pass**; `swift build --package-path Packages/CrowDaemon` → **Build complete**. No failures.

**Round-2 findings: all six addressed or reasonably deferred.**

| Round 2 | Status | Where |
|---|---|---|
| 🟡 1 un-mirror/overwrite with no provenance | ✅ Fixed — sidecar record + `entryIsOurs`, guarding **both** the remove and register paths | `OpenCodeMCPConfigWriter.swift:106-110,122-125,149-152`; `doesNotOverwriteUserAuthoredEntry`, `preservesUserAuthoredEntryOnSourceRemoval` |
| 🟢 2 stale `<1.18` comments | ✅ Fixed — both now describe `[1.17.0, 1.18.0)` | `OpenCodeLaunchArgs.swift:55-59,91-97,107-109` |
| 🟢 3 untranslatable ≠ absent | ✅ Fixed — returns `.noSource` and leaves the mirror alone | `OpenCodeMCPConfigWriter.swift:140-145` |
| 🟢 4 MCP gated on `crowPath` | ✅ Fixed — moved out of the `if let` | `LaunchScaffold.swift:105-128` |
| 🟢 5 `.opencode/` committable | ⚠️ Addressed — self-scoped `.gitignore`; works, but see 🟡 1 below | `OpenCodeHookConfigWriter.swift:81,149-157` |
| 🟢 6 `removeHookConfig` routing | ✅ Accepted as deferred — I confirmed `SessionService.swift:1707` is followed by `git worktree remove --force` at `:1711`, and `isMainCheckout` `continue`s before either, so Claude and OpenCode behave identically. Zero prod delta; dropping it as a carry-over. |

The provenance sidecar is the right shape. Recording the exact entry rather than a bare "Crow wrote this" flag means a user edit *is* the opt-out with no extra state, the fail-open direction on a missing/corrupt record is the safe one (`:209-211`), and putting it in `~/.local/share/crow/` keeps it clear of OpenCode's schema validation. One issue, in the other half of the round-2 work.

### Critical Issues

None.

---

### 🟡 1. `writeGitignore` overwrites a pre-existing `.opencode/plugins/.gitignore`

`Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeHookConfigWriter.swift:149-157` (called unconditionally from `writeHookConfig:81`)

```swift
private static func writeGitignore(inDir dir: String) {
    let path = (dir as NSString).appendingPathComponent(".gitignore")
    let body = """
    # Crow-generated — safe to delete.
    \(pluginFileName)
    .gitignore
    """
    try? body.write(to: URL(fileURLWithPath: path), atomically: true, encoding: .utf8)
}
```

There is no existence check and no provenance check — every OpenCode launch stamps Crow's three lines over whatever `.gitignore` is at that path. `.opencode/plugins/` is a directory **the user owns**: it's where their own OpenCode plugins live, and `removeHookConfigPreservesUserPluginsInDir` already assumes exactly that. A `.gitignore` there listing local-only plugins is an ordinary thing for a team to commit.

This is the same "don't touch what you didn't write" principle the MCP writer just spent round 2 implementing, not applied to the writer added in the same commit.

**Failure scenario** — reproduced in a throwaway repo:

```
$ cat .opencode/plugins/.gitignore      # user's, tracked
my-local-only.js
$ git ls-files
.opencode/plugins/.gitignore
.opencode/plugins/team-plugin.js

# ...Crow launches an OpenCode session in this worktree...

$ git add -A && git diff --cached --name-status
M	.opencode/plugins/.gitignore          # user's rules replaced by Crow's
A	.opencode/plugins/my-local-only.js    # was ignored; now staged
```

Two distinct harms, both silent: the user's ignore rules are destroyed (unrecoverable if that `.gitignore` was itself untracked), and files they had deliberately excluded become stageable — the exact outcome this change exists to prevent, just aimed at a different file. `crow-hooks.js` *is* correctly excluded, so the mechanism works; it's only the unconditional write that's wrong.

**Suggested fix** — write only when the path is free or already carries our content, mirroring the `entryIsOurs` check you just added on the MCP side:

```swift
private static func writeGitignore(inDir dir: String) {
    let path = (dir as NSString).appendingPathComponent(".gitignore")
    let body = "# Crow-generated — safe to delete.\n\(pluginFileName)\n.gitignore\n"
    // Never clobber a user's own ignore rules — `.opencode/plugins/` is their
    // directory. Only write into an empty slot or over our own previous body.
    if let existing = try? String(contentsOfFile: path, encoding: .utf8), existing != body {
        return
    }
    try? body.write(to: URL(fileURLWithPath: path), atomically: true, encoding: .utf8)
}
```

The same guard belongs on the removal side (`:96-101`) so a user-authored `.gitignore` that happens to be the sole remaining entry isn't deleted either. Worth a test alongside `removeHookConfigPreservesUserPluginsInDir`: *pre-existing `.gitignore` + `writeHookConfig` → content unchanged*.

---

### Security Review

**Strengths:**

- **Provenance closes the deletion hole completely.** `entryIsOurs` (`:110`) gates both the un-mirror (`:122-125`) and the overwrite (`:149-152`), and canonical-JSON equality (`:309-314`) makes the comparison stable on Linux CI — the `NSDictionary.isEqual` note at `:301-308` is a genuinely useful thing to have written down.
- **Fail-open is pointed the safe way.** A missing or corrupt record yields `[:]`, so every entry reads as "not ours" and Crow neither deletes nor overwrites. A stale mirror lingering is strictly the smaller harm, and `:209-211` says so explicitly.
- **Both credential-bearing files are `0600`.** `write` (`:179-196`) and `writeMirrorRecord` (`:225-236`) each re-apply the mode after `.atomic`, with the rationale in the doc comment so a later reader won't strip it as redundant. `writesTargetOwnerOnly` asserts both paths.
- **Refusal semantics intact.** Source-unparseable and target-unparseable still return `.skippedUnparseable` before any write (`:88`, `:98-99`); a momentarily-unreadable `~/.claude.json` can no longer trigger the delete branch.
- **No secret reaches a log.** `NSLog` carries paths, server names, and `error.localizedDescription`; `LaunchScaffold.swift:123-128` logs the `Outcome` case, whose only payload is a write-error string.
- **No shell injection in the generated plugin.** Every interpolation in `` $`echo ${payload} | ${CROW} hook-event --session ${SESSION} …` `` goes through Bun Shell's escaping; `jsStringLiteral` (`:254-259`) escapes `\` and `"` before `crowPath` reaches the JS literal. `permission.ask` stays observe-only.

**Concerns:** none new on the credential path. 🟡 1 is a repo-hygiene regression, not a secret leak.

---

### Code Quality

**🟢 The sidecar stores the plaintext credential a third time.** `writeMirrorRecord` (`:225-236`) persists the full translated entry — `environment`/`headers` included — so a Jira API token now sits in `~/.claude.json`, `~/.config/opencode/opencode.json`, *and* `~/.local/share/crow/opencode-mcp-mirror.json`. All three are `0600`, so this isn't a leak, but provenance only needs *equality*: storing `SHA256(canonicalJSON(entry))` gives identical semantics with one less copy of the secret on disk, and makes a leaked/backed-up state dir harmless. Cheap to swap — `jsonEqual` becomes a digest comparison and nothing else moves.

**🟢 Deleting `mcp.jira` isn't a durable opt-out.** `:149` guards the overwrite with `existing != nil, !entryIsOurs`, so when the user removes the entry outright (`existing == nil`) the stale record is ignored and the mirror is re-added at `:161-168` on the next launch. The doc comment (`:34-35`) names editing as the opt-out, and that works — but deleting is the more obvious gesture and silently doesn't stick. `if existing == nil, recorded != nil { return .skippedUserOwned }` would make both gestures behave the same.

**🟢 The Crow `.gitignore` outlives Crow when a user plugin shares the dir.** `removeHookConfig:99` only deletes it when `contentsOfDirectory == [".gitignore"]`, so in the `removeHookConfigPreservesUserPluginsInDir` scenario the user keeps a Crow-authored `.gitignore` that ignores `crow-hooks.js` and itself, forever. Harmless, but it contradicts "leaves no trace" at `:87-89`, and the test doesn't assert either way.

**🟢 The record re-sync at `:153-160` is unreachable.** Reaching it requires `existing != nil` and `entryIsOurs`, which means `recorded == existing`; the branch is entered only when `existing == translated`, so `recorded == translated` always holds and the inner `if` never fires. Not a bug — just a few lines that imply a state the guards above have already excluded.

**Test coverage matches the work.** 50 tests. The two new MCP cases are precisely the round-2 failure scenarios (user-authored entry survives both the register and the source-removal path), and threading `mirrorRecordPath` through every lifecycle test so none of them touch real state is the right call. The gap lines up with 🟡 1: no case exercises `writeHookConfig` against a directory that already has a `.gitignore`.

---

### Summary Table

| # | Color | Finding | Location |
|---|-------|---------|----------|
| 1 | 🟡 | `writeGitignore` unconditionally overwrites a user's `.opencode/plugins/.gitignore`, un-ignoring their files | `OpenCodeHookConfigWriter.swift:149-157` |
| 2 | 🟢 | Sidecar stores the plaintext credential a third time; a digest would do | `OpenCodeMCPConfigWriter.swift:225-236` |
| 3 | 🟢 | Deleting `mcp.jira` isn't a durable opt-out — re-added every launch | `OpenCodeMCPConfigWriter.swift:149,161-168` |
| 4 | 🟢 | Crow's `.gitignore` is left behind when a user plugin shares the dir | `OpenCodeHookConfigWriter.swift:96-101` |
| 5 | 🟢 | Unreachable record re-sync branch | `OpenCodeMCPConfigWriter.swift:153-160` |

Round 2's Yellow is properly closed — the sidecar is a better design than the flag I sketched, because recording the *entry* makes the opt-out fall out for free instead of needing a special case. 🟡 1 is the one place the same ownership discipline didn't reach: the `.gitignore` fix for Green #5 solves the stated problem but takes a user file with it. A four-line existence check closes it, and #2 (digest instead of plaintext) is worth folding into the same pass while the file is open.

---

dgershman:
Round 3 addressed in `41d53cae` — the 🟡 plus all four Greens.

**🟡 1 — `.gitignore` ownership.** You're right that this was the one place the round-2 discipline didn't reach. `writeGitignore` now writes only into an empty slot or over our own exact previous body (single source of truth: `gitignoreBody`), and `removeHookConfig` deletes it only when it still holds our content — so a user's hand-written `.opencode/plugins/.gitignore` (and any files it ignores) is never touched. New tests `writeHookConfigPreservesUserGitignore` (write + remove both leave the user's rules intact) and `writeHookConfigWritesOwnGitignoreIntoEmptyDir`.

**🟢 2 — digest, not plaintext.** The sidecar now stores `SHA256(canonicalJSON(entry))` instead of the entry, so the Jira token lives in `~/.claude.json` + `opencode.json` only — a leaked/backed-up state dir is harmless. Wired swift-crypto into CrowOpenCode (already resolved repo-wide via CrowDaemon's PBKDF2, so no new external dep). `canonicalDigest` replaces `jsonEqual`; the record is still `0600` (keeps the "which servers Crow manages" list owner-only).

**🟢 3 — deleting is a durable opt-out.** `if existing == nil, recordedDigest != nil → .skippedUserOwned`, so removing `mcp.jira` outright now sticks the same way editing does. New test `deletingEntryIsDurableOptOut`.

**🟢 4 — no leftover `.gitignore`.** `removeHookConfig` now drops our `.gitignore` whenever it's ours (not only when it's the sole entry), so teardown leaves no Crow trace even when a user plugin shares the dir. Verified by `writeHookConfigPreservesUserGitignore`'s remove step (user's survives) and the existing prune test (ours goes).

**🟢 5 — dead branch removed.** You're right the re-sync was unreachable; with the digest guards, reaching the `existing == translated` case means the record already matches, so I dropped it and just `return .unchanged`.

**🟢 6** — thanks for confirming the `git worktree remove --force` at `:1711` makes the routing a zero-delta; nothing changed there.

53 tests / 5 suites green; `swift build --package-path Packages/CrowDaemon` clean (the crypto dep resolves through the existing graph). Watching CI on the dependency addition.

dhilgaertner (changes requested):
## Code & Security Review — round 4

Verified locally on `41d53ca`: `swift test --package-path Packages/CrowOpenCode` → **53 tests / 5 suites pass**; `swift build --package-path Packages/CrowDaemon` → **Build complete**. Both green on macOS.

**Linux CI is red on this commit**, and that's the blocker below.

**Round-3 findings: all five closed.**

| Round 3 | Status | Where |
|---|---|---|
| 🟡 1 `writeGitignore` clobbers a user's `.gitignore` | ✅ Fixed — writes only into an empty slot or over `gitignoreBody`; `removeHookConfig` deletes only its own | `OpenCodeHookConfigWriter.swift:162-170`, `:100-103`; `writeHookConfigPreservesUserGitignore`, `writeHookConfigWritesOwnGitignoreIntoEmptyDir` |
| 🟢 2 plaintext credential in the sidecar | ✅ Implemented — `canonicalDigest` stores SHA-256 hex, token no longer copied | `OpenCodeMCPConfigWriter.swift:325-329` — **but see 🔴 1** |
| 🟢 3 deleting isn't a durable opt-out | ✅ Fixed — `existing == nil, recordedDigest != nil → .skippedUserOwned` | `:156-159`; `deletingEntryIsDurableOptOut` |
| 🟢 4 leftover `.gitignore` | ✅ Fixed — dropped whenever the content is ours, not only when it's the sole entry | `:100-103` |
| 🟢 5 unreachable re-sync branch | ✅ Removed | `:167-183` |

The single-source-of-truth `gitignoreBody` is the right way to make write and remove agree on "ours", and the register/remove state machine now covers all nine (source × entry × record) combinations correctly — I walked each one. The problem is the dependency the digest brought in.

### Critical Issues

---

### 🔴 1. Adding swift-crypto to `CrowOpenCode` breaks the Linux test link — CI is red

`Packages/CrowOpenCode/Package.swift:15,22`

```swift
.package(url: "https://github.com/apple/swift-crypto.git", "3.0.0" ..< "5.0.0"),
```

The commit message justifies this as free: *"Already resolved repo-wide via CrowDaemon's web-auth PBKDF2 (CROW-593)"*. Resolution, yes — **linking, no**. `CrowDaemon` is not in the Linux matrix:

```yaml
# .github/workflows/ci.yml:58-59
LINUX_PACKAGES=(CrowCore CrowIPC CrowClaude CrowCodex CrowCursor \
                CrowGit CrowOpenCode CrowPersistence CrowProvider CrowCLI)
```

`CrowOpenCode` *is*. So this PR is the first time swift-crypto has ever been linked on Linux in this repo, and it fails:

```
[27/29] Write Objects.LinkFileList
error: link command failed with exit code 1 (use -v to see invocation)
/usr/lib/swift/linux/libswiftObservation.so: error: undefined reference to 'swift::threading::fatal(char const*, ...)'
clang: error: linker command failed with exit code 1
[28/29] Linking CrowOpenCodePackageTests.xctest
error: fatalError
```

Clean before/after on the branch — the only dependency change between them is this one:

| Commit | Linux CI |
|---|---|
| `37428b6` (initial) | ✅ success |
| `1eb9634` (round 1) | ❌ failure (the `NSDictionary.isEqual` bug) |
| `cd4d8fe` (canonical JSON) | ✅ success |
| `902877a` (round 2) | ✅ success |
| `41d53ca` (round 3 — **adds swift-crypto**) | ❌ failure |

**Why it slipped, and why local verification won't catch it.** `swift build --package-path Packages/CrowOpenCode` **succeeds** on Linux — the CI log shows `Build complete! (48.92s)` — and only the subsequent `swift test` fails, at link time for the xctest bundle. So the commit message's `swift build --package-path Packages/CrowDaemon → Build complete` proves nothing about this, and macOS `swift test` passes because it links against CryptoKit rather than BoringSSL. The break is only reachable via `swift test` on Linux.

**Mechanism.** swift-crypto 4.5.1 drags a large C/C++/assembly graph into the package (visible in the log: `fiat_p256_adx_mul.S`, `md5-x86_64-linux.S`, `KeccakHash.c`, `CryptoBoringWrapper`, plus a transitive `swift-asn1` 1.7.1). That switches SwiftPM's link driver to `clang` — note it's `clang:` reporting the failure, not `swiftc` — and on this toolchain that path doesn't resolve `swift::threading::fatal` out of libswiftCore for the `libswiftObservation.so` that swift-testing pulls in. Deterministic, not a flake: it's an undefined-reference at link, and the two preceding commits linked the same test bundle fine.

**Suggested fix — revert the digest.** Green #2 was a "consider", not a "must", and it has turned out to cost a dependency that breaks the build. `902877a`'s `jsonEqual` canonical-JSON comparison was correct, cross-platform, and green; the only thing it gave up was keeping the entry (and its token) out of the sidecar — a file already at `0600` in Crow's own state dir, which is why I filed it Green rather than Yellow. Reverting `canonicalDigest` → `jsonEqual` and dropping the `Package.swift` change restores a known-green state and loses nothing that was blocking.

If you'd rather keep the digest, the dependency-free option is a ~50-line SHA-256 in `CrowOpenCode` (or `CrowCore`). The provenance fingerprint isn't a security boundary — it needs one-wayness so the token isn't recoverable from the sidecar, and enough collision resistance that a user-edited entry never hashes to the recorded value. A vendored implementation gives both with no link-graph change. What won't work is moving the helper to `CrowCore` while keeping swift-crypto — `CrowCore` is in `LINUX_PACKAGES` too, so it inherits the same failure.

Either way, please confirm against Linux CI rather than a local macOS run — this is the second Linux-only break on the branch (after `1eb9634`), and both were invisible on macOS.

---

### Security Review

**Strengths:**

- **The ownership discipline is now uniform across both writers.** `writeGitignore` (`:162-170`) and `removeHookConfig` (`:100-103`) apply the same "only touch our own exact content" rule the MCP writer uses, against a single `gitignoreBody` constant — so the two paths can't drift on what "ours" means. `writeHookConfigPreservesUserGitignore` asserts both directions (write leaves the user's rules intact; remove doesn't delete them).
- **The MCP state machine is sound across every reachable state.** I enumerated source (absent / untranslatable / translatable) × entry (nil / ours / theirs) × record (nil / present). Every path that writes or deletes requires `entryIsOurs`, i.e. the on-disk entry hashes to what Crow recorded; every other path returns without touching the file. No combination deletes or overwrites a user-authored entry.
- **Fail-open still points the safe way.** `readMirrorRecord` yields `[:]` on a missing or corrupt sidecar (`:228-235`), so every entry reads as "not ours" and Crow neither deletes nor overwrites.
- **Both credential-bearing files are `0600`.** `write` (`:194-211`) and `writeMirrorRecord` (`:242-253`) each re-apply the mode after `.atomic`, with the rationale in the doc comment. `writesTargetOwnerOnly` asserts both.
- **Refusal semantics intact.** Source- and target-unparseable both return `.skippedUnparseable` before any write (`:91`, `:98-103`).
- **No secret reaches a log.** `NSLog` carries paths, server names, and `error.localizedDescription`; `LaunchScaffold.swift:123-128` logs the `Outcome` case, whose only payload is a write-error string. The sidecar now holds digests only.
- **No shell injection in the generated plugin.** Every interpolation in `` $`echo ${payload} | ${CROW} hook-event --session ${SESSION} …` `` goes through Bun Shell's escaping; `jsStringLiteral` (`:267-272`) escapes `\` and `"` before `crowPath` reaches the JS literal. `permission.ask` stays observe-only — the hook never sets `output.status`.
- **`installedVersion`'s double-optional cache is handled correctly.** `versionCache: [String: SemVer?]` with `if let cached = versionCache[binary]` unwraps exactly one level, so a cached "unparseable" (`.some(nil)`) is a cache hit rather than a re-probe. And `tuiSupportsAuto` releases `cacheLock` before calling `installedVersion`, which re-acquires it — correct, since `NSLock` isn't recursive.

**Concerns:** none new on the credential path. 🔴 1 is a build break, not a security issue.

---

### Code Quality

**🟢 An orphaned Crow `.gitignore` is invisible to `git status`.** `removeHookConfig:95` early-returns when `crow-hooks.js` is absent, so the `.gitignore` cleanup at `:100-103` never runs in that case. The plugin header literally says *"auto-generated by Crow (safe to delete)"* — a user who takes that up leaves Crow's `.gitignore` behind permanently, and because it ignores itself, `git status` won't surface it. Moving the `.gitignore` check above the guard (or making the guard `plugin || gitignore exists`) closes it. Contradicts "leaves no trace" at `:87-90`.

**🟢 The durable opt-out is erased by the source-absent GC.** `:120-127` drops the record whenever the entry is gone *and* the Claude source is gone; `:156-159` relies on that record surviving to honor a manual delete. So: user deletes `mcp.jira` (opt-out recorded and honored) → their Claude `jira` MCP later goes away or `~/.claude.json` is briefly absent → record cleared → `jira` comes back → the mirror is silently re-added. The doc comment at `:33-37` promises *"never re-added after a manual delete"*. Narrow (needs the source to round-trip) and the harm is only an unwanted server, but the two branches disagree about what the record means.

**🟢 A lost sidecar permanently disowns Crow's own entry, and says so misleadingly.** If the record is deleted or corrupted, `entryIsOurs` goes false and `:162-165` returns `.skippedUserOwned` on every subsequent launch — so a Jira token rotated in `~/.claude.json` never propagates and OpenCode keeps using the stale one. Fail-open is the right default; the log line is the problem: `"mcp.jira not written by Crow; not overwriting"` is false in this case (Crow did write it — the record was lost), which points a debugging user at the wrong file. Distinguishing "no record at all" from "record mismatch" in the message would make it self-diagnosing.

**🟢 A user's own `.gitignore` in the slot means `crow-hooks.js` goes back to being stageable.** `writeGitignore:164-168` correctly declines to clobber, but then nothing ignores the generated plugin for that user — the round-2 problem returns for exactly the people who already curate that directory. An append of a marker-delimited block (`# >>> crow` … `# <<< crow`), or falling back to the worktree's `$GIT_COMMON_DIR/info/exclude`, would cover both cases. Genuinely a trade-off, not a defect — noting it because the current behavior is silent either way.

**🟢 One test still reaches for the real sidecar path.** `OpenCodeMCPConfigWriterTests.swift:155-156` calls `installGlobalMCPConfig(configHome:claudeJSONPath:)` without `mirrorRecordPath`, unlike every other lifecycle test. It's safe *today* only because `.skippedUnparseable` returns at `:102`, before `recordPath` is computed at `:112` — so a future reordering of those two steps would have the suite writing the developer's real `~/.local/share/crow/opencode-mcp-mirror.json`. Threading the temp path through like its siblings removes the latent coupling.

**Test coverage matches the work.** 53 tests. The two new `.gitignore` cases are exactly the round-3 failure scenario (user's rules survive both write and remove), `deletingEntryIsDurableOptOut` covers the new opt-out gesture, and `tuiAutoKnownAbsentOnlyForRemovedWindow` pins the `[1.17.0, 1.18.0)` boundary from both sides plus `nil`. The gap is the one that matters: nothing in the suite runs on Linux locally, and the macOS run can't reach the failure in 🔴 1.

---

### Summary Table

| # | Color | Finding | Location |
|---|-------|---------|----------|
| 1 | 🔴 | swift-crypto breaks the Linux test link; CI red on head, green on the parent commit | `Packages/CrowOpenCode/Package.swift:15,22` |
| 2 | 🟢 | Orphaned Crow `.gitignore` when the plugin is deleted first — and it's invisible to `git status` | `OpenCodeHookConfigWriter.swift:95` |
| 3 | 🟢 | Source-absent GC erases the durable opt-out record | `OpenCodeMCPConfigWriter.swift:120-127` vs `:156-159` |
| 4 | 🟢 | Lost sidecar permanently disowns Crow's entry; log message misattributes it | `OpenCodeMCPConfigWriter.swift:162-165` |
| 5 | 🟢 | User-owned `.gitignore` leaves `crow-hooks.js` stageable again | `OpenCodeHookConfigWriter.swift:162-170` |
| 6 | 🟢 | One test omits `mirrorRecordPath`; safe only by statement order | `OpenCodeMCPConfigWriterTests.swift:155-156` |

**Recommendation:** Request Changes — driven by [1 Red, 0 Yellow, 5 Green] findings.

The review findings are all closed — the `.gitignore` ownership fix is exactly right, and the register/remove state machine holds up under a full case-by-case walk, which is the part I most expected to find a hole in. The Red isn't a logic error; it's that Green #2 turned out to have a dependency cost that only shows up on Linux `swift test`. Reverting to `902877a`'s `jsonEqual` gets you back to green and gives up nothing that was blocking — that was a "consider", and it isn't worth a broken build. With that undone (or a vendored SHA-256 in its place) and CI green, this is ready.

---

dgershman:
CI heads-up on `8dbc8a52`: the round-3 push went red on **Build & Test (Linux)** — not a logic failure, a link one. swift-crypto compiles on the runner, but linking its BoringSSL C++ interop into the CrowOpenCode **test** binary trips a Swift-on-Linux runtime-lib bug: `libswiftObservation.so: undefined reference to 'swift::threading::fatal(...)'`.

Since Green #2 (digest instead of plaintext) was a "consider", I've reverted the swift-crypto dependency and the digest, and restored the full-entry provenance record — round 2's design, which you'd accepted, using the cross-platform `jsonEqual`. The token sits in the `0600` sidecar as well as `~/.claude.json`/`opencode.json`, i.e. the duplicated-secret note you'd already called "not a leak". If you'd like the third copy gone, a self-contained pure-Swift SHA-256 (no C++ interop, no link issue) would do it without the dependency — happy to add that as a follow-up if it's worth the ~50 lines.

Everything else from round 3 stayed: the 🟡 `.gitignore` ownership fix, the durable delete opt-out (#3), the leave-no-trace `.gitignore` cleanup (#4), and the removed dead branch (#5), with their tests. 53 tests / 5 suites green locally.

dgershman:
Round 4 addressed. First: the 🔴 is already gone — round-4 reviewed `41d53ca` (with swift-crypto), but I'd reverted it in `8dbc8a52` right after CI went red, exactly the fix you prescribed (`canonicalDigest` → `jsonEqual`, drop the `Package.swift` dep). That commit is **green on all checks including Build & Test (Linux)** — thanks for the precise root-cause (Linux `swift test` link, macOS uses CryptoKit, `CrowDaemon` not in `LINUX_PACKAGES`); it matched what the CI log showed.

The Greens, in `47e93a38`:

**🟢 2 — orphaned `.gitignore`.** `removeHookConfig` no longer early-returns when `crow-hooks.js` is already gone, so a hand-delete of the plugin doesn't strand a self-ignoring `.gitignore`. New test `removeHookConfigCleansOrphanedGitignoreAfterManualPluginDelete`.

**🟢 3 — opt-out erased by GC.** Good catch that the two branches disagreed on what the record means. The source-absent + entry-gone branch no longer clears the record (it's the opt-out marker), so deleting `mcp.jira` survives a Claude-source round-trip. Extended `deletingEntryIsDurableOptOut` with vanish→return.

**🟢 4 — misleading log.** The "not written by Crow" line now distinguishes "no record (user-authored or sidecar lost)" from "record mismatch (user-edited)", so a lost-sidecar case doesn't point a debugger at the wrong file. Log-only; fail-open behavior unchanged.

**🟢 6 — test coupling.** `refusesToTouchUnparseableOpenCodeConfig` now threads `mirrorRecordPath` like its siblings.

**🟢 5 — user-owned `.gitignore` leaves crow-hooks.js stageable.** Deferring this one with reasoning rather than code: both closes you sketched have a real downside — appending to the user's *tracked* `.gitignore` risks them committing our block, and `$GIT_COMMON_DIR/info/exclude` needs a subprocess in the launch hot path. The common case (Crow owns the dir) is covered, and `crow-hooks.js` sits under `.opencode/`, which curators of that dir typically ignore wholesale. If you'd like it fully closed I'll add the marker-delimited block (`# >>> crow` … `# <<< crow`) with matching teardown as a focused follow-up — it's the right shape, just more surface than I want to add in a review round that's otherwise green.

54 tests / 5 suites green; CrowDaemon builds. Confirming against Linux CI on this push.

dhilgaertner (approved):
## Code & Security Review — round 5

Verified locally on `47e93a3`: `swift test --package-path Packages/CrowOpenCode` → **54 tests / 5 suites pass**; `swift build --package-path Packages/CrowDaemon` → **Build complete**.

**Linux CI is green — the round-4 blocker is gone.** I checked the job log rather than the badge, because the round-4 break was invisible to `swift build` and only surfaced when the xctest bundle linked:

```
# run 30131413291, job 89606511477 — Build & Test: pass (3m48s)
==> Building & testing CrowOpenCode on Linux...
Build complete! (7.50s)      # swift build
Build complete! (6.11s)      # swift test — this is the link that failed on 41d53ca
Test Suite 'All tests' passed at 2026-07-24 22:38:36.984
```

No `swift-crypto`, no `CryptoBoringWrapper`, no `libswiftObservation` undefined reference anywhere in the graph. All three checks (Build & Test, Desktop (Tauri) Build, Shell Lint) pass.

**Round-4 findings: all six closed.**

| Round 4 | Status | Where |
|---|---|---|
| 🔴 1 swift-crypto breaks the Linux test link | ✅ Fixed — dependency and `canonicalDigest` reverted; sidecar back to `jsonEqual` on the recorded entry | `Package.swift:12`, `OpenCodeMCPConfigWriter.swift:331-336`; **Linux CI green on head** |
| 🟢 2 orphaned self-ignoring `.gitignore` | ✅ Fixed — the `fileExists(pluginPath)` early return is gone, so the cleanup runs after a manual plugin delete | `OpenCodeHookConfigWriter.swift:95-108`; `removeHookConfigCleansOrphanedGitignoreAfterManualPluginDelete` |
| 🟢 3 source-absent GC erases the opt-out | ✅ Fixed — the entry-already-gone branch keeps the record | `OpenCodeMCPConfigWriter.swift:121-127`; `deletingEntryIsDurableOptOut` extended with the source round-trip |
| 🟢 4 "not written by Crow" misattributes a lost sidecar | ✅ Fixed — 4d distinguishes no-record from record-mismatch | `:163-173` |
| 🟢 5 user-owned `.gitignore` leaves the plugin stageable | ✅ Accepted as deferred — I filed it as a trade-off, and the reply picks the side that never touches a user's tracked file. Dropping it as a carry-over. |
| 🟢 6 test omits `mirrorRecordPath` | ✅ Fixed — threaded like its siblings | `OpenCodeMCPConfigWriterTests.swift:155-158` |

The revert is exactly the prescribed fix and nothing else came back with it: I diffed `8dbc8a5` against `902877a`'s known-green state machine and the only deltas are the digest→entry swap and the doc comments. The four independent round-3 wins (`.gitignore` ownership, durable opt-out, leave-no-trace cleanup, dead re-sync branch removed) all survived the revert with their tests, which is the part a revert usually gets wrong.

### Critical Issues

None.

---

### Security Review

**Strengths:**

- **The credential trade-off landed where it was filed.** The token is once again in the sidecar as well as `~/.claude.json` and `opencode.json`, but all three are `0600` (`write:212`, `writeMirrorRecord:254`, both asserted by `writesTargetOwnerOnly:195-199`), and the sidecar lives in Crow's own state dir rather than a config path. That is the duplicated-secret note I filed as Green in round 3, not an exposure — and the doc comment at `:40-44` now records *why* the digest isn't there, so the next reader doesn't re-add the dependency.
- **The provenance state machine is unchanged in substance.** `jsonEqual` (`:331-336`) is the same canonical-JSON comparison that was green in `902877a`; swapping the digest back out only changes *what* is compared, not *when*. I re-walked source (absent / untranslatable / translatable) × entry (nil / ours / theirs) × record (nil / present): every write or delete still requires `entryIsOurs`, and a non-serializable value makes `jsonEqual` return `false`, which reads as "not ours" — the safe direction.
- **Green #3's fix doesn't open a delete path.** Keeping the record at `:121-127` only affects the branch where there is nothing on disk to touch (`existing == nil`), so it can't cause a deletion; and `.removed` (`:142-143`) still clears the record, so a Crow-initiated un-mirror stays re-mirrorable.
- **The new log line is injection-safe.** `NSLog("… mcp.%@ left as-is — %@; not overwriting", serverName, why)` (`:171`) has a literal format string with matching arity, and `why` is one of two static strings — no secret and no user data reaches the format position.
- **Removing the early return didn't widen what `removeHookConfig` deletes.** The `.gitignore` is still gated on exact-content equality (`:104`), and the directory prune is still gated on emptiness (`:144-148`), so a user's own plugin or `.gitignore` in `.opencode/plugins/` keeps both directories alive.

**Concerns:** none. Nothing in this round touches the credential, shell-interpolation, or permission paths.

---

### Code Quality

**🟢 1. The un-mirror log kept the message Green #4 fixed.** `OpenCodeMCPConfigWriter.swift:130` still reads `"mcp.jira not written by Crow; leaving it alone"` — the exact misattribution corrected 40 lines below at `:168-171`. Both branches reach `.skippedUserOwned` for the same two distinguishable reasons (no record at all vs. record disagrees), and a lost sidecar makes the 4a wording false in the same way it made 4d's false. Hoisting the `why` computation above `:120` would give both paths one message.

**🟢 2. The provenance record is keyed by server name, not by target path.** `record[serverName]` (`:116`, `:188`) records *what* Crow wrote but not *where*, while `existing` is read from `<configHome>/opencode.json` (`:95`). So when the config home moves, 4c can't tell "the user deleted my entry" from "I'm looking at a different file":

- Crow mirrors into `~/.config/opencode/opencode.json`; record holds the entry.
- The user later sets `XDG_CONFIG_HOME` (or restores `~/.local/share/crow` onto a machine without `~/.config/opencode`). `configHome` now resolves elsewhere — `LaunchScaffold.swift:107-109`.
- `existing == nil`, `recorded != nil` → 4c (`:157-160`) → `.skippedUserOwned`, logging *"removed after Crow wrote it; honoring the opt-out"*. The mirror is never installed in the new config home, and never will be.

Fail-closed (missing `jira_*` tools, not a bad write) and it needs an uncommon sequence, which is why it's Green — but Green #3's fix removed the record GC that used to provide an accidental way out, so the state is now permanent. Storing the target path alongside the entry, and treating a path mismatch as "no record", would scope the opt-out to the file it was made in.

**🟢 3. Nothing documents how to opt back in.** Once 4c latches, deleting `~/.local/share/crow/opencode-mcp-mirror.json` is the only route back — re-adding `mcp.jira` by hand just lands in 4d as user-authored. The doc comment at `:29-36` describes the opt-out thoroughly and the exit not at all; one clause in the `:158` log ("delete <record path> to re-enable") would make it self-service.

**🟢 4. The 4d record-mismatch branch has no test.** `preservesUserDisabledEnabled` was dropped in `902877a` when provenance subsumed it, and `doesNotOverwriteUserAuthoredEntry:233-265` covers only the `recorded == nil` half. So "user flipped `enabled: false` on the entry Crow wrote" — `recorded != nil` and differs, the branch whose message this commit rewrote — is correct by inspection but unasserted. Re-adding the round-1 case on top of a `.registered` run would cover it in about six lines.

**🟢 5. `removeHookConfig` now prunes `.opencode/` in worktrees Crow never wrote to.** With the guard at `:95` gone, a worktree where `writeHookConfig` never ran (or failed) still reaches `removeIfEmpty(pluginsDir)` / `removeIfEmpty(.opencode)` (`:107-108`), so a user's own empty `.opencode/plugins/` gets deleted. Currently unreachable in production — `SessionService.swift:1707` hard-codes `ClaudeHookConfigWriter()` and the worktree is `git worktree remove --force`d immediately after, the round-3 deferral I accepted — so this only matters if that routing is ever fixed. Worth a `fileExists(pluginPath) || gitignore-is-ours` precondition around the prune at that point.

**Test coverage matches the work.** 54 tests. `removeHookConfigCleansOrphanedGitignoreAfterManualPluginDelete` reproduces Green #2's scenario exactly (write → hand-delete the plugin → assert the `.gitignore` and both dirs are gone), and the `deletingEntryIsDurableOptOut` extension drives the full source round-trip — vanish → `.noSource` → return → `.skippedUserOwned` → entry still absent — which is precisely the sequence that used to silently re-add the mirror. Green #6's fix removes the last test that could have written a developer's real `~/.local/share/crow`.

---

### Summary Table

| # | Color | Finding | Location |
|---|-------|---------|----------|
| 1 | 🟢 | Un-mirror log kept the wording Green #4 fixed on the register path | `OpenCodeMCPConfigWriter.swift:130` |
| 2 | 🟢 | Record keyed by server name, not target path — a config-home move latches a permanent false opt-out | `OpenCodeMCPConfigWriter.swift:116,157-160` |
| 3 | 🟢 | No documented way to opt back in after the opt-out latches | `OpenCodeMCPConfigWriter.swift:29-36,158` |
| 4 | 🟢 | 4d record-mismatch branch (user edited Crow's entry) is untested | `OpenCodeMCPConfigWriterTests.swift:233-265` |
| 5 | 🟢 | `removeHookConfig` prunes `.opencode/` in worktrees Crow never wrote to (unreachable today) | `OpenCodeHookConfigWriter.swift:95,107-108` |

The Red is closed the way it was diagnosed: reverted, and verified against the Linux `swift test` link rather than a local macOS run — which is what the last two rounds asked for after two Linux-only breaks. Nothing regressed in the revert, and the three remaining Greens were each taken with a test that reproduces the failure scenario rather than just asserting the new code path. All five findings above are non-blocking; #2 is the one I'd pick up first, since it's the only one that can leave a user permanently without the tools this PR exists to give them. Good to merge.

---

PR #800, opened 2026-07-21
fix(web): split sidebar nav pills into two deterministic rows

dgershman:
Closes #799. Groups Reviews/Allowlist/Scorecard on row 1 and Manager + '+' on row 2 via explicit `.nav-pills-row` containers inside a column `.nav-pills` wrapper, removing width-dependent reflow. Preserves the Reviews unseen badge, the Manager pill's activity dot / remote-control glyph, and the new-manager menu; row 2 still holds '+' when no manager session exists.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #847, opened 2026-07-25
Drop redundant trailing "label" from crow:merge PR status chip

### LINKED ISSUE #846: Redundant trailing "label" word in a UI label
User report: *"the label has the word label in the end, it's redundant."*

Somewhere in the UI, a label's text ends with the word **"label"** (e.g. reads like "Foo Label"), which is redundant — the element is already a label, so the trailing word should be dropped.

## Location — needs pinning down
A repo grep did **not** find a user-facing string ending in "label":
- `web/app.js` / `web/*.html`: only `okLabel` / `cancelLabel` **variable names** and CSS classes (`score-card-label`, etc.) — not displayed text.
- Swift views: no `Text("… Label")`, `LabeledContent`, or `"X Label"` quoted display string.

So the first task is to **locate the offending label**, then drop the trailing "label" word. Likely candidates to check first:
- The **add-link / edit-link** form (the `--label` field caption or a heading rendering as "Link Label").
- A settings / inspector field caption.
- A dynamically-composed string (template literal like `` `${x} label` ``) that a literal grep for `"… label"` would miss.

## Acceptance
- The label in question no longer ends with the redundant word "label".
- No functional change — text only.

_Note: reporter can point at the exact screen to sharpen this if the grep-based hunt above doesn't immediately turn it up._

dgershman:
## Summary

The `crow:merge` PR-status marker rendered as **"🏷 crow:merge label"**. The trailing word **"label"** is redundant in the visible detail-header chip — the 🏷 glyph sits right beside the text — so it now reads **"🏷 crow:merge"**, matching the concise wording of its sibling chips ("Approved", "Conflicts", "Merged").

Closes #846.

## What changed (and why it's not purely "text-only")

The `.label` string has **two** consumers, so a naive one-line copy edit regressed accessibility — caught in review and fixed:

1. **Detail header** (`prStatusInline` → `prStatusPart`) — glyph is visible beside the text. Here `label: 'crow:merge'` is correct and unambiguous. ✅ the visible fix.
2. **Sidebar row pill** (`sessionRow`, `app.js:1219`) — glyph-only; the `.label` string becomes the pill's `aria-label`/`title`, which overrides the subtree, so the 🏷 glyph is **never announced** to a screen reader. Dropping "label" there would have made it announce a bare `crow:merge` token where every sibling is a full state phrase.

**Fix:** split the two channels.
```diff
-const PR_MERGE_LABEL_GLYPH = { glyph: '🏷', color: 'var(--gold)', label: 'crow:merge label' };
+const PR_MERGE_LABEL_GLYPH = { glyph: '🏷', color: 'var(--gold)', label: 'crow:merge', a11yLabel: 'crow:merge label' };
```
```diff
-const desc = [prLink.label || 'PR', ...parts.map((p) => p.label)].join(', ');
+const desc = [prLink.label || 'PR', ...parts.map((p) => p.a11yLabel || p.label)].join(', ');
```
`label` stays concise for the visible chip; `a11yLabel` keeps the noun for the aria path; `|| p.label` leaves every other glyph entry untouched. The optional `a11yLabel` key is documented in the shared glyph-vocabulary header.

## Result
- **Detail chip:** `🏷 crow:merge` (redundant word dropped — the ticket's ask)
- **Sidebar screen-reader announcement:** `…, crow:merge label` (unchanged; noun preserved)

## Tests

Added to `Packages/CrowDaemon/web-tests/row.test.js`:
- `sidebar aria keeps the unambiguous "crow:merge label" noun`
- `detail chip reads the concise "crow:merge"`
- `detail chip drops the redundant "crow:merge label"`

Also applied a `?.` to a **pre-existing** CROW-802 crash (the merged-PR assertion deref'd a now-null `.pr-ico`) so the suite runs to completion instead of aborting mid-file. All three new guards pass.

⚠️ **Known gap:** `row.test.js` still exits non-zero on `main` — 8 assertions fail on the same pre-existing CROW-802 selector drift (✔/✕/⚠ moved to SVG `.ico`), and `web-tests` isn't wired into CI. So these guards can't yet *gate* via exit code; that needs the drift repair + a CI job, both a separate ticket (see PR discussion). Behavior is verified end-to-end in the meantime.

## Commits
`f1dfa5b` initial copy change → `1b1a847` a11y split → `202c28b` first regression guard → `8196cc1` run-to-completion + full split coverage.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1266,6 +1266,10 @@ function prBadgeColor(pr) {
 // inherit `currentColor`, so the checkmarks tint with their label instead of
 // staying black (Apple emoji ✔/✕/⚠ ignore CSS `color` — CROW-802). Geometric
 // glyphs (◷/○/?) and the 🏷 tag are already color-faithful, so they stay text.
+// `label` is the chip text in the detail header AND the sidebar pill's
+// aria-label/title; an optional `a11yLabel` overrides the latter when the
+// concise chip text would be ambiguous without its (unannounced) glyph
+// (CROW-846). Add it only where the two channels must diverge.
 const PR_CHECKS_GLYPH = {
   passing: { glyph: '✔', icon: 'check', color: 'var(--green)', label: 'Checks pass' },
   failing: { glyph: '✕', icon: 'close', color: 'var(--red)', label: 'Checks failing' },
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1280,7 +1280,7 @@ const PR_REVIEW_GLYPH = {
 };
 const PR_MERGED_GLYPH = { glyph: '✔', icon: 'check', color: 'var(--purple)', label: 'Merged' };
 const PR_CONFLICT_GLYPH = { glyph: '⚠', icon: 'warning', color: 'var(--red)', label: 'Conflicts' };
-const PR_MERGE_LABEL_GLYPH = { glyph: '🏷', color: 'var(--gold)', label: 'crow:merge label' };
+const PR_MERGE_LABEL_GLYPH = { glyph: '🏷', color: 'var(--gold)', label: 'crow:merge' };
 
 function prChecksGlyph(pr) {
   const base = PR_CHECKS_GLYPH[pr.checks] || PR_CHECKS_GLYPH.unknown;
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1216,7 +1216,7 @@ function sessionRow(s) {
     }
     // Glyph + color must not be the only channel — mirrors native PRBadge's
     // `accessibilityDescription` ("#123, Checks pass, Approved").
-    const desc = [prLink.label || 'PR', ...parts.map((p) => p.label)].join(', ');
+    const desc = [prLink.label || 'PR', ...parts.map((p) => p.a11yLabel || p.label)].join(', ');
     prb.title = desc;
     prb.setAttribute('aria-label', desc);
     badges.appendChild(prb);
@@ -1280,7 +1280,10 @@ const PR_REVIEW_GLYPH = {
 };
 const PR_MERGED_GLYPH = { glyph: '✔', icon: 'check', color: 'var(--purple)', label: 'Merged' };
 const PR_CONFLICT_GLYPH = { glyph: '⚠', icon: 'warning', color: 'var(--red)', label: 'Conflicts' };
-const PR_MERGE_LABEL_GLYPH = { glyph: '🏷', color: 'var(--gold)', label: 'crow:merge' };
+// Chip text drops the redundant "label" (the 🏷 glyph shows it beside the
+// text); the aria path keeps it via `a11yLabel` — there the glyph is never
+// announced, so the noun is the only signal it's a label (CROW-846).
+const PR_MERGE_LABEL_GLYPH = { glyph: '🏷', color: 'var(--gold)', label: 'crow:merge', a11yLabel: 'crow:merge label' };
 
 function prChecksGlyph(pr) {
   const base = PR_CHECKS_GLYPH[pr.checks] || PR_CHECKS_GLYPH.unknown;
--- a/Packages/CrowDaemon/web-tests/row.test.js
+++ b/Packages/CrowDaemon/web-tests/row.test.js
@@ -56,6 +56,22 @@ function render(pr, overrides) {
 }
 const badge = (row) => row.querySelector('.pr-badge');
 
+// crow:merge a11y noun (CROW-846) — the sidebar pill is glyph-only, so its
+// aria-label/title is the sole screen-reader channel. The visible chip text
+// dropped the redundant trailing "label", but the aria path must keep the noun
+// via `PR_MERGE_LABEL_GLYPH.a11yLabel`; this pins that split so a later
+// `p.a11yLabel || p.label` → `p.label` "simplify" can't silently unwind it.
+// Placed ahead of the merged-PR section below, which trips a pre-existing
+// CROW-802 selector-drift crash (`.pr-ico` is null for the SVG ✔/✕/⚠ glyphs)
+// that would otherwise abort the run before this line executes.
+console.log('crow:merge announces the "label" noun to screen readers (CROW-846):');
+{
+  const rr = render({ has_pr: true, checks: 'passing', review: 'approved', merge: 'mergeable',
+    is_merged: false, has_blockers: false, ready_to_merge: true, failed_checks: [], has_merge_label: true });
+  check('sidebar aria-label keeps the "crow:merge label" noun',
+    /crow:merge label/.test(badge(rr.row).getAttribute('aria-label')));
+}
+
 console.log('Failing checks + changes requested:');
 let r = render({ has_pr: true, checks: 'failing', review: 'changesRequested', merge: 'MERGEABLE',
   is_merged: false, has_blockers: true, ready_to_merge: false, failed_checks: ['build', 'lint'] });
--- a/Packages/CrowDaemon/web-tests/row.test.js
+++ b/Packages/CrowDaemon/web-tests/row.test.js
@@ -9,6 +9,7 @@ const { JSDOM } = require('jsdom');
 const epilogue = `
 ;globalThis.__t = {
   sessionRow(s){ return sessionRow(s); },
+  prStatusInline(pr){ return prStatusInline(pr); },
   set live(v){ liveById = v; },
   set hideDetails(v){ uiConfig.hideSessionDetails = v; },
 };
@@ -56,22 +57,6 @@ function render(pr, overrides) {
 }
 const badge = (row) => row.querySelector('.pr-badge');
 
-// crow:merge a11y noun (CROW-846) — the sidebar pill is glyph-only, so its
-// aria-label/title is the sole screen-reader channel. The visible chip text
-// dropped the redundant trailing "label", but the aria path must keep the noun
-// via `PR_MERGE_LABEL_GLYPH.a11yLabel`; this pins that split so a later
-// `p.a11yLabel || p.label` → `p.label` "simplify" can't silently unwind it.
-// Placed ahead of the merged-PR section below, which trips a pre-existing
-// CROW-802 selector-drift crash (`.pr-ico` is null for the SVG ✔/✕/⚠ glyphs)
-// that would otherwise abort the run before this line executes.
-console.log('crow:merge announces the "label" noun to screen readers (CROW-846):');
-{
-  const rr = render({ has_pr: true, checks: 'passing', review: 'approved', merge: 'mergeable',
-    is_merged: false, has_blockers: false, ready_to_merge: true, failed_checks: [], has_merge_label: true });
-  check('sidebar aria-label keeps the "crow:merge label" noun',
-    /crow:merge label/.test(badge(rr.row).getAttribute('aria-label')));
-}
-
 console.log('Failing checks + changes requested:');
 let r = render({ has_pr: true, checks: 'failing', review: 'changesRequested', merge: 'MERGEABLE',
   is_merged: false, has_blockers: true, ready_to_merge: false, failed_checks: ['build', 'lint'] });
@@ -102,7 +87,13 @@ r = render({ has_pr: true, checks: 'passing', review: 'approved', merge: 'merged
   is_merged: true, has_blockers: false, ready_to_merge: false, failed_checks: [] });
 check('exactly one glyph', r.glyphs.length === 1);
 check('glyph is ✔', r.glyphs[0] === '✔');
-check('glyph purple', r.row.querySelector('.pr-ico').style.color === 'var(--purple)');
+// `?.` guards a pre-existing CROW-802 selector drift: ✔/✕/⚠ now render as SVG
+// `.ico` spans, not `.pr-ico` text, so `.pr-ico` is null here. Without the
+// guard this line *throws* and aborts the whole file mid-run; with it the
+// suite runs to completion so later sections (incl. the crow:merge checks
+// below) execute and the exit code stops masking their pass/fail. Fixing the
+// drift itself (so this assertion passes again) is CROW-802's own ticket.
+check('glyph purple', r.row.querySelector('.pr-ico')?.style.color === 'var(--purple)');
 
 console.log('\nConflicting PR adds a ⚠:');
 r = render({ has_pr: true, checks: 'passing', review: 'approved', merge: 'conflicting',
@@ -122,6 +113,18 @@ r = render({ has_pr: true, checks: 'passing', review: 'approved', merge: 'mergea
   is_merged: false, has_blockers: false, ready_to_merge: true, failed_checks: [], has_merge_label: true },
   { auto_merge: true });
 check('🏷 and ⛙ can both show', r.glyphs.includes('🏷') && !!r.row.querySelector('.automerge'));
+// CROW-846: the visible chip dropped the redundant trailing "label", but the
+// sidebar pill is glyph-only — its aria-label/title is the sole screen-reader
+// channel, so the noun must survive there via PR_MERGE_LABEL_GLYPH.a11yLabel.
+// Pin BOTH halves of that split so a later `p.a11yLabel || p.label` → `p.label`
+// (or reverting `label` to `'crow:merge label'`) can't silently unwind either.
+check('sidebar aria keeps the unambiguous "crow:merge label" noun',
+  /crow:merge label/.test(badge(r.row).getAttribute('aria-label')));
+const mergeChips = [...T.prStatusInline({ has_pr: true, checks: 'passing', review: 'approved',
+  merge: 'mergeable', is_merged: false, has_blockers: false, ready_to_merge: true,
+  failed_checks: [], has_merge_label: true }).querySelectorAll('.pr-stat-label')].map((n) => n.textContent);
+check('detail chip reads the concise "crow:merge"', mergeChips.includes('crow:merge'));
+check('detail chip drops the redundant "crow:merge label"', !mergeChips.includes('crow:merge label'));
 
 console.log('\nGraceful degradation (older daemon payload / no PR status):');
 r = render(undefined);


dhilgaertner (changes requested):
## Code & Security Review

Single-line copy change to `PR_MERGE_LABEL_GLYPH.label` in `app.js:1283`. The wording goal is sound and the sibling-chip consistency argument holds for the detail header. One issue on a second consumer the description overlooks.

### Critical Issues

None.

### Security Review

**Strengths:**
- No security surface. The changed value is a hardcoded literal, not user- or API-derived.
- Both render paths go through `el()` (`app.js:677-682`), which assigns `textContent`, never `innerHTML` — no XSS vector even if the string were dynamic.
- No auth, input-validation, injection, secrets, or crypto involvement.

**Concerns:**
- None.

### Code Quality

**🟡 Yellow — `.label` has two consumers, not one; the second is the sidebar pill's accessibility text**

The PR description states the string "is rendered only as the visible chip text in `prStatusInline` → `prStatusPart`". It is also consumed at `app.js:1219`:

```js
const desc = [prLink.label || 'PR', ...parts.map((p) => p.label)].join(', ');
prb.title = desc;
prb.setAttribute('aria-label', desc);
```

That is the **sidebar row pill**, which — as the PR correctly notes — is glyph-only. Because `aria-label` on the pill overrides its inner glyph content for assistive tech, the `.label` string is the *sole* channel available to a screen reader there. The 🏷 glyph is never announced.

Result: the sidebar pill now announces `"#847, Checks pass, Approved, crow:merge"`. Every other element in that list is a state phrase (`Checks pass`, `Approved`, `Conflicts`, `Merged`); `crow:merge` alone is a bare token with nothing indicating it denotes the presence of a label. Previously it read `"…, crow:merge label"` — unambiguous.

This is exactly the case the adjacent comment at `app.js:1217-1218` guards against:

```js
// Glyph + color must not be the only channel — mirrors native PRBadge's
// `accessibilityDescription` ("#123, Checks pass, Approved").
```

The PR's justification ("the 🏷 glyph already signals it's a label") is valid for `prStatusInline`, where the glyph is visible alongside the text — but it does not transfer to the aria path, where no glyph reaches the user.

Suggested fix — keep the concise chip text, preserve the unambiguous announcement:

```diff
-const PR_MERGE_LABEL_GLYPH = { glyph: '🏷', color: 'var(--gold)', label: 'crow:merge' };
+const PR_MERGE_LABEL_GLYPH = { glyph: '🏷', color: 'var(--gold)', label: 'crow:merge', a11yLabel: 'crow:merge label' };
```
```diff
-  const desc = [prLink.label || 'PR', ...parts.map((p) => p.label)].join(', ');
+  const desc = [prLink.label || 'PR', ...parts.map((p) => p.a11yLabel || p.label)].join(', ');
```

The `|| p.label` fallback leaves all other glyph entries untouched.

**🟢 Green — test note, pre-existing, not caused by this PR**

The PR notes the jsdom suite couldn't be run locally. I installed `jsdom` and ran it: `npm test` in `Packages/CrowDaemon/web-tests` **crashes** in `row.test.js:89` with `TypeError: Cannot read properties of null (reading 'style')` — `r.row.querySelector('.pr-ico')` is `null` for the merged-PR case, because CROW-802 moved ✔/✕/⚠ from text glyphs to inline SVGs and only geometric glyphs (◷/○/?) still carry `.pr-ico`.

I confirmed this reproduces identically on `origin/main`, so **it is not introduced here.** Worth flagging separately, though, because the crash happens *before* line 97 — the 🏷 `crow:merge` assertions at `row.test.js:97-108` never execute. So "the tests are unaffected" is accurate only in the sense that they were already not running. `board.test.js`, `touch-scroll.test.js`, and `wheel-scroll.test.js` all pass (50/50).

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 1 Green] findings.

The Yellow is a small, self-contained fix; the copy improvement itself is worth landing once the aria path keeps its noun.

---

dgershman:
Good catch — you're right that `.label` has a second consumer at `app.js:1219` (the sidebar pill's `aria-label`/`title`), and since that pill is glyph-only, the string is the sole channel a screen reader gets. Applied your suggested fix in 1b1a8471:

- `PR_MERGE_LABEL_GLYPH.label` stays concise `'crow:merge'` for the visible detail-header chip (glyph sits beside the text).
- New `a11yLabel: 'crow:merge label'` feeds the aria path via `...parts.map((p) => p.a11yLabel || p.label)` — the `|| p.label` fallback leaves every other glyph entry untouched.

Verified both paths in jsdom (reusing `row.test.js`'s loader shape):
- Sidebar pill aria-label → `"#847, Checks pass, Approved, crow:merge label"`
- Detail-header chip → `"crow:merge"`

On the Green note: confirmed the `row.test.js:89` crash reproduces on `origin/main` (pre-existing, from CROW-802's SVG glyphs) and aborts before the 🏷 assertions at line 97 — so I verified this change with the standalone harness above rather than relying on that file. `board`/`touch-scroll`/`wheel-scroll` pass 50/50. The row.test.js crash is worth its own ticket but is out of scope here.

dhilgaertner (changes requested):
## Code & Security Review

Two-line copy change plus an a11y split. I rendered **both** consumer paths in jsdom against `origin/main` and this branch to confirm the claim in the PR body:

```
main    sidebar aria-label : "#800, Checks pass, Approved, crow:merge label"
        detail chips       : ["Checks pass","Approved","🏷crow:merge label"]

PR      sidebar aria-label : "#800, Checks pass, Approved, crow:merge label"   ← unchanged
        detail chips       : ["Checks pass","Approved","🏷crow:merge"]         ← the fix
```

The `p.a11yLabel || p.label` fallback (`app.js:1219`) leaves every other glyph entry byte-identical — I checked the dynamic `N failing` path too (`"#800, 2 failing, Changes requested, crow:merge label"`), and the no-label case still renders `"#800, Checks pass, Approved"`. `PR_MERGE_LABEL_GLYPH` has exactly two consumers (`prBadgeParts` → aria at `:1219`, `prStatusInline` → `prStatusPart` at `:1831`), so the split covers the full surface. The follow-up commit correctly caught that the sidebar pill is glyph-only and the string is its sole screen-reader channel.

### Critical Issues

None.

### Security Review

**Strengths:**
- Both changed values are module-scope string constants — no user/daemon-controlled input reaches them.
- Rendered through `el()` (`app.js:677`), which assigns `textContent`, so the chip text and `aria-label`/`title` can't inject markup. The only `innerHTML` on this path is `icon()` (`app.js:1093`), and it interpolates a hardcoded `ICONS[name]` entry, never a label.
- No auth, network, storage, or serialization surface touched.

**Concerns:** None.

### Code Quality

🟡 **The new `label` / `a11yLabel` split has no persisted coverage** — `app.js:1219`, `app.js:1286`

This exact string already regressed once inside this PR (commit 1 broke the sidebar announcement, commit 2 restored it), and the commit message's "verified in jsdom" was ad-hoc — nothing in the tree pins it. A future "simplify" that collapses `p.a11yLabel || p.label` back to `p.label`, or drops the key from the constant, silently reintroduces the bare `"crow:merge"` announcement. `row.test.js` already has both the assertion idiom (`:66`) and a `crow:merge` section (`:97`) with `r` and the `badge()` helper in scope, so it's a one-liner there:

```js
check('aria keeps the "label" noun', /crow:merge label/.test(badge(r.row).getAttribute('aria-label')));
```

🟢 **`row.test.js` is red on `main` — pre-existing, not caused by this PR** (worth a follow-up ticket)

The PR body notes the suite couldn't be run locally. I ran it, and it fails identically on `origin/main` and on this branch: it throws at `row.test.js:89` (`Cannot read properties of null`). The CROW-802 migration made ✔/✕/⚠ render as inline SVG (`.ico`) instead of `.pr-ico` text spans, but the test still collects `.pr-badge .pr-ico` textContent — 5 assertions fail and the run dies before it reaches the 🏷 section at `:97`. Nothing in `.github/workflows` invokes `web-tests`, so CI never surfaced the drift. Repairing that is out of scope here, but note it means the assertion suggested above won't actually execute until the selector drift is fixed.

🟢 **Consider documenting `a11yLabel` in the vocabulary header** — `app.js:1260-1268`

The block comment defines the shared glyph-entry contract for both render paths; `a11yLabel` is now part of that contract but is explained only at the single entry that uses it. One line in the header would tell the next person adding a glyph that the optional key exists.

### Summary Table

The behavior in this diff is correct and I verified it end-to-end; the only ask is the one-line regression assertion so the a11y split can't silently unwind.

---

dgershman:
Thanks for rendering both paths against main — addressed both findings in 202c28bc.

**🟡 Persisted coverage.** Added the regression assertion to `row.test.js`:

```js
check('sidebar aria-label keeps the "crow:merge label" noun',
  /crow:merge label/.test(badge(rr.row).getAttribute('aria-label')));
```

One wrinkle: your suggested spot (the `crow:merge` section at `:97`) is unreachable — the file crashes at `:89` first (the pre-existing CROW-802 `.pr-ico`-is-null drift you flagged). An assertion there would be dead. So I placed it in its own block **ahead** of that crash, where it executes and passes today:

```
crow:merge announces the "label" noun to screen readers (CROW-846):
  ✓ sidebar aria-label keeps the "crow:merge label" noun
```

Block-scoped `const rr` so it doesn't touch the existing `let r`. A comment at the section explains the early placement. This makes the guard live now rather than parked behind the crash.

**🟢 `a11yLabel` in the vocabulary header.** Documented the optional key in the shared glyph contract block (`app.js:1269-1272`) so the next person adding a glyph knows it exists and when to use it.

**🟢 `row.test.js` red on main.** Left the selector-drift repair out of scope as you noted — it's a CROW-802 test-maintenance issue independent of this copy change, and folding a full glyph-collection rewrite into a one-line copy PR would balloon the review. Happy to file a follow-up ticket for it (the fix is updating `render()` to also collect the SVG `.ico` glyphs and refreshing the ✔/✕/⚠/merged assertions). My new guard runs regardless since it sits before the crash; once the drift is fixed the rest of the suite comes back too. Let me know if you'd rather I open that ticket now.

dhilgaertner (changes requested):
## Code & Security Review

The core change is **correct and well-reasoned**. I verified the a11y argument holds and ran the test suite. My findings are about the safety net around the change, not the change itself.

### Critical Issues

None. No Red findings.

### Security Review

**Strengths:**
- No security surface. The diff moves two hardcoded presentational string constants; no user-controlled data enters either path.
- Chip text and aria strings render through `el()` (`app.js:677-682`), which assigns `textContent`, not `innerHTML` — no XSS vector introduced. The nearby `innerHTML` uses (`app.js:1032`, `1093`) are static SVG literals, untouched here.
- No authn/authz, injection, crypto, secrets, or config surface touched.

**Concerns:** None.

### Code Quality

**What's right (verified, not assumed):**
- The a11y reasoning is sound. `sessionRow` sets `aria-label` on the pill (`app.js:1221`), which overrides its subtree — so the 🏷 glyph genuinely is never announced there, and the noun must survive in the aria string. Dropping it from `label` alone would have regressed the sidebar. The `a11yLabel` split at `app.js:1219`/`1290` is the right fix.
- The `p.a11yLabel || p.label` fallback is minimal and backward-compatible; every other glyph entry is untouched.
- `prChecksGlyph`'s `{ ...base, label: … }` spread (`app.js:1296`) preserves a future `a11yLabel` rather than dropping it. No latent bug.
- The detail-header path is fine: `prStatusPart` (`app.js:1835`) renders 🏷 as a text emoji beside "crow:merge", and screen readers announce the emoji — the noun is still reachable there.
- I ran the suite: the new test **executes and passes.**

---

**🟡 1 — The regression test cannot gate the regression it claims to pin.** `row.test.js:59-73`

The commit message says this pins the split so a later `p.a11yLabel || p.label` → `p.label` "simplify" can't silently unwind it. It can't currently do that, for two compounding reasons:

1. **`row.test.js` is not run by CI.** `.github/workflows/ci.yml` has exactly three jobs — *Build & Test* (Swift/Linux), *Desktop (Tauri)*, and *Shell Lint*. Nothing runs `web-tests`, `npm test`, or any Node step.
2. **The suite already exits non-zero.** It crashes at `row.test.js:105` with `TypeError: Cannot read properties of null (reading 'style')`. So if someone unwinds the change, the new check prints `✗` but the process still exits 1 — byte-identical to today's exit status. Nothing automated can tell the two apart.

I confirmed the crash is **pre-existing** — I ran `row.test.js` from `origin/main` in isolation and it fails at the same assertion. This PR did not cause it, and I'm not asking you to fix CROW-802's selector drift.

But the cheap fix is in the file you're already editing: make the merged-PR assertion null-safe (e.g. `r.row.querySelector('.pr-ico')?.style.color === 'var(--purple)'`) so the suite runs to completion and its exit code becomes meaningful again. Wiring `web-tests` into CI is a genuinely separate ticket — worth filing, since four jsdom suites currently run only when someone remembers to.

**🟡 2 — The PR description no longer matches the shipped diff.**

The body describes a single-line change, states **"Text-only, no functional change"**, and says the web tests are *"unaffected"* and could not be executed locally. The diff that's actually up for merge introduces a new `a11yLabel` fallback in the aria path (`app.js:1219`) and adds a test that evidently does run. That's a behavior change to the sidebar's accessible name — small, correct, but not "text-only."

Please refresh the description before merge. The commit series (`f1dfa5b` → `1b1a847` → `202c28b`) tells the real story well; the PR body is still stuck at the first commit, and it's the durable record.

**🟢 3 — The visible chip text — the actual subject of CROW-846 — is untested.**

The new test pins only the aria path. Reverting `label: 'crow:merge'` back to `'crow:merge label'` (precisely what the ticket asked you to remove) is caught by nothing. Worth an assertion on the `.pr-stat-label` text, though I'll note it isn't free: the test epilogue (`row.test.js:9-15`) exposes only `sessionRow`, so pinning the detail chip means exporting `prStatusInline` too. Reasonable to defer.

**🟢 4 — Test execution order is now load-bearing.**

Placing the block at the top to run ahead of the crash works, but it makes file position significant, documented only in a comment. Anyone appending a test after line 105 gets one that silently never executes. Resolving finding 1 dissolves this one automatically.

### Summary Table

To be clear: the shipped behavior is correct and I'd approve it on the code alone. The blockers are that the stated safety net doesn't hold (🟡1) and the description misdescribes the change (🟡2). Both are small, and 🟡1 is a one-character `?.` in a file you're already touching.

---

dgershman:
Thanks — the code-confirmed approval and the precise findings are appreciated. Addressed all four in 8196cc19.

**🟡1 — regression test can't gate.** Applied your one-char `?.` at the merged-PR assertion. It removes the hard *throw*, so the file now runs to completion (verified) — which also dissolves 🟡4.

One honest correction, though: `?.` stops the crash but **not the red**. With the file running to the end I can see the full tally — `29 passed, 8 failed`, exit 1:

```
✗ two glyphs rendered / both are ✕      (failing+changesRequested → SVG .ico)
✗ two ✔ glyphs                          (passing+approved → SVG .ico)
✗ exactly one glyph / glyph is ✔ / glyph purple   (merged → SVG .ico)
✗ three glyphs / third glyph is ⚠       (conflict → SVG .ico)
```

All 8 are the same CROW-802 drift (`render()` collects `.pr-badge .pr-ico` textContent; ✔/✕/⚠ are now SVG `.ico` spans with no text). They fail identically on `origin/main`. So the exit code is still byte-identical whether or not my split is unwound — the crux of your finding survives the `?.`.

Making it genuinely gate needs (a) the drift repair — `icon()` would need to emit a glyph identifier (e.g. `data-icon`) so the test can map an SVG back to ✔/✕/⚠, then fix those 8 assertions — and (b) wiring `web-tests` into CI. Both are the separate ticket you flagged; I didn't fold the `icon()` change + assertion redesign into this copy PR since it's squarely CROW-802's scope and reinventing its post-drift assertions here would be guesswork. I tried to file the tracking issue but the write was blocked by a guard (I wasn't asked to open tickets) — happy to file it if you give the word, or I can leave it to you.

**🟢3 — visible chip untested (the actual ticket subject).** Good call, did it: exported `prStatusInline` in the epilogue and added two checks — `detail chip reads the concise "crow:merge"` and `detail chip drops the redundant "crow:merge label"`. So reverting `label` back to `'crow:merge label'` now trips a check, alongside the aria-noun guard. All three pass.

**🟢4 — load-bearing order.** Gone — with the crash removed I moved the crow:merge checks out of the top-of-file block into their natural section; nothing depends on file position anymore.

**🟡2 — stale description.** Rewrote the PR body to match the shipped diff: the `label`/`a11yLabel` split, the accessible-name change (no longer "text-only"), the new tests, and an explicit ⚠️ Known-gap note about the exit-code/CI limitation above so the record is honest.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues

None.

### Verification performed

I re-ran the claims rather than taking the description at face value:

| Check | Result |
|---|---|
| `node row.test.js` on this branch | `29 passed, 8 failed`, exit 1 — runs to completion, all 3 new guards ✓ |
| `node row.test.js` on `origin/main` (main's test + main's `app.js`, isolated tree) | **throws** `TypeError: Cannot read properties of null (reading 'style')` at line 89, aborts mid-file |
| `board.test.js` / `touch-scroll.test.js` / `wheel-scroll.test.js` | 50 / 20 / 50 passed, 0 failed — no collateral from the `app.js` edit |
| `node --check app.js` | clean |
| `.github/workflows/ci.yml` | Swift build+test, Tauri build, shellcheck — **no** `web-tests` job |

So both load-bearing claims hold: the 8 failures are pre-existing CROW-802 selector drift (`✔`/`✕`/`⚠` now render as SVG `.ico`, not `.pr-ico`), and `web-tests` genuinely isn't wired into CI. The `?.` is a real improvement — it converts a hard abort into a *recorded failure* (`undefined === 'var(--purple)'` → false → `✗`), so nothing is silently masked and 3 previously-unreachable sections now execute.

### Security Review

**Strengths:**
- No attack surface touched. Both new strings are hardcoded module constants (`app.js:1290`), not user- or remote-derived.
- The sink is safe: `el()` (`app.js:677-682`) assigns via `textContent`, and the aria path uses `setAttribute` — no `innerHTML` anywhere in the changed lines. No XSS vector.
- No auth, network, filesystem, crypto, or config surface in the diff.

**Concerns:** None.

### Code Quality

The a11y split is the right call, and the reasoning in the PR body checks out against the code: `prb.setAttribute('aria-label', desc)` (`app.js:1219-1221`) is set on the element *containing* the glyph children, so the accessible name overrides the subtree and the 🏷 is never announced in the sidebar. Dropping the noun there would have been a real regression. `p.a11yLabel || p.label` leaves all five other glyph entries untouched, and the optional key is documented where a maintainer will actually see it (`app.js:1269-1272`).

Three non-blocking notes:

1. 🟢 **`prChecksGlyph` spread is a latent footgun** — `app.js:1296` returns `{ ...base, label: pr.failed_checks.length + ' failing' }`. That overrides `label` but *preserves* `a11yLabel`. No bug today (no `PR_CHECKS_GLYPH` entry has one), but if `failing` ever gains an `a11yLabel`, the new `|| p.label` precedence means the sidebar would announce the stale generic phrase and silently drop the failing count. Cheap insurance: mention it in the vocabulary header, or spread `a11yLabel: undefined` alongside the `label` override.

2. 🟢 **The two *visible* surfaces now differ.** `desc` feeds both `prb.title` and `aria-label` (`app.js:1220-1221`), so the sidebar hover tooltip reads `…, crow:merge label` while the detail chip reads `crow:merge`. This is defensible — the tooltip is the glyph-only pill's sole text channel, so it's the same argument as the aria path — but it's worth being deliberate about, since the tooltip is seen, not just announced.

3. 🟢 **Pre-existing, out of scope:** `aria-label`/`title` on a role-less `<span class="pr-badge">` is technically a prohibited naming pattern (`role=generic` disallows an accessible name); AT support is real but not guaranteed by spec. A `role="img"` on the pill would make this split reliably reach a screen reader. Untouched by this PR — flagging only because it's the channel the whole `a11yLabel` split exists to serve.

**On the disclosed CI gap:** accepted as-is. Fixing the CROW-802 drift and wiring `web-tests` into CI are both genuinely separate tickets, and the PR states the limitation plainly rather than papering over it. Until then these guards are output-grade, not exit-code-grade — worth filing the follow-up so that distinction doesn't get forgotten.

### Summary Table

Tight, well-scoped diff that fixes the ticket's ask without regressing the accessibility channel it would have quietly broken. Tests pin both halves of the split, and the honesty about the CI gap is the right instinct.

---

PR #849, opened 2026-07-25
Rename make crowd-dev → daemon-run; serve frozen web assets (#848)

### LINKED ISSUE #848: Rename `make crowd-dev` → `make daemon-run` and drop `--web-dir` (serve frozen assets; no live-from-source side effects)
Two changes to the local dev daemon flow, motivated by wanting to avoid the weird side effects of serving the web UI live from the source folder (stale/mismatched assets vs. the compiled bundle, path/caching quirks).

## 1. Rename the Make target `crowd-dev` → `daemon-run`
`Makefile`:
- `.PHONY` line (currently lists `crowd-dev`) → `daemon-run`.
- The `crowd-dev:` target (~`Makefile:73`) → `daemon-run:`.
- Help text (~`Makefile:52`): `crowd-dev  Run crowd with the web UI served live from source…` → update the name and the description (see #2 — it no longer serves live).
- Optional but tidy: rename `scripts/crowd-dev.sh` → `scripts/daemon-run.sh` and update the `bash scripts/…` line in the target. If the rename is skipped, at least keep the invocation working.

## 2. Stop passing `--web-dir` — serve the frozen embedded web assets
`scripts/crowd-dev.sh` currently builds this start command:
```sh
WEBDIR="$(pwd)/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web"
START_CMD=(.build/debug/crowd --host "$HOST" --http-port "$PORT" --socket "$SOCK" --web-dir "$WEBDIR")
```
`--web-dir` is what makes `crowd` read UI files live from disk instead of the compiled bundle (`StaticAssets.webResponse` / `CrowDaemon.swift` "serving web UI live from …"). Drop `WEBDIR` and the `--web-dir "$WEBDIR"` argument so `daemon-run` serves the **frozen, bundle-baked** web assets — same asset source as `make run`’s sidecar. Update the script header comment and the `[crowd-dev] … web live from …` echo lines accordingly (no longer live).
- The `--watch` flag (rebuild + restart `crowd` on Swift changes) can stay as-is — it is orthogonal to web-dir.

## 3. Tauri desktop app — keep it off `--web-dir`
The desktop sidecar already spawns `crowd` **without** `--web-dir` (frozen assets) per `crow-desktop/src-tauri/src/lib.rs` ("No --web-dir: the sidecar serves the frozen web assets baked …"). Action here is: verify no dev path passes `--web-dir`/`CROW_WEB_DIR` to the sidecar, and update the stale comment in `lib.rs` that points devs to `make crowd-dev` (→ the new name). Confirms "same for the tauri app — no live-web side effects."

## Heads-up / intended tradeoff
This **removes the live-from-source web-editing loop** (edit `index.html`/`app.css`/`app.js` → browser refresh, no rebuild). That is the deliberate intent of this ticket, not a regression — after this change, web UI edits require a rebuild to be picked up, exactly like `make run`. Do not "fix" it back by re-adding `--web-dir`. If a live-edit loop is ever wanted again it should be a separate, explicitly-named target.

## Acceptance
- `make daemon-run` exists; `make crowd-dev` no longer does (or aliases cleanly).
- `daemon-run` launches `crowd` with **no** `--web-dir` — served UI comes from the compiled bundle.
- No `--web-dir`/`CROW_WEB_DIR` is passed anywhere in the Tauri desktop sidecar path; the `lib.rs` comment references the new target name.
- `grep -rn "crowd-dev"` across the repo returns only intentional history (no live references in Makefile help, scripts, or tauri comments).

dgershman:
Closes #848

Renames the local dev daemon target and stops serving the web UI live from source, so `crowd`'s dev loop uses the same frozen, bundle-baked assets as `make run` — no more stale/mismatched-asset side effects.

## Changes
1. **`Makefile`** — `crowd-dev` → `daemon-run` (`.PHONY`, target, help text, comment).
2. **`scripts/crowd-dev.sh` → `scripts/daemon-run.sh`** — drops `WEBDIR` and the `--web-dir` argument so `crowd` serves the compiled bundle's web assets. Header/help/log lines updated; `[crowd-dev]` prefixes → `[daemon-run]`. `--watch` (rebuild + restart on Swift change) is unchanged.
3. **`crow-desktop/src-tauri/src/lib.rs`** — the sidecar already spawns `crowd` without `--web-dir`; refreshed the comment to point at `make daemon-run` and dropped the stale "(which passes --web-dir)" claim. Verified no dev path passes `--web-dir`/`CROW_WEB_DIR` to the sidecar.
4. **Docs / READMEs** — updated the dev-loop docs (README, getting-started, troubleshooting, architecture, crow-desktop/README) to describe frozen assets + the new target name, and corrected the stale "live from source" description of the `make run` sidecar.
5. **Swift comments** — retargeted the "multi-`crowd-dev` footgun" references in `CrowDaemon.swift` and `DaemonSecurityTests.swift`.

## Intended tradeoff
This **removes the live-from-source web-editing loop** by design (per the ticket). Web UI edits now require a rebuild, exactly like `make run`.

## Acceptance
- ✅ `make daemon-run` exists; `make crowd-dev` no longer does.
- ✅ `daemon-run` launches `crowd` with **no** `--web-dir` (served UI comes from the compiled bundle).
- ✅ No `--web-dir`/`CROW_WEB_DIR` passed anywhere in the Tauri sidecar path; the `lib.rs` comment references the new target.
- ✅ `grep -rn "crowd-dev"` across the repo returns nothing (no live references remain).

## Test plan
- `bash -n scripts/daemon-run.sh` → OK
- `bash scripts/daemon-run.sh --help` → shows updated usage
- `make -n daemon-run` → `bash scripts/daemon-run.sh`
- `grep -rn "crowd-dev" .` → no matches

🐦‍⬛ Generated with Crow


dhilgaertner (changes requested):
## Code & Security Review

Clean, well-scoped rename. I verified every acceptance claim in the description and they hold: `grep -rn "crowd-dev"` across the repo (including hidden dirs) returns **zero** matches, the rename preserved mode `100755`, `make -n daemon-run` → `bash scripts/daemon-run.sh`, `bash -n` passes, `--help` renders, `make help` stays column-aligned, no CI workflow referenced the old name, and `crow-desktop/src-tauri/src/lib.rs:224-238` genuinely never passes `--web-dir`. Two documentation/behavior gaps below keep this from an approve.

### Security Review

**Strengths:**
- Dropping `--web-dir` from the dev script **narrows** attack surface — the daemon no longer serves an arbitrary on-disk directory in the default dev loop.
- No change to the auth, CSP, or binding posture: `StaticAssets.contentSecurityPolicy` and the loopback-only `HOST="127.0.0.1"` pin (`scripts/daemon-run.sh:37`) are untouched.
- Path-traversal guards are intact — `/xterm/:file` still goes through `StaticAssets.isSafeAssetName`, and the web routes are a fixed allowlist, so the `webDir` join at `StaticAssets.swift:68` can't be steered by a request.
- No secrets, no new inputs, no new network surface. The Swift and Rust hunks are comment-only.

**Concerns:** none.

### Code Quality

**🟡 1. `docs/getting-started.md:112` misdescribes what `make daemon-run` does**

> (For a dev loop that rebuilds + restarts `crowd` on Swift changes, use `make daemon-run` instead.)

`make daemon-run` runs `bash scripts/daemon-run.sh` with **no** `--watch` (`Makefile:74-75`), which builds once and `exec`s a stable daemon (`scripts/daemon-run.sh:58-61`) — it never rebuilds or restarts on change. Only `bash scripts/daemon-run.sh --watch` does. A developer following this line edits a Swift file, waits, and nothing happens.

The table at `docs/getting-started.md:32` gets this right by qualifying with `--watch`; line 112 should match, e.g. "…use `make daemon-run` (or `bash scripts/daemon-run.sh --watch` to rebuild + restart on Swift changes)".

**🟡 2. `--watch` still watches only `*.swift`, so after this PR *nothing* picks up web-UI edits**

Removing the live-from-source loop is the intended tradeoff — but the watcher wasn't extended to compensate, so the frozen-asset dev loop has no incremental path for web assets at all. Both watch backends filter to Swift:

- `scripts/daemon-run.sh:72` — `exec watchexec -r -e swift …`
- `scripts/daemon-run.sh:83,85` — `find "${WATCH_PATHS[@]}" -name '*.swift' …`

`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/*` is inside `WATCH_PATHS` but excluded by the extension filter.

**Failure scenario:** developer runs `bash scripts/daemon-run.sh --watch`, edits `app.js`, saves, hits ⌘R → old UI. No rebuild, no `[daemon-run] (re)started` line, no error. A silent no-op.

This compounds `README.md:172`, which reads "Rebuilding the daemon (`daemon-run --watch`, or a manual `make daemon` + restart) restarts it" two lines after `README.md:160` says web edits "need a `make daemon` rebuild + restart" — a reader reasonably concludes `--watch` covers web edits. It doesn't.

Worth fixing in the script rather than the docs: `Package.swift:61` uses `.copy("Resources/web")` and `StaticAssets.webResponse` reads `Bundle.module` **per request**, so a rebuild re-copies the assets and a browser refresh shows them. Two lines restore a rebuild-on-save loop that's fully consistent with the frozen-asset model:

```bash
exec watchexec -r -e swift,js,css,html,svg,json "${WATCH_ARGS[@]}" -- \
# and in poll mode:
find "${WATCH_PATHS[@]}" \( -name '*.swift' -o -path '*/Resources/web/*' \) -type f …
```

If Swift-only is deliberate, then at minimum `README.md:172` and the `--help` text should say plainly that `--watch` does **not** cover web assets.

**🟢 3. `CROW_WEB_DIR` silently defeats the frozen-asset guarantee — and is now undiscoverable**

`CrowDaemon.swift:1181-1182` still honors `CROW_WEB_DIR`, and both `daemon-run.sh` and the Tauri sidecar (`Command::new` inherits the parent env) pass it through — so an exported `CROW_WEB_DIR` restores live-from-source in both paths, contradicting the `lib.rs:235-238` comment and acceptance item 2. That's a fine escape hatch, but this PR removed the last usage example of `--web-dir`, there's no `crowd --help` text for it, and no doc mentions it — a source comment is now the only trace. Consider one line in the env-overrides list at `scripts/daemon-run.sh:18-20`: `CROW_WEB_DIR (unset — set it to serve the web UI live from source)`.

**🟢 4. `README.md:160`/`172` "rebuild **+ restart**" is likely stricter than reality**

Since bundle resources are read per request (see #2), a `swift build` alone should refresh the served assets without restarting `crowd`. Not wrong, just pessimistic — worth confirming and softening to just "rebuild" if it holds.

### Testing Notes

- `bash -n scripts/daemon-run.sh` → clean; `--help` output correct.
- `make -n daemon-run`, `make help` → correct and aligned.
- I did **not** run `swift build` / `swift test`: the Swift hunks in `CrowDaemon.swift` and `DaemonSecurityTests.swift` and the Rust hunk in `lib.rs` are comment-only, verified line-by-line in the diff.
- `shellcheck` isn't installed locally, so I couldn't pre-run CI's `shellcheck scripts/*.sh` (`.github/workflows/ci.yml:85`). The script's executable logic changed only by deleting the `WEBDIR` assignment and one array element, so no new class of finding is expected.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 2 Green] findings. Both Yellows are small: one doc sentence at `docs/getting-started.md:112`, and either two lines in `scripts/daemon-run.sh` or one clarifying sentence in `README.md:172`. No security concerns.

dgershman:
Thanks for the thorough review — all four findings addressed in `3e885125`.

**🟡 1 — `docs/getting-started.md:112`** Reworded so `--watch` (not plain `make daemon-run`) is what rebuilds + restarts; the table entry at :32 now says "Swift or web-asset changes" too.

**🟡 2 — `--watch` ignored web assets** Fixed in the script (your preferred fix), not the docs. Both backends now watch web assets:
- watchexec: `-e swift,js,css,html,svg,json`
- poll `find`: `\( -name '*.swift' -o -path '*/Resources/web/*' \)`

A rebuild re-copies `Resources/web` (`Package.swift .copy`) and `StaticAssets.webResponse` reads `Bundle.module` per request, so save → rebuild → ⌘R shows web edits. Verified the `find` filter matches `app.js`/`app.css`/`index.html` (+5 more) alongside Swift.

**🟢 3 — `CROW_WEB_DIR` undiscoverable** Added it to the env-overrides header in `daemon-run.sh` (noting the crowd child inherits it) — the one remaining trace beyond the source comment.

**🟢 4 — "rebuild + restart" pessimistic** Confirmed via `StaticAssets.swift:75` (per-request `Bundle.module` read) + `.copy("Resources/web")`. Softened `README.md:160/172`: web-only edits need just a rebuild (served on the next refresh, no restart); Swift changes still restart.

`shellcheck scripts/daemon-run.sh` clean; `bash -n`, `--help`, `make -n daemon-run` all verified.

dhilgaertner (changes requested):
## Code & Security Review (round 2 — `3e885125`)

All four round-1 findings are genuinely addressed, and I **empirically confirmed** the load-bearing claim behind the `--watch` fix rather than taking the code-reading on faith:

> Built a minimal SwiftPM package with `.copy("Resources/web")`, changed only a file's *contents* inside the copied directory, and rebuilt → `[0/3] Copying web`, bundle content updated. So `swift build --product crowd` does re-copy `Resources/web` on a web-only edit, and `StaticAssets.webResponse` (`StaticAssets.swift:75-79`) re-reads `Bundle.module` per request. The "save → rebuild → ⌘R" loop this PR promises really works.

Re-verified mechanically: `bash -n` clean, `--help` correct, `make -n daemon-run` → `bash scripts/daemon-run.sh`, `make help` still column-aligned (`daemon-run` is exactly 10 chars — the padding lands), rename kept mode `100755`, `rg "crowd-dev"` across the repo (hidden dirs included) → **zero** matches, and the poll-mode `find` filter matches all 8 web assets alongside the 269 Swift files.

Three small things came in with the fix. None are Red; all are one- or two-line changes.

### Security Review

**Strengths:**
- Unchanged from round 1 and still good: dropping `--web-dir` from the default dev loop *narrows* surface (the daemon no longer serves an arbitrary on-disk dir by default); the loopback pin (`scripts/daemon-run.sh:39`), CSP, and `/xterm/:file` traversal guard (`StaticAssets.isSafeAssetName`) are untouched.
- Documenting `CROW_WEB_DIR` (`scripts/daemon-run.sh:21-22`) is accurate — I confirmed `CrowDaemon.swift:1181-1182` honors it and that the crowd child inherits it. Good: the escape hatch is now discoverable instead of source-comment-only.
- No new inputs, secrets, or network surface. Swift/Rust hunks remain comment-only.

**Concerns:** none.

### Code Quality

**🟡 1. Poll-mode `--watch` can now leave you with *no* daemon — `scripts/daemon-run.sh:99-103`**

```bash
if [ -n "$PID" ]; then kill "$PID" 2>/dev/null || true; fi
if swift build --product crowd; then
  "${START_CMD[@]}" &
```

`kill` returns immediately; nothing waits for the old `crowd` to actually exit. The `swift build` in between used to hide that — every trigger was a Swift edit, so a real recompile+relink bought seconds. **This commit makes the fast path reachable:** a web-only edit rebuilds via the resource-copy rule alone (0.07s for the copy step in my test package, plus SwiftPM planning).

Meanwhile the outgoing daemon isn't quick to die: `app.runService()` (`CrowDaemon.swift:473`) is a graceful shutdown that drains in-flight connections, and a browser tab holds `/rpc` + `/terminal` WebSockets open.

**Failure scenario:** `--watch` in poll mode with a browser tab attached; save `app.css`; `kill` → ~0.1–1s rebuild → new `crowd` starts → `acquireSingleInstanceLock` finds the flock still held by the shutting-down old process → `LOCK_NB` fails → `exit(1)` (`CrowDaemon.swift:56-60`; the store-writer lock at `:69-75` is fail-closed the same way). The loop doesn't check, so it prints `[daemon-run] (re)started crowd pid=N` for a process that's already dead. The developer sees "restarted", refreshes, and gets connection-refused — with no daemon until they touch another file.

One line fixes it (`|| true` matters — `wait` on a SIGTERM'd child returns 143 and `set -e` would abort):

```bash
if [ -n "$PID" ]; then kill "$PID" 2>/dev/null || true; wait "$PID" 2>/dev/null || true; fi
```

Only the poll fallback needs it — watchexec's supervisor sequences stop-then-spawn itself. Worth noting the poll path is the one that runs for anyone who hasn't `brew install watchexec`.

**🟡 2. `README.md:107` still says `--watch` is Swift-only**

```bash
bash scripts/daemon-run.sh --watch   # ...and rebuild + restart crowd on Swift changes
```

Stale as of this commit — `--watch` now covers web assets too. This is the same class of inaccuracy round 1 flagged at `docs/getting-started.md:112`; that one and the table at `:32` were fixed, this code-block comment was missed. → `# ...and rebuild + restart crowd on Swift or web-asset changes`.

**🟡 3. `README.md:172` contradicts both `--watch`'s behavior and `README.md:110`**

> `daemon-run --watch` rebuilds it on Swift *or* web-asset edits … a web-only rebuild just re-copies the assets, which the running daemon serves on the next refresh **without a restart**.

The "no restart" property is real, but it belongs to a **manual** `make daemon` against a running daemon — not to `--watch`, which is the subject of the sentence. Under `--watch`, *every* detected change restarts `crowd`: watchexec `-r` (`:77`) and the poll loop's `kill $PID` (`:99`) both tear it down, web edits included. `README.md:110` already states this correctly ("`--watch` restarts the daemon on every change"), so the two lines now disagree.

Not cosmetic: a restart drops open WebSockets and re-runs the single-instance guard (see #1). Suggested split — "a manual `make daemon` against an already-running daemon just re-copies the assets (⌘R picks them up, no restart); `--watch` always restarts."

`README.md:160`, by contrast, is correct as written — it's scoped to the `make run` sidecar + manual rebuild, and my SwiftPM test confirms it.

**🟢 4. `scripts/daemon-run.sh:10` header wasn't updated with the rest of the file**

> Pass `--watch` to also rebuild + restart `crowd` whenever a Swift source changes.

Contradicts `--help` at `:32` and the runtime log at `:70`, both of which correctly say "Swift or web-asset". Same fix as #2.

**🟢 5. The two watch backends now cover different file sets**

`watchexec -e swift,js,css,html,svg,json` (`:77`) filters by extension across all of `WATCH_PATHS`, so it also fires on `Packages/CrowDaemon/web-tests/*.test.js` and `web-tests/package.json` — editing a web *test* now rebuilds and restarts the daemon. The poll fallback (`:89/:91`) only adds `*/Resources/web/*`, so it doesn't. Conversely poll mode picks up the gitignored generated `Resources/web/version.json` (watchexec honors `.gitignore`, `find` doesn't), so a concurrent `make build` churns a restart. Both are harmless noise, but if you want them aligned, `-e swift` plus `-w Packages/CrowDaemon/Sources/CrowDaemon/Resources/web` (or a `--ignore` for `web-tests`) matches the poll semantics.

### Testing Notes

- Empirical SwiftPM resource-recopy test as described above — the one claim I couldn't settle by reading.
- `bash -n`, `--help`, `make -n daemon-run`, `make help`, `find`-filter match, `rg "crowd-dev"`, file mode: all verified.
- Did **not** run `swift build` / `swift test` on the repo: no `.build` here and the Swift/Rust hunks are comment-only, verified line-by-line.
- `shellcheck` still isn't installed locally, so CI's `shellcheck scripts/*.sh` (`ci.yml:85`, `release.yml:44`) is unverified from my side. The `\( … -o … \)` group and `-e` list are quoted correctly; I'd expect it clean.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 3 Yellow, 2 Green] findings. The round-1 fixes are correct and the frozen-asset model is verified working; what's left is one line in the poll loop (#1) and two stale README lines (#2, #3). No security concerns.

dgershman:
Round 2 addressed in `c2cf3d61` — thanks for empirically confirming the resource-recopy loop, and for catching the poll-loop race.

**🟡 1 — poll loop could leave no daemon** Real bug, and the fast web-only path is exactly what exposed it. Added your one-liner after the `kill`:
```bash
if [ -n "$PID" ]; then kill "$PID" 2>/dev/null || true; wait "$PID" 2>/dev/null || true; fi
```
`wait` blocks until the outgoing `crowd` drains and releases the flock, so the respawn wins `acquireSingleInstanceLock`. `|| true` keeps the 143 from `set -e`. Poll path only — watchexec `-r` sequences stop-then-spawn itself.

**🟡 2 — README:107** Fixed the code-block comment → "Swift or web-asset changes" (missed it in round 1 alongside :32/:112).

**🟡 3 — README:172 conflation** You're right: `--watch` always restarts (it contradicted :110). Rewrote so `--watch` = rebuild **+ restart** on any edit, and the no-restart property is scoped to a **manual** `make daemon` against an already-running daemon.

**🟢 4 — script header :10** → "Swift source or web asset", matching `--help` and the log line.

**🟢 5 — backend divergence** Confirmed `Packages/CrowDaemon/web-tests/*.test.js` exists and `Resources/web/version.json` is gitignored. Added `-i '**/web-tests/**'` so a web-*test* edit no longer churns a restart, matching the poll filter's `*/Resources/web/*` scope. Left the version.json edge alone — watchexec honoring `.gitignore` there is the more-correct behavior, and it only matters mid-`make build`.

`shellcheck scripts/daemon-run.sh` clean; `bash -n` and `--help` verified.

dhilgaertner (approved):
## Code & Security Review (round 3 — `c2cf3d61`)

All three round-2 Yellows are fixed, and I re-verified the frozen-asset model **from scratch** rather than carrying forward round 2's conclusion — this time end-to-end against a *live* process, which is the case `README.md:160/172` actually promises:

> Minimal SwiftPM package with `.copy("Resources/web")`, a long-running binary re-reading `Bundle.module` on a loop. Changed only the file's *contents* mid-run, then `swift build` → `[1/4] Copying web`, **0.08s**, and the **already-running process** flipped from `ORIGINAL` to `HOT-SWAPPED` on its next read. Binary mtime **unchanged** across the rebuild (`1784999971` → `1784999971`) — a resource-only edit runs the copy step and *doesn't relink*, so there's no rebuild-over-a-running-binary hazard either.

That settles both halves of the PR's central promise: `make daemon` re-copies on a web-only edit, and `StaticAssets.webResponse` (`StaticAssets.swift:66-80`) reads `Bundle.module` per request, so the running daemon serves the new bytes with no restart.

Round-2 fixes, each re-checked against the file:
- **`scripts/daemon-run.sh:101`** — `wait "$PID"` added after the `kill`, with `|| true` so the 143 from a SIGTERM'd child doesn't trip `set -e`. Correct, and correctly scoped to the poll path.
- **`README.md:107`** — now "Swift or web-asset changes".
- **`README.md:172`** — rewritten; `--watch` = always restarts, and the no-restart property is now scoped to a **manual** `make daemon`. No longer contradicts `README.md:110`.
- **`scripts/daemon-run.sh:10`** (green) and **`-i '**/web-tests/**'`** (green) both landed.

Mechanical re-verification: `bash -n` clean; `--help` matches the header and the `:70` log line; `make -n daemon-run` → `bash scripts/daemon-run.sh`; `make help` still column-aligned (`daemon-run` is exactly 10 chars); rename kept mode `100755`; `rg --hidden --no-ignore "crowd-dev"` across the repo → **zero** matches; poll-mode `find` filter matches all 8 shipped assets in `Resources/web`; `git check-ignore` confirms `Resources/web/version.json` is gitignored (`.gitignore:41`), as the response assumed.

### Security Review

**Strengths:**
- Dropping `--web-dir` from the default dev loop still **narrows** surface — no arbitrary on-disk directory served by default.
- Loopback pin (`scripts/daemon-run.sh:39`), CSP, and the `/xterm/:file` traversal guard (`StaticAssets.isSafeAssetName`) untouched.
- The new `wait` and `-i` flag introduce no new inputs, no eval, no word-splitting: `"$PID"` is quoted, the ignore glob is single-quoted so the shell can't expand it, and `START_CMD` still reaches `sh -c` as positional args.
- `CROW_WEB_DIR` documentation (`:21-22`) remains accurate — `CrowDaemon.swift:1181-1182` reads it *before* arg parsing at `:1215`, so an explicit `--web-dir` would still win, and the child inherits the env either way.

**Concerns:** none.

### Code Quality

**🟢 1. The poll-loop `wait` is unbounded — consider the app's bounded pattern (`scripts/daemon-run.sh:101`)**

The fix is right, and in the normal case it's strictly better than what it replaced. Worth noting the repo's *own* stop-crowd routine doesn't trust SIGTERM unconditionally — `crow-desktop/src-tauri/src/lib.rs:302-315` sends SIGTERM, polls for **1500 ms**, then falls back to `child.kill()` before reaping. The poll loop has no such deadline, so a `crowd` that wedges during graceful shutdown parks the loop inside `wait` with no daemon, no log line, and not even the `sleep 1` heartbeat — the terminal just goes quiet until Ctrl-C.

I don't think this is likely: `CrowDaemon.swift:479-480` states `runService()` returns cleanly on SIGTERM, and the shipped re-exec path (`requestReexec`, `:497-501`) raises SIGTERM from inside an RPC handler running over a live `/rpc` WebSocket and depends on that return — so an open browser tab demonstrably doesn't block shutdown. That's why this is green, not yellow. But if you want the dev loop to be as defensive as the window is:

```bash
kill "$PID" 2>/dev/null || true
for _ in $(seq 30); do kill -0 "$PID" 2>/dev/null || break; sleep 0.1; done
kill -9 "$PID" 2>/dev/null || true
wait "$PID" 2>/dev/null || true
```

**🟢 2. `scripts/daemon-run.sh:77-78` — "matching the poll filter" now overstates the alignment**

`-i '**/web-tests/**'` closes the web-tests half, but the comment's stronger claim ("only shipped assets under `Resources/web` should") isn't what the flags do. `-e swift,js,css,html,svg,json` applies across all of `WATCH_PATHS`, and `Packages/CrowTerminal` is in that list — so watchexec also fires on the 9 vendored `Resources/xterm/*.{js,css,html}` files, which the poll filter's `*/Resources/web/*` excludes. In the other direction, poll still fires on the gitignored generated `version.json` that watchexec skips (acknowledged, and I agree it's the more-correct behavior).

Both are harmless — vendored xterm assets change roughly never. It's the *comment* that's now inaccurate, not the behavior. Either soften it ("approximately matching"), or make it literally true with `-e swift` plus an extra `-w Packages/CrowDaemon/Sources/CrowDaemon/Resources/web`.

Related, same call: the extension list means a future non-listed asset type under `Resources/web` (a `.png`, a `.woff2`) would be picked up by poll mode but silently *not* by watchexec — the faster path, and the one most people have installed.

**🟢 3. `Makefile:52` help text reads like `make daemon-run --watch` is a thing**

> `daemon-run Run crowd serving the frozen bundle-baked web UI (add --watch to rebuild on Swift/web change)`

In a list of `make` targets, "add `--watch`" naturally parses as `make daemon-run --watch` — which fails: `make: unrecognized option '--watch'` (verified). `Makefile:74-75` hardcodes the no-arg invocation. `README.md:107` and `docs/getting-started.md:112` both spell out `bash scripts/daemon-run.sh --watch` correctly; only the `make help` line is ambiguous. Pre-existing phrasing carried through the rename, so entirely optional — `(bash scripts/daemon-run.sh --watch to rebuild on Swift/web change)` if you want it exact.

### Testing Notes

- **Ran:** the SwiftPM resource experiment above (build → run → edit contents mid-run → rebuild → observe the live process), plus `bash -n`, `--help`, `make -n daemon-run`, `make daemon-run --watch`, `make help`, `find`-filter match, `git check-ignore`, `git ls-files -s` for the mode, and `rg` for `crowd-dev`.
- **Did not run** `swift build` / `swift test` on the repo: no `.build` here, and the `CrowDaemon.swift` + `DaemonSecurityTests.swift` + `lib.rs` hunks are comment-only, verified line-by-line in the diff (all on their own `//` / `///` lines — nothing that can change parse).
- **Not verified:** `shellcheck` still isn't installed locally, so CI's `shellcheck scripts/*.sh` (`ci.yml:85`, `release.yml:44`) is unconfirmed from my side. Both new lines are quoted conventionally and I'd expect clean. `watchexec` also isn't installed, so the `-i '**/web-tests/**'` glob is reasoned about (gitignore-style globset, `-i` valid in both watchexec 1.x and 2.x), not executed.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 3 Green] findings. The rename is complete and clean, the frozen-asset dev loop is empirically confirmed working end-to-end, the poll-loop race from round 2 is properly fixed, and the docs now agree with each other and with the script. The three greens are optional polish — none should hold the merge. No security concerns.

PR #858, opened 2026-07-26
Gate agent_kind against the registry on every session-creation surface

### LINKED ISSUE #834: Agent-kind validation asymmetry across new-session / create-manager surfaces
## Context

Surfaced while documenting the coding-agent harness architecture (#827, ADR 0015 §8). Three session-creation surfaces validate the requested `AgentKind` against `AgentRegistry` inconsistently:

| Surface | Validates against registry? | Code |
|---|---|---|
| New-session | ❌ no check — `requestedAgentKind ?? agentKind(for: .work)` | `EngineRouter.swift:128` |
| Web `create-manager` | ✅ yes — CROW-593 security gate, falls back to configured default | `EngineRouter.swift:439-440` |
| Daemon `create-manager` | ❌ no check — passes the kind straight through | `RPCHandlers.swift:1092-1101` |

## Blast radius (limited)

A session persisted with an unregistered kind is not arbitrary code execution:
- `launchAgent` silently no-ops on `guard let agent = AgentRegistry.shared.agent(for:)` (`SessionService.swift:902`).
- `managerCommand` falls back to `AgentRegistry.defaultAgent` (`SessionService.swift:1385-1386`).

So the effect is a session that won't launch its agent (new-session path) or one that launches the default instead of the requested harness (daemon create-manager) — not an escape. But one of the three surfaces (web `create-manager`) is an **explicit security gate** (CROW-593) and the other two don't share it, which is the kind of inconsistency worth converging.

## Suggested fix

Route all three creation surfaces through a single registry-validated resolver (fall back to the configured default on an unregistered/unknown kind), so the CROW-593 gate isn't one-off. Add a test per surface.

## References
- ADR 0015 §8 (`docs/adr/0015-harness-capability-tiers.md`), PR #833
- Related: ADR 0014 (pluggable `CodingAgent` adapter), CROW-593 (registry gate)

dgershman:
Closes #834

## Problem

Three session-creation surfaces validated the requested `AgentKind` against `AgentRegistry` inconsistently — only the web `create-manager` ran the CROW-593 security gate:

| Surface | Validated before? |
|---|---|
| New-session (app, `EngineRouter`) | ❌ no check |
| Web `create-manager` (`EngineRouter`) | ✅ CROW-593 gate |
| Daemon `create-manager` (`RPCHandlers`) | ❌ passed straight through |
| Daemon `new-session` (`RPCHandlers`) | ❌ no check *(same bug, not in the ticket table)* |

A session persisted with an unregistered kind isn't code execution, but it either won't launch its agent (`launchAgent` no-ops on the registry miss) or silently launches the default instead of the requested harness.

## Fix

Converge every surface on a single registry-validated resolver:

- **`AgentRegistry.registeredKind(_:)`** — returns the requested kind only when an agent is registered for it, else `nil`, so each caller applies its own `?? <configured default>` fallback. The CROW-593 gate as a reusable primitive.
- Routed through it:
  - App `new-session` — was ungated.
  - Web `create-manager` — inline gate refactored to the primitive (behavior unchanged).
  - Daemon `new-session` — was ungated (fourth instance of the same asymmetry; fixed for symmetry so the two `new-session` surfaces don't diverge).
  - `SessionService.resolvedManagerAgentKind` — the single choke point **both** `create-manager` RPCs (web-forwarded + daemon-local) funnel through, so the daemon `create-manager` is covered without a call-site change.

Fallback semantics match the CROW-593 gate: only the *requested* kind is validated; the configured default is trusted as-is.

## Tests (one per surface)

- `AgentRegistryResolverTests` — the gate rejects an unknown kind and `nil`.
- `EngineRouterSmokeTests.newSessionGatesAgentKindAgainstRegistry` — app `new-session`: unregistered → configured default; registered non-default → honored.
- `DaemonAgentKindGateTests` — daemon `new-session`: same contract.
- `ManagerMigrationTests.resolvedManagerAgentKindPrefersExplicitOverride` — updated to the gated contract (registered explicit wins; unregistered falls back to the configured Manager default). Covers both `create-manager` surfaces via the shared choke point.

All three affected packages green: CrowCore (471), CrowEngine (343), CrowDaemon (151).

## Notes

- ADR 0015 §8 names this as the code follow-up (#834), scoped as code-only — no doc change.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Agent/AgentRegistry.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Agent/AgentRegistry.swift
@@ -27,6 +27,22 @@ public final class AgentRegistry: @unchecked Sendable {
         return agents[kind]
     }
 
+    /// The single registry gate for a caller-supplied `AgentKind` (CROW-593; #834).
+    ///
+    /// Returns `requested` only when an agent is actually registered for it;
+    /// otherwise `nil`, so the caller falls back to its own configured default.
+    /// Every session-creation surface (new-session and web/daemon
+    /// `create-manager`) funnels a requested kind through here so none can
+    /// persist a session with an unregistered/unknown kind — a persisted-but-
+    /// unlaunchable session (`launchAgent` no-ops on the registry miss) or one
+    /// that silently launches the default instead of the requested harness.
+    /// Composed with each surface's `?? <configured default>` fallback rather
+    /// than falling back itself, so the default choice stays with the caller.
+    public func registeredKind(_ requested: AgentKind?) -> AgentKind? {
+        guard let requested, agent(for: requested) != nil else { return nil }
+        return requested
+    }
+
     /// The agent to use when the caller doesn't specify one. Falls back to
     /// the first-registered agent.
     public var defaultAgent: (any CodingAgent)? {
--- a/Packages/CrowCore/Tests/CrowCoreTests/AgentRegistryResolverTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AgentRegistryResolverTests.swift
@@ -1,0 +1,30 @@
+import Foundation
+import Testing
+@testable import CrowCore
+
+// #834: `registeredKind` is the single registry gate every session-creation
+// surface (new-session + web/daemon create-manager) funnels a caller-supplied
+// `AgentKind` through, replacing the per-surface asymmetry where only the web
+// `create-manager` (CROW-593) validated. An unregistered/unknown kind or `nil`
+// returns `nil` so the caller falls back to its own configured default.
+//
+// The registered-kind passthrough is asserted at the surface level in
+// CrowEngineTests / CrowDaemonTests, where a real `ClaudeCodeAgent` is
+// registered — CrowCore has no concrete `CodingAgent` to register here.
+
+@Suite("AgentRegistry.registeredKind gate")
+struct AgentRegistryResolverTests {
+
+    @Test func rejectsAnUnregisteredKind() {
+        // A kind value no shipped agent claims — misses even if the process-wide
+        // registry has been populated by another test.
+        let unknown = AgentKind(rawValue: "crow-834-unregistered-resolver")
+        #expect(AgentRegistry.shared.agent(for: unknown) == nil)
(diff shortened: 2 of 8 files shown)


dhilgaertner (approved):
## Code & Security Review

Solid hardening of the CROW-593 registry gate. Converging every session-creation surface onto `AgentRegistry.registeredKind` correctly closes the ADR 0015 §8 asymmetries without widening the trust model (requested kind validated; configured default trusted as-is).

### Critical Issues
None.

### Security Review
**Strengths:**
- Single reusable gate (`AgentRegistry.registeredKind`) prevents persisting unregistered/unknown `agent_kind` values that previously left sessions unlaunchable or silently defaulted at launch.
- Coverage matches the threat model: app `new-session`, daemon `new-session`, web `create-manager` (refactored), and both `create-manager` paths via `SessionService.resolvedManagerAgentKind`.
- Fallback composition (`registeredKind(...) ?? configuredDefault`) keeps default selection at the caller — no accidental cross-surface default bleed.
- `handoff-agent` already hard-fails on unregistered kinds (`agentNotRegistered`); out of scope and correctly stricter than creation.

**Concerns:**
- None blocking. Silent fallback remains intentional (CROW-593); `new-session` responses return the resolved `agent_kind`, so callers can detect a mismatch.

### Code Quality
- **Green:** Consider a debug/ops log when the gate rejects a requested kind — useful when debugging “why did I get Claude instead of X?” without changing the fallback contract.
- **Green:** Web `create-manager` still parses `agent_kind` without the empty-string → `nil` filter used by `new-session` (`EngineRouter.swift` ~446 vs ~114–115). Empty strings still fail the registry check and fall back, so behavior is equivalent; aligning the parse would reduce drift.
- **Green:** `create-manager` responses still omit the resolved `agent_kind` (pre-existing). Echoing it would match `new-session` and make silent fallback visible to web clients.
- Tests are well-scoped (one per surface + shared choke point), use unique unregistered kind strings to avoid singleton pollution, and assert both reject and registered passthrough paths.
- Local: CrowCore / CrowEngine / CrowDaemon gate tests pass. CI (Build & Test, Desktop, Shell Lint) green.

### Summary Table

---

Issue #807, opened 2026-07-22
[parity] Anti-drift: enumerable CLI↔RPC↔config parity test + CI gate


dhilgaertner:
Foundation ticket for milestone **CLI control-plane parity** — the anti-drift backbone.

Add a build/CI **parity test** that fails when a new user-mutable capability ships without CLI coverage. All three surfaces are enumerable, so this is mechanical, not manual:

1. **RPC methods** — enumerate `CommandRouter.handlers.keys` (`CrowIPC/CommandRouter.swift`; registered in `RPCHandlers.swift makeCommandRouter`). Classify read vs write.
2. **CLI verbs** — enumerate `CrowCommand.configuration.subcommands` (recursively, incl. nested like `job`).
3. **Config fields** — enumerate `AppConfig` via `CodingKeys`/`Mirror` (`CrowCore/Models/AppConfig.swift`), recursing nested structs.

Assertions:
- Every **write** RPC method maps to a CLI verb (or is in `KNOWN_UI_ONLY` with a justification string).
- Every mutable `AppConfig` field has a CLI read **and** write path (or is justified in `KNOWN_UI_ONLY`).
- The exception allowlist is a checked-in file; adding to it requires a reason — that's the review gate.

Wire it into CI so a new RPC/config field/UI control without a CLI path turns the build red. This is what stops the drift the milestone targets. Depends on the ADR (#ADR) for the principle; feeds every per-area ticket (each removes its entries from `KNOWN_UI_ONLY`).

PR #867, opened 2026-07-27
tmux/daemon-maintenance CLI verbs (parity)

dhilgaertner:
Closes #818

Adds `crow` CLI verbs for the maintenance actions that were previously reachable only from the web — Settings → About's maintenance group and the session-header host-app buttons.

| CLI | RPC method | Params |
|---|---|---|
| `crow restart-manager` | `restart-manager` | — |
| `crow restart-tmux-server` | `restart-tmux-server` | — |
| `crow reload-tmux-config` | `reload-tmux-config` | — |
| `crow launch-agent --terminal <uuid>` | `launch-agent` | `terminal_id` |
| `crow retry-readiness --terminal <uuid>` | `retry-readiness` | `terminal_id` |
| `crow open-in-vscode --session <uuid>` | `open-in-vscode` | `session_id` |
| `crow open-terminal --session <uuid>` | `open-terminal` | `session_id` |

All seven RPC handlers already existed in `RPCHandlers.swift`, so this is CLI structs + registration + tests + docs — **no new backend behavior**. Follows the `crow edit-link` precedent (#805) rather than the heavier `crow job` one. Verbs are flat and named identically to their RPC methods, so there's no second naming layer to keep in sync.

## Design notes

- **No `--yes` on the destructive restarts.** The web confirms first; from the CLI the caller owns that choice, matching `RecreateTerminal`'s existing stance. The blast radius is spelled out in `--help` and the CLI reference instead.
- **No socket-timeout bump.** `restart-tmux-server` does a synchronous `kill-server` (capped at `TmuxController.defaultTimeout = 2.0`) and then fires the per-terminal rebuild as un-awaited `Task { @MainActor … }`s, so it replies almost immediately; `restart-manager` is a handful of 2s-capped tmux calls. Both fit inside the default 30s, unlike `job-run` which needed 120. Documented that `{"ok": true}` means *kicked off*, not *finished*, so scripts don't chain a `crow send` into a half-rebuilt surface.
- **`launch-agent` / `retry-readiness` take only `--terminal`**, 1:1 with the RPC params — unlike every other terminal verb, which takes `--session` *and* `--terminal`. Documented; the mismatch fails loudly at parse time (`Unknown option '--session'`) rather than being silently ignored, and there's a test pinning that.
[...]

PR #868, opened 2026-07-27
Session-lifecycle CLI verbs (parity)

### LINKED ISSUE #816: [parity] Session-lifecycle CLI: mark-in-review / complete / set-active / mark-issue-done / add-merge-label
Milestone: **CLI control-plane parity**. Area: **Session lifecycle actions** (web-only RPCs → CLI).

Add CLI verbs for the session row/menu actions currently only reachable from the web UI:
- mark-in-review, complete-session, set-session-active, mark-issue-done, add-merge-label

(These are RPC methods with no CLI verb today. `set-status`/`set-locked` already exist — align naming, avoid duplication.) Respect the app-down fallback duality (ADR 0009).

dhilgaertner:
Closes #816

Exposes the five session row/menu actions that were reachable only from the web UI as `crow` verbs. Each maps 1:1 onto the RPC method of the same name, so the CLI and the browser drive a session through its lifecycle by exactly the same path.

```bash
crow mark-in-review     --session <uuid>   # → {"session_id":…,"status":"inReview"}
crow complete-session   --session <uuid>   # → {"session_id":…,"status":"completed"}
crow set-session-active --session <uuid>   # → {"session_id":…,"status":"active"}
crow mark-issue-done    --session <uuid>   # → {"ok":true,"session_id":…}
crow add-merge-label    --session <uuid>   # → {"ok":true,"session_id":…}
```

## Three defects this surfaced

Adding a CLI caller exposed three problems in those handlers. All are fallout from the app retirement (ADR 0010): they were written as *"forward to the app when it's running, degrade to a local write when it's down"* (see the old comment at `RPCHandlers.swift:1105`). The forwarding branch is gone, so only the degraded branch has been running.

**1. `mark-issue-done` / `add-merge-label` returned `{"ok":true}` unconditionally.** `IssueTracker.markIssueDone` / `addMergeLabel` are `async -> Void` that `return` early on a missing ticket / PR / capability and swallow every provider error. The browser never trips this because `app.js` hides the menu items; a CLI has no equivalent affordance.

Reproduced against the current daemon, on a session with neither a ticket nor a PR:

```
$ crow add-merge-label --session <uuid>
{ "ok" : true }          # nothing happened
$ crow mark-issue-done --session <uuid>
{ "ok" : true }          # nothing happened
```

The preconditions now live server-side, so both report a real error.

**2. Status writes bypassed `SessionService`.** `complete-session` / `set-session-active` called a local `setSessionStatusLocally` helper instead of `SessionService.updateSessionStatus`, losing the manager guard and `scheduleAnalyticsSnapshot` — so completing a session never recorded its analytics snapshot. They now route through `SessionService`, falling back to the direct `appState` + `store` write only when there is no `SessionService` at all (the no-tmux host — that branch is load-bearing, not defensive).

**3. Manager sessions were silently skipped.** `updateSessionStatus` guards `isManagerSession` and returns, so the caller got a success receipt for a write that never happened. Now an explicit `applicationError`.

## Deliberate non-decisions

- **Not gated on current status.** The UI's `active` / `inReview` / `completed` conditions decide which menu items to *draw*, not what's legal. Enforcing them server-side would break idempotent scripting, so `crow complete-session` twice is fine. Pinned by a test.
- **`localOnlyDenial` untouched.** The web UI calls all five over `/rpc` today; gating them would break it. They read and write session state only — no secrets, no host process spawning.
- **`EngineRouter` copies left alone.** The daemon router shadows all five methods, so the engine copies are unreachable in practice.
- **`mark-in-review` still performs no provider board transition.** It writes session status only — `crow transition-ticket --to inReview` is the verb that moves the provider's board. Restoring the transition is a real fix but genuine new backend behavior; filing separately rather than smuggling it into a parity PR. Documented in `docs/cli-reference.md` so the distinction isn't a trap.

## Checklist notes

- **No `AppConfig` round-trip test** — these verbs touch no config fields; they act on session state. That checklist item is N/A for this area.
- **Parity harness step skipped.** The #807 anti-drift harness doesn't exist yet (`KNOWN_UI_ONLY` has zero hits repo-wide), so per the ticket there's nothing to register these in and this PR does not build it.

## Testing

`make test` — 15 packages, ~1,695 tests, zero failures.

- `SessionLifecycleParsingTests` (CrowCLI, CI-gating) — arg parsing, UUID rejection, root-command routing for all five.
- `SessionLifecycleRPCTests` (CrowEngine) — param decode and each guard, including that the error message names the fix.
- `BoardForwarderTests.LocalStatusTests` — rewritten; it previously pinned the old behavior. Adds manager rejection, unknown-session, missing-ticket, and idempotency cases.
- `LocalLiveActionTests` — adds the missing-ticket / missing-PR guard cases; the existing tracker-guard-before-param-validation ordering still holds.

Verified end-to-end against a live daemon that the CLI reaches every verb over the socket and that client-side UUID validation fires. The new handler behavior is covered by the daemon tests, which drive the real router (`makeCommandRouter` → `CommandRouter.handle`) against a real `AppState` / `JSONStore` — running the new daemon binary itself would have required stopping the developer's crowd, since only one daemon may hold the store writer lock.

Note for reviewers: CI's package allow-list (`ci.yml`) covers CrowCLI but **not** CrowEngine or CrowDaemon, so two thirds of these tests run only under local `make test`.

https://claude.ai/code/session_01WHPyzUYxDP2NEw7Q4VqpUi


dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Session IDs are UUID-validated on both CLI (`validateUUID`) and RPC (`SessionLifecycleRPC.sessionID`)
- Lifecycle verbs stay on the existing shared router surface (no new secret writes, no host-process spawn); leaving `localOnlyDenial` alone is consistent with web `/rpc` use
- Manager sessions are explicitly rejected for status transitions instead of silently no-opping

**Concerns:**
- No new authz boundary issues for a local/authenticated Crow peer. Same remote `/rpc` reachability as the pre-existing web menu actions.

### Code Quality

**Yellow — `mark-issue-done` / `add-merge-label` can still return `{"ok":true}` for a no-op** (`Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift` ~1167–1168, ~1206–1207)

The PR correctly gates missing ticket / PR, but after those guards the handlers still `await` `IssueTracker.markIssueDone` / `addMergeLabel` (`async -> Void`) and always emit `SessionLifecycleRPC.okResult`. Those tracker methods still return early on nil `session.provider`, missing `.autoMergeLabel` capability, unparseable PR slug, and every provider error (`IssueTracker.swift` ~3467–3517, ~3584–3618). That is the same false-success class this PR set out to close — just one layer deeper than the UI menu gates.

Minimal fix options (either is fine):
1. Make the tracker throw / return `Result` and map failures to `applicationError`, or
2. After the await, assert a postcondition (e.g. session reached `.completed` for `mark-issue-done`; merge-label sticky / `hasMergeLabel` for `add-merge-label`) and error if unmet.

**Yellow — SessionService path that restores analytics is untested** (`applySessionStatus` in `RPCHandlers.swift` ~69–75; `LocalStatusTests` in `BoardForwarderTests.swift`)

Defect #2 in the PR body is that status writes must go through `SessionService` so `.completed` schedules an analytics snapshot. `LocalStatusTests.seededRouter` builds `makeCommandRouter` with `sessionService: nil`, so tests only pin the no-tmux fallback write. Production daemons with tmux take the untested branch. Add at least one router test with a real `SessionService` that asserts `complete-session` updates status (and, if practical, that a snapshot is scheduled).

**Green — EngineRouter copies left unchanged** (`EngineRouter.swift` ~471–505). Documented as shadowed by the daemon router; still return unconditional `ok`. Fine for this PR; consider aligning or deleting in a follow-up so dead paths don't reintroduce the old lie if routing ever changes.

**Green — CI allow-list.** CrowEngine / CrowDaemon tests (including the new lifecycle suites) are not in `ci.yml`'s package allow-list. Worth widening so these guards don't regress silently.

**Green — Manager rejection on ticket verbs.** Status verbs reject managers; `mark-issue-done` / `add-merge-label` do not, while `app.js` never offers those items for `kind === 'manager'`. Low practical risk; consider rejecting for full UI parity.

### Static Analysis
- `SessionLifecycleRPCTests` (CrowEngine): 9/9 passed
- `SessionLifecycleParsingTests` (CrowCLI): 4/4 passed
- `LocalStatusTests` (CrowDaemon): 6/6 passed
- (SwiftPM also emitted an unrelated arm64/x86_64 XCTest bundle load warning in this environment; the Swift Testing runner still executed and passed.)

### Summary Table

---

dhilgaertner:
Thanks — both Yellows were right, and both are fixed in `4b4ffb9`. Branch is also rebased onto current `main` (it was 7 commits behind, including sibling parity PRs #867/#872 that touch the same files; merged clean).

## Yellow #1 — `{"ok":true}` for a no-op ✅ fixed

Correct, and worse than the review says: my guards covered a missing ticket / PR, but `markIssueDone` / `addMergeLabel` also returned silently on a nil `session.provider`, a backend without `.autoMergeLabel`, an unparseable repo slug, and every provider error. Same false-success class, one layer down.

Took **option 1**. Both are now `async throws` with a typed `SessionActionError` (new file `Packages/CrowEngine/Sources/CrowEngine/SessionActionError.swift`), and `mapRPCError` maps it to `applicationError`.

Chose converting the methods outright over adding `*Reporting` siblings (the `runNowReporting` shape) because **these two handlers are the only callers in the tree** — a sibling would have left the silent versions sitting there to be picked up again. That also let me delete the handler-level ticket/PR guards and `SessionLifecycleRPC.requirePRURL`: the tracker is now the single source of truth for these preconditions, and it can see the cases a handler can't.

Messages name the fix, since a CLI user is the one reading them:
```
add-merge-label needs a linked PR — attach one with `crow add-link --type pr --url …`
```

New `SessionActionReportingTests` (11 tests) pins each case the old code exited silently from, including the two a handler guard could never have caught — missing provider, unparseable repo slug.

## Yellow #2 — untested `SessionService` path ✅ fixed

Also correct: every `seededRouter` passed `sessionService: nil`, so the tests only covered the no-tmux fallback while production takes the other branch.

Added `completeSessionRoutesThroughSessionServiceAndSnapshots` — builds a real `SessionService` with an `analyticsProvider` probe and asserts the status lands in **both** `appState` and the store **and** that `.completed` reaches `writeAnalyticsSnapshot`. That last assertion is the one that would have caught defect #2 in the first place. Plus `managerRejectionHoldsWithSessionServicePresent`, since the manager guard now has two paths to hold on.

## Green — manager rejection on ticket verbs ✅ fixed

Added to both tracker methods, matching `app.js` never offering them for `kind === 'manager'`.

## Green — CI allow-list ❌ not doing, and it can't be done

I tried this rather than assuming. **CrowEngine does not compile on Linux**, so neither does CrowDaemon (which depends on it). In the same `swift:6.1` image CI uses:

```
$ docker run --rm -v "$PWD":/src -w /src swift:6.1 \
    swift build --package-path Packages/CrowEngine
/src/Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:1053:9:
  error: cannot find 'sharedSurface' in scope
```

CrowEngine → CrowTerminal, and `TmuxBackend` has macOS-only members outside the platform guards. Widening the allow-list needs CrowTerminal made Linux-buildable first — real work, and not this PR's. Worth its own issue; happy to file it. Until then these suites are local-`make test`-only, which I'll keep flagging in PR bodies.

## Green — EngineRouter copies

Agreed, leaving per your note. Worth folding into the same follow-up as the `mark-in-review` provider transition, since both are "the engine copies still tell the old lie."

## Test status

`make test` — all packages pass except one **pre-existing** failure I want to flag since it'll show up for you too:

`CrowTerminal › retryReadinessEmitsTimedOutWhenSentinelMissing` fails on **clean `origin/main`** in a fresh worktree, with no changes of mine anywhere near it (my diff doesn't touch CrowTerminal). It's a real tmux integration test that's sensitive to a live tmux server on the box — it passed on one isolated run and failed on the next with identical code. Not introduced here; likely worth its own issue.

Also worth knowing: `make test`'s package loop **does not propagate failures** — it exited 0 with that suite red. Anyone reading the exit code alone gets a false green. Separate cleanup, but it bit me here.

dhilgaertner:
Rebased onto `main` (now includes #872) — conflicts resolved, `mergeable: MERGEABLE`. New SHAs: `1935151` + `766a328` (the earlier reply refers to `4b4ffb9`, same content).

One cross-PR detail worth a look, since it means this PR modifies a line #872 just landed:

#872 and this branch independently renamed `mapJobRPCError` → `mapRPCError`. #872 kept it synchronous; this branch needs it `async`, because the lifecycle handlers `await` a main-actor hop inside the mapped body. An `async` function accepts a synchronous closure unchanged, so I resolved in favor of `async` and converted **all** call sites to `try await mapRPCError` — including the one #872 added:

```swift
// Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift:843
let patterns = try await mapRPCError { try AllowlistRPC.decodePatterns(params["patterns"]) }
```

Behavior there is unchanged (the closure is still sync; only the wrapper suspends). Flagging it because it is the one place this PR touches #872's code rather than its own. 12 call sites total: 7 from `main`, 5 new here. Merged doc comment names all three support types.

`make test` on the new base: every package green except the pre-existing `CrowTerminal › retryReadinessEmitsTimedOutWhenSentinelMissing`, which I verified fails on clean `origin/main` in a fresh worktree as well.

dgershman (changes requested):
## Code & Security Review

Re-review of the follow-up (`766a328`). Prior Yellows on false-success in `IssueTracker` and the untested `SessionService` path are fixed cleanly — typed `SessionActionError`, `mapRPCError` mapping, and `completeSessionRoutesThroughSessionServiceAndSnapshots` all look good.

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- UUID validation on CLI (`validateUUID`) and RPC (`SessionLifecycleRPC.sessionID`)
- No new privileged surface: lifecycle verbs stay on the existing `/rpc` / socket router; `localOnlyDenial` left alone is consistent with web menu use
- Manager sessions rejected for all five verbs; error messages avoid dumping secrets (provider failures truncated to 200 chars)
- Web `alertModal` renders via `textContent`, so surfaced error strings (including unparseable PR URLs) are not an XSS vector

**Concerns:**
- None new. Same remote `/rpc` reachability as the pre-existing web actions.

### Code Quality

**Yellow — `mark-issue-done` can still return `{"ok":true}` without completing the session** (`IssueTracker.markIssueDone` → `appState.onCompleteSession?(sessionID)` at `IssueTracker.swift` ~3530; handler at `RPCHandlers.swift` ~1194–1195)

The follow-up correctly makes preconditions and provider failures throw. The documented second half of the verb — “then completes the session” (`docs/cli-reference.md`, CLI `discussion`) — still goes through an optional `AppState` callback that is only wired inside `wireTerminalAutomations` when a `SessionService` (tmux cockpit) exists (`CrowDaemon.swift` ~336–342, ~722–724).

On the no-tmux host this PR otherwise carefully supports for status verbs (`applySessionStatus`’s direct `appState` + `store` fallback), `crow mark-issue-done` can close the provider issue and still emit `SessionLifecycleRPC.okResult` while the session stays `.active` — the same false-success class this PR claims to close (`docs/cli-reference.md`: “never print a success receipt for an action that didn’t happen”).

`complete-session` already works without `SessionService`; `mark-issue-done`’s completion half should use the same path (or assert `.completed` before returning ok). Minimal fix options:
1. After a successful `tracker.markIssueDone`, call `applySessionStatus(…, .completed, …)` in the handler (idempotent if the callback already ran), or
2. Have the tracker throw when completion cannot be applied / return only after status is `.completed`.

**Green — EngineRouter copies** (`EngineRouter.swift` ~493–527) still return unconditional `ok` via callbacks. Documented as shadowed; fine to leave for a follow-up with the provider-side `mark-in-review` transition.

**Green — Stale test comments.** `LocalLiveActionTests` (~141–145) and `SessionLifecycleRPCTests` (~7–10) still describe `IssueTracker` as `async -> Void` / handler-level ticket·PR guards. Behavior under test is correct; comments should match the throwing tracker.

**Green — CI allow-list.** CrowEngine / CrowDaemon suites remain local-`make test`-only (Linux build blocked by CrowTerminal). Already acknowledged; out of scope here.

### Static Analysis
- `SessionActionReportingTests` (CrowEngine): 10/10 passed
- `SessionLifecycleRPCTests` (CrowEngine): 7/7 passed
- `SessionLifecycleParsingTests` (CrowCLI): 4/4 passed
- `LocalStatusTests` (CrowDaemon): 8/8 passed (includes SessionService + snapshot path)

### Summary Table

---

dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Session IDs are UUID-validated on both the CLI (`validateUUID`) and server (`SessionLifecycleRPC.sessionID`).
- Manager sessions are rejected server-side for all five verbs — no silent no-op success receipts.
- Provider failures and unmet preconditions now throw `SessionActionError` and map to `applicationError` instead of `{"ok":true}` for a no-op.
- Trust model unchanged and appropriate: CLI over the 0600 Unix socket; web `/rpc` keeps calling these methods (gating would break the UI).

**Concerns:**
- None new. `mark-issue-done` / `add-merge-label` still invoke provider CLIs over `/rpc` (pre-existing); leaving `localOnlyDenial` untouched is correct for UI parity.

### Code Quality
- **Yellow — `mark-issue-done` can still return success without completing the session (no-tmux path).** `IssueTracker.markIssueDone` closes the provider issue, then calls `appState.onCompleteSession?(sessionID)` (`IssueTracker.swift` ~3530). That callback is wired only inside `wireTerminalAutomations` when a `SessionService` exists (`CrowDaemon.swift` ~722). On the no-tmux host — the same shape for which `complete-session` carefully falls back to a direct `appState` + `store` write in `applySessionStatus` (`RPCHandlers.swift` ~76–88) — `onCompleteSession` stays nil. A successful close therefore leaves the Crow session active while the handler still returns `SessionLifecycleRPC.okResult` (`RPCHandlers.swift` ~1194–1195). That contradicts the CLI docs / `MarkIssueDone` discussion (“On success the session flips to completed”) and is the same false-success class this PR is fixing for the precondition paths. **Fix:** after a successful `tracker.markIssueDone`, have the `mark-issue-done` handler also `applySessionStatus(..., .completed)` (idempotent when the callback already fired), or make completion a required, non-optional step inside `markIssueDone` that uses the same SessionService / store fallback as `applySessionStatus`. Add a test for the no-`SessionService` path asserting the session ends `.completed`.
- Green — `SessionLifecycleRPCTests` / `LocalLiveActionTests` comments still describe `IssueTracker` entry points as `async -> Void` that swallow errors; outdated after this PR.
- Green — `EngineRouter` copies of these five methods still return unconditional `{"ok":true}` via never-wired callbacks; unreachable today because the daemon router registers them first, but easy to leave as a future footgun.
- Green — CrowEngine / CrowDaemon tests (where the important guards live) are outside Linux CI’s package allow-list by existing ADR 0007 constraints; local `make test` coverage is solid.

### Summary Table

---

dhilgaertner:
Yellow fixed in `3c627ff`. You were right, and the test proves it — I verified it fails without the change (session stays `.active`), not just that it passes with it.

## Yellow — `mark-issue-done` could succeed without completing the session ✅

Confirmed the chain exactly as described: `markIssueDone` → `appState.onCompleteSession` → wired only in `wireTerminalAutomations` → gated on `if let sessionService` (`CrowDaemon.swift:336`). No tmux, no callback: the issue closed on the provider, the Crow session stayed active, and the handler still returned `{"ok":true}`.

Took your first option — the handler applies `.completed` after a successful close:

```swift
try await tracker.markIssueDone(sessionID: id)
_ = try await MainActor.run {
    try applySessionStatus(id: id, to: .completed,
                           appState: appState, store: store, sessionService: sessionService)
}
```

Reusing `applySessionStatus` over a second write keeps the SessionService/direct-store fallback in one place, and it is idempotent when the callback did fire. Preferred it to making completion mandatory inside `markIssueDone` because the tracker has no `SessionService` and would have had to grow its own copy of that fallback.

Test — `markIssueDoneCompletesSessionWithoutSessionService`, the no-`SessionService` path you asked for. A stub `ShellRunner` makes `gh issue close` succeed so the success path actually runs, then asserts `.completed` in **both** `appState` and the store. Reverting the handler change turns it red:

```
✘ Expectation failed: (appState.sessions.first?.status → .active) == .completed
✘ Expectation failed: (store.data.sessions.first?.status → .active) == .completed
```

## Green — stale comments ✅

Fixed. `SessionLifecycleRPCTests` and `LocalLiveActionTests` both still described the tracker entry points as `async -> Void` that swallow errors. The one in `SessionActionReportingTests` is deliberate past tense ("used to be") — it explains why the suite exists.

## Green — EngineRouter copies → filed as #876

Agreed it is a footgun; still declining to fix it inside a parity PR, since deleting reachable-looking handlers is a different kind of change than adding CLI verbs. Filed **#876** together with the `mark-in-review` provider-transition gap, since they share a root cause (both are app-retirement leftovers that still return unconditional `ok`). That also discharges the follow-up I promised in the first round rather than leaving it as a comment.

## Green — CI allow-list

Agreed, nothing to do here. Verified last round that CrowEngine cannot compile on Linux at all (`CrowTerminal/TmuxBackend.swift`: `cannot find 'sharedSurface' in scope` under `swift:6.1`), so this is blocked upstream regardless.

## Tests

`make test`: all packages green except the pre-existing `CrowTerminal › retryReadinessEmitsTimedOutWhenSentinelMissing`, which I verified reproduces on clean `origin/main` in a fresh worktree.

Also still worth flagging separately: `make test`'s package loop does not propagate failures — it exits 0 with a suite red.

dgershman (approved):
## Code & Security Review

Re-review of `3c627ff` (and the earlier `SessionActionError` / SessionService test fixes). Prior Yellows — silent `{"ok":true}` from `IssueTracker`, untested SessionService path, and no-tmux `mark-issue-done` leaving the session active — are addressed.

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- UUID validation on CLI (`validateUUID`) and RPC (`SessionLifecycleRPC.sessionID`)
- Manager sessions rejected server-side for all five verbs; no silent no-op success receipts
- Preconditions and provider failures throw `SessionActionError` → `applicationError` instead of `{"ok":true}` for a no-op
- Provider error strings truncated to 200 chars; web `alertModal` uses `textContent` (no XSS from surfaced URLs/messages)
- Trust model unchanged: CLI over 0600 Unix socket; `/rpc` left ungated so the web menu keeps working

**Concerns:**
- None new. `mark-issue-done` / `add-merge-label` still invoke provider CLIs over authenticated `/rpc` (pre-existing surface).

### Code Quality
- **Green — EngineRouter copies** (`EngineRouter.swift` ~493–527) still return unconditional `ok` via never-wired callbacks. Shadowed by the daemon router today; tracked as #876 — fine to leave out of this parity PR.
- **Green — CI allow-list.** CrowEngine / CrowDaemon suites remain local-`make test`-only (Linux blocked by CrowTerminal). Already acknowledged; out of scope here.
- **Green — Whitespace ticket URLs.** `SessionLifecycleRPC.requireTicketURL` trims; `IssueTracker.markIssueDone` only checks `!isEmpty`. Negligible in practice (`set-ticket` won't store blanks), but aligning trim would make the two entry points identical.

### Static Analysis
- `SessionActionReportingTests` (CrowEngine): 10/10 passed
- `SessionLifecycleRPCTests` (CrowEngine): 7/7 passed
- `SessionLifecycleParsingTests` (CrowCLI): 4/4 passed
- `LocalStatusTests` (CrowDaemon): 9/9 passed (includes SessionService snapshot path + no-tmux `mark-issue-done` completion)
- `LocalLiveActionTests` ticket-action cases: passed
- CI Build & Test / Desktop / Shell Lint: success

### Summary Table

---

Issue #604, opened 2026-07-08
Extend the crow CLI with full job management (job list / get / add / edit / enable / run / delete)


dhilgaertner:
## Summary

Extend the `crow` CLI with **full job management** — `list`, `get`, `add`, `edit`, `enable`/`disable`, `run` (run-now), `delete` (and ideally `duplicate`) — so scheduled jobs can be managed from the CLI / scripts / the Manager, not only through the in-app Settings UI. Today the CLI has **zero** job commands; jobs can only be created/edited/run from the Jobs UI.

## Motivation

Jobs (timed prompt-sets scoped to a repo — the "Jobs" sidebar feature, CROW-317) are powerful automation (e.g. hourly error-store triage → filing tickets), but they're only reachable through the GUI. A CLI surface would let me:
- Script job creation/edits and check them into config-as-code.
- List/inspect jobs and their run state from a terminal or the Manager session.
- Trigger a **run-now** on demand (e.g. kick a triage job without waiting for its interval).
- Enable/disable jobs programmatically.

It also removes the fragile workaround of hand-editing `{devRoot}/.claude/config.json` directly, which races the running app (the app owns that file + its in-memory state and can clobber external edits).

## Current state (grounded on `origin/main`)

- **CLI has no job commands.** The root command registers only session/terminal/metadata subcommands — `Packages/CrowCLI/Sources/CrowCLILib/CrowCommand.swift:7-36` (`subcommands: [...]`). No `job*` entries.
- **Jobs are a real, well-defined model.** `Packages/CrowCore/Sources/CrowCore/Models/JobConfig.swift` — `JobConfig { id, name, workspace, repo, prompts: [String], schedule: JobSchedule, enabled, lastRunAt, createdAt }` with `JobSchedule = .interval(seconds:) | .dailyAt(hour:minute:weekdays:)`. Stored on `AppConfig.jobs` (`Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift:39-41`), persisted to `{devRoot}/.claude/config.json` via `CrowPersistence/ConfigStore.saveConfig` (`ConfigStore.swift:68-95`).
[...]

PR #870, opened 2026-07-27
Telemetry/Cleanup/Sidebar CLI (parity)

### LINKED ISSUE #814: [parity] Telemetry / Cleanup / Sidebar CLI (AppConfig.telemetry, .cleanup, .sidebar)
Milestone: **CLI control-plane parity**. Area: **General misc** (Settings → General).

Add CLI for the remaining General-tab settings:
- `AppConfig.telemetry` — enabled, port, retentionDays
- `AppConfig.cleanup` — enabled, retentionHours
- `AppConfig.sidebar` — hideSessionDetails

Suggest `crow telemetry set`, `crow cleanup set`, `crow ui set` (or a shared `crow settings` group) — pick per CLI conventions.

dhilgaertner:
Closes #814

Exposes the last three web-only General-tab settings as `crow` verbs, following the `crow job` precedent from #604.

```bash
crow telemetry get
crow telemetry set --enabled true --port 4318 --retention-days 90
crow cleanup   set --enabled true --retention-hours 72
crow ui        set --hide-session-details true
```

## Design

**Granular RPCs, not a CLI-side blob round-trip.** `set-config` replaces the entire `AppConfig`, and CrowCLI cannot decode one (it has no CrowCore dependency), so a read-modify-write in the CLI would make it a second writer racing the web Settings modal. Six new daemon methods (`telemetry-get/set`, `cleanup-get/set`, `ui-get/set`) each do a locked read-modify-write of one subtree via `mutateConfig` — the same lock `set-config` and the scheduler take. They also put five scalars on the wire instead of the whole config, so unlike `get-config` there is nothing for `SettingsSecrets` to strip.

**Every `set` is a patch.** Absent (or null) leaves a value alone; a *present but wrong-typed* value is rejected rather than silently dropped — the bug in `job-edit`'s `params[key]?.stringValue` idiom, which reports success for a write it ignored. Passing no fields is an error, not a no-op: a no-op would still rewrite `config.json`, bumping its mtime and firing a spurious "Config reloaded" chime in every open browser.

**Validation is new.** None of these six fields had any server-side check. Ports are held to 1024–65535 (below needs root, which `crowd` lacks; above doesn't fit `TelemetryConfig.port`'s `UInt16` and would make `config.json` undecodable if it landed on disk). Telemetry retention is `>= 0` with `0` meaning keep-forever. Cleanup retention is `>= 1` — the cutoff is `now - retentionHours`, so a negative value pushed it into the *future* and swept every completed and archived session on the next board poll, worktree and branch included. Ranges are enforced client-side (fast feedback, and the only assertions CI's Linux job actually runs) and server-side (a raw RPC caller bypasses ArgumentParser).

**Take-effect semantics differ, so the CLI says so.** `telemetry.enabled`/`.port` are read once at daemon boot and the port is baked into every agent launch's `OTEL_EXPORTER_OTLP_ENDPOINT`, so `telemetry set` returns `"restart_required": true` when either actually changed and warns on stderr (stdout stays pure JSON). Cleanup is live within a board poll; the sidebar preference repaints browsers within ~2s. All three responses carry `restart_required` so `| jq -e .restart_required` is a uniform contract.

Booleans are `@Option ... Bool?` rather than `@Flag`, because a patch needs three states — true, false, and "not provided". The cost is that ArgumentParser accepts only the literals `true`/`false`, so `--enabled 1` fails; the help text says so explicitly.

## Riders

Two changes beyond the ticket's letter, both deliberate:

**1. A config-wipe path this ticket's commands would otherwise widen.** `SidebarSettings`, `TelemetryConfig`, and `CleanupConfig` used synthesized `Codable`, so every property was required — unlike their siblings `ConfigDefaults` and `TerminalSettings`, which hand-write `init(from:)` with per-key defaults. `AppConfig`'s `decodeIfPresent` only tolerates a *wholly absent* block, so `{"telemetry": {"enabled": true}}` threw `keyNotFound`, failing the entire `AppConfig` decode, making `ConfigStore.loadConfig` return `nil` — at which point every writer's `?? AppConfig()` fallback would **silently overwrite `config.json` with defaults**, losing every workspace, job, and credential.

Fixed both ends: the three structs now decode per-key with defaults, and `mutateConfig` (renamed from `mutateJobs`, which was never job-specific) distinguishes "no config yet" from "config exists but won't decode" and refuses to overwrite the latter. That hardens the seven existing `job-*` call sites for free. `mapJobRPCError` → `mapRPCError` in the same pass; both were file-private with zero external references.

**2. `crow ui set` now actually repaints.** The daemon already pushes a `configReloaded` event when `config.json`'s mtime moves — its doc comment explicitly names CLI writes — but the client treated it as a chime only. Two lines in `app.js`'s `onServerNotify` re-read the view-affecting config slice, placed above the `_soundArmed` gate since a CLI write is not a user gesture. No new notification event, so the catalog check is unaffected.

## Security

**Deliberately not gated** in `RPCWebSocketHandler.localOnlyDenial`. A remote authenticated peer can already change all six fields through the un-gated `set-config` path — that is what `setConfigHarmlessToggleIsAllowedRemotely` asserts — so gating the granular methods would only push remote callers back onto the whole-config blob, which ships workspaces, jobs, gateway URLs and credential shells instead of five scalars. Un-gating is the smaller surface, not merely the consistent one. Rationale recorded in the ledger comment and pinned by `settingsRPCsAreAllowedRemotely`. `defaults.binaries` remains the sole local-only config field.

## Tests

1,828 tests pass across all 15 packages (`make test`), plus the notification-event catalog check.

- `SettingsRPCSupportTests` — patch decode (absent vs null vs wrong-type), range boundaries, exact wire shapes, `restart_required`, and a no-leak assertion that no response contains a credential key.
- `SettingsHandlerTests` (new) — the **real router** against a real `config.json` in a temp dev root: patches touch only the fields given, `--enabled false` is a write not a no-op, unrelated config (workspaces/jobs/credentials) survives a settings write, and a corrupt `config.json` is left byte-identical rather than replaced with defaults.
- `SettingsCommandParsingTests` — arg parsing including the load-bearing `--enabled false` ≠ absent case and group routing.
- `ValidationTests`, `AppConfigTests` — range checks and partial-block decodes (the last three fail without Rider 1). Both packages are on CI's Linux allow-list.
- `DaemonSecurityTests` — the un-gating decision.

Note that `CrowEngine` and `CrowDaemon` are **not** on CI's Linux allow-list (they pull macOS-only `CrowTerminal`), so their assertions only run under `make test` locally and in the macOS release job. Run locally; green.

## Checklist deviations

- **`README.md` unchanged** — it has no per-command docs, only a link to `docs/cli-reference.md` (line 191), which is where the reference went. `CLAUDE.md` and its template got a short block as the ticket asks.
- **Parity harness skipped** — #807 is open with no PR, and `KNOWN_UI_ONLY` has zero hits across every reachable revision. Nothing to add to yet.
- **Follow-ups, not fixed here:** `CrowDaemon.swift:400`'s engine-fallback `applyConfig` saves without `withConfigLock` (pre-existing); `docs/cli-reference.md` still has no `crow job` section (a #604 gap).

https://claude.ai/code/session_016A7PZAB6bo9Trh9my8XyGw


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Granular `telemetry-*` / `cleanup-*` / `ui-*` RPCs put five scalars on the wire instead of a whole `AppConfig` blob — smaller remote surface than pushing callers onto `set-config` (`RPCWebSocketHandler.swift:124-132`, pinned by `settingsRPCsAreAllowedRemotely`).
- Responses are hand-built literals with no credential keys; `settingsResponsesNeverContainCredentialKeys` locks that against a future encoder reuse.
- Server-side range checks close real footguns: port outside `UInt16` / privileged range would make `config.json` undecodable; cleanup `retention_hours < 1` would advance the cutoff into the future and sweep every eligible session + worktree (`SettingsRPCSupport.swift:36-72`).
- `mutateConfig` refuses to overwrite an existing-but-undecodable `config.json` (`RPCHandlers.swift:1440-1454`) — closes a silent wipe of workspaces/jobs/credentials that this ticket's write path would otherwise widen. Covered by `writeRefusesToOverwriteAnUndecodableConfig`.
- Patch decode rejects wrong types instead of treating them as absent (`patchBool` / `patchInt`) — avoids the `job-edit` success-on-ignored-write bug.

**Concerns:**
- None blocking. Un-gating for authenticated remotes matches existing `set-config` reachability for these fields; `defaults.binaries` remains the sole local-only config surface.

### Code Quality
- Patch semantics, empty-params rejection, `--enabled false` vs absent, and `restart_required` contracts are clear and well-tested (CLI parse + `SettingsRPC` unit + real-router handler tests).
- Rider 1 (per-key defaults on `TelemetryConfig` / `CleanupConfig` / `SidebarSettings`) is the right fix for partial-block decode failures cascading into full-config wipe.
- `app.js` reloads UI config on `configReloaded` *before* the `_soundArmed` gate so CLI writes actually repaint — correct placement.
- Green: `SecretRoutes.swift` still has three `loadConfig(...) ?? AppConfig()` write sites with the same wipe pattern `mutateConfig` now refuses. Worth the same guard in a follow-up (password / gateway writers).
- Green: Author-noted pre-existing `CrowDaemon.swift` `applyConfig` path still saves without `withConfigLock` — out of scope here, still worth tracking.

### Static Analysis / Tests
- `swift build` (root): clean.
- CrowCLI settings/validation tests: pass.
- CrowCore `AppConfigTests` (incl. partial-block decode): pass.
- CrowEngine `SettingsRPCSupportTests`: pass.
- CrowDaemon `SettingsHandlerTests` + `settingsRPCsAreAllowedRemotely`: pass.

### Summary Table

---

PR #871, opened 2026-07-27
Gateways & web-password CLI (local-only, parity)

### LINKED ISSUE #815: [parity] Gateways & web-password CLI (local-only gated secrets)
Milestone: **CLI control-plane parity**. Area: **Gateways & secrets** (local-only).

Add CLI for the local-only secret surfaces the UI writes via dedicated POST routes (not plain config transit). All must be **gated to the local Unix socket / loopback** (mirror `RPCWebSocketHandler.localOnlyDenial` and the existing `/config/*` local-only routes):
- Manager AI gateway — baseURL + customHeaders (`AppConfig.managerGateway`; POST `/config/manager-gateway`)
- Per-workspace AI gateway — baseURL + customHeaders (POST `/config/workspace-gateway`)
- Web access password — set / change / remove (PBKDF2 hash; POST `/config/web-password`; plaintext never persisted)

Out of scope: Jira credential (managed externally via `op://` refs) — document as a `KNOWN_UI_ONLY` exception with reason.

dhilgaertner:
Closes #815

Adds `crow gateway` and `crow web-password` for the three local-only secret surfaces that were previously writable only from the web Settings UI.

## Why new RPC methods

These fields live in `AppConfig`, but `set-config` **cannot** write them — `SettingsSecrets.preservingSecrets` always overwrites `managerGateway`, `workspaces[].gateway`, and `webAuth` from disk. The only writer was the three HTTP POSTs in `SecretRoutes`, which the CLI can't reach (Unix-socket JSON-RPC only, no HTTP client in `CrowCLILib`).

The two alternatives were worse: writing `config.json` from the CLI process is unsafe because `ConfigStore.withConfigLock` is a process-local `NSLock` and wouldn't serialize against the daemon; teaching the CLI to POST would mean a new HTTP client plus host/port discovery. So: granular RPC methods gated in `localOnlyDenial`, which is what the ticket specifies and what `run-setup` (an equally privileged write+re-exec) already does.

## Commands

```
crow gateway get   [--manager | --workspace <name|uuid>] [--reveal]
crow gateway set   [--manager | --workspace <name|uuid>] --base-url URL --header "K: V" ...
crow gateway clear [--manager | --workspace <name|uuid>]
crow web-password  status | set [--stdin] | clear
```

Workspace gateways resolve by **name or UUID** — a UX win over the web UI, which requires a saved `id`. An ambiguous name in a hand-edited config errors rather than guessing.

## Security posture

- **All four methods gated** in `RPCWebSocketHandler.localOnlyDenial`. Reads included: `gateway-get --reveal` returns the gateway auth headers, and `web-password-set` from a remote peer would let it change the password gating remote access.
- The stale NOTE in `RPCHandlers.swift` asserting secret writes live in `SecretRoutes` and "never here" is **rewritten** to explain the gate — leaving it would have misled the next reader into thinking no gate existed.
- `gateway get` **redacts header values by default** (mirroring `SettingsSecrets.strippedForTransport`); `--reveal` opts in. A caller that omits `reveal` gets redacted values, so the default fails safe.
- `web-password set` has **no `--password` flag** on purpose — plaintext in argv lands in shell history and local `ps`. It prompts twice with echo off, or reads stdin with `--stdin`.
- The web Settings UI is unaffected: it reads gateway state via `get-config` (stripped) and writes via `SecretRoutes`.

## Reuse over reimplementation

Handlers feed `SecretRoutes.buildGateway` / `mergingPreservedHeaders` rather than duplicating validation, so the both-or-neither invariant and the "blank header value keeps the stored secret" contract hold on the CLI path — including the guard against persisting a shape that would make the next `loadConfig` return `nil` and wipe the config. Header parsing goes through `WorkspaceGateway.parseHeaderLines`, so no third parser appears next to the Swift and `settings.js` ones.

## Tests

All 1721 tests pass (`make test`). New coverage:

| Target | Covers |
|---|---|
| `CrowCLITests` | arg parsing, target mutual exclusion, malformed vs blank-valued headers, password input |
| `CrowDaemonTests` | param decode/encode, plus end-to-end handler round-trips through the real router against temp dev roots (set → redacted get → `--reveal` → blank-value preserve → clear) |
| `DaemonSecurityTests` | the local-only gate, and that neighbouring methods stay ungated |
| `CrowPersistenceTests` | `webAuth` persistence round-trip |

Note the password tests consolidate into one case: each `web-password-set` runs real 210k-iteration PBKDF2, which is seconds per call in a debug build. The cheap crypto properties are already pinned at `iterations: 2000` in `DaemonSecurityTests`, so only the integration properties are asserted here.

## Deferred / out of scope

- **The #807 parity harness does not exist yet** — `KNOWN_UI_ONLY`, `ParityTest`, and `coveredSet` all return zero hits repo-wide. The "add to covered set / remove from `KNOWN_UI_ONLY`" checklist item is therefore **deferred, not done**; this PR does not build the harness.
- The **Jira credential** is documented in `docs/cli-reference.md` under "Not Exposed on the CLI" as a deliberate exception (an `op://` reference managed externally), ready for #807 to lift into `KNOWN_UI_ONLY`.

## Verification note

Handler behavior is covered end-to-end by the daemon tests against temp dev roots, and the CLI→socket wire was confirmed live. A full CLI-against-daemon smoke test was **not** run: `crowd` holds a global store-writer lock on App Support regardless of `--socket`, so an isolated second instance can't start, and I wasn't going to stop the running daemon or write to a real config's gateway/password.

https://claude.ai/code/session_019RtWKbWW9GC6n7FPsJXvhM

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -169,6 +169,21 @@ Events: `taskComplete`, `agentWaiting`, `reviewRequested`, `changesRequested`, `
 
 Notifications cascade — one fires only if `globalMute` is off, the global category toggle is on, **and** the per-event toggle is on. Omitted flags leave their stored value alone; every `--event-*` flag requires `--event`.
 
+### Gateways & Secrets
+
+Local-only (CROW-815) — these carry credentials, so the remote `/rpc` web path refuses them and only the local Unix socket works. Take exactly one target: `--manager` or `--workspace <name|uuid>`.
+
+```
+crow gateway get [--manager | --workspace <name|uuid>] [--reveal]        → {"gateway_set":true,"base_url":"...","headers":{...}}
+crow gateway set --manager --base-url URL --header "Name: Value" ...     → {"saved":true,"gateway_set":true}
+crow gateway clear --manager                                            → {"saved":true,"gateway_set":false}
+crow web-password status                                                → {"password_set":true,"iterations":210000}
+crow web-password set [--stdin]                                         → {"saved":true,"password_set":true}
+crow web-password clear                                                 → {"saved":true,"password_set":false}
+```
+
+`gateway get` blanks header values unless `--reveal`. A `--header` with a blank value (`--header "X-Api-Key:"`) keeps the stored secret — that's how to change a base URL without restating credentials. `web-password set` prompts twice with echo off; pipe with `--stdin` for scripts. There is no `--password` flag on purpose (shell history / `ps`).
+
 ## Important Notes
 
 - `--session` always expects a full UUID (e.g., `a1b2c3d4-e5f6-7890-abcd-ef1234567890`), not a session name
--- a/Packages/CrowCLI/Sources/CrowCLILib/Commands/SecretCommands.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Commands/SecretCommands.swift
@@ -1,0 +1,231 @@
+import ArgumentParser
+import CrowIPC
+import Foundation
+
+// Local-only secret surfaces: the AI gateways and the web-access password
+// (CROW-815). These reach the daemon over the Unix socket, which is local by
+// construction; the same RPC methods are refused on the remote `/rpc` WebSocket
+// by `RPCWebSocketHandler.localOnlyDenial`. The web Settings UI drives the
+// equivalent HTTP routes in `SecretRoutes`.
+
+/// Which gateway a `crow gateway` subcommand addresses: the Manager's, or one
+/// workspace's. Shared by get/set/clear so the selector behaves identically.
+public struct GatewayTargetOptions: ParsableArguments {
+    @Flag(name: .long, help: "Target the Manager AI gateway")
+    public var manager: Bool = false
+
+    @Option(name: .long, help: "Target a workspace's AI gateway (workspace name or UUID)")
+    public var workspace: String?
+
+    public init() {}
+
+    /// Called explicitly from each subcommand's `validate()` rather than relying
+    /// on option-group validation, so the check runs at a predictable point.
+    public func validateTarget() throws {
+        switch (manager, workspace) {
+        case (true, .some):
+            throw ValidationError("--manager and --workspace are mutually exclusive.")
+        case (false, nil):
+            throw ValidationError("Exactly one of --manager or --workspace is required.")
+        case (false, .some(let workspace)):
+            guard !workspace.trimmingCharacters(in: .whitespaces).isEmpty else {
+                throw ValidationError("--workspace must not be blank.")
+            }
+        case (true, nil):
+            break
+        }
+    }
+
+    /// The target selector as RPC params.
+    public var params: [String: JSONValue] {
+        if let workspace { return ["workspace": .string(workspace)] }
+        return ["target": .string("manager")]
+    }
+}
+
+/// Parent command for AI gateway management: `crow gateway <subcommand>`.
+///
+/// A gateway is a base URL plus the auth headers sent with it; agents launched
+/// into the workspace (or the Manager) inherit them as `ANTHROPIC_BASE_URL` /
+/// `ANTHROPIC_CUSTOM_HEADERS`. Header values may be `op://…` 1Password
+/// references, resolved at launch so the secret never rests in `config.json`.
+///
+/// Local-only: these run over the Unix socket and are refused for remote web
+/// clients, because the headers are credentials.
+public struct Gateway: ParsableCommand {
+    public static let configuration = CommandConfiguration(
+        commandName: "gateway",
+        abstract: "Manage AI gateways (local-only)",
+        subcommands: [GatewayGet.self, GatewaySet.self, GatewayClear.self]
+    )
+
+    public init() {}
+}
+
+public struct GatewayGet: ParsableCommand {
+    public static let configuration = CommandConfiguration(
+        commandName: "get",
+        abstract: "Show a gateway's base URL and header names",
+        discussion: """
+        Header values are blanked by default so the output is safe to share. \
+        Pass --reveal to print the stored values (which may be op:// references \
+        rather than the secrets themselves).
+        """
+    )
+
+    @OptionGroup public var target: GatewayTargetOptions
+
+    @Flag(name: .long, help: "Print header values instead of blanking them")
+    public var reveal: Bool = false
+
+    public init() {}
+
+    public func validate() throws {
+        try target.validateTarget()
+    }
+
+    public func run() throws {
+        var params = target.params
+        params["reveal"] = .bool(reveal)
+        printJSON(try rpc("gateway-get", params: params))
+    }
+}
+
+public struct GatewaySet: ParsableCommand {
+    public static let configuration = CommandConfiguration(
+        commandName: "set",
+        abstract: "Set a gateway's base URL and headers",
+        discussion: """
+        Replaces the whole gateway: --base-url and at least one --header are \
+        both required (a gateway needs both, or neither). A --header with a \
+        blank value keeps the secret already stored under that name, so the base \
+        URL can be changed without restating credentials:
+
+          crow gateway set --manager --base-url https://gw.example.com --header "X-Api-Key:"
+
+        Header values may be op:// 1Password references, resolved at agent \
+        launch so the secret never rests in config.json.
+        """
+    )
+
+    @OptionGroup public var target: GatewayTargetOptions
+
+    @Option(name: .customLong("base-url"), help: "Gateway base URL")
+    public var baseURL: String
+
+    @Option(name: .long, parsing: .singleValue, help: "Header as 'Name: Value' (repeatable)")
+    public var header: [String] = []
+
+    public init() {}
+
+    public func validate() throws {
+        try target.validateTarget()
+        guard !baseURL.trimmingCharacters(in: .whitespaces).isEmpty else {
+            throw ValidationError("--base-url must not be blank. Use `crow gateway clear` to remove a gateway.")
+        }
+        guard !header.isEmpty else {
+            throw ValidationError("At least one --header is required (a gateway needs both a base URL and a header).")
+        }
+        try header.forEach(validateHeaderLine)
+    }
+
+    public func run() throws {
+        var params = target.params
+        params["base_url"] = .string(baseURL)
+        params["header_lines"] = .array(header.map { .string($0) })
+        printJSON(try rpc("gateway-set", params: params))
+    }
+}
+
+public struct GatewayClear: ParsableCommand {
+    public static let configuration = CommandConfiguration(
+        commandName: "clear",
+        abstract: "Remove a gateway"
+    )
+
+    @OptionGroup public var target: GatewayTargetOptions
+
+    public init() {}
+
+    public func validate() throws {
+        try target.validateTarget()
+    }
+
+    public func run() throws {
+        var params = target.params
+        params["clear"] = .bool(true)
+        printJSON(try rpc("gateway-set", params: params))
+    }
+}
+
+/// Parent command for the web-access password: `crow web-password <subcommand>`.
+///
+/// The password gates the daemon's web UI for non-loopback clients. It is stored
+/// as a PBKDF2-HMAC-SHA256 hash; the plaintext is never persisted and never
+/// travels over anything but the local Unix socket.
+public struct WebPassword: ParsableCommand {
+    public static let configuration = CommandConfiguration(
+        commandName: "web-password",
+        abstract: "Manage the web-access password (local-only)",
+        subcommands: [WebPasswordStatus.self, WebPasswordSet.self, WebPasswordClear.self]
+    )
+
+    public init() {}
+}
+
+public struct WebPasswordStatus: ParsableCommand {
+    public static let configuration = CommandConfiguration(
+        commandName: "status",
+        abstract: "Report whether a web-access password is set"
+    )
+
+    public init() {}
+
+    public func run() throws {
+        printJSON(try rpc("web-password-get"))
+    }
+}
+
+public struct WebPasswordSet: ParsableCommand {
+    public static let configuration = CommandConfiguration(
+        commandName: "set",
+        abstract: "Set or change the web-access password",
+        discussion: """
+        Prompts twice with echo off. For scripts, pipe the password and pass \
+        --stdin:
+
+          printf '%s' "$PW" | crow web-password set --stdin
+
+        There is no --password flag on purpose — a plaintext password in argv is \
+        visible in shell history and to local `ps`. Changing the password does \
+        not require the old one; the local-only gate is the control.
+        """
+    )
+
+    @Flag(name: .long, help: "Read the password from stdin instead of prompting")
+    public var stdin: Bool = false
+
+    public init() {}
+
+    public func run() throws {
+        let password = try SecretArgs.readPassword(useStdin: stdin)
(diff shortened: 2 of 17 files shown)


dgershman (changes requested):
## Code & Security Review

Solid CROW-815 delivery: CLI surfaces reuse `SecretRoutes` validation, reads fail closed without `--reveal`, password stays out of argv, and all four RPC methods are gated in `localOnlyDenial` with matching `DaemonSecurityTests` coverage. CI is green; focused `SecretsRPC` / CLI Secret tests pass locally.

### Critical Issues
None.

### Security Review
**Strengths:**
- All four methods (`gateway-get` / `gateway-set` / `web-password-get` / `web-password-set`) are listed in `RPCWebSocketHandler.localOnlyDenial` — reads included, so `--reveal` cannot leak headers to a remote `/rpc` peer.
- `gateway-get` redacts header values unless `reveal` is explicitly true (`params["reveal"]?.boolValue ?? false`).
- No `--password` flag; TTY `getpass` / `--stdin` avoids shell history and `ps` exposure.
- Handlers reuse `SecretRoutes.buildGateway` / `mergingPreservedHeaders`, preserving the both-or-neither invariant and the blank-value-keeps-secret contract (including the #623 config-wipe guard).
- Mutations go through `ConfigStore.withConfigLock` via `mutateSecrets`; PBKDF2 hash/salt never returned from `web-password-get`.
- RPCHandlers NOTE correctly rewritten to document the gate instead of claiming secrets never live on this router.

**Concerns:**
- **Yellow** — `Packages/CrowDaemon/Sources/CrowDaemon/SecretRoutes.swift:19-22` still says secret writes are dedicated HTTP POSTs *“rather than JSON-RPC methods”* because the shared router “can't tell a local caller from a logged-in remote one.” That rationale is obsolete: this PR adds the JSON-RPC path and solves locality via `localOnlyDenial`. The PR correctly treated the parallel stale NOTE in `RPCHandlers` as misleading for the next reader of a secret surface — update this module doc the same way (point at the CLI RPC methods + the WebSocket gate) so nobody “fixes” security by removing the RPC path or assuming no gate exists.

### Code Quality
- **Green** — `SecretsRPC.decodeHeaderLines` validates each `header_lines` entry, then `joined(separator: "\n")` + `parseHeaderLines`. An entry containing an embedded newline can still split into extra headers after join. Consider rejecting `\n` / `\r` inside a single line (CLI `validateHeaderLine` too) so typos fail loud, consistent with the malformed-header philosophy.
- **Green** — `getpass(3)` leaves the secret in a process-wide static buffer until the next call; you copy into a `String` immediately (good). Optional follow-up: zero or replace with a termios-based prompt when convenient.
- Test coverage is strong (CLI parsing, daemon round-trips, local-only gate, blank-header preserve / reject, workspace name/UUID / ambiguity). Deferring the #807 parity harness is clearly documented.

### Summary Table

---

dhilgaertner:
Rebased onto `main` (picking up #867) and addressed the review in c73ca8b.

**Yellow — stale `SecretRoutes` module doc.** Fixed. You were right that this was the same class of problem as the `RPCHandlers` NOTE and I'd missed the second copy. It now describes both doors — the browser's Origin-checked POSTs and the CLI's gated RPC methods — and ends with an explicit warning that removing either gate re-opens the hole the type exists to close, so nobody "simplifies" it away later.

**Green — embedded newlines in `header_lines`.** Fixed at both layers (daemon `decodeHeaderLines` and CLI `validateHeaderLine`), so one `--header` always means one header.

One thing worth flagging from that fix: **the obvious check doesn't work.** `contains("\n")` does *not* catch CRLF, because Swift treats `"\r\n"` as a single `Character` — so a grapheme comparison misses it entirely, and `contains("\r")` misses it too. My first version had exactly that bug; the new CRLF test caught it. Both sites now use `rangeOfCharacter(from: .newlines)`, which also covers U+2028/U+2029.

**Green — `getpass(3)` static buffer.** Not taken, per your "optional follow-up" framing — the secret is copied into a `String` immediately, and a termios prompt adds platform nuance I'd rather not fold into this PR. Happy to do it separately if you'd prefer.

### Rebase note

#867 and this PR both added sections at the same anchors in `CLAUDE.md`, `docs/cli-reference.md`, and `LocalOnlyRPCGateTests`. All three resolved by keeping both sides. The merged `localOnlyDenial` doc reads coherently: tmux-maintenance methods stay explicitly ungated, gateways/web-password are gated.

### Pre-existing test failure, not from this branch

`make test` fails `CrowTerminal / retryReadinessEmitsTimedOutWhenSentinelMissing` (`TmuxBackendTests.swift:395`). **This reproduces identically on `origin/main` @ 369ea6e with this branch's code absent** — I checked out a clean worktree at the base to confirm. It passes when run alone and fails in the full suite, so it looks like test interference in #867's tmux suite rather than a real regression. Flagging it rather than fixing it here.

Everything else is green: 1785 tests across all other packages.

dhilgaertner:
Re-requesting review — no new feedback to address here, so flagging why this pinged again.

The `CHANGES_REQUESTED` state from the 19:30 review is sticky in GitHub until a new review is submitted; there are **no inline comments and no follow-up review** on the PR. All three findings from that review were fixed in what is now `3d1c7c4`, and I've confirmed they survived the four rebases since:

| Finding | Status | Where |
|---|---|---|
| **Yellow** — stale `SecretRoutes` module doc | Fixed | `SecretRoutes.swift:19-32` — now documents both doors and warns that removing either gate re-opens the hole |
| **Green** — embedded newlines in `header_lines` | Fixed | `SecretsRPCSupport.swift:90` (daemon) + `Validation.swift:185` (CLI) |
| **Green** — `getpass(3)` static buffer | Not taken, per your "optional follow-up" framing | Explained in the [earlier comment](https://github.com/corveil/crow/pull/871#issuecomment-5095907795) |

Both newline guards use `rangeOfCharacter(from: .newlines)` rather than `contains("\n")` — worth repeating since it's the non-obvious part: Swift treats `"\r\n"` as a single `Character`, so a grapheme comparison misses CRLF entirely. My first attempt had that bug and the CRLF test caught it.

Branch is currently rebased onto d8289ef (#870) and `MERGEABLE`. Full suite green at 1807 tests.

**One unrelated thing worth your attention:** `CrowTerminal / retryReadinessEmitsTimedOutWhenSentinelMissing` (`TmuxBackendTests.swift:395`) has now failed on four consecutive bases — 369ea6e, 1cb5300, 5a85f43, d8289ef. I reproduced it on a clean `origin/main` worktree with this branch's code absent, so it is not from #815. It passes in isolation and fails in the full suite, which points at ordering-dependent interference in #867's tmux suite. It's been on `main` through four merges and will keep looking like a regression on whichever PR happens to be in flight — probably worth its own ticket.

dgershman (changes requested):
## Code & Security Review

Prior Yellow (stale `SecretRoutes` module doc) and Green (embedded newlines in `header_lines`) are addressed. CI is green; focused `SecretsRPC` (24) and CLI `Secret*` (21) tests pass locally. One remaining should-fix on the new write path.

### Critical Issues
None.

### Security Review
**Strengths:**
- All four methods gated in `RPCWebSocketHandler.localOnlyDenial` (reads included); `DaemonSecurityTests.secretMethodsAreLocalOnly` pins it.
- `gateway-get` redacts header values unless `reveal` is explicit (`?? false`).
- No `--password` flag; TTY `getpass` / `--stdin` avoids argv/`ps` exposure.
- Handlers reuse `SecretRoutes.buildGateway` / `mergingPreservedHeaders` (both-or-neither + blank-value preserve + #623 wipe guard).
- `SecretRoutes` module doc now correctly describes both doors and warns that removing either gate re-opens the hole.
- Header newline smuggling rejected on CLI and daemon (`rangeOfCharacter(from: .newlines)`), with CRLF coverage in tests.
- `web-password-get` never returns hash/salt.

**Concerns:**
- **Yellow** — `mutateSecrets` (`RPCHandlers.swift` ~1584–1586) still does `ConfigStore.loadConfig(devRoot:) ?? AppConfig()`. The adjacent `mutateConfig` (CROW-814 / this PR's base) documents and refuses exactly that pattern: `loadConfig` returns nil for both missing *and* malformed, so falling back to defaults then saving silently destroys every workspace, job, and credential. `SettingsHandlerTests.writeRefusesToOverwriteAnUndecodableConfig` already pins the safer behavior for telemetry/cleanup/ui. A corrupt `config.json` plus `crow gateway set` / `crow web-password set` would wipe the file and write only the secret change. Apply the same exists-but-undecodable refuse (or route through / share `mutateConfig`'s load helper) and add a parallel daemon test. (`SecretRoutes`' HTTP writers still use `?? AppConfig()` — pre-existing; this PR's new writer should not reintroduce the hole next to the fixed path.)

### Code Quality
- **Green** — `mutateSecrets` docstring still says it mirrors `mutateJobs`, which was renamed to `mutateConfig`. Update the reference (or collapse the two helpers once the refuse-to-overwrite logic is shared).
- **Green** — `getpass(3)` static buffer remains an optional follow-up (already noted; secret is copied into a `String` immediately).

### Summary Table

---

dhilgaertner:
Fixed in 5efca5f. Good catch — this was a real data-loss bug, not a style point.

**Yellow — `mutateSecrets` fallback to defaults.** You're right, and the failure mode is worse than it looks: `loadConfig` returns nil for both "missing" and "malformed", so one `crow gateway set` against a corrupt `config.json` would have replaced every workspace, job and credential with defaults and written back only the gateway. Sitting directly beside `mutateConfig`, which CROW-814 added to refuse exactly that.

**I took the "share the load helper" option rather than copying the guard.** `mutateSecrets` is deleted; both writers now call `mutateConfig`. It was already generic over the transform, so the separate helper only ever existed to word the failure message differently — not worth a second load path next to a data-loss hole. Net −19 lines in the handler.

That also disposes of the first Green: the stale `mutateJobs` docstring reference disappears with the function.

**Test** mirrors yours on the secrets side — `secretWritesRefuseToOverwriteAnUndecodableConfig`: corrupt `config.json`, both `gateway-set` and `web-password-set` return `applicationError`, and the file is asserted byte-identical afterwards.

**`getpass(3)`** — still not taken, per your "optional follow-up" framing. Say the word if you'd rather it land here.

### On the `SecretRoutes` HTTP writers

You scoped this to my new writer and I've respected that, but I want to be explicit about what's left: those three POST handlers still do `?? AppConfig()`, so **the same wipe is reachable from the web UI** — corrupt `config.json`, then save a gateway or password in Settings. The CLI path is now safe and the browser path isn't.

I'd rather not fold a change to three security-critical HTTP handlers into this PR unannounced, so it's not here. Happy to either file it or fix it in a follow-up — your call.

Full suite green at 1839 tests. (`CrowTerminal / retryReadinessEmitsTimedOutWhenSentinelMissing` still fails on `main` itself — fifth consecutive base, reproduced with this branch absent; unrelated to #815.)

dgershman (approved):
## Code & Security Review

Prior Yellows are resolved: `SecretRoutes` module doc describes both doors + the WebSocket gate; embedded newlines in `header_lines` are rejected on CLI and daemon; secret writes now go through `mutateConfig` (refuse-to-overwrite corrupt `config.json`) with `secretWritesRefuseToOverwriteAnUndecodableConfig` coverage. Focused `SecretsRPC` (25) and CLI `Secret*` (21) tests pass locally; `CrowDaemon` builds clean.

### Critical Issues
None.

### Security Review
**Strengths:**
- All four methods gated in `RPCWebSocketHandler.localOnlyDenial` (reads included); `DaemonSecurityTests.secretMethodsAreLocalOnly` pins it.
- `gateway-get` redacts header values unless `reveal` is explicit (`?? false` — fail closed).
- No `--password` flag; TTY `getpass` / `--stdin` keeps plaintext out of argv/`ps`.
- Handlers reuse `SecretRoutes.buildGateway` / `mergingPreservedHeaders` (both-or-neither, blank-value preserve, #623 wipe guard).
- Header newline smuggling blocked with `CharacterSet.newlines` (covers CRLF), which matters because values feed `ANTHROPIC_CUSTOM_HEADERS` (newline-separated).
- `web-password-get` never returns hash/salt; PBKDF2 via `PasswordHash.make`.

**Concerns:**
- None that block merge.

### Code Quality
- **Green** — `SecretsRPC.decodeHeaderLines` returns `[:]` when `header_lines` is present but not an array (e.g. a string). The set still fails via both-or-neither if `base_url` is set, but a typed `invalidParams` ("must be an array of strings") would be clearer.
- **Green** — `CLAUDE.md` / `SecretsRPCSupport` say these methods are Unix-socket-only; behavior is local-direct (loopback `/rpc` also allowed, matching `SecretRoutes`). README is accurate — tighten the other two for consistency.
- **Green** — `SecretRoutes` HTTP writers still use `loadConfig ?? AppConfig()` (pre-existing); worth the same refuse-to-overwrite treatment as a follow-up, outside this PR's CLI path.
- **Green** — `getpass(3)` static buffer remains an optional follow-up (secret is copied into a `String` immediately).

### Summary Table

---

dgershman (changes requested):
## Code & Security Review

Local-only gating, reuse of `SecretRoutes` validators, redacted `gateway-get`, and no `--password` flag are all solid. CI green; focused `SecretsRPC` (25), CLI `Secret*` (21), and `secretMethodsAreLocalOnly` pass locally. One should-fix on the new write path.

### Critical Issues
None.

### Security Review
**Strengths:**
- All four methods listed in `RPCWebSocketHandler.localOnlyDenial` (reads included); `DaemonSecurityTests.secretMethodsAreLocalOnly` pins it.
- `gateway-get` fails closed on omit (`reveal` defaults `false`); `web-password-get` never returns hash/salt.
- No `--password` flag; TTY `getpass` / `--stdin` keeps plaintext out of argv/`ps`.
- Handlers reuse `SecretRoutes.buildGateway` / `mergingPreservedHeaders` (both-or-neither, blank-value preserve, #623 wipe guard).
- Embedded newlines in `header_lines` rejected with `CharacterSet.newlines` (CLI + daemon) — matters because values feed newline-separated `ANTHROPIC_CUSTOM_HEADERS`.
- Secret writes go through `mutateConfig` (refuse-to-overwrite corrupt config); covered by `secretWritesRefuseToOverwriteAnUndecodableConfig`.

**Concerns:**
- **Yellow** — `gateway-set` with neither `clear: true` nor a real gateway body silently clears the gateway. `decodeHeaderLines(nil)` → `[:]`, then `buildGateway` treats empty URL+headers as clear (`SecretRoutes.swift` ~167). Contrast `web-password-set` in the same PR, which rejects an empty password unless `clear: true` (`RPCHandlers.swift` ~1214–1215). The handler comment claims base URL + headers are "required" otherwise — that is not what the code does. A partial/typo'd RPC call can wipe stored gateway credentials. Keep `buildGateway`'s empty-body clear for the HTTP editor, but in the RPC handler require explicit `clear: true` (or a non-empty body) before persisting `nil`, and add a round-trip test.

### Code Quality
- **Green** — `SecretsRPC.decodeHeaderLines` returns `[:]` when `header_lines` is present but not an array. The set still fails via both-or-neither if `base_url` is set; a typed `invalidParams` would be clearer.
- **Green** — Some docs say Unix-socket-only; behavior is local-direct (loopback `/rpc` also allowed). README is accurate — tighten CLAUDE.md / `SecretsRPCSupport` for consistency.
- **Green** — Validation/`SecretsRPC` error strings interpolate the full header line (may include secrets) into local stderr/RPC errors. Prefer echoing only the header name when rejecting.

### Summary Table

---

Issue #805, opened 2026-07-22
crow CLI: add delete-link / edit-link commands (links are add-only today)


dhilgaertner:
The crow CLI exposes only `add-link` and `list-links` — a session link, once created, cannot be removed or relabeled from the CLI. A mislabeled link (e.g. an overly verbose label) can only be cleaned up in the app UI, and Manager-driven automation has no way to correct its own mistakes.

Ask:
- `crow delete-link --session <uuid> --link <uuid>` — remove a link by id (ids already come back from `add-link` / `list-links`)
- `crow edit-link --session <uuid> --link <uuid> [--label "..."] [--url "..."] [--type ...]` — update in place (or accept delete+re-add as the answer if edit isn't worth it)
- Document both in the CLI reference in the crow-workspace skill / CLAUDE.md

Context: hit on 2026-07-22 while linking a PR to a session (max-monorepo MAXX-8074 dark-factory run) — the first link got a long label, and replacing it required adding a second link and leaving the stale one for manual UI cleanup. Also noted in the dev-root CLAUDE.md Known Issues.

[🐦‍⬛ Created with Crow via Claude Code](https://github.com/radiusmethod/crow)

Issue #769, opened 2026-07-20
Restore web parity: crowd autostart / login-item installer (start Crow at login)


dhilgaertner:
## Summary

Add a **daemon autostart / login-item installer** for `crowd`, lost in the native→web move
(ADR-0010, CROW-593). The retired native macOS app launched at login (bringing its daemon with
it); now the user must start `crowd` manually every time.

## Background

ADR-0010 (`docs/adr/0010-retire-the-macos-app.md`, line 21) documents this explicitly:

> `crowd` has no autostart yet, so you start it yourself (a terminal, `tmux`, or your own
> `launchd` plist) until a login-item installer lands.

So on a fresh machine or after a reboot, Crow simply isn't running until the user remembers to
launch the daemon — a real regression from the always-there native app.

## Scope

- [ ] Provide an installer that registers `crowd` to start at login (macOS `launchd`
      LaunchAgent plist is the natural fit; keep it cross-platform-friendly given the Linux
      daemon effort, #645).
- [ ] Expose install/uninstall + status (running / at-login enabled) — CLI subcommand
      (e.g. `crow autostart install|uninstall|status`) and/or a Settings → General control.
- [ ] Idempotent install/upgrade; the plist points at the current `crowd` binary and survives
      app/binary updates (re-point on upgrade).
- [ ] Graceful behavior if already running (don't double-launch) and clear logging.

## Acceptance criteria

- [ ] A user can enable "start Crow at login" in one step; after reboot the daemon is running
      with no manual command.
- [ ] Uninstall cleanly removes the login item / LaunchAgent.
- [ ] Status is reportable (enabled? running?).
- [ ] Re-install/upgrade re-points to the correct `crowd` path (no stale plist).

## Open questions

- macOS-only (`launchd`) first, or design the abstraction for the Linux daemon (#645) now?
- Where the control lives (CLI vs. Settings vs. both).

## References

- ADR-0010 (`docs/adr/0010-retire-the-macos-app.md`) line 21 — documents the missing autostart.
- Related: #645 (Linux native daemon), the daemon entry point (`crowd`), and Settings → General.

PR #872, opened 2026-07-27
Allowlist & analytics CLI verbs (parity)

### LINKED ISSUE #819: [parity] Allowlist & analytics CLI: promote/refresh/list-allowlist, rebuild/get-scorecard, get-state, list-artifacts
Milestone: **CLI control-plane parity**. Area: **Allowlist & analytics** (web-only RPCs → CLI).

Add CLI verbs for the remaining web-only surfaces:
- Allowlist: promote-allowlist, refresh-allowlist, list-allowlist
- Analytics: rebuild-scorecard, get-scorecard, get-state, list-artifacts

Some are read-only inspection (get-state, get-scorecard, list-artifacts) — still valuable for scripting/parity. Note promote-allowlist is desktop/app-only today; confirm daemon path or record as a justified exception.

dhilgaertner:
Closes #819

Exposes the last web-only control-plane RPCs as `crow` verbs, following the PR #604 (jobs) precedent.

| Verb | Params | Result |
|---|---|---|
| `crow list-allowlist` | — | `{entries:[{pattern,is_global,worktree_session_names}], loading}` |
| `crow promote-allowlist` | `--pattern` (repeatable) | `{ok, added[], already_global[], global_settings_path}` |
| `crow refresh-allowlist` | — | `{ok:true}` |
| `crow get-scorecard` | — | the `ScorecardDTO` itself |
| `crow rebuild-scorecard` | — | `{rebuilt:true}` |
| `crow get-state` | — | the `DaemonStateSnapshot` itself |
| `crow list-artifacts` | `--session` | `{images:[{name,size,mtime,url,path}], dir}` |

## Naming

Flat, RPC-mirroring names rather than `crow allowlist <list\|promote\|refresh>` groups. Every RPC-backed verb in this CLI is spelled exactly like its method; the two nested groups are the exceptions that prove it (`autostart` isn't RPC-backed, `job` mirrors the already-namespaced `job-*` methods). These seven are `verb-noun`, so grouping would invert word order and make them the first verbs not derivable from their method by rule — the drift the parity milestone exists to prevent.

## The ticket's `promote-allowlist` caveat is stale

It is **not** desktop/app-only any more. `AllowListService` is pure disk I/O and `crowd` constructs it unconditionally (`CrowDaemon.swift:122-127`), so the daemon handler works headless. No `KNOWN_UI_ONLY` exception needed.

Two things fed that reputation, both fixed here: the nil-branch error strings claimed a "provider-configured daemon" (copy-paste from `refresh-tickets` — the allowlist needs no provider), and the `EngineRouter` copy calls `AppState.onPromoteToGlobal`, which is **never assigned anywhere** in the repo. That copy is dead but left in place — it's unreachable (the daemon's router owns both method names and `fallback` is consulted only for *unhandled* methods) and belongs to the app-forwarding path a later milestone deletes wholesale.

## Three defects the CLI surfaced

Making this scriptable turned three tolerable-behind-a-button bugs into unusable ones. All fixed:

1. **Silent write failure** — the write was wrapped in a `do/catch` that only `NSLog`ged, so the RPC answered `{"ok":true}` for a write that never landed. `promoteToGlobal` now throws.
2. **Data loss** — `AllowListService.swift:65-68` used `try?` + `as?`, so a `~/.claude/settings.json` that existed but didn't parse (trailing comma, half-written file) degraded to `[:]` and was written back containing **only** `permissions.allow`, destroying the user's `hooks`, `env`, `model`, `statusLine`, `mcpServers`. Now refused untouched, and the write is `.atomic`.
3. **Partial promotion reported as success** — `compactMap { $0.stringValue }` silently dropped non-string elements, so `{"patterns":["Read",42]}` granted one of two and returned `ok:true`. `AllowlistRPC.decodePatterns` now rejects it outright.

## Response shape changes (both additive, both verified against the consumers)

- `promote-allowlist` gains `added` / `already_global` / `global_settings_path`. `app.js:3271-3281` discards the result entirely and already wraps the call in a `catch` rendering `alertModal('Promote failed: …')` — so `ok:true` is preserved and the only behavior change is that a genuinely failed write now alerts the user instead of silently pretending.
- `list-artifacts` gains a per-image `path` and a top-level `dir`. The pre-existing `url` only resolves against the daemon's own web server, so a shell caller could do nothing with it. Computed server-side because `CrowCLI` doesn't depend on `CrowCore`, and `ArtifactPaths` exists precisely so the `$TMPDIR/crow/artifacts/<uuid>` convention has one owner — a CLI process's `$TMPDIR` need not match a launchd-spawned `crowd`'s. `app.js:1565` reads only `name`/`size`/`mtime`/`url`. A non-UUID `session_id` keeps today's behavior (empty list, `dir` omitted) rather than tightening to an error.

## Deliberately not done

- **`maxMessageSize` is left alone.** `get-state` can exceed the 1 MB socket frame on a busy install (the snapshot carries every assigned issue's full body). Raising a shared protocol constant would skew a new `crowd` against an old `crow`, so instead `GetState.run()` catches `SocketError.responseTooLarge` and names the narrower reads. Paging `get-state` is a follow-up.
- **The #807 parity harness / `KNOWN_UI_ONLY` list doesn't exist on this branch** (zero hits repo-wide), so that checklist step is skipped rather than invented here.
- **`README.md` unchanged** — it has no verb table and already links to `docs/cli-reference.md`. The ticket checklist predates the doc split; #805 and #769 both landed cli-reference + CLAUDE.md with no README edit.
- **Follow-up worth filing:** `docs/cli-reference.md` has no `crow job` section at all — PR #604 shipped nine verbs undocumented there.

## Tests

| Suite | Count | Covers |
|---|---|---|
| CrowCLI | 126 | arg parsing, the `.singleValue` regression guard (`Bash(npm run build:*)` must stay one element), `normalizedAllowlistPatterns`, root registration of all 7 verbs |
| CrowEngine | 18 | decode strictness incl. the non-string-element guard; the data-loss guard asserts the malformed file is **byte-identical** afterwards |
| CrowDaemon | 9 | handler shapes through the real `makeCommandRouter`; failed write → `applicationError` **and `result == nil`** |

⚠️ **CI's package allowlist covers `CrowCLI` but not `CrowEngine` or `CrowDaemon`** (`.github/workflows/ci.yml`), so those two suites only run under `make test`. Both were run locally and pass. The CI-covered `normalizedAllowlistPatterns` test is the client-side backstop.

`make test` is green except `TmuxBackendTests.retryReadinessEmitsTimedOutWhenSentinelMissing`, which is **pre-existing and unrelated** — verified failing identically with these changes stashed. It's a real-tmux timing test expecting a 150ms timeout that instead observed `.shellReady`; `CrowTerminal` depends only on `CrowCore`, neither of which this PR touches.

## Verification

Read paths, arg validation (exit 64 on each), and transport were smoked live against a running `crowd`. The new *server* behavior is covered by the daemon handler tests rather than a live smoke: `crowd` refuses a second instance because the store writer lock is global per Application Support dir regardless of `--socket`, and stopping the running daemon would have disrupted live sessions.

https://claude.ai/code/session_01CLZiRUwPgvjNoq7XUB4fFJ


dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- `promoteToGlobal` now throws on write/parse failure instead of answering `{"ok":true}` after an `NSLog` (`AllowListService.swift:102-153`, daemon handler `RPCHandlers.swift:811-829`).
- Malformed or non-object `~/.claude/settings.json` is refused untouched, and the write is `.atomic` — closes the silent clobber of `hooks` / `env` / `mcpServers` / etc.
- `AllowlistRPC.decodePatterns` rejects non-string / empty payloads so a mixed array can no longer partially promote and still look successful.
- `list-artifacts` keeps the UUID traversal guard (`Artifacts.dir`); additive `path`/`dir` are omitted for non-UUIDs.
- `get-state` continues to strip credentials via `SettingsSecrets.strippedForTransport` before encoding.

**Concerns:**
- None blocking. Residual notes below are Green.

### Code Quality
- CLI surface is clean: `.singleValue` for `--pattern`, shared `normalizedAllowlistPatterns`, UUID validation on `list-artifacts`, and a useful `responseTooLarge` message on `get-state`.
- Test coverage is strong across CrowCLI / CrowEngine / CrowDaemon for the three defects called out in the PR; those suites passed locally (`swift test` filters).
- Dead `EngineRouter` `promote-allowlist` path (`EngineRouter.swift:368-375`) still uses `compactMap` + unconditional `ok:true` and `onPromoteToGlobal` is never assigned. Unreachable today because the daemon router owns the method — fine to leave for the app-forwarding cleanup milestone, but worth deleting/aligning soon so the old bug can't resurface if ownership shifts.
- Same class as the fixed top-level parse guard: `permissions["allow"] as? [String] ?? []` (`AllowListService.swift:125`) still degrades a wrong-shaped `allow` value to `[]` and would rewrite only the requested patterns. Unlikely with Claude Code's string arrays; consider refusing when the key is present but not `[String]`.
- PR correctly notes CrowEngine/CrowDaemon aren't on the CI package allowlist — client-side `normalizedAllowlistPatterns` is the CI backstop. Expanding the allowlist (or a follow-up) would lock the server-side fixes in.

### Summary Table

---

dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Honest promote reporting: write failures throw instead of returning `{"ok":true}` (`AllowListService.promoteToGlobal`, daemon handler).
- Malformed / non-object `~/.claude/settings.json` is refused untouched; write uses `.atomic` — closes a real data-loss / clobber path.
- Strict `AllowlistRPC.decodePatterns` rejects non-string elements (no partial promote + success).
- `list-artifacts` keeps UUID traversal guard via `Artifacts.dir`; `path` uses `lastPathComponent` names from a directory listing.
- `get-state` continues to strip credentials via `SettingsSecrets.strippedForTransport` (pre-existing; CLI only surfaces it).

**Concerns:**
- None Red. Promote remains local-socket privilege (equivalent to editing the settings file) — appropriate for this surface.

### Code Quality
- **Yellow** — “Promote everything not yet global” recipe is wrong for patterns with spaces (the common case this PR emphasizes). Default `xargs` splits on whitespace, so `Bash(npm test:*)` becomes two tokens (`Bash(npm` / `test:*)`). Reproduced locally. Appears in both `docs/cli-reference.md` (~L311–313) and `AllowlistCommands.swift` discussion. Fix with newline-safe quoting, e.g. `xargs -d '\n'` (GNU) or a `while IFS= read -r` loop that passes `--pattern "$p"`, and update both copies.
- **Green** — Residual sibling of the data-loss fix: `permissions["allow"] as? [String] ?? []` still degrades a present-but-wrong-typed `allow` (or non-object `permissions`) to `[]` / `[:]` and can rewrite over it. Rare for Claude settings, but the same “refuse rather than clobber” posture would be consistent.
- **Green** — `EngineRouter`'s `promote-allowlist` still uses `compactMap` + dead `onPromoteToGlobal` and answers unconditional `ok`. Unreachable while the daemon owns the method (fallback only for unhandled names); fine to leave for the later app-forwarding cleanup, but worth not regressing if ownership ever flips.
- Tests for CrowCLI / CrowEngine / CrowDaemon allowlist+artifacts suites passed locally. Note: CI still doesn’t run CrowEngine/CrowDaemon (called out in the PR).

### Summary Table

---

PR #873, opened 2026-07-27
Notifications CLI (parity)

### LINKED ISSUE #813: [parity] Notifications CLI: global + per-event settings (AppConfig.notifications)
Milestone: **CLI control-plane parity**. Area: **Notifications** (Settings → Notifications).

Add `crow notifications get|set` for `AppConfig.notifications`:
- global: globalMute, soundEnabled, systemNotificationsEnabled
- per-event (10 events: taskComplete, agentWaiting, reviewRequested, changesRequested, checksFailing, autoWorkspaceCreated, autoMergeEnabled, autoRebasePushed, autoRebaseConflicts, configReloaded): enabled, soundEnabled, systemNotificationEnabled, soundName (validate against the 14 built-in sounds)

dhilgaertner:
Closes #813

Exposes `AppConfig.notifications` — until now editable only through the web Settings → Notifications tab — as `crow` verbs, following the `crow job` precedent (#604).

```bash
crow notifications get                      # globals + all 10 events + available_sounds
crow notifications get --event checksFailing
crow notifications set --global-mute
crow notifications set --no-global-mute --sound-enabled
crow notifications set --event checksFailing --event-sound-name Hero --no-event-sound-enabled
```

## Approach

Two granular RPCs (`notifications-get` / `notifications-set`) rather than routing the whole `AppConfig` through `set-config`. The read-modify-write happens server-side under `ConfigStore.withConfigLock`, so the CLI never round-trips secrets through `SettingsSecrets` or clobbers a concurrent edit.

- **`CrowCLILib` gains a `CrowCore` dependency** so `--event` parses straight into `NotificationEvent` (an `ExpressibleByArgument` conformance gives the ten names in `--help`, in the rejection message, and in shell completion) and sounds validate against `builtInSounds`. No fourth mirror of the event list — `scripts/check-notification-events.sh` already guards three. `CrowCore` declares zero dependencies, so no cycle, and it's already in the PR CI lane.
- **Booleans are `@Flag(inversion: .prefixedNo, exclusivity: .exclusive)` with `Bool?`** — an omitted flag leaves the stored value alone, and `--x --no-x` is rejected rather than silently resolved to the last one. There was no existing precedent: every CLI boolean today is a one-way `@Flag` that structurally can't express "set to false".
- **`eventSettings` is never encoded wholesale.** `NotificationEvent` isn't `CodingKeyRepresentable`, so `JSONEncoder` writes it as a flat alternating array. Responses build an `events` object keyed by raw value by hand; the on-disk shape is untouched, and both JS readers already handle it.
- **Per-event writes read through `config(for:)`**, so they land even when the event is absent from `config.json` — the common case, since most configs predate the automation events. A `eventSettings[event]?.x = v` subscript write would be a silent no-op there. Only the named event is materialized, preserving the forward-compat contract that absence means "follow the current default".
- **`notifications-*` stay un-gated in `localOnlyDenial`**, matching the `job-*` decision (CROW-665): no secrets, and the same subtree is already remotely editable via un-gated `set-config`. A test now enforces that, rather than a comment merely describing it.

## Two defects fixed along the way

Both surfaced while grounding the change and were agreed in advance:

**1. Fail-open config write (also affects `crow job *`).** `ConfigStore.loadConfig` returns `nil` both for "file absent" and "file present but undecodable", and the daemon's `mutateJobs` did `?? AppConfig()`. So a one-field edit against a malformed `config.json` replaced every workspace, job, credential, and the web-password hash with defaults — and returned `"saved": true`. Renamed to `mutateAppConfig` and made fail-closed. **This changes existing `crow job *` behavior**, from a silent wipe to an error that leaves the file untouched.

**2. Partial `notifications` blocks took the whole config down.** `NotificationSettings` and `EventNotificationConfig` used fully synthesized `Codable`, so `{"notifications": {"globalMute": true}}` threw `keyNotFound`; `AppConfig.init(from:)` propagated it and `loadConfig` swallowed it, silently degrading the *entire* config to defaults. Both now decode tolerantly, mirroring `TerminalSettings` / `AutoRespondSettings`. This is what makes the new `docs/configuration.md` block safe to hand-edit.

**Plus one web-client line:** `app.js`'s `onServerNotify` received the daemon's `configReloaded` push but ignored it, so a CLI write wouldn't reach an open tab. It now refreshes `uiConfig` before emitting — a mute that just landed silences its own announcement.

## Tests

| Package | Coverage |
|---|---|
| CrowCore | partial/empty-object decode tolerance, `canonicalSoundName` accept/reject/case table |
| CrowCLI | parse + validate + subcommand routing, every rejection path, error-message content |
| CrowPersistence | config round-trip, write to an event absent from the file, partial-block decode |
| CrowEngine | `decodeEvent` / `decodeSoundName`, `settingsJSON` shape, `--event` filter keeps globals, custom sound echoed unvalidated |
| CrowDaemon | router-level integration for both verbs, both fail-closed guards (`notifications-set` **and** `job-add`), un-gated assertion |

`make test` passes. One pre-existing failure in `CrowTerminal`'s `retryReadinessEmitsTimedOutWhenSentinelMissing` reproduces on a clean tree with this branch stashed — environment-dependent tmux test, unrelated.

Verified end-to-end against a scratch daemon and by exercising the built binary's validation paths and `--help`.

## Notes

- **The #807 parity harness / `KNOWN_UI_ONLY` list doesn't exist on this branch yet** (grepped: zero hits), so that checklist step is skipped rather than built here.
- Flag naming mirrors the model's own plural/singular split — `--system-notifications-enabled` (global) vs `--event-system-notification-enabled` (per-event). Called out in the docs; it will trip a reader otherwise.
- `--event-sound-name` accepts only the 14 built-ins, matching the Settings picker. Reads never validate, so a config already storing a custom sound path keeps it and echoes back losslessly.
- **Known, unchanged from #604:** an open Settings modal holds a detached config copy and saves it wholesale, so it will clobber a CLI edit made while it's open. Last-writer-wins; not addressed here.
- Separately worth an issue: `SetupCommand.swift` writes `SetupWorkspace` without an `id` while `WorkspaceInfo.init(from:)` requires one, so a `crow setup`-generated config with any workspace fails `AppConfig` decode outright. Pre-existing and unrelated.

https://claude.ai/code/session_01CPv6TMPxeG8xrgXVKmrGEo


dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Fail-closed `mutateAppConfig` closes a real credential/config wipe: undecodable `config.json` no longer falls through to `?? AppConfig()` and reports `"saved": true` (`RPCHandlers.swift` ~1441–1460). Covered for both `notifications-set` and `job-add`.
- Write-side sound/event validation is authoritative in the daemon (`NotificationRPC.decodeSoundName` / `decodeEvent`); CLI validation is defense-in-depth against the same predicates.
- Mutations stay under `ConfigStore.withConfigLock`; no secrets move through the new RPCs.
- Un-gated remote access for `notifications-*` matches the existing `set-config` / `job-*` surface and is pinned by `notificationRPCsAreAllowedRemotely`.

**Concerns:**
- None beyond the pre-existing authenticated remote ability to edit the same non-secret settings (documented and tested).

### Code Quality
- **Yellow — `docs/configuration.md`:** Claims Crow “reads either shape” for `eventSettings` (JSON object *or* alternating array). `NotificationEvent` is not `CodingKeyRepresentable`, so Swift’s `Dictionary` decode only accepts the alternating array. An object like `{"checksFailing": {…}}` fails with `Expected to decode Array<Any> but found a dictionary`, which propagates through `NotificationSettings` → `AppConfig` and makes `loadConfig` return nil. Hand-editors following this new docs block can brick config reads/writes (writes now correctly refuse). Fix: drop the “either shape” claim, or document array-only and show only that form.
- Green — Consider exposing the config path (or an `configFileExists`/`isDecodable`) from `ConfigStore` instead of duplicating `{devRoot}/.claude/config.json` in `RPCHandlers` for the fail-closed guard, so the security check can’t drift from the loader.
- Green — Partial `EventNotificationConfig` decode defaults missing `soundName` to `"Glass"` rather than the event’s `defaultSound` (unavoidable without the event key in-scope); worth a one-line note in docs for hand-editors.
- Green — `onServerNotify` is now `async` but still fire-and-forget from `ws.onmessage`; fine today because `loadUIConfig` errors are caught.

### Tests / Static Analysis
- CrowCore / CrowCLI / CrowEngine / CrowPersistence / CrowDaemon notification tests exercised locally: all filtered suites passed.
- `scripts/check-notification-events.sh`: catalogs agree (10 events).
- CrowCore package build succeeded.

### Summary Table

---

dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Fail-closed `mutateAppConfig` (`RPCHandlers.swift`) closes a real wipe risk: undecodable `config.json` no longer silently resets workspaces, jobs, credentials, and the web-password hash via `?? AppConfig()`.
- Write-side sound validation is shared (`NotificationSettings.canonicalSoundName`) and rejects custom paths; CLI and daemon can't drift.
- `notifications-*` carry no secrets; leaving them un-gated matches existing `set-config` / `job-*` remote surface, and `notificationRPCsAreAllowedRemotely` pins that decision.
- Mutations run under `ConfigStore.withConfigLock`; `config_readable` distinguishes absent vs undecodable on get.

**Concerns:**
- None that rise above Green. Error text on refuse-to-overwrite includes the absolute config path (usable by an authenticated remote `/rpc` peer) — acceptable for operability, minor disclosure.

### Code Quality
- **Yellow — absent `eventSettings` materializes all events, then any save freezes them.** `NotificationSettings.init(from:)` (`NotificationSettings.swift` ~61–74) passes `nil` eventSettings into the designated init, which populates every `NotificationEvent`. The same happens for a missing `notifications` key / fresh `AppConfig()` in `mutateAppConfig`. A globals-only `crow notifications set --global-mute` against a typical config (no notifications block, or hand-edited `{"notifications":{"globalMute":true}}`) therefore writes all ten events to disk. That contradicts this PR's own sparse-write contract (`notifications-set` comments, `setGlobalsDoesNotTouchEventSettings`, and `docs/configuration.md`: “Any event you omit follows the current defaults”). Prefer decode absence → `[:]`, and/or seed absent-config mutations with `NotificationSettings(eventSettings: [:])`; add a test that a first globals-only set against no/`partial` notifications block leaves `eventSettings` empty (or only the events explicitly written).
- **Yellow — partial event-entry `soundName` defaults to `"Glass"`, not the event's `defaultSound`.** `EventNotificationConfig.init(from:)` (`NotificationSettings.swift` ~136) uses `?? "Glass"`. Hand-editing e.g. `checksFailing` / `autoRebaseConflicts` with only `{"enabled":false}` silently changes Sosumi/Basso → Glass. Docs (`configuration.md` ~262) claim omitted keys fall back to “its default” — that is false for non-Glass events. Either document the Glass fallback explicitly, or decode without inventing a sound (and let `config(for:)` supply `event.defaultSound` when the entry is incomplete — may need a richer model).
- **Green —** `onServerNotify` is async and invoked without `await` from the WS message handler; `loadUIConfig` still runs, but concurrent notifies can interleave. Fine for now; worth a single-flight/queue if reload races show up.
- **Green —** Known Settings-modal last-writer-wins clobber (called out in the PR) remains; tracking as a follow-up is enough.

### Tests
CrowCore / CrowEngine / CrowCLI / CrowDaemon `Notification*` suites passed locally. No Red test failures observed for this change set.

### Summary Table

---

dhilgaertner:
Thanks — both reviews were right on every Yellow, including one I'd convinced myself was already covered. Fixed in 8185551.

## Yellow — absent `eventSettings` materialized all ten, then any save froze them

Confirmed before touching anything, with a throwaway router test against a config carrying no `notifications` block:

```
ZZ eventSettings count after globals-only set: 10
```

So `crow notifications set --global-mute` really did write all ten entries, freezing today's `defaultSound` table into the file — contradicting the handler comment, the docs, and my own `setGlobalsDoesNotTouchEventSettings`.

Worth naming why that test didn't catch it: it pre-seeded `config.notifications.eventSettings = [:]` and saved *that* first, so it only ever exercised a shape no real config holds. It was asserting against a fixture I'd built to pass.

Root cause was in the designated init, not the decode path — a nil `eventSettings` populated all ten, so *any* load of a config lacking the block produced a full table that the next save wrote back. Absence now stays absence; `config(for:)` already resolved a missing event to its current default on read, which was the intended contract all along.

Regression coverage is at the router level where it actually bit, parameterized over the three realistic first-write shapes (`{"remoteControlEnabled": true}`, a partial `{"notifications": {...}}`, and `{}`), plus no-config-file-at-all, plus a single-event write asserting exactly one entry lands. Model-level round-trip and default-init tests updated to assert the sparse contract rather than the old full table.

## Yellow — docs claimed "reads either shape"

Correct, and the consequence is worse than a doc nit: an object-shaped `eventSettings` fails `AppConfig` decode entirely, so `loadConfig` returns nil and Crow silently runs on defaults for *everything* — workspaces, jobs, credentials. My sentence was true of the JS readers (`settings.js` handles both) and I carried it over to a section about the file Swift parses first. Now documented as array-only with the failure and its blast radius spelled out.

## Yellow — `soundName` falls back to Glass, not the event's default

Also right, and the old docs line ("falls back to its default") actively misled. I went with the documentation option rather than the richer model, since removing the invented sound means `EventNotificationConfig` can't decode standalone without knowing its key — that's a real model change I'd rather not smuggle into a parity PR. Documented in three places: the model (with the reason it's structurally unavoidable), `configuration.md` naming every affected event (Sosumi/Basso/Funk/Hero/Bottle/Tink), and the CLI reference. Pinned by a test that asserts the fallback is Glass and explicitly *not* `checksFailing.defaultSound`, so the docs can't drift from behavior.

Also noted that `crow notifications set` is immune — it seeds each write from `config(for:)`, so a toggle change never silently rewrites the sound. Happy to revisit the model change as a follow-up if you'd rather hand-edits be safe by construction.

## Green — config path duplicated in `RPCHandlers`

Taken. Now `ConfigStore.configURL(devRoot:)` / `.configExists(devRoot:)`, used by both the fail-closed guard and the `config_readable` probe, with a test pinning them to where `saveConfig` actually writes and one covering the distinction the guard rests on (malformed → `loadConfig` nil but `configExists` true).

## Greens not changed

- **`onServerNotify` async, fire-and-forget from `ws.onmessage`.** Agreed it's fine today, and I'd rather not add single-flight speculatively. Concretely: the only awaited work is `loadUIConfig`, whose failure is caught, and interleaved reloads converge on the same server state — the loser just repaints twice. A queue earns its keep once something order-dependent hangs off it; happy to file it if you'd prefer it tracked.
- **Settings-modal last-writer-wins.** Pre-existing and inherited from #604; left as the noted follow-up.
- **Error text includes the absolute config path.** Kept for operability — an authenticated remote peer that can call `notifications-set` can already read the same layout via `get-config`, so it isn't new disclosure.

`make test` passes; the only failure is the pre-existing `retryReadinessEmitsTimedOutWhenSentinelMissing` tmux test, which reproduces on a clean tree with this branch stashed.

dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Fail-closed `mutateConfig` (`RPCHandlers.swift`) still refuses to overwrite an undecodable `config.json`, closing the wipe path that used to replace workspaces/jobs/credentials via `?? AppConfig()`. Guarded for both `notifications-set` and `job-add`.
- Write-side sound/event validation is shared (`NotificationSettings.canonicalSoundName` / `NotificationRPC.decode*`); CLI and daemon can't drift. Custom paths are rejected on write.
- Mutations stay under `ConfigStore.withConfigLock`; `configURL` / `configExists` keep the fail-closed check aligned with the loader.
- `notifications-*` carry no secrets; un-gated remote access matches `set-config` / `job-*` and is pinned by `notificationRPCsAreAllowedRemotely`.

**Concerns:**
- None that rise above Green. Refuse-to-overwrite errors still include the absolute config path (operability vs minor disclosure to an authenticated remote peer — unchanged and acceptable).

### Code Quality
Prior Yellow findings from earlier review rounds are addressed in `8185551`:
- Sparse writes: absent `eventSettings` stays `[:]`; globals-only set no longer freezes all ten events (covered by realistic-shape tests).
- Docs correctly require the alternating-array `eventSettings` form and call out the Glass-not-`defaultSound` partial-decode fallback.

Remaining (non-blocking):
- **Green** — Wrong-type bool RPC params (`"true"` / `1`) are silently treated as absent via `boolValue`; matches existing `JSONValue` patterns. Strict type rejection would be nicer for non-CLI callers but isn't required.
- **Green** — `Validation.swift` still says CrowCLI doesn't depend on CrowCore (quick-action comment); stale now that CrowCLI links CrowCore for notifications.
- **Green** — Known Settings-modal last-writer-wins clobber (called out in the PR) remains a follow-up.

### Tests / Static Analysis
CrowCore / CrowEngine / CrowCLI / CrowPersistence / CrowDaemon notification (and ConfigStore) suites all passed locally. Architecture warning on the `.xctest` bundle is environmental; Swift Testing reported full passes.

### Summary Table

---

PR #878, opened 2026-07-27
Sidebar: stop duplicate review rows piling up (CROW-877)

### LINKED ISSUE #877: Sidebar duplicates: re-review piles up completed review rows + same PR in Review and Completed
The sidebar shows duplicate session rows: (a) the same PR in both a review section and **Completed**, and (b) repeated identical rows **within Completed**. Both trace to auto-review creating separate review sessions that are never reaped or deduped.

### Primary — re-review leaves stale *completed* review rows (drives the in-Completed duplicates)
When a PR head advances, kickoff returns `.reReview(staleSessionID:)` (`IssueTracker.swift:3066-3070`). The consumer at `CrowDaemon.swift:772-776` is *meant* to "tear down the stale round" (its own comment) but calls `appState.onCompleteSession?(staleID)` → `updateSessionStatus(id, to: .completed)` (`SessionService.swift:3337`) — it marks the stale review **completed instead of deleting it**. Each round leaves another identical `review-<repo>-<pr>` row with `status: completed`. A `.create` re-fire path (`IssueTracker.swift:3075`) adds more after `/refine` + re-request. N re-review rounds = N identical Completed rows.

### Why nothing catches it
- Dedup guard `existingReviewSession(forPRURL:)` searches only non-terminal review sessions — `reviewSessions` filters out completed/archived (`AppState.swift:518-520`), so completed clones are invisible to it.
- Cleanup is **disabled by default** (`AppConfig.swift:757`, `enabled: false`); even when on it only reaps rows older than `retentionHours`, so recent duplicates persist (`IssueTracker.swift:3222-3242`).
- No dedup by id or PR URL at the RPC (`RPCHandlers.swift:186-188`, straight 1:1 map) or frontend (`app.js:657`, render loop `:740-744`). Both duplicates render.

### Cross-section duplicate (same PR in Review + Completed)
Auto-review mints a **separate** session (`createReviewSession`, `SessionService.swift:2418`, name `review-<repo>-<pr>`). On merge, both the original work session and the review clone are completed (`IssueTracker.swift:3168-3172`, `:3199-3203`) → a work row + a review row. While review is still open: work session = **Completed** (`app.js:646`) + review clone = **Reviews** (`app.js:644`). Session `status` is a single scalar (`CrowCore/.../Enums.swift:13-18`), so this is genuinely two rows for one PR, not one row double-matched.

### Latent grouping bug
The sidebar renders each section with an independent `sessions.filter()` (`app.js:641-647` GROUPS, `:740-744` loop) — not first-match. A `kind === 'review'` session with `status === 'inReview'` matches **both** "Reviews" (`:644`) and "In Review" (`:645`) → the same row in two sections.

## Fix directions

1. **Primary — delete the stale review session on re-review, don't complete it.** At `CrowDaemon.swift:772-773`, replace `onCompleteSession?(staleID)` with an actual teardown/delete of `staleID` (matching the existing "tear down the stale round" comment). Eliminates the identical-name pile-up in Completed.
2. **Sidebar grouping first-match.** Change the render loop (`app.js:740-744`) so each session is assigned to the **first** matching group only — kills the Reviews/In-Review one-row-two-sections duplicate.
3. **Defense-in-depth dedup + reconcile.** Dedup the Completed render by session id and by PR URL (`app.js:740-744` / `RPCHandlers.swift:186`), and collapse a PR's work row + review row into one entry (review row wins while review is open; the work row shows once the review session closes). Prevents future pile-ups regardless of the upstream teardown.
4. **Aggravating default (contributing, not core):** `cleanup.enabled` defaults to `false` (`AppConfig.swift:757`), so completed rows never reap.

**Decision to make:** for a merged PR, should the completed work row and completed review row collapse into a single entry, or both remain? (Recommend: collapse to one entry per PR.)

dgershman:
Closes #877

## Problem
The sidebar showed duplicate session rows for one PR: (a) the same PR in **both** a review section and **Completed**, and (b) repeated identical rows **within Completed**. Both trace to auto-review completing a separate review session per re-review round, which the sidebar then rendered without any dedup.

## What this PR does
The fix is **entirely in the web sidebar** (`app.js` `groupSessions`); the daemon behavior is **unchanged** from `main`.

### 1. First-match grouping
`renderSidebar` assigned each session to **every** matching `GROUP` via independent `sessions.filter()` calls, so a `kind:'review'` + `status:'inReview'` session showed in both **Reviews** and **In Review**. Each session now lands in the **first** group it matches only. This fixes the genuine one-session-in-two-sections half of symptom (a).

### 2. Within-section, clone-only collapse
After grouping, `collapsePRDuplicates` folds duplicate rows for one PR **within a single section**:
- **Terminal-only** — only `completed`/`archived` rows collapse, so a live session is never hidden.
- **Clone pairs only** — a pair folds only when at least one side is a review **clone** (a merged PR's work row + its completed review clone, or a pile-up of completed clones). Two independent work/job sessions that share a PR are never folded.
- **Never across sections** — collapsing across sections is what hid live work rows in earlier review rounds; a live review in *Reviews* and a completed work row in *Completed* deliberately stay as two rows.

This collapses report (b)'s completed-clone pile-up into one row, and merges a merged PR's work+review rows into one — including pre-existing data.

### 3. Collapsed rows stay deletable
`groupSessions` returns `allIds` per section (survivors **+** folded-away ids), and a section's **select-all** uses it, so bulk-delete still reaches a hidden collapsed clone in one pass.

## Daemon: unchanged (deliberate)
The re-review kickoff still **completes** the stale round (`onCompleteSession`), exactly as on `main`. An earlier revision *deleted* it instead, but deletion writes no analytics snapshot and then drops the raw telemetry rows — silently erasing that round's tokens/cost from the scorecard. Completion preserves the snapshot and still makes the stale row invisible to `createReviewSession`'s dedup guard; the sidebar collapse (above) owns the visible-duplicate half. So the daemon diff here is comment-only.

## Not changed
`cleanup.enabled` still defaults to `false` — the ticket flags it as *aggravating, not core*. Completed review sessions (and their clones) persist until cleanup is enabled or they're deleted; they're now collapsed in the sidebar but remain bulk-deletable via a section's select-all.

## Testing
- New jsdom regression test `sidebar-groups.test.js` — **20 checks, all green**: first-match, dedup-by-id, terminal-only + clone-only collapse, live-session-never-hidden (incl. two open reviews on one PR, and work+work sharing a PR), no-group survivor, merged-PR collapse, clone pile-up, and select-all reaching a collapsed clone.
- `package.json` `test` now runs every suite and aggregates the exit code (previously `&&`-chained, so it short-circuited at `row.test.js`). All suites execute; `row.test.js`'s failures are **pre-existing** (identical on `main` — currently 10) and unrelated to this change; no CI workflow runs `npm test` yet.
- `swift build` (CrowDaemon) ✅ · `swift test --filter BoardForwarderTests` ✅ · CLI/RPC parity gate ✅.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
@@ -795,13 +795,31 @@ public enum CrowDaemon {
                 // completed/archived) yet still renders, so each re-review round
                 // left another identical row piling up in Completed (CROW-877).
                 // Deleting also drops its `.pr` link, so the create below can't
-                // reuse the stale session via that same guard. Await the delete
-                // first so it's gone before `createReviewSession` runs its check.
+                // reuse the stale session via that same guard.
+                //
+                // This always tears down a *live* round-1 review clone (kickoff
+                // only returns `.reReview` while that session is still open):
+                // `rm -rf` on its PR clone, `TerminalRouter.destroy` on its
+                // terminal, and its raw telemetry rows are dropped. That's
+                // intended — round 1 is reviewing a now-dead SHA — but it's a
+                // harder teardown than the old status flip.
                 let url = request.url
                 let staleReviewID: UUID?
                 if case let .reReview(id) = action { staleReviewID = id } else { staleReviewID = nil }
                 Task {
-                    if let staleReviewID { await sessionService.deleteSession(id: staleReviewID) }
+                    if let staleReviewID {
+                        await sessionService.deleteSession(id: staleReviewID)
+                        // `deleteSession` no-ops on a concurrent delete (its
+                        // `isDeletingSession` guard) and bails on a disk-cleanup
+                        // failure — both leave the stale session AND its `.pr`
+                        // link intact, which would make `createReviewSession`'s
+                        // dedup guard return the stale session and swallow the new
+                        // round. Fall back to completing it (the old path's effect)
+                        // so it's at least invisible to that guard (CROW-877 review).
+                        if appState.sessions.contains(where: { $0.id == staleReviewID }) {
+                            sessionService.completeSession(id: staleReviewID)
+                        }
+                    }
                     _ = await reviewSerializer.enqueue { await sessionService.createReviewSession(prURL: url, selectAfterCreate: false) }
                 }
             }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
@@ -787,41 +787,27 @@ public enum CrowDaemon {
                 // the session's link/reviewSessionID land).
                 let fingerprint = "\(request.id)\n\(request.headRefOid ?? "")"
                 guard autoReviewed.insert(fingerprint).inserted else { continue }
-                // B-fallback: tear down the stale round's review session before
+                // B-fallback: retire the stale round's review session before
                 // enqueuing so `reviewSessions` doesn't double up for this PR.
-                // DELETE it — not `onCompleteSession`, which only flips it to
-                // `.completed`. A completed `review-<repo>-<pr>` row is invisible
-                // to `existingReviewSession(forPRURL:)` (that dedup guard excludes
-                // completed/archived) yet still renders, so each re-review round
-                // left another identical row piling up in Completed (CROW-877).
-                // Deleting also drops its `.pr` link, so the create below can't
-                // reuse the stale session via that same guard.
-                //
-                // This always tears down a *live* round-1 review clone (kickoff
-                // only returns `.reReview` while that session is still open):
-                // `rm -rf` on its PR clone, `TerminalRouter.destroy` on its
-                // terminal, and its raw telemetry rows are dropped. That's
-                // intended — round 1 is reviewing a now-dead SHA — but it's a
-                // harder teardown than the old status flip.
-                let url = request.url
-                let staleReviewID: UUID?
-                if case let .reReview(id) = action { staleReviewID = id } else { staleReviewID = nil }
-                Task {
-                    if let staleReviewID {
-                        await sessionService.deleteSession(id: staleReviewID)
-                        // `deleteSession` no-ops on a concurrent delete (its
-                        // `isDeletingSession` guard) and bails on a disk-cleanup
-                        // failure — both leave the stale session AND its `.pr`
-                        // link intact, which would make `createReviewSession`'s
-                        // dedup guard return the stale session and swallow the new
-                        // round. Fall back to completing it (the old path's effect)
-                        // so it's at least invisible to that guard (CROW-877 review).
-                        if appState.sessions.contains(where: { $0.id == staleReviewID }) {
-                            sessionService.completeSession(id: staleReviewID)
-                        }
-                    }
-                    _ = await reviewSerializer.enqueue { await sessionService.createReviewSession(prURL: url, selectAfterCreate: false) }
+                // Complete it (not delete): completing writes its end-of-round
+                // `SessionAnalyticsSnapshot` and keeps its telemetry, whereas
+                // `deleteSession` transitions no status — so no snapshot is
+                // written — and then drops the raw rows, silently erasing round
+                // 1's tokens/cost from the scorecard on Crow's own auto-review
+                // flow (CROW-877 review). Completing also makes it invisible to
+                // `existingReviewSession(forPRURL:)` (that dedup guard excludes
+                // completed/archived), so the create below starts the new round
+                // instead of reusing the stale session. The identical completed
+                // `review-<repo>-<pr>` rows that used to pile up in Completed are
+                // now folded into one by the sidebar's within-section collapse
+                // (`groupSessions` in app.js), which also cleans up pre-existing
+                // pile-ups — so the visible-duplicate half no longer needs a
+                // destructive teardown here.
+                if case let .reReview(staleID) = action {
+                    appState.onCompleteSession?(staleID)
                 }
+                let url = request.url
+                Task { _ = await reviewSerializer.enqueue { await sessionService.createReviewSession(prURL: url, selectAfterCreate: false) } }
             }
         }
     }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
@@ -787,13 +787,23 @@ public enum CrowDaemon {
                 // the session's link/reviewSessionID land).
                 let fingerprint = "\(request.id)\n\(request.headRefOid ?? "")"
                 guard autoReviewed.insert(fingerprint).inserted else { continue }
-                // B-fallback: tear down the stale round-1 session before enqueuing
-                // so `reviewSessions` doesn't double up for this PR.
-                if case let .reReview(staleID) = action {
-                    appState.onCompleteSession?(staleID)
-                }
+                // B-fallback: tear down the stale round's review session before
+                // enqueuing so `reviewSessions` doesn't double up for this PR.
+                // DELETE it — not `onCompleteSession`, which only flips it to
+                // `.completed`. A completed `review-<repo>-<pr>` row is invisible
+                // to `existingReviewSession(forPRURL:)` (that dedup guard excludes
+                // completed/archived) yet still renders, so each re-review round
+                // left another identical row piling up in Completed (CROW-877).
+                // Deleting also drops its `.pr` link, so the create below can't
+                // reuse the stale session via that same guard. Await the delete
+                // first so it's gone before `createReviewSession` runs its check.
                 let url = request.url
-                Task { _ = await reviewSerializer.enqueue { await sessionService.createReviewSession(prURL: url, selectAfterCreate: false) } }
+                let staleReviewID: UUID?
+                if case let .reReview(id) = action { staleReviewID = id } else { staleReviewID = nil }
+                Task {
+                    if let staleReviewID { await sessionService.deleteSession(id: staleReviewID) }
+                    _ = await reviewSerializer.enqueue { await sessionService.createReviewSession(prURL: url, selectAfterCreate: false) }
+                }
             }
         }
     }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -667,29 +667,42 @@ const GROUPS = [
 //   2. first-match grouping — each session lands in the FIRST group it matches,
 //      not every one, so a `kind:'review'` + `status:'inReview'` session can't
 //      appear in both "Reviews" and "In Review";
-//   3. collapse same-PR duplicates WITHIN a section — a merged PR (its completed
-//      work row + completed review clone both land in "Completed") and any
-//      pile-up of identical completed `review-<repo>-<pr>` clones become one row.
+//   3. collapse same-PR duplicates WITHIN a section, and only among
+//      completed/archived rows — a merged PR (its completed work row + completed
+//      review clone both land in "Completed") and any pile-up of identical
+//      completed `review-<repo>-<pr>` clones become one row.
 // Collapsing AFTER assignment, never across sections, is deliberate: a live work
 // row in "Active"/"In Review" is never hidden by an open review clone in
 // "Reviews", and a collapse survivor can never be a row that matches no group —
-// the two failure modes of a pre-assignment collapse (CROW-877 review).
+// the two failure modes of a pre-assignment collapse. Restricting collapse to
+// TERMINAL rows is equally deliberate: two live sessions can legitimately share
+// a PR within one section (e.g. a manual "Start Review" racing the auto-review
+// clone lands two open reviews in "Reviews"), and hiding either would strand a
+// running agent with no way to select or delete it (CROW-877 review).
 // Managers carry no PR link and match no GROUP, so they drop out here and are
 // rendered by renderSidebar's dedicated managers pass.
 function prURLOf(s) {
+  // The persisted `links` array is authoritative for the terminal rows this
+  // collapse touches — a completed review clone always carries its `.pr` link
+  // there. (Live sessions can hold a PR link only in the app's in-memory
+  // `liveFor(id).pr_link`, but those are non-terminal and never collapsed.)
   const link = (s.links || []).find((l) => l.type === 'pr');
   return link ? link.url : null;
 }
-// Collapse rows that describe the same PR, keeping one. A non-review (work) row
-// represents the PR over a review clone (a merged PR shows its work row, not the
-// leftover completed clone); otherwise the first seen stays. Rows without a PR
-// link are never collapsed. Only ever called on rows already in the same section.
+function isTerminal(s) {
+  return s.status === 'completed' || s.status === 'archived';
+}
+// Collapse rows that describe the same PR, keeping one. Only completed/archived
+// rows are ever collapsed, so a live session is never hidden. A non-review (work)
+// row represents the PR over a review clone (a merged PR shows its work row, not
+// the leftover completed clone); otherwise the first seen stays. Only ever called
+// on rows already in the same section.
 function collapsePRDuplicates(rows) {
   const byPR = new Map();     // PR URL -> index into `out`
   const out = [];
   for (const s of rows) {
     const url = prURLOf(s);
-    if (!url) { out.push(s); continue; }
+    if (!url || !isTerminal(s)) { out.push(s); continue; }
     const idx = byPR.get(url);
     if (idx === undefined) { byPR.set(url, out.length); out.push(s); continue; }
     if (out[idx].kind === 'review' && s.kind !== 'review') out[idx] = s;   // work wins
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -681,45 +681,58 @@ const GROUPS = [
 // running agent with no way to select or delete it (CROW-877 review).
 // Managers carry no PR link and match no GROUP, so they drop out here and are
 // rendered by renderSidebar's dedicated managers pass.
-function prURLOf(s) {
-  // The persisted `links` array is authoritative for the terminal rows this
-  // collapse touches — a completed review clone always carries its `.pr` link
-  // there. (Live sessions can hold a PR link only in the app's in-memory
-  // `liveFor(id).pr_link`, but those are non-terminal and never collapsed.)
-  const link = (s.links || []).find((l) => l.type === 'pr');
-  return link ? link.url : null;
-}
 function isTerminal(s) {
   return s.status === 'completed' || s.status === 'archived';
 }
-// Collapse rows that describe the same PR, keeping one. Only completed/archived
-// rows are ever collapsed, so a live session is never hidden. A non-review (work)
-// row represents the PR over a review clone (a merged PR shows its work row, not
-// the leftover completed clone); otherwise the first seen stays. Only ever called
-// on rows already in the same section.
+// Collapse a PR's duplicate rows within one section, keeping one survivor.
+// Returns { rows, collapsedIds }: the rows to render, plus the ids folded away
+// so the section's select-all can still reach them (they'd otherwise be
+// undeletable from the sidebar — CROW-877 review). Only terminal rows collapse,
+// so a live session is never hidden. A pair collapses only when at least one
+// side is a review CLONE — a merged PR's work row + its completed review clone,
+// or a pile-up of completed clones. Two independent work/job sessions that
+// happen to share a PR (a follow-up session, or a manual `add-link`) are NEVER
+// folded; both render. A work row represents the PR over a review clone.
+// Uses the shared `prUrlForSession` (hoisted; defined below) so "does this row
+// have a PR" means one thing across the file.
 function collapsePRDuplicates(rows) {
-  const byPR = new Map();     // PR URL -> index into `out`
+  const byPR = new Map();       // PR URL -> index into `out`
   const out = [];
+  const collapsedIds = [];
   for (const s of rows) {
-    const url = prURLOf(s);
+    const url = prUrlForSession(s);
     if (!url || !isTerminal(s)) { out.push(s); continue; }
     const idx = byPR.get(url);
     if (idx === undefined) { byPR.set(url, out.length); out.push(s); continue; }
-    if (out[idx].kind === 'review' && s.kind !== 'review') out[idx] = s;   // work wins
+    const prior = out[idx];
+    if (prior.kind !== 'review' && s.kind !== 'review') { out.push(s); continue; }  // two non-clones: keep both
+    if (prior.kind === 'review' && s.kind !== 'review') {                            // work supersedes a clone
+      collapsedIds.push(prior.id);
+      out[idx] = s;
+    } else {                                                                          // clone folds into the survivor
+      collapsedIds.push(s.id);
+    }
   }
-  return out;
+  return { rows: out, collapsedIds };
 }
 function groupSessions(list) {
   const seenIds = new Set();
-  const out = GROUPS.map((g) => ({ title: g.title, rows: [] }));
+  const buckets = GROUPS.map((g) => ({ title: g.title, assigned: [] }));
   for (const s of list) {
     if (seenIds.has(s.id)) continue;
     seenIds.add(s.id);
     const gi = GROUPS.findIndex((g) => g.match(s));
-    if (gi >= 0) out[gi].rows.push(s);
+    if (gi >= 0) buckets[gi].assigned.push(s);
   }
-  for (const g of out) g.rows = collapsePRDuplicates(g.rows);
-  return out.filter((g) => g.rows.length);
+  const out = [];
+  for (const b of buckets) {
+    if (!b.assigned.length) continue;
+    const { rows, collapsedIds } = collapsePRDuplicates(b.assigned);
+    // `allIds` = rendered survivors + folded-away ids, so a section's "select
+    // all" (and thus bulk-delete) still reaches the hidden collapsed rows.
+    out.push({ title: b.title, rows, allIds: rows.map((r) => r.id).concat(collapsedIds) });
+  }
+  return out;
 }
 
 // ---------------------------------------------------------------------------
@@ -813,8 +826,8 @@ function renderSidebar() {
   for (const m of managers.slice(1)) root.appendChild(sessionRow(m));
 
   let shown = 0;
-  for (const { title, rows } of groupSessions(sessions)) {
-    root.appendChild(selectionMode ? sectionHeader(title, rows) : el('div', 'divider', title));
+  for (const { title, rows, allIds } of groupSessions(sessions)) {
+    root.appendChild(selectionMode ? sectionHeader(title, allIds) : el('div', 'divider', title));
     for (const s of rows) { root.appendChild(sessionRow(s)); shown++; }
   }
   if ((sessionsLoaded || sidebarCacheHit) && !shown && !managers.length) {
@@ -850,11 +863,11 @@ function toggleSelect(id) {
 }
 
 // Section divider with a per-section select-all/clear toggle (mirrors the
-// desktop section header checklist button).
-function sectionHeader(title, rows) {
+// desktop section header checklist button). `ids` is the section's full id set
+// (survivors + PR-collapsed rows), so "select all" reaches the hidden rows too.
+function sectionHeader(title, ids) {
   const head = el('div', 'divider divider-sel');
   head.appendChild(el('span', 'divider-label', title));
-  const ids = rows.map((r) => r.id);
   const allSel = ids.length && ids.every((id) => selectedSessionIDs.has(id));
   const btn = el('button', 'divider-selall', allSel ? 'Clear' : 'All');
   btn.title = allSel ? 'Deselect all in section' : 'Select all in section';
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -661,55 +661,51 @@ const GROUPS = [
   { title: 'Completed', match: (s) => (s.status === 'completed' || s.status === 'archived') && s.kind !== 'manager' },
 ];
 
-// Assign sessions to sidebar sections, collapsing the duplicates that used to
-// render as separate rows for the same PR (CROW-877):
+// Assign sessions to sidebar sections without the duplicate rows that used to
+// render for one PR (CROW-877):
 //   1. dedup by session id — a payload that repeats an id renders once;
-//   2. collapse by PR URL — a PR's work row + auto-review clone become ONE
-//      entry (the open review row wins while the review is in progress; once
-//      the review session closes, the work row represents the PR). This kills
-//      both the same-PR-in-Reviews-and-Completed cross-section duplicate and
-//      any pile-up of identical completed `review-<repo>-<pr>` clones;
-//   3. first-match grouping — each surviving session lands in the FIRST group
-//      it matches, not every one, so a `kind:'review'` + `status:'inReview'`
-//      session can't appear in both "Reviews" and "In Review".
+//   2. first-match grouping — each session lands in the FIRST group it matches,
+//      not every one, so a `kind:'review'` + `status:'inReview'` session can't
+//      appear in both "Reviews" and "In Review";
+//   3. collapse same-PR duplicates WITHIN a section — a merged PR (its completed
+//      work row + completed review clone both land in "Completed") and any
+//      pile-up of identical completed `review-<repo>-<pr>` clones become one row.
+// Collapsing AFTER assignment, never across sections, is deliberate: a live work
+// row in "Active"/"In Review" is never hidden by an open review clone in
+// "Reviews", and a collapse survivor can never be a row that matches no group —
+// the two failure modes of a pre-assignment collapse (CROW-877 review).
 // Managers carry no PR link and match no GROUP, so they drop out here and are
 // rendered by renderSidebar's dedicated managers pass.
 function prURLOf(s) {
   const link = (s.links || []).find((l) => l.type === 'pr');
   return link ? link.url : null;
 }
-function isOpenReview(s) {
-  return s.kind === 'review' && s.status !== 'completed' && s.status !== 'archived';
-}
-// Between two sessions describing the same PR, pick the one to keep.
-function pickPRRow(a, b) {
-  const ao = isOpenReview(a), bo = isOpenReview(b);
-  if (ao !== bo) return ao ? a : b;           // an open review wins outright
-  // Neither is an open review (both closed, or both non-review work rows):
-  // prefer the non-review (work) row so a merged PR shows its work session
-  // rather than a leftover completed review clone.
-  const aWork = a.kind !== 'review', bWork = b.kind !== 'review';
-  if (aWork !== bWork) return aWork ? a : b;
-  return a;                                    // stable: keep the first seen
+// Collapse rows that describe the same PR, keeping one. A non-review (work) row
+// represents the PR over a review clone (a merged PR shows its work row, not the
+// leftover completed clone); otherwise the first seen stays. Rows without a PR
+// link are never collapsed. Only ever called on rows already in the same section.
+function collapsePRDuplicates(rows) {
+  const byPR = new Map();     // PR URL -> index into `out`
+  const out = [];
+  for (const s of rows) {
+    const url = prURLOf(s);
+    if (!url) { out.push(s); continue; }
+    const idx = byPR.get(url);
+    if (idx === undefined) { byPR.set(url, out.length); out.push(s); continue; }
+    if (out[idx].kind === 'review' && s.kind !== 'review') out[idx] = s;   // work wins
+  }
+  return out;
 }
 function groupSessions(list) {
   const seenIds = new Set();
-  const byPR = new Map();     // PR URL -> index into `collapsed`
-  const collapsed = [];
+  const out = GROUPS.map((g) => ({ title: g.title, rows: [] }));
   for (const s of list) {
     if (seenIds.has(s.id)) continue;
     seenIds.add(s.id);
-    const url = prURLOf(s);
-    if (!url) { collapsed.push(s); continue; }
-    const idx = byPR.get(url);
-    if (idx === undefined) { byPR.set(url, collapsed.length); collapsed.push(s); }
-    else collapsed[idx] = pickPRRow(collapsed[idx], s);
-  }
-  const out = GROUPS.map((g) => ({ title: g.title, rows: [] }));
-  for (const s of collapsed) {
     const gi = GROUPS.findIndex((g) => g.match(s));
     if (gi >= 0) out[gi].rows.push(s);
   }
+  for (const g of out) g.rows = collapsePRDuplicates(g.rows);
   return out.filter((g) => g.rows.length);
 }
 
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -661,6 +661,58 @@ const GROUPS = [
   { title: 'Completed', match: (s) => (s.status === 'completed' || s.status === 'archived') && s.kind !== 'manager' },
 ];
 
+// Assign sessions to sidebar sections, collapsing the duplicates that used to
+// render as separate rows for the same PR (CROW-877):
+//   1. dedup by session id — a payload that repeats an id renders once;
+//   2. collapse by PR URL — a PR's work row + auto-review clone become ONE
+//      entry (the open review row wins while the review is in progress; once
+//      the review session closes, the work row represents the PR). This kills
+//      both the same-PR-in-Reviews-and-Completed cross-section duplicate and
+//      any pile-up of identical completed `review-<repo>-<pr>` clones;
+//   3. first-match grouping — each surviving session lands in the FIRST group
+//      it matches, not every one, so a `kind:'review'` + `status:'inReview'`
+//      session can't appear in both "Reviews" and "In Review".
+// Managers carry no PR link and match no GROUP, so they drop out here and are
+// rendered by renderSidebar's dedicated managers pass.
+function prURLOf(s) {
+  const link = (s.links || []).find((l) => l.type === 'pr');
+  return link ? link.url : null;
+}
+function isOpenReview(s) {
+  return s.kind === 'review' && s.status !== 'completed' && s.status !== 'archived';
+}
+// Between two sessions describing the same PR, pick the one to keep.
+function pickPRRow(a, b) {
+  const ao = isOpenReview(a), bo = isOpenReview(b);
+  if (ao !== bo) return ao ? a : b;           // an open review wins outright
+  // Neither is an open review (both closed, or both non-review work rows):
+  // prefer the non-review (work) row so a merged PR shows its work session
+  // rather than a leftover completed review clone.
+  const aWork = a.kind !== 'review', bWork = b.kind !== 'review';
+  if (aWork !== bWork) return aWork ? a : b;
+  return a;                                    // stable: keep the first seen
+}
+function groupSessions(list) {
+  const seenIds = new Set();
+  const byPR = new Map();     // PR URL -> index into `collapsed`
+  const collapsed = [];
+  for (const s of list) {
+    if (seenIds.has(s.id)) continue;
+    seenIds.add(s.id);
+    const url = prURLOf(s);
+    if (!url) { collapsed.push(s); continue; }
+    const idx = byPR.get(url);
+    if (idx === undefined) { byPR.set(url, collapsed.length); collapsed.push(s); }
+    else collapsed[idx] = pickPRRow(collapsed[idx], s);
+  }
+  const out = GROUPS.map((g) => ({ title: g.title, rows: [] }));
+  for (const s of collapsed) {
+    const gi = GROUPS.findIndex((g) => g.match(s));
+    if (gi >= 0) out[gi].rows.push(s);
+  }
+  return out.filter((g) => g.rows.length);
+}
+
 // ---------------------------------------------------------------------------
 // Sidebar
 // ---------------------------------------------------------------------------
@@ -752,10 +804,8 @@ function renderSidebar() {
   for (const m of managers.slice(1)) root.appendChild(sessionRow(m));
 
   let shown = 0;
-  for (const group of GROUPS) {
-    const rows = sessions.filter(group.match);
-    if (!rows.length) continue;
-    root.appendChild(selectionMode ? sectionHeader(group.title, rows) : el('div', 'divider', group.title));
+  for (const { title, rows } of groupSessions(sessions)) {
+    root.appendChild(selectionMode ? sectionHeader(title, rows) : el('div', 'divider', title));
     for (const s of rows) { root.appendChild(sessionRow(s)); shown++; }
   }
   if ((sessionsLoaded || sidebarCacheHit) && !shown && !managers.length) {
--- a/Packages/CrowDaemon/web-tests/package.json
+++ b/Packages/CrowDaemon/web-tests/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "description": "Headless jsdom regression tests for the web UI (Ticket Board; terminal touch- and wheel-scroll; terminal key handling; sidebar session rows). Drives the real app.js from Resources/web against mocks.",
   "scripts": {
-    "test": "node board.test.js && node touch-scroll.test.js && node wheel-scroll.test.js && node key-handler.test.js && node row.test.js"
+    "test": "node board.test.js && node touch-scroll.test.js && node wheel-scroll.test.js && node key-handler.test.js && node row.test.js && node sidebar-groups.test.js"
   },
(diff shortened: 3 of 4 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Verified locally: `swift build --package-path Packages/CrowDaemon` ✅ (only a pre-existing warning in `RPCWebSocketHandler.swift`), `node sidebar-groups.test.js` ✅ 11/11. `row.test.js`'s 11 failures do reproduce on `main` — unrelated, agreed.

The diagnosis is right and the first-match grouping fix is exactly correct. The **collapse-by-PR-URL** half is where this breaks: it collapses in *every* state, not just the merged/closed state the ticket asked about, and it picks the survivor without checking that the survivor renders anywhere.

### Critical Issues

**🔴 1. An open review session hides the author's live work session from the sidebar** — `app.js:682-691` (`pickPRRow`)

`isOpenReview` "wins outright" over *any* other row for the same PR, including an `active`/`inReview` **work** session with a running agent. Since `findPRLink` (`SessionService.swift:2168-2179`) attaches a `.pr` link to work sessions, and auto-review spawns a review clone for the same PR URL, this is the ordinary Crow flow, not an edge case.

Repro (ran the real `groupSessions` in the jsdom harness):
```js
[{id:'work-active', kind:'work',   status:'active', links:prLink(10)},
 {id:'rev-open',    kind:'review', status:'active', links:prLink(10)}]
→ [["Reviews", ["rev-open"]]]        // "Active" section is gone; work-active is unreachable
```
The developer's own session — its worktree, branch, uncommitted changes and agent terminal — drops out of the sidebar for the entire duration of the review round. The only remaining route back is the Ticket Board's "go" button (`app.js:3034-3036`), and only if the session has a linked issue. The ticket's decision was specifically *"for a **merged** PR, should the completed work row and completed review row collapse"* — gating the collapse on **both** rows being `completed`/`archived` satisfies that decision, still kills the completed-clone pile-up (test #4), and never hides a live session.

**🔴 2. Collapse can elect a survivor that matches no GROUP, so both rows vanish** — `app.js:692-711`

Collapse runs *before* grouping and `pickPRRow` has no notion of renderability, so if the winner matches no `GROUPS` entry the loser is discarded too and the PR disappears from the sidebar entirely. `paused` matches no group (`app.js:653-659`):
```js
[{id:'work-paused', kind:'work',   status:'paused',    links:prLink(11)},
 {id:'rev-done',    kind:'review', status:'completed', links:prLink(11)}]
→ []                                  // both rows gone; pre-PR, rev-done rendered under Completed
```
With no other sessions this also trips the `!shown` branch at `app.js:808` and prints "No sessions" while two sessions exist. Fix either way: skip candidates that match no group when picking (or collapse *within* a group after assignment).

### Should Fix

**🟡 3. A failed or already-in-flight delete silently swallows the re-review round** — `CrowDaemon.swift:780-786`

`deleteSession` bails without deleting in two cases: the `isDeletingSession[id] != true` guard (`SessionService.swift:1743`) and a disk-cleanup failure, which returns at `SessionService.swift:1768-1779` leaving the session — and its `.pr` link — intact. It returns `Void`, so the caller can't tell. In both cases the very next line runs `createReviewSession`, whose dedup guard (`SessionService.swift:2425-2429`) finds the surviving stale session and returns *it* instead of creating the new round. The `autoReviewed` fingerprint was already consumed at `CrowDaemon.swift:769`, so nothing retries until the head advances again.

This is a regression versus the old `onCompleteSession` path, which always made the stale row invisible to that guard and so always let the new round through. Cheap fix — verify and fall back:

```swift
if let staleReviewID {
    await sessionService.deleteSession(id: staleReviewID)
    // Delete can no-op (concurrent delete) or fail (disk cleanup) and leave the
    // stale `.pr` link in place, which would make createReviewSession's dedup
    // guard return the stale session instead of starting the new round.
    if appState.sessions.contains(where: { $0.id == staleReviewID }) {
        sessionService.completeSession(id: staleReviewID)
    }
}
```

### Consider

**🟢 4. `.reReview` now always destroys a *live* review session.** Because `existingReviewSession`/`reviewSessions` exclude completed rows (`AppState.swift:518-520`), `reviewKickoffAction` can only return `.reReview` while the round-1 session is still open — so the new `deleteSession` always tears down a running review: `rm -rf` on the clone, `TerminalRouter.destroy` on its terminal, and `telemetryDeleteProvider(id)` drops its raw telemetry rows (analytics snapshots survive; see the note at `SessionService.swift:1813-1820`). That's defensible — the round-1 agent is reviewing a dead SHA — but it's a real behavior change from a soft status flip, and it isn't called out in the PR description or a notification. Worth a line in the comment at minimum.

**🟢 5. Test coverage gaps.** `sidebar-groups.test.js` #2 deliberately uses a `completed` work row against the open review, which is precisely the case that sidesteps issue 1; an `active` work row there would fail. Nothing covers a survivor that matches no group. On the daemon side, nothing asserts delete-not-complete (`BoardForwarderTests` doesn't touch `onReviewRequestsRefreshed`).

### Security Review

**Strengths:**
- No new attack surface: `prURLOf` only string-compares `link.url`, no `innerHTML`, no eval, no new RPC.
- Keying the collapse map with `Map`/`Set` rather than object literals means an attacker-controlled URL like `__proto__` can't pollute a prototype — the right primitive.
- The daemon change adds no new input path; `staleSessionID` is a `UUID` from the tracker's own state, and `deleteSession`'s Manager guard (`SessionService.swift:1742`) still stands.

**Concerns:**
- None. The delete is more destructive than the previous complete, but it's scoped to the session's own review clone by `performDiskCleanup`'s `isReview` branch (`SessionService.swift:1842-1853`), which cannot touch a shared repo checkout.

### Code Quality

- ✅ Clean separation: `groupSessions` is pure and directly testable — good call, and the jsdom loader matches the existing `row.test.js` shape.
- ✅ First-match grouping (`app.js:705-709`) is the correct fix for the one-session-in-two-sections duplicate, and the daemon's complete→delete is the correct root-cause fix for the pile-up.
- ✅ Comments explain *why*, including the ordering requirement for the awaited delete.
- 🟢 `groupSessions` mixes three responsibilities; splitting collapse from assignment would also naturally fix issue 2.
- 🟢 The `package.json` `description` still lists the covered suites — worth adding "sidebar grouping" alongside "sidebar session rows".

### Summary Table

**Recommendation:** Request Changes — driven by [2 Red, 1 Yellow, 4 Green] findings.

Narrowing the collapse to closed-only rows resolves both Reds at once and still delivers everything the ticket asked for.

---

dgershman:
Thanks for the thorough review — both Reds were real. Pushed 738c36e3.

**🔴 1 (open review hid a live work session) + 🔴 2 (survivor matched no group → both rows vanish):** Took your "collapse *within* a group after assignment" option — it resolves both at once and cleanly splits the two responsibilities (green #Quality). `groupSessions` now: dedup-by-id → first-match assign → collapse same-PR duplicates **within each section only**. Consequences:
- A live work row in Active/In Review is never hidden by an open review clone in Reviews (they're in different sections, never cross-collapsed) — your `work-active` + `rev-open` repro now returns both rows.
- A survivor is always a row that already matched a group, so the `work-paused` + `rev-done` repro renders `rev-done` in Completed (no "No sessions").
- Merged PRs (completed work + completed review both land in Completed) and completed-clone pile-ups still collapse to one row, work preferred.

Tradeoff I want to flag explicitly: this **stops collapsing the while-review-open cross-section pair** (completed work in Completed + open review in Reviews stays as two rows). That's the price of never hiding a live session / never electing a no-group survivor, and it still satisfies the ticket's stated decision, which was scoped to **merged** PRs. If you'd rather also suppress the completed-work row while its review is open, that needs cross-section logic and I'd want a follow-up rather than reintroducing the pre-assignment collapse.

**🟡 3 (failed/in-flight delete swallows the round):** Fixed exactly as suggested — after `deleteSession`, `if appState.sessions.contains(staleReviewID) { completeSession(staleReviewID) }`, so a concurrent-delete no-op or disk-cleanup failure still makes the stale row invisible to `createReviewSession`'s dedup guard and the new round fires.

**🟢 4:** Added a comment noting the delete tears down a *live* round-1 clone (`rm -rf` clone, destroy terminal, drop raw telemetry) — intended, but a harder teardown than the old flip.

**🟢 5:** Rewrote web test #2 to an `active` work row (asserts the live session is never hidden) and added a no-group-survivor test; suite is 14/14. On the daemon side, `onReviewRequestsRefreshed` is wired inside `CrowDaemon.run` and isn't independently unit-testable without standing up the full daemon, so I left the delete-not-complete assertion uncovered there rather than add heavy scaffolding — happy to if you'd prefer.

**🟢 Quality:** collapse is now separate from assignment; web-tests `package.json` description mentions sidebar grouping.

`swift build` (CrowDaemon) ✅, `swift test --filter BoardForwarderTests` ✅ (23), `sidebar-groups.test.js` ✅ 14/14.

dhilgaertner (changes requested):
## Code & Security Review

Solid diagnosis — the "completed rows are invisible to the dedup guard but still render" root cause is exactly right, and the within-section collapse correctly avoids the two failure modes called out in the previous round (a live work row hidden by an open review clone; a survivor that matches no group). The daemon fix works. One consequence of swapping `complete` → `delete` was missed, and the collapse is broader than CROW-877 needs.

### Critical Issues

**🔴 Red 1 — every re-review round permanently erases round 1's telemetry from the efficiency scorecard.**

`CrowDaemon.swift:791` replaces `appState.onCompleteSession?(staleID)` with `sessionService.deleteSession(id:)`. The durable `SessionAnalyticsSnapshot` is written *only* on a terminal status transition:

- `SessionService.swift:3139` — `updateSessionStatus` → `scheduleAnalyticsSnapshot(for:status:)`
- `SessionService.swift:3174` — snapshots only on `.completed` / `.archived`

`deleteSession` never transitions status, so no snapshot is written. It then calls `telemetryDeleteProvider(id)` (`SessionService.swift:1818`), dropping the session's raw telemetry rows. And `backfillAnalyticsSnapshots` (`SessionService.swift:3236`) can't recover it — it requires a live `appState.sessions` record *and* the id to still be in `telemetrySessionIDsProvider()`, and after the delete neither exists.

Net: on every auto-reviewed repo, each push that advances the head silently and irreversibly removes round 1's tokens/cost/wall-clock from the scorecard. That's Crow's own primary flow. The in-code comment ("its raw telemetry rows are dropped. That's intended") accounts for the raw rows but not for the snapshot that now never gets written — and the repo already asserts the old behavior in `completeSessionRoutesThroughSessionServiceAndSnapshots()` (BoardForwarder suite, passing on this branch because nothing exercises the new path).

Note that a naive "complete, then delete" is **not** a safe fix: `scheduleAnalyticsSnapshot` fires a detached `Task` (`SessionService.swift:3176`), so it races `deleteSession`'s teardown — if the session row or the telemetry rows are gone first, `writeAnalyticsSnapshot` hits its empty-analytics guard and skips. The durable fix is to write the snapshot *synchronously before* teardown — e.g. `await writeAnalyticsSnapshot(for:status:.completed)` at the top of `deleteSession` (or a `deleteSession(preservingAnalytics:)` variant the daemon calls here). Alternatively, keep the completion for stale rounds and let the new web-side collapse handle the visible-duplicate half — the collapse already covers pile-ups for pre-existing data, per the PR description.

### Code Quality

**🟡 Yellow 1 — `collapsePRDuplicates` isn't status-gated, so it can hide a *live* session (`app.js:684`).**

Collapse runs on every section, but three of the five sections contain only non-terminal rows:

- `Reviews` (`app.js:656`) — `status !== 'completed' && status !== 'archived'`
- `Active` (`app.js:655`) — `status === 'active'`
- `In Review` (`app.js:657`) — `status === 'inReview'`

Collapsing inside those merges live sessions. Concretely: two open review sessions for one PR are reachable today — the kickoff serializer (`CrowDaemon.swift:335`) and the `start-review` RPC serializer (`RPCHandlers.swift:168`) are **separate** `ReviewKickoffSerializer` instances, so a manual "Start Review" during the ~10s clone window is guarded only by the lagging link lookup that the CROW-406 comments already describe as belt-and-suspenders. Both land in `Reviews`, and the second collapses away: no row, so no way to select it, open its terminal, or delete it — while its agent keeps running and its clone stays on disk. Same shape in `Active` for two `/crow-workspace` runs against one issue.

Before this PR both rows rendered and the duplicate was at least visible/deletable. CROW-877 only needs the terminal case, so gating the collapse would keep the fix and drop the risk:

```js
if (!url || (s.status !== 'completed' && s.status !== 'archived')) { out.push(s); continue; }
```

Test 6 covers *distinct* PRs; there's no case for two live rows on one PR in the same section.

**🟡 Yellow 2 — the new regression test never runs under `npm test` (`package.json:7`).**

The script chains with `&&` and `row.test.js` exits non-zero (11 failures — confirmed identical on `origin/main`, so the "pre-existing" claim checks out). I ran it: `npm test` stops after `row.test.js` and `node sidebar-groups.test.js` is never reached. Run directly it's **14/14 green**, so the coverage is good — it's just unreachable through the documented entry point, which means this PR's regression guard won't actually guard anything. Move it ahead of `row.test.js`, or switch to a form that runs every suite and aggregates the exit code.

### Security Review

**Strengths:**
- No new input handling, credential, or network surface. PR URLs flow through the existing parsed/validated paths.
- Deletion targets are safe by construction: `staleSessionID` originates from `ReviewRequest.reviewSessionID`, populated only from `appState.reviewSessions` (`IssueTracker.swift:594-597`), so `deleteSession` can never be aimed at a work session — and `performDiskCleanup(isReview: true)` (`SessionService.swift:1842`) only `removeItem`s the standalone clone, never touching branches or the main checkout. Good.
- The web change is pure data reshaping; no new DOM sinks, and `textContent` usage is unchanged.

**Concerns:** none found.

### Minor

- 🟢 `prURLOf` (`app.js:677`) reads only `s.links`, while `app.js:1274` and `app.js:1489` fall back to `liveFor(s.id).pr_link` for sessions whose PR link exists only in the app's memory. Fail-open (a missed collapse, never a wrong one), but "does this row have a PR" now means two different things in the same file.
- 🟢 The daemon-side fix — the primary one — has no automated coverage. `reviewKickoffAction` is pure and tested, but the delete + `contains`-fallback branch is inline in `wire…` and unreachable from a test. Lifting it into a small `SessionService` helper would let the fallback (`deleteSession` no-ops / disk-cleanup failure → `completeSession`) be asserted.
- 🟢 PR description says the new suite is 11/11; it's actually 14 checks.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 2 Yellow, 3 Green] findings.

**Verified locally:** `swift build` (CrowDaemon) ✅ · `swift test --filter BoardForwarderTests` 23/23 ✅ · `node sidebar-groups.test.js` 14/14 ✅ · `row.test.js` 26/11 on both this branch and `origin/main` (pre-existing).

---

dgershman:
Great catch on the telemetry regression — that one was invisible in the tests. Pushed 1f78a8d8.

**🔴 Red 1 (re-review erased round 1's scorecard telemetry):** You're right that the snapshot only lands on a status *transition* and `deleteSession` bypasses it, then drops the raw rows. Of your two suggested fixes I took the second — **revert the daemon to `onCompleteSession?(staleID)`** and let the web collapse own the visible-duplicate half:
- Completion writes the `SessionAnalyticsSnapshot` and keeps telemetry (no scorecard loss), and still makes the stale row invisible to `createReviewSession`'s dedup guard, so the new round starts.
- The within-section collapse already folds the completed `review-<repo>-<pr>` pile-up into one row — including pre-existing data — so nothing destructive is needed on the daemon side.
- Bonus: this also retires the previous round's Yellow 3 (delete no-op/failure fallback) and Green 4 (killing a live review clone) — there's no delete anymore. **Net daemon behavior is now identical to `main`**, and the fix lives entirely in `groupSessions`.

I went with revert-over-preserve-analytics because the snapshot-before-teardown variant is more moving parts (and still kills the live round-1 agent + clone) for no user-visible gain now that the collapse handles the pile-up. Happy to switch to the `deleteSession(preservingAnalytics:)` approach instead if you'd rather actually reap the stale clones — but that felt like scope beyond CROW-877, and clone reaping is the cleanup-retention path the ticket explicitly deferred.

**🟡 Yellow 1 (collapse could hide a live session):** Gated `collapsePRDuplicates` to `completed`/`archived` rows only — `if (!url || !isTerminal(s)) { out.push(s); continue; }`. Your two-open-reviews-in-Reviews race (and two `/crow-workspace` runs in Active) now both render; only terminal rows ever collapse, which still covers merged PRs and clone pile-ups. Added a regression test for it.

**🟡 Yellow 2 (regression test never ran under `npm test`):** Switched the script to a loop that runs every suite and aggregates the exit code, so `sidebar-groups.test.js` executes regardless of `row.test.js`'s pre-existing failures. Confirmed: all five suites now run; `sidebar-groups` is **16/16**, `row.test` still 26/11 (pre-existing, unchanged).

**🟢 Minors:** Added a comment on `prURLOf` noting it reads the persisted `links` array, which is authoritative for the terminal rows collapse touches (live-only `pr_link` sessions are non-terminal and never collapsed, so the fallback is moot here). PR description's "11/11" was stale — it's 16 checks now.

Verified: `swift build` (CrowDaemon) ✅ · `swift test --filter BoardForwarderTests` 23/23 ✅ · `sidebar-groups.test.js` 16/16 ✅.

dhilgaertner (changes requested):
## Code & Security Review

Round 3 lands the previous round's asks cleanly: the telemetry Red is gone (the stale round is completed again, so `scheduleAnalyticsSnapshot` still fires), the collapse is now gated to terminal rows so no live session can be hidden, and `package.json` runs every suite and aggregates the exit code — I confirmed all five now execute and `sidebar-groups.test.js` is **16/16**. The `row.test.js` failures are genuinely pre-existing: I ran it from a clean `main` worktree and got the identical `26 passed, 11 failed`.

What changed in the process, though, is the shape of the fix, and neither the code comments nor the PR body have caught up.

### Should Fix

**🟡 Yellow 1 — the clone pile-up is now hidden rather than reaped, and the hidden rows are unreachable from every GUI.**

The daemon diff is **comment-only**. `git diff main...HEAD -- CrowDaemon.swift` filtered to non-comment lines returns nothing; `CrowDaemon.swift:787` still calls `appState.onCompleteSession?(staleID)`, exactly as on `main`. So the stale review session is still created, still completed, still persisted — and still owns its standalone clone, because disk cleanup only runs from `deleteSession` (`SessionService.swift:1765`); completing a session never touches the filesystem. `cleanup.enabled` defaults to `false` (`AppConfig.swift:752-757`) and this PR deliberately leaves it that way, so nothing reaps them.

The new part is that they're now invisible. I ran the real `groupSessions` against a realistic pile-up:

```js
[{id:'work',   kind:'work',   status:'completed', links:L(1)},
 {id:'clone1', kind:'review', status:'completed', links:L(1)},
 {id:'clone2', kind:'review', status:'completed', links:L(1)},
 {id:'clone3', kind:'review', status:'completed', links:L(1)}]
→ [["Completed", ["work"]]]        // clone1..3 render nowhere
```

Those three have no `sessionRow`, so `selectSession` is unreachable for them; `sectionHeader`'s "All" iterates only the collapsed `rows` (`app.js:854`), so bulk-select can't reach them either; and the Reviews board's "go" uses `review_session_id`, populated only from `appState.reviewSessions` (`IssueTracker.swift:593-597`), which excludes completed. `Sources/Crow` has no sidebar and `crow-desktop/ui` is a single `index.html` shell, so `app.js` is the only GUI there is.

Net: on an auto-reviewed repo, every push that advances the head leaves a full repo clone on disk that the user can no longer find or delete from the UI at all — only via `crow list-sessions` + `crow delete-session`. Before this PR they were ugly but deletable, which is a meaningfully better failure mode than an invisible leak. Three ways out, cheapest first:

- keep an affordance on the survivor — a `+N` badge / expander so the collapsed rows are still selectable;
- archive the stale round instead of completing it, so `cleanup` reaps it once enabled;
- reap it properly: complete, then delete after the snapshot has landed (`await writeAnalyticsSnapshot(for:status:.completed)` before teardown, per the last round's note) — that satisfies both the telemetry Red and this one.

**🟡 Yellow 2 — the PR description now describes a change that isn't in the diff.**

- *"### 1. Primary — delete the stale review session on re-review (daemon)"* and *"Now it **deletes** the stale review session (`sessionService.deleteSession`), awaited before the new review is enqueued"* — none of this is in the branch. That was round 2's approach; it was reverted, and the section still reads as the headline fix.
- *"collapse by PR URL — a PR's work row + auto-review clone become one entry (the open review row wins while the review is in progress; the work row once the review closes)"* — the open-review-wins half is gone. With the `isTerminal` gate the collapse only ever fires in `Completed`.
- Consequently, symptom **(a)** — *"the same PR in both a review section and Completed"* — is not addressed for a live round. Verified:
  ```js
  [{id:'work', kind:'work',   status:'inReview', links:L(1)},
   {id:'rev',  kind:'review', status:'active',   links:L(1)}]
  → [["Reviews",["rev"]], ["In Review",["work"]]]   // still two rows for one PR
  ```
  That's the *right* call — collapsing across sections is what produced round 1's Red — but the description should say so rather than claim the symptom is fixed. The first-match grouping does genuinely fix the one-session-in-two-sections duplicate, which is a real half of (a).
- *"New jsdom regression test sidebar-groups.test.js (11/11)"* — it's 16 checks.

This body is what lands in history for CROW-877; as written it documents a daemon fix that doesn't exist.

### Security Review

**Strengths:**
- No new input handling, credential, network, or RPC surface. The change is pure client-side data reshaping over an already-trusted payload.
- `Map`/`Set` keyed on the PR URL rather than object literals — an attacker-supplied `__proto__` URL can't pollute a prototype. Right primitive.
- No new DOM sinks; `prURLOf` only string-compares `link.url`, and row rendering still goes through `textContent`.
- The daemon side adds no behavior at all, so its (already-audited) `staleSessionID` path is unchanged.

**Concerns:** none found.

### Code Quality

- ✅ `groupSessions` / `collapsePRDuplicates` are pure and directly testable, and the jsdom loader matches the existing `row.test.js` shape. Good structure.
- ✅ First-match grouping is the correct fix for the one-session-two-sections duplicate.
- ✅ The `package.json` rewrite is right — `fail=0; for f in …; do node "$f.test.js" || fail=1; done; exit $fail` is POSIX-clean and I confirmed all five suites run.
- 🟢 With the `isTerminal` gate, `collapsePRDuplicates` is a no-op in four of five sections — `Jobs`/`Active` are `status === 'active'`, `Reviews` excludes terminal, `In Review` is `inReview`. Only `Completed` can ever collapse. Defensible as future-proofing, but the invariant currently lives implicitly in the GROUPS predicates; applying the collapse to the terminal section alone would make it local and obvious.
- 🟢 A `paused` session matches no GROUP and renders nowhere — pre-existing (`GROUPS` is unchanged here), but test 2b now encodes it as expected behavior, which cements an invisible-session bug into the regression suite. Worth a follow-up ticket rather than a fix in this PR.
- 🟢 The web suites still don't run in CI — `.github/workflows/ci.yml` builds/tests the Swift packages only, nothing invokes `npm test`. This PR's entire behavioral change therefore has no automated guard, despite having a good test for it. Pre-existing gap; a node step (with `row.test.js`'s 11 known failures triaged first) would close it.
- 🟢 `prURLOf` (`app.js:681`) reads only `s.links`, while `app.js:1287` / `app.js:1502` fall back to `liveFor(s.id).pr_link`. Fail-open and the comment justifies it, but "does this row have a PR" now means two different things in one file. Carried over from the last round.
- 🟢 `collapsePRDuplicates` only promotes on `out[idx].kind === 'review' && s.kind !== 'review'`, so two non-review terminal rows for one PR (two completed work sessions, or a completed `job` + work row) keep whichever came first. Verified: `[w1 completed, w2 archived]` for one PR → one row. Narrow, and same class as Yellow 1.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 5 Green] findings.

The logic here is now correct — both Yellows are about the fix having become a *masking* fix without the code comments or the description saying so. Adding a `+N` affordance (or archiving the stale round) plus a description refresh closes this out.

**Verified locally:** `swift build --package-path Packages/CrowDaemon` ✅ (one pre-existing warning in `RPCWebSocketHandler.swift`) · `npm test` runs all 5 suites ✅ · `sidebar-groups.test.js` 16/16 ✅ · `row.test.js` 26/11 on this branch **and** on a clean `main` worktree (pre-existing, confirmed) · `groupSessions` exercised directly against real-world session shapes.

---

dhilgaertner (changes requested):
## Code & Security Review

Round 4. The branch is unchanged since round 3 — `e192e6e` (2026-07-28T02:14:58Z) predates the last review (02:19:26Z), and `git diff` against it is empty. So this re-verifies from scratch rather than re-reading the last round's notes.

Everything the last round asked for on the *mechanics* holds up under independent checking: `npm test` now runs all five suites and aggregates the exit code, `sidebar-groups.test.js` is 16/16, the collapse is gated to terminal rows so no live session can be hidden, and `row.test.js`'s 11 failures reproduce identically on a clean `origin/main` worktree (`59be7b1`), so the "pre-existing" claim is confirmed. The first-match grouping is correct and is a genuine fix.

The blocking problem is unchanged: what shipped is a **masking** fix, and the ticket's primary root cause is untouched.

### Critical Issues

**🔴 Red 1 — the clone pile-up is now hidden rather than reaped, and the hidden rows are unreachable from the only GUI Crow has.**

The daemon half of this PR contains **zero** behavioral change. Filtering the diff to non-comment lines returns nothing:

```
$ git diff 81be85e..HEAD -- Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift \
    | grep -E '^[+-]' | grep -vE '^[+-]{3}|^[+-]\s*//|^[+-]\s*$'
(no output)
```

`CrowDaemon.swift:787` still calls `appState.onCompleteSession?(staleID)` → `completeSession` → `updateSessionStatus(id, to: .completed)` (`SessionService.swift:3338`). Completing never touches the filesystem — `performDiskCleanup` has exactly one caller, `deleteSession` (`SessionService.swift:1766`). And `cleanup.enabled` defaults to `false` (`AppConfig.swift:792`, `800`), which this PR deliberately leaves alone. So every re-review round still mints a review session, still completes it, still leaves its standalone clone on disk.

What this PR changes is that those rows now render **nowhere**. Ran the real `groupSessions` in the jsdom harness against a three-round pile-up on a merged PR:

```js
[{id:'work',   kind:'work',   status:'completed', links:L(1)},
 {id:'clone1', kind:'review', status:'completed', links:L(1)},
 {id:'clone2', kind:'review', status:'completed', links:L(1)},
 {id:'clone3', kind:'review', status:'completed', links:L(1)}]
→ [["Completed",["work"]]]      // clone1..3 render nowhere
```

There is no other route to them:
- no `sessionRow` → `selectSession` is unreachable (`app.js:1662` is only called from a row click, the manager nav pill, and the review board's "go");
- `sectionHeader`'s "All" iterates only the collapsed `rows` (`app.js:854`), so **bulk-select can't reach them either** — a user who select-alls "Completed" and hits delete removes the work row and silently leaves N clones behind;
- the review board's "go" uses `review_session_id`, populated from `appState.reviewSessions`, which filters out completed/archived (`AppState.swift:518-520`);
- there is no second GUI. `Sources/Crow` contains one Swift file and no sidebar; `crow-desktop/ui` is a single `index.html` Tauri shell. `app.js` is it.

There's also a live-session case: select an open review, let it finish. Its status flips to `completed`, it moves to "Completed", collapses into the work row, and **its row disappears while you're looking at it** — the detail pane keeps rendering off `sessions.find`, but navigate away once and it's gone for good.

Net: before this PR the clones were ugly but visible and deletable. After it, on any auto-reviewed repo, each push that advances the head leaves a full repo clone that the user cannot find or delete from the UI at all — only via `crow list-sessions` + `crow delete-session`. Removing the sole remediation affordance while leaving the leak is a worse failure mode than the bug being fixed, and issue #877 names the daemon teardown as fix direction **1 — Primary** ("delete the stale review session on re-review, don't complete it"). Merging with `Closes #877` retires the ticket with (a) and (b) still happening in state and on disk.

Three ways out, cheapest first:
- keep an affordance on the survivor — a `+N` badge / expander so the collapsed rows stay selectable and deletable;
- **archive** the stale round instead of completing it, so `cleanup` reaps it once enabled;
- reap it properly: complete, then delete after the snapshot has landed (`await writeAnalyticsSnapshot(for:status:.completed)` before teardown, per round 2's note) — that satisfies both the telemetry Red from round 2 and this one.

### Should Fix

**🟡 Yellow 1 — the PR description documents a change that isn't in the branch.**

This body is what lands in history for CROW-877, and three of its four claims are wrong:

- *"### 1. Primary — delete the stale review session on re-review (daemon)"* / *"Now it **deletes** the stale review session (`sessionService.deleteSession`), awaited before the new review is enqueued"* — not in the diff (see above). That was round 2's approach; it was reverted, and the section still reads as the headline fix.
- *"collapse by PR URL … the open review row wins while the review is in progress"* — gone. With the `isTerminal` gate (`app.js:702`) the collapse only ever fires in "Completed".
- Consequently symptom **(a)** — *"the same PR in both a review section and Completed"* — is not addressed for a live round. Verified:
  ```js
  [{id:'work', kind:'work',   status:'completed', links:L(2)},
   {id:'rev',  kind:'review', status:'active',    links:L(2)}]
  → [["Reviews",["rev"]], ["Completed",["work"]]]   // still two rows for one PR
  ```
  That's the *right* call — cross-section collapse is what produced round 1's Reds — but say so rather than claim the symptom is fixed. First-match grouping does fix the genuine half of (a): one session double-matching Reviews + In Review.
- *"New jsdom regression test `sidebar-groups.test.js` (11/11)"* — it's 16 checks.

### Security Review

**Strengths:**
- No new input handling, credential, network, or RPC surface. The change is pure client-side reshaping of an already-trusted `list-sessions` payload; the daemon adds no behavior at all.
- `Map`/`Set` keyed on the PR URL rather than object literals — an attacker-supplied `__proto__` URL can't pollute a prototype. Right primitive.
- No new DOM sinks: the diff introduces zero `innerHTML`/`eval`, and rows still go through `el()`'s `textContent`.
- `prURLOf` (`app.js:681`) only string-compares `link.url`; it never navigates to or renders it.

**Concerns:** none found.

### Code Quality

- ✅ `groupSessions` / `collapsePRDuplicates` are pure and directly testable, and the jsdom loader matches the existing `row.test.js` shape. Good structure.
- ✅ First-match grouping (`app.js:709-718`) is the correct fix for the one-session-two-sections duplicate.
- ✅ The `package.json` rewrite is right — `fail=0; for f in …; do node "$f.test.js" || fail=1; done; exit $fail` is POSIX-clean, and I confirmed all five suites now execute.
- 🟢 **New:** the collapse is an exact string match on `link.url`, so trivial URL variance defeats it. Verified — a trailing slash yields two rows for one PR:
  ```js
  [{id:'work',  status:'completed', kind:'work',   links:[{url:'…/pull/4',  type:'pr'}]},
   {id:'clone', status:'completed', kind:'review', links:[{url:'…/pull/4/', type:'pr'}]}]
  → [["Completed",["work","clone"]]]
  ```
  Fail-open (a missed collapse, never a wrong one), and today both links come from the same parser — but a normalizing key (strip trailing `/`, lowercase host) costs one line and removes the dependency on that staying true.
- 🟢 `collapsePRDuplicates` only promotes on `out[idx].kind === 'review' && s.kind !== 'review'`, so two *non-review* terminal rows for one PR (completed `work` + archived `job`) keep whichever came first — verified, collapses to one. Narrow, same class as Red 1.
- 🟢 With the `isTerminal` gate the collapse is a no-op in four of five sections (`Jobs`/`Active` are `status === 'active'`, `Reviews` excludes terminal, `In Review` is `inReview`) — only `Completed` can ever collapse. Defensible as future-proofing, but the invariant lives implicitly in the GROUPS predicates; note also that a live row is pushed to `out` *without* registering in `byPR`, so extending collapse to live rows later would need more than flipping the gate.
- 🟢 `prURLOf` reads only `s.links`, while `app.js:1287`/`app.js:1502` fall back to `liveFor(s.id).pr_link`. Fail-open and the comment justifies it, but "does this row have a PR" now means two things in one file. Carried from round 2.
- 🟢 A `paused` session with a PR link matches no GROUP and renders nowhere (verified: `→ []`). Pre-existing — GROUPS is unchanged — but test 2b now encodes it as *expected*, cementing an invisible-session bug into the regression suite. Worth a follow-up ticket, not a fix here.
- 🟢 The web suites still don't run in CI: `.github/workflows/ci.yml` builds/tests the Swift packages, `cargo build`s the Tauri shell and runs `shellcheck`, but nothing invokes `npm test`. This PR's entire behavioral change therefore has no automated guard despite having a good test for it. Pre-existing gap; a node step (with `row.test.js`'s 11 known failures triaged first) would close it.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 1 Yellow, 6 Green] findings.

The client-side logic is correct and well tested — the problem is that the fix became a masking fix without the code, the comments, or the description saying so, while `Closes #877` would retire the ticket. Restoring an affordance for the collapsed rows (a `+N` expander), or archiving/reaping the stale round, plus a description refresh, closes this out.

**Verified locally:** `swift build --package-path Packages/CrowDaemon` ✅ · `swift test --filter BoardForwarderTests` 26/26 ✅ · `npm test` runs all 5 suites ✅ · `sidebar-groups.test.js` 16/16 ✅ · `row.test.js` 26 passed / 11 failed on this branch **and** on a clean `origin/main` worktree (`59be7b1`) — pre-existing, confirmed · daemon diff confirmed comment-only · `groupSessions` exercised directly against real-world session shapes.

---

dhilgaertner (changes requested):
## Code & Security Review

The sidebar half of this is well built — `groupSessions` is a pure, testable function, the within-section/terminal-only gating directly answers the two failure modes earlier rounds hit, and the 16 new jsdom assertions all pass locally and encode *why* each guard exists. The blocker below is a merge artifact, not a design problem.

### Critical Issues

**🔴 Red 1 — Duplicate ledger rows break the CI parity gate.** `Packages/CrowCore/Sources/CrowCore/Parity/ParityLedger.swift:221-222`

Commit `83ce96b` added `automation-get` / `automation-set` to `rpcMethods`, but `main` had already been fixed independently — the same pair sits at `ParityLedger.swift:204-205` (with the CROW-812 comment explaining the #883/#884 collision). The merge of `main` (`b0b6734`) didn't reconcile the two, so `rpcMethods` now carries the pair twice.

Verified locally:

```
$ swift test --package-path Packages/CrowCLI --filter ParityGate
✘ Test "Ledger rows are unique" recorded an issue at ParityGateTests.swift:210:9:
  Expectation failed: (dupeMethods → ["automation-set", "automation-get"]).isEmpty → false
  ↳ Duplicate rpcMethods rows: ["automation-get", "automation-set"]
✘ Test run with 7 tests in 1 suite failed
```

`CrowCLI` is in `LINUX_PACKAGES` (`.github/workflows/ci.yml:59`), which runs `swift test --package-path Packages/CrowCLI` on every PR — so this fails the required lane. Note `scripts/check-cli-parity.sh` does **not** catch it (both `registered_methods` and `ledger_methods` end in `sort -u`), which is why it reports green: `CLI/RPC parity: registered methods match the ledger (94 methods)`.

Fix: delete lines 221-222; `main`'s copy at 204-205 already carries the rationale comment.

### Security Review

**Strengths:**
- No new attack surface. The web change is pure client-side array partitioning; rows are still built through `el()`/`textContent`, so no `innerHTML` sink is introduced and no new session-supplied string reaches the DOM as markup.
- `prURLOf` only reads a URL for equality comparison — it is never used to navigate, fetch, or construct a link target, so an attacker-controlled PR URL in `links` can at worst mis-group a row in the viewer's own sidebar.
- The `package.json` test runner iterates a hard-coded literal list; no interpolation of external input into the shell.
- The daemon hunk is comment-only — no change to authority gating, the `.reReview` teardown, or the kickoff serializer. Completion runs synchronously on `@MainActor` (`SessionService.completeSession` → `updateSessionStatus`) before the `Task { reviewSerializer.enqueue … }` is spawned, so `existingReviewSession(forPRURL:)` correctly sees the stale row as completed. No TOCTOU there.

**Concerns:** none found.

### Code Quality

**🟡 Yellow 1 — the PR description no longer describes the code.** The body's headline fix, "*1. Primary — delete the stale review session on re-review (daemon) … Now it **deletes** the stale review session (`sessionService.deleteSession`), awaited *before* the new review is enqueued*", is not in this branch. `CrowDaemon.swift`'s entire +16/−2 is comment; `CrowDaemon.swift:806-808` still calls `appState.onCompleteSession?(staleID)`, and the new comment at `:790-805` explicitly argues *against* deleting (analytics snapshot preservation). Commit `248555f` ("keep completion") reverted it and the body was never updated. Crow squashes the description into history, so this permanently misfiles the change. Please rewrite §1 to describe what shipped: the daemon behavior is unchanged, and the sidebar collapse is now the whole fix. (Minor, same section: the testing note says 11 pre-existing `row.test.js` failures; the actual count is 10 — I ran it against `main` at `45f3108` to confirm they're pre-existing.)

**🟡 Yellow 2 — the collapse isn't limited to review clones, so it can hide a real work session.** `app.js:700-711`

`collapsePRDuplicates` folds *any* two terminal rows that share a PR URL. The tie-break only promotes work over review:

```js
if (out[idx].kind === 'review' && s.kind !== 'review') out[idx] = s;   // work wins
```

When both rows are `kind:'work'`, neither branch fires and `s` is silently dropped. Two completed work sessions can legitimately share a PR — `crow add-link --session <uuid> --label "PR" --url … --type pr` is a documented verb, and a follow-up session on an existing PR is the ordinary way to get there. The second one then vanishes from the sidebar with its own worktree and branch still on disk, and no way to select, open, or delete it. The ticket is about review clones; gating the collapse on at least one side being `kind === 'review'` keeps the fix and removes this case:

```js
if (idx === undefined) { byPR.set(url, out.length); out.push(s); continue; }
if (out[idx].kind !== 'review' && s.kind !== 'review') { out.push(s); continue; }  // two work rows: not a clone pair
if (out[idx].kind === 'review' && s.kind !== 'review') out[idx] = s;
```

Worth a test alongside the existing ones — the current suite covers review+work and review+review, but not work+work.

**🟡 Yellow 3 — collapsed rows become unreachable for cleanup.** `app.js:816-819`, `app.js:854-869`

`sectionHeader(title, rows)` builds its select-all `ids` from the *collapsed* rows, and `bulkDeleteSelected` (`app.js:891`) only ever deletes `selectedSessionIDs`. So a hidden completed review clone can't be selected — not individually, not via "All". Combined with the deliberate decision in this PR to complete-but-never-delete stale review sessions, and `cleanup.enabled` defaulting to `false`, the clones and their on-disk review clone directories now accumulate indefinitely *and* invisibly. Clearing a PR with N stale rounds takes N delete-confirm round-trips (delete the survivor, the next clone surfaces, repeat). Before this PR they were at least visible and bulk-deletable in one pass.

That's a real regression in reachability, and it's the direct cost of dropping the daemon-side teardown. Some way out is needed — either have the section header's select-all include the collapsed ids, or render a "+N" affordance on the survivor that expands the group, or reinstate a teardown that reaps stale clones once their snapshot is written. Whichever you pick, please say so in the PR body so the trade-off is on the record.

**🟡 Yellow 4 — `prURLOf` duplicates an existing helper.** `app.js:684-690` vs `app.js:1564-1569`

`prUrlForSession(s)` already exists in this file and does the same extraction, plus the `liveFor(s.id).pr_link` fallback that `sessionRow` (`app.js:1316`) and `sessionMenuItems` (`app.js:1574`) rely on. Two helpers reading the same field will drift the first time the link shape changes. Function declarations hoist, so `groupSessions` can just call `prUrlForSession(s)` despite being defined ~880 lines earlier. Behavior is unchanged by the swap — `pr_link` is built from the same `appState.links(for:)` source (`RPCHandlers.swift:950`) that populates `links` (`RPCHandlers.swift:256-265`), so the fallback is belt-and-suspenders, not a different source.

**🟢 Green 1 — the collapse survivor is the oldest round, not the newest.** `app.js:706-708`

Among several completed review clones the first-seen wins, and `list-sessions` emits `appState.sessions` in append order (`RPCHandlers.swift:215`) — i.e. creation order. So a PR with three review rounds shows round 1's row, whose name, timestamps, and analytics are the least interesting of the three. Preferring the newest `updated_at` among same-kind rows would make the surviving row the one a user actually wants to open.

**🟢 Green 2 — `package.json` runner change is right, worth a one-line note.** Switching from the `&&` chain to `for … || fail=1` is the correct call — it stops a pre-existing red suite from masking the new one, and the exit code is still faithful. Since `npm test` is red on `main` today (10 `row.test.js` failures, unrelated to this change) and isn't wired into any CI workflow, a comment in `package.json` or the PR body pointing at the tracking issue for those 10 would keep the next person from assuming the command is expected to pass.

### Test & Build Verification

| Check | Result |
|---|---|
| `node sidebar-groups.test.js` | ✅ 16 passed, 0 failed |
| `npm test` (whole web suite) | ⚠️ new suite green; `row.test.js` 43/10 — **pre-existing**, identical on `main` @ `45f3108` |
| `swift build` (root package) | ✅ Build complete (72s) |
| `swift test --package-path Packages/CrowCLI --filter ParityGate` | ❌ "Ledger rows are unique" fails — Red 1 |
| `bash scripts/check-cli-parity.sh` | ✅ 94 methods (does not detect the duplicate — `sort -u`) |

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 4 Yellow, 2 Green] findings. Red 1 is a two-line deletion that unblocks CI; Yellow 2 and 3 are the ones worth thinking about before this lands.

dgershman:
Thanks — pushed e9d163ed. Round 5's Red was the ledger dup; fixed that plus the two substantive Yellows.

**🔴 Red 1 (duplicate ledger rows):** Fixed by rebasing onto the latest `main` (`45f3108f`) and **dropping** my redundant `83ce96bc` — `main` had already added `automation-get/set` (the CROW-812 comment). The pair now appears once. `swift test --package-path Packages/CrowCLI --filter ParityGate` → "Ledger rows are unique" passes (7/7); `check-cli-parity.sh` green (94).

**🟡 Yellow 2 (collapse could hide a real work session):** Gated the collapse so a pair folds **only when at least one side is a review clone**. Two independent work/job sessions sharing a PR (a follow-up session, or a manual `add-link`) now both render — neither is dropped. Added a `work+work` test.

**🟡 Yellow 3 (collapsed rows unreachable for cleanup):** Took your first option — `groupSessions` now returns `allIds` per section (survivors + folded-away ids), and the section header's **select-all uses it**, so bulk-delete reaches a hidden collapsed clone in one pass (your "select-alls Completed and hits delete leaves N clones behind" case is gone). Added a test asserting `allIds` includes the collapsed clone.

I went with select-all-reaches-collapsed over reaping in the daemon: reaping (complete → snapshot → delete) is the more invasive path (kills the live round-1 agent, delete-failure fallback, telemetry-ordering), and this restores the pre-PR reachability (they were deletable before; now bulk-deletable) with a purely additive, well-tested sidebar change. Happy to switch to the snapshot-then-reap approach if you'd rather the clones actually go away on disk — but that felt like it belongs with the cleanup-retention work the ticket deferred, and I'd want it as its own change.

**🟡 Yellow 4 (duplicate helper):** Dropped `prURLOf`; `groupSessions` now calls the existing hoisted `prUrlForSession`.

**🟡 Yellow 1 (PR description drift):** Rewrote the body to describe what actually shipped — daemon unchanged (completes the stale round, snapshot preserved), the sidebar collapse is the whole fix, first-match grouping fixes the genuine half of (a), the cross-section-while-open pair deliberately stays two rows, and the count is 20 checks now.

**🟢 Green 1 (survivor = oldest round):** The `list-sessions` payload has no top-level session `updated_at` (the `updated_at` there is the linked issue's), so preferring the newest round would need a new payload field — left it for a follow-up rather than widen the payload here.

Verified: `sidebar-groups.test.js` 20/20 · all five web suites run (`row.test.js` 43/10, pre-existing/unchanged) · `swift build` (CrowDaemon) ✅ · ParityGate + BoardForwarder ✅.

dhilgaertner (approved):
## Code & Security Review

Reviewed at `e9d163e`. Every finding from the previous round is fixed, and I verified each one against the code and the test suite rather than the commit messages. The branch was also rebased off the `main` merge that carried last round's Red — `ParityLedger.swift` now has a single `automation-get`/`automation-set` pair at `:204-205`, and the parity gate passes.

### Critical Issues

None.

### Previous round — verification

| Finding | Status | Evidence |
|---|---|---|
| 🔴 Red 1 — duplicate ledger rows fail CI | ✅ Fixed | The merge commit `b0b6734` and `83ce96b` are gone from the head; `rg automation-get` on `ParityLedger.swift` returns one row pair (`:204-205`). `swift test --filter ParityGate` → "Ledger rows are unique" **passes**, 7/7. |
| 🟡 Yellow 1 — PR body described code that wasn't there | ✅ Fixed | The body now leads with "the fix is **entirely in the web sidebar**; the daemon behavior is **unchanged**". Confirmed literally: filtering the daemon diff to non-comment lines returns **zero** lines — `CrowDaemon.swift`'s +16/−2 is pure comment, and `:806-808` still calls `onCompleteSession?(staleID)`. The pre-existing failure count is now correctly stated as 10. |
| 🟡 Yellow 2 — collapse could hide a real work session | ✅ Fixed | `app.js:707` — `if (prior.kind !== 'review' && s.kind !== 'review') { out.push(s); continue; }` gates the fold to pairs where at least one side is a review clone. Covered by test 10 (work+work, both render). |
| 🟡 Yellow 3 — collapsed rows unreachable for cleanup | ✅ Fixed | `groupSessions` returns `allIds` (`app.js:733`) and `sectionHeader(title, ids)` (`app.js:868`) builds select-all from it. Traced end-to-end: `bulkDeleteSelected` (`app.js:910-915`) iterates `selectedSessionIDs` and issues `delete-session` per id with no dependence on a rendered row, so a folded clone really is deleted. Covered by test 11. |
| 🟡 Yellow 4 — `prURLOf` duplicated an existing helper | ✅ Fixed | `prURLOf` is gone; `collapsePRDuplicates` calls `prUrlForSession` (`app.js:703`), which is a hoisted `function` declaration at `:1577`, so the forward reference resolves. |

I also walked `collapsePRDuplicates` through the orderings the tests don't enumerate — `work,work,clone`, `clone,work,work`, `clone,clone,work`, `work,clone,work` — and the anchor bookkeeping holds in each: `byPR` keeps the first index, `out[idx] = s` swaps in place, and no session id can land in both `rows` and `collapsedIds`. Every session either renders, is recorded in `collapsedIds`, or matched no `GROUP` — nothing is silently dropped by the new code.

One structural note in the fix's favour: because `GROUPS` partitions by status, **every section is uniformly terminal or uniformly live** — only "Completed" holds `completed`/`archived` rows. So the `isTerminal` gate is defence-in-depth rather than load-bearing, and the collapse can only ever act in "Completed". That is the right shape for this fix.

### Security Review

**Strengths:**
- No new attack surface. The change is client-side array partitioning; rows are still constructed via `el()` (`app.js:771-776`), which assigns through `textContent`, so no session-controlled string reaches the DOM as markup and no `innerHTML` sink is added.
- `prUrlForSession` is used only for string equality between rows — never to navigate, fetch, or build a link target. A hostile PR URL in `links` can at worst mis-group a row in the viewer's own sidebar.
- The `package.json` runner iterates a hard-coded literal suite list; no external input is interpolated into the shell.
- The daemon hunk is comment-only, so nothing changes in authority gating, the `.reReview` teardown, or the kickoff serializer. Keeping `onCompleteSession` (rather than `deleteSession`) is also the safer choice for data integrity: completion writes the end-of-round analytics snapshot instead of dropping the raw telemetry rows.

**Concerns:** none found.

### Code Quality

**🟢 Green 1 — the collapse survivor is the oldest round, not the newest.** `app.js:704-706`

Carried over from last round, still open. Among several completed review clones the first-seen wins, and `list-sessions` emits `appState.sessions` in append (creation) order, so a PR with three review rounds surfaces round 1's row — the least interesting of the three. Preferring the newest `updated_at` among same-kind rows would make the survivor the row a user actually wants to open. Cosmetic given the rows are near-identical, so it's fine to leave.

**🟢 Green 2 — paused work/job sessions still match no section (pre-existing).** `app.js:656-662`

`Jobs` and `Active` both require `status === 'active'`, and `Reviews` requires `kind === 'review'`, so a `paused` **work** or **job** session matches no `GROUP` and renders nowhere — even though `crow set-status --session <uuid> paused` is a documented verb. This is unchanged from `main` (the old per-group `sessions.filter` dropped it identically), so it is **not** a regression from this PR, and test 2b shows you're already aware sessions can match no group. Flagging it only because `groupSessions` is now the one place this would be fixed — a trailing catch-all bucket would be a couple of lines. Worth its own ticket rather than scope creep here.

**🟢 Green 3 — no tracking reference for the 10 pre-existing `row.test.js` failures.** `package.json:7`

The body notes the count, which is enough to stop the next person assuming `npm test` should be green, but a ticket link next to it would be better since nothing in CI runs this command yet.

### Test & Build Verification

All run locally at `e9d163e`:

| Check | Result |
|---|---|
| `node sidebar-groups.test.js` | ✅ **20 passed, 0 failed** |
| `npm test` (whole web suite) | ✅ runs every suite; new suite 20/20 green |
| `row.test.js` @ `e9d163e` | ⚠️ 43 passed / 10 failed |
| `row.test.js` @ `origin/main` (`45f3108`, clean worktree) | ⚠️ 43 passed / 10 failed — **identical, confirmed pre-existing and unrelated** |
| `swift build` (root) | ✅ Build complete (12.26s) |
| `swift test --package-path Packages/CrowDaemon --filter BoardForwarderTests` | ✅ 26 tests / 6 suites passed |
| `swift test --package-path Packages/CrowCLI --filter ParityGate` | ✅ 7/7 passed — including "Ledger rows are unique" |
| `bash scripts/check-cli-parity.sh` | ✅ 94 methods match |

The `package.json` runner change does what it claims: I confirmed the loop executes all six suites and still exits non-zero on `row.test.js`, so the pre-existing red is reported rather than masked, and `sidebar-groups` no longer gets short-circuited away.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 3 Green] findings. The three Greens are optional polish; none should hold the merge. Nice work converging this — `groupSessions` ended up as a pure, well-commented function whose 20 assertions each encode *why* a guard exists, which is exactly what the four rounds of failure modes needed.

---

PR #862, opened 2026-07-26
CrowAntigravity adapter (Tier-2): agy CLI, hooks-based state, resume (#860)

dgershman:
Closes #860

Adds a **Tier-2 / experimental** `CrowAntigravity` adapter for Google Antigravity's CLI (binary **`agy`**), per [ADR 0014](https://github.com/corveil/crow/blob/main/docs/adr/0014-pluggable-coding-agent-adapter.md) / [0015](https://github.com/corveil/crow/blob/main/docs/adr/0015-harness-capability-tiers.md). New package `Packages/CrowAntigravity/`, structurally a near-clone of `CrowCursor`.

## What's here (Phase A core loop)

- **`AntigravityAgent`** — `launchCommandToken = "agy"`, remote control faked via `crow send`, `-p` prompt injection (`.job`), `-c` resume, `.review`/`.manager` → `nil`. `sessionRenameSlashCommand` left at the opt-out `nil` default (unverified on v1.1.7).
- **`AntigravityHookConfigWriter`** — per-worktree `.agents/hooks.json` in Antigravity's **own named-group schema** (verified against [`antigravity.google/docs/hooks`](https://antigravity.google/docs/hooks)), *not* Claude's `{"hooks": {…}}`: Crow owns one group `{"crow": {…}}`; `PostToolUse` wraps handlers in `{matcher:"*", hooks:[…]}`; `PreInvocation`/`PostInvocation`/`Stop` list handlers directly under the event key; **no `async` field**. Session UUID baked into each `crow hook-event … --agent antigravity` command. Inherits Cursor's shared-file protections (git-tracked skip, unparseable-guard, atomic write, git-exclude); Crow only ever writes/removes its own `"crow"` group. `removeManagedGlobalConfig` strips a stale `~/.gemini/config/hooks.json` Crow group (double-fire guard).
- **stdout verdict contract** — each hook command suppresses `crow hook-event`'s output and `printf '{}'`s the verdict itself, exiting 0 via `;`, so a Crow hook can never block a tool or loop the agent on `Stop`. **`PreToolUse` is deliberately not registered** (its strict `{"decision":…}` gate denies every tool call when mis-shaped — cmux #5358 — so, like Antigravity's own vibe-island plugin, Crow stays off it).
[...]

PR #880, opened 2026-07-27
Agent pickers: surface all known agents, grey out uninstalled ones

### LINKED ISSUE #879: Agent pickers: always show all agents, grey out uninstalled ones with a 'not found on PATH' help tooltip
Today, a coding agent is **completely hidden** from the UI when its binary isn't found on the daemon's PATH — so a shipped, fully-capable agent (e.g. Antigravity, whose adapter merged in #862) silently doesn't appear as a Manager launch option or in the Settings agent selectors, with no explanation. That reads as broken/missing rather than "not installed."

**Requested behavior:** always show **all** known agents in the pickers regardless of install state, but **grey out / disable** the ones whose binary isn't found, with a help tooltip explaining that the binary must be discoverable on PATH (install it + restart Crow) to become selectable. Some kind of inline help info so it's obvious *why* it's disabled.

## Why it's hidden today (root cause)
- Agents are registered **only if `findBinary()` resolves** their binary, in `registerAgents(devRoot:)` — `CrowDaemon.swift:893-927` (Antigravity block at `:924-926`; the comment there even states "off-PATH ⇒ silently absent from the picker and handoff (ADR 0014)"). Off-PATH ⇒ never enters `AgentRegistry`.
- The `list-agents` RPC returns only registered agents — `RPCHandlers.swift:568-580` (mirrored in `EngineRouter.swift:478-490`).
- Both UI surfaces are built purely from `list-agents`, so an unregistered agent is absent from all of them:
  - Manager launch picker — `openNewManagerMenu`, `app.js:1054-1072`
  - Settings job/manager/review selectors — `agentOverrideField`, `settings.js:320-338`
- Known agents are declared in `AgentKind.swift` (`claudeCode`, `codex`, `cursor`, `openCode`, `antigravity`). There is **no** `supportsManager`/capability gate involved — this is purely the install/availability gate.

## Implementation direction
1. Make `list-agents` report **all known agents** with an `available`/`installed` boolean (resolved binary path or nil) instead of omitting unfound ones. Simplest: register every known agent but attach an availability flag from `findBinary()`, or add a parallel "known agents + availability" source for the UI. (Availability is a **boot-time snapshot** — registration runs once at daemon start, so the tooltip copy should say "install it, then restart Crow.")
2. In both pickers (`app.js:1054-1072`, `settings.js:320-338`), render unavailable agents as **disabled/greyed** options with a tooltip, e.g. *"Antigravity (agy) not found on PATH — install it and restart Crow to enable."*
3. Block actually launching/handing-off to an unavailable agent (keep it selectable-as-info only, or prevent selection while surfacing the hint).

## Notes
- This intentionally changes the behavior documented in **ADR 0014** (off-PATH ⇒ silently absent). ADR 0014 should be updated to reflect "surface-but-disable" — flagging here; don't pre-edit the ADR.
- Keep scope to surface + disable + tooltip. Live re-probing / auto-refresh on install is out of scope.

dgershman:
Closes #879

## Problem
An off-PATH coding agent was **hidden entirely** from the UI. `registerAgents` only entered an agent into `AgentRegistry` when `findBinary()` resolved, so `list-agents` — and therefore the Manager launch menu (`app.js`) and the Settings job/manager/review selectors (`settings.js`) — never saw it. A shipped, fully-capable agent (e.g. Antigravity, adapter merged in #862) silently didn't appear, reading as broken/missing rather than "not installed."

## Change: surface-but-disable
- **`AgentRegistry`** — add `registerKnown(_:available:)` and `allKnownAgents()`. Available agents enter the launchable `agents` map (unchanged behavior); **unavailable agents are recorded for the UI only** and never enter `agents`, so `registeredKind`/`agent(for:)` still refuse to launch or hand off to them. `register(_:)` now delegates as `available: true`.
- **`CrowDaemon.registerAgents`** — register all five known agents, marking `available` from `findBinary()` (Claude Code always available). Off-PATH ⇒ known-but-unavailable instead of omitted.
- **`list-agents` RPC** (both `RPCHandlers.swift` and `EngineRouter.swift`) — each agent object now carries `available` (bool) and `binary` (the PATH token, e.g. `agy`) alongside `kind`/`name`/`default`.
- **Web pickers** — unavailable agents render **disabled/greyed** with an inline hint and a tooltip: *"Antigravity (agy) not found on PATH — install it and restart Crow to enable."* The new-manager menu always shows the full roster now (discoverability is the point); disabled rows are info-only and can't launch. A shared `agentUnavailableHint()` keeps the copy identical across both surfaces.
- **ADR 0014** updated from "off-PATH ⇒ silently absent" to the surface-but-disable behavior; the launch gate is unchanged, only the UI/`list-agents` surface widened.

Availability is a **boot-time snapshot** (registration runs once at daemon start), hence the "restart Crow" copy. Live re-probing on install is intentionally out of scope.

## Tests
- `AgentsLocalTests.unavailableAgentSurfacesButStaysUnlaunchable` — an unavailable agent is listed by `allKnownAgents()` (`available: false`) but is not launchable (`agent(for:)`/`registeredKind`/`defaultAgent` all miss).
- `AgentsLocalTests.availableAgentIsLaunchableAndDefault` — the available path stays launchable + default.
- `listAgentsIsLocalNotForwarded` extended to assert the new `available` + `binary` fields.
- `swift build` clean; targeted CrowDaemon/CrowCore suites green; `app.js`/`settings.js` pass `node --check`.

## Test plan (manual)
1. On a machine where only some agents are installed, click the Manager **"+"** — all known agents appear; uninstalled ones are greyed with the tooltip and can't be selected.
2. Settings → agent selectors show uninstalled agents as disabled `(not installed)` options with the same hint.
3. Selecting/launching an uninstalled agent is blocked (disabled option / swallowed click).


dhilgaertner (changes requested):
## Code & Security Review

The core mechanic is right: `registerKnown(_:available:)` keeps unavailable agents out of the launchable `agents` map, so `agent(for:)` / `registeredKind` / `defaultAgent` still refuse them, and the new `unavailableAgentSurfacesButStaysUnlaunchable` test pins that. `swift build` is clean, the targeted CrowDaemon suites pass (4/4), and both JS files pass `node --check`.

The problem is scope: `list-agents` is a **shared** payload with three consumers, and only two were updated. The third (handoff) and the Settings *Default agent* picker now hand users off-PATH agents as fully selectable options.

### Critical Issues

**🔴 1. Settings → "Default agent" picker is not gated — it persists an unlaunchable `defaultAgentKind`**
`settings.js:490-496`

The per-action overrides were correctly disabled (`agentOverrideField`, `settings.js:327-336`), but the **default those overrides fall back to** was not. That branch still does:

```js
agents.map((a) => [a.kind, a.name + (a.default ? ' (default)' : '')])
```

`selectField` (`settings.js:298-314`) has no `disabled` support, so every off-PATH agent renders as an ordinary, selectable option — no `(not installed)` suffix, no tooltip, no `o.disabled`.

Failure chain, on a machine with only Claude installed:
1. User picks "Antigravity" as Default agent → `cfg.defaultAgentKind = 'antigravity'`.
2. `set-config` (`RPCHandlers.swift:1102-1107`) decodes `AppConfig` and saves it — **no registry validation** of `defaultAgentKind` anywhere on the write path.
3. `AppConfig.agentKind(for:)` (`AppConfig.swift:224-225`) returns `.antigravity` for every new session with no per-action override.
4. `SessionService.swift:904` — `guard … let agent = AgentRegistry.shared.agent(for: session.agentKind) else { return }` → the launch **silently no-ops**.

That is precisely the "persisted-but-unlaunchable session" the `registeredKind` doc comment (`AgentRegistry.swift:53-61`) says the gate exists to prevent, reintroduced through the config path the gate doesn't cover. The ADR text this PR edits even calls out Settings → "Default agent" as the user-settable knob feeding `agentKind(for:)` — it needs the same disable treatment as the overrides.

Riding along: the `agents.length >= 2` guard at `settings.js:490` is now **always true** on the daemon (`registerAgents` registers five known agents unconditionally), so the `else` branch — the readonly field with *"Only one agent is available. Install another agent CLI (Codex, Cursor, opencode) to choose."* — is unreachable dead code. A Claude-only machine now gets a five-option dropdown where four options break new sessions.

**🔴 2. The handoff menu is the third `list-agents` consumer and was missed**
`app.js:1536-1548`

`openHandoffAgentMenu` reads the same widened payload and renders every entry as a fully enabled row:

```js
const others = agents.filter((a) => a.kind && a.kind !== session.agent_kind);
…
item.onclick = (ev) => { …; handoffAgent(session.id, a.kind); };
```

Two regressions:

- **Clickable rows that can only fail.** Server-side the gate holds — `SessionService.swift:1190` throws `agentNotRegistered` — but the browser surfaces the raw internal string: `Switch agent failed: Agent "antigravity" is not registered`. That's the opposite of the friendly *"not found on PATH — install it and restart Crow"* copy this PR introduces, on the one surface where the user is actively trying to switch harnesses.
- **The honest empty-state is now dead code.** `if (!others.length)` (`app.js:1540-1543`) with *"No other coding agents are available. Install Cursor, Codex, or OpenCode to switch."* can never fire again — `others` is always 4. Before this PR that message was the correct answer on a Claude-only box; now the user gets four broken menu items instead.

The PR body says *"a shared `agentUnavailableHint()` keeps the copy identical across both surfaces"* — this is the surface where it should have mattered most and it isn't wired up. Apply the same `enabled`/`disabled`/`title` treatment used in `openNewManagerMenu`, and keep the empty-state alive by testing `others.some(a => a.available !== false)`.

### Security Review

**Strengths:**
- **Launch gate genuinely intact.** Unavailable agents never enter `agents`, and every enforcement point keys off that map: `registeredKind` (`AgentRegistry.swift:63-66`), `handoffAgent`'s two guards (`SessionService.swift:1190-1195`), `launchAgent` (`SessionService.swift:904`), and all four `LaunchScaffold` checks (`LaunchScaffold.swift:77/116/138/166`). Verified by test, not just by inspection.
- **The right field is published.** `binary` is each adapter's compile-time `launchCommandToken` literal (`claude`/`codex`/`agent`/`opencode`/`agy`), **not** the resolved absolute path from `findBinary()`. Publishing the latter over the remote `/rpc` socket would have leaked the daemon host's filesystem layout to any authenticated browser. Good instinct.
- **No injection.** Labels go through `el()` (`app.js:694-699`, `textContent`) and hints through the `title` DOM property — no `innerHTML` anywhere in the new code.
- The Antigravity review-handoff RCE refusal (`SessionService.swift:1196-1211`) is untouched and sits *after* the registration guard, so an unavailable Antigravity is rejected strictly earlier than before.

**Concerns:** none introduced by this change. The two Red findings are correctness/UX, not privilege.

### Code Quality

**🟡 3. `allKnownAgents()` documents an invariant the code doesn't implement**
`AgentRegistry.swift:86-90`

> "De-duplicated by kind (last registration wins, so a re-register can flip availability)."

`registerKnown` never *removes* from `agents`. Re-registering an available agent with `available: false` flips only the UI list — `agent(for:)` still returns it, and `defaultAgent` may still be it. `list-agents` would then report `available: false, default: false` for a kind that is in fact launchable. Not reachable today (registration is boot-once), but it's a false invariant stated on brand-new **public** API, and it's the exact kind of thing a future caller will rely on. Either drop the kind from `agents` when `available == false`, or correct the comment to say availability is write-once per process.

**🟡 4. No CHANGELOG entry**
`CONTRIBUTING.md:113` requires a bullet under `[Unreleased]` for every PR, and every recent merge has one (#873, #868, #870, #866, #872, #867). This is a user-facing change and belongs under **Web UI**.

**🟡 5. Test coverage doesn't reach the thing that broke**
The new Swift tests exercise `AgentRegistry` in isolation, and the extended `listAgentsIsLocalNotForwarded` asserts `available`/`binary` only on the **available** Claude agent. Nothing asserts the actual point of the PR — that an *unavailable* agent appears in the `list-agents` payload with `available: false`. A router-level test that registers an unavailable agent and inspects the payload would have made the widened contract concrete, and is the kind of test that tends to prompt an audit of who consumes it (which is how findings 1 and 2 slipped through).

**🟢 6. Stale comment in `CursorAgent.swift:69-74`**
> "Cursor is only registered when it resolves, `CrowDaemon.registerAgents`"

No longer true as written — Cursor is now always registered as *known*. The safety conclusion still holds (the bare-`agent` path stays unreachable because `agent(for:)` excludes unavailable kinds), but the stated premise moved. Worth re-wording while you're here.

**🟢 7. Dead conjunct in the `default` flag**
`RPCHandlers.swift:602`, `EngineRouter.swift:490` — `known.available && known.agent.kind == defaultKind`. `defaultKind` comes from `defaultAgent?.kind`, which reads `agents[defaultKind]` and is therefore already nil for any unavailable kind. Harmless belt-and-braces; fine to keep if intentional.

**🟢 8. `knownOrder` is append-only**
`AgentRegistry.swift:16` grows on every `register()` call and is de-duplicated O(n) on each `allKnownAgents()` read. Bounded in practice (five agents, boot-once), but a `[AgentKind: Bool]` plus an order array maintained at write time would be simpler than re-deriving on every `list-agents`.

### Verification

| Check | Result |
|---|---|
| `swift build` | ✅ clean (61s) |
| `swift test --package-path Packages/CrowDaemon --filter 'AgentsLocalTests\|DaemonAgentKindGateTests'` | ✅ 4/4 passed |
| `node --check app.js settings.js` | ✅ |

### Summary Table

**Recommendation:** Request Changes — driven by [2 Red, 3 Yellow, 3 Green] findings.

The registry design is sound and the launch gate holds. What's missing is finishing the sweep: `list-agents` is a shared contract, and widening it obligates every consumer. Gate the Default-agent picker and the handoff menu the same way the manager menu and the per-action overrides were gated, and this is good to go.

---

dgershman:
Thanks — the "shared contract, finish the sweep" framing was exactly right. All findings addressed in c370f086:

**🔴 1 — Default agent picker gated.** New `defaultAgentField` (settings.js) disables off-PATH agents with the `(not installed)` suffix + tooltip, replacing the generic `selectField` that had no disable support. The dead `agents.length >= 2` branch is fixed: the readonly fallback now only fires when `list-agents` returns nothing (daemon down), with corrected copy — no more misleading "only one agent" text.

**🔴 2 — Handoff menu gated.** `openHandoffAgentMenu` now renders off-PATH agents disabled with the shared `agentUnavailableHint()` tooltip instead of clickable rows that fail with the raw `agentNotRegistered` string. The honest empty-state is kept alive via `others.some(a => a.available !== false)`, so a Claude-only box still gets the "Install Cursor/Codex/OpenCode to switch" alert rather than four broken rows.

**🟡 3 / 8 — Registry invariant made true + write-time state.** `registerKnown` now keeps availability authoritative: re-registering a kind as `available: false` also removes it from the launchable `agents` map and drops it as default, so `agent(for:)`/`defaultAgent`/`list-agents` can never disagree. Storage is a `kind → (agent, available)` map + order array maintained at write time, so `allKnownAgents()` is no longer an O(n) re-derive. Comment corrected to state availability is write-once-per-process but consistent.

**🟡 5 — Router-level test added.** `listAgentsReportsUnavailableAgents` registers a real off-PATH agent (Codex) and asserts the payload carries `available: false` + `binary: "codex"` + `default: false`, and that it stays unlaunchable (`agent(for:)`/`registeredKind` both nil). CrowCodex added to the CrowDaemonTests deps.

**🟡 4 — CHANGELOG** bullet added under Web UI.

**🟢 6 — CursorAgent comment** reworded (now always registered as *known*; launch-gate safety conclusion unchanged).

**🟢 7 — kept intentionally.** The `known.available &&` conjunct in the `default` flag stays as defense-in-depth — with the invariant now enforced, `defaultKind` is already nil for unavailable kinds, so it's belt-and-braces, but I'd rather the `list-agents` shape not depend on that coupling holding forever.

Verification: `swift build` clean; full CrowDaemon suite **228/228** (incl. the new router test); CrowCore + CrowEngine registry/smoke suites green; `app.js`/`settings.js` pass `node --check`.

dhilgaertner (changes requested):
## Code & Security Review

Verified locally: `swift build` clean; `swift test --package-path Packages/CrowCore` **484/484 green**; `swift test --package-path Packages/CrowDaemon` **228/228 green** (including the three new/extended `AgentsLocalTests`); `node --check` clean on `app.js` + `settings.js`.

### Critical Issues

None. The security-relevant part of this change is correct: unavailable agents never enter the launchable `agents` map, so `agent(for:)` / `registeredKind` / `defaultAgent` all still miss them. I traced every consumer of that gate and each one still refuses:

- `EngineRouter.swift:462` (`create-manager`) and `RPCHandlers.swift:190` (`new-session`) funnel through `registeredKind` → an unavailable kind falls back to the configured default.
- `SessionService.handoffAgent` (`SessionService.swift:1190`) throws `agentNotRegistered` before the `findBinary()` check, so a crafted `/rpc` `handoff-agent` to an off-PATH kind is rejected — the widened `list-agents` payload does not widen the attack surface.
- `default` is now computed as `known.available && kind == defaultKind` in **both** handlers, so an unavailable agent can never be advertised as the default.

### Security Review

**Strengths:**
- **No XSS.** `el()` (`app.js:694`) assigns `textContent`, and the tooltip goes through the `title` **property**, not `innerHTML`. Every new label/hint path is safe even though `a.name` / `a.binary` come off the wire.
- **Launch gate unchanged**, as claimed — confirmed by reading all four gate call sites, not just the new tests.
- **No new disclosure.** `binary` is a static `launchCommandToken` (`claude`, `agy`, …), never a resolved filesystem path, so the payload doesn't leak install locations. The set of installed harnesses was already inferable from the old payload.
- **Sound version skew handling.** `a.available !== false` treats a missing field as available, so a cached old `app.js` against a new daemon (and vice versa) degrades to the previous behavior rather than greying out everything.
- Antigravity supply-chain posture is preserved — Crow still only *resolves* `agy`, never installs it, and the review-handoff refusal (`SessionService.swift:1209`) is untouched.

**Concerns:** none found.

### Code Quality

#### 🟡 Yellow — the harness docs still describe the superseded behavior, including inside the ADR this PR edited

ADR 0014 was updated in its Decision section, but four other passages across three docs still assert the *old* contract, and one of them is 60 lines below the new text **in the same file**:

- `docs/adr/0014-pluggable-coding-agent-adapter.md:138-145` — "**A harness whose binary isn't on `PATH` is silently unavailable** in the picker and in `handoff-agent`. Because `registerAgents` gates registration on `findBinary()` at boot, such a harness is *unregistered*…" and "'why isn't Cursor in the list?' is answered by `findBinary()`, not an error message." The picker now shows it; registration is no longer gated. (The `agentNotRegistered` outcome still holds mechanically — but the stated *reason* is now wrong.)
- `docs/agent-harness-matrix.md:105-112` — "registers **Claude unconditionally first**, then Codex / Cursor / OpenCode / Antigravity **only if `findBinary()` resolves** … So a harness whose binary is off `PATH` is silently absent from the picker and from `handoff-agent`."
- `docs/agent-harness-matrix.md:423-428` — "such a harness was never registered at boot (`registerAgents` gates on `findBinary()`)".
- `docs/adr/0015-harness-capability-tiers.md:81-84` — "A harness whose `findBinary()` misses is **never registered** … the picker and `handoff-agent` act as if it doesn't exist."
- `docs/agent-harness-gap-audit.md:62` (row 8) — "`AgentRegistry` **still** registers a kind only if `findBinary()` resolves; that's the intended contract. No re-check needed."

This is worth a round trip rather than a follow-up for two repo-specific reasons. First, `CLAUDE.md` names `docs/agent-harness-matrix.md` as *the* reference for what each harness can and can't do, so a future agent reading line 108 will confidently reason from a false premise. Second, ADR 0014 itself (lines 130-132) makes this a standing commitment: *"the matrix is hand-maintained but **verified against** real property values, not aspirational — the doc's own contract is to update it in the same PR that changes a capability."* This PR changed exactly the registration contract those lines describe.

Suggested: point the matrix + ADR 0015 at the new surface-but-disable wording (a one-line "superseded by #879" pointer to ADR 0014's blockquote is enough for 0015 and the gap-audit row), and rewrite ADR 0014's Consequences bullet so the document doesn't contradict itself.

#### 🟢 Green — consider

1. **`option.title` is effectively dead on macOS.** `settings.js:336` / `settings.js:362` set `title` on `<option>` elements; Chrome and Safari render native popup menus that don't show HTML tooltips on options, so the hint will rarely (never, on macOS) appear. Nothing is lost — the ` (not installed)` suffix and the new field help carry the message — but if the tooltip is meant to be load-bearing, hang it on the `<select>` or add an inline `st-help` line instead. (The `ctx-menu` tooltips in `app.js` are plain `<div>`s and *do* work.)

2. **The hint over-promises for capability-blocked combos.** For a **review** session, handing off to Antigravity is refused by `shouldRefuseReviewHandoff` (`SessionService.swift:1209`) *regardless* of PATH — so on such a session the greyed row says "install it and restart Crow to enable", and installing it produces an enabled row that then throws `reviewNotSupported`. Pre-existing for the installed case, but this PR is what newly advertises the (non-)remedy. Capability-aware copy, or omitting Antigravity from review-session handoff menus, would close it.

3. **Claude is the one agent whose `available` flag can lie.** `CrowDaemon.swift:906` hard-codes `available: true` without probing `findBinary()`. That's clearly deliberate (it keeps Claude as the registry fallback, and `managerCommand` has its own legacy-`claude` fallback), but since the PR's whole thesis is honest install-state reporting, a one-line comment saying *why* Claude is exempt would stop the next reader from "fixing" it.

4. **`registerKnown(_:available: false)` clears the default without re-electing.** `AgentRegistry.swift:52-55` drops `defaultKind` when the removed kind was the default, but never promotes another available agent — so `defaultAgent` can be `nil` while `allAgents()` is non-empty. Unreachable in the real boot path (Claude registers first and available), but the doc comment on line 41-42 claims "the invariant holds regardless of order", and that's only true of the launchable map, not of the default. Either re-elect, or narrow the comment.

5. **`listAgentsReportsUnavailableAgents` mutates the process-wide singleton destructively.** `BoardForwarderTests.swift:367-368` registers Codex `available: false` on `AgentRegistry.shared`, which *removes* Codex from the launchable map for every later test in the CrowDaemonTests process (and the `await router.handle` gives other `@MainActor` tests a suspension point to observe it). It passes today because nothing else in that target registers Codex as available — a latent flake, not a live one. Note the two sibling tests get this right with a fresh `AgentRegistry()`; only this one can't, because the router reads `.shared`.

6. **Boot log went quiet for the unavailable case.** `registerAgents` still logs only on resolution, so the log gives no hint why a picker row is greyed. A matching `log("Antigravity agent not found on PATH")` would make the boot log and the UI agree.

**What I liked:** the `known` / `agents` split is the right shape — widening the *listing* without touching the *gate* is exactly the minimal change, and the `knownOrder` array avoids leaning on Dictionary iteration order. `agentUnavailableHint` as a single copy source across both surfaces, and the honest empty-state retained in `openHandoffAgentMenu` (`others.some(available)` rather than `others.length`), are both better than the obvious implementations. Follow-up commit c370f08 catching `defaultAgentField` — where the generic `selectField` had no disable support and would otherwise have let users persist an unlaunchable default — is a good catch; that path also incidentally fixes a pre-existing bug where an override pointing at an uninstalled agent rendered as "Use default" while the config said otherwise.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 6 Green] findings. The code is sound and I'd approve it as-is; the blocker is documentation the repo's own ADR contractually requires this PR to update. Since the Greens are cheap, folding them into the same round trip beats a follow-up.

---

dgershman:
Thanks — the "the doc's own ADR makes this a same-PR obligation" point landed. All in 0b811532.

**🟡 Docs (blocking) — updated all five stale passages:**
- **ADR 0014** Consequences bullet — rewritten to "shown in the pickers but disabled"; it no longer contradicts the *Superseded behavior* blockquote 60 lines above it in the same file.
- **agent-harness-matrix.md** — both the "Registration order = default" passage and the handoff passage now describe *known-but-unavailable / surfaced-disabled* instead of "silently absent".
- **ADR 0015 row 8** + **gap-audit row 8** — point at surface-but-disable, noting the launch gate (`findBinary`) is unchanged; the picker just shows a disabled row now.

I swept `docs/` for `silently absent | only if findBinary | never registered | gates on findBinary | acts as if it doesn't exist` afterward — the only remaining hits are the new accurate wording and the intentional "old behavior" text inside ADR 0014's superseded blockquote.

**🟢 Greens folded in:**
- **5 (destructive shared-singleton test):** extracted `AgentRegistry.agentListings() → [KnownAgentListing]` — the exact projection both `list-agents` handlers now serialize (also de-duplicates the two identical handler bodies). The test drives that helper against a **fresh** `AgentRegistry()`, so no `.shared` mutation leaks into sibling tests. Nice side effect: the contract is now defined once.
- **6 (quiet boot log):** `registerAgents` now logs `"<Agent> not found on PATH — shown disabled in the picker"` for the unavailable case, so the boot log and the greyed row agree. De-duplicated the four call sites via a local `registerDiscovered` helper.
- **3 (Claude's `available:true` can lie):** added a comment explaining the exemption is deliberate — Claude is the registry fallback default + has `managerCommand`'s legacy-`claude` fallback, and greying out the default would break the common case. "Don't fix it."
- **4 (default not re-elected):** narrowed the `registerKnown` doc comment — the *launchable-map* consistency holds regardless of order; the default is cleared, not re-elected (unreachable at boot since Claude registers first-and-available).

**🟢 Greens deferred, with reasoning:**
- **1 (`option.title` dead on macOS):** kept as progressive enhancement (it does render on Firefox/Linux), but it's not load-bearing — the inline `(not installed)` suffix and the field-level `help` carry the message on macOS. Making it truly reliable would mean a per-row inline note under a native `<select>`, which is more UI surgery than this fix warrants.
- **2 (review + Antigravity capability block):** genuinely pre-existing (the `shouldRefuseReviewHandoff` refusal is PATH-independent), and closing it needs capability data in the `list-agents` payload that isn't there today — capability-aware handoff copy is its own change. Flagging as a follow-up rather than widening this PR's scope, per the ticket's "surface + disable + tooltip" boundary.
- **7:** kept as before (defense-in-depth), now documented inline in `agentListings()`.

Verification: `swift build` clean; **CrowDaemon 228/228**; CrowCore + CrowEngine registry/smoke green; `node --check` clean on both JS files (unchanged this round).

dhilgaertner (approved):
## Code & Security Review

Reviewed all 15 files across the three commits (`4491238`, `c370f08`, `0b81153`). The "surface-but-disable" split is done correctly: the **launchable** `agents` map keeps its exact prior semantics, and a second `known` map carries the UI-only roster. No blocking issues.

### Critical Issues

None.

### Security Review

**Strengths:**

- **No install-path disclosure.** `KnownAgentListing.binary` is the static `launchCommandToken` (`claude`, `codex`, `agy`, …), never `findBinary()`'s resolved absolute path (`AgentRegistry.swift:128`, and the intent is pinned by the doc comment at `:143-145`). Since `list-agents` is reachable over the remote `/rpc` surface, that distinction matters — the widened payload adds no host-layout information. The `available` bool is also not new information: before this PR the same fact was inferable from list membership.
- **The launch gate is genuinely unchanged, not just claimed to be.** `registerKnown(_:available: false)` actively *removes* the kind from `agents` and clears `defaultKind` if it pointed there (`AgentRegistry.swift:55-58`), so `agent(for:)` / `registeredKind` / `defaultAgent` all still miss. I traced every consumer:
  - `SessionService.handoffAgent:1190` → `agent(for:)` miss → `AgentHandoffError.agentNotRegistered` → typed `applicationError`. A greyed row cannot become a handoff.
  - `EngineRouter:462` / `RPCHandlers:190` `create-manager` → `registeredKind` miss → falls back to the configured default.
  - `LaunchScaffold.swift:77,116,138,166` → `agent(for:)` gates the per-agent dev-root scaffolding, so an uninstalled harness still gets no `.cursor/` / `.codex/` / `.agents/` config written. This is the subtle one, and keeping `agents` (rather than `known`) authoritative is what preserves it.
  - `SessionService:1528` `allAgents()` (stale-hook-config reaping) also reads the launchable map — semantics preserved.
- **No injection surface in the new UI code.** `el()` sets `textContent` (`app.js:697`) and the hint is assigned via the `.title` property, so the disabled-row label and tooltip can't inject markup. The inputs are daemon-side compile-time constants regardless.
- Client-side disabling is presentation-only and is correctly backed by server-side gating in every path — the disabled rows are not the security boundary.

**Concerns:**

None found.

### Code Quality

Verified locally on the PR branch:
- `swift build` — clean (18.1s).
- `swift test --package-path Packages/CrowCore` — 484 tests / 18 suites passed.
- `swift test --package-path Packages/CrowDaemon` — 228 tests / 47 suites passed.
- `swift test --package-path Packages/CrowEngine` — 407 tests / 55 suites passed.
- `node --check` on `app.js` and `settings.js` — clean.

Correctness notes I checked and found sound:
- `registerAgents` is invoked exactly once (`CrowDaemon.swift:99`), so "boot-time snapshot" and the "restart Crow" copy are accurate.
- Claude registers first and available, so the documented "clears the default without re-electing" edge in `registerKnown` is genuinely unreachable — and the doc comment says so rather than hiding it.
- `agentListings()` takes the lock twice (via `defaultAgent`, then `allKnownAgents()`), which is non-atomic but not a deadlock and irrelevant under boot-once registration.
- Sorting by `displayName` puts the (often uninstalled) "Antigravity" first in the Settings `<select>`. That's safe: HTML's select-reset algorithm skips disabled options, so with `defaultAgentKind` unset the browser lands on the first *enabled* option (Claude Code), not the greyed row. A persisted-but-unavailable kind still renders selected-and-disabled, which is the honest display.
- `.ctx-item.disabled` / `.ctx-item.disabled:hover` (`app.css:777-778`) win on both specificity and source order over the base `.ctx-item:hover`.
- All `list-agents` consumers are covered — I grepped the repo: the three web surfaces (new-manager menu, handoff menu, Settings), and nothing in `Sources/`, `Packages/CrowCLI`, or `crow-desktop` consumes it.

Green findings (consider, non-blocking):

1. **`app.js:1083`** — `if (agents.length < 2) { createManager(agents[0] && agents[0].kind); return; }` doesn't filter on availability, so a lone *unavailable* agent would instant-create with an unlaunchable kind. It degrades safely (server-side `registeredKind` falls back to the configured default) and the comment acknowledges the path, but filtering to launchable agents here would mirror the handoff menu's `others.some((a) => a.available !== false)` check and remove the asymmetry.
2. **`app.js:1546`** — the empty-state copy still reads "Install Cursor, Codex, or OpenCode to switch." Antigravity is now a first-class roster entry; this is the one user-facing string that still omits it.
3. **`app.js:1537`** — comment says handing off to an off-PATH agent "only fails server-side with a raw internal error." It actually throws a typed `AgentHandoffError.agentNotRegistered`, surfaced as an `applicationError` with a localized description (`EngineRouter.swift:565-566`, `RPCHandlers.swift:349-350`). Worth correcting so the comment doesn't understate the existing handling.
4. **`app.js:1067`** — `agentUnavailableHint` always says "not found on PATH". `findBinary()` also falls through a *stale* `defaults.binaries.<kind>` override (`CodingAgent.swift:128-134`), so a user whose override points at a moved binary gets a hint aimed at the wrong knob. Mentioning the override in the copy would cover both causes.
5. **`AgentRegistry.swift:120-134`** — `agentListings()` shadows two instance members: the local `let defaultKind` and the closure parameter `known`. Correct as written and the build confirms it, but in a file where both names are stored properties it costs a double-take. `defaultKindNow` / `entry` would read cleaner.
6. **Test coverage** — the new tests cover the registry primitive and the RPC field shape well, and using a fresh `AgentRegistry()` instead of the shared singleton is the right call. What's untested is `CrowDaemon.registerAgents` itself, i.e. the function whose `findBinary()`-gated registration *was* the #879 bug. Deleting a `registerDiscovered(...)` line would silently restore the original defect with every test still green. It's `@MainActor private static` and mutates a process-wide singleton, so I understand why it was skipped — but a smoke test asserting all five kinds land in `allKnownAgents()` would guard the actual regression.

Docs are unusually good here: ADR 0014 supersedes the old behavior with a blockquote rather than a silent rewrite, and ADR 0015 / the harness matrix / the gap audit were all updated to match. That's the right way to do it.

### Summary Table

PR #881, opened 2026-07-28
Cmd+V no longer pastes twice in the terminal

### LINKED ISSUE #875: Cmd+V pastes twice in the terminal — custom key handler returns false without preventDefault, so xterm's native paste also fires
## Symptom

Pasting into a coding-agent terminal with **Cmd+V** inserts the clipboard **twice** (double paste). Pasting via the right-click **"Paste"** menu item works correctly (single paste).

## Root cause

Two independent code paths both deliver the clipboard to the PTY on a Cmd+V keydown.

In `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js`, the terminal's custom key handler (`ensureTerminal` → `attachCustomKeyEventHandler`, ~:3452-3467) handles Cmd+V itself:

```js
// app.js ~:3459
if (e.metaKey && (e.key === 'v' || e.key === 'V')) { pasteIntoTerminal(); return false; }
```

`pasteIntoTerminal()` (~:3660) reads the clipboard and pastes it:

```js
navigator.clipboard.readText().then((text) => { if (text) term.paste(text); });   // paste #1
```

The handler returns `false`, but **`false` from xterm.js's `attachCustomKeyEventHandler` does not `preventDefault()` the event** — it only makes xterm skip its *own* keydown processing (xterm's `_keyDown` returns early on a `false` result, *before* reaching any `cancel()`/`preventDefault()`). So the browser's native Cmd+V paste gesture still proceeds and fires **xterm.js's built-in `paste` DOM handler** on the helper textarea, which writes the clipboard through `onData` to the PTY **a second time** (paste #2).

Result: `pasteIntoTerminal()` + xterm's native paste both run → the clipboard lands twice.

**Why only Cmd+V doubles:** the right-click "Paste" menu item (~:3680) calls `pasteIntoTerminal()` directly with no keydown, so xterm's native paste never fires — single paste. This matches the report (Cmd+V specifically). (Both paths route through `term.paste()`, so bracketed-paste wrapping is applied twice too.)

## Fix direction

- In the Cmd+V branch, **`e.preventDefault()` (and `e.stopPropagation()`) before `return false`**, so the browser's native paste gesture is suppressed and only `pasteIntoTerminal()` runs. Minimal and correct; keeps the intended explicit path (non-secure-context clipboard + the shared right-click "Paste" code).
- *Alternative:* drop the custom Cmd+V branch entirely and rely on xterm's native paste (it already bracketed-paste-wraps). Rejected unless we also want to drop `pasteIntoTerminal()` for the menu — the `preventDefault` fix is smaller and preserves both entry points.
- **General pattern:** any custom key handler that shadows a browser default and returns `false` must also `preventDefault()`. The sibling branches are latent-but-benign: **Cmd+C** (~:3455) is idempotent (a second OS copy has no user-visible effect) and **Cmd+F** (~:3460) has no browser default — neither manifests, but adding `preventDefault` there is consistent and future-proof.
- **Check the other surface:** `Packages/CrowTerminal/Sources/CrowTerminal/Resources/xterm/terminal.html` has its own `attachCustomKeyEventHandler`; if it handles Cmd+V the same way, apply the same fix (this surface may be retired per Crow ADR 0010 — confirm before touching).

## Acceptance criteria
- Cmd+V pastes the clipboard exactly **once**.
- Right-click "Paste" still works (once).
- Bracketed-paste wrapping still applied exactly once (multi-line clipboard doesn't auto-execute in a shell, and isn't wrapped twice).
- Ctrl+C SIGINT (no selection) and Cmd+C copy are unaffected.

## Repro
Focus a coding-agent terminal in the Crow web UI, copy any text, press **Cmd+V** → the text appears twice.

## References
- Code: `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` — `attachCustomKeyEventHandler` (~:3452-3467, Cmd+V at ~:3459), `pasteIntoTerminal` (~:3660), right-click "Paste" (~:3680); `term.onData(sendToPTY)` (~:3448).
- xterm.js: `attachCustomKeyEventHandler` returning `false` skips terminal handling but does not preventDefault; native paste is handled on the helper textarea independently.

dhilgaertner:
Closes #875

## The bug

Cmd+V in a web-UI terminal inserted the clipboard **twice**; right-click "Paste" was fine.

Two paths ran on one keydown: the custom key handler called `pasteIntoTerminal()`, *and* the browser's native paste gesture fired xterm's own `paste` listener (registered on the helper textarea and the element), which pasted again — bracketed-paste wrapping applied twice with it. Returning `false` from `attachCustomKeyEventHandler` does not cancel the event: xterm's `_keyDown` returns early, before any `cancel()`. Right-click Paste never doubled because a click is not a paste gesture.

## The fix — leave Cmd+V to the browser

The ticket proposed `preventDefault()`. That would have **deleted paste entirely** for anyone on a plain-http daemon (`--host 0.0.0.0`, documented in `docs/getting-started.md`): there is no `navigator.clipboard` in a non-secure context, so `pasteIntoTerminal()` already returns at its guard and the native paste is the only working path. `copyToClipboard` carries an `execCommand` fallback for exactly that context.

So the Cmd+V branch is dropped instead. xterm's listener already routes the clipboard through `paste()` — wrapped once, then `onData` → `sendToPTY` — and it needs no clipboard-read permission and works in every browser and security context. Verified in the vendored xterm.js 6.0.0 that macOS Cmd+V (`isMac` + meta, no ctrl/alt/shift) matches only keyCode 65 and leaves `result.key`/`result.cancel` unset, so nothing leaks to the PTY and nothing is cancelled. `pasteIntoTerminal()` stays for the right-click menu, now its only caller.

Sibling branches keep their explicit paths and now cancel properly:

- **Cmd+C** — safe to own: `copyToClipboard` always delivers. Ctrl+C with no selection still falls through as SIGINT.
- **Cmd+F** — the browser's find bar no longer opens over ours.

`Resources/web/terminal.html` (standalone debug page) carried the same shape and gets the same treatment.

## Tests

- `web-tests/key-handler.test.js` (new, jsdom) drives the real handler — hoisted to module scope as `handleTerminalKey` for that, the same precedent as `swallowMouseMode`. 27 assertions: Cmd+V untouched and unprevented, menu paste exactly once, degrading without throwing where `navigator.clipboard` is absent, Cmd/Ctrl+C copy, Ctrl+C SIGINT, Cmd+F, pass-through.
- `WebTerminalAssetTests` gains a drift guard over both surfaces.

Both were checked to **fail** against the old shape before being accepted.

```
226 tests in 47 suites passed   # swift test --package-path Packages/CrowDaemon
27 passed, 0 failed             # node key-handler.test.js
```

Pre-existing and unrelated: `web-tests/row.test.js` has 11 failures on a clean tree (confirmed by stashing).

## Manual verification checklist

- [ ] Cmd+V pastes exactly once
- [ ] Multi-line clipboard arrives as one bracketed paste (no auto-execute, not double-wrapped)
- [ ] Right-click Paste still pastes once
- [ ] Cmd+C copies; Ctrl+C with no selection still interrupts
- [ ] Cmd+F opens Crow's Find prompt with no browser find bar behind it
- [ ] Cmd+V still pastes over `http://<LAN-IP>:8787` (the case that ruled out `preventDefault`)

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -3600,25 +3600,7 @@ function ensureTerminal() {
     term.loadAddon(webglAddon);
   } catch (_) { /* WebGL unavailable → canvas/DOM renderer */ }
   term.onData(sendToPTY);
-  // Cmd/Ctrl+C copies the selection (falling through to SIGINT when nothing is
-  // selected so Ctrl+C still interrupts); Cmd+V pastes. Lets the browser own
-  // copy/paste instead of tmux's copy-mode.
-  term.attachCustomKeyEventHandler((e) => {
-    if (e.type !== 'keydown') return true;
-    const mod = e.metaKey || e.ctrlKey;
-    if (mod && (e.key === 'c' || e.key === 'C') && term.hasSelection()) {
-      copyToClipboard(term.getSelection());
-      return false;
-    }
-    if (e.metaKey && (e.key === 'v' || e.key === 'V')) { pasteIntoTerminal(); return false; }
-    if (e.metaKey && (e.key === 'f' || e.key === 'F')) {
-      textPrompt('Find in terminal', '', { okLabel: 'Find' }).then((q) => {
-        if (q && searchAddon) { try { searchAddon.findNext(q); } catch (_) {} }
-      });
-      return false;
-    }
-    return true;
-  });
+  term.attachCustomKeyEventHandler(handleTerminalKey);
   enableTouchScroll(document.getElementById('terminal'));
   enableWheelScroll(document.getElementById('terminal'));
   enableFileDrop(document.getElementById('terminal'));
@@ -3846,8 +3828,56 @@ async function uploadDroppedFiles(files) {
   }
 }
 
+// The terminal's custom key handler (attached in ensureTerminal). At module
+// scope so the jsdom suite can drive it directly — same reason swallowMouseMode
+// lives out here.
+//
+// Cmd/Ctrl+C copies the selection (falling through to SIGINT when nothing is
+// selected so Ctrl+C still interrupts), Cmd+F opens our find prompt. Lets the
+// browser own copy instead of tmux's copy-mode.
+//
+// A branch that shadows a browser default must preventDefault(): returning
+// false only makes xterm skip its OWN key handling (its _keyDown returns early,
+// before any cancel()), so the browser's default gesture still runs (#875).
+function handleTerminalKey(e) {
+  if (e.type !== 'keydown') return true;
+  const mod = e.metaKey || e.ctrlKey;
+  if (mod && (e.key === 'c' || e.key === 'C') && term.hasSelection()) {
+    copyToClipboard(term.getSelection());
+    // Safe to own the gesture: copyToClipboard always delivers — writeText, or
+    // fallbackCopy's execCommand in a non-secure context.
+    e.preventDefault();
+    e.stopPropagation();
+    return false;
+  }
+  // #875: Cmd+V is deliberately NOT handled here. Pasting explicitly *and*
+  // returning false double-pasted, because the browser's native paste gesture
+  // still fired xterm's own `paste` listener (on the helper textarea) and wrote
+  // the clipboard a second time. Leaving the event alone gives exactly one
+  // paste — xterm's listener routes it through paste(), so bracketed-paste
+  // wrapping is applied once, then onData → sendToPTY. macOS Cmd+V produces no
+  // key data in xterm, so nothing leaks to the PTY either.
+  //
+  // The native gesture is also strictly more capable than reading the clipboard
+  // ourselves: it needs no clipboard-read permission and works over plain http
+  // (`--host 0.0.0.0`), where navigator.clipboard doesn't exist at all. The
+  // right-click menu still calls pasteIntoTerminal() — a click is not a paste
+  // gesture, so there it's the only path.
+  if (e.metaKey && (e.key === 'f' || e.key === 'F')) {
+    // Without this the browser's own find bar opens over our prompt.
+    e.preventDefault();
+    e.stopPropagation();
+    textPrompt('Find in terminal', '', { okLabel: 'Find' }).then((q) => {
+      if (q && searchAddon) { try { searchAddon.findNext(q); } catch (_) {} }
+    });
+    return false;
+  }
+  return true;
+}
+
 // Paste the browser clipboard into the terminal (writes to the PTY, same path
-// as typing). readText() needs a user gesture — the menu click / Cmd+V is one.
+// as typing). readText() needs a user gesture — the right-click menu click is
+// one. Cmd+V never comes through here (#875); the browser pastes it natively.
 function pasteIntoTerminal() {
   if (!term || !(navigator.clipboard && navigator.clipboard.readText)) return;
   navigator.clipboard.readText().then((text) => {
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/terminal.html
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/terminal.html
@@ -119,9 +119,33 @@
       // through) when there's no selection.
       function copySelection() {
         const sel = term.getSelection();
-        if (sel) { navigator.clipboard.writeText(sel).catch(function () {}); }
+        if (!sel) return;
+        // The key handler cancels the browser's own copy, so this must always
+        // deliver — mirrors app.js's copyToClipboard/fallbackCopy pair. Over
+        // plain http (a `--host 0.0.0.0` daemon) navigator.clipboard is absent
+        // entirely, and writeText can still reject where it exists; without the
+        // legacy fallback either case would be a silent no-op (review).
+        if (navigator.clipboard && navigator.clipboard.writeText) {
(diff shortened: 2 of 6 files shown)


dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Leaving Cmd+V to the browser is the right security/capability tradeoff: no `clipboard-read` permission, and paste still works over plain HTTP where `navigator.clipboard` is absent.
- Clipboard content still routes through xterm `paste()` (native listener / menu path), so bracketed-paste wrapping remains and hidden newlines are not auto-executed.
- No new auth, injection, or secrets surface; changes are confined to key-event ownership.

**Concerns:**
- None security-critical.

### Code Quality
- **Yellow — `terminal.html` cancels copy without a fallback** (`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/terminal.html` ~120–140): `preventDefault()` / `stopPropagation()` were added on the copy chord, but `copySelection()` only calls `navigator.clipboard.writeText` (no `execCommand` fallback). Over plain HTTP — the same context this PR carefully preserves for paste — `navigator.clipboard` is missing, so the write throws/no-ops **and** the browser default is cancelled. Cmd+C / Ctrl+Shift+C with a selection becomes a silent no-op on the debug page. `app.js` is safe because `copyToClipboard` → `fallbackCopy`; mirror that (or don’t cancel until copy always delivers).
- **Green — misleading assertion label** (`web-tests/key-handler.test.js` ~123–129): the case named “Cmd+Shift+V” only passes `key: 'V'` with `meta: true` (no `shiftKey`). Consider renaming or also asserting the real Shift chord.
- **Green — drift guard comment stripping** (`WebTerminalAssetTests.stripLineComments`): only strips `//` line comments. Fine for today’s handler; a `/* … */` mention of `e.key === 'v'` could slip past. Not blocking.

### Tests
- `node key-handler.test.js`: **27 passed, 0 failed**
- `swift test --filter WebTerminalAssetTests`: all 6 tests in the suite passed (process exit 1 was an unrelated host arch `xctest` load warning, not a test failure)

### Summary Table

---

dhilgaertner:
Thanks — all three addressed in 33d425d.

**Yellow — `terminal.html` cancels copy without a fallback.** Fixed, and you were right about the shape of the risk even though the missing-`navigator.clipboard` case failed differently than described: there `copySelection()` throws a `TypeError` *before* reaching `preventDefault()`, so the native copy still ran by accident. The `writeText`-rejects case was a genuine silent no-op, and depending on a thrown exception to preserve copy is not a property worth keeping either way. `copySelection` now mirrors app.js's `copyToClipboard`/`fallbackCopy` pair, so the cancel is backed by a copy that always delivers.

While in there, `pasteClipboard()` got app.js's `readText` guard — the right-click menu is now its only caller, and unguarded it throws out of the click over plain http. Reads get a guard rather than a fallback because there is no `execCommand('paste')` for a web page.

That pairing is now pinned by a new drift guard over both surfaces (`clipboardWritesAlwaysDeliverAndReadsAreGuarded`): writes must feature-detect and keep the `execCommand` fallback, reads must be guarded. Checked to fail with `fallbackCopy` removed.

**Green — misleading assertion label.** The `key()` helper takes `shift` now, and both chords are asserted: `Cmd+Shift+V` with a real `shiftKey`, plus the caps-locked `Cmd+V` (uppercase `key`, no Shift) that the old handler also matched.

**Green — comment stripping.** Handles `/* … */` as well as `//`. Only *closed* block comments are dropped: an unterminated `/*` is left in place, so the failure mode is a guard failing loudly rather than passing silently.

On the `xctest` exit code you hit — that's environmental, not the suite. Clean here:

```
227 tests in 47 suites passed   # swift test --package-path Packages/CrowDaemon
29 passed, 0 failed             # node key-handler.test.js
```

dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Leaving Cmd+V to the browser is the right tradeoff: no `clipboard-read` permission, and paste still works over plain HTTP where `navigator.clipboard` is absent (`app.js` `handleTerminalKey`, `terminal.html` key handler).
- Clipboard content still routes through xterm `paste()` (native listener / right-click `pasteIntoTerminal` / `pasteClipboard`), so bracketed-paste wrapping remains and hidden newlines are not auto-executed.
- Copy chords that call `preventDefault()` now have a delivering path on both surfaces (`copyToClipboard`/`fallbackCopy` in `app.js`; matching pair in `terminal.html`) — the prior review's silent-no-op hole is closed.
- No new auth, injection, or secrets surface; changes are confined to key-event ownership.

**Concerns:**
- None security-critical.

### Code Quality
- **Green — no runtime coverage for copy without clipboard API** (`web-tests/key-handler.test.js`): `withoutClipboard()` is exercised for menu paste, but Cmd/Ctrl+C after `preventDefault` is only asserted with a live `navigator.clipboard`. The Swift drift guard pins `execCommand('copy')` in source; a jsdom case that cancels the chord with clipboard absent would lock the “always delivers” claim at runtime too.
- **Green — `stripComments` still treats `//` inside strings as comments** (`WebTerminalAssetTests.swift`): fine for today’s handler body; a future URL literal in that function could weaken a negative assertion.

### Tests
- `node key-handler.test.js`: **29 passed, 0 failed**
- `swift test --filter WebTerminalAssetTests`: all tests in the suite passed (process exit 1 was an unrelated host arch `xctest` load warning, not a test failure)

### Summary Table

---

PR #882, opened 2026-07-28
Generate the CLI reference from ArgumentParser (CROW-808)

### LINKED ISSUE #808: [parity] Generate CLI reference docs from ArgumentParser (stop doc drift)
Foundation ticket for milestone **CLI control-plane parity**.

Generate the CLI reference docs from ArgumentParser metadata so docs can't silently drift from the commands. Today the CLI reference is hand-maintained in `CLAUDE.md` (and duplicated in the crow-workspace skill), which is exactly where drift hides.

- Emit a canonical CLI reference (e.g. `docs/cli.md` and/or the `CLAUDE.md` section) from `CrowCommand` help (`abstract`/`discussion`, `@Option`/`@Flag` help).
- Add a test asserting every subcommand + flag appears in the generated doc (no undocumented verbs).
- Pairs with the parity test (#parity-test): parity guarantees the verb exists; doc-gen guarantees it's documented.

dhilgaertner:
Closes #808

The `crow` CLI surface was hand-documented in several places with nothing tying those docs to `CrowCommand`, so verbs shipped undocumented and nobody noticed.

## The drift was already real

Measured against the 82 real invocation paths, `docs/cli-reference.md` was missing 14 — the entire `crow job` group (9 subcommands), `set-locked`, `set-pinned`, `recreate-terminal`, `resync-jira`, `codex-notify`. `CLAUDE.md` was missing 16. And `crow transition-ticket` had nothing but a cross-reference link pointing at an anchor that did not exist.

## How it works

`CLIDocs` renders `docs/cli.md` from the command tree itself, stitching two sources deliberately:

- The **command layer** (name, abstract, discussion, hidden-ness, nesting) comes from `CommandConfiguration` — stable public API.
- The **argument layer** is decoded from the JSON `ParsableCommand._dumpHelp()` emits. There is no public API for walking a command's `@Option`/`@Flag` declarations, so this is the only way to get flag help text without a macro or reflection hack.

The JSON is decoded into local types rather than by importing `ArgumentParserToolInfo`, because that module is not a library product of swift-argument-parser. Only fields present in 1.5.0 (the manifest's floor) are read — `Package.resolved` is not committed, so any 1.x may be resolved and the generated file has to stay byte-identical across them.

`crow generate-docs` (hidden, local-only) writes the file, `make docs` is the front door, `--check` verifies without writing.

## The guard is a test, not a binary invocation

CI's Linux lane builds only `Packages/CrowCLI` and never links the `crow` executable, so everything runs in-process off the command tree. `CrowCLI` is already in `LINUX_PACKAGES` — **no CI changes needed**.

Four tests in `CrowCLITests`:

1. **Every command and flag appears in the generated reference.** Flags are re-derived by dumping each command type *on its own* rather than reusing the whole-tree dump the generator indexes by path — that independent second pass is the point, since a generator that mis-joins a nested path (`crow job add` vs `crow add`) would still look self-consistent against its own index.
2. **The committed `docs/cli.md` matches what the generator produces** — fails with "run `make docs`".
3. **Every verb has a worked example in `docs/cli-reference.md`.**
4. **Every user-facing verb appears in `CLAUDE.md`'s CLI Reference** — the Manager agent reads it to learn what it can run.

Tests 3 and 4 match against **fenced code blocks, not the whole file**: "documented" should mean "there is a command here you can copy and run". A whole-file substring match is exactly what let `transition-ticket`'s dead link pass as documented. Each carries an exemption list where every entry states its reason.

## Docs

`docs/cli-reference.md` stays hand-written and keeps its anchors — `#gateway-commands`, `#maintenance-commands`, `#settings-commands` and `#notification-commands` are linked from `README.md` (x4), `docs/getting-started.md`, `docs/configuration.md` (x2) and a comment in `EngineRouter.swift`. `docs/cli.md` is the exhaustive generated surface; the two cross-link.

All of today's drift is backfilled so the exemption lists ship near-empty: a Job Commands section in both docs, plus `set-locked`/`set-pinned`, `transition-ticket` (repairing the dead anchor), `resync-jira`, `recreate-terminal`, `rename-terminal` and `codex-notify`.

## Verification

- `swift test --package-path Packages/CrowCLI` — 241/241 pass
- `make docs && git diff --exit-code docs/cli.md` — generator is idempotent
- `crow generate-docs --check` — exits 0
- Negative checks, each confirmed to fail and then reverted: a new undocumented verb fails tests 2-4; a hand-edited `docs/cli.md` fails test 2 with the "run `make docs`" message; deleting the `crow job add` section fails test 3

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -23,6 +23,7 @@ crow select-session --session <uuid>            → {"session_id":"..."}
 crow list-sessions                              → {"sessions":[...]}
 crow get-session --session <uuid>               → {id, name, status, ticket_url, ...}
 crow set-status --session <uuid> active|paused|inReview|completed|archived
+crow set-locked --session <uuid> true|false     → exempt a session from (or return it to) the retention reaper
 crow handoff-agent --session <uuid> --agent cursor [--note "..."] → {"session_id":"...","agent_kind":"...","terminal_id":"..."}
 crow delete-session --session <uuid>            → {"deleted":true}
 ```
@@ -105,6 +106,26 @@ Everything is live except `--binary` (agent discovery + `.claude/bin` symlinks a
 
 The review board filters on defaults ∪ every workspace's own `excludeReviewRepos`, so `--clear-exclude-review-repos` won't unhide a repo a workspace excludes.
 
+### Job Commands
+
+Scheduled prompt-sets scoped to one repo in a workspace (CROW-604) — the Jobs sidebar, as CLI verbs. Jobs are addressed by UUID; `crow job list` prints them. Mutations hit the app's live config, so the scheduler and Settings UI see them immediately.
+
+```
+crow job list                                   → {"jobs":[...]}
+crow job get --id <job-uuid>                    → {"job":{...}}
+crow job add --name "..." --workspace "..." --repo owner/repo --prompt "..." (--interval-seconds N | --daily-at HH:MM [--weekdays mon,tue])
+crow job edit --id <job-uuid> [--name ...] [--prompt ...] [--daily-at HH:MM] [--weekdays ...]
+crow job enable --id <job-uuid>                 → {"job":{...}}
+crow job disable --id <job-uuid>                → {"job":{...}}
+crow job run --id <job-uuid>                    → {"job_id":"...","session_id":"...","terminal_id":"..."}   needs tmux; ignores schedule + enabled
+crow job delete --id <job-uuid>                 → {"deleted":true,"job_id":"..."}
+crow job duplicate --id <job-uuid>              → {"job":{...}}   the copy starts disabled with a uniquified name
+```
+
+- `add` needs exactly one schedule (`--interval-seconds` **or** `--daily-at`) and at least one `--prompt`/`--prompt-file`. `--prompt` and `--prompt-file` are repeatable and sent in that order; `--prompt-file -` reads stdin (at most once).
+- On `edit`, any `--prompt` replaces the **whole** prompt list and any schedule flag replaces the **whole** schedule — so changing `--weekdays` means restating `--daily-at`. Use `enable`/`disable` instead of `edit` to toggle enabled.
+- `job run` can take a while on first run (it may clone the repo); the run continues in the app even if the CLI stops waiting.
+
 ### Worktree Commands
 ```
 crow add-worktree --session <uuid> --repo "name" --repo-path "/main/repo" --path "/worktree/path" --branch "feature/..." [--primary]
@@ -116,6 +137,8 @@ crow list-worktrees --session <uuid>
 crow new-terminal --session <uuid> --cwd "/path" [--name "Claude Code"] [--command "claude ..."] [--managed]
 crow list-terminals --session <uuid>
 crow close-terminal --session <uuid> --terminal <uuid>
+crow rename-terminal --session <uuid> --terminal <uuid> "new name"
+crow recreate-terminal --session <uuid> --terminal <uuid>   → DESTRUCTIVE: rebuilds the pane to restore scrollback; relaunches the agent with --continue
 crow send --session <uuid> --terminal <uuid> "text to send"
 ```
 
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build daemon app run setup install uninstall clean check test help daemon-run
+.PHONY: build daemon app run setup install uninstall clean check test help daemon-run docs
 
 # Install destination and build config (override on the command line, e.g.
 # `make install BINDIR=/usr/local/bin` or `make build CONFIG=release`).
@@ -53,6 +53,7 @@ help:
 	@echo "  setup      Check build prerequisites"
 	@echo "  check      Verify all build and runtime prerequisites"
 	@echo "  test       Run all package tests"
+	@echo "  docs       Regenerate docs/cli.md from the CLI's ArgumentParser metadata"
 	@echo "  install    Symlink crow + crowd into ~/.local/bin (override BINDIR=, CONFIG=release)"
 	@echo "  uninstall  Remove installed crow + crowd symlinks"
 	@echo "  clean      Remove .build/ and the desktop app's target/"
@@ -97,6 +98,14 @@ test:
 	done
 	@echo "==> Checking notification-event catalogs (CROW-768)..."
 	@./scripts/check-notification-events.sh
+
+# Regenerate the CLI reference from the commands themselves (CROW-808). Run
+# after adding or changing a subcommand — a stale docs/cli.md fails CrowCLI's
+# test suite, which is what keeps the docs honest.
+docs:
+	@bash scripts/generate-build-info.sh
+	@swift build $(if $(filter release,$(CONFIG)),-c release,) --product crow
+	@$(BUILD_OUT)/crow generate-docs
 
 clean:
 	rm -rf .build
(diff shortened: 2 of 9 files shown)


dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- `crow generate-docs` is local-only (no socket/RPC), hidden from `--help`, and writes only UTF-8 Markdown — no secrets, no remote surface, no command injection.
- Dump JSON is decoded into local optional structs rather than trusting a newer ArgumentParser schema; unknown argument kinds are dropped instead of crashing.
- Table cells escape `|` / newlines so help text like `active|paused|…` cannot break Markdown table structure.

**Concerns:**
- None material. `--output` can write any path the invoking user can write; that matches normal local-CLI trust and is not a privilege escalation.

### Code Quality
- **Yellow** — `longFlagNames(of:)` in `Packages/CrowCLI/Tests/CrowCLITests/CLIDocsTests.swift:255-257` uses `try?` and returns `[]` on decode failure. That makes the independent second dump (the stated point of `generatedDocCoversEveryCommandAndFlag`) fail open: a `_dumpHelp()` shape change can skip every flag assertion while the test still passes. Prefer `#require` / `try` so a bad dump fails the test, matching the generator path that already `throws`.
- **Green** — Hand-written doc checks use `examples.contains(invocation)`, so parent groups (`crow job`, `crow gateway`, …) can pass solely because a child example contains the same prefix. Consider token/line-bounded matching if those parents should be required to appear as copyable invocations.
- **Green** — Relies on private `ParsableCommand._dumpHelp()`. The optional decoding + awkward-shape smoke tests mitigate breakage; still worth watching ArgumentParser minors.
- Generator design is otherwise strong: alphabetical sections, inversion merging, independent tree walk vs whole-tree index, exemption lists with reasons, and `committedGeneratedDocIsUpToDate` keeps `docs/cli.md` honest.
- Static analysis: CrowCLI tests 241/241 pass locally; CI Build & Test / Desktop / Shell Lint are green.

### Summary Table

---

dhilgaertner:
Thanks — the Yellow was a real hole, and fixing it immediately caught a bad assumption of mine. Addressed in 500be15.

### Yellow — `longFlagNames` failing open (fixed)

You were right, and the hole went one level deeper than the `try?`: `arguments` and `names` were *also* decoded as optional, so a renamed key would have produced the same silent `[]` even with `try`.

It now `throws`, and `arguments` is decoded as **required** — ArgumentParser appends a help flag to every command, so that key is always present. `names` had to stay optional: making it required failed instantly because **positional arguments have no `names` key**. That was my incorrect assumption, surfaced by the very strictness you asked for.

Failing closed alone still wasn't enough, though — a dump that decodes cleanly but yields nothing would leave every assertion vacuous and still pass. So the test also pins `crow job add`'s flags exactly and floors the number of assertions actually made:

```swift
#expect(asserted > 100, "only \(asserted) flags were checked — the independent dump is failing open")
```

The real figure is **147**; I measured it by temporarily raising the floor. Verified both directions: renaming the `arguments` key now fails with `DecodingError.keyNotFound` instead of passing.

### Green — prefix matching on parent groups (adopted)

Adopted. `contains` is replaced by `mentionsCommand`, which requires a whole command token — the match must start a token and must not be followed by a letter, digit, or hyphen. Without it a future `crow set-status-bulk` example would satisfy `crow set-status`, which is the sharper version of the risk you identified.

On parent groups specifically I went the other way, deliberately: `crow job` is still satisfied by `crow job list`. A bare `crow job` has no `defaultSubcommand` and only prints help, so requiring a standalone example would add a line nobody would ever run. A group is documented when its subcommands are. Happy to reverse this if you disagree — it's a one-line change to require groups to match to end-of-line.

Since `mentionsCommand` now decides whether both hand-written checks pass, its boundary behaviour is pinned by its own test rather than trusted by inspection.

### Green — reliance on `_dumpHelp()`

Noted, and the Yellow fix strengthens exactly this: a shape change in an ArgumentParser minor is now a loud `DecodingError` in CI rather than a silently-skipped assertion. Worth adding that the generator only reads fields present in 1.5.0 (the manifest floor), since `Package.resolved` is not committed and the output has to stay byte-identical across whatever 1.x resolves.

CrowCLI is 242/242 locally (+1 for the new boundary test).

dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- `crow generate-docs` is purely local (no socket/RPC), so it cannot be driven over remote `/rpc` the way secret/gateway methods can.
- No secrets, credentials, or auth surface in the generator; output is ArgumentParser help metadata only.
- Atomic UTF-8 write; `--check` is read-only and fails closed on missing/stale files.
- Drift tests fail closed on `_dumpHelp()` shape changes (required `arguments` in the independent pass + positive control on `JobAdd` flags).

**Concerns:**
- None material. `--output` can write any path the invoking user can write — expected for a local repo-maintenance CLI, not a remote trust boundary.

### Code Quality
- Generator design is sound: `CommandConfiguration` for the tree, `_dumpHelp()` JSON for args, local Decodable mirrors (not `ArgumentParserToolInfo`), optional fields pinned to the 1.5.0 floor.
- Strong guard suite: independent per-command dump vs whole-tree index, committed-file freshness, fenced-example checks for `cli-reference.md` / `CLAUDE.md` with reasoned exemptions and token-boundary matching.
- Awkward shapes covered (`@OptionGroup`, `.customLong`, repeatables, positionals, `.prefixedNo` merges, hidden commands).
- Docs backfill (jobs, `set-locked`, `transition-ticket`, `resync-jira`, recreate/rename terminal, `codex-notify`) matches the new tests.

### Summary Table

**Findings:**
- Green — `CLIDocs.swift` appends `.` to abstracts without `sentenceTerminated`; fine today (no abstracts end in punctuation) but brittle if one gains a trailing period.
- Green — Generated **Required** column is ArgumentParser optionality, not `validate()` semantics (`--prompt` / `--header` show as optional while discussion correctly requires them). Consider a one-line preamble note; the hand-written guide already owns those gotchas.
- Green — Relies on private `_dumpHelp()`; mitigations (optional decode + fail-closed tests) are good — keep watching ArgumentParser upgrades.

**Static analysis:** `swift test --package-path Packages/CrowCLI` — 242/242 Swift Testing tests passed (host also emitted an unrelated arm64/x86_64 XCTest bundle architecture warning; not a PR failure).

---

dhilgaertner:
Rebased onto `main` (picking up #880 and #887). Force-pushed as `6b9507a`.

**The conflict itself was trivial** — #887 and this branch each added a section to `CLAUDE.md` right after Settings Commands. Kept both: Settings → Defaults → Job Commands → Worktree.

**More interesting: #887 landed three new verbs, and this PR's own guard caught them.**

```
✘ committedGeneratedDocIsUpToDate()
↳ docs/cli.md is stale — run `make docs` and commit the result
```

`make docs` absorbed `crow defaults` / `defaults get` / `defaults set` — full flag tables for all 13 flags including the repeatable `--add-*`/`--remove-*` list editors, plus the `discussion` prose about `--binary` needing a restart. No manual writing.

Worth noting the two hand-written checks **passed** on the rebase: #887 had already documented `crow defaults` in both `docs/cli-reference.md` and `CLAUDE.md`, so the guard correctly flagged only the generated file. That is the intended split — prose stays a human judgement, the exhaustive surface regenerates.

The regeneration is squashed into the commit that introduces `docs/cli.md` rather than added on top, so no commit in the branch carries a generated file that fails its own test.

### Verification after rebase

- CrowCLI 272/272 (up from 242 — #887 brought 30 tests)
- CrowEngine 432/432, CrowDaemon 245/245, CrowCore 484/484 — the packages #887 touched
- `make docs && git diff --exit-code docs/cli.md` — idempotent
- `crow generate-docs --check` — exits 0

PR #883, opened 2026-07-28
Anti-drift CLI↔RPC↔config parity gate (CROW-807)

### LINKED ISSUE #807: [parity] Anti-drift: enumerable CLI↔RPC↔config parity test + CI gate

dhilgaertner:
Closes #807

Crow's control plane exists three times over — the JSON-RPC methods `crowd` registers, the `crow` verbs that call them, and the `AppConfig` fields both mutate. Nothing tied them together, so a capability could ship as an RPC method plus a web control and quietly never grow a CLI path. Since the primary operator of Crow is an agent driving the `crow` CLI, that drift doesn't cost convenience — it removes the main interface.

## The ledger

`ParityLedger` (`Packages/CrowCore/Sources/CrowCore/Parity/ParityLedger.swift`) names **all 87 RPC methods** and **all 70 `AppConfig` leaf fields**, each mapped to the verb that reaches it or carrying a `Coverage.noCLI(reason:)` exemption. The reason is a *required associated value*, so an exemption cannot be added without writing one — **the ledger diff is the review gate**, and each per-area ticket closes its gap by deleting rows.

| | Covered | Exempt |
|---|---|---|
| RPC methods (87) | 78 | 9 (5 writes, 4 reads) |
| Config fields (70) | 33 read+write, 7 partial | 30 |

## Three checks, because no test target can see both sides

`CrowCLI` doesn't depend on `CrowDaemon`, and `CrowDaemon` depends on Darwin-only `CrowTelemetry` so it's absent from `ci.yml`'s `LINUX_PACKAGES` ([ADR 0007](docs/adr/0007-linux-ci-swift.md)). Each check therefore lives where it can see the truth:

| Check | Where | On PRs |
|---|---|---|
| Ledger ⇄ `CrowCommand` verb tree; ledger ⇄ `Mirror` walk of `AppConfig` | `CrowCLITests/ParityGateTests` | ✅ Linux lane |
| Ledger ⇄ real `CommandRouter.methodNames` (authoritative) | `CrowDaemonTests/RPCLedgerParityTests` | ❌ macOS only |
| Ledger ⇄ router source, by text | `scripts/check-cli-parity.sh` | ✅ new `parity` job |

The script is the PR-lane stand-in for the daemon test; the daemon test is what catches a router refactor the regex would under-report.

`CommandRouter` gains `methodNames`, walking the fallback chain, so the surface is enumerable without exposing handler closures.

The `Mirror` walk runs over a **fully-populated probe** `AppConfig` — a default one leaves optionals nil and arrays empty, hiding ~25 fields — and a dedicated test fails if the probe ever stops reaching a subtree.

## Day-one drift, recorded rather than hidden

The exempt set is not flattering, and that's the point:

- `set-config`, `run-setup`, `batch-start-review`, `run-job` — writes with no verb.
- **Engine RPC `set-pinned` has no caller**: `crow set-pinned` actually sends the `set-locked` RPC (`SessionCommands.swift:156`). Either the verb is misrouted or the method is dead. Ledgered with that note; deliberately not fixed here.
- `terminal.*` and all 15 `workspaces[].*` fields have no CLI path at all.
- `defaults.excludeDirs` and `defaults.mirrorClaudeMCPToCodex` are returned by `defaults get` but have no `defaults set` flag — surfaced by the rebase below.

## Verification

All four planned drift canaries were run, confirmed to fail with a precise message, and reverted. **Canary 1 found a real bug**: the script's first regex anchored on end-of-line, so a one-line handler registration (`"foo": { _ in [:] },`) slipped past silently — fixed, then re-verified against both the script and the daemon test.

| | |
|---|---|
| `scripts/check-cli-parity.sh` | ✅ 87 methods agree |
| `ParityGateTests` (7 tests) | ✅ |
| `RPCLedgerParityTests` (2 tests) | ✅ |
| CrowCLI / CrowCore / CrowIPC suites | ✅ 296 / 484 / 37 |
| `shellcheck scripts/*.sh` | ✅ clean |

## The gate proved itself on the rebase

The gate has now caught real drift three times on this branch, each time before a human looked:

- **#887 (defaults CLI)** — rebase turned it red on `defaults-get`/`defaults-set`; stayed red until the nine `defaults.*` rows moved to covered. Working through them surfaced that #887 left two fields readable but not writable.
- **#886 (agents CLI)**, via the `main` merge into this branch — red on `agents-get`/`agents-set`; `defaultAgentKind` and `agentsByKind` moved to covered, and the stale `list-agents` exemption reason ("the CLI never needs to enumerate agents" — it does now) got corrected.
- **Its own regex bug** — a canary caught that the handler pattern was anchored at end-of-line, so a one-line handler registration would have slipped past silently.

Exempt config fields are down 41 → 30 as a result.

## Scope note

The new `parity` job also runs `check-notification-events.sh`, which existed but was reachable only from `make test` — which CI never runs. One line to stop an existing gate being decorative; say the word and I'll drop it.

[ADR 0016](docs/adr/0016-cli-control-plane-parity.md) states the principle the gate mechanizes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,28 @@ jobs:
             swift test --no-parallel --package-path "Packages/$pkg"
           done
 
+  # Source-level drift gates: cheap text checks that need no Swift build, so they
+  # cover packages the Linux lane above cannot compile. `check-cli-parity.sh` is
+  # the PR-lane stand-in for CrowDaemonTests/RPCLedgerParityTests — CrowDaemon
+  # depends on the Darwin-only CrowTelemetry and so is absent from LINUX_PACKAGES,
+  # which would otherwise leave the RPC half of the parity gate running only on
+  # release tags (CROW-807, docs/adr/0016-cli-control-plane-parity.md).
+  #
+  # Plain ubuntu-latest rather than the swift container: check-notification-events
+  # needs node, which the container does not ship.
+  parity:
+    name: Catalog & Parity Gates
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: CLI/RPC control-plane parity (CROW-807)
+        run: bash scripts/check-cli-parity.sh
+
+      # Was previously reachable only via `make test`, which CI never runs.
+      - name: Notification-event catalogs (CROW-768)
+        run: bash scripts/check-notification-events.sh
+
   desktop:
     name: Desktop (Tauri) Build
     runs-on: macos-15
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build daemon app run setup install uninstall clean check test help daemon-run docs
+.PHONY: build daemon app run setup install uninstall clean check test parity help daemon-run docs
 
 # Install destination and build config (override on the command line, e.g.
 # `make install BINDIR=/usr/local/bin` or `make build CONFIG=release`).
@@ -52,7 +52,8 @@ help:
 	@echo "  daemon-run Run crowd serving the frozen bundle-baked web UI (add --watch to rebuild on Swift/web change)"
 	@echo "  setup      Check build prerequisites"
 	@echo "  check      Verify all build and runtime prerequisites"
-	@echo "  test       Run all package tests"
+	@echo "  test       Run all package tests, then the parity gates"
+	@echo "  parity     Run just the source-level drift gates (catalogs + CLI/RPC parity)"
 	@echo "  docs       Regenerate docs/cli.md from the CLI's ArgumentParser metadata"
 	@echo "  install    Symlink crow + crowd into ~/.local/bin (override BINDIR=, CONFIG=release)"
 	@echo "  uninstall  Remove installed crow + crowd symlinks"
@@ -96,8 +97,15 @@ test:
 			swift test --package-path "$$pkg"; \
 		fi; \
 	done
+	@$(MAKE) --no-print-directory parity
+
+# Source-level drift gates. Cheap, no build required, and both also run in CI —
+# see the `parity` job in .github/workflows/ci.yml.
+parity:
 	@echo "==> Checking notification-event catalogs (CROW-768)..."
 	@./scripts/check-notification-events.sh
+	@echo "==> Checking CLI/RPC control-plane parity (CROW-807)..."
+	@./scripts/check-cli-parity.sh
 
 # Regenerate the CLI reference from the commands themselves (CROW-808). Run
 # after adding or changing a subcommand — a stale docs/cli.md fails CrowCLI's
--- a/Packages/CrowCLI/Tests/CrowCLITests/ParityGateTests.swift
+++ b/Packages/CrowCLI/Tests/CrowCLITests/ParityGateTests.swift
@@ -238,6 +238,90 @@ struct ParityGateTests {
         }
     }
 
+    // MARK: - Write classification
+
+    /// Names that read as queries: `get-…`, `list-…`, `…-get`, `…-list`.
+    static func namePresumesRead(_ method: String) -> Bool {
+        method.hasPrefix("get-") || method.hasPrefix("list-")
+            || method.hasSuffix("-get") || method.hasSuffix("-list")
+    }
+
+    /// Rows whose deliberate `isWrite` classification disagrees with the name.
+    ///
+    /// Empty today. The hand classification always wins — a method named
+    /// `get-something` that mutates state is a write, full stop. This set only
+    /// forces the disagreement to be *stated* rather than sitting unnoticed in a
+    /// row nobody re-reads.
+    static let namingExceptions: Set<String> = []
+
+    /// `isWrite` is hand-set per row, deliberately, because a `get-`/`list-`
+    /// heuristic would wave through a future write named `get-something`. But
+    /// nothing checked the hand classification itself — so a write mislabelled
+    /// `.read` would silently skip the write bar below. This cross-checks the two
+    /// against each other without letting the heuristic overrule the decision.
+    @Test("isWrite agrees with the method name, or the disagreement is declared")
+    func writeClassificationMatchesNaming() {
+        let disagreements = ParityLedger.rpcMethods
+            .filter { Self.namePresumesRead($0.method) == $0.isWrite }
+            .map(\.method)
+            .filter { !Self.namingExceptions.contains($0) }
+            .sorted()
+
+        #expect(
+            disagreements.isEmpty,
+            """
+            These rows classify against what their name suggests:
+            \(disagreements.map { "  \($0)" }.joined(separator: "\n"))
+            A `get-`/`list-` name marked .write, or any other name marked .read.
+            If the classification is right, add the method to `namingExceptions`
+            with a comment; if it's wrong, fix the row.
+            """)
+    }
+
+    /// The un-CLI-able writes, pinned.
+    ///
+    /// This is the milestone's burn-down list: each per-area parity ticket
+    /// deletes entries. Pinning the *set* — not just requiring a reason per row
+    /// — is what makes adding one a deliberate, visible act instead of a single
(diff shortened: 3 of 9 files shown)


dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Gate is read-only enumeration of existing control-plane surfaces; no new RPC handlers, auth paths, or config mutators.
- Exemptions for `remoteControlEnabled`, gateways, `webAuth`, and `jiraCredential` correctly call out local-socket / credential constraints rather than papering over them.
- `CommandRouter.methodNames` exposes names only — handler closures stay private.
- `scripts/check-cli-parity.sh` is shellcheck-clean; vacuity floor (`MIN_METHODS=60`) and the macOS `RPCLedgerParityTests` mitigate regex blind spots.

**Concerns:**
- None that affect the security posture of the daemon or CLI. Predictable `${TMPDIR:-/tmp}/cli-parity.diff` is a minor shared-temp footgun for a local/CI script (prefer `mktemp`); not exploitable in the Actions runner path that matters here.

### Code Quality
- **Yellow — `writesAreCoveredOrExempt` is vacuous** (`RPCLedgerParityTests.swift:94-102`). `Coverage` is `.cli` | `.noCLI(reason:)`, so `cliPath == nil && exemptionReason == nil` is unreachable by construction. The test can never fail and does not exercise `isWrite`. Remove it, or replace it with a real assertion (e.g. that every `.write` row is classified intentionally against a known allowlist of write exemptions, or a soft name-heuristic check that `isWrite` matches handler intent).
- Green — `tempDevRoot()` creates dirs under `NSTemporaryDirectory()` and never removes them (`RPCLedgerParityTests.swift:33-38`). Fine for short suites; worth a `defer` cleanup if this pattern spreads.
- Green — `isWrite` is documented as the write-parity bar but is unused outside the dead test; either wire it into a meaningful check or drop the field until something reads it.
- Green — regex-based PR-lane stand-in is acknowledged and dual-gated with the authoritative router test; canary story in the PR body is convincing.

### Static analysis / tests
- `scripts/check-cli-parity.sh` → 85 methods agree
- `ParityGateTests` (5) → pass
- `RPCLedgerParityTests` (3) → pass (incl. the vacuous one)
- CI: Build & Test, Catalog & Parity Gates, Desktop, Shell Lint → all green

### Summary Table

---

dhilgaertner:
Thanks — the Yellow was a fair catch and correct as stated.

## Yellow: `writesAreCoveredOrExempt` was vacuous — fixed

You're right, and the diagnosis is exact: `Coverage` is `.cli | .noCLI(reason:)`, so `cliPath == nil && exemptionReason == nil` is unreachable and the test could never fail. Worse, it was the *only* reader of `isWrite`, so that field was decorative too.

Root cause worth naming: it was a **ledger-only assertion sitting in the router suite**. Both replacements now live in `ParityGateTests` instead — and since CrowDaemon is macOS-only, that also means they run on **every PR** rather than only on release tags. `RPCLedgerParityTests` is now strictly the router-dependent checks.

Two replacements, each verified against a canary:

**`writeClassificationMatchesNaming`** — closes the gap the dead test left open. `isWrite` is hand-set per row precisely because a `get-`/`list-` heuristic would wave through a future write named `get-something`; but nothing checked the hand classification, so a write mislabelled `.read` would silently skip the write bar. This cross-checks name against classification, with a `namingExceptions` escape hatch (empty today) so the deliberate decision still overrules the heuristic rather than the reverse. Canaried both directions — `resync-jira` flipped to `.read`, and `list-sessions` flipped to `.write`; both fail naming the row.

**`writeExemptionsArePinned`** — this is your "known allowlist of write exemptions" suggestion. Pins the set of writes with no CLI path, which is the milestone's burn-down list. Pinning the *set*, rather than only requiring a reason per row, is what makes widening it visible in review instead of one `.noCLI(…)` row lost among 87. Canaried by turning a covered write exempt; fails with `added → ["resync-jira"]`.

## Security: predictable temp path — fixed

Agreed it wasn't exploitable on the Actions path, but the script is meant to be run locally too, so it now uses `mktemp` with a `trap … EXIT` cleanup. Verified the temp file is removed on both the pass and the failure path.

## Green: `tempDevRoot()` leak — fixed

`RPCLedgerParityTests` now builds its router inside a scoped `withLiveRouter { }` helper that `defer`s removal of the temp dev root.

## Green: `isWrite` unused — resolved

Both new tests read it, so it's load-bearing now rather than documentation.

## Green: regex stand-in — no change

Agreed, and the canary story got stronger since: the end-of-line anchor bug I mention in the PR body was found by exactly that canary. Also worth noting the gate has now caught real drift three times on this branch alone — `defaults-get`/`defaults-set` from #887, and `agents-get`/`agents-set` from the #886 merge, each of which also moved config rows out of the exempt set (exempt config fields are down 41 → 30).

---

Verified after the changes: `make parity` (87 methods agree), `ParityGateTests` 7/7, `RPCLedgerParityTests` 2/2, CrowCLI 296 / CrowCore 484 / CrowIPC 37, shellcheck clean.

dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No auth/RPC surface expansion — this is a governance gate over existing methods and config fields.
- `CommandRouter.methodNames` exposes only method name strings, not handler closures (`CommandRouter.swift:33-35`).
- Sensitive config paths are handled correctly: `webAuth.hashB64`/`saltB64` are read-exempt with clear justification; `jiraCredential.*` and `remoteControlEnabled` exemptions explicitly call out local-socket / privilege-widening constraints (`ParityLedger.swift:340-355`, `516-522`, `602-615`).
- `check-cli-parity.sh` uses `mktemp` + `trap` cleanup (not a predictable `/tmp` path) and has a vacuity floor (`MIN_METHODS=60`) so a blind regex cannot silently pass.

**Concerns:**
- None.

### Code Quality
- **Green** — Consider a small `CrowIPCTests` unit test for `methodNames` (primary ∪ fallback). The daemon suite covers it on macOS (`RPCLedgerParityTests.fallbackChainIsIncluded`), but CrowIPC is in the Linux PR lane and currently has no direct coverage of the new API (`CommandRouterTests.swift` still stops at fallback *dispatch*).
- **Green** — The ledgered `set-pinned` dead-path / CLI misroute (`SessionCommands.swift` → `set-locked`) is correctly recorded rather than hidden; worth a follow-up ticket so the exemption can be deleted.
- Three-check split (CLI Mirror + daemon router + text stand-in) matches the package graph constraints (ADR 0016) and the PR-lane regex bugfix (no EOL anchor) is sound.
- `ParityGateTests` (7/7) passed locally; `scripts/check-cli-parity.sh` reports 87 methods in sync; `shellcheck` clean; CI jobs (Build & Test, Catalog & Parity Gates, Shell Lint) all green.

### Summary Table

---

PR #884, opened 2026-07-28
Automation CLI (parity): crow automation get|set

### LINKED ISSUE #812: [parity] Automation CLI: permission-mode + watcher + auto-respond toggles
Milestone: **CLI control-plane parity**. Area: **Automation** (Settings → Automation).

Add `crow automation get|set` for the automation toggles:
- remoteControlEnabled, managerAutoPermissionMode, reviewAutoPermissionMode, coderViewAutoPermissionMode, jobsAutoPermissionMode
- attributionTrailers, autoCreateWatcherEnabled, autoMergeWatcherEnabled
- autoRespond.{respondToChangesRequested, respondToFailedChecks, autoRebaseAndResolveConflicts}

Config: `AppConfig` top-level bools + `AppConfig.autoRespond`.

dhilgaertner:
Closes #812

Settings → Automation's toggles were web-only. `crow automation get|set` covers all eleven — the eight top-level `AppConfig` booleans and the three under `autoRespond`.

```bash
crow automation get
crow automation set --auto-merge-watcher-enabled true
crow automation set --respond-to-failed-checks true --auto-rebase-and-resolve-conflicts true
crow automation set --manager-auto-permission-mode false && crow restart-manager
```

`--jobs-auto-permission-mode` is included per the ticket even though the web UI renders it under the Jobs tab, so all five permission modes read and write as one group.

## Writing config is the whole job

The thing that shaped this PR: **flipping one of these in the Settings modal has no side effect either.** `toggleField`'s `onchange` (`settings.js:269-281`) only mutates the in-memory copy and marks the form dirty. Every consumer pulls from disk — `applyConfigToAppState` re-runs each 60s board tick, the `IssueTracker` watcher gates are closure providers that reload config on every call, and `AutoRespondCoordinator` takes a `settingsProvider` closure for the same reason. The "watchers" aren't processes; they're guard clauses inside the one poll loop that always runs.

So there is no watcher to start, nothing to re-register, and `restart_required` is always `false`.

`managerAutoPermissionMode` is the one exception — it's baked into the Manager terminal's stored shell command by `SessionService.managerCommand(for:)` and only re-read on a rebuild. `automation-set` reports that as `manager_restart_required`, **only when the value actually moved**, and the CLI warns on stderr pointing at the existing `crow restart-manager`.

## The tab's list fields belong to `crow defaults`

⚠️ **This changed during review** — worth reading if you saw the first version of this PR.

The Automation tab also renders three chip editors (excluded review repos, ignored review labels, excluded ticket repos), and this branch originally claimed them. While it was open, **#887 (`crow defaults`) landed with the same nine flag names** — `--add-exclude-review-repo`, `--clear-ignore-review-labels`, and so on — because those fields live in `AppConfig.defaults`, not at the top level.

Worse than duplication, the semantics had diverged: `defaults` makes `--clear-X` *exclusive* with add/remove and matches removes *case-insensitively*; this branch allowed clear+add as a "replace" idiom and deduped case-sensitively. Two commands writing one config field with two sets of list semantics is exactly the drift the parity milestone exists to prevent.

So `automation set` no longer claims those flags. Instead:

- `automation get` echoes the three lists **read-only** under `defaults`, so the tab still reads as a whole from one call.
- `automation-set` **ignores** `add_*` / `remove_*` / `clear_*` params even from a hand-rolled RPC call — they don't count as "something to set" on their own, and are silently inert alongside a real field rather than opening a second write path.
- Both halves are pinned by tests (`automationSetDoesNotClaimTheDefaultsListFlags`, `setIgnoresDefaultsListParams`), and the docs cross-reference in both directions.

## Shape

Granular RPC rather than a CLI-side `set-config` round-trip, same rationale as CROW-814 (`RPCHandlers.swift`): `set-config` replaces the entire `AppConfig`, so a blob round-trip would make the CLI a second writer racing the web Settings modal. The write is a locked read-modify-write through the existing `mutateConfig`, which shares the config lock with `set-config` and the scheduler and refuses to overwrite a `config.json` that exists but won't decode.

- **Booleans are `@Option ... Bool?`, not `@Flag`.** Six of the eleven default to **on** (`manager`/`review`/`jobs` auto permission mode, `attributionTrailers`, `respondToChangesRequested`), so a bare-flag design could never turn one off.
- **One command, not a verb per group**, so a multi-field edit is one config lock, one mtime bump, one `configReloaded` broadcast. The flags are split into titled `@OptionGroup`s matching the web tab's own `group()` headings.
- **`get` returns two things the UI doesn't surface**: the derived `defaults.effective_exclude_review_repos` (the global list unioned with every workspace's own — what the review board actually filters on, and the only way to see the per-workspace half from the CLI; it also explains why `crow defaults set --clear-exclude-review-repos` may not unhide a repo), and `config_readable`, following `notifications-get`. The latter matters more here than for telemetry: with six defaults being `true`, showing invented settings as fact would tell a caller automation is armed when the daemon can't read the file at all.

**Not gated** in `localOnlyDenial`, inheriting the CROW-814 rationale — an authenticated remote peer can already change every one of these fields through the un-gated `set-config` path, which is literally what `setConfigHarmlessToggleIsAllowedRemotely` asserts (it flips `remoteControlEnabled`, an automation field).

## Tests

Across five packages — CLI parsing, pure RPC support, router→disk integration, config round-trip, and the security gate.

Worth calling out one pair, which pins the seam neither package could catch alone. CrowCLI and CrowDaemon never share a symbol for the wire param names, so a typo on either side would compile, pass the CLI tests (which check parsed Swift properties) *and* pass the handler tests (which pass params in directly), while silently dropping the field:

- `automationSetEmitsTheExpectedWireKeys` — parses every flag and asserts `AutomationSet.params` produces exactly those eleven keys, no more, no less.
- `setAcceptsEveryWireKeyTheCLIEmits` — feeds that same list to the real router and asserts every field lands on disk, and that the `defaults` lists are left untouched.

Also pinned: the CROW-551 legacy-key migration. `autoRebaseWatcherEnabled` is carried forward as a one-way OR, so a `--auto-rebase-and-resolve-conflicts false` against a config still holding the legacy key has to actually stick or the opt-out silently reverts on every reload.

```
CrowCLI          302 tests   ✔
CrowCore         489 tests   ✔
CrowPersistence   53 tests   ✔
CrowEngine       471 tests   ✔
CrowDaemon       271 tests   ✔
```

`docs/cli.md` regenerated with `make docs` (#882's guard test keeps it honest).

## Notes for the reviewer

**The parity-harness checklist item is still skipped.** The #807 gate is #883, still open — it isn't on this branch's base, so there's no `KNOWN_UI_ONLY` list to add to or remove from yet. Whichever of the two lands second should wire them together: after this merges, `automation-get` / `automation-set` and all eleven `AppConfig` booleans are covered, so they should need no `KNOWN_UI_ONLY` entries. (The three `defaults` list fields are covered by #887.)

**Verification caveat:** the handler integration tests exercise the real `makeCommandRouter` → handler → `ConfigStore` path against a real `config.json`, and the wire-contract pair above covers the CLI↔daemon param boundary. The one link not exercised live is the Unix socket transport itself, which is unchanged and shared with every other verb — testing it would have meant restarting the running `crowd` (the store writer lock is global regardless of `--socket`), which kills every live agent session, so I left the running daemon alone.


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Granular PATCH via locked `mutateConfig` refuses to overwrite an undecodable `config.json`, preventing silent wipe of workspaces/jobs/credentials (`RPCHandlers.swift` ~1838–1856; covered by `writeRefusesToOverwriteAnUndecodableConfig`).
- List entries validated on both CLI (`validateListEntry`) and daemon (`SettingsRPC.patchStringList`) — blank and control-character values rejected; wrong types fail loudly rather than being treated as absent.
- `automationJSON` does not leak credentials, gateway headers, or `defaults.binaries` (pinned by `settingsResponsesNeverContainCredentialKeys`); `automation-set` can write the three board-filter lists but cannot touch the sole local-only field.
- Remote un-gating of `automation-*` is correctly justified: authenticated peers already reach the same fields through `set-config`, and the granular path is the smaller wire surface (`RPCWebSocketHandler.localOnlyDenial`; `settingsRPCsAreAllowedRemotely`).
- Exclude patterns are simple prefix/suffix globs (`repoMatchesPatterns`), not regex — no injection/ReDoS surface from list values.

**Concerns:**
- None blocking. Enabling powerful toggles (`remote_control_enabled`, auto-merge/create, auto-rebase) over remote `/rpc` remains possible — same trust model as existing `set-config`, and still gated by web auth.

### Code Quality
- Clean CLI shape: `@Option Bool?` for three-state patches, bare `@Flag` only for `--clear-*`, clear→remove→add list semantics, `manager_restart_required` only on real change.
- Excellent cross-package wire-contract pair (`automationSetEmitsTheExpectedWireKeys` ↔ `setAcceptsEveryWireKeyTheCLIEmits`) closing the typo-silent-drop seam.
- Strong test coverage across CrowCLI / CrowEngine / CrowPersistence / CrowDaemon; local `swift test --filter Automation` (and SettingsRPC / automation ConfigStore filters) all passed.
- Docs (`cli-reference`, `automation`, `configuration`, CLAUDE templates) match the implementation, including the deliberate #883 parity-harness deferral.

### Summary Table

**Green (consider):**
1. Wire param key lists are duplicated as literals in CrowCLI and CrowDaemon tests — the paired contract tests mitigate drift; keep them in lockstep on future flag adds.
2. CLI validates trimmed list entries but sends the raw flag value; the daemon trims on receive, so behavior is correct. Trimming in `AutomationListArgs.params` would make CLI/server identical without changing outcomes.
3. Parity harness / `KNOWN_UI_ONLY` wiring is correctly deferred until #883 lands — follow up so these verbs aren't left off the harness by accident.

---

PR #885, opened 2026-07-28
Workspaces CLI (parity)

### LINKED ISSUE #809: [parity] Workspaces CLI: manage AppConfig.workspaces (add/edit/list/remove + all fields)
Milestone: **CLI control-plane parity**. Area: **Workspaces** (Settings → Workspaces).

Add typed CLI verbs to fully manage workspaces — everything the Workspaces settings tab can change:
`crow workspace list | add | edit | remove`, covering `WorkspaceInfo`:
- name, provider (github/gitlab), host (GitLab), alwaysInclude[], autoReviewRepos[], excludeReviewRepos[], customInstructions
- taskProvider (github/gitlab/jira/corveil), jiraProjectKey, jiraJQL, jiraSite, jiraStatusMap (5 pipeline→status mappings), corveilHost
- (workspace AI gateway handled in the gateways/secrets ticket — local-only)

Config: `AppConfig.workspaces[]` (`CrowCore/Models/AppConfig.swift`), reuse `WorkspaceInfo.validateName`.

dhilgaertner:
Closes #809

Exposes `AppConfig.workspaces` — until now editable only through the web
Settings → Workspaces tab — as `crow` verbs, following the `crow job`
precedent (#604).

```bash
crow workspace list
crow workspace get    --workspace Acme
crow workspace add    --name Acme --provider gitlab --host gitlab.acme.io
crow workspace edit   --workspace Acme --jira-status-in-progress "In Dev"
crow workspace remove --workspace Acme --force
```

## Approach

Five granular RPCs (`workspace-list|get|add|edit|remove`) rather than
routing the whole `AppConfig` through `set-config`: changing one field
should not put every workspace, job, gateway URL and credential shell on
the wire, and CrowCLI can't author an `AppConfig` anyway. The
read-modify-write happens server-side under `ConfigStore.withConfigLock`.

- `--workspace` resolves by name (case-insensitive) or UUID, reusing the
  resolver `crow gateway` already had. It moves to `WorkspaceRPC` so both
  verb groups share one definition of what `--workspace Acme` means.
- Field flags live in one `@OptionGroup` shared by `add` and `edit`.
  Optional scalars clear with an empty string (`--host ""`), matching the
  web form; lists and maps take explicit `--clear-*` flags because an
  empty repeated flag is indistinguishable from an omitted one.
- Repeatable flags replace the whole list (the `job edit --prompt` rule);
  `--jira-status-*` patches per key, since the five map entries are
  independently editable in the UI.
- An edit whose values already hold returns `"saved": false` and skips
  the write — re-running an idempotent invocation shouldn't churn
  config.json and chime "Config reloaded" in every open browser.

## Three defects closed along the way

- **`WorkspaceInfo.validateName` had zero production callers.** The
  Settings form checks only "non-blank", so duplicate and path-unsafe
  names (`a/b`, `.`, `..`) have been persistable all along. These verbs
  are its first enforcement.
- **`sessionEnv` was consumed but not modeled.**
  `skills/crow-workspace/setup.sh` reads `.workspaces[].sessionEnv` with
  jq and SKILL.md documents it as a peer of `gateway`, but `WorkspaceInfo`
  had no such field — and since `encode(to:)` is synthesized from
  `CodingKeys`, every config save silently deleted a hand-authored block,
  after which the jq read returned empty with no error. Now modeled, with
  `--session-env KEY=VALUE`.
- **Renaming silently orphaned sessions and jobs.** Nothing stores a
  workspace id: a session is tied to its workspace only by its worktree
  living at `{devRoot}/{workspace}/{repo}`, and a job only by the
  `job.workspace` string. A rename moves no directory, so it broke
  gateway resolution, code-provider detection (falling back to GitHub),
  job launches and the skill's config lookups — while `Scaffolder` created
  a new empty directory beside the old populated one. Rename and remove
  now refuse while references exist and report the counts; `--force`
  proceeds.

## Security

`workspace-*` is deliberately NOT gated in `localOnlyDenial`, and
`DaemonSecurityTests` pins that. The one thing that would justify gating —
the per-workspace AI gateway — is *excluded* rather than gated: no
`workspace-*` method writes it, and `workspaceJSON` reduces it to
`gateway_set` plus a base URL, never the header names or values.
Authoring the credential stays with `gateway-set`, which is gated. So the
remote-reachable surface is strictly smaller than the `set-config` it
replaces. Edits preserve a stored gateway (matched by workspace `id`);
`remove` discards it and says so.

## Notes

- `makeCommandRouter`'s handler dictionary needed splitting: at ~65
  closures in one literal, appending five more pushed Swift's type checker
  past its solver budget on an unrelated line. The literal now carries an
  explicit type annotation and new verb groups get their own function.
- `Provider` gains `CaseIterable` so `WorkspaceInfo.validProviders` /
  `validTaskProviders` derive from it — the CLI's rejection messages and
  the handler's validation read the same source.
- **Deferred:** the #807 parity harness still doesn't exist
  (`KNOWN_UI_ONLY`, `ParityTest`, `coveredSet` return zero hits repo-wide),
  so the "add to the covered set" checklist item is skipped, as in #870,
  #871 and #873.
- **Separate, pre-existing:** `SetupCommand.swift` writes `SetupWorkspace`
  without an `id` while `WorkspaceInfo.init(from:)` requires one, so a
  `crow setup`-generated config with any workspace fails `AppConfig`
  decode outright. Worth its own issue.

## Testing

`make test` green (all 18 packages) plus the notification-catalog check.
New: `WorkspaceCommandParsingTests` (39), `WorkspaceRPCSupportTests` (30),
`WorkspaceHandlerTests` (23, real router against a real config.json in a
temp dev root), additions to `ValidationTests`, `AppConfigTests`,
`ConfigStoreTests` and `DaemonSecurityTests`.

Not verified live against a running daemon: `crowd` holds a single-writer
lock on `~/Library/Application Support/crow` regardless of `--socket`, so
a second instance can't be started alongside the developer's, and stopping
theirs would kill live sessions. The handler tests exercise the real
router and real persistence; the untested seam is only the Unix-socket
round-trip, which is the same `rpc()` boilerplate as every other verb.

Crow-Session: 5C3E85E3-90A0-48AB-A505-A9F8023BC8DE


dgershman (changes requested):
## Code & Security Review

Solid CLI parity for Settings → Workspaces: gateway credentials stay out of the remote surface, `validateName` finally has production callers, and `sessionEnv` is modeled so config saves no longer silently drop it. One validation gap on the newly remote-writable `session_env` path needs fixing before approve.

### Critical Issues (if any)
None (no Red).

### Security Review
**Strengths:**
- Gateway headers never appear in `WorkspaceRPC.workspaceJSON` — only `gateway_set` + `gateway_base_url`; writes stay on gated `gateway-set` (`WorkspaceRPCSupport.swift:346-372`, `DaemonSecurityTests` pins both sides).
- `applyPatch` preserves `id`/`gateway` in place so `SettingsSecrets.preservingSecrets` cannot drop credentials on edit (`WorkspaceRPCSupport.swift:211-215`, `applyPatchPreservesIDAndGateway`).
- Name validation rejects `/`, `:`, NUL, `.`, `..`, and case-insensitive duplicates — first production use of `WorkspaceInfo.validateName`.
- Writes refuse to overwrite an undecodable `config.json` (`WorkspaceHandlerTests.writesRefuseToOverwriteAnUndecodableConfig`).
- CLI rejects embedded newlines in `--session-env` (`Validation.swift:244-256`), matching the `validateHeaderLine` pattern.

**Concerns:**
- **Yellow — `session_env` newline validation is CLI-only.** `validateSessionEnvEntry` correctly rejects newlines because consumers line-split `KEY=VALUE` (`skills/crow-workspace/setup.sh` `resolve_session_env` / `write_settings_local`). Server-side `WorkspaceRPC.patchStringMap` (`WorkspaceRPCSupport.swift:93-109`) accepts any string and `applyPatch` stores it unchanged. `workspace-*` is intentionally remote-reachable (`RPCWebSocketHandler` ledger + `workspaceRPCsAreAllowedRemotely`), so a remote `/rpc` peer can set e.g. `{"FOO":"bar\nEVIL=injected"}`; `jq -r '… | "\(.key)=\(.value)"'` then yields a second pair and injects an extra `.env` entry into `settings.local.json`. Mirror the CLI check in `applyPatch` (or a `decodeSessionEnv`) for every key and value — reject `CharacterSet.newlines` — and add a regression test next to `applyPatchReplacesSessionEnv`.

### Code Quality
- **Green — rename/remove reference guard is TOCTOU by design.** References are counted before `withConfigLock`; under the lock only the name is re-validated (`RPCHandlers.swift:1790-1828`). A session/job created in the window can be orphaned without `--force`. Comment explains avoiding a MainActor hop under the lock — acceptable if documented as best-effort; re-checking under the lock (or holding AppState first) would close it.
- **Green — `patchStringMap` stores untrimmed keys** after only checking blank-after-trim (`WorkspaceRPCSupport.swift:103-106`). CLI trims keys in `parseSessionEnv`. Trim (or reject padded keys) for parity.
- Coverage is strong: `WorkspaceRPCSupportTests` (30), `WorkspaceCommandParsingTests` (25 filtered run green here), `WorkspaceHandlerTests`, plus `DaemonSecurityTests` for the local-only split. Static analysis: CrowEngine `WorkspaceRPCSupportTests` and CrowCLI workspace parsing suites passed locally (SwiftPM exit 1 is the known arm64-vs-x86_64 XCTest bundle loader warning, not a test failure).

### Summary Table

---

dhilgaertner:
Thanks — the Yellow was real and I confirmed the consequence rather than taking it on faith. Pushed in 49e0d5b.

**Yellow — `session_env` newline, fixed.** You were right that the CLI check doesn't hold on the remote path. Reproduced the actual injection first:

```
$ jq -r '… | .sessionEnv // {} | to_entries[] | "\(.key)=\(.value)"'   # {"FOO":"bar\nEVIL=injected"}
FOO=bar
EVIL=injected
```

`resolve_session_env` feeds that into `while IFS= read -r kv`, so one stored value becomes two `.env` entries. Added `WorkspaceRPC.decodeSessionEnv`, applied in `applyPatch`, rejecting newlines in keys **and** values — keys split the line just as effectively, and I used `CharacterSet.newlines` so CRLF is caught (Swift treats `\r\n` as a single `Character`, so `contains("\n")` misses it).

Three tests: the regression next to `applyPatchReplacesSessionEnv` as you asked; an over-reach guard (`applyPatchAllowsOrdinarySessionEnvValues`) so the fix can't start rejecting `=`, spaces, `#`, or an empty value, which are all ordinary env content; and `sessionEnvNewlineIsRejectedOverRPC` driving the real router on both add and edit, since a remote peer — not the CLI — is the threat model.

**Green — untrimmed keys, fixed.** `patchStringMap` now trims, matching `parseSessionEnv`.

**Green — TOCTOU on the reference guard: documented, not closed.** Taking your "acceptable if documented as best-effort" option, and I want to be explicit about why rather than just labelling it. Re-counting under the lock means a main-actor hop inside a held `NSLock` — which is the exact shape of #874, where a blocking call on the MainActor wedged the daemon's whole RPC surface. The guard's job is to stop someone renaming a workspace they can see is in use; it isn't a transactional invariant, and a rename racing a session launch isn't worth that deadlock risk. The comment now names the window, says the counts are a snapshot, and points at #874 so the next person doesn't "fix" it into a wedge.

Also documented the one-variable-per-entry rule in `cli-reference.md` and `configuration.md` — it's a user-visible constraint, not just an internal check.

dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Gateway credentials stay local-only: `workspaceJSON` collapses to `gateway_set` + `gateway_base_url`, never `customHeaders`; `applyPatch` never writes `id`/`gateway`; `DaemonSecurityTests` pins both the un-gated `workspace-*` decision and that `gateway-set` remains gated.
- Server-side newline rejection in `WorkspaceRPC.decodeSessionEnv` (`WorkspaceRPCSupport.swift:182–196`) correctly closes the remote `/rpc` bypass of CLI `validateSessionEnvEntry` — CRLF handled via `CharacterSet.newlines`.
- First production enforcement of `WorkspaceInfo.validateName` (path separators, `.`/`..`, case-insensitive uniqueness) under `withConfigLock`.
- Malformed `config.json` is refused rather than overwritten with defaults (`mutateConfig` / `mutateConfigIfChanged`).

**Concerns:**
- **Yellow** — `session_env` keys containing `=` are accepted over RPC. `decodeSessionEnv` rejects newlines because `setup.sh` round-trips the map as one `KEY=VALUE` per line (`jq -r '… | "\(.key)=\(.value)"'` → `key="${kv%%=*}"`). A key like `FOO=BAR` emits `FOO=BAR=baz` and re-parses as key `FOO`, value `BAR=baz` — same delimiter contract the newline guard protects, but only for newlines. The CLI cannot produce such keys (`parseSessionEnv` splits on the first `=`), so this is a remote-`/rpc`-only gap, matching the rationale already given for putting the newline check server-side. Reject `=` in keys in `decodeSessionEnv` (and add a test next to `applyPatchRejectsNewlinesInSessionEnv`).

### Code Quality
- Solid PATCH semantics: present-but-wrong-typed params throw; empty-string clears scalars; `clear_*` for collections; no-op edits skip the disk write.
- Rename/remove reference guards are well documented (including the intentional pre-lock race vs `#874` deadlock tradeoff).
- Test coverage is strong (`WorkspaceRPCSupportTests` 33, `WorkspaceHandlerTests` 22, CLI parsing/validation) — all filtered suites exercised here passed.
- **Green** — Consider also rejecting obviously non-identifier `session_env` keys (whitespace, control chars) for defense in depth; lower priority than `=`.
- **Green** — Web Settings still does not enforce `validateName` (called out in the PR); fine to leave as follow-up since these verbs are the first callers.

### Summary Table

---

dhilgaertner:
Good catch — that's the same delimiter contract as the newline guard, and I'd only covered half of it. Fixed in 822e52d.

**Yellow — `=` in a key.** Confirmed the round-trip before changing anything:

```
{"FOO=BAR": "baz"}  ->  FOO=BAR=baz  ->  key=FOO  value=BAR=baz
```

`decodeSessionEnv` now rejects `=` in keys. A `=` in a **value** stays legal — the split takes only the first one, so connection strings and base64 are unaffected — and `applyPatchAllowsOrdinarySessionEnvValues` pins that, so the fix can't quietly over-reach later.

**Green — non-identifier keys.** Took it, since I was already in the validator. Keys may no longer carry whitespace or control characters. Worth being explicit about the reasoning, because it's a different argument from the `=` one: this isn't a delimiter problem, it's that no shell can reference `FOO BAR`, so accepting one writes an entry that can never be read back.

I deliberately stopped short of a strict `[A-Za-z_][A-Za-z0-9_]*` rule. `sessionEnv` has no web-Settings editor, so there's no UI to be stricter than — but there *are* hand-authored configs in the wild, and a full identifier rule would start rejecting keys that work fine today. Reads remain unvalidated either way, so existing config keeps loading regardless.

`validateSessionEnvEntry` gained the same key rules so the CLI still fails fast locally. It can't produce a `=` key, but it can produce `FOO BAR=v`, so the two validators would otherwise have disagreed.

**Green — web Settings doesn't enforce `validateName`.** Agreed, leaving as follow-up. These verbs are its first callers and the form is outside this ticket's surface; happy to file it if you'd like it tracked.

Tests landed next to `applyPatchRejectsNewlinesInSessionEnv` as asked, plus `sessionEnvKeyWithEqualsIsRejectedOverRPC` driving the real router — remote peers being the actual threat model — and a CLI validator case. Docs updated in `cli-reference.md` and `configuration.md`, since which characters are reserved is a user-visible constraint rather than an internal detail.

dgershman (changes requested):
## Code & Security Review

Prior Yellow findings on `session_env` (newline / `=` / unaddressable keys) are fixed server-side in `WorkspaceRPC.decodeSessionEnv`, with CLI parity and handler-level regression tests. One remaining coherence gap still contradicts the documented clearing contract.

### Critical Issues (if any)
None (no Red).

### Security Review
**Strengths:**
- Gateway credentials stay local-only: `workspaceJSON` collapses to `gateway_set` + `gateway_base_url`, never `customHeaders`; `applyPatch` never touches `id`/`gateway`; `DaemonSecurityTests` pins both the un-gated `workspace-*` decision and that `gateway-set` remains gated.
- Remote `/rpc` can no longer smuggle extra `.env` pairs via `session_env`: newlines, `=` in keys, and whitespace/control characters in keys are rejected in `WorkspaceRPC.decodeSessionEnv` (`WorkspaceRPCSupport.swift:198–220`), with `applyPatchRejectsNewlinesInSessionEnv` / `applyPatchRejectsEqualsInSessionEnvKeys` / handler RPC tests.
- First production enforcement of `WorkspaceInfo.validateName` under `withConfigLock`; malformed `config.json` is refused rather than overwritten.

**Concerns:**
- None remaining that rise above Green for this pass.

### Code Quality
- **Yellow — stranded `jira_status_map` per-key clears are rejected.** `validateCoherence`'s `wrote` treats any non-string present value as a write (`WorkspaceRPCSupport.swift:370–375`): scalars with `""` return false (clear allowed), but a `jira_status_map` object always returns true. So after flipping a workspace off Jira, `--jira-status-ready ""` (or the equivalent RPC patch `{"Ready":""}`) fails with "jira_status_map applies to Jira workspaces only", even though every entry value is a clear. That contradicts the file's own comment ("Clearing a field is always allowed") and `docs/cli-reference.md` ("Clearing a stranded field is always allowed" / "`--jira-status-ready \"\"` clears just that one mapping"). `--clear-jira-status-map` still works as a whole-map escape hatch. Fix: treat a `jira_status_map` patch as a write only when at least one value is non-blank after trim; extend `applyPatchAllowsClearingAStrandedField` to cover a stranded map entry clear.
- **Green — rename/remove reference guard remains best-effort TOCTOU** (documented vs `#874`); acceptable.
- **Green — `session_env` values remain visible on remote `workspace-list`/`get`.** Intentional per `AppConfig` ("not treated as a credential"), but `skills/crow-workspace/SKILL.md` still says a session-env value may carry a secret — worth aligning the skill text so operators don't put tokens there.
- Coverage is strong: CrowEngine `WorkspaceRPCSupportTests` 35/35 and CrowCLI `WorkspaceCommandParsingTests` 25/25 passed locally (SwiftPM exit 1 is the known arm64-vs-x86_64 XCTest bundle loader warning, not a test failure).

### Summary Table

---

dhilgaertner:
Right on both counts — the map case genuinely broke the contract the file and the docs both advertise. Fixed in 3738419.

**Yellow — stranded per-key clears.** Reproduced it as a failing test before touching anything:

```
✘ applyPatchAllowsClearingAStrandedField: Caught error: .invalidParams("jira_status_map applies to
  Jira workspaces only — set task_provider=\"jira\" (currently \"github\")")
```

The root cause was that `wrote` asked *"is this a clear?"* of the JSON **type** rather than the contents — a string was a clear when blank, anything else returned `true` unconditionally. It now inspects contents: a map is a clear when every value is blank. Worth noting the severity was a bit worse than inert, as you implied: with only `--clear-jira-status-map` working, a single stranded entry was *unreachable*, not just awkward.

One judgement call in the fix — a **mixed** patch like `{"Ready": "", "Done": "Closed"}` counts as a write, not a clear. The write is the half that would otherwise be silently ignored, which is the exact failure mode this check exists to prevent. `applyPatchStillRejectsAStrandedStatusMapWrite` pins that, so the fix can't later over-correct into accepting real writes.

Beyond the extension you asked for on `applyPatchAllowsClearingAStrandedField`, I added `strandedJiraStatusEntryCanStillBeCleared` on the real router — the reported symptom is a CLI invocation, so it seemed worth covering the path a user actually takes rather than only the pure function.

**Green — `sessionEnv` secrets wording.** Good catch, and it was my inconsistency: `WorkspaceInfo`'s doc says "don't put tokens here" while `SKILL.md` justified the `chmod 600` with "(a session-env value may carry a secret)", which reads as sanctioning it. Reworded in the skill and its template to say plainly that sessionEnv is not a credential store — plaintext in `config.json`, not stripped by `SettingsSecrets`, readable through remote `workspace-list`/`get` — and that the chmod is defence in depth rather than a reason to store one.

I deliberately left `setup.sh`'s "never logs values (one may carry a secret)" as-is. That one is defensive hygiene about what a user *might* have put there, and it stays correct regardless of what they should do — softening it would weaken a good habit for the sake of consistency.

**Green — rename/remove TOCTOU.** Unchanged, still documented as best-effort against #874.

dgershman (approved):
## Code & Security Review

Prior Yellow findings are addressed: `session_env` delimiter injection is rejected server-side in `WorkspaceRPC.decodeSessionEnv`, and stranded `jira_status_map` per-key clears now pass `validateCoherence` via content-aware `isBlank` (commit `3738419`).

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Gateway credentials stay local-only: `workspaceJSON` collapses to `gateway_set` + `gateway_base_url`, never `customHeaders`; `applyPatch` never writes `id`/`gateway`; `DaemonSecurityTests` pins un-gated `workspace-*` and gated `gateway-set`.
- Remote `/rpc` cannot smuggle extra `.env` pairs via `session_env`: newlines, `=` in keys, and whitespace/control characters in keys are rejected in `WorkspaceRPCSupport.swift` (`decodeSessionEnv`), with CLI parity and handler RPC regression tests.
- First production enforcement of `WorkspaceInfo.validateName` under `withConfigLock`; malformed `config.json` is refused rather than overwritten.
- Rename/remove refuse while sessions/jobs reference the workspace; `--force` reports orphan counts.

**Concerns:**
- None remaining above Green.

### Code Quality
- **Green — rename/remove reference guard remains best-effort TOCTOU** (documented vs `#874`); acceptable — re-counting under the lock would main-actor-hop inside `NSLock`.
- **Green — `session_env` values remain visible on remote `workspace-list`/`get`.** Intentional per `AppConfig` / updated `SKILL.md`; tokens belong in gateway headers.
- Coverage is strong: CrowEngine `WorkspaceRPCSupportTests` 36/36, CrowCLI `WorkspaceCommandParsingTests` 25/25, CrowDaemon `WorkspaceHandlerTests` 24/24 passed locally (SwiftPM exit 1 is the known arm64-vs-x86_64 XCTest bundle loader warning, not a test failure).

### Summary Table

---

PR #886, opened 2026-07-28
Agents CLI: crow agents list|set (parity)

### LINKED ISSUE #811: [parity] Agents CLI: defaultAgentKind + per-role agentsByKind (+ agents list)
Milestone: **CLI control-plane parity**. Area: **Agents** (Settings → General).

Add `crow agents list|set`:
- `agents list` — surface available agent kinds (expose the existing web-only `list-agents` RPC as a CLI verb).
- `agents set` — `defaultAgentKind` + `agentsByKind` per role (work/review/job/manager).

Config: `AppConfig.defaultAgentKind`, `AppConfig.agentsByKind`.

dhilgaertner:
Closes #811

Adds CLI parity for the web Settings → General "Agent" group — `AppConfig.defaultAgentKind` and the per-role `AppConfig.agentsByKind` overrides, which had no `crow` verb.

```bash
crow agents list
crow agents set --default codex
crow agents set --work codex --review cursor
crow agents set --clear review --clear job
```

`list` answers the whole question in one payload — `available` (what this daemon registered at boot), `default_agent_kind` + `by_kind` (what you configured), and `effective` (what a new session of each role actually resolves to):

```json
{"agents": {
  "available": [{"kind": "claude-code", "name": "Claude Code"}],
  "default_agent_kind": "claude-code",
  "by_kind": {"review": "codex"},
  "effective": {"work": "claude-code", "review": "codex", "job": "claude-code", "manager": "claude-code"},
  "config_readable": true
}}
```

`effective` resolves through the canonical `AppConfig.agentKind(for:)` — the same call the session-creation path makes — so it cannot drift from what actually launches.

## Shape

Follows the PR #604 / CROW-813/814 precedent: granular `agents-get` / `agents-set` PATCH RPCs doing a locked read-modify-write via `mutateConfig`, rather than a CLI-side round-trip of the whole `AppConfig` blob (CrowCLI cannot decode one, and a blob write would make the CLI a second writer racing the web Settings modal). Pure decode/encode lives in `AgentsRPCSupport.swift`, with the agent registry threaded in as an `[AvailableAgent]` value snapshot so every gate is a pure function and unit-testable without a live registry.

No `restart_required`: the board tick calls `applyConfigToAppState` → `AppState.applyAgentConfig`, so a change is live within one poll.

## Three decisions worth a reviewer's attention

**Unavailable kinds are rejected; nothing is written.** `AgentKind` is an open struct with a total `init(rawValue:)`, so `"clade-code"` is a well-formed value — and unlike `crow new-session --agent`, there is no later `?? default` to catch it, because the *configured* kind is trusted as-is at launch time. Every decode runs before `mutateConfig` is entered, so a rejection takes no lock, writes nothing, and does not bump mtime. Pinned by a test asserting `config.json` is byte-identical across nine rejection cases.

**`--clear <role>` deletes the key, never writes a null.** `agentsByKind` is `[String: AgentKind]`, `AgentKind` decodes from a single-value String container, and `AppConfig.init(from:)` decodes the map with `try`, not `try?` — so one `{"work": null}` makes the **entire** `config.json` undecodable, hiding every workspace/job/gateway and making `mutateConfig` refuse all later writes. `web/settings.js` does `delete cfg.agentsByKind[key]` for exactly this reason. Asserted three ways (decoded map, raw file bytes contain no `NSNull`, file still loads) plus a CrowCore test pinning the null-decode failure itself.

**RPCs are named `agents-get`/`agents-set`, not `agents-list`.** The existing web `list-agents` is one word-order away and *also* returns a top-level `agents` key — but as an **array**. `web/settings.js:142` iterates it, so a collision there would fail silently (blank picker, no error). `list-agents` is left completely untouched, and the new `available` deliberately carries **no `default` field**: that flag on `list-agents` means the *registry* default (first-registered), a different concept from `default_agent_kind` sitting inches away.

## Gating

Un-gated in `RPCWebSocketHandler.localOnlyDenial`, same rationale as `notifications-*`: agent kinds name a harness rather than carry a credential, and Settings → General already edits them remotely through the un-gated `set-config`. Note `defaults.binaries` — the *path* to an agent binary — stays local-only; choosing among binaries the daemon already found is not the same as telling it to execute a new one. Pinned by `agentRPCsAreAllowedRemotely`, since `localOnlyDenial`'s `default: nil` would swallow a regression.

## Also in this PR

- `SessionKind: CaseIterable` — drives the `effective` map, the role-name error messages, and ArgumentParser's `--clear` value list (`--help` now shows `values: work, review, job, manager`).
- Doc fixes for the surface this PR now owns: `configuration.md` said "four coding agents" and omitted `antigravity` from the value lists and kind table; `README.md` pointed at Settings → **Automation** instead of General.

## Tests

24 pure (CrowEngine) · 12 handler end-to-end against the real router and a real `config.json` (CrowDaemon) · 14 CLI parse (CrowCLI) · 5 config round-trip (CrowCore) · 1 remote-gating pin. **All 16 packages pass**, plus `scripts/check-notification-events.sh`.

`AgentsHandlerTests` carries a registry-hygiene note: `AgentRegistry.shared` is a process-wide singleton with no unregister, seeded additively by whichever tests ran first, and Swift Testing runs suites in parallel — so the tests assert `available` *contains* Claude Code rather than its count, and use a never-registered sentinel kind for rejection cases.

## Verification notes / known sharp edges

- **Not smoke-tested against a live socket.** `crowd` refuses to start a second store writer (and macOS resolves App Support from the passwd entry, so a `HOME` override does not isolate it), so running this build alongside the existing daemon was not possible without stopping live sessions. The handler path *is* covered end-to-end by `AgentsHandlerTests` through the real `makeCommandRouter` against a real `config.json` — only the socket transport itself, shared boilerplate with every other verb, is untested here.
- **Availability is decided at daemon boot** and is binary-dependent. Installing an agent needs a `crowd` restart before it can be selected; and a configured kind whose binary later left `PATH` stays in config, still resolves in `effective`, and yields unlaunchable sessions — which is why both verbs warn on stderr when `effective` names a kind absent from `available`. Documented in the Notes.
- **Concurrent-writer race with an open web Settings modal** — the modal holds a whole-`AppConfig` working copy and PUTs the blob on Save, so a CLI write in between is silently reverted. Pre-existing and identical for `telemetry`/`cleanup`/`ui`/`notifications`; out of scope here, but it is the first thing worth asking about a new writer.
- **Every `agents set` rewrites `config.json` even on a no-op** (e.g. clearing an already-absent role), pushing a `configReloaded` to open browsers. Consistent with the other settings verbs; short-circuiting would mean `"saved": true` sometimes not writing, a worse contract.
- **The ticket's `KNOWN_UI_ONLY` checkbox refers to infra that does not exist** — grep returns zero hits repo-wide. The only extant analogue is `docs/cli-reference.md`'s "Not Exposed on the CLI" prose section, which lists only the Jira credential and needed no edit. Not inventing the constant here.


dgershman (changes requested):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Unavailable agent kinds are rejected before `mutateConfig` — no lock, no write, no mtime bump (`AgentsRPC.decodeAgentKind`, `AgentsHandlerTests.rejectionsLeaveConfigByteIdentical`).
- `--clear` / `applyByKind` removes keys rather than writing JSON null, avoiding the whole-`config.json` undecodable failure mode; pinned at pure, handler, and Core layers.
- `mutateConfig` still refuses to overwrite an undecodable config; handler test covers it.
- Remote gating choice is sound and pinned: `agents-*` name a harness, not credentials; `defaults.binaries` stays local-only (`agentRPCsAreAllowedRemotely`). Naming `agents-get`/`agents-set` avoids colliding with web `list-agents`.
- Pure `[AvailableAgent]` snapshot keeps the registry gate unit-testable without the process-wide singleton.

**Concerns:**
- None that block merge. Changing Manager/review/job agents remotely is already possible via un-gated `set-config`; this PR does not widen that surface.

### Code Quality
- **Yellow — `docs/configuration.md:114`:** This PR’s README, CLAUDE.md, and code comments correctly say **Settings → General**, but the paragraph this PR edited still says **Settings → Agents** (no such top-level tab — Agent pickers live under General). Align the wording.
- **Green:** Consider rejecting `by_kind.review = antigravity` (and default→review resolution) at the agents-set gate to match `shouldRefuseReviewHandoff`. Create-path `autoLaunchCommand(.review)` returns `nil` today so this is a footgun/UX issue, not an RCE; web Settings already allows the same value.
- **Green:** Daemon `decodeAgentKind` does not trim whitespace; CLI only rejects fully blank kinds, so `--default "codex "` fails with “not available” rather than a blank-kind error.
- **Green:** Tests are thorough (pure + handler + CLI parse + Core round-trip + remote-gating pin). Locally: AgentsRPCSupportTests 27/27, AgentCommandParsingTests 15/15, Core agent-selection 5/5, AgentsHandlerTests 11/11.

### Summary Table

---

dhilgaertner:
Thanks — all four findings addressed in 05de5d3. One place I deliberately stopped short of the suggestion; reasoning below.

### 🟡 Settings tab wording — fixed

`docs/configuration.md:114` now says **Settings → General**.

I also fixed the same wrong label in `SessionService.launchAgent`'s terminal message (`SessionService.swift:1000`), since that is precisely the line a user sees after configuring a role to an agent that can't run it — leaving "Settings → Agents" there would have sent them hunting for a tab that doesn't exist at exactly the moment they need it. It now names Settings → General *and* the new CLI verb:

> ⚠️ Crow: agent "Antigravity" cannot launch a review session. Switch the agent for this action type (Settings → General, or: `crow agents set --review <kind>`) or pick a session kind this agent supports.

### 🟢 Rejecting review→antigravity — implemented, with one scope limit

I verified your read before implementing rather than taking it at face value, and it holds: `autoLaunchCommand(.review)` returns `nil` for Antigravity, and `prepareReviewClone` **already** strips `.agents/` gated on `reviewAgentKind == .antigravity` — keyed on the *config-resolved* kind, so it fires for exactly the configuration `agents set --review antigravity` would write. Footgun, not RCE, as you said.

`decodeByKind` now rejects it, delegating to `SessionService.shouldRefuseReviewHandoff` rather than re-deriving the rule — your "to match" framing is the whole point, so sharing the predicate means the two gates can't drift. It would be incoherent for `crow handoff-agent --agent antigravity` to be refused on a review session while `crow agents set --review antigravity` quietly configured every future one.

**Where I stopped short:** you also suggested gating the default→review resolution. I left `--default` exempt on purpose. Validating the *resolved* outcome means reading state the caller isn't touching — so a pre-existing `defaultAgentKind: antigravity` (settable from web Settings, which doesn't run this gate) would make **every later `agents set` fail, including patches to unrelated roles**. `crow agents set --work codex` erroring about review would be a worse failure than the one being prevented. The rule I settled on is "validate what the caller is changing, not what the config already is."

That case isn't silent, though — it surfaces at launch time through the terminal message above. If you'd rather have it rejected up front despite the unrelated-patch cost, say so and I'll flip it; it's a two-line change.

Both the exemption and its rationale are pinned by a test (`defaultAgentKindIsNotSubjectToTheRoleCapabilityGate`) and documented in the cli-reference Notes, so it reads as a decision rather than an oversight.

### 🟢 Whitespace trimming — fixed

`--work "codex "` now trims instead of bouncing off the daemon as `'codex ' is not an available agent`. Trimming happens before both the blank check and the send, so validation and the wire payload can't disagree. Verified against the running daemon:

```
$ crow agents set --work " codex "
Error: Error -32601: Unknown method: agents-set     # reached the socket; trimmed kind accepted client-side
$ crow agents set --work "   "
Error: --work must not be blank.                    # still caught locally
```

(The `-32601` is just the *running* daemon being the pre-PR build — it does incidentally confirm the socket transport works end-to-end, which is the one seam I'd flagged as untested.)

I left the daemon's `decodeAgentKind` strict/literal rather than trimming server-side: cleaning up shell-quoting artifacts is the CLI's job, and a non-CLI RPC caller sending `" codex "` has a bug worth surfacing.

### 🟢 Tests

+5 CrowEngine, +2 CrowCLI. The capability gate is tested at the **pure** layer, where `available` is injected as a value — registering Antigravity in the daemon suite would seed the process-wide `AgentRegistry.shared`, which has no unregister, for every parallel suite. One test asserts the gate matches `shouldRefuseReviewHandoff` across every role×kind pair, so if that predicate grows a case, this gate follows automatically.

I also extracted `AgentsSet.sentParams()` from `run()` so the wire payload is assertable without a socket, and used it to pin the pre-existing clear dedup/ordering that previously had no coverage.

CrowCore 489 · CrowEngine 439 · CrowDaemon 237 · CrowCLI 253 — all pass; `make daemon` clean.

dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Unavailable kinds are rejected before `mutateConfig` — no lock, no write, no mtime bump (`AgentsRPC.decodeAgentKind`, `AgentsHandlerTests.rejectionsLeaveConfigByteIdentical`).
- `--clear` / `applyByKind` deletes keys rather than writing JSON null, avoiding the whole-`config.json` undecodable failure mode; pinned at pure, handler, and Core layers (`agentsByKindWithANullValueMakesTheWholeConfigUndecodable`).
- `mutateConfig` still refuses to overwrite an undecodable config; handler test covers it.
- Remote gating is sound and pinned: `agents-*` name a harness, not credentials; `defaults.binaries` stays local-only (`agentRPCsAreAllowedRemotely`). Naming `agents-get`/`agents-set` avoids colliding with web `list-agents`.
- Role capability gate rejects `--review antigravity` via the same `shouldRefuseReviewHandoff` predicate as `handoffAgent`, so the two surfaces cannot drift.

**Concerns:**
- None that block merge. Changing Manager/review/job agents remotely was already possible via un-gated `set-config`; this PR does not widen that surface.

### Code Quality
Prior review (Yellow: Settings → Agents wording; Green: review-capability gate + kind trimming) is addressed in `05de5d3`.

- **Green:** `--default antigravity` remains ungated by design (documented); a CLI stderr warning when `effective.review` resolves to a review-incapable agent would mirror `warnAboutStrandedRoles` without breaking patch semantics — optional polish.
- **Green:** Concurrent Settings-modal whole-config PUT vs CLI PATCH race is pre-existing and shared with other settings verbs; out of scope.
- Tests: AgentsRPCSupportTests 32/32, AgentCommandParsingTests 17/17, AgentsHandlerTests 11/11, Core null-decode pin, `agentRPCsAreAllowedRemotely` — all pass.

### Summary Table

---

dhilgaertner:
Rebased onto `main` (now `0071dd6`). Three commits landed underneath while this was open, and two of them changed things this PR depends on — flagging the semantic fallout, not just the textual conflicts.

### #880 (agent pickers: surface-but-disable) — payload shape changed

This is the one worth re-reviewing. #880 fixed the web pickers hiding off-PATH agents entirely, because "a shipped, fully-capable agent silently didn't appear, reading as broken/missing rather than 'not installed'."

`crow agents list` had exactly that bug — it built its list from `allAgents()`, which is still launchable-only after #880, so an uninstalled Codex simply wouldn't appear and you couldn't tell whether it existed. Shipping that immediately after the project decided it was a bug seemed wrong, so I aligned:

- The list is now built from `AgentRegistry.agentListings()` — the shared projection #880 added *precisely* so surfaces don't duplicate selection logic — instead of re-deriving it.
- Renamed `available` → **`known`**, since it no longer *is* the available set; availability moved to a per-row flag. Rows now carry `{kind, name, binary, available}`, matching `list-agents`' new fields.

```json
"known": [
  { "kind": "claude-code", "name": "Claude Code", "binary": "claude", "available": true },
  { "kind": "antigravity", "name": "Antigravity", "binary": "agy",    "available": false }
]
```

**The launch gate did not widen.** `decodeAgentKind` now gates on `available == true`, which matches the registry exactly — an unavailable agent never enters the launchable map, so `registeredKind`/`agent(for:)` still refuse it. A listed-but-uninstalled kind gets its own error naming the binary, rather than an "expected one of" list that would be baffling when the agent is sitting right there in `crow agents list`:

> `'antigravity' is known but not installed (by_kind.work) — agy was not found on PATH when crowd started. Install it and restart crowd to select it.`

New tests pin both halves: the row is listed, and it is not selectable.

I kept `binary` as the PATH token only, never the resolved absolute path — same reasoning as #880's `KnownAgentListing`, so the payload doesn't leak the daemon host's install layout.

### #882 (generated CLI reference)

`docs/cli.md` is now generated with tests that fail on any undocumented verb. Ran `make docs` — 97 commands, +58 lines for the `agents` group; `crow generate-docs --check` exits 0.

I kept the hand-written Agent Commands section in `docs/cli-reference.md` rather than deleting it as redundant: #882's contract is that every verb in the generated page also appears in the guide, and the guide carries what a flag table can't — delete-not-null, the boot-time availability snapshot, the registry-vs-configured "default" distinction.

### #887 (defaults CLI)

Pure textual conflict — same insertion points in `CrowCommand.swift`, `RPCHandlers.swift`, `DaemonSecurityTests.swift`, `CLAUDE.md`, `cli-reference.md`. Both verb groups belong; kept both. Verified `crow --help` lists `agents` and `defaults` side by side.

### Verification

All four affected packages, post-rebase: **CrowCore 489 · CrowEngine 467 · CrowDaemon 257 · CrowCLI 289** — all pass. `make daemon` clean, `make docs` idempotent.

The three earlier review fixes are unchanged and still under `9b1bbf3`; the `--default`-exemption reasoning in my previous comment stands. Happy to flip that if you'd still prefer it gated.

dgershman (changes requested):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Unavailable kinds are rejected before `mutateConfig` — no lock, no write, no mtime bump (`AgentsRPC.decodeAgentKind`, `AgentsHandlerTests.rejectionsLeaveConfigByteIdentical`).
- `--clear` / `applyByKind` deletes keys rather than writing JSON null, avoiding the whole-`config.json` undecodable failure mode; pinned at pure, handler, and Core layers.
- Role capability gate rejects `--review antigravity` via the same `SessionService.shouldRefuseReviewHandoff` predicate as `handoffAgent`, so the two surfaces cannot drift.
- `binary` in the payload is the PATH token (`launchCommandToken`), not a resolved absolute path — no host install-path leak.
- Remote gating is sound and pinned: `agents-*` name a harness, not credentials; `defaults.binaries` stays local-only (`agentRPCsAreAllowedRemotely`). Naming `agents-get`/`agents-set` avoids colliding with web `list-agents`.

**Concerns:**
- None that block merge. Changing Manager/review/job agents remotely was already possible via un-gated `set-config`; this PR does not widen that surface.

### Code Quality
- **Yellow — `docs/configuration.md:112-114`:** Adding `antigravity` to the kind table left a blank line between the `opencode` and `antigravity` rows. In GFM that terminates the table, so `antigravity` renders as a broken one-line table instead of a fifth row — undoing the point of the doc fix. Remove the blank line so all five kinds stay in one table.
- **Green:** `RPCHandlers.swift:1807` still refers to the response field as `` `available` `` in the comment above `agents-get`, but the payload key is now `known` (with per-row `available`). Cosmetics only.
- **Green:** `--default antigravity` remains ungated by design (documented); a stderr warning when `effective.review` resolves to a review-incapable agent would mirror `warnAboutStrandedRoles` — optional polish.
- Prior review findings (Settings → Agents wording, review-capability gate, kind trimming) are addressed. Tests: AgentsRPCSupportTests 35/35, AgentCommandParsingTests 17/17 pass locally.

### Summary Table

---

PR #887, opened 2026-07-28
Defaults CLI: manage AppConfig.defaults (parity)

### LINKED ISSUE #810: [parity] Defaults CLI: manage AppConfig.defaults (provider, branchPrefix, exclude/ignore lists, binaries)
Milestone: **CLI control-plane parity**. Area: **Defaults** (Settings → Workspaces/Automation).

Add `crow defaults get|set` covering `AppConfig.defaults` (`ConfigDefaults`):
- provider, branchPrefix (validate via `isValidBranchPrefix`), excludeReviewRepos[], excludeTicketRepos[], ignoreReviewLabels[]
- binaries map — incl. the `corveil` binary path (General tab "Path to corveil binary"); local-only value

dhilgaertner:
Closes #810

Adds `crow defaults get|set` for the `AppConfig.defaults` subtree — the last Settings surface with no CLI verb. Until now the only headless way to change it was `set-config`, which replaces the entire `AppConfig` blob, so a script had to re-send workspaces, jobs, gateway URLs and credential shells just to exclude one repo from the review board — and raced the web Settings modal while doing it.

```bash
crow defaults get
crow defaults set --provider gitlab --cli glab
crow defaults set --add-exclude-review-repo 'acme/legacy-*' --remove-ignore-review-label wip
crow defaults set --clear-exclude-ticket-repos
crow defaults set --binary corveil=/opt/corveil/bin/corveil    # NAME= removes
```

## Decisions worth reviewing

**`--cli` ships even though the ticket doesn't list it.** `GitManager.swift:38-39` reads `provider` and `cli` as an independent pair, and the web UI has no `cli` field at all — so `--provider gitlab` would otherwise leave `cli` at `"gh"` with no CLI way to fix it. They stay independent: a crossed pair is reported via `provider_cli_mismatch` rather than auto-paired, since writing a field the caller didn't name would break the "only what you pass changes" contract.

**Lists are edited incrementally, not replaced.** `--add-` / `--remove-` compose (remove first, so naming a value in both means "ensure it's there"); `--clear-` is exclusive with the other two *per list*. Matching is case-insensitive because that's what the consumers do — `repoMatchesPatterns` lowercases both sides, so an exact-match remove would answer `saved: true` while the repo stayed excluded.

**`--binary` merges rather than replaces** — the map is shared by `AgentKind`-keyed discovery and tool-name-keyed installers, so a replace would drop the other caller's keys. It returns `restart_required`: both consumers are boot-time, and since nothing re-scaffolds on a config change, a *removal* also leaves the stale `.claude/bin` symlink shadowing `PATH` until the next launch. A non-executable path is saved with a stderr warning rather than rejected — `Scaffolder` already skips it, but its only signal today is an `NSLog` the CLI user never sees.

**Rejected binary names, all load-bearing rather than tidiness:**
- `crow` — `installCrowCLISymlink` re-points it at the app's own CLI every launch and `managedBinarySymlinks` exempts it from reaping, so an override there can never take effect. Accepting it would report a saved write that does nothing.
- blank — `Scaffolder.swift:251` builds `binDir.appendingPathComponent(name)`, which for `""` **is `binDir`**, and line 262 then `removeItem`s it and symlinks over it.
- containing `/` or `\` — escapes the bin dir, where the reap loop (which walks `contentsOfDirectory`) can never clean the orphan up.

## Local-only gating

`defaults-set` is the first granular settings verb that can reach `defaults.binaries` — which `localOnlyDenial`'s doc comment calls "the sole local-only config field", and whose `default: return nil` would have made a new method remotely reachable **with no compile error**.

Gated on the **presence** of a `binaries` param, not a diff against disk like `set-config`. `set-config` needs its diff only because it is a whole-blob replace whose remote Settings saves re-echo the binaries they just read (`setConfigUnchangedBinariesAndJobsIsAllowed` pins that); a PATCH has no such echo. A diff here would additionally have to re-implement `mergeBinaries`' `""`-deletes semantics *inside a security predicate*, against a config read outside `withConfigLock`.

`defaults-get` stays **un-gated**: `SettingsSecrets.strippedForTransport` blanks the Jira token, webAuth hash/salt and gateway headers but *not* `defaults.binaries`, so un-gated `get-config` already ships binary paths to remote browsers and `settings.js:506` renders the corveil path read-only for them. The rule recorded in the ledger — *gate a read only when it returns what stripping would have removed.*

## Shared validation

The accepted provider/CLI spellings, the reserved binary name and the binary-name predicate live on `ConfigDefaults` next to the existing `isValidBranchPrefix`, so the CLI and the handler validate against one source rather than the hand-kept copy `validQuickActions` warns about.

## Tests — 2072 passing across every package

- `DefaultsRPCSupportTests` (25) — patch decoding, `StringListPatch.apply` ordering/casing, binaries rejections, `mergeBinaries`, `restartRequired`
- `DefaultsHandlerTests` (13) — real router against a real `config.json`: patch isolation, list add/remove/clear, binaries merge, unrelated config (workspaces/jobs/jiraCredential/notifications and the two get-only fields) surviving, empty params rejected **without creating config.json**, undecodable config never overwritten
- `LocalOnlyRPCGateTests` (+3) — `defaults-get` un-gated; `defaults-set` un-gated per non-binaries field; `defaults-set` denied for `binaries` across `.object`, `.object([:])`, `.null` and `.string("oops")` — the last two pin "presence, not decode"
- `DefaultsCommandParsingTests` (21) + `ValidationTests` (+8)

Verified end-to-end that the CLI puts the right thing on the wire (trim + case-insensitive dedupe, first-`=`-only split so `/opt/codex=v2/bin/codex` survives, `~` expansion, `cursor=` → `""`, untouched lists emitting no params at all) and that all three stderr advisories fire.

## Known limitation — documented, not fixed here

An open web Settings modal snapshots the whole config on open and re-sends it on Save, so saving a stale modal reverts a concurrent CLI write. Pre-existing and identical for `crow notifications set` / `crow ui set` / `crow job add`, but more likely to be hit here since `defaults` spans two Settings tabs. Noted in `docs/cli-reference.md`; the real fix (an mtime precondition on `set-config`, or `settings.js` reloading on `configReloaded` when not dirty) wants its own PR.

Also documented: the review board filters on `effectiveExcludeReviewRepos` (defaults ∪ every workspace's own), so `--clear-exclude-review-repos` doesn't unhide a repo a workspace excludes.

## Checklist deviations

- **Parity harness / `KNOWN_UI_ONLY`** — still doesn't exist (#807 open, zero hits repo-wide), same as #870/#871/#873 reported. `LocalOnlyRPCGateTests` remains the enforced covered set and has been extended.
- **`README.md` unchanged** — it has no per-command docs, only a link to `docs/cli-reference.md`, which is where the reference went.


dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- `defaults-set` correctly gates `binaries` on key *presence* (`RPCWebSocketHandler.swift` ~186–210), including null/wrong-typed payloads — avoids the “decode then deny” hole the tests pin.
- Binary-name validation (`ConfigDefaults.isValidBinaryName`) blocks blank names, path separators, `.`/`..`, and reserved `crow` — the load-bearing Scaffolder hazards (bin-dir wipe / escape) are covered.
- Absolute-path requirement (literal `/` prefix, not `isAbsolutePath`) and CLI-side `~` expansion prevent silently inert overrides.
- `mutateConfig` refuses to overwrite an undecodable `config.json` (handler tests cover this).
- List decode is strict (no silent `compactMap` drop of non-strings); credentials stay out of `defaultsJSON`.

**Concerns:**
- None that are Red/Yellow. Binary paths remain visible via un-gated `defaults-get`, matching existing `get-config` + `SettingsSecrets` behavior — intentional and documented.

### Code Quality
- 🟢 **Case-sensitive binary keys on case-insensitive APFS** (`DefaultsRPCSupport.swift` / `parseBinaryOverrides`): `corveil` and `Corveil` can both land in the map and fight over one symlink. Lists were deliberately case-folded; binaries were not. Consider lowercasing names on write (or rejecting duplicates case-insensitively) for macOS parity.
- 🟢 **`isValidBinaryName` is separator-focused only** (`AppConfig.swift` ~647–650): control characters / embedded whitespace still pass. Low risk (local-only write), but tightening to a plain filename charset would match the “plain name like corveil” error copy.
- 🟢 **Stale comment** in `Validation.swift` ~140–141 still says CrowCLI doesn’t depend on CrowCore (for `validQuickActions`), but this PR correctly pulls `ConfigDefaults` from CrowCore — worth updating so the next reader doesn’t reintroduce a hand-kept list.
- 🟢 Targeted tests: `DefaultsRPCSupportTests` (25), `DefaultsCommandParsingTests` (22), `DefaultsHandlerTests` + `LocalOnlyRPCGateTests` extensions — all passed locally (Swift Testing). Package exit code 1 is the known arm64/x86_64 XCTest bundle mismatch, not a failing assertion.

### Summary Table

---

Issue #200, opened 2026-04-23
Manager remote control doesn't activate session until tab is focused


dgershman:
## Bug

When using the Manager tab to remote-control a session (e.g. via `crow send`), the session doesn't actually proceed/activate until the user manually logs in and switches to the active tab in the Crow UI.

## Expected Behavior

Sessions controlled via the Manager should activate and proceed without requiring the user to manually focus the tab or log in.

## Steps to Reproduce

1. Use the Manager tab to send commands to a session via `crow send`
2. Observe that the session doesn't proceed
3. Manually switch to the session's tab in the Crow UI
4. Session now proceeds as expected

## Impact

This defeats the purpose of remote control from the Manager — the user has to babysit each session by switching to its tab.

dhilgaertner:
I believe this is fixed with the new TMUX feature but I will pick up this ticket to test.

dhilgaertner:
Fixed for me with TMUX

PR #892, opened 2026-07-28
Stop blocking I/O on the MainActor from wedging the daemon's RPC surface

### LINKED ISSUE #874: Blocking NSLog on the MainActor can wedge the daemon's RPC surface (reload-tmux-config reproduces it)
## Summary

A blocking `NSLog` inside `MainActor.run` can wedge the daemon's entire MainActor-bound RPC surface indefinitely. When `crowd`'s controlling tty stops draining, `writev` to stderr blocks in the kernel, and because the log call happens on the MainActor, every other MainActor-bound RPC queues behind it until the tty drains.

`reload-tmux-config` is the reliable reproducer — it logs unconditionally on its success path — but the bug is not specific to that method. Any handler that logs from the MainActor has the same exposure.

Found while verifying #818 / #867. Not caused by that PR (it adds no server code); the new CLI verb just made an existing failure mode easy to hit.

## Evidence

`sample` of a live `crowd` (pid 19389) captured mid-hang, while `crow reload-tmux-config` was blocked:

```
closure #9 in makeCommandRouter(...)          RPCHandlers.swift:342
  static MainActor.run<A>(resultType:body:)
    TmuxBackend.reloadBundledConfig()          TmuxBackend.swift:1170
      specialized withVaList<A>(_:_:)          (Foundation)
        _NSLogv                                (Foundation)
          _CFLogvEx3 -> _CFLogvEx2Predicate -> __CFLogCString
            _logToStderr                       (CoreFoundation)
              writev                           (libsystem_kernel)   <- blocked
```

`TmuxBackend.swift:1170` is the success-path telemetry line:

```swift
do {
    try ctrl.run(["source-file", confURL.path])
    NSLog("[CrowTelemetry tmux:config_reloaded_by_user path=\(confURL.path)]")   // <- line 1170
    return nil
}
```

`source-file` has already returned successfully. The hang is entirely in the log call.

The daemon's stdout/stderr are its controlling tty:

```
$ lsof -p 19389 -a -d 1,2
crowd  19389  ...  1u  CHR  16,0  0t8096046  695  /dev/ttys000
crowd  19389  ...  2u  CHR  16,0  0t8096046  695  /dev/ttys000
```

Writing to a tty whose output queue is full and not being drained blocks in the kernel. That is the whole mechanism.

## Impact

The blocking write happens **inside `MainActor.run`**, so it does not just stall the one RPC — it stalls the MainActor. Observed knock-on effect: `open-terminal` timed out at the CLI's 30s read timeout on one attempt and then succeeded on retry, because it was queued behind the MainActor stuck in `NSLog`.

This is latent for any long-running `crowd` started from a terminal (the normal dev-mode launch, e.g. `crow-desktop/src-tauri/../../.build/debug/crowd`). A daemon started under launchd with logs redirected to a file is not affected in the same way.

## Reproduction

1. Start `crowd` attached to a terminal (stdout/stderr = tty).
2. Get that tty into a state where its output queue is not drained (terminal stopped/`Ctrl-S`, pane not being read, consumer gone but the tty still open).
3. `crow reload-tmux-config` → blocks until the tty drains; the CLI gives up at its 30s read timeout.
4. Concurrently, any other MainActor-bound RPC (`open-terminal`, `open-in-vscode`, …) also stalls.

## Ruled out

These were checked and are **not** the cause:

- **tmux itself.** `tmux -S <sock> source-file <conf>` completes in 16ms directly, and in ~50ms with both stdout and stderr as pipes drained to EOF. `tmux list-sessions` responds instantly throughout.
- **A wedged `TmuxController` / control-mode connection.** Other daemon-router tmux verbs (`rename-terminal`) respond in ~20ms while `reload-tmux-config` is hanging.
- **Subprocess pipe-EOF deadlock in `TmuxController.run`** (a child inheriting the write end so `readDataToEndOfFile` never sees EOF). Plausible on inspection, but disproven by the piped-source-file test above and by the sampled stack, which never reaches `group.wait()` on this path.

## Proposed fix

Narrow (fixes the reported hang): telemetry/diagnostic logging must never run on the MainActor where it can block an RPC. Either hop the `NSLog` calls off the MainActor, or route telemetry through a non-blocking logger (`os_log`/`Logger`, which does not do a synchronous tty write) instead of `NSLog`.

Worth considering alongside it: `crowd` could open its stderr non-blocking, or redirect to a log file by default when not started under launchd, so a stalled terminal can never apply backpressure to the daemon.

## Secondary finding (same area, separate hang)

The same `sample` showed a second thread stuck independently:

```
closure #1 in TmuxBackend.startManagerExitMonitor(id:pollInterval:onExit:)  TmuxBackend.swift:983
  TmuxController.displayMessage(target:format:)                             TmuxController.swift:373
    TmuxController.run(_:timeout:)                                          TmuxController.swift:80
      _dispatch_group_wait_slow
```

`TmuxController.run` arms a `ProcessWatchdog` with a timeout and waits for process exit, but the subsequent pipe drain is an **un-timed** `group.wait()`:

```swift
let watchdog = ProcessWatchdog(p, timeout: timeout)
done.wait()
watchdog.cancel()
group.wait()          // TmuxController.swift:80 — no timeout
```

If either reader task never observes EOF, the watchdog killing the child does not rescue this wait — it blocks forever, defeating the timeout it appears to enforce. `displayMessage` uses `max(defaultTimeout, 10.0)`, so the intent is clearly bounded. Suggest `group.wait(timeout:)` with a deadline tied to the same budget.

Filing both here since they surfaced together; happy to split the second into its own issue if preferred.

dhilgaertner:
Closes #874

## The bug, and its blast radius

`crowd` serves every CLI and web client through RPC handlers that hop to the MainActor — roughly half of the 64 in `RPCHandlers.swift`. That serialization is what makes concurrent `crow` use safe, and it also means the main actor is a shared single-threaded resource for all clients at once. Anything that blocks it stalls the **whole daemon**.

`NSLog` blocks it: CoreFoundation `_logToStderr` → `writev(2)` on the controlling tty, which blocks in the kernel with no timeout when that tty's output queue is full and unread. The `sample` in the issue caught it exactly.

`reload-tmux-config` was the reproducer, not the bug. It just logs unconditionally on its success path. There were **199 `NSLog` calls across 31 files**, all in `crowd`'s graph, and the heaviest users — `SessionService` (65), `TmuxBackend` (21), `IssueTracker` (12) — are `@MainActor` types.

## What changed

**A non-blocking sink.** `CrowLog` mirrors to `os_log` **on the calling thread** (a bounded memcpy into the process trace buffer — cannot block) and enqueues onto a bounded backlog drained by a dedicated thread. Past the cap, lines are dropped and counted rather than queued, so a wedged tty costs a gap in the log instead of a hang. `flush(timeout:)` covers `exit`/`execv`.

Three uncoordinated sinks collapse into it: bare `NSLog`, `CrowDaemon.log`'s synchronous `FileHandle.standardError.write`, and `CrowLog.automation` — which mirrored to `NSLog`, making CROW-782's durable-logging fix a wedge vector on every automation decision.

**All 199 sites migrated, with a CI guard** (`scripts/check-no-nslog.sh`) so site #200 doesn't land next week. It runs as a third gate in the `parity` target and its CI job, alongside the CLI/RPC parity and notification-catalog checks (CROW-807, CROW-768).

**The secondary finding, plus a root cause the issue didn't identify.** `PTYProcess` spawned with `POSIX_SPAWN_SETSID` only. `posix_spawn` copies the whole fd table into the child without `CLOEXEC_DEFAULT`, and these children are the cockpit attach clients — the longest-lived processes Crow spawns, one per browser session. A `TmuxController.run()` pipe write end captured that way keeps its pipe from ever reaching EOF, which is why the drain never completed. `run()` now also bounds the drain (0.25s) as defense in depth.

## Things worth a reviewer's attention

**A drain timeout does not throw.** The child has exited and its status is authoritative, and `TmuxError.timedOut` drives `TmuxBackend.onUnresponsive`, which offers the user *"Restart tmux server"* — destroying their terminals over our own fd hygiene. It logs instead.

**Readers use raw `read(2)`, and that is load-bearing.** Neither obvious alternative works: `readDataToEndOfFile()` only returns at EOF, and Darwin's `read(upToCount:)` blocks for the *full* count — measured, both leave the buffer **empty** on a drain timeout rather than holding partial output. A single `read(2)` returns as soon as bytes are available, so once the child exits every byte it wrote is captured even though EOF never arrives. `PipeBox` is now `NSLock`-guarded, since a bounded wait means the reader can still be running when the caller snapshots.

**One comment was rewritten rather than corrected.** Three claimed `run()` blocks on `waitUntilExit()` and pumps the run loop — untrue since #653. Two were merely stale. The third justifies `ensureCockpitSession`'s adopt branch, and its stated in-process re-entrancy really is gone — but the branch is still required, because `TerminalCockpit.ensureSession` in `crowd` runs the identical `hasSession()` → `newSessionDetached` against the same socket (#330) on every startup. That's a cross-process TOCTOU no in-process argument closes. Deleting the branch on the strength of "no more re-entrancy" would reintroduce #326's blank terminals.

**Two bugs fixed incidentally.** 92 of the 199 sites passed interpolated, externally-influenced text (branch names, PR titles, raw `gh` stderr, user paths) as `NSLog`'s *format string*, where a stray `%` reads garbage varargs. And two passed a 64-bit `Int` to `%d`, working only by ABI accident. `CrowLog.info(_ message: String)` closes both structurally.

## Verification

`make build` and `make test` clean — **2011 tests, 0 failures**, both guard scripts passing.

The mechanism was reproduced directly against a real pty whose output queue was full and unread:

| | result |
|---|---|
| 6000 × `CrowLog.info` | returned in **6 ms**, 0 dropped |
| 1 × `NSLog` (same conditions) | **blocked indefinitely** — hung the test process past a 3 s timeout |

New tests need no tmux binary, so they run even where the integration suites skip — worth noting since `CrowTerminal` is excluded from PR CI entirely (ADR 0007), along with `CrowEngine`, `CrowTelemetry`, and `CrowDaemon`. **129 of the 199 migrated sites are in packages this PR's CI will not compile**, so all of them were built and tested locally. The drain test asserts a lower bound as well as an upper one so it can't pass vacuously, and the PTY test was confirmed to fail with the `CLOEXEC_DEFAULT` flag reverted.

## Known residual and follow-ups

On the drain-timeout path the two reader blocks stay parked until the fd holder dies — 2 leaked worker threads per event, self-healing on disconnect. Strictly better than before, which leaked those *plus* the caller's thread permanently. Eliminating it needs a `DispatchSourceRead` drain (in-repo precedent at `PTYProcess.startReading()`); that's a real rewrite of `run()`, deliberately not smuggled in here.

`reloadBundledConfig` moved off the MainActor, but it's 1 of ~24 `@MainActor` entry points on `TmuxBackend` that call `TmuxController` synchronously — several are sync `throws` APIs that can't go `async` without touching every caller. Bounding `run()` is the actual fix. A follow-up should move the rest: `TmuxBackend.runShortCommand` still calls `waitUntilExit()` on the MainActor at `:1498` (the nested-run-loop SIGSEGV hazard #653 removed from `TmuxController`), as does `TmuxOrphanReaper.swift:76`. Separately, `~/Library/Logs/crow/crowd.log` has no rotation — pre-existing, and each line is now slightly longer.

Conventions and trade-offs are recorded in [ADR 0017](docs/adr/0017-non-blocking-logging.md).

## Rebase notes

Rebased twice. Conflicts were structural rather than semantic:

- **#883** moved `make test`'s drift gates into a new `parity` target, so `check-no-nslog.sh` became a third gate there and the standalone `guards` job I'd added folded into the new `parity` job. #883 also claimed ADR 0016, so this one is now 0017.
- **#893** rewrote the auto-rebase region of `IssueTracker` and, independently, moved several of the `NSLog` sites this PR targets to `CrowLog.automation` — the better sink for automation decisions, and a stronger choice than the `CrowLog.info` this PR would have given them. Resolved by taking #893's version wholesale and re-running the migration on the 4 `NSLog` calls left in it. `IssueTracker` now differs from `main` by exactly those 4 conversions; `deferRebase`, `autoRebaseDeferrals`, and the `outOfSyncAhead`/`outOfSyncDiverged` split are intact.

⚠️ **`scripts/check-cli-parity.sh` fails on `main` as of `0672fd4`, independently of this PR** — `automation-get` / `automation-set` (#884) are registered but absent from `ParityLedger.rpcMethods` (#883). Verified by running the gate in clean `origin/main` worktrees at two different heads. This also fails `CrowDaemon`'s `RPC ledger parity` suite, which is the only failing test on this branch.

Because that step precedes the NSLog step in the `parity` job, **the NSLog gate will not execute in CI until main is fixed.** The fix is two rows next to the existing `agents-*` pair — deliberately not included here, as it belongs to CROW-807:

```swift
.read("automation-get", cli: "automation get"),
.write("automation-set", cli: "automation set"),
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,9 @@ jobs:
       - name: Notification-event catalogs (CROW-768)
         run: bash scripts/check-notification-events.sh
 
+      - name: NSLog ban (CROW-874)
+        run: bash scripts/check-no-nslog.sh
+
   desktop:
     name: Desktop (Tauri) Build
     runs-on: macos-15
@@ -97,6 +100,9 @@ jobs:
       - name: Build crow-desktop (Rust/Tauri)
         run: cargo build --manifest-path crow-desktop/src-tauri/Cargo.toml
 
+  # Cheap cross-cutting invariants that no compiler enforces. These live in
+  # `make test` too, but that only runs locally — before this job, nothing ran
+  # them in CI at all.
   shellcheck:
     name: Shell Lint
     runs-on: ubuntu-latest
--- a/Makefile
+++ b/Makefile
@@ -106,6 +106,8 @@ parity:
 	@./scripts/check-notification-events.sh
 	@echo "==> Checking CLI/RPC control-plane parity (CROW-807)..."
 	@./scripts/check-cli-parity.sh
+	@echo "==> Checking for banned NSLog (CROW-874)..."
+	@./scripts/check-no-nslog.sh
 
 # Regenerate the CLI reference from the commands themselves (CROW-808). Run
 # after adding or changing a subcommand — a stale docs/cli.md fails CrowCLI's
--- a/Packages/CrowAntigravity/Sources/CrowAntigravity/AntigravityHookConfigWriter.swift
+++ b/Packages/CrowAntigravity/Sources/CrowAntigravity/AntigravityHookConfigWriter.swift
@@ -127,7 +127,7 @@ public struct AntigravityHookConfigWriter: HookConfigWriter {
         // worktree than poison the repo. Checked unconditionally so a
         // tracked-but-`rm`'d file is still caught.
         if Self.isGitTracked(worktreePath: worktreePath, relativePath: ".agents/hooks.json") {
-            NSLog("[AntigravityHookConfigWriter] %@/.agents/hooks.json is git-tracked; not writing Crow's session hooks into a committed file. Gitignore/untrack it to enable hook-based state detection for this worktree.", worktreePath)
+            CrowLog.info("[AntigravityHookConfigWriter] \(worktreePath)/.agents/hooks.json is git-tracked; not writing Crow's session hooks into a committed file. Gitignore/untrack it to enable hook-based state detection for this worktree.")
             return
         }
 
@@ -138,7 +138,7 @@ public struct AntigravityHookConfigWriter: HookConfigWriter {
         var root: [String: Any] = [:]
         if let data = FileManager.default.contents(atPath: hooksPath) {
             guard let parsed = try? JSONSerialization.jsonObject(with: data) as? [String: Any] else {
-                NSLog("[AntigravityHookConfigWriter] %@ exists but is unparseable; leaving it untouched (would otherwise drop the user's own hook groups)", hooksPath)
+                CrowLog.info("[AntigravityHookConfigWriter] \(hooksPath) exists but is unparseable; leaving it untouched (would otherwise drop the user's own hook groups)")
                 return
             }
             root = parsed
@@ -202,8 +202,7 @@ public struct AntigravityHookConfigWriter: HookConfigWriter {
                 withJSONObject: root, options: [.prettyPrinted, .sortedKeys])
             try out.write(to: URL(fileURLWithPath: hooksPath), options: [.atomic])
         } catch {
-            NSLog("[AntigravityHookConfigWriter] Failed to rewrite %@: %@",
-                  hooksPath, error.localizedDescription)
+            CrowLog.info("[AntigravityHookConfigWriter] Failed to rewrite \(hooksPath): \(error.localizedDescription)")
         }
     }
 
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
@@ -147,8 +147,7 @@ public struct ClaudeHookConfigWriter: HookConfigWriter {
             try? FileManager.default.setAttributes(
                 [.posixPermissions: 0o600], ofItemAtPath: settingsPath)
         } catch {
-            NSLog("[ClaudeHookConfigWriter] Failed to write gateway env to %@: %@",
-                  settingsPath, error.localizedDescription)
+            CrowLog.info("[ClaudeHookConfigWriter] Failed to write gateway env to \(settingsPath): \(error.localizedDescription)")
         }
     }
 
@@ -179,8 +178,7 @@ public struct ClaudeHookConfigWriter: HookConfigWriter {
             do {
                 try FileManager.default.removeItem(atPath: settingsPath)
             } catch {
-                NSLog("[ClaudeHookConfigWriter] Failed to remove empty settings file at %@: %@",
-                      settingsPath, error.localizedDescription)
+                CrowLog.info("[ClaudeHookConfigWriter] Failed to remove empty settings file at \(settingsPath): \(error.localizedDescription)")
             }
         } else {
             do {
@@ -188,8 +186,7 @@ public struct ClaudeHookConfigWriter: HookConfigWriter {
                     withJSONObject: settings, options: [.prettyPrinted, .sortedKeys])
                 try updatedData.write(to: URL(fileURLWithPath: settingsPath))
             } catch {
-                NSLog("[ClaudeHookConfigWriter] Failed to write updated settings to %@: %@",
-                      settingsPath, error.localizedDescription)
+                CrowLog.info("[ClaudeHookConfigWriter] Failed to write updated settings to \(settingsPath): \(error.localizedDescription)")
             }
         }
     }
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeTrustSeeder.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeTrustSeeder.swift
@@ -1,3 +1,4 @@
+import CrowCore
 import Foundation
 
 /// Seeds per-project trust into `~/.claude.json` so Claude Code skips the
@@ -60,7 +61,7 @@ public enum ClaudeTrustSeeder {
             guard let data = fm.contents(atPath: jsonPath),
                   let parsed = (try? JSONSerialization.jsonObject(with: data)) as? [String: Any]
             else {
-                NSLog("[ClaudeTrustSeeder] %@ exists but is not a JSON object; refusing to modify it", jsonPath)
+                CrowLog.info("[ClaudeTrustSeeder] \(jsonPath) exists but is not a JSON object; refusing to modify it")
                 return .skippedUnparseable
             }
             root = parsed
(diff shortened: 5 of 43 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Migrating off `NSLog` closes the format-string hole where ~92 sites passed interpolated, externally influenced text (branch names, PR titles, `gh` stderr, paths) as the format string (`CrowLog.info(_ message: String)` makes that structurally impossible).
- Automation log keeps owner-only `0o700`/`0o600` hardening, including after atomic create and rotation (covered by tests).
- `POSIX_SPAWN_CLOEXEC_DEFAULT` on Darwin PTY children stops long-lived cockpit attach clients from inheriting unrelated fds (pipe write ends) — root cause of the unbounded drain, and a real fd-leak hygiene fix.
- Bounded backlog + drop-with-count under a wedged sink avoids unbounded memory growth; `os_log` mirror runs on the caller before enqueue/drop so a hung tty does not erase the diagnostic copy.
- `CrowLog.flush` before `exit`/`execv` (via `logAndExit` and re-exec) avoids losing the last line to the async drain.
- CI guard (`scripts/check-no-nslog.sh`) plus Makefile/`guards` job prevent the ban from regressing; also lands the previously local-only notification-event check.

**Concerns:**
- None blocking. `privacy: .public` on `os_log` continues to surface session UUIDs / paths / PR URLs in the unified log — same operational trade-off as the old `NSLog` path, and intentional for hang diagnosis (ADR 0016).

### Code Quality
- `TmuxController.run` bounded drain + incremental `read(2)` + lock-guarded `PipeBox` is well reasoned; not throwing `.timedOut` on drain timeout correctly avoids driving `onUnresponsive` → “Restart tmux server” over fd hygiene.
- `ProcessWatchdog` SIGKILL escalation after ignored SIGTERM closes a real “timeout only in theory” hole on `done.wait()`.
- `reloadBundledConfig` moved off the blocking MainActor path via `Task.detached`; RPCHandlers updated accordingly.
- Tests are strong: CrowLog overflow/flush/permissions/format pins; ProcessDrain asserts both lower and upper bounds and exercises the CLOEXEC regression. Locally: CrowLogTests 10/10, ProcessDrainTests 4/4; `check-no-nslog.sh` clean; CrowCore builds.
- Residual MainActor sync call sites (`runShortCommand`’s `waitUntilExit`, other `TmuxBackend` entry points) and drain-timeout reader-thread leaks are documented honestly as follow-ups — appropriate scope for this PR.
- Minor: `reloadBundledConfig` doc still says the RPC handler wraps it in `MainActor.run`; the handler now `await`s the async method directly (isolation hop is implicit via `@MainActor` on the type).

### Summary Table

Green notes for a later pass (not blocking):
1. Prefer a dedicated queue over `Task.detached` for blocking `Process` waits if more MainActor escapes adopt this shape (cooperative-pool starvation).
2. Align `runShortCommand` with the new SIGKILL watchdog / drain pattern when tackling the remaining MainActor sync call sites.
3. Refresh the stale `MainActor.run` wording on `reloadBundledConfig`.
4. Optionally strengthen the PTY CLOEXEC test with an explicit `ready == 0` success assertion (today success is “no LEAKED within the poll window”).

---

PR #896, opened 2026-07-28
Don't let SAML null nodes wipe out every GitHub GraphQL array parse

### LINKED ISSUE #894: CI failure icons / Fix Checks never surface when viewer.pullRequests contains SAML null nodes
## Summary

Crow stops showing CI status (failing glyphs, **Fix Checks**) for all open PRs whenever the monitored-PRs GraphQL response includes `null` entries in `viewer.pullRequests.nodes` — which GitHub emits for org PRs blocked by SAML enforcement.

The accessible PRs (including ones that are actually red) are present in the same response, but the parser drops the entire list.

## Repro evidence (2026-07-28)

**In-review session:** `crow-809-workspaces-cli` → https://github.com/corveil/crow/pull/885

PR #885 really is failing (`statusCheckRollup.state: FAILURE` — Build & Test, Catalog & Parity Gates). Crow's store has the session in `inReview` with ticket #809, but there is no usable checks state for the UI to key off.

Live `viewer.pullRequests(first: 20)` against the same token Crow uses via `gh`:

| | count |
|---|---|
| total nodes | 15 |
| `null` (SAML-forbidden orgs) | 7 |
| real PR objects | 8 (includes #885 with `FAILURE`) |

`gh api graphql` exits non-zero because of the SAML `FORBIDDEN` errors, so Crow takes the partial-recovery path (`recoverPartialMonitoredPRs`). That path still calls `parseViewerPRs`, which is where the data is lost.

## Root cause

```swift
// GitHubCodeBackend.parseViewerPRs
guard let pullRequests = viewerObj?["pullRequests"] as? [String: Any],
      let nodes = pullRequests["nodes"] as? [[String: Any]] else { return [] }
```

`JSONSerialization` turns GraphQL `null` nodes into `NSNull`. Casting the array to `[[String: Any]]` then **fails for the whole array**, so the function returns `[]` even though most nodes are valid dictionaries.

Confirmed in a Swift one-liner: `[{"number":885}, null, {"number":837}]` → cast to `[[String: Any]]` fails; `compactMap { $0 as? [String: Any] }` keeps both real PRs.

Downstream:

- `IssueTracker.applyPRStatuses` early-returns on empty `viewerPRs`
- session `prStatus.checks` stays `unknown`
- web UI only shows Fix Checks when `pr.checks === 'failing'` (`Packages/CrowDaemon/.../web/app.js`)
- auto-merge / auto-rebase also see an empty viewer list (`not-in-viewer-prs` / no status)

SAML recovery was added and unit-tested (`testListMonitoredPRsRecoversAccessiblePRsOnSAML`), but the fixture uses a clean `nodes` array with **no nulls**, so this cast failure never showed up.

Same `as? [[String: Any]]` pattern appears on other GraphQL `nodes` walks in this backend (labels, review requests, search PRs, etc.) and will fail the same way if any element is null.

## Secondary gap

`prStates(refs:)` (stale/session-linked PR backfill) does **not** request `statusCheckRollup` at all — only merge/review/label fields. Even when that path runs, it cannot restore CI state after the primary monitored-PRs parse drops everything.

## Expected

1. Parse GraphQL `nodes` as `[Any]` and `compactMap` to `[String: Any]`, skipping `null`.
2. Keep returning accessible viewer PRs under SAML (existing recovery path), including `statusCheckRollup` / failed check names.
3. Extend the SAML monitored-PRs fixture to include interleaved `null` nodes and assert the non-null PRs (and their check state) survive.
4. Optionally add `statusCheckRollup` to `prStates` so session-linked PRs still get CI status if the bulk viewer query is incomplete.

## Impact

Any account with at least one open PR in a SAML-restricted org loses CI visibility (and Fix Checks / auto-respond-to-failed-checks) for **all** open PRs — not just the restricted ones. That matches “Crow is not finding the CI failures / no Fix Checks button.”

dhilgaertner:
Closes #894

## The bug

GitHub's GraphQL connections declare *nullable elements*. When the OAuth token can't reach an org, GitHub nullifies **that node** and returns every sibling normally, reporting the hole in `errors` with a path like `["viewerPRs","pullRequests","nodes",3]`.

Swift's conditional array downcast is element-wise and all-or-nothing, so **one** `NSNull` makes `as? [[String: Any]]` return nil for the entire array:

```swift
// GitHubCodeBackend.parseViewerPRs — before
guard let pullRequests = viewerObj?["pullRequests"] as? [String: Any],
      let nodes = pullRequests["nodes"] as? [[String: Any]] else { return [] }
```

The ticket's live repro: 15 nodes, 7 null, 8 real — including PR #885 with `statusCheckRollup.state: FAILURE`. All 8 dropped. Downstream, `applyPRStatuses` early-returns on the empty list, `prStatus.checks` stays `unknown`, and `app.js` only renders Fix Checks when `pr.checks === 'failing'`. **One PR in one restricted org broke CI visibility for the entire account.**

## What changed

All 16 GraphQL connection walks across both GitHub backends now go through a `LenientJSON` helper that filters unresolvable elements instead of failing the array. That includes `parseIssueNodes` — **the assigned-issues board had the identical bug** and emptied the same way.

Riding along, both touching the stale/session-linked PR path — the only route by which a PR past `viewer.pullRequests(first: 50)`, or one in a restricted org (a *permanent* hole in that connection, not a transient one), can ever surface CI state:

- **`prStates` now requests `statusCheckRollup`.** Parsing was already free — `parsePRNode` is shared — so only the query text needed it. Additive by construction: `mergePRRecords` backfills `checksState` only when the winner's is empty, so a fresh viewer record can't be shadowed by a stale one.
- **`prStates` recovers its accessible aliases under SAML** rather than letting one blocked repo throw away every other ref in the batch. Symmetric with what `listMonitoredPRs` already does.

## Reviewer notes

**One intentional behavior change.** `hasProjectItems` now sees through a nulled project item. Previously `[null, {…}]` failed the cast → "no board" → `setTaskStatus` quietly wrote a `crow:in-review` label onto an issue that *is* on a board. It now raises `unimplemented`, which is what that function's own doc comment already claimed it does. Both sides are pinned by tests.

**`prDataComplete` is safe.** Swallowing the `prStates` throw means the flag stays `true`. This can't cause a wrong auto-completion: `decideSessionCompletions` and `decideReviewCompletions` both require positive evidence (`guard let pr = prsByURL[prLink.url] else { continue }`), so a restricted PR — absent from the partial map — is skipped. I deliberately did *not* add a coverage check (`states.count < refs.count → incomplete`): refs go permanently unresolved for benign reasons (deleted PR, transferred repo), which would latch the flag off forever and block auto-completion for every PR-linked session.

**`reportSAMLWarning()` still fires.** `prStates` no longer reaches `handleGitHubBackendError`, but `listAssigned`/`listMonitoredPRs` run every cycle and both set `samlRestricted`. A SAML org that breaks `prStates` necessarily breaks those.

**`applyPRStatuses`'s `guard !viewerPRs.isEmpty` is unchanged** — it's why the bug was *silent*, but the guard itself is correct (`applyAutoRebase`'s per-head bookkeeping relies on an empty poll not wiping its maps). Post-fix the list isn't empty in the reported scenario. Flagging so nobody re-investigates.

## Testing

`swift test --package-path Packages/CrowProvider` — 154 XCTest + 52 swift-testing, green.
`swift test --package-path Packages/CrowEngine` — 476 tests, green.

The existing SAML tests missed this bug because their fixtures used clean `nodes` arrays. The issue-side one even carried `"path":["openIssues","nodes",3]` in its `errors` while its array had no null — it *described* the bug without reproducing it. Both fixtures now interleave nulls at head/middle/tail.

Every new test was verified to fail before the fix, not just pass after it. Reverting `LenientJSON.objects` to the raw cast turns 9 of them red; separately disabling the `prStates` recovery block turns `testPRStatesRecoversAccessibleRefsOnSAML` red. A test that passes both ways is exactly the failure mode that let this ship.

## Known unrelated failure

`make parity` fails on `check-cli-parity.sh` with `automation-get` / `automation-set` registered but not in `ParityLedger.rpcMethods`. **This is pre-existing** — it reproduces identically on a clean tree at this branch point (from #884), and nothing here touches RPC registration.

## Follow-up

36 instances of the same `as? [[String: Any]]` pattern remain in `JiraTransitionClient`, `JiraSearchClient`, `JiraStatusFetcher`, `GitLab*Backend`, `JiraTaskBackend`, `CorveilTaskBackend`, and `CursorHookConfigWriter`. Same latent trap, different providers — worth a sweep ticket, plus a `check-no-nslog.sh`-style grep gate once they're all converted (a gate covering 2 of 10 affected files would pass while the trap survives everywhere else, so it's not worth adding yet).


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- GraphQL owner/repo/number/branch values stay bound via `gh -F` variables — no string interpolation into the query body (`GitHubCodeBackend.swift` ~103–128, ~224–235).
- `LenientJSON` is pure filtering of already-decoded JSON; it does not broaden trust boundaries or execute untrusted content.
- Intentional `hasProjectItems` change fixes a real integrity bug: a SAML-nulled project item no longer looks like “no board,” which previously caused a quiet `crow:` label write onto an issue that *is* on a board (`GitHubTaskBackend.swift` ~332–348).

**Concerns:**
- None in the GitHub GraphQL paths touched here. Remaining `as? [[String: Any]]` casts in Jira/GitLab/Corveil backends (called out in the PR follow-up) are the same latent trap, but out of scope for this fix.

### Code Quality
- `LenientJSON` is the right abstraction: clear “do not simplify back to the cast” docs, three small accessors matching real GraphQL shapes, and a direct unit test that pins the Swift cast failure mode (`LenientJSON.swift`, `testLenientJSONDropsNullsInsteadOfFailingTheArray`).
- Coverage is excellent: nested nulls, wholly-null connections, issue-board recovery, `prStates` rollup + SAML partial recovery, non-SAML still throws, and the label-fallback companion for all-null project items. All 82 `BackendsTests` passed locally (`arch -arm64 swift test --package-path Packages/CrowProvider --filter BackendsTests`).
- `prStates` SAML recovery correctly reuses `decodeGraphQLData` (needed for the trailing `gh:` line) and keeps auto-completion safe: restricted refs are absent from the map, and completion still requires positive evidence.
- Green: `findRecentPRsForBranches` still throws away the whole batch on SAML (`GitHubCodeBackend.swift` ~250–251) — same class as the `prStates` fix, lower severity (branch reconcile). Worth a follow-up for symmetry.
- Green: `testPRStatesStillThrowsOnNonSAMLFailure` asserts “not `.samlRestricted`” but not the concrete expected error; tightening that would make a silent wrong-path harder.

### Summary Table

---

PR #895, opened 2026-07-28
ADR 0016: complete the parity contract (CROW-806)

dhilgaertner:
Closes #806

## Why this is an amendment, not a new ADR

**The ADR #806 asks for already exists.** PR #883 — which closed the *dependent* ticket #807 (the anti-drift gate) — wrote `docs/adr/0016-cli-control-plane-parity.md` as part of landing `ParityLedger`, and merged it to `main` before this foundation ticket was picked up. A reviewer expecting a brand-new `0017-*.md` should read this diff as completing 0016 to its intended scope. `docs/adr/README.md` says *"Each file is one decision"*, so a second ADR on the same decision would be wrong.

Checking #806's five required statements against what shipped:

| #806 asks the ADR to state | Before | Now |
|---|---|---|
| Parity contract + rationale (headless/scriptable, agent-drivability, reproducibility) | partial — *reproducibility* unnamed | ✅ |
| Deviations only via a justified allowlist (`KNOWN_UI_ONLY`) | ✅ `Coverage.noCLI(reason:)` | ✅ |
| Enumerable surfaces (RPC / CLI verbs / `AppConfig`) | ✅ | ✅ |
| **Per-feature file-touch checklist (CROW-604 precedent)** | ❌ absent | ✅ |
| **App-down fallback duality as a divergence risk (ADR 0009)** | ❌ absent | ✅ |

## What's in the ADR now

**Definition of done for a CLI-covered feature.** A twelve-row file-touch table from the `crow job` precedent (CROW-604, #624) and every parity PR since. The load-bearing column is *which rows a machine already checks* — six of twelve. The other six are convention, and convention is what the jobs CLI proved fragile: `crow job` shipped and then went missing from the reference, along with `set-locked`, `recreate-terminal`, `resync-jira` and `codex-notify`, until CROW-808 built the doc gate.

**A covered row does not mean one implementation.** ADR 0009 retired `forwardToApp`, but the shape survives — the daemon's `makeCommandRouter` and the `makeEngineRouter` it falls back to both register **29 of the same method names**. The daemon's copy always answers; the engine's is shadowed. `CommandRouter.methodNames` *unions* them, so the ledger sees one row where two bodies of code exist.

They have already drifted, verified on this tree:

- `get-config` returns `app_running: false` from `RPCHandlers.swift:1084` and `true` from `EngineRouter.swift:91`, and the daemon copy returns three keys the engine omits (`configured`, `default_dev_root`, `vs_code_available`) — one method name, two response contracts.
[...]

PR #898, opened 2026-07-28
Resolve the workspace AI gateway for review sessions

### LINKED ISSUE #891: Review sessions never resolve the workspace AI gateway (clone lives outside the workspace folder)
## Summary

Review sessions can never pick up a workspace's AI gateway (CROW-402). The gateway resolves to `nil` for every review, so the launch line actively **unsets** `ANTHROPIC_BASE_URL` / `ANTHROPIC_CUSTOM_HEADERS` and the review's traffic bypasses the gateway entirely.

## Root cause

`SessionService.workspaceGatewayResolved(for:)` derives the workspace name purely from the worktree path — the first path component under devRoot:

```swift
// Packages/CrowEngine/Sources/CrowEngine/SessionService.swift:1592-1601
guard let devRoot = ConfigStore.loadDevRoot(),
      let config = ConfigStore.loadConfig(devRoot: devRoot),
      let worktree = appState.primaryWorktree(for: sessionID),
      let wsName = Self.workspaceName(forWorktreePath: worktree.worktreePath, devRoot: devRoot),
      let workspace = config.workspaces.first(where: { $0.name == wsName }),
      let gateway = workspace.gateway, !gateway.isEmpty
else { return nil }
```

But review clones don't live under a workspace folder. `prepareReviewClone` puts them in a flat sibling directory:

```swift
// Packages/CrowEngine/Sources/CrowEngine/SessionService.swift:2674-2676
let reviewsDir = (devRoot as NSString).appendingPathComponent("crow-reviews")
let cloneDirName = "\(repoName)-pr-\(prNumber)"
let clonePath = (reviewsDir as NSString).appendingPathComponent(cloneDirName)
```

So `workspaceName(forWorktreePath:devRoot:)` returns `"crow-reviews"`, which matches no entry in `config.workspaces` → `nil`.

Worker/job sessions are unaffected — their worktrees are at `{devRoot}/{workspace}/{repo}-{n}-{slug}`, so the first component *is* the workspace name.

## Why it's worse than "not applied"

A `nil` resolution isn't a no-op. `ClaudeLaunchArgs.gatewayEnvPrefix(nil)` emits an explicit unset:

```swift
// Packages/CrowClaude/Sources/CrowClaude/ClaudeLaunchArgs.swift:66-69
guard let resolved else {
    return "unset ANTHROPIC_BASE_URL ANTHROPIC_CUSTOM_HEADERS && "
}
```

and `ClaudeHookConfigWriter.writeGatewayEnv(dirPath:resolved: nil)` strips the env keys from the clone's `settings.local.json`. So a review session strips the gateway even if the user's shell had it exported — it can't be worked around from `~/.zshrc`.

## Reproduction

1. Set an AI gateway on a workspace (Settings → workspace → AI gateway).
2. Start a review on a PR in that workspace's repo (`crow start-review --url …`).
3. Inspect `{devRoot}/crow-reviews/{repo}-pr-{N}/.claude/settings.local.json`.

**Expected:** an `env` block with `ANTHROPIC_BASE_URL` and `ANTHROPIC_CUSTOM_HEADERS`.
**Actual:** only `hooks` — no `env` block. No traffic reaches the gateway.

Observed on three reviews started after the gateway was configured:

| clone | settings.local.json written | env block |
|---|---|---|
| `crow-pr-861` | 19:07:54 | none |
| `corveil-pr-1865` | 19:12:12 | none |
| `corveil-pr-1868` | 19:15:30 | none |

Not a stale-binary issue — the running `crowd` contains the CROW-402 gateway code (`strings` shows `ANTHROPIC_CUSTOM_HEADERS` and the `unset …` prefix).

## Suggested fix

The data needed to map a review clone back to its workspace is already on the session. For `review-corveil-1868`:

- primary worktree: `{devRoot}/crow-reviews/corveil-pr-1868`, `repo: "corveil"`
- session link: `{ type: "pr", url: "https://github.com/corveil/corveil/pull/1868" }`

and the workspace config carries repo globs:

```json
{ "name": "Corveil", "alwaysInclude": ["corveil/*"], "autoReviewRepos": ["corveil/*"], "gateway": { … } }
```

So `workspaceGatewayResolved(for:)` could fall back, when path-based lookup fails, to matching the session's PR `owner/repo` against each workspace's repo globs (`alwaysInclude` / `autoReviewRepos`, minus `excludeReviewRepos`) and use the first match's gateway. Path-based resolution stays the fast path for worker/job sessions.

Worth deciding explicitly: what should happen when a review's repo matches **no** workspace (e.g. a one-off PR outside any configured org glob)? Current behavior — unset — is a defensible default, but it should be a deliberate choice rather than a fall-through.

## Notes

- Applies to the Claude Code agent path specifically; the gateway env is gated to `agent.kind == .claudeCode` (`SessionService.swift:955`).
- Gateway env is only applied at agent **launch**, so sessions already running when a gateway is saved keep their old environment until relaunched. That's expected, but worth a line in `docs/configuration.md`.
- Separately (not this bug): the settings UI stores the header value verbatim after the first `:` (`Resources/web/settings.js:398-407`), so a value pasted with surrounding shell quotes is stored *with* the quotes and sent as `x-citadel-api-key: "Bearer sk-…"`. Corveil trims whitespace and a `bearer ` prefix but not quotes, so the key silently fails to match. A validation warning on save (or stripping matched surrounding quotes) would turn a silent 401 into an obvious one.

dhilgaertner:
Closes #891

## The bug

A workspace can route its Claude Code sessions through an internal LLM gateway (CROW-402). Review sessions never got one.

`SessionService.workspaceGatewayResolved(for:)` derived the workspace purely from path math — the first component under the dev root. That works for work and job sessions, whose worktrees live at `{devRoot}/{workspace}/{repo}-{n}-{slug}`. Review clones don't: `prepareReviewClone` puts them in a flat sibling directory, `{devRoot}/crow-reviews/{repo}-pr-{N}`. So the derived name was the literal string `"crow-reviews"`, which matches no workspace, and resolution returned `nil` for **every review, always**.

`nil` is not a no-op — it actively strips the gateway:

- `ClaudeLaunchArgs.gatewayEnvPrefix(nil)` prefixes the launch with `unset ANTHROPIC_BASE_URL ANTHROPIC_CUSTOM_HEADERS &&`
- `ClaudeHookConfigWriter.writeGatewayEnv(resolved: nil)` removes the `env` block from the clone's `settings.local.json`

So a review didn't merely miss the gateway; it defeated the workaround too. Exporting the vars in `~/.zshrc` couldn't help, because the launch line unsets them after the shell has run.

## The fix

Keep path lookup as the fast path, add a repo-slug fallback.

**`AppConfig.workspace(forRepoSlug:)`** (new, CrowCore) answers "which workspace owns `owner/repo`?" by matching the slug against `alwaysInclude` ∪ `autoReviewRepos` using the existing `repoMatchesPatterns` glob semantics.

**`workspaceGatewayResolved`** now tries the path, then — if nothing matched — the PR link's `owner/repo` via a new pure `repoSlug(fromPRLinks:)`. That helper reuses `Session.parseReviewPR`, the same parser `createReviewSession` used to name the clone, so resolution can't disagree with creation about what the repo is.

### Two decisions worth flagging

**`excludeReviewRepos` is deliberately not consulted.** It's a review-board *visibility* filter, not a membership statement. Subtracting it would mean a repo you'd hidden from the board silently lost its gateway when you reviewed it by hand — and because `effectiveExcludeReviewRepos` is a global union, one workspace's exclusion would suppress every workspace's match. Pinned by `workspaceForRepoSlugIgnoresExcludeReviewRepos`.

**No match still means unset, now deliberately and logged.** A PR in a repo no workspace claims runs vanilla and logs `No workspace claims repo …`. Falling back to `managerGateway` was considered and rejected: it would route a one-off PR from an unrelated org through your corporate gateway without asking. The Manager's gateway stays the Manager's alone.

Ambiguity is resolved deterministically so the answer can't flip when workspaces are reordered: an exact slug beats a glob, then config order. `workspaceForRepoSlugPrefersExactOverGlob` puts the glob *first* in config order and asserts the exact match still wins.

## Things a reviewer should look at

**The fallback is not gated on `session.kind == .review`.** It's gated on the precondition — path lookup empty *and* a PR link parses. The path fast path strictly dominates for any session under a workspace folder, so a kind gate would be dead weight there. And `.pr` links are not review-only: `IssueTracker.swift:1214`, `:1654` and `SessionService.swift:2171` attach them to **work** sessions once their PR opens. What a kind gate would suppress is therefore real and desirable — a work session whose worktree sits outside the dev root currently gets its gateway stripped, and now inherits its repo's workspace gateway instead. That is a behavior change beyond reviews; the log line makes it observable.

**The path fast path became case-insensitive.** It was `$0.name == wsName`, exact. APFS is case-preserving but case-*insensitive*, so a workspace named `Spotlight` with an on-disk folder `spotlight` is the same directory yet failed to match — silently stripping the gateway for every session in it. This widens matching only; it cannot turn a current match into a non-match. Note `resolvedCodeProvider` (`SessionService.swift:2180`) still does the case-sensitive compare against the same field; fixing that belongs to whatever ticket owns code-provider resolution, not this one.

**GitLab MR URLs resolve to a garbage slug**, and therefore to "unset" — exactly today's behavior, no regression. The mis-parse is upstream: `createReviewSession` uses the same `parseReviewPR`, so a GitLab review clone is *already* named `--pr-12`. Deliberately not fixed here, but pinned by `repoSlugForGitLabMRIsMalformedButStable` so a future GitLab fix updates it consciously rather than drifting.

**A resolved secret now lands in review clones**, which contain the PR author's code. The file is still `0600` and Crow never executes the clone, but the header is now readable by an agent working in a directory of untrusted code — a hostile PR could try to talk it into `cat`-ing `settings.local.json`. The risk is inherent to the mechanism and already applies to work and job sessions; extending it to reviews should be a conscious choice, so it's documented in `configuration.md` alongside the opt-out (drop the repo from the workspace's `alwaysInclude`/`autoReviewRepos`).

## Testing

23 new tests, all passing.

**CrowCore (13)** — `AppConfigWorkspaceForRepoSlugTests`. Runs in the PR Linux lane, which is why the lookup lives in CrowCore rather than CrowEngine (ADR 0007 excludes CrowEngine from `ci.yml`). Covers the `alwaysInclude` ∪ `autoReviewRepos` union, case-insensitivity both directions, globs, the `excludeReviewRepos` guard, both ambiguity rules, and membership-without-a-gateway.

**CrowEngine (10)** — `SessionServiceGatewayResolutionTests`, local `make test` only. Covers `repoSlug(fromPRLinks:)` (link-type selection, first-*parseable* rather than first, the GitLab shape) and `workspaceName(forWorktreePath:devRoot:)`, which had no tests at all despite being the function the bug lived in. `workspaceNameOfReviewCloneIsCrowReviews` pins the root cause as an executable assertion.

`workspaceGatewayResolved` itself is not directly tested: it reads `ConfigStore.loadDevRoot()`, which is all-static with no injection seam, and `SessionService.init` takes no devRoot or config — exercising it would mean touching the live devroot pointer, which ADR 0012 forbids. Every *decision* it makes now lives in the three pure helpers above, which is what made this testable at all. `workspaceName` and `repoSlug` are `nonisolated` for the same reason the file's existing `shouldStripCursorReviewCloneOnHandoff` is.

**Parity gate untouched, as intended.** The new member is a *method* on `AppConfig`, and `ParityGateTests` walks stored properties via `Mirror` — confirmed by the existing computed `effectiveExcludeReviewRepos`, which likewise has no ledger row. `make parity` passes with no `ParityLedger` edit.

**Parity gate: two rows added, eleven exemptions retired.** `make parity` was already red on `main` — `automation-get`/`automation-set` shipped with #884 but never got `ParityLedger` rows, the gap `b21ab1a` documented and left to CROW-807. Since the gate blocks this PR, it's fixed here rather than worked around:

- Added `.read("automation-get", cli: "automation get")` and `.write("automation-set", cli: "automation set")` — `check-cli-parity.sh` now reports 89 methods agreeing, up from a 87-vs-89 mismatch.
- Retired the eleven `automation.*` **config-field** exemptions. Each read "Web Settings toggle; no verb reads or writes it", which stopped being true when the automation verbs shipped; they're now covered rows pointing at `automation get`/`automation set`. This is the same cleanup #883's own history describes for the `agents-*` and `defaults.*` rows — closing a gap by deleting exemptions is what the ledger is for.

This is unrelated to the gateway bug and reviewable independently — it's the first two commits' worth of change in one file, and reverting it leaves the CROW-891 fix intact.

## Docs

`configuration.md` gains a **Which gateway applies** section — the two-step resolution, the `excludeReviewRepos` carve-out, the ambiguity rule, and the explicit "no match means unset, and the Manager's gateway is never inherited". The section previously said "per-workspace" throughout and never mentioned reviews at all.

Also adds the note the ticket asked for: gateway env is applied at agent **launch**, so a session already running when you edit a gateway keeps its old environment until it's relaunched or handed off.

## Not addressed

The ticket's third note — that the settings UI stores a header value verbatim after the first `:`, so a value pasted with surrounding shell quotes is sent as `x-citadel-api-key: "Bearer sk-…"` and silently 401s. That's a `Resources/web/settings.js` validation change with no overlap with this code path; it deserves its own ticket rather than a drive-by here.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/DevRootLayout.swift
+++ b/Packages/CrowCore/Sources/CrowCore/DevRootLayout.swift
@@ -1,0 +1,36 @@
+import Foundation
+
+/// The one place the dev root's own reserved directory names live, so the code
+/// that *creates* review clones, the repo scan that *skips* them, workspace-name
+/// validation, and gateway resolution can never drift (CROW-891).
+///
+/// The dev root holds one directory per workspace plus a small number of
+/// Crow-owned directories that are *not* workspaces. Those are indistinguishable
+/// from a workspace folder by path shape alone, so anything deriving a workspace
+/// from a path has to know which names are reserved.
+public enum DevRootLayout {
+    /// Holds transient PR-review clones at `{devRoot}/crow-reviews/{repo}-pr-{N}`.
+    ///
+    /// Deliberately *not* a workspace: review clones sit here rather than under
+    /// the owning workspace's folder, which is why a review session resolves its
+    /// workspace by repo slug instead of by path (CROW-891).
+    public static let reviewsDirName = "crow-reviews"
+
+    /// Directory names a workspace may not take, because a workspace folder of
+    /// that name would collide with Crow's own dev-root layout.
+    public static let reservedWorkspaceNames: Set<String> = [reviewsDirName]
+
+    /// Whether `name` collides with a reserved dev-root directory.
+    ///
+    /// Case-insensitive: APFS is case-preserving but case-insensitive, so a
+    /// workspace named `Crow-Reviews` would be the very same directory as the
+    /// review-clone root.
+    public static func isReservedWorkspaceName(_ name: String) -> Bool {
+        reservedWorkspaceNames.contains(name.lowercased())
+    }
+
+    /// The review-clone root under a dev root.
+    public static func reviewsDir(devRoot: String) -> String {
+        (devRoot as NSString).appendingPathComponent(reviewsDirName)
+    }
+}
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -636,6 +636,13 @@ public struct WorkspaceInfo: Identifiable, Codable, Sendable, Equatable {
         if name == "." || name == ".." {
             return "Name cannot be “.” or “..”"
         }
+        // Crow owns some dev-root directories that aren't workspaces. A workspace
+        // folder of that name would collide with them on disk, and anything
+        // deriving a workspace from a path would bind those sessions to it —
+        // review sessions in particular (CROW-891).
+        if DevRootLayout.isReservedWorkspaceName(name) {
+            return "“\(name)” is reserved by Crow and cannot be a workspace name"
+        }
         let lowercased = name.lowercased()
         if existingNames.contains(where: { $0.lowercased() == lowercased }) {
             return "A workspace with this name already exists"
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -97,6 +97,36 @@ public struct AppConfig: Codable, Sendable, Equatable {
         defaults.excludeReviewRepos + workspaces.flatMap(\.excludeReviewRepos)
     }
 
+    /// The workspace that owns a repo, addressed by its `owner/repo` slug.
+    ///
+    /// Membership is `alwaysInclude` ∪ `autoReviewRepos` — the two lists that name
+    /// repos a workspace *works on*. `excludeReviewRepos` is deliberately **not**
+    /// subtracted: it's a review-board *visibility* filter, not a membership
+    /// statement, so a repo hidden from the board still belongs to the workspace
+    /// and still gets its gateway (CROW-891).
+    ///
+    /// Patterns use ``repoMatchesPatterns`` glob semantics — case-insensitive, one
+    /// `*`. Ambiguity resolves deterministically so two workspaces claiming the
+    /// same repo can't flip the answer between launches: a workspace naming the
+    /// slug **exactly** beats one matching only through a glob, and among equals
+    /// the earlier entry in `workspaces` (config file order) wins.
+    ///
+    /// Returns nil when no workspace claims the slug. Callers must treat that as
+    /// "unset", not "inherit" — see `SessionService.workspaceGatewayResolved`.
+    public func workspace(forRepoSlug slug: String) -> WorkspaceInfo? {
+        let lowerSlug = slug.lowercased()
+        var globMatch: WorkspaceInfo?
+        for workspace in workspaces {
+            let patterns = workspace.alwaysInclude + workspace.autoReviewRepos
+            guard repoMatchesPatterns(slug, patterns: patterns) else { continue }
+            if patterns.contains(where: { $0.lowercased() == lowerSlug }) {
+                return workspace  // exact beats glob; first exact in config order wins
+            }
+            if globMatch == nil { globMatch = workspace }
+        }
+        return globMatch
+    }
+
     public init(
         workspaces: [WorkspaceInfo] = [],
         defaults: ConfigDefaults = ConfigDefaults(),
--- a/Packages/CrowCore/Tests/CrowCoreTests/AppConfigWorkspaceForRepoSlugTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AppConfigWorkspaceForRepoSlugTests.swift
@@ -1,0 +1,149 @@
+import Foundation
+import Testing
+
+@testable import CrowCore
+
+/// Coverage for `AppConfig.workspace(forRepoSlug:)` — the repo→workspace lookup
+/// that lets a review clone under `{devRoot}/crow-reviews/` find its workspace's
+/// AI gateway, which path-based resolution alone can never do (CROW-891).
+///
+/// Pure function over in-memory config, so no temp dirs and no `ConfigStore`.
+
+// MARK: - Fixtures
+
+private func ws(
+    _ name: String,
+    alwaysInclude: [String] = [],
+    autoReviewRepos: [String] = [],
+    excludeReviewRepos: [String] = [],
+    gateway: WorkspaceGateway? = nil
+) -> WorkspaceInfo {
+    WorkspaceInfo(
+        name: name,
+        alwaysInclude: alwaysInclude,
+        autoReviewRepos: autoReviewRepos,
+        excludeReviewRepos: excludeReviewRepos,
+        gateway: gateway
+    )
+}
+
+private let probeGateway = WorkspaceGateway(
+    baseURL: "https://gateway.invalid",
+    customHeaders: ["x-probe": "value"]
+)
+
+// MARK: - Membership
+
+@Test func workspaceForRepoSlugMatchesAlwaysInclude() {
+    let config = AppConfig(workspaces: [ws("Acme", alwaysInclude: ["acme/widget"])])
+    #expect(config.workspace(forRepoSlug: "acme/widget")?.name == "Acme")
+}
+
+@Test func workspaceForRepoSlugMatchesAutoReviewRepos() {
+    // Membership is the *union* — a repo named only in autoReviewRepos still
+    // belongs to the workspace.
+    let config = AppConfig(workspaces: [ws("Acme", autoReviewRepos: ["acme/widget"])])
+    #expect(config.workspace(forRepoSlug: "acme/widget")?.name == "Acme")
+}
+
+@Test func workspaceForRepoSlugIsCaseInsensitive() {
+    let config = AppConfig(workspaces: [ws("Acme", alwaysInclude: ["Acme/Widget"])])
+    #expect(config.workspace(forRepoSlug: "acme/widget")?.name == "Acme")
+
+    let lowerConfig = AppConfig(workspaces: [ws("Acme", alwaysInclude: ["acme/widget"])])
+    #expect(lowerConfig.workspace(forRepoSlug: "ACME/WIDGET")?.name == "Acme")
+}
+
+@Test func workspaceForRepoSlugHonorsGlobs() {
+    let config = AppConfig(workspaces: [
+        ws("Acme", alwaysInclude: ["acme/*"]),
+        ws("Widgets", alwaysInclude: ["*/gadget"]),
+    ])
+    #expect(config.workspace(forRepoSlug: "acme/anything")?.name == "Acme")
+    #expect(config.workspace(forRepoSlug: "other/gadget")?.name == "Widgets")
+    #expect(config.workspace(forRepoSlug: "other/unclaimed") == nil)
+}
+
+@Test func workspaceForRepoSlugIgnoresExcludeReviewRepos() {
+    // excludeReviewRepos hides a repo from the review *board*; it is not a
+    // membership statement. A manually started review on an excluded repo still
+    // gets the workspace's gateway (CROW-891 decision).
+    let config = AppConfig(workspaces: [
+        ws(
+            "Acme",
+            alwaysInclude: ["acme/widget"],
+            excludeReviewRepos: ["acme/widget"],
+            gateway: probeGateway
+        )
+    ])
+    #expect(config.workspace(forRepoSlug: "acme/widget")?.gateway == probeGateway)
+}
+
+// MARK: - No match
+
+@Test func workspaceForRepoSlugReturnsNilWhenUnclaimed() {
+    let config = AppConfig(workspaces: [ws("Acme", alwaysInclude: ["acme/*"])])
+    #expect(config.workspace(forRepoSlug: "stranger/repo") == nil)
+}
+
+@Test func workspaceForRepoSlugReturnsNilForEmptyWorkspaces() {
+    #expect(AppConfig().workspace(forRepoSlug: "acme/widget") == nil)
+}
+
+@Test func workspaceForRepoSlugReturnsNilForWorkspaceWithNoPatterns() {
+    // Both lists empty → claims nothing. Such a workspace is still reachable via
+    // the worktree-path fast path.
+    let config = AppConfig(workspaces: [ws("Acme", gateway: probeGateway)])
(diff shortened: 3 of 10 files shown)


dgershman (changes requested):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- No Manager-gateway inheritance for unmatched repos — correct; avoids routing unrelated orgs through a corporate proxy without consent (`SessionService.swift:1592-1596`).
- `excludeReviewRepos` correctly left out of membership so a visibility filter cannot strip credentials / suppress resolution across workspaces.
- Secrets still land only in `0600` `settings.local.json`; resolved values are not logged; opt-out for sensitive repos is documented in `configuration.md`.
- Parity ledger retirement of the old `remoteControlEnabled` “no verb” exemptions matches the intentional remote `automation-*` surface already defended in `RPCWebSocketHandler` / `DaemonSecurityTests` (#884) — not a new privilege expansion.

**Concerns:**
- **Yellow** — Path fast path can short-circuit the slug fallback if a workspace is named `crow-reviews`. `workspaceName` returns that literal for every review clone (`SessionService.swift:1607-1610`, `:1658-1660`), then `config.workspaces.first { $0.name.lowercased() == wsName.lowercased() }` will bind all reviews to that one workspace and never consult `workspace(forRepoSlug:)`. Comments claim path math “matches nothing,” but `WorkspaceInfo.validateName` does not reserve `crow-reviews`, and `GitManager` already special-cases the directory. Treat `wsName.lowercased() == "crow-reviews"` as a non-match (and ideally reject the name in `validateName`) so the documented invariant holds.

### Code Quality
- **Green** — Writing resolved gateway headers into review clones expands the existing work/job secret-on-disk surface into untrusted PR trees. Documented with an opt-out; consider a follow-up that uses launch-line-only (no `settings.local.json` env) for `.review` sessions.
- **Green** — GitLab MR slug mis-parse (`proj/-`) is pinned and non-regressive; worth a dedicated `parseReviewPR` fix later.
- **Green** — Case-insensitive path match is a good APFS fix; `resolvedCodeProvider` still case-sensitive as noted — fine to leave out of scope.
- Fix shape is clean: pure helpers + solid CrowCore/CrowEngine coverage (13 + 10 tests passed locally). Bundled parity cleanup is justified and reviewable independently.

### Static Analysis
- CrowCore `AppConfigWorkspaceForRepoSlug*` — 13/13 passed.
- CrowEngine `SessionServiceGatewayResolution` — 10/10 passed.
- (Local `swift test` also printed an x86_64 vs arm64 XCTest bundle warning; Swift Testing still ran and passed.)

### Summary Table

---

dhilgaertner:
Thanks — the Yellow is a real bug and it's fixed in the follow-up commit.

**Yellow: `crow-reviews` as a workspace name.** You're right, and my case-insensitive path match slightly widened it: `Crow-Reviews` would have collided too. Worse, the failure is silent and total — one such workspace captures *every* review session and the slug fallback never runs, which is precisely the invariant the comments claim.

Fixed both halves you suggested:

- **Resolution treats it as a non-match** (`SessionService.swift`, path fast path) — guarded on `DevRootLayout.isReservedWorkspaceName(wsName)` rather than a bare string compare, so a config written before this still resolves correctly.
- **`validateName` rejects it** (`AppConfig.swift`), case-insensitively, so the collision can't be created going forward.

Rather than add a fifth hardcoded `"crow-reviews"`, I pulled the name into `DevRootLayout` in CrowCore — same single-source-of-truth rationale as the existing `ArtifactPaths` — and routed the four existing literals through it (`prepareReviewClone`, `Scaffolder`, `GitManager`'s discovery skip, plus the new checks). `GitManager`'s skip became case-insensitive as a side effect, which is the same APFS correctness point.

I deliberately did **not** change `workspaceName` itself to return nil for reserved names. It's `public` and CROW-809's `workspace-edit`/`workspace-remove` use it to count orphaned sessions; changing its contract from "derive the folder name" to "derive the workspace" would silently alter that feature's behavior. The guard lives at the consumer instead.

Regression tests: 7 in CrowCore (`DevRootLayoutTests` — reservation, case-insensitivity, near-misses like `crow-reviews-archive` staying usable, and both `validateName` directions) and 1 in CrowEngine pinning that the review clone's derived name *is* reserved. CrowCore 517, CrowEngine 523, CrowDaemon 301, CrowGit 21 — all green, `make parity` clean.

**On the three Greens** — agreed on all, and none changed here:

- **Launch-line-only for `.review`** is the right shape and I'd support it, but it's a real behavior change: `settings.local.json` is what makes the gateway survive a manual `claude` re-run inside the clone, so dropping it trades the secret-at-rest exposure for reviews silently losing the gateway on any re-run. Worth its own ticket with that trade-off stated, not a quiet change here.
- **`parseReviewPR` GitLab fix** — agreed. Note the mis-parse is upstream of resolution: `createReviewSession` names the clone with the same parser, so GitLab review clones are already `{repo}--pr-{N}`. Fixing it properly means touching clone naming, which is why it's pinned rather than patched.
- **`resolvedCodeProvider` case-sensitivity** — thanks, left out of scope as you suggested.

One correction to the PR description while I'm here: the bundled parity cleanup you reviewed is no longer in this PR. #895 landed the identical automation-field retirement on `main` first, so on rebase I took its version wholesale and my ledger diff against `main` is now empty. Only the gateway change remains.

dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- **The one new secret-exposure surface is disclosed and mitigated.** A resolved gateway header now lands in review clones (which hold untrusted PR-author code). The PR calls this out, keeps the file `0600`, never executes the clone, notes Crow already does this for work/job sessions, and documents the opt-out (drop the repo from `alwaysInclude`/`autoReviewRepos`) in `configuration.md`. Extending an existing mechanism with eyes open, not a new hole.
- **No secret leaks into logs.** The "no workspace claims repo" line logs only the `owner/repo` slug, never gateway values (`SessionService.swift:379`).
- **No cross-workspace bleed.** A slug only matches a workspace via that workspace's *own* `alwaysInclude`/`autoReviewRepos` patterns, so a PR from an unrelated org can't attract another workspace's gateway unless the user's own `*` catch-all claims it — and that catch-all behavior is documented. The deliberate "no `managerGateway` fallback" keeps the Manager's credential from bleeding into one-off reviews.

### Code Quality
- **Root cause correctly identified and fixed.** Path math yielded the literal `"crow-reviews"` for review clones (`workspaceName(forWorktreePath:)`), which matched no workspace → `nil` → gateway actively *unset*. The repo-slug fallback via `parseReviewPR` reuses the same parser `createReviewSession` used to name the clone, so creation and resolution can't disagree about the repo.
- **`DevRootLayout` centralizes the reserved name** across all four call sites that previously hard-coded `"crow-reviews"` (Scaffolder, GitManager, `validateName`, gateway resolution) — the drift this prevents is real.
- **Ambiguity resolution is correct and order-independent.** Verified the exact-beats-glob short-circuit in `AppConfig.workspace(forRepoSlug:)`: an exact `patterns.contains(==slug)` returns immediately while glob matches only fall through at loop end, so an exact match wins regardless of config position. `workspaceForRepoSlugPrefersExactOverGlob` pins exactly this by putting the glob first.
- **Behavior changes are widening-only and documented.** Case-insensitive path matching and the work-session slug fallback (for worktrees outside the dev root) can only turn a former non-match into a match, never the reverse; both are flagged in the description and the log line makes the latter observable.
- **Tests are thorough and honest.** 19 CrowCore + 11 CrowEngine pass locally; the untestable-by-injection reality of `workspaceGatewayResolved` is acknowledged and every *decision* it makes is pushed into pure, tested helpers. `workspaceNameOfReviewCloneIsCrowReviews` pins the root cause as an executable assertion.

**Verified locally:** `swift build` clean for CrowCore and CrowEngine; `AppConfigWorkspaceForRepoSlug`/`DevRootLayout` (19) and `Gateway workspace resolution` (11) all pass.

- **GitLab MR slug `proj/-`** could in theory match a hypothetical `proj/*` glob and wrongly inherit a gateway. In practice a GitLab workspace's globs key on the real owner (`group/*`), which `proj/-` won't match, so this stays at today's "unset" — no regression. Already pinned by `repoSlugForGitLabMRIsMalformedButStable`, and the PR consciously defers the upstream GitLab parse. Fine to leave.
- The parity-gate cleanup (two rows added, eleven `automation.*` exemptions retired) is unrelated to the gateway fix but is self-contained, reverts independently, and turns a `main`-red gate green. Reasonable to land here.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. Correct root-cause fix, deterministic and order-independent resolution, honest disclosure of the review-clone secret trade-off with a documented opt-out, and all 30 tests green against a clean build.

---

PR #899, opened 2026-07-28
crow:merge no longer lands in silence — surface the verdict, and direct-merge when GitHub's auto-merge is disabled

### LINKED ISSUE #888: crow:merge label applies but PRs never auto-merge — silent when repo disallows auto-merge / blocked on review
## Summary

Since the web UI parity work, applying the **`crow:merge`** label (via Crow or on GitHub) reliably puts the label on the PR, but the PR often **never merges**. From the user’s POV the automation looks broken: the label is there, nothing else happens.

Live investigation against a running Crow instance (`~/Library/Logs/crow/crowd-automation.log` + GitHub state) shows the watcher *is* armed and sometimes enables GitHub auto-merge — but several failure modes leave a labeled PR permanently unmerged with **no in-UI explanation**.

## What Crow actually does

`crow:merge` does **not** merge the PR itself. Per `docs/automation.md` / `IssueTracker.applyAutoMerge`:

1. Watcher must be on (`Settings → Automation → Enable crow:merge auto-merge…` / `autoMergeWatcherEnabled`).
2. On each poll, for a session-linked PR with the label (and not draft / conflicting / changes-requested), Crow runs `gh pr merge --auto --squash --delete-branch`.
3. Eligibility also requires a `Crow-Session: <uuid>` commit trailer matching a **currently known** Crow session.
4. Enablement is one-shot (`Session.autoMergeEnabledAt` + in-memory permanent skip markers). After a permanent failure Crow stops retrying for that PR for the process lifetime (CROW-621).

GitHub then holds the merge until checks/reviews allow it — Crow does not babysit completion.

### 1. Repo has “Allow auto-merge” disabled — label sticks, enable never sticks (smoking gun for Corveil)

`corveil/corveil` currently has:

```json
{ "allow_auto_merge": false, "allow_squash_merge": true }
```

Open PRs with `crow:merge` and **no** `autoMergeRequest`:

| PR | State | Notes |
|----|--------|--------|
| [corveil#1866](https://github.com/corveil/corveil/pull/1866) | CLEAN + APPROVED + MERGEABLE | Ideal merge candidate; auto-merge never enabled |
| [corveil#1868](https://github.com/corveil/corveil/pull/1868) | BLOCKED / CHANGES_REQUESTED | Label only |
| [corveil#1865](https://github.com/corveil/corveil/pull/1865) | BLOCKED / CHANGES_REQUESTED | Label only |

Automation log (every ~60s):

```text
auto-merge: … https://github.com/corveil/corveil/pull/1866:repo-disallows-auto-merge …
```

Crow tried once, got `Auto merge is not allowed for this repository`, then permanently skipped. The UI still shows the gold `crow:merge` tag as if the pipeline is armed.

`corveil/crow` has `allow_auto_merge: true`, which is why Crow-repo PRs sometimes show `autoMergeRequest` while Corveil-repo ones never do.

### 2. Auto-merge enables, but GitHub stays BLOCKED on review — looks “stuck”

On `corveil/crow`, Crow *did* enable auto-merge for several PRs (log: `auto-merge: ENABLED on …`), yet they sit `mergeStateStatus: BLOCKED` with `reviewDecision: CHANGES_REQUESTED` or `REVIEW_REQUIRED`, e.g.:

- [#881](https://github.com/corveil/crow/pull/881) — `autoMergeRequest` set at 00:26:56Z; still open, CHANGES_REQUESTED
- [#883](https://github.com/corveil/crow/pull/883) — enabled; REVIEW_REQUIRED / BLOCKED
- Similar pattern on #878, #861

Crow’s poll then reports `in-flight` / already-enabled and does nothing further. Correct for GitHub auto-merge semantics, but after web parity the product surface mostly advertises the **label**, not “enabled vs waiting vs permanently failed”.

### 3. Web UX gap after native→web

The session action is **“Add label crow:merge to PR”**. Success = label present. There is little/no web surfacing for:

- `repo-disallows-auto-merge` (permanent)
- `no-crow-session-trailer` / session already cleaned up
- `not-in-viewer-prs`
- `changes-requested` / `conflicting` / `BEHIND` (update-branch path)
- “auto-merge enabled; waiting on GitHub for reviews/checks”

Sidebar only shows a small `⛙` when `session.auto_merge` is true (`autoMergeEnabledAt != nil`). A labeled PR with a permanent skip looks almost identical to a healthy pending auto-merge.

## Expected

When I apply `crow:merge`, either:

1. Crow enables GitHub auto-merge and the PR merges when green/approved, **or**
2. Crow clearly tells me why it can’t (repo setting, review state, missing trailer, etc.), **and**
3. For repos that disallow GitHub auto-merge, Crow should either refuse the label with a clear error, fall back to a direct merge when the PR is fully eligible, or guide me to flip **Settings → General → Allow auto-merge** on the repo.

Quiet permanent skips after labeling are the broken experience.

## Proposed directions (non-prescriptive)

- **Detect repo `allow_auto_merge` early** (when adding the label or on first enable failure) and surface a session/PR banner + notification instead of only `crowd-automation.log`.
- **Web PR status vocabulary**: distinguish `crow:merge` (request) vs “auto-merge enabled” vs “blocked: reason” vs “permanently skipped: reason”.
- **Fallback** when the repo disallows native auto-merge: once the PR is OPEN, non-draft, non-conflicting, not CHANGES_REQUESTED, checks green (and Crow-authored), run a non-`--auto` squash merge — or refuse `add-merge-label` with an actionable error.
- **Re-evaluate after CHANGES_REQUESTED clears**: if GitHub cancelled/blocked the auto-merge request, don’t leave the session forever in a false “done” state without UI.

## Evidence pointers

- Watcher / skip reasons: `IssueTracker.applyAutoMerge`, `attemptEnableAutoMerge`, `autoMergeSkipReason`, `isPermanentAutoMergeFailure` (`Packages/CrowEngine/.../IssueTracker.swift`)
- Label action: web `add-merge-label` → daemon RPC
- Docs: `docs/automation.md` (#299), `docs/troubleshooting.md` (auto-merge table)
- Local log: `~/Library/Logs/crow/crowd-automation.log` (`auto-merge:` lines)

## Acceptance ideas

- [ ] Labeling a PR on a repo with `allow_auto_merge: false` produces a visible Crow error/banner (not only a log line) and does not look “armed”.
- [ ] Labeled + successfully enabled PRs show an explicit “auto-merge enabled / waiting on …” state in the web UI.
- [ ] A CLEAN+APPROVED Crow-authored PR on a disallowing repo either merges via fallback or is explicitly refused — it must not sit labeled forever with no signal.
- [ ] Automation log reasons remain, but the common cases above are answerable from the UI alone.

dhilgaertner:
Closes #888

Applying `crow:merge` reliably landed the label, but the PR often never merged and Crow said nothing. The label is a *request*; the watcher decides whether to act — and every skip reason went only to `~/Library/Logs/crow/crowd-automation.log`.

The ticket's smoking gun: `corveil/corveil` has `allow_auto_merge: false`. Crow called `gh pr merge --auto` once, matched the English phrase in the error, latched a permanent in-memory skip, and never retried or told anyone. corveil#1866 sat CLEAN + APPROVED + MERGEABLE, labeled, forever. In the sidebar a permanently-dead PR and a healthy pending one rendered identically.

## What changed

**Detect repo policy proactively.** `Repository.autoMergeAllowed` rides the `repository { … }` object both GraphQL queries already select, so it costs no extra request. `nil` means *unknown*, never *forbidden* — only an explicit `false` changes behaviour, because guessing wrong either strands a healthy PR or merges one Crow shouldn't touch. `isPermanentAutoMergeFailure` survives as a backstop for records where the field wasn't fetched (GitLab, partial SAML recovery, pre-#888 caches).

**Merge it ourselves when GitHub won't.** A repo that forbids the auto-merge queue now gets a direct squash merge (`gh pr merge --squash --delete-branch`, no `--auto`), behind a predicate strictly narrower than the auto-merge path — that path hands GitHub a queued request and lets *GitHub* enforce checks and reviews, while a direct merge has no such backstop:

```swift
guard pr.repoAutoMergeAllowed == false else { return false }   // never on unknown/true
guard shouldAttemptAutoMerge(pr: pr, session: session) else { return false }
guard pr.mergeStateStatus == "CLEAN" else { return false }     // GitHub's own green merge button
guard pr.mergeable == "MERGEABLE" else { return false }
guard pr.checksState == "SUCCESS" else { return false }
guard pr.reviewDecision == "APPROVED" else { return false }
```

Crow authorship is re-verified before the merge, same as the auto-merge path. A repo with no required reviewers reports `reviewDecision == ""` and is deliberately left alone — refusing to merge something no human approved is the right side to err on.

**Publish the verdict.** The twelve skip reasons (six enum cases plus six ad-hoc string literals scattered across `applyAutoMerge` / `attemptEnableAutoMerge`) are now one enum carrying the same grep-stable log strings **plus** a human sentence and a permanence flag. It lands in `AppState.autoMergeState` and ships on `list-sessions-live` as an additive `auto_merge_state` `{phase, reason, message, permanent}`.

| state | pill | meaning |
|---|---|---|
| `enabled` | ⛙ green | queued; waiting on GitHub |
| `merged` | ⛙ purple | Crow merged it directly |
| `stalled` | ⛙ orange | will retry |
| `blocked` | ⛙ red | permanent; needs a human |
| `off` | ⛙ grey | the watcher is switched off |

Reasons the pill **already** renders (conflicts, changes-requested, no label) are deliberately *not* republished — two surfaces disagreeing about one PR is what CROW-773 consolidated the vocabulary to prevent.

The bare untinted ⛙ in the row trail folds into the pill, so "Crow armed this" and "Crow gave up on this" stop rendering identically. A permanent block also fires a new **Auto-Merge Blocked** notification, once per (PR, reason) rather than once per 60s poll, and `crow add-merge-label` returns an additive `warning` alongside `ok: true` when the label can't lead to a merge.

## Worth a reviewer's eye

- **A disallowing repo whose PR isn't *yet* green is transient, not latched.** Labels are normally applied while CI is still running, so latching there would freeze the PR before it could ever qualify for the fallback — the exact case this change exists to fix. `IssueTrackerDirectMergeTests.aNotYetGreenPRInADisallowingRepoIsTransientNotPermanent` pins it.
- **The fallback merges real code, unconditionally** (no new setting — that was the explicit product call). The four gates above plus the `repoAutoMergeAllowed == false` precondition are the whole safety margin. If they prove too loose, the escape hatch is a follow-up `autoMergeDirectFallbackEnabled` toggle.
- `auto_merge_state` is a **sibling** of `pr`, not a member: `prStatusJSON` is shared byte-identically with `get-pr-status` (which answers "what does the forge say"), and a verdict must survive a session with no `PRStatus` at all — the `not-in-viewer-prs` case.
- `EngineRouter`'s `add-merge-label` structurally cannot carry the warning (`onAddMergeLabel` is `(UUID) -> Void`), and is unreachable in the daemon — `CommandRouter` only consults a fallback for *unregistered* methods. Commented rather than changed.

## Rebase notes (2026-07-28)

Rebased onto `main` after #896, #885 and #895 landed. Two conflicts, both resolved toward upstream:

- **`GitHubCodeBackend.parsePRNode`** — #896 replaced the manual GraphQL array parsing with `LenientJSON.nodes`. Took that, and kept the new `autoMergeAllowed` read on the manual optional cast that `repoName` beside it uses, since `repository` is a plain object rather than a connection (`LenientJSON` is for connections).
- **`ParityLedger.swift`** — this branch originally carried two fixes there; both were solved upstream in the meantime. #885 ledgered the `automation-*` rows (with better history than my note: they were missed because #884 and the gate itself, #883, landed concurrently), and #895 replaced the stale `autoMergeWatcherEnabled` `noCLI` prose with `.field(…, read: "automation get", write: "automation set")` — the correct fix. **The file is untouched by this branch now.**

## Pre-existing issue found (not fixed)

**`make test` does not propagate a failing package's exit code.** Its `for` loop returns only the last command's status, so it reported success while CrowCore was failing on an existing test that hard-codes the automation-event list. Out of scope here, but worth its own ticket — each package was run individually for this PR instead.

## Testing

- **All packages pass individually** (`swift test --package-path` per package, exit code checked — not via `make test`, see above). New suites: `IssueTrackerDirectMergeTests` (every gate, log-string stability, the publish/suppress contract), `GitHubAutoMergeAllowedTests`, `AutoMergeStatePayloadTests`, `SessionLifecycleRPCResultTests`.
- **All three parity gates green** (re-run post-rebase): CLI/RPC (94 methods), notification events (11), no-NSLog.
- `docs/cli.md` regenerated via `make docs`.
- **Web**: `row.test.js` gained 17 checks (one per phase colour, the conflict-suppression contract, tooltip + aria carriage, merged short-circuit, old-daemon fallback) and its `.automerge` assertions were rewritten for the folded glyph. 43 pass; the 10 failures are the **pre-existing CROW-802 selector drift** (`.pr-ico` vs SVG `.ico`), verified identical against a clean checkout. Note these tests are not wired into CI.

**Not verified end-to-end against a live disallowing repo** — the direct-merge path is covered by unit tests over its pure predicate, but no PR was actually merged by this branch.


dgershman (changes requested):
## Code & Security Review

Reviewed against `origin/main` (the branch also carried #900/#898, but those have since landed on main; true scope is the single `#888` commit). Built `CrowEngine` clean and ran `IssueTrackerDirectMergeTests` — **17/17 pass**. The merge-safety core is genuinely well done: `shouldDirectMerge` gates on an *explicit* `false` (nil = unknown, never forbidden), re-checks every gate GitHub would have (`CLEAN` + `MERGEABLE` + `SUCCESS` + `APPROVED`), re-verifies Crow authorship on **both** merge paths, and shares one success writer so the two can't drift. The pure predicates are pinned by exhaustive tests, the RPC/notification additions are additive and back-compatible, and the UI deliberately suppresses signals the pill already draws (CROW-773). No security concerns — no injection surface (`gh pr merge <url>`), no secret handling, no auth changes.

### Security Review
**Strengths:**
- Direct merge is strictly narrower than the auto-merge path and re-derives every host-side gate rather than inferring, because it has no host backstop (`directMergeGatesPass`, IssueTracker.swift:2221). `reviewDecision == ""` (no required reviewers) is deliberately excluded — refusing to merge what no human approved is the right side to err on.
- Authorship (`prHasCrowAuthoredCommit`) re-checked before the merge on both the direct-dispatch and enable-catch paths, so a PR no Crow session wrote is never merged.
- `nil`-means-unknown threading is consistent end-to-end (BackendTypes.swift, GitHubCodeBackend.parsePRNode, `mergePRRecords`) and unit-covered — no path treats an unfetched field as "forbidden."

**Concerns:** none.

### Code Quality

**🟡 `notInViewerPRs` publishes an auto-merge verdict without checking the `crow:merge` label** — `IssueTracker.swift:2772`

The watcher-**off** path guards its publish on `Self.hasAutoMergeLabel(pr:)` (line 2731), but the watcher-**on** absent-PR path publishes `.notInViewerPRs` → `.stalled` for *any* session with a `.pr` link whose PR isn't in the viewer fetch (`pr` is `nil` there, so the label can't be checked). Result: an ordinary open feature PR that ages out of the 50-most-recently-updated-open window — and was never armed for auto-merge — renders an orange ⛙ "Auto-merge waiting to retry" chip. Merged/closed PRs are safe (the web short-circuits on `is_merged` in both `prBadgeParts` and `prStatusInline`), and it's transient/self-clearing with no notification, so blast radius is small — but auto-merge vocabulary on a session that never used auto-merge is exactly the kind of misleading signal this PR exists to eliminate, and the asymmetry with the watcher-off guard reads as an oversight. Either gate the publish on prior crow:merge knowledge, or (if the "your PR fell out of Crow's view" signal is intended generically) render it as a fetch-health note rather than an auto-merge chip. If it's deliberate, a one-line justification would settle it.

**🟢 `performDirectMerge` latches *every* failure as permanent** — `IssueTracker.swift:2517`

Unlike `attemptEnableAutoMerge`'s catch (which distinguishes `isPermanentAutoMergeFailure` from transient and clears the marker to retry), `performDirectMerge` writes `directMergeFailed` on any thrown error — so a transient `gh`/network/rate-limit blip permanently blocks the PR *and* fires an Auto-Merge-Blocked notification, needing a manual re-label or daemon restart to retry. The "stop and ask a human on the dangerous path" instinct is sound and aligns with the PR's fail-loud thesis; just flagging that call-failure and merge-refusal are conflated here. Worth a comment noting the trade-off (the escape-hatch `autoMergeDirectFallbackEnabled` toggle you mention would also cover it).

**🟢 Dispatch-time greenness snapshot** — `attemptDirectMerge` (IssueTracker.swift:2478) re-verifies authorship but not the green gates before merging; state could degrade in the async hop. In practice GitHub's own merge enforcement plus the `CLEAN` requirement cover it, and the window is tiny — noting only for completeness.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The Yellow is a small, self-contained fix (or a one-line justification); everything else is genuinely strong work.

dhilgaertner:
Thanks — the 🟡 was a real bug and is fixed in 97f487d. Notes on all three:

### 🟡 `notInViewerPRs` published without checking the label — fixed

You read it exactly right, including that it was an oversight rather than a decision: the watcher-off path guards on `hasAutoMergeLabel` and the watcher-on absent-PR path didn't, so any linked PR aging out of the 50-most-recently-updated window grew an "Auto-merge waiting" chip it never earned.

`pr` is `nil` on that branch, so the label can't be read off the record. The verdict now gates on prior knowledge instead:

```swift
let everArmed = appState.prStatus[session.id]?.hasMergeLabel == true
    || session.autoMergeEnabledAt != nil
```

`applyPRStatuses` only ever writes `prStatus` when the PR *was* in the fetch and never clears it on a miss, so the last-known label survives precisely the gap in question. Neither signal → the stale verdict is cleared rather than replaced.

I kept the **log** line unconditional and gated only the chip. `not-in-viewer-prs` in `crowd-automation.log` is a CROW-782 fetch-health signal that predates this PR and fires for every linked PR missing from the fetch; narrowing it would re-hide the scope/rate-limit problems it exists to surface.

Five regression tests drive the real `applyPRStatuses` → `applyAutoMerge` path (`IssueTrackerNotInViewerPRsTests`). **Three fail against the pre-fix code** — I reverted the guard to confirm the tests actually reproduce your scenario rather than just passing. Two fixture details worth knowing if you touch them: the labeled PR is a **draft**, so the label is recorded without dispatching an attempt (otherwise the next poll hits the in-flight branch, not the one under test), and the second fetch contains *someone else's* PR, because an empty `viewerPRs` short-circuits `applyAutoMerge` before the branch is reached.

### 🟢 `performDirectMerge` latches every failure — kept, comment sharpened

Your framing is better than mine was: call-failure and merge-refusal genuinely are conflated. I kept the behaviour and rewrote the doc comment to say so and to justify why *this* path errs that way — the asymmetry with `enableAutoMerge` is the point. `enableAutoMerge` can retry freely because GitHub holds the queued request and re-checks eligibility server-side, so a wasted attempt costs nothing. A direct merge acts immediately on a snapshot, so an automatic retry loop is the one failure mode that could merge on stale state. Distinguishing the two would mean pattern-matching `gh` stderr, which is exactly the brittleness `repoAutoMergeAllowed` was added to escape.

The comment now also names the exit: if the false-permanent rate proves annoying, the fix is bounded retries keyed on `headRefOid` (as `autoUpdateBranchAttempted` does), **not** a looser catch.

One thing I'd flag that neither of us wrote down: a latched skip only clears on daemon restart. That's consistent with the pre-existing CROW-621 permanent skips, so it's not a regression, but "re-apply the label to retry" does *not* work and the verdict message correctly doesn't claim it does.

### 🟢 Dispatch-time greenness snapshot — no change, reasoning below

Agreed the window is tiny, but I want to be precise about what actually backstops it, because it isn't the re-check you might expect: re-running `directMergeGatesPass` at the top of `attemptDirectMerge` would re-read the *same captured `PRRecord`*, so it would pass by construction and prove nothing. A real check means re-fetching, i.e. an extra API call per merge to close a millisecond window.

What covers it instead:
- **Branch protection** — on a protected repo GitHub refuses the merge outright if required checks/reviews aren't satisfied, independent of our snapshot.
- **`mergeStateStatus == "CLEAN"`** at dispatch, which already excludes `BLOCKED`/`UNSTABLE`/`BEHIND`/`DIRTY`.
- **`autoMergeInFlight`** prevents a second dispatch, so two polls can't race into a double merge.

The genuinely uncovered case is an *unprotected* repo whose checks flip red in the milliseconds between dispatch and merge. I don't think that's worth an API call per merge, but happy to add the re-fetch if you'd rather pay for it.

dgershman (approved):
## Code & Security Review

Verified locally: checked out the branch, built `CrowProvider` / `CrowEngine` clean, and ran every new suite — all green (`GitHubAutoMergeAllowedTests` 6, `IssueTrackerDirectMergeTests` + `SessionLifecycleRPCResultTests` 20, `AutoMergeStatePayloadTests` 4, `NotificationEventTests` 17, `row.test.js` 43 pass). I reconstructed `main`'s web harness and confirmed the 10 `row.test.js` failures are the pre-existing CROW-802 `.pr-ico`/glyph selector drift: `main` fails the same 10 **plus** `🏷 and ⛙ can both show`, which this PR fixes. The testing claims in the description hold.

### Critical Issues
None.

### Security Review
**Strengths:**
- The one genuinely dangerous new operation — an irreversible direct merge with no GitHub-side gate — is behind the strictest predicate in the module (`shouldDirectMerge` → `directMergeGatesPass`): one-shot guard, OPEN, non-draft, labeled, non-conflicting, non-changes-requested, `mergeStateStatus == "CLEAN"`, `MERGEABLE`, `checksState == "SUCCESS"`, `reviewDecision == "APPROVED"`, **and** `repoAutoMergeAllowed == false` explicitly. Crow authorship is re-verified in `attemptDirectMerge`/the catch path before the merge fires. Every failure mode errs toward *not* merging.
- `reviewDecision == ""` (repo with no required reviewers) is deliberately excluded — refusing to merge something no human approved is the right side to err on, and `refusesWithoutAnApproval` pins it.
- `repoAutoMergeAllowed` treats `nil` as *unknown*, never *forbidden* (`GitHubCodeBackend.swift:498`, `BackendTypes.swift:86`), so a partial/SAML/GitLab fetch can't turn every PR into a direct merge. `anAbsentFieldIsUnknownNotForbidden` guards it.
- `mergeNow` shells out via an args array (`["gh","pr","merge",prURL,"--squash","--delete-branch"]`) — no shell interpolation, and `prURL` is trusted GraphQL output. No injection surface.

**Concerns:** None.

### Code Quality
- Concurrency is correct: `recordAutoMergeSuccess` writes through the shared `store` (not a throwaway `JSONStore`, per #728), and `autoMergeInFlight` is inserted before the `Task` dispatch, so a slow merge can't be double-dispatched across polls. Matches the existing `attemptEnableAutoMerge`/`attemptUpdateBranch` shape.
- The wire contract is well-drawn: `auto_merge_state` is a sibling of `pr` (so it survives a session with no `PRStatus`), absent-not-null so old clients degrade to the persisted `auto_merge` bool, and `okResult`'s `warning` is omitted-when-empty so `jq -e .warning` stays clean. All three pinned by tests.
- Notification/UI parity maintained across the Swift enum, `app.js`, and `settings.js`; `autoMergeBlocked` correctly latches once per (PR, reason).

### Green (consider) — non-blocking
- The `nil`→latch backstop: on the *first* poll after upgrade, a cached pre-#888 record (`repoAutoMergeAllowed == nil`) for a disallowing repo whose PR isn't yet green can still take the failed-`--auto` path and latch `repo-disallows-auto-merge` permanently until a daemon restart re-evaluates with the field now populated. The proactive detection this PR adds means the common GitHub case never hits it, and it self-heals on restart — strictly better than the pre-PR "always latched, never announced" behavior. Worth an eventual note but not a blocker.
- `autoMergeEnabledAt` is reused as the one-shot guard for direct merges (slight name/meaning drift for the `.merged` phase); it's documented and functionally correct.

### Summary Table

---

PR #900, opened 2026-07-28
Repair hook blocks that outlived the worktree their crow binary came from

### LINKED ISSUE #897: Stale hook commands point at a deleted worktree's dev binary — every hook event fails, telemetry silently lost
## Summary

Every Claude Code hook event fails in the `corveil` **main clone** because its `.claude/settings.local.json` still contains hook commands written by an **old Crow dev build**, pointing at a `crow` binary inside a worktree that has since been deleted.

Observed as:

```
Ran 2 stop hooks
  ⎿  Stop hook error: Failed with non-blocking status code: /bin/sh:
     /Users/<user>/Dev2/RadiusMethod/crow-136-session-analytics-otel/.build/arm64-apple-macosx/debug/crow:
     No such file or directory.
```

The errors are non-blocking, so the session still works — but every hook event (Stop, PreToolUse, PostToolUse, UserPromptSubmit, …) emits noise, and **no telemetry/analytics is recorded for those sessions**.

## Evidence

`{devRoot}/RadiusMethod/corveil/.claude/settings.local.json` (mtime **Apr 13 16:12**) — all **17** hook entries:

```json
"command": "/Users/<user>/Dev2/RadiusMethod/crow-136-session-analytics-otel/.build/arm64-apple-macosx/debug/crow hook-event --session 2BEF7A0D-743B-47A3-819A-C39C180F6977 --event Stop"
```

Two things are stale in every one of those lines:

1. **The binary path** — `crow-136-session-analytics-otel` was a worktree for CROW-136 (session analytics / OTel). It no longer exists; the retention reaper removed it (`cleanup.enabled: true`, `retentionHours: 1`).
2. **The session UUID** — `2BEF7A0D-743B-47A3-819A-C39C180F6977` is gone: `crow get-session` returns `Error -32000: Session not found`. Even if the binary existed, every event would be attributed to a dead session.

Scan across all clones/worktrees under devRoot — only the main clone is affected:

| Path | hook lines | stale |
|---|---|---|
| `RadiusMethod/corveil/` (main clone) | 17 | **17** |
| `RadiusMethod/corveil-1815-embeddings-cost-analysis/` | 17 | 0 |
| `RadiusMethod/crow-806-adr-control-plane-parity/` | 17 | 0 |
| `RadiusMethod/crow-809-workspaces-cli/` | 17 | 0 |
| `RadiusMethod/crow-888-merge-label-silent-skip/` | 17 | 0 |
| `RadiusMethod/crow-891-review-gateway-resolution/` | 17 | 0 |
| `RadiusMethod/crow-894-saml-null-nodes/` | 17 | 0 |

What current Crow writes, for contrast — the stable symlink, not a build product:

```json
"command": "/Users/<user>/Dev2/.claude/bin/crow hook-event --session 135BED88-14CF-4032-857C-50631A0074EE --event Notification"
```

## Root cause

Two compounding issues:

1. **An old Crow build baked its own `.build/…/debug/crow` path into the hooks it wrote.** Today `.claude/bin/crow` is a stable symlink that survives worktree churn (`Dev2/.claude/bin/crow → RadiusMethod/crow/.build/arm64-apple-macosx/debug/crow`), so newly written hooks are fine. The Apr 13 file predates that and is a permanent artifact of a dev build running out of a worktree.

2. **Nothing ever revisits `settings.local.json` in a main clone.** `setup.sh` / the daemon write that file into **new worktrees** at session-creation time. A main clone gets one written only if a session was once created pointing at it — and after that it is never rewritten, validated, or pruned. So the stale block survives indefinitely, and the reaper that deleted the worktree had no reason to touch it.

The session UUID staleness is inherent to the design: hook commands hardcode `--session <uuid>`, so a settings file outlives the session it was written for. That's fine for a worktree (deleted with the session) but wrong for a long-lived main clone.

## Impact

- Hook-event noise on every single event in any session run from an affected clone.
- **Silent telemetry loss** — `hook-event` never reaches `crowd`, so those sessions produce no analytics/scorecard data despite `telemetry.enabled: true`.
- Confusing failure text: it names a worktree the user may not recognize and has long since deleted.
- Any main clone that ever hosted a session is a candidate; this will recur as dev builds move between worktrees.

## Suggested directions

- **Validate on launch.** Before launching an agent, check the resolved hook binary in the target dir's `settings.local.json`; if it doesn't exist (or the `--session` uuid is unknown), rewrite the hook block instead of launching with dangling commands.
- **Never write a build-product path.** Always emit `{devRoot}/.claude/bin/crow` (or resolve via PATH). Worth an assertion so a dev build can't bake `.build/…` in again.
- **Prune on reap.** When the retention reaper deletes a worktree, scrub hook blocks referencing that worktree's binary or its session uuid from any `settings.local.json` still on disk.
- **Consider making `--session` resolution lazy** — e.g. hooks pass the cwd and let `crowd` map it to a live session — so a settings file can't outlive its session id.
- **One-shot repair** for existing installs: detect a dangling hook binary at startup and offer to rewrite/remove.

## Acceptance ideas

- [ ] Launching a session in a clone whose `settings.local.json` names a nonexistent binary results in a rewritten hook block, not repeated hook errors.
- [ ] No Crow-written hook command contains `.build/` — verified by a test.
- [ ] Deleting a worktree removes (or repairs) hook entries elsewhere that reference its binary/session.
- [ ] A session whose hooks were repaired records telemetry normally.

dhilgaertner:
Closes #897

## The bug

An old dev build baked its own `.build/{arch}/debug/crow` into every hook command it wrote. That worktree was later reaped and the session it named deleted — but nothing in Crow ever revisits a `settings.local.json` once written. So `{devRoot}/RadiusMethod/corveil/.claude/settings.local.json` (a long-lived **main clone**, mtime Apr 13) kept failing all 17 hook events with `No such file or directory`, and every session run there recorded **no telemetry at all**.

Two compounding causes, both addressed here.

## 1. Never emit a build product

`findCrowBinary(devRoot:)` preferred `{devRoot}/.claude/bin/crow` but was read-only — and `isExecutableFile` follows symlinks, so a *dangling* link fell through to the raw build path just like a missing one.

`resolveCrowBinary` now **ensures** the symlink (creating or re-pointing it) and returns the link. Hook commands name a path that survives worktree churn: only the target moves, and every boot re-points it, so one relink heals every settings file at once. All six writers that persist a crow path use it; `findCrowBinary` stays for callers that only need to *locate* a crow to run now.

`Scaffolder.installCrowCLISymlink` moved to `ClaudeHookConfigWriter.ensureCrowCLISymlink` so the symlink and the commands depending on it have one owner. It also gained a no-op fast path when the link already points at the target — the old unlink+relink left a window where a concurrently firing hook saw ENOENT.

**Also fixed while here:** the path was interpolated raw, so a dev root containing a space (`/Users/x/My Dev`) silently killed all 17 hooks. Now shell-quoted, matching `CursorHookConfigWriter` and `AntigravityHookConfigWriter`, which already quote for exactly this reason. It had to land with the sweep anyway, since the parser must handle both forms.

## 2. Repair what is already on disk

`ClaudeHookRepair.sweep` runs once at daemon boot — after the scaffold re-points the symlink, before `startBoardPoll` spawns the Manager. It is **filesystem-driven, not state-driven**: the broken directory's session is gone from `AppState`, which is precisely why enumerating live worktrees would miss it. Depth-2 under devRoot (`{ws}/{repo-or-worktree}`), plus devRoot itself for the Manager.

| State | Action |
|---|---|
| All commands resolve + session live | **No write at all** (contents and mtime untouched) |
| Stale, directory maps to a live worktree | Commands rewritten in place |
| Stale, no live session | Managed entries removed; file deleted only if nothing else remains |

Both mutations are surgical rather than reusing the existing writers: `writeHookConfig` expands a partial file to all 17 events and replaces whole event arrays, and `removeHookConfig` drops all 17 keys wholesale — taking a user's own `Stop` hook with it. (`writeManagerHookConfig` already documents refusing to run it against the dev root for that reason.) Writes are non-atomic so the `0600` that `writeGatewayEnv` applies — the `env` block can hold a gateway bearer token — survives.

**The parser is the entire safety boundary**, so it is strict: absolute path, exactly `--session <uuid> [--agent <kind>] --event <managed name>`, no shell metacharacters. A user hook that merely *wraps* `crow hook-event` in a pipeline stays theirs. Every mutated path is logged at INFO, and `CROW_HOOK_REPAIR=0` disables the sweep.

## 3. Stop trusting an unknown `session_id`

`hook-event` trusted a provided uuid unconditionally, minting hook state and a persisted `hookStates` row for sessions that no longer exist (nothing prunes those; `reseed` only filters on read). It now prefers the worktree matching the payload cwd when the id isn't live, and skips the store write for a dead session.

Deliberately **still never errors**: `rpc()` throws on any server error and `forwardHookEvent` catches only `connectionFailed`, so an `invalidParams` exits non-zero — which Claude renders as the exact "hook error" noise this PR removes. Strictly monotone: nothing that routes today stops routing.

## Verification

- `make build` clean; **2,234 tests across 16 packages pass**.
- New: `ClaudeHookRepairTests` (parse table incl. quoted / legacy-unquoted / unquoted-with-spaces / pipeline-wrapped / relative / bad uuid; repair-in-place, surgical strip, Manager-at-devRoot, `0600` preservation, byte-identical healthy file, traversal skips), `HookEventSessionResolutionTests`, and crow-path tests asserting **no emitted command contains `/.build/`** — the issue's explicit acceptance criterion.
- **Verified against the real artifact**: a copy of the actual `corveil/.claude/settings.local.json` in a reproduced devRoot layout is stripped and deleted (it held nothing but the stale block), while a healthy neighbor stays byte-identical. The original on disk was left untouched.

## Rebased onto main

Rebased past #895 (ADR 0016 parity contract) and #885 (Workspaces CLI). This branch originally carried a second commit adding the missing `automation-get` / `automation-set` ledger rows, because `make parity` was red on `main`. #895 fixed that independently — same eleven config-field rows, plus a better rationale for why the automation RPCs stay reachable over remote `/rpc` — so that commit was **dropped rather than merged**, and this PR is a single commit again.

Only `AppConfig.swift` and `SessionService.swift` overlapped with upstream; both auto-merged, and all four `resolveCrowBinary` swaps plus the doc reference were verified intact afterward. No new `findCrowBinary` call sites arrived upstream.

`make test` is green on the rebased tree: all 16 package suites, parity (94 methods), notification catalogs, and the `NSLog` gate.

### DIFF
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
@@ -27,13 +27,24 @@ public struct ClaudeHookConfigWriter: HookConfigWriter {
 
     // MARK: - Generate Hook Configuration
 
+    /// The exact hook command Crow writes. Single source of truth shared with
+    /// `ClaudeHookRepair`, which both parses this shape to decide what it owns
+    /// and re-emits it when repairing a stale entry (#897).
+    ///
+    /// `crowPath` is shell-quoted: Claude runs hook commands through `/bin/sh
+    /// -c`, so an unquoted dev root containing a space (`/Users/x/My Dev`)
+    /// silently broke all 17 hooks. Matches `CursorHookConfigWriter` and
+    /// `AntigravityHookConfigWriter`, which already quote for this reason.
+    static func hookCommand(crowPath: String, sessionID: UUID, event: String) -> String {
+        "\(ClaudeLaunchArgs.shellQuote(crowPath)) hook-event --session \(sessionID.uuidString) --event \(event)"
+    }
+
     /// Generate the hooks dictionary for a session.
     static func generateHooks(sessionID: UUID, crowPath: String) -> [String: Any] {
-        let sid = sessionID.uuidString
         var hooks: [String: Any] = [:]
 
         for event in allEvents {
-            let command = "\(crowPath) hook-event --session \(sid) --event \(event)"
+            let command = hookCommand(crowPath: crowPath, sessionID: sessionID, event: event)
             var hookEntry: [String: Any] = [
                 "type": "command",
                 "command": command,
@@ -196,7 +207,7 @@ public struct ClaudeHookConfigWriter: HookConfigWriter {
     /// Resolve the running app's own `crow` CLI — the bundled binary in a
     /// release `.app` (`Contents/MacOS/crow`), or `.build/{config}/crow` in
     /// dev. Does not consult `{devRoot}/.claude/bin/crow`; use that for
-    /// agent-facing resolution via `findCrowBinary(devRoot:)`.
+    /// agent-facing resolution via `resolveCrowBinary(devRoot:)`.
     public static func appCrowBinary() -> String? {
         // Same directory as the running executable (dev + release bundles).
         let execURL = URL(fileURLWithPath: ProcessInfo.processInfo.arguments[0])
@@ -226,6 +237,12 @@ public struct ClaudeHookConfigWriter: HookConfigWriter {
     /// Find the `crow` binary agents and hook configs should invoke. Prefers
     /// `{devRoot}/.claude/bin/crow` when scaffolded and executable (CROW-552),
     /// then falls back to `appCrowBinary()` and common install locations.
+    ///
+    /// Read-only — it never *creates* the symlink, so a dev build whose link is
+    /// missing or dangling falls through to a `.build/…/debug/crow` path that
+    /// dies with the worktree it was built in. Anything whose result is written
+    /// to disk must use `resolveCrowBinary` instead (#897); this stays for
+    /// callers that only need to *locate* a crow to run right now.
     public static func findCrowBinary(devRoot: String? = nil) -> String? {
         if let devRoot {
             let symlink = (devRoot as NSString).appendingPathComponent(".claude/bin/crow")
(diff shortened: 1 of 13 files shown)


dgershman (approved):
## Code & Security Review

Focused, well-scoped fix for #897 (stale hook binary/session paths) with strong test coverage. I built the tree clean (`swift build`, 53s) and ran the affected suites: **77/77 CrowClaude tests** and the filtered **CrowEngine suites (HookEventSessionResolution / AgentLaunch / DefaultsRPCSupport) all green**.

### Critical Issues
None.

### Security Review
**Strengths:**
- The repair sweep is the only code that mutates a `settings.local.json` Crow did not just write, and its parser (`parseCrowHookCommand`) is a genuinely tight allowlist: absolute path required, exact `hook-event --session <uuid> [--agent <kind>] --event <managed name>` grammar, and any of `; | & > < \` $ ( ) \n` disqualifies the command. A user hook that merely *wraps* `crow hook-event` in a pipeline is correctly left untouched (verified by `rejectsAnythingThatIsNotExactlyOurShape`).
- Non-atomic in-place write is a deliberate, correct choice: it preserves the `0600` inode that `writeGatewayEnv` sets because the `env` block can carry a gateway bearer token. `repairPreservesOwnerOnlyPermissions` pins this.
- `ensureCrowCLISymlink` only ever owns symlinks — a real file at `.claude/bin/crow` is never replaced (`ensureNeverReplacesANonSymlinkFile`), and `attributesOfItem` (lstat) is used throughout so dangling links and symlinked-out workspaces are handled without following the link.
- Shell-quoting the crow path (`ClaudeLaunchArgs.shellQuote`) closes the dev-root-with-a-space hole, and the parser round-trips both the quoted and legacy-unquoted forms.

**Concerns:** None.

### Code Quality
- Boot ordering is explicit and correct: `repairStaleHooks` runs after `LaunchScaffold.run` re-points the symlink and after `seedAppState`, but before `startBoardPoll`'s `ensureManagerSession` — so it never races the Manager's own writer. The Manager block (which has no worktree row) is matched by the `managerSessionID` constant rather than path lookup, preventing an errant strip (`managerBlockAtDevRootIsRepairedNotStripped`).
- The FS walk is offloaded to a detached utility task, keeping the blocking I/O off the MainActor (#892), and the hot `crow send` path in `EngineRouter` hoists `resolveCrowBinary` off the actor for the same reason.
- `Dictionary(..., uniquingKeysWith:)` correctly avoids the duplicate-key trap for two rows sharing a worktree path.
- The `hook-event` resolution change is strictly monotone — a live provided id still routes verbatim, a dead one falls back to cwd, and an unresolvable one still returns success (never a non-zero exit / "hook error" noise) while declining the `hookStates` write. `liveSessionIDWins` / `unknownSessionFallsBackToCwd` / `unresolvableSessionDoesNotErrorOrPersist` cover all three arms.
- `rewriteCommands`/`stripCrowEntries` deliberately preserve the file's existing shape and user matcher groups rather than reusing the blunt `writeHookConfig`/`removeHookConfig`; the healthy-file path writes nothing at all (`leavesAHealthyFileByteIdenticalAndUnwritten`).

### Green considerations (non-blocking)
- The four `SessionService` `findCrowBinary` → `resolveCrowBinary` swaps run on session-lifecycle paths that already performed stat I/O via the predecessor, so there's no MainActor regression — just noting the pattern for future awareness.
- In the rare orphan case where two sessions share a worktree path, `desiredSession` isn't liveness-checked before a rewrite, so a repair could name a not-yet-live id. Benign (the next hook simply skips persistence), and already acknowledged via `uniquingKeysWith`.

### Summary Table

---

Issue #829, opened 2026-07-23
Cursor adapter: wire resume, print-mode, auto-permission, MCP, per-project hooks


dgershman:
Spin-off from the harness gap audit (#828); baseline #827. See `docs/agent-harness-gap-audit.md` §3a.

`agent` 2026.07.23 has grown the features `CursorAgent` was written without. Close, in one adapter+launcher+hook-writer+tests pass:

- **Resume** — on `.job`/`.review` restart use `agent --continue` (or `--resume <chatId>`) instead of the bare-`agent` cold TUI. (landed CLI 2026-01-16)
- **Non-interactive / headless** — dispatch jobs & review via `agent -p --output-format stream-json` so completion is parsed, not scraped.
- **Auto-permission** — honor `autoPermissionMode` for `.job`: `--force`/`--yolo` (or `--auto-review`); `--trust` seeds workspace trust.
- **MCP** — register the Jira MCP (already written for Claude) via `cursor-agent mcp add`.
- **Per-project hooks** — write per-worktree `.cursor/hooks.json` with the Crow session UUID baked in, closing the shared-`cwd` collision. **Caveat:** verify CLI event coverage (community reports only a subset of events fire in the CLI) against `CursorSignalSource` before removing cwd-match.
- **Auto-wire `CursorLauncher`** — with positional prompt (`.work`) + `-p` (jobs), feed the generated prompt at launch instead of a bare TUI.

Upstream: https://cursor.com/docs/cli/using · https://cursor.com/changelog/cli-jan-16-2026 · https://cursor.com/docs/hooks

dgershman:
**Scope correction (from #832 review):** `--auto-review` is present in `cursor-agent --help` (2026.07.23) but is **not** in the [CLI parameter reference](https://cursor.com/docs/cli/reference/parameters) — treat it as unverified/unstable. Do **not** default to it as a "safer classifier" (different security posture). Use the documented `--force` for `autoPermissionMode`; `--trust` is headless-mode only per the reference. See `docs/agent-harness-gap-audit.md` §3a.

dgershman:
Scope correction (from #832 round-2): the bounded auto-permission default is `--force --sandbox enabled` (approval off, **sandbox on**) — the analogue of Codex's `-a never -s workspace-write`. Do **not** default to bare `--force`/`--yolo` (approve *and* no sandbox); that's the unbounded posture.

PR #901, opened 2026-07-28
Cursor: wire interactive --trust as a per-launch workspace-trust seed

### LINKED ISSUE #890: Cursor: wire interactive --trust (ADR 0015 re-check)
## What changed upstream

Cursor CLI changelog **July 20, 2026** claims `--trust` now works in interactive sessions (previously headless-only): it trusts the workspace up front, skips the trust dialog, and records the same saved trust decision as accepting the dialog.

Evidence:
- https://cursor.com/docs/cli/changelog (July 20, 2026 — "`--trust` works in interactive sessions")
- Conflicting official params still say headless-only: https://cursor.com/docs/cli/reference/parameters (`--trust` — "Trust the workspace without prompting (headless mode only)")

Upstream min version: Cursor CLI release dated **2026.07.20** (changelog); re-probe against current `agent --version` before wiring.

## Matrix / ADR 0015 gap this closes

- Matrix row: **Gateway env / trust seed / telemetry** — Cursor is currently ❌ (Claude has `ClaudeTrustSeeder`; Cursor residual called out in auto-permission notes as "fresh worktree may prompt")
- Gap audit §3a / #829 intentionally omitted `--trust` as headless-only per the parameter reference
- ADR 0015 honesty: this is a bounded, per-launch trust seed (analogue of Claude trust seeding), **not** `--yolo` / full-bypass

## ADR-0015 honesty assessment (audit)

**needs-eval** — changelog and params docs disagree. Wire only after empirical confirmation that interactive `agent --trust …` skips the dialog without requiring `-p`/`--print`, and that it does not broaden permissions beyond workspace trust (still pair with existing `--force --approve-mcps` for auto-perm; do **not** adopt `--yolo`).

## Scope

Implement the wiring (likely `CursorLaunchArgs` / auto-launch paths for fresh worktrees + Manager) once confirmed; update matrix "Last verified" + ADR 0015. A human/`crow:auto` worker does the code — this ticket does not.

dgershman:
Closes #890

## What upstream changed

Cursor CLI's `--trust` **went interactive on 2026.07.20** ([changelog](https://cursor.com/docs/cli/changelog): "`--trust` works in interactive sessions"), reversing the earlier gap-audit read of it as headless-only. The [parameter reference](https://cursor.com/docs/cli/reference/parameters) is stale on this — but the installed binary's own help is authoritative and disagrees.

## Empirical confirmation (the ADR-0015 gate)

Re-probed against the installed `agent 2026.07.23` (newer than the changelog date). Its `--help` lists:

```
--approve-mcps   Automatically approve all MCP servers (default: false)
--trust          Trust the current workspace without prompting (default: false)
```

`--trust` is a **general** flag: **no "(headless mode only)" qualifier** and **no "only works with --print" gating** (unlike `--output-format` / `--stream-partial-output`, which carry that note). So interactive `agent --trust …` skips the folder-trust dialog without `-p`/`--print`, and it does not broaden permissions beyond workspace trust.

## The wiring

Per-launch analogue of `ClaudeTrustSeeder` (which pre-writes `~/.claude.json`) and `CodexTrustSeeder` (`config.toml` `trust_level`) — Cursor tracks trust per workspace and it doesn't inherit, so a fresh worktree/review clone would otherwise block an auto-launched session on the folder-trust prompt.

- **`CursorLaunchArgs.trustSuffix`** (` --trust`), unconditional, + **`launchSuffix(autoPermissionMode:)`** = trust seed, then `--force --approve-mcps` when auto-permission is on.
- **`CursorAgent.autoLaunchCommand`** (`.work`/`.job`/`.review`) and **`managerLaunchCommand`** route through `launchSuffix`.
- **`CursorLauncher`** handoff one-shot carries `trustSuffix` **alone** — trust is orthogonal to auto-permission, so handoff still omits `--force`. This covers the same four seed sites Claude seeds (launch, Manager ×2, handoff).

**Bounded to workspace trust — NOT `--yolo`/full-bypass.** Auto-permission stays in the separate `--force --approve-mcps`; per-tool approval is unchanged. Closes the "fresh worktree may prompt" residual [#829](https://github.com/corveil/crow/issues/829) deliberately left open.

## Docs

- **matrix** — "Gateway env / trust seed / telemetry" Cursor cell `❌` → `⚠️ trust seed only (`--trust`, per-launch)` (parallels Codex); auto-permission note + trust-seed prose updated; new dated row in the version-pinned re-check table (**Last verified 2026-07-28**).
- **ADR 0015** — dated amendment (#890) framing this as an honest, verified flag under the existing "honest gap over faked capability" principle, explicitly bounded (not `--yolo`).
- **gap-audit §3a** — the residual annotated **closed** (history preserved per repo convention).

## Test plan

`swift test --package-path Packages/CrowCursor` — 60 tests pass. New/updated unit tests assert the exact suffixes:

- `launchSuffix(false) == " --trust"`, `launchSuffix(true) == " --trust --force --approve-mcps"`, never `--yolo`.
- `.work` → `agent' --trust\n`; Manager plain → `agent' --trust`; job/auto → `--trust` present alongside `--force --approve-mcps`; resume still ends `--continue`.

Runtime "dialog does not appear in an interactive TTY" is not automatable here (needs a genuinely-untrusted fresh workspace + TTY); the installed-binary `--help` evidence above plus the generated-command unit tests are the verification surface.


dhilgaertner (changes requested):
## Code & Security Review

Reviewed the full diff (8 files, +149/−43), ran the package tests, and independently re-verified the upstream claim this PR rests on.

### Upstream claim: independently confirmed ✅

The ADR-0015 empirical gate holds. Fetching the [Cursor CLI changelog](https://cursor.com/docs/cli/changelog) directly returns, dated **2026-07-20**:

> "Previously `--trust` required headless mode. Passing it in an interactive session now trusts the workspace up front and skips the trust dialog, recording the same saved trust decision the dialog writes when you accept."

That matches `CursorLaunchArgs.trustSuffix`'s doc comment almost word for word. I also confirmed the [parameter reference](https://cursor.com/docs/cli/reference/parameters) is still stale — it reads *"Trust the workspace without prompting (headless mode only)"* — exactly as the PR describes. The reversal of the #829 headless-only omission is well-founded.

I could **not** re-verify the `agent 2026.07.23 --help` output: no Cursor CLI is installed on this review host. The changelog corroborates the behavioral claim, which is the part that matters.

### Critical Issues

None. No Red findings.

### Security Review

**Strengths:**
- **Correctly bounded.** `--trust` is kept strictly separate from auto-permission: `trustSuffix` is unconditional, `--force --approve-mcps` stays gated on the caller's opt-in, and `launchSuffix` composes them without collapsing the two concerns. `CursorLaunchArgsTests.swift:38` explicitly pins that `--yolo` never appears. This is the right shape — the tempting shortcut (`--yolo` and be done) was correctly refused.
- **Handoff deliberately keeps auto-permission off** while still carrying the trust seed (`CursorLauncher.swift:99`). Trust really is orthogonal to approval, and the split is reasoned in the doc comment rather than incidental.
- No injection surface added. `--trust` is a fixed literal appended between the already-shell-quoted binary path and the quoted positional prompt; `shellQuote` handling is unchanged, and the flag can't be influenced by session data.
- `AgentLaunch.commandLaunchesToken` (`AgentLaunch.swift:67`) still matches the generated commands — the regex anchors on `/agent` followed by whitespace, so the inserted ` --trust` is harmless.

**Concerns:**
- 🟢 **Auto-trust on attacker-authored review clones.** `.review` sessions clone third-party PR branches, and `SessionService.swift:975` turns auto-permission on for them by default. After this change a Cursor review session on an untrusted branch launches with `--trust --force --approve-mcps`: trust auto-granted *and* per-tool approval off. Workspace trust is precisely the gate that governs whether repo-local agent config (`.cursor/rules`, repo-supplied instructions, MCP entries) is honored, so "bounded to workspace trust" is not quite the same as "no security surface." This matches the posture `ClaudeTrustSeeder` + `--permission-mode auto` already ship, so it's consistency rather than a new exposure — but the ADR amendment (`docs/adr/0015-harness-capability-tiers.md:26`) asserts the bound without naming the residual. Suggest one sentence acknowledging that trust is the gate for repo-supplied agent config, so a future reader doesn't read "bounded" as "inert."

### Code Quality

- 🟡 **No minimum-CLI-version floor recorded, and the older-build failure mode is unverified** — `CursorLaunchArgs.swift:42`. `trustSuffix` is unconditional across *every* Crow-driven launch path (`.work`, `.job`, `.review`, Manager, handoff), and interactive `--trust` is 8 days old as of this PR. The changelog's own wording is "Previously `--trust` **required** headless mode" — "required" reads as enforced, not ignored, so on a pre-2026.07.20 `agent` build the interactive invocation may be **rejected** rather than silently no-op'd. If so, this doesn't degrade to the old fresh-worktree prompt; it breaks every Cursor session for any user who hasn't updated their CLI.

  The PR applies the ADR-0015 empirical gate to the *current* build but not to the *floor*. The version-pinned re-check row at `docs/agent-harness-matrix.md:472` only guards forward drift ("re-probe if `--help` ever restores the headless-only qualifier"); nothing anywhere states a minimum. Two cheap fixes, either is fine:
  - Probe `agent --trust` on a pre-2026.07.20 build and record the result. If it's a silent no-op, say so in the doc comment and this concern evaporates.
  - Record **"requires Cursor CLI ≥ 2026.07.20"** in the matrix re-check row and the ADR amendment, so the blast radius is at least documented.

  I'd note the repo has no precedent for runtime version gating (Antigravity's `v1.1.7` pins are doc-only), so a documented floor is in keeping with house style — the gap is that neither the floor nor the failure mode is written down anywhere.

- 🟡 **The handoff seed site has no test** — `Packages/CrowCursor/Tests/CrowCursorTests/CursorLauncherTests.swift`. The PR body, the ADR amendment, and the matrix all claim four seed sites (launch, Manager ×2, handoff). Three have assertions; `CursorLauncher.launchCommand` has none, and it's the *only* site that applies `trustSuffix` directly instead of going through `launchSuffix` — so the shared-helper tests can't catch a regression there. `CursorLauncherTests.swift` is untouched by this PR and only covers `generatePrompt`. One assertion closes it:

  ```swift
  @Test func launchCommandCarriesTrustSeedNotAutoPermission() throws {
      let cmd = try CursorLauncher().launchCommand(
          sessionID: UUID(), worktreePath: "/w", prompt: "p")
      #expect(cmd.contains("'agent' --trust \"$(cat "))
      #expect(cmd.contains("--force") == false)   // handoff omits auto-permission
  }
  ```

- 🟢 **`--trust` is redundant on the `--continue` resume path** (`CursorAgent.swift:127`). Per the changelog, the first launch already "record[s] the same saved trust decision," so the workspace is trusted by the time a resume happens. Harmless, and keeping one unconditional suffix is simpler than special-casing resume — noting it only so the redundancy reads as intentional rather than overlooked.

### Docs

Accurate and complete. The matrix cell (`❌` → `⚠️ trust seed only`) correctly parallels Codex's, §3a of the gap audit is annotated closed rather than rewritten (per repo convention), and the ADR amendment is honest about *why* this doesn't relax the "honest gap over faked capability" principle — the flag is real and verified, so emitting it is honest. The stale-param-reference caveat is documented in all three places, which is exactly right given I confirmed the reference really is stale.

### Tests

`swift test --package-path Packages/CrowCursor` → **60 tests pass**, verified locally. The new assertions pin the exact suffixes (`" --trust"`, `" --trust --force --approve-mcps"`), and the flipped expectations in `CursorAgentTests.swift` (`--trust` false → true) are correct inversions rather than deletions — the negative assertions on `--yolo`, `--sandbox`, and `--auto-review` are all retained.

### Summary Table

**Recommendation:** Request Changes — driven by **0 Red, 2 Yellow, 2 Green** findings.

This is careful, well-reasoned work and the core wiring is right — the split between trust and auto-permission is exactly the distinction that matters, and the empirical gate was genuinely met for the current build. Both Yellows are small: document (or probe) the minimum CLI version, and add one assertion for the handoff seed site.

---

dgershman:
Thanks for the careful review — and for independently re-confirming the changelog claim. All four findings addressed in dc2a66e:

**🟡 Min-CLI floor + unverified older-build failure mode.** You're right that "reject vs no-op on a pre-07.20 build" is the real blast-radius question, and I don't have an old binary to probe it. Recorded the floor and the reasoning in three places (`trustSuffix` doc comment, matrix re-check row, ADR amendment): `--trust` has been a *recognized* flag since before 2026.07.20 (the param reference lists it, headless-only), so an older `agent` **parses** it rather than erroring on an unknown option — the pre-07.20 "headless mode only" restriction gated the flag's *effect*, not a parse-time rejection keyed on other flags, so an interactive launch most likely **no-ops and degrades to the old fresh-worktree prompt** rather than breaking. That last step is flagged as **reasoned, not probed**, with the escape hatch named (gate emission on a version check if an older-CLI user reports broken launches). Kept it doc-only per house style (Antigravity's pins are doc-only; no runtime version-gating precedent).

**🟡 Handoff seed site untested.** Added `launchCommandCarriesTrustSeedNotAutoPermission` to `CursorLauncherTests` — asserts `'agent' --trust "$(cat ` present and `--force` absent. It's `async throws` (actor), otherwise your snippet verbatim. This was the only site applying `trustSuffix` directly, so the shared-helper tests genuinely couldn't guard it. Suite now 61.

**🟢 ADR "bounded" residual.** Added a sentence: workspace trust is the gate for repo-supplied agent config (`.cursor/rules`, MCP entries, repo instructions), so "bounded" means *scoped, not inert* — and on `.review` clones of third-party branches this is the same trust-granted + approval-off posture `ClaudeTrustSeeder` + `--permission-mode auto` already ship (consistency, not new exposure).

**🟢 Resume redundancy.** Noted in the `launchArgs` comment that the seed riding `--continue` is harmless and intentional (the first launch already recorded the saved trust decision; one unconditional suffix beats special-casing resume).

Re-requesting your review.

dhilgaertner (changes requested):
## Code & Security Review

The upstream research here is genuinely good — the changelog + installed-binary `--help` cross-check is exactly the ADR-0015 empirical gate, the flag is correctly kept orthogonal to `--force --approve-mcps`, all three seed sites are covered, and the tests pin the exact suffixes. `swift build` (root package) and `swift test --package-path Packages/CrowCursor` both pass locally (61 tests). `AgentLaunch.commandLaunchesToken` still matches the new command text (`/agent` + `(?=["'])`), so readiness detection is unaffected.

The blocker is not the flag — it's that it's **unconditional across `.review`**.

### Critical Issues

**🔴 Red 1 — `--trust` is applied to `.review` sessions, reversing this repo's explicit "never trust a review clone" policy.**

`CursorLaunchArgs.trustSuffix` is emitted unconditionally at all three sites (`CursorLaunchArgs.swift:55`, `:96-98`; `CursorAgent.swift:85`; `CursorLauncher.swift:99`), including `.review`. That contradicts a policy this codebase has established, reviewed twice, and documented in four places:

- `SessionService.swift:929-941` — **Codex** trust seeding is explicitly gated:
  ```swift
  // Never trust a `.review` clone: its working tree is `gh repo clone`
  // output checked out at the PR author's head — attacker-controlled.
  if session.kind != .review { CodexTrustSeeder.seedTrust(...) }
  ```
- `SessionService.swift:1245` — the same guard on the Codex **handoff** arm ("Skip `.review` for the same attacker-controlled-clone reason as `launchAgent`").
- `SessionService.swift:2668-2669` — Antigravity review handoff is refused because "Codex/Cursor/OpenCode … support review handoff with their own **strip + trust protections**." On `main`, Cursor's *trust* protection **is** the folder-trust dialog. This PR is what removes it.
- `SessionService.swift:1264-1266` (#829 round 10, Red 1) — "the attacker's `.cursor/hooks.json` fires on the very first handoff launch (hooks have no approval gate) and its `.cursor/mcp.json` is **auto-trusted on the next relaunch**." Trust is named as the load-bearing gate.

Failure scenario, entirely within shipped defaults (`AppState.reviewAutoPermissionMode = true`, `SessionService.swift:975`) — a `.review` Cursor session launches as `agent --trust --force --approve-mcps "$(cat .crow-review-prompt.md)"` inside a clone of an attacker's PR head:

1. Clone-prep strips `.cursor/` (`SessionService.swift:2855`), but that is a **working-tree `removeItem` only** — the index/HEAD entry survives, by design (`:2689-2692`).
2. The review agent runs git in the clone; `/crow-review-pr` Step 1 is literally `gh pr checkout <url>`, and any `git restore .` / `git checkout <ref>` round-trip restores the attacker's `.cursor/` from HEAD.
3. Next relaunch (app restart → `--continue`, `crow recreate-terminal`, `crow retry-readiness`) re-emits `--trust --force --approve-mcps` into a workspace whose `.cursor/hooks.json` (no approval gate) and `.cursor/mcp.json` (`--approve-mcps` auto-trusts it) are back. Unsandboxed execution on the reviewer's machine — the exact chain #829 round 10 blocked.

The folder-trust prompt was the intentional human gate here, and the codebase says so: *"the review clone falls back to Codex's folder-trust prompt (acceptable — review is the human-gated path anyway)"* (`:935-936`).

**Suggested fix** — mirror the existing guard rather than inventing a new mechanism:
- `CursorAgent.autoLaunchCommand` already has `session.kind` in scope; make the seed `session.kind != .review` (e.g. `launchSuffix(autoPermissionMode:seedTrust:)`, or take the kind directly).
- `CursorLauncher.launchCommand` has no session kind — either plumb it through `CodingAgent.launchCommand`, or gate it in `SessionService.handoffAgent`, which already holds `session.kind` and already branches on Cursor at `:1258-1274`.
- Please add `.review`-specific tests asserting the seed is **absent**, paralleling the Codex guard's coverage.

If the intent really is to trust review clones, that's a separate, deliberate decision that should also flip the Codex guard and update the round-10/round-11 rationale — not a side effect of a Cursor flag-wiring PR.

### Security Review

**Strengths:**
- Correctly bounded: `--trust` is workspace trust only, kept out of `autoPermissionSuffix`, and `--yolo` / `--auto-review` / `--sandbox` are still refused — with a test asserting `launchSuffix(true)` never contains `--yolo`.
- Handoff deliberately omits `--force --approve-mcps`; the orthogonality argument is sound and well tested (`CursorLauncherTests.swift:35-45`, including the note on why that site needs its own assertion).
- No new injection surface: `shellQuote` still wraps every interpolated path, the suffix is a static literal, and flags precede the positional prompt on every path.
- The empirical grounding (changelog date + installed-binary `--help`, with the stale param reference called out) is the right standard and is exactly what ADR 0015 asks for.

**Concerns:**
- Red 1 above.
- **🟡 Yellow 1 — the ADR/matrix risk framing compares against the one precedent that supports the change and omits the two that don't.** `CursorLaunchArgs.swift:21-22` and the ADR amendment both cite `CodexTrustSeeder` as the analogue, but Codex's seeder is precisely the one that refuses `.review` — the cited parallel doesn't hold where it matters most. The amendment's "on `.review` clones … the same posture `ClaudeTrustSeeder` + `--permission-mode auto` already ship, consistency rather than a new exposure" is accurate for Claude (`SessionService.swift:927` is ungated) but silently drops both the Codex guard and the `.cursor/`-strip rationale that names trust as the gate. Please state the Codex divergence explicitly and record why review clones are (or aren't) in scope — an ADR that omits the contrary precedent will mislead the next maintainer.

### Code Quality

- **🟡 Yellow 2 — an unverified version floor with total-launch blast radius and no runtime guard.** `CursorLaunchArgs.swift:43-54` documents a ≥ 2026.07.20 minimum and then admits the pre-07.20 behavior is *"reasoned, not probed … no pre-07.20 binary on hand to confirm reject-vs-no-op."* If an older `agent` rejects `--trust` outside `--print` rather than no-oping it, **every** Crow-driven Cursor launch fails — work, job, review, Manager, and handoff — with no fallback and no user-facing signal. The PR's own evidence makes the failure mode plausible: `--output-format` / `--stream-partial-output` carry an "only works with `--print`" note, and if Cursor hard-errors on those, `--trust` may have behaved the same before 07.20. The repo's convention is to gate on an empirical probe; here one leg ships on reasoning. Either probe `agent --version` (once, alongside the existing availability check) and skip the suffix below the floor, or provide an escape hatch so an affected user isn't locked out of Cursor entirely. Deferring is defensible only if the reject-vs-no-op case is actually confirmed first.
- **🟢 Green 1 — the persistence claim is the one assertion in this PR with no evidence behind it.** `CursorLaunchArgs.swift:18-19` ("records the same saved trust decision as accepting it") and `CursorAgent.swift:82-84` ("the first launch already recorded the saved trust decision") both rest on it, while the quoted `--help` text says only "Trust the current workspace without prompting" — which reads per-invocation. Behavior is fine either way since the suffix is unconditional, but it changes what the matrix's new `⚠️ trust seed only` means: a per-launch flag is not equivalent to Codex's persisted `config.toml` entry (a manual `agent` run in a Crow worktree would still prompt). Worth marking as inferred, or probing. Note it also cuts the other way on Red 1 — if it *does* persist, the review clone stays trusted for every later invocation, including manual ones.
- **🟢 Green 2 —** `autoLaunchCommandReviewSessionFirstLaunch` (`CursorAgentTests.swift:72-90`) is the only launch-kind test that asserts nothing about the suffix. The shared `launchArgs` makes that harmless today, but `.review` is exactly the path Red 1 is about, so it should carry explicit assertions once the gate lands.
- Docs are otherwise thorough and internally consistent — the matrix cell, the prose, the dated re-check row, and the gap-audit §3a annotation all move together, and the superseded text is annotated rather than deleted per repo convention. 👍

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 2 Yellow, 2 Green] findings.

The `--trust` research and wiring are solid and I'd like to see this land; it just needs the `.review` guard that the Codex path already carries, and an honest note about the version floor.

dgershman:
Excellent catch on Red 1 — you're right, and thank you for tracing it to the exact `#829` round-10 chain. Fixed in 776b607.

**🔴 Red 1 — `--trust` withheld from `.review` clones.** I was trusting attacker-controlled review checkouts, reversing the twice-reviewed policy. Now mirrors the Codex `session.kind != .review` guard at **all three** seed sites:
- `launchSuffix(seedTrust:autoPermissionMode:)` — `--trust` only when `seedTrust`.
- `CursorAgent.autoLaunchCommand` passes `seedTrust: session.kind != .review` — covers the first `.review` launch **and** every `--continue` relaunch (the core of your chain).
- Handoff: rather than a hacky string-strip or a 5-agent protocol churn, I added a **defaulted** `CodingAgent.launchCommand(…, sessionKind:)` overload (your first suggested option). Only Cursor overrides it (withholds trust for `.review`); Claude/Codex/OpenCode/Antigravity inherit the extension default unchanged. `SessionService.handoffAgent` passes the live `session.kind`.

Auto-permission on `.review` is untouched — only `--trust` is withheld, leaving the folder-trust dialog as the human gate. New tests assert the seed is **absent** on `.review`, including the exact `reviewAutoPermissionMode=true` exploit config (`agent … --force --approve-mcps` present, `--trust` absent), plus first-launch, `--continue` resume, and both handoff paths.

**🟡 Yellow 1 — Codex divergence named.** The ADR amendment and matrix now state it explicitly: Cursor follows the Codex `!= .review` precedent, a deliberate divergence from the **ungated** `ClaudeTrustSeeder`, and record that review clones are out of scope for the seed (they fall back to the folder-trust gate).

**🟡 Yellow 2 — reject-vs-no-op now probed, not reasoned.** \`agent --output-format json --version\` — a \`--print\`-only flag used outside print mode — exits **0** and prints the version with no error. So the Cursor arg-parser silently *ignores* mode-gated flags; a pre-2026.07.20 build recognizing \`--trust\` would no-op it and fall back to the old prompt, **not** reject the launch. That's the confirmation you asked for before deferring the runtime guard. Min floor (≥ 2026.07.20) is recorded in the doc comment, matrix row, and ADR. (Happy to add a version-probe skip if you'd still prefer belt-and-suspenders, but the parser evidence resolves the hard-break concern.)

**🟢 Green 1 —** persistence claim now cited to the changelog's "saved trust decision" wording (and moot for security since `.review` is never seeded).
**🟢 Green 2 —** `.review` launch-kind tests now carry explicit suffix assertions.

`swift build` clean; CrowCursor 65 pass; CrowEngine handoff/launch/strip suites pass. Re-requesting your review.

dhilgaertner (changes requested):
## Code & Security Review

Round 3. The previous round's **🔴 Red 1 is genuinely fixed** — `.review` no longer receives the trust seed on any of the four seed sites, and the gate is threaded correctly rather than patched at one spot. Verified locally:

- `swift build` (root package) → clean
- `swift test --package-path Packages/CrowCursor` → **65 tests pass**
- `swift test --package-path Packages/CrowEngine` → **552 pass** · `Packages/CrowCore` → **518 pass**

Round-2 follow-ups all landed: the ADR now names the `ClaudeTrustSeeder` divergence explicitly (`0015:41-44`), the version floor is recorded in the ADR, the matrix re-check row, and `CursorLaunchArgs.swift:56`, the persistence claim is now sourced to the changelog quote, and `autoLaunchCommandReviewSessionFirstLaunch` carries the explicit negative assertion it was missing. Nice follow-through.

Two Yellows remain, both about the *durability* of the new guard rather than its correctness today.

### Critical Issues

None. No Red findings.

### Security Review

**Strengths:**
- **The guard is threaded, not bolted on.** `seedTrust` reaches all four sites through one predicate (`session.kind != .review`) rather than four independent conditionals, so there's a single place to get it wrong. `launchSuffix(seedTrust:autoPermissionMode:)` keeps trust and approval orthogonal — `launchSuffix(seedTrust: false, autoPermissionMode: true) == " --force --approve-mcps"` is directly asserted (`CursorLaunchArgsTests.swift:64`), which is the right shape: withholding trust must not quietly disarm auto-permission.
- **Handoff is the hard case and it's covered.** A review session created under Claude/Codex and later handed to Cursor never went through `prepareReviewClone`'s Cursor strip; the PR handles both legs — `stripCursorConfigFromReviewClone` at `SessionService.swift:1270` *and* now the withheld seed at `:1233`. Strip + trust-gate together, matching what `shouldRefuseReviewHandoff` claims Cursor provides.
- **Layering is sound.** The strip is a working-tree `removeItem` (index/HEAD survive by design), so a `gh pr checkout` inside the review clone can restore the attacker's `.cursor/`. Withholding trust is what makes that restoration non-load-bearing on relaunch — exactly the chain round 2 identified.
- No new injection surface: `trustSuffix` is a static literal placed between the already-`shellQuote`d binary and the quoted positional prompt. `AgentLaunch.commandLaunchesToken` (`AgentLaunch.swift:67`) still matches — the regex anchors on `/agent` with `(?=\s|$|["'])` lookahead, and ` --trust` satisfies the `\s` branch, so readiness detection is unaffected on every path.

**Concerns:**

- 🟢 **The "folder-trust dialog is review's human gate" guarantee is asserted but never probed under `--force`.** `reviewAutoPermissionMode` is default-on (`SessionService.swift:975`), so the shipped review launch is `agent --force --approve-mcps "$(cat …)"`. The ADR amendment, the matrix prose, and four code comments now state as fact that review "keeps the folder-trust dialog as its human gate." That holds only if `--force` doesn't itself suppress the dialog — and this repo's own gap audit calls `--yolo` an alias for `--force` and describes it as "the unbounded posture." No probe of that combination is recorded anywhere, and I couldn't run one (no `agent` binary on this review host).

  The **code is right either way** — withholding `--trust` is never worse than emitting it — and the assumption predates this PR, so this is not a regression. But ADR 0015 exists specifically to stop unverified capability claims from being enshrined, and this PR is what promotes the assumption into an ADR-level security guarantee. Either probe `agent --force` in a genuinely untrusted folder and record it, or soften to "not auto-trusted; whether `--force` alone still surfaces the dialog is unverified."

### Code Quality

- 🟡 **Yellow 1 — the guard's real dispatch path has no test, and I confirmed the failure is silent.** Every new test exercises a concrete `CursorAgent`. Production doesn't: `AgentRegistry.agent(for:)` returns `(any CodingAgent)?` (`AgentRegistry.swift:61`), so `SessionService.handoffAgent:1229` calls through an **existential**. The doc comment at `CodingAgent.swift:133-137` names this exact fragility — *"A protocol requirement with an extension default, so a call on `any CodingAgent` still dynamically dispatches to an overriding conformer"* — but nothing pins it.

  I verified the counterfactual rather than assuming it. Deleting the four-line requirement declaration at `CodingAgent.swift:94-99` (leaving only the extension default at `:138`) — the kind of cleanup that looks redundant to a future reader:

  - still **compiles clean**;
  - **all 65 CrowCursor tests still pass**, including `launchCommandHandoffWithholdsTrustForReview`;
  - and `(CursorAgent() as any CodingAgent).launchCommand(…, sessionKind: .review)` starts returning `--trust` again — the review-clone hole, silently reopened, with a green suite.

  One test closes it, and it's the only test that would have failed above:

  ```swift
  @Test func existentialDispatchReachesCursorOverride() async throws {
      let erased: any CodingAgent = CursorAgent()   // as SessionService holds it
      let review = try await erased.launchCommand(
          sessionID: UUID(), worktreePath: "/w", prompt: "p", sessionKind: .review)
      #expect(review.contains("--trust") == false)
  }
  ```

- 🟡 **Yellow 2 — `seedTrust` fails open, and the two overloads are one missing argument label apart.** `CursorLauncher.launchCommand` defaults `seedTrust: Bool = true` (`CursorLauncher.swift:96`), and `CursorAgent`'s three-argument `launchCommand` (`CursorAgent.swift:158-169`) forwards without it — so it seeds trust regardless of kind. Confirmed: `(CursorAgent() as any CodingAgent).launchCommand(sessionID:worktreePath:prompt:)` emits `--trust`.

  The overloads at `CodingAgent.swift:81` and `:94` differ only by a trailing `sessionKind:`. Swift resolves a call that omits it to the three-argument form **silently** — no warning, no error — and that form is a public protocol requirement, so it stays reachable indefinitely. It is now the only fail-open path left in the Cursor launch surface, guarding the thing round 2 spent a full round establishing.

  Cost of closing it is near zero: `CursorAgent`'s three-argument override has **no production caller** today (`SessionService.swift:1229` is the sole call site and it passes `sessionKind:`), so it can forward `seedTrust: false` without changing any shipped behavior. Cleaner still, drop the `= true` default on `CursorLauncher.launchCommand` and let the compiler force every call site to state its intent — the same "decide explicitly at each site" discipline the Codex guard already follows by repeating itself at `SessionService.swift:937` and `:1246`.

  Flagging that `launchCommandCarriesTrustSeedNotAutoPermission` (`CursorLauncherTests.swift:35`) currently *asserts* the fail-open default, so this is a deliberate decision to revisit, not an oversight to patch.

- 🟢 **The `.review` carve-out isn't stated where a reader lands first.** `CursorLaunchArgs.swift:13` opens with *"appended to **every** Crow-launched Cursor command"* and only qualifies it 16 lines later at `:29` ("**Withheld from `.review` sessions**"). Same in the matrix: the cell at `agent-harness-matrix.md:39` reads `⚠️ trust seed only (--trust, per-launch)` with the exclusion appearing ~290 lines down in prose. Both are the security-relevant half of the behavior. Suggest `(--trust, per-launch, except .review)` in the cell and folding "except `.review`" into that first sentence — grid-scanners and hover-the-symbol readers are exactly who this needs to reach.

### Docs

Accurate and internally consistent. The ADR amendment is notably better than round 2's: it now states the Codex parallel *and* the `ClaudeTrustSeeder` divergence, and explains why workspace trust is the gate for repo-supplied agent config — that "bounded also means *scoped*" paragraph is the sentence round 1 asked for. Gap-audit §3a is annotated closed rather than rewritten, per repo convention. The dated re-check row carries the floor, the probe, and a forward-drift trigger.

### Summary Table

The security fix is correct and the reasoning behind it is now properly recorded — this is close. Both Yellows are about making the guard survive the next refactor rather than about whether it works today: one existential-typed test, and a fail-closed default. Together they're maybe fifteen lines.

dgershman:
Both Yellows were exactly right — thank you for verifying the counterfactuals rather than asserting them; that's what made them easy to close precisely. Fixed in 79e377d.

**🟡 Yellow 1 — existential-typed test.** You're right that every prior test used a concrete `CursorAgent` while production calls through `any CodingAgent`, so dropping the protocol *requirement* would silently reopen the hole. Added `existentialDispatchReachesCursorOverride`, which erases to `any CodingAgent` (as `handoffAgent` holds it) and pins `.review` → no `--trust`, `.work` → `--trust`. I confirmed your counterfactual too: it's the one test that fails if the requirement declaration is deleted.

**🟡 Yellow 2 — fail-closed.** Removed the `= true` default on `CursorLauncher.launchCommand` so every call site must state intent (the same per-site discipline the Codex guard uses), and made the kindless 3-arg `CursorAgent.launchCommand` forward `seedTrust: false`. As you noted, the 3-arg has no production caller (`handoffAgent:1229` always passes `sessionKind:`), so shipped behavior is unchanged — this just closes the last fail-open path. Added `launchCommandThreeArgFailsClosed`. `launchCommandCarriesTrustSeedNotAutoPermission` now states `seedTrust: true` explicitly rather than leaning on a default.

**🟢 Green 1 — "human gate" claim softened.** I couldn't probe the interactive dialog-under-`--force` on this host either. Since the code is correct either way, I softened the assertion to the verified fact — **Crow never auto-trusts review clones** — and marked the "folder-trust dialog remains the gate" consequence as *unverified under `--force`* (with "withholding `--trust` is never worse than emitting it" as the load-bearing invariant). Applied to the `launchSuffix` doc, both CursorAgent comments, the ADR amendment, and the matrix prose + re-check row. If you'd like, I can file a follow-up to empirically confirm `agent --force` in an untrusted folder once a review host has the CLI.

**🟢 Green 2 — carve-out surfaced up front.** `trustSuffix`'s opening sentence now reads "appended to every Crow-launched Cursor command **except `.review`**," and the matrix cell is now `⚠️ trust seed only (--trust, per-launch, except .review)`.

CrowCursor 67 pass; `swift build` clean. Re-requesting your review.

dhilgaertner (approved):
## Code & Security Review

Round 4. **All four round-3 findings are addressed**, and I verified the two Yellows by counterfactual rather than by reading the diff.

Verified locally on `feature/crow-890-cursor-interactive-trust`:

- `swift build` (root package) → clean
- `swift test --package-path Packages/CrowCursor` → **67 pass** (was 65; +2 new)
- `Packages/CrowEngine` → **552 pass** · `Packages/CrowCore` → **518 pass**
- CI: Build & Test, Catalog & Parity Gates, Desktop (Tauri) Build, Shell Lint — all green; `mergeable: MERGEABLE`

### Critical Issues

None. No Red findings.

### Round-3 follow-ups

**🟡 Yellow 1 (existential dispatch) — closed, and the new test is load-bearing.** I reproduced the prior round's counterfactual: deleted the protocol requirement at `CodingAgent.swift:94-99`, leaving only the extension default at `:138`. It still compiles — and `existentialDispatchReachesCursorOverride` (`CursorAgentTests.swift:144`) now **fails**:

```
✘ Expectation failed: (work.contains("--trust") → false) == true
✘ Test run with 67 tests in 7 suites failed
```

Worth noting *which* assertion caught it: the `.work` one, not the `.review` one. With the requirement gone, the existential falls through to the extension default → the three-argument form → `seedTrust: false`, so review still gets no `--trust`. The two fixes compose: Yellow 2's fail-closed default means even a broken dispatch path degrades toward under-trusting, and Yellow 1's test still catches the breakage on the benign side. That's a better outcome than the single assertion I asked for.

**🟡 Yellow 2 (fail-open `seedTrust`) — closed on both legs.** `CursorLauncher.launchCommand` (`CursorLauncher.swift:97`) drops the `= true` default entirely, so every call site must state intent; `CursorAgent`'s three-argument override (`CursorAgent.swift:177`) forwards `seedTrust: false`. `launchCommandThreeArgFailsClosed` (`CursorAgentTests.swift:160`) pins it, and `launchCommandCarriesTrustSeedNotAutoPermission` — which previously *asserted* the fail-open default — now passes `seedTrust: true` explicitly (`CursorLauncherTests.swift:43`). No shipped behavior changed: `SessionService.swift:1229` remains the sole production caller and it passes `sessionKind:`.

**🟢 Green 1 (unverified `--force`-vs-dialog claim) — softened, not overclaimed.** The ADR now reads "*the intent is they fall back to Cursor's folder-trust dialog … though whether `--force` … still surfaces that dialog is unverified; withholding `--trust` is never worse than emitting it*" (`0015`), and the same hedge appears in `CursorLaunchArgs.swift:107-113`, `CursorAgent.swift:81-84`, the matrix prose, and the re-check row. That is exactly the right resolution for an ADR whose purpose is to stop unverified claims being enshrined.

**🟢 Green 2 (carve-out buried) — fixed at both landing spots.** `CursorLaunchArgs.swift:13-14` now opens "*appended to every Crow-launched Cursor command **except `.review`***", and the matrix cell reads `⚠️ trust seed only (--trust, per-launch, except .review)`.

### Security Review

**Strengths:**

- **The `.review` guard holds on every reachable path, and `.review` can't be laundered into another kind.** I checked the one thing that would defeat it: `Session.kind` is mutated in exactly one place — `migrateLegacyManagerKind` (`SessionService.swift:73`), gated on `AppState.managerSessionID`. There is no status transition, quick action, or RPC that promotes a `.review` session to `.work`, so a review clone cannot become trust-eligible after creation.
- **Defense in depth is layered correctly.** Review clones already get `.cursor/` stripped at creation *and* on handoff (`shouldStripCursorReviewCloneOnHandoff`, `SessionService.swift:1270`). The strip is a working-tree removal that a `gh pr checkout` inside the clone can undo; withholding trust is what makes that restoration non-load-bearing. Both legs now present.
- **Trust and approval stay orthogonal.** `launchSuffix(seedTrust: false, autoPermissionMode: true) == " --force --approve-mcps"` is directly asserted (`CursorLaunchArgsTests.swift:47`) — withholding trust does not quietly disarm auto-permission on review, which would have been a silent behavior change.
- **No new injection surface, and readiness detection is unaffected.** `trustSuffix` is a static literal placed between the already-`shellQuote`d binary and the `shellQuote`d prompt path. `AgentLaunch.commandLaunchesToken` (`AgentLaunch.swift:67`) anchors on `/agent` with a `(?=\s|$|["'])` lookahead — the closing `'` still satisfies it on every path, including the Manager's newline-less form. I also swept for command-string introspection that could break on a new flag (`EngineHelpers.swift:60`, `EngineRouter.swift:815` check `--rc`/`--remote-control` only) — nothing matches on shape.
- **Bounded as claimed.** `--yolo` absence is asserted on the composed suffix (`CursorLaunchArgsTests.swift:38`), not just on `autoPermissionSuffix`.

**Concerns:** none blocking.

### Code Quality

- 🟢 **`SessionService.handoffAgent`'s `sessionKind: session.kind` argument (`SessionService.swift:1233`) is the one link with no test.** Every other hop in the chain is pinned. There's no harness for `handoffAgent` in `CrowEngineTests` today, and unlike Yellow 1's case the only way to break this is to deliberately hardcode a wrong kind — not something a plausible cleanup does. Noting it for completeness, not asking for it.
- 🟢 **The two-overload protocol shape is the residual API cost.** `CodingAgent` now carries `launchCommand(…)` and `launchCommand(…sessionKind:)` one argument label apart, and Swift resolves an omitted label to the three-argument form silently. Fail-closed inverts the failure mode from "silently over-trusts a review clone" to "silently under-trusts and stalls on the dialog" — the right direction, and the same as pre-PR behavior. The alternative — adding `sessionKind` to the single existing requirement and having the four kind-agnostic conformers ignore it — would remove the ambiguity outright and make every call site compiler-checked, at the cost of touching four files. Worth considering if a second kind-dependent launch ever appears; not worth churn now.
- 🟢 **The `.review` rationale is now restated in five places** (`CursorLaunchArgs.swift` ×2, `CursorAgent.swift` ×2, `CursorLauncher.swift`), plus the ADR and matrix. This matches the existing Codex-guard precedent (`SessionService.swift:929` / `:1246`), so it's consistent with repo convention — but it is the surface most likely to drift out of sync with the ADR on the next change.
- 🟢 **PR body says "60 tests pass"; actual is 67.** Stale from an earlier round.

### Docs

Accurate and internally consistent. The version floor appears in all three required places (ADR, matrix re-check row, `CursorLaunchArgs.swift`) with the parser probe recorded as the basis for "older builds no-op rather than reject," and a forward-drift trigger ("re-probe if `--help` ever restores the headless-only qualifier"). Gap-audit §3a is annotated closed rather than rewritten, per repo convention.

One transparency note on my own verification: the `agent` binary is not installed on this review host, so the `--help` evidence underpinning the interactive-`--trust` claim is not independently reproducible from here. It is dated and version-pinned per ADR 0015's own convention, which is the mechanism this repo relies on for exactly that case.

### Summary Table

Four rounds in, the guard is threaded through one predicate, pinned at every hop it can realistically regress at, and honest about what it hasn't probed. The two Yellows didn't just get patched — they compose, so a future refactor that breaks dispatch degrades toward under-trusting instead of reopening the review-clone hole. Ship it.

PR #905, opened 2026-07-28
hook-event: fire-and-forget the RPC so a busy daemon can't stall the hook (CROW-903)

### LINKED ISSUE #903: crow hook-event blocks on the daemon RPC reply it never uses — busy MainActor stalls every hook to the 5s hook timeout (should be fire-and-forget)
## Summary

`crow hook-event` performs a **blocking** JSON-RPC round-trip and waits for the daemon's reply — but the reply is never used for anything except printing an error. Because the daemon serializes every RPC on the `@MainActor` (`EngineRouter`), any busy window (board poll, git/worktree op, whole-`store.json` write, or a burst of concurrent hook-events) stalls the hook-event RPC until Claude Code's 5s hook timeout kills it. The hook is meant to be fire-and-forget (the code comment says so) — the wait-for-response is accidental and should be removed.

## Evidence (measured from Claude Code transcripts, ~50 sessions / 2 days on the Manager host)

PreToolUse:Bash hook durations are **bimodal** — fast or timeout, nothing between:
- 0–200ms: 232 runs (88%)
- ~5s timeout: 31 runs (11%)
- 200ms–4.9s: 0 runs

A slow-but-completing op would smear across 200ms–3s. A clean fast-or-5s split is the signature of a **block**, not slow code.

The timeouts **cluster in busy windows** and concentrate in a few sessions (not spread across the 263 calls):
- 2026-07-27 17:55–17:56 — ~16 timeouts in a ~90s burst
- 2026-07-28 19:14–19:50 — steady ~1/20–60s for 36 min, almost all from one session (24 of ~50 timeouts came from a single session; two others contributed 12 and 9)

Other per-call/per-prompt hooks show the same tail: UserPromptSubmit (32 runs, 6 timeouts), PreToolUse:Edit (68, 6), PreToolUse:Read (27, 3), Stop (29, 3). All hooks are the same `crow hook-event … --event X` call, so they share the one bottleneck.

Cost: each timeout is a ~5s stall before the tool runs (hook cancelled, Claude proceeds). ~49 stalls ≈ ~4 min of dead wait, concentrated in the busy sessions. Annoying, not corrupting — but it makes heavy Manager/batch periods drag.

## Root cause (pinned to source)

`Packages/CrowCLI/Sources/CrowCLILib/Commands/HookEventCommand.swift:61` — `forwardHookEvent`:

```swift
let result = try rpc("hook-event", params: params)   // blocks waiting for the reply
if result["error"] != nil { printJSON(result) }      // reply only used to print errors
```

Daemon handler `Packages/CrowEngine/Sources/CrowEngine/EngineRouter.swift:1208` runs on the `@MainActor` router (line 15; nearly every handler does `await MainActor.run { … }`) and does a whole-store `mutate`. So hook-event must acquire the MainActor, contending with board polls, git ops, and every other terminal's hook-events (each rewriting the full store — the single-writer design). During busy windows it queues; the CLI blocks on a reply it doesn't use; the 5s hook timeout fires. The doc comment above the code already states "Hooks are fire-and-forget" — the blocking wait contradicts the intended design.

## Proposed fix

1. **Primary (client-side, ~one function):** make `hook-event` a genuine fire-and-forget notification — send it as a JSON-RPC notification (no `id`) or write the payload to the socket and return without awaiting the response. Keep the existing `SocketError.connectionFailed` silent no-op (missing daemon is expected). The daemon still processes the event when the MainActor frees; Claude Code never waits. This removes the stall outright and matches the code's stated intent. Verify no hook consumes the hook-event *output* to gate a tool (PreToolUse here is a pure notification — nothing does).
2. **Structural (daemon-side, follow-up):** ensure board polling / long git ops don't hold the MainActor across their network/subprocess `await` — hop to MainActor only for the final state mutation, so hook-events aren't starved during polls. Consider not doing a synchronous whole-store write on every hook-event notification (coalesce), which would also cut the per-event MainActor pressure.

## Non-fix

Raising the 5s hook `timeout` in the scaffolded settings is the wrong lever — it makes each stall *longer* before giving up, not rarer. Don't.

## Related

- #874 (CLOSED) — "Blocking NSLog on the MainActor can wedge the daemon's RPC surface." Same MainActor-serialization failure class, but scoped to an NSLog on `reload-tmux-config`'s success path. Its fix does NOT cover this: hook-event still blocks whenever the MainActor is *legitimately* busy (polls, store writes, bursts) — which is what the 07-27/07-28 windows above show.
- #728 (CLOSED), #304/#404/#250 (CLOSED) — adjacent MainActor-contention symptoms (store-write race; UI freeze under concurrent reviews).

dgershman:
Closes #903

## Problem

`crow hook-event` did a **blocking** JSON-RPC round-trip and waited for the daemon's reply — a reply used only to print RPC errors. Because the daemon serializes every request on its `@MainActor` (`EngineRouter`), any busy window (board poll, git/worktree op, whole-`store.json` write, or a burst of concurrent hook-events) queued the hook-event RPC until Claude Code's 5s hook timeout killed it.

The measured PreToolUse:Bash durations were **bimodal** — fast or ~5s, nothing between (0–200ms: 88%, ~5s timeout: 11%, 200ms–4.9s: **0**). A clean fast-or-5s split is the signature of a **block**, not slow code; the timeouts clustered in busy windows and concentrated in a few sessions. The doc comment above the code already stated "Hooks are fire-and-forget" — the blocking wait contradicted the intended design.

## Fix

Make the hook a genuine fire-and-forget notification. `SocketClient.post` connects, writes the request, and returns **without reading a response**. The daemon still processes the event once its MainActor frees; the agent never waits, so the stall is gone.

- **CrowIPC** — extract the connect+write path into a private `connectAndWrite`, shared by `send` (which then reads a reply) and the new `post` (fire-and-forget, closes immediately). Data already accepted into the Unix stream socket is delivered before the FIN, so the server still receives the full request; if it later replies to the closed connection its `SIGPIPE`-ignoring write path drops the response harmlessly.
- **CrowCLI** — add `rpcNotify`; `forwardHookEvent` now posts instead of `rpc`, dropping the error-print that the reply was only ever used for. Missing daemon (`SocketError.connectionFailed`) stays a silent no-op; genuine socket errors (create/write) still propagate.

Scope is the ticket's **primary (client-side)** fix only. The structural daemon-side follow-up (don't hold the MainActor across polls/git `await`; coalesce per-event store writes) is intentionally left out of this PR.

## Non-fix

Raising the 5s hook timeout was explicitly rejected in the ticket — it makes each stall *longer*, not rarer.

## Test plan

- `swift build` — clean.
- `swift test --package-path Packages/CrowIPC` — 39/39 pass, including two new tests:
  - `postReturnsWithoutWaitingForHandler` — against a handler that sleeps 60s, `post` returns in ~50ms **and** the event is still delivered (the daemon processes it). `send` would have stalled on the reply.
  - `postToNonexistentSocketThrows` — an absent socket still surfaces `connectionFailed`, which `forwardHookEvent` catches to stay a silent no-op.
- `swift test --package-path Packages/CrowCLI` — 341/341 pass, including `forwardHookEventSilentWhenAppNotRunning` (#227 regression) and the CLI/RPC control-plane parity gates.

### DIFF
--- a/.github/workflows/cache-warm.yml
+++ b/.github/workflows/cache-warm.yml
@@ -38,7 +38,7 @@ jobs:
         run: |
           bash scripts/generate-build-info.sh
           LINUX_PACKAGES=(CrowCore CrowIPC CrowClaude CrowCodex CrowCursor \
-                          CrowAntigravity CrowGit CrowOpenCode CrowPersistence CrowProvider CrowCLI)
+                          CrowAntigravity CrowGit CrowOpenCode CrowGrok CrowPersistence CrowProvider CrowCLI)
           for pkg in "${LINUX_PACKAGES[@]}"; do
             echo "==> Building $pkg on Linux..."
             swift build --package-path "Packages/$pkg"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Build & test Linux packages
         run: |
           LINUX_PACKAGES=(CrowCore CrowIPC CrowClaude CrowCodex CrowCursor \
-                          CrowAntigravity CrowGit CrowOpenCode CrowPersistence CrowProvider CrowCLI)
+                          CrowAntigravity CrowGit CrowOpenCode CrowGrok CrowPersistence CrowProvider CrowCLI)
           for pkg in "${LINUX_PACKAGES[@]}"; do
             echo "==> Building & testing $pkg on Linux..."
             swift build --package-path "Packages/$pkg"
--- a/Packages/CrowAntigravity/Sources/CrowAntigravity/AntigravityHookConfigWriter.swift
+++ b/Packages/CrowAntigravity/Sources/CrowAntigravity/AntigravityHookConfigWriter.swift
@@ -36,14 +36,16 @@ import CrowCore
 /// `PostToolUse` (whose reply is a harmless `{}`).
 ///
 /// **Every command emits its own stdout verdict.** Antigravity reads the hook's
-/// stdout as JSON. `crow hook-event` is observational and may print a JSON-RPC
-/// error on failure, so each command **suppresses** hook-event's output and then
-/// `printf '{}'`s the empty-object verdict itself — the documented no-op reply
-/// for `PostToolUse`/`PreInvocation`/`PostInvocation`, and a non-`continue` reply
-/// for `Stop` (i.e. "allow the agent to stop"). Using `;` (not `&&`) plus the
-/// trailing `printf` guarantees the verdict is emitted and the hook exits 0 even
-/// when hook-event fails — so a Crow hook can never block a tool or loop the
-/// agent.
+/// stdout as JSON. `crow hook-event` is observational and prints nothing to
+/// stdout (fire-and-forget since #903 — it no longer even prints RPC errors),
+/// so each command still `printf '{}'`s the empty-object verdict itself — the
+/// documented no-op reply for `PostToolUse`/`PreInvocation`/`PostInvocation`,
+/// and a non-`continue` reply for `Stop` (i.e. "allow the agent to stop"). The
+/// leading output-suppression is now belt-and-suspenders, but kept so a future
+/// hook-event that does print can't leak into the verdict. Using `;` (not `&&`)
+/// plus the trailing `printf` guarantees the verdict is emitted and the hook
+/// exits 0 even when hook-event fails — so a Crow hook can never block a tool or
+/// loop the agent.
 ///
 /// **`.agents/hooks.json` is a shared workspace file** (like Cursor's
 /// `.cursor/hooks.json`), so this writer keeps Cursor's protections: a
--- a/Packages/CrowCLI/Sources/CrowCLILib/Commands/BoardCommands.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Commands/BoardCommands.swift
@@ -131,7 +131,7 @@ public struct CreateManager: ParsableCommand {
         abstract: "Create an additional Manager session"
     )
 
-    @Option(name: .long, help: "Coding agent kind (claude-code, cursor, codex, opencode); default agent when omitted")
+    @Option(name: .long, help: "Coding agent kind (claude-code, cursor, codex, opencode, antigravity, grok); default agent when omitted")
     var agent: String?
 
     public init() {}
@@ -141,7 +141,7 @@ public struct CreateManager: ParsableCommand {
         // downstream packages extend, so a hardcoded allowlist would go stale
         // (same reasoning as `crow handoff-agent`).
         if let agent, agent.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
-            throw ValidationError("--agent must not be blank (claude-code, cursor, codex, or opencode).")
+            throw ValidationError("--agent must not be blank (e.g. claude-code, cursor, codex, opencode, antigravity, grok).")
         }
     }
 
--- a/Packages/CrowCLI/Sources/CrowCLILib/Commands/HookEventCommand.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Commands/HookEventCommand.swift
@@ -50,20 +50,24 @@ public struct HookEventCmd: ParsableCommand {
     }
 }
 
-/// Forward a hook-event RPC over the Unix socket.
+/// Forward a hook-event RPC over the Unix socket, fire-and-forget.
 ///
-/// Silently no-ops when the Crow app is not running (socket connection
-/// refused or socket file absent). Hooks are fire-and-forget — a missing
-/// listener is an expected state, not an error, so we must not exit
-/// non-zero or write to stderr (it pollutes the agent's hook output).
-/// Other socket errors (timeout, write/read failures) and JSON-RPC
-/// errors still propagate so genuine misbehavior is visible.
+/// Hooks are fire-and-forget: we write the event and return without waiting for
+/// the daemon's reply. That reply was only ever used to print RPC errors, but
+/// because the daemon serializes every request on its `@MainActor`, waiting for
+/// it stalled the hook whenever the daemon was busy (board poll, git op,
+/// whole-store write, a burst of hook-events) — right up to the agent's hook
+/// timeout (#903). Dropping the wait removes that stall; the daemon still
+/// processes the event once its MainActor frees.
+///
+/// Silently no-ops when the Crow app is not running (socket connection refused
+/// or socket file absent): a missing listener is an expected state, not an
+/// error, so we must not exit non-zero or write to stderr (it pollutes the
+/// agent's hook output). Other socket errors (create/write failures) still
+/// propagate so genuine misbehavior is visible.
 func forwardHookEvent(params: [String: JSONValue]) throws {
     do {
-        let result = try rpc("hook-event", params: params)
-        if result["error"] != nil {
-            printJSON(result)
-        }
+        try rpcNotify("hook-event", params: params)
     } catch SocketError.connectionFailed {
         return
     }
--- a/Packages/CrowCLI/Sources/CrowCLILib/Commands/HookEventCommand.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Commands/HookEventCommand.swift
@@ -52,13 +52,21 @@ public struct HookEventCmd: ParsableCommand {
 
 /// Forward a hook-event RPC over the Unix socket, fire-and-forget.
 ///
-/// Hooks are fire-and-forget: we write the event and return without waiting for
-/// the daemon's reply. That reply was only ever used to print RPC errors, but
-/// because the daemon serializes every request on its `@MainActor`, waiting for
-/// it stalled the hook whenever the daemon was busy (board poll, git op,
-/// whole-store write, a burst of hook-events) — right up to the agent's hook
-/// timeout (#903). Dropping the wait removes that stall; the daemon still
-/// processes the event once its MainActor frees.
+/// Hooks are fire-and-forget: we write the event and return without reading the
+/// daemon's reply. Because the daemon serializes every request on its
+/// `@MainActor`, the old blocking round-trip stalled the hook whenever the
+/// daemon was busy (board poll, git op, whole-store write, a burst of
+/// hook-events) — right up to the agent's hook timeout (#903). Dropping the read
+/// removes that stall; the daemon still processes the event once its MainActor
+/// frees.
+///
+/// The only behavior this drops is error *surfacing*. The old call went through
+/// `rpc()`, which throws on a daemon RPC error, so a failed hook-event exited
+/// non-zero with a stderr message; the `printJSON(result)` branch it also had
+/// was dead, since `rpc()` returns only the success `result`, which never
+/// carries an `error` key. Losing that surfacing is a win here: it silences the
+/// `session_id required or resolvable from payload cwd` noise that global-hook
+/// agents hit (#897) and keeps any error text off the hook's output.
 ///
 /// Silently no-ops when the Crow app is not running (socket connection refused
 /// or socket file absent): a missing listener is an expected state, not an
--- a/Packages/CrowCLI/Sources/CrowCLILib/Commands/SessionCommands.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Commands/SessionCommands.swift
@@ -186,7 +186,7 @@ public struct HandoffAgent: ParsableCommand {
         abstract: "Switch a session to a different coding agent (e.g. when credits run out)"
     )
     @Option(name: .long, help: "Session UUID") var session: String
-    @Option(name: .long, help: "Target agent kind (claude-code, cursor, codex, opencode, antigravity)")
+    @Option(name: .long, help: "Target agent kind (claude-code, cursor, codex, opencode, grok, antigravity)")
     var agent: String
     @Option(name: .long, help: "Optional note for the incoming agent about where to resume")
     var note: String?
@@ -196,7 +196,7 @@ public struct HandoffAgent: ParsableCommand {
     public func validate() throws {
         try validateUUID(session, label: "session UUID")
         guard !agent.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty else {
-            throw ValidationError("agent must be non-empty (claude-code, cursor, codex, opencode, or antigravity)")
+            throw ValidationError("agent must be non-empty (claude-code, cursor, codex, opencode, grok, or antigravity)")
         }
     }
 
--- a/Packages/CrowCLI/Sources/CrowCLILib/Helpers.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Helpers.swift
@@ -19,6 +19,19 @@ public func rpc(_ method: String, params: [String: JSONValue] = [:], timeoutSeco
     return response.result ?? [:]
 }
 
+/// Fire-and-forget a JSON-RPC request to the running Crow app via Unix socket.
+///
+/// Writes the request and returns immediately without reading a response. Used
+/// by hooks (`crow hook-event`): the app still processes the event once its
+/// serialized MainActor frees, but the agent never waits, so a busy daemon
+/// can't stall the hook up to its timeout (#903).
+///
+/// - Throws: `SocketError.connectionFailed` when the app isn't running (callers
+///   treat this as an expected no-op); other socket errors propagate.
+public func rpcNotify(_ method: String, params: [String: JSONValue] = [:]) throws {
+    try SocketClient().post(method: method, params: params)
+}
+
 /// Pretty-print a JSON dictionary to stdout with sorted keys.
 public func printJSON(_ dict: [String: JSONValue]) {
     let encoder = JSONEncoder()
--- a/Packages/CrowCLI/Sources/CrowCLILib/Helpers.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Helpers.swift
@@ -21,10 +21,12 @@ public func rpc(_ method: String, params: [String: JSONValue] = [:], timeoutSeco
 
 /// Fire-and-forget a JSON-RPC request to the running Crow app via Unix socket.
 ///
-/// Writes the request and returns immediately without reading a response. Used
-/// by hooks (`crow hook-event`): the app still processes the event once its
-/// serialized MainActor frees, but the agent never waits, so a busy daemon
-/// can't stall the hook up to its timeout (#903).
+/// Writes the request and returns without reading a response, so the agent
+/// doesn't wait on the daemon's reply: the app still processes the event once
+/// its serialized MainActor frees, but a busy daemon can't stall the hook up to
+/// its timeout (#903). See `SocketClient.post` for the one residual wait — a
+/// payload past the socket send buffer can still block in `write()` until the
+/// daemon drains it.
 ///
 /// - Throws: `SocketError.connectionFailed` when the app isn't running (callers
 ///   treat this as an expected no-op); other socket errors propagate.
--- a/Packages/CrowCLI/Tests/CrowCLITests/HookEventTests.swift
+++ b/Packages/CrowCLI/Tests/CrowCLITests/HookEventTests.swift
@@ -3,6 +3,16 @@ import Foundation
 import CrowIPC
 @testable import CrowCLILib
 
+/// Sync-settable, async-readable latch. `DispatchSemaphore.wait()` is
+/// unavailable in async contexts, so a parked hook handler polls this instead
+/// of blocking — the test releases it synchronously from a `defer`.
+private final class Latch: @unchecked Sendable {
+    private let lock = NSLock()
+    private var open = false
+    func release() { lock.lock(); open = true; lock.unlock() }
+    var isOpen: Bool { lock.lock(); defer { lock.unlock() }; return open }
+}
+
 // MARK: - Hook Event Payload Parsing
 
 @Test func parseValidJSONPayload() throws {
@@ -51,47 +61,69 @@ import CrowIPC
 
 // MARK: - Connection Refused Handling
 
-/// Regression test for #227: when the Crow app is not running, the hook-event
-/// command must silently no-op instead of exiting non-zero. Otherwise Claude
-/// Code surfaces "Stop hook error: …" on every session exit.
-@Test func forwardHookEventSilentWhenAppNotRunning() throws {
-    let nonExistent = NSTemporaryDirectory() + "crow-test-\(UUID().uuidString).sock"
-    setenv("CROW_SOCKET", nonExistent, 1)
-    defer { unsetenv("CROW_SOCKET") }
+/// Both tests mutate the process-global `CROW_SOCKET` env var, which
+/// `SocketClient` reads. Swift Testing parallelizes by default and `make test`
+/// doesn't pass `--no-parallel`, so without `.serialized` these two would race
+/// each other's `setenv`/`unsetenv` — both a data race against another thread's
+/// `environ` read, and a correctness hole: if the absent-socket test's `setenv`
+/// landed between this suite's other test's `setenv` and its `forwardHookEvent`,
+/// the fire-and-forget gate would connect to the nonexistent path, no-op, and
+/// pass vacuously. Serializing the suite closes both.
+@Suite(.serialized)
+struct HookEventForwarding {
 
-    try forwardHookEvent(params: [
-        "session_id": .string(UUID().uuidString),
-        "event_name": .string("Stop"),
-        "payload": .object([:]),
-    ])
-}
+    /// Regression test for #227: when the Crow app is not running, the
+    /// hook-event command must silently no-op instead of exiting non-zero.
+    /// Otherwise Claude Code surfaces "Stop hook error: …" on every session exit.
+    @Test func forwardHookEventSilentWhenAppNotRunning() throws {
+        let nonExistent = NSTemporaryDirectory() + "crow-test-\(UUID().uuidString).sock"
+        setenv("CROW_SOCKET", nonExistent, 1)
+        defer { unsetenv("CROW_SOCKET") }
 
-/// #903: `forwardHookEvent` must be wired to the fire-and-forget path, not a
-/// blocking round-trip. Against a live daemon whose handler never replies within
-/// the test, `rpc` would hang on the read until the socket timeout; `rpcNotify`
-/// returns as soon as the request is written. Reverting the call site back to
-/// `rpc` fails here — which the absent-socket test above cannot catch, since it
-/// never reaches the read.
-@Test func forwardHookEventDoesNotBlockOnSlowDaemon() throws {
-    let path = NSTemporaryDirectory() + "crow-test-\(UUID().uuidString).sock"
-    let router = CommandRouter(handlers: [
-        "hook-event": { @Sendable _ in
-            try await Task.sleep(nanoseconds: 60_000_000_000) // 60s — never replies in test
-            return [:]
-        },
-    ])
-    let server = SocketServer(socketPath: path, router: router)
-    try server.start()
-    defer { server.stop(); try? FileManager.default.removeItem(atPath: path) }
-    Thread.sleep(forTimeInterval: 0.05) // let the accept loop come up
+        try forwardHookEvent(params: [
+            "session_id": .string(UUID().uuidString),
+            "event_name": .string("Stop"),
+            "payload": .object([:]),
+        ])
+    }
 
-    setenv("CROW_SOCKET", path, 1)
-    defer { unsetenv("CROW_SOCKET") }
+    /// #903: `forwardHookEvent` must be wired to the fire-and-forget path, not a
+    /// blocking round-trip. The handler signals receipt and then parks until the
+    /// test releases it, so a reintroduced `rpc` read would block here while
+    /// `rpcNotify` returns as soon as the request is written. Asserts both that
+    /// `forwardHookEvent` returns before the handler replies *and* that the
+    /// daemon still received the event — so it can't pass by silently no-op'ing
+    /// (which is why it lives in the serialized suite above).
+    @Test func forwardHookEventDoesNotBlockOnSlowDaemon() throws {
+        let received = DispatchSemaphore(value: 0)
+        let latch = Latch()
+        let path = NSTemporaryDirectory() + "crow-test-\(UUID().uuidString).sock"
+        let router = CommandRouter(handlers: [
+            "hook-event": { @Sendable _ in
+                received.signal()
+                // Park (without replying) until the test releases us, so a
+                // reintroduced `rpc` read would stall here. Polling a latch, not
+                // a 60s sleep — the handler returns within a poll of release, so
+                // it doesn't pin a socket worker for the rest of the run.
+                while !latch.isOpen { try await Task.sleep(nanoseconds: 20_000_000) }
+                return [:]
+            },
+        ])
+        let server = SocketServer(socketPath: path, router: router)
+        try server.start()
+        // Release the parked handler first so `server.stop()` isn't waiting on it.
+        defer { latch.release(); server.stop(); try? FileManager.default.removeItem(atPath: path) }
+        Thread.sleep(forTimeInterval: 0.05) // let the accept loop come up
 
-    let start = Date()
-    try forwardHookEvent(params: [
-        "event_name": .string("PreToolUse"),
-        "payload": .object([:]),
-    ])
-    #expect(Date().timeIntervalSince(start) < 5)
+        setenv("CROW_SOCKET", path, 1)
+        defer { unsetenv("CROW_SOCKET") }
+
+        let start = Date()
+        try forwardHookEvent(params: [
+            "event_name": .string("PreToolUse"),
+            "payload": .object([:]),
+        ])
+        #expect(Date().timeIntervalSince(start) < 5) // returned before the handler's (parked) reply
+        #expect(received.wait(timeout: .now() + 2) == .success) // and the daemon still got it
+    }
 }
--- a/Packages/CrowCLI/Tests/CrowCLITests/HookEventTests.swift
+++ b/Packages/CrowCLI/Tests/CrowCLITests/HookEventTests.swift
@@ -65,3 +65,33 @@ import CrowIPC
         "payload": .object([:]),
     ])
 }
+
+/// #903: `forwardHookEvent` must be wired to the fire-and-forget path, not a
+/// blocking round-trip. Against a live daemon whose handler never replies within
+/// the test, `rpc` would hang on the read until the socket timeout; `rpcNotify`
+/// returns as soon as the request is written. Reverting the call site back to
+/// `rpc` fails here — which the absent-socket test above cannot catch, since it
+/// never reaches the read.
+@Test func forwardHookEventDoesNotBlockOnSlowDaemon() throws {
+    let path = NSTemporaryDirectory() + "crow-test-\(UUID().uuidString).sock"
+    let router = CommandRouter(handlers: [
+        "hook-event": { @Sendable _ in
+            try await Task.sleep(nanoseconds: 60_000_000_000) // 60s — never replies in test
+            return [:]
+        },
+    ])
+    let server = SocketServer(socketPath: path, router: router)
+    try server.start()
+    defer { server.stop(); try? FileManager.default.removeItem(atPath: path) }
+    Thread.sleep(forTimeInterval: 0.05) // let the accept loop come up
+
+    setenv("CROW_SOCKET", path, 1)
+    defer { unsetenv("CROW_SOCKET") }
+
+    let start = Date()
+    try forwardHookEvent(params: [
+        "event_name": .string("PreToolUse"),
+        "payload": .object([:]),
+    ])
+    #expect(Date().timeIntervalSince(start) < 5)
+}
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
@@ -17,8 +17,21 @@ public struct ClaudeHookConfigWriter: HookConfigWriter {
     ]
 
     /// Post-execution events that can safely run async (fire-and-forget).
-    /// PreToolUse is intentionally NOT async — it must arrive before
-    /// PermissionRequest so the state machine ordering is reliable.
+    /// PreToolUse stays non-async so the agent waits for its hook process to
+    /// exit before firing the next hook — which keeps PreToolUse *accepted* by
+    /// the daemon ahead of the following PermissionRequest.
+    ///
+    /// Note (#903): pre-fire-and-forget, non-async also gave end-to-end *apply*
+    /// ordering — the hook process blocked on the daemon's reply, written only
+    /// after the transition was applied, so event N was applied before N+1 was
+    /// sent. `crow hook-event` no longer waits for that reply, and the daemon
+    /// applies hook-events on independently-scheduled `MainActor` tasks
+    /// (`SocketServer` fans each connection out concurrently), so accept order
+    /// no longer implies apply order. Keeping PreToolUse non-async still narrows
+    /// the reorder window to that MainActor scheduling race (vs. also racing the
+    /// writes if it were async), and any inversion self-heals on the next event;
+    /// restoring a hard guarantee needs per-session server-side sequencing —
+    /// the daemon-side follow-up scoped out of #903.
     private static let asyncEvents: Set<String> = [
         "PostToolUse", "PostToolUseFailure",
     ]
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
@@ -32,17 +32,31 @@ public struct ClaudeHookConfigWriter: HookConfigWriter {
     /// writes if it were async).
     ///
     /// Known limitation until the fix lands: a PreToolUse/PermissionRequest
-    /// inversion can transiently show the wrong card state — e.g. PreToolUse
-    /// applied after PermissionRequest wipes the permission badge and reads
-    /// `.working` while the agent is parked at the prompt, or (the reverse)
-    /// a PostToolUse racing a later PreToolUse strands the card at `.waiting`.
+    /// inversion can show the wrong card state. Two directions, and they are not
+    /// equally benign:
+    ///   - PostToolUse racing a *later* PreToolUse can strand the card at
+    ///     `.waiting` while the agent is actively working — this self-corrects,
+    ///     because the agent keeps emitting events (the next PreToolUse/Stop
+    ///     fixes it).
+    ///   - PreToolUse applied *after* PermissionRequest blanket-clears the
+    ///     permission badge and sets `.working` — and this does NOT self-correct.
+    ///     The agent is parked at the prompt, so no further hook fires until the
+    ///     user answers. The only backstop is a separate `Notification` carrying
+    ///     `notification_type == "permission_prompt"`, which re-raises the badge
+    ///     (`ClaudeHookSignalSource`); that field is read verbatim from the
+    ///     payload, never synthesized, so on any Claude build that doesn't emit
+    ///     it the card can read `.working` for the whole prompt and Crow never
+    ///     surfaces "waiting". This is Claude-specific — Cursor keeps a
+    ///     `PermissionRequest` case for parity but doesn't emit it, so the
+    ///     permission-badge case can't arise there.
+    ///
     /// A pure client-side signal source can't reliably tell "this PreToolUse is
     /// the one the pending prompt is for" (the event carries no ordering key and
     /// `PermissionRequest` carries no verified tool identity), so the correct fix
     /// is per-session server-side sequencing of hook-event application — the
     /// daemon-side follow-up scoped out of #903 (see docs/agent-harness-matrix.md
-    /// "Hook async delivery"). The window is a few milliseconds and self-corrects
-    /// on the next state-changing event.
+    /// "Hook async delivery"). The reorder *window* is a few milliseconds; the
+    /// resulting wrong state persists as described above.
     private static let asyncEvents: Set<String> = [
         "PostToolUse", "PostToolUseFailure",
     ]
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
@@ -29,13 +29,20 @@ public struct ClaudeHookConfigWriter: HookConfigWriter {
     /// (`SocketServer` fans each connection out concurrently), so accept order
     /// no longer implies apply order. Keeping PreToolUse non-async still narrows
     /// the reorder window to that MainActor scheduling race (vs. also racing the
-    /// writes if it were async). The one inversion that would NOT self-heal —
-    /// PreToolUse applied after PermissionRequest, wiping the permission badge
-    /// while the agent is parked at the prompt — is guarded directly in
-    /// `ClaudeHookSignalSource` (PreToolUse leaves a pending `permission_prompt`
-    /// intact); other inversions self-heal on the next event. A hard transport
-    /// guarantee (per-session server-side sequencing) is the daemon-side
-    /// follow-up scoped out of #903.
+    /// writes if it were async).
+    ///
+    /// Known limitation until the fix lands: a PreToolUse/PermissionRequest
+    /// inversion can transiently show the wrong card state — e.g. PreToolUse
+    /// applied after PermissionRequest wipes the permission badge and reads
+    /// `.working` while the agent is parked at the prompt, or (the reverse)
+    /// a PostToolUse racing a later PreToolUse strands the card at `.waiting`.
+    /// A pure client-side signal source can't reliably tell "this PreToolUse is
+    /// the one the pending prompt is for" (the event carries no ordering key and
+    /// `PermissionRequest` carries no verified tool identity), so the correct fix
+    /// is per-session server-side sequencing of hook-event application — the
+    /// daemon-side follow-up scoped out of #903 (see docs/agent-harness-matrix.md
+    /// "Hook async delivery"). The window is a few milliseconds and self-corrects
+    /// on the next state-changing event.
     private static let asyncEvents: Set<String> = [
         "PostToolUse", "PostToolUseFailure",
     ]
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
@@ -29,9 +29,13 @@ public struct ClaudeHookConfigWriter: HookConfigWriter {
     /// (`SocketServer` fans each connection out concurrently), so accept order
     /// no longer implies apply order. Keeping PreToolUse non-async still narrows
     /// the reorder window to that MainActor scheduling race (vs. also racing the
-    /// writes if it were async), and any inversion self-heals on the next event;
-    /// restoring a hard guarantee needs per-session server-side sequencing —
-    /// the daemon-side follow-up scoped out of #903.
+    /// writes if it were async). The one inversion that would NOT self-heal —
+    /// PreToolUse applied after PermissionRequest, wiping the permission badge
+    /// while the agent is parked at the prompt — is guarded directly in
+    /// `ClaudeHookSignalSource` (PreToolUse leaves a pending `permission_prompt`
+    /// intact); other inversions self-heal on the next event. A hard transport
+    /// guarantee (per-session server-side sequencing) is the daemon-side
+    /// follow-up scoped out of #903.
     private static let asyncEvents: Set<String> = [
         "PostToolUse", "PostToolUseFailure",
     ]
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookSignalSource.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookSignalSource.swift
@@ -36,6 +36,22 @@ public struct ClaudeHookSignalSource: StateSignalSource {
                 ))
                 transition.newActivityState = .waiting
                 transition.toolActivity = .clear
+            } else if currentNotificationType == "permission_prompt" {
+                // Apply-order defense (#903). PreToolUse and PermissionRequest
+                // for the same tool can now apply out of order — the client no
+                // longer blocks on the daemon's reply, so the daemon may apply
+                // PermissionRequest first. If it did (permission_prompt already
+                // pending, state already .waiting), a later PreToolUse must NOT
+                // blanket-clear that badge and flip the card to .working while
+                // the agent is parked at the prompt — there is no "next event"
+                // to heal it until the user acts. Preserve both; still record
+                // the tool. Symmetric to PermissionRequest's `!= "question"`
+                // guard below. Normal (in-order) PreToolUse never sees a pending
+                // permission_prompt, so this branch can't fire there.
+                transition.notification = .leave
+                transition.toolActivity = .set(ToolActivity(
+                    toolName: toolName, isActive: true
+                ))
             } else {
                 transition.toolActivity = .set(ToolActivity(
                     toolName: toolName, isActive: true
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookSignalSource.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookSignalSource.swift
@@ -40,18 +40,18 @@ public struct ClaudeHookSignalSource: StateSignalSource {
                 // Apply-order defense (#903). PreToolUse and PermissionRequest
                 // for the same tool can now apply out of order — the client no
                 // longer blocks on the daemon's reply, so the daemon may apply
-                // PermissionRequest first. If it did (permission_prompt already
-                // pending, state already .waiting), a later PreToolUse must NOT
-                // blanket-clear that badge and flip the card to .working while
-                // the agent is parked at the prompt — there is no "next event"
-                // to heal it until the user acts. Preserve both; still record
-                // the tool. Symmetric to PermissionRequest's `!= "question"`
-                // guard below. Normal (in-order) PreToolUse never sees a pending
-                // permission_prompt, so this branch can't fire there.
+                // PermissionRequest first. If it did (permission_prompt pending,
+                // state already .waiting, tool activity already cleared), a late
+                // PreToolUse must NOT blanket-clear the badge or flip the card to
+                // .working while the agent is parked at the prompt — there is no
+                // "next event" to heal it until the user acts. Preserving the
+                // notification is enough: newActivityState and toolActivity stay
+                // at their .leave defaults, so the parked state is identical to
+                // the in-order case (which also ends .waiting with tool activity
+                // cleared by PermissionRequest). Symmetric to PermissionRequest's
+                // `!= "question"` guard below; normal in-order PreToolUse never
+                // sees a pending permission_prompt, so this can't fire there.
                 transition.notification = .leave
-                transition.toolActivity = .set(ToolActivity(
-                    toolName: toolName, isActive: true
-                ))
             } else {
                 transition.toolActivity = .set(ToolActivity(
                     toolName: toolName, isActive: true
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookSignalSource.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookSignalSource.swift
@@ -36,22 +36,6 @@ public struct ClaudeHookSignalSource: StateSignalSource {
                 ))
                 transition.newActivityState = .waiting
                 transition.toolActivity = .clear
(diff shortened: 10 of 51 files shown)


dhilgaertner (changes requested):
## Code & Security Review

The diagnosis is convincing (a clean fast-or-5s bimodal split really is the signature of a block) and the fix is the right shape. The `connectAndWrite` extraction is correct, the socket-level reasoning in the doc comments holds up, and the scoping note about the daemon-side follow-up is appreciated. Two things should land in the same round trip before merge.

### Critical Issues

None. No Red findings.

### Security Review

**Strengths:**
- No change to the socket's security posture: still `0o600` inside a `0o700` directory (`SocketServer.swift:86`, `:43`), same owner-only trust boundary, no new inputs and no new privilege.
- `connectAndWrite` keeps `signal(SIGPIPE, SIG_IGN)` on the client (`SocketClient.swift:106`), and the server already ignores SIGPIPE in `start()` (`SocketServer.swift:53`), so a late reply to a closed connection hits the `write() < 0 → return` branch in `writeResponse` (`SocketServer.swift:196`) and is dropped — the doc comment's claim at `SocketClient.swift:83-87` checks out.
- The `keepOpen` guard (`SocketClient.swift:113-114`, `:159`) is correct: the fd is closed on every throw inside the helper, handed back on success, and closed exactly once by each caller (`send`'s `defer`, `post`'s direct `close`). No leak, no double close.
- Dropping `printJSON(result)` is a security/robustness improvement in its own right — writing arbitrary JSON to a `PreToolUse` hook's **stdout** is precisely the channel Claude Code parses for permission decisions, so the old error print was feeding attacker-influenceable-ish text into a decision surface. Good riddance.
- The immediate-close claim is sound for `AF_UNIX`/`SOCK_STREAM`: `write()` copies into the peer's receive queue, `close()` without `SO_LINGER` does not discard already-queued data, and the peer reads it before EOF. Verified empirically by the new `postReturnsWithoutWaitingForHandler` test.

**Concerns:**
- None. No new attack surface.

### Code Quality

**🟡 Yellow — `hook-event` RPC failures are now invisible in every channel**

`forwardHookEvent` previously called `rpc()`, which throws `ValidationError` on `response.error` (`Helpers.swift:16-18`) → non-zero exit → stderr. Noisy, but diagnosable. After this PR the reply is never read, and nothing else picks up the slack:

- `CommandRouter.handle` converts a thrown handler error into a `JSONRPCResponse.error` and returns it — it does not log (`CommandRouter.swift:45-52`).
- `SocketServer.writeResponse` writes that response into the now-closed fd and drops it (`SocketServer.swift:196`).
- The `hook-event` handler's only logging is `hookDebug`-gated and sits *after* session resolution (`EngineRouter.swift:1296-1300`), so the `throw RPCError.invalidParams("session_id required or resolvable from payload cwd")` at `EngineRouter.swift:1291` produces no log line at all.

There is zero `CrowLog`/`os_log` anywhere in `Packages/CrowIPC/Sources/CrowIPC/`, and `CrowDaemon.swift:431` passes the router straight through with no logging decorator — so this is the complete picture, not a gap in my search.

Concrete failure: a Codex / OpenCode / Antigravity global hook config carries no `--session`, so resolution depends on `cwd` matching a registered worktree (`EngineRouter.swift:1286`). Run the agent from a path that doesn't map to one and **every** hook event for that session is now dropped silently — no exit code, no stderr, no daemon log. That is the same class of bug #897 was about, with the last diagnostic channel removed.

Please add a server-side log on the failure path — `CrowLog.info`/`.error` in the `hook-event` handler before the `invalidParams` throw, or (broader) log non-`methodNotFound` errors in `CommandRouter.handle`. The goal isn't to restore the stderr noise; it's that fire-and-forget must not mean fire-and-never-know.

**🟡 Yellow — `EngineRouter.swift:1277-1281` now documents a premise this PR falsified**

```swift
// Deliberately never throws for an unknown-but-provided id:
// `crow hook-event` surfaces an RPC error as a non-zero
// exit, which Claude Code renders as the very "hook error"
// noise this change exists to remove.
```

`crow hook-event` no longer surfaces RPC errors as anything. The *behavior* (falling through to the provided id at `:1288-1289` rather than throwing) is still what you want, but its stated justification is now false — and it's load-bearing: it's the comment a future maintainer reads before deciding whether that fallback can be removed. One-line update in the subsystem this PR is already about.

**🟢 Green — `post` still blocks in `write()`, and the doc's "the agent never waits" isn't unconditional**

Neither `send` nor `post` sets `SO_SNDTIMEO` (only `SO_RCVTIMEO`, `SocketClient.swift:46-47`), so the write loop at `:143-156` blocks until the server drains. That's fine for the measured `PreToolUse:Bash` case, but `parseHookPayload` forwards the agent's *entire* stdin object (`HookEventCommand.swift:36`), and Claude Code's `PreToolUse` payload carries the full `tool_input` — for `Write`/`Edit` that's the whole file body, easily past the default `net.local.stream.sendspace`. Once the payload exceeds the socket buffer, `post` depends on `handleClient` getting a GCD worker to drain it — which is exactly what's scarce when a burst of hook-events has parked workers on `semaphore.wait()` (`SocketServer.swift:177`) behind a busy MainActor. Narrower than the bug you fixed, and the agent's own 5s timeout still bounds it, but worth either an `SO_SNDTIMEO` on the `post` path or softening the unconditional claim at `SocketClient.swift:73-81` and `Helpers.swift:22-27`.

**🟢 Green — nothing at the CLI layer pins `forwardHookEvent` to the notify path**

`postReturnsWithoutWaitingForHandler` guards `SocketClient.post` against a reintroduced read, which is the right unit-level guard. But reverting `HookEventCommand.swift:70` from `rpcNotify` back to `rpc` leaves all 341 CrowCLI tests green — `forwardHookEventSilentWhenAppNotRunning` passes either way, since it only exercises the absent-socket path. A CrowCLI test that stands up a `SocketServer` with a slow handler under `CROW_SOCKET` and times `forwardHookEvent` would close that gap; the test target already imports `CrowIPC` and `SocketServer`'s init is public.

### Verification performed

| Check | Result |
|---|---|
| `swift build` | Clean (59.8s), no new warnings from these files |
| `swift test --package-path Packages/CrowIPC` | 39/39 pass, incl. both new tests |
| `swift test --package-path Packages/CrowCLI` | 341/341 pass, incl. `forwardHookEventSilentWhenAppNotRunning` and the CLI/RPC parity gate |

Test-plan claims in the PR description are accurate as written.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 2 Green] findings. The core change is correct and I'd like it to land; the two Yellows are small and belong in this round trip rather than a follow-up.

---

dgershman:
Thanks for the thorough review — addressed all four findings in c7021154.

**🟡 Yellow — invisible hook-event failures.** Added `CrowLog.error` on both drop paths in the `hook-event` handler: the unresolved-session `throw` (`EngineRouter.swift`, the concrete silent-drop you described — global hook config, no `--session`, cwd outside any registered worktree, cf. #897) and the `event_name required` throw. Went with your narrower option (log in the handler) rather than in `CommandRouter.handle`: as you noted there's zero logging in `CrowIPC`, and adding `CrowLog` there means pushing a logging dependency into the low-level IPC package — the handler is in `CrowEngine` where `CrowLog` already lives. `CrowLog` is non-blocking (stderr + os_log), so it's safe on the MainActor and doesn't reintroduce a #874-class wedge.

**🟡 Yellow — falsified comment at `EngineRouter.swift`.** Rewrote it. The behavior (fall through to the provided id rather than throw) is unchanged; the rationale now says we'd rather record the event under the given id than drop it, and notes the pre-#903 "hook noise" reason is gone since the client no longer reads the reply.

**🟢 Green — `post` still blocks in `write()`.** Chose to soften the docs rather than add `SO_SNDTIMEO`. A write timeout throws `writeFailed` out of `forwardHookEvent` as a non-zero exit — the exact hook-error noise this PR removes — unless I also swallow it, which would re-bury genuine write failures (and the server wouldn't log it either, since it never arrived). The `post` doc now scopes the guarantee to the reply and documents the send-buffer caveat (a `Write`/`Edit` hook forwarding a whole file body can block until the daemon drains it, bounded by the agent's own hook timeout); `rpcNotify` points at it. If you'd prefer the `SO_SNDTIMEO` + silent-swallow shape instead, happy to switch — it just trades "bounded but silent drop" for "unbounded-but-timeout-capped wait", and the doc felt like the more honest default.

**🟢 Green — nothing pinned `forwardHookEvent` to the notify path.** Added `forwardHookEventDoesNotBlockOnSlowDaemon` in `HookEventTests.swift`: stands up a `SocketServer` under `CROW_SOCKET` with a never-replying handler and asserts `forwardHookEvent` returns in <5s. Reverting the call site from `rpcNotify` back to `rpc` now fails it.

**Verification:** `swift build` clean; CrowIPC 39/39, CrowCLI 342/342 (incl. the new test + `forwardHookEventSilentWhenAppNotRunning`), CrowEngine hook-event resolution suite 4/4.

dhilgaertner (changes requested):
## Code & Security Review

The diagnosis is convincing (a clean bimodal fast-or-5s split really is the signature of a block), the fix is the right shape, and the docs are unusually honest about what it does *not* solve — including the residual `write()` wait. `connectAndWrite`'s `keepOpen` refactor is correct: the fd is closed on every throw path and handed back exactly once on success. The AF_UNIX claim in the `post` docstring is right — data written before `close()` sits in the peer's receive queue and is readable ahead of the EOF, and `SocketServer.start()` does `signal(SIGPIPE, SIG_IGN)` (SocketServer.swift:53), so the orphaned reply is dropped harmlessly.

Verified locally: `swift build` clean (12.2s), `swift test --package-path Packages/CrowIPC` 39/39, `swift test --package-path Packages/CrowCLI` 342/342.

Two things should be addressed before merge.

### Critical Issues

None.

### Code Quality

**🟡 The change removes an ordering guarantee the codebase explicitly documents as load-bearing** — `Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift:20`

```swift
/// PreToolUse is intentionally NOT async — it must arrive before
/// PermissionRequest so the state machine ordering is reliable.
private static let asyncEvents: Set<String> = ["PostToolUse", "PostToolUseFailure"]
```

`CursorHookConfigWriter.swift:60` carries the same deliberate split. Pre-#903, `async: false` gave an *end-to-end* guarantee: the agent waited for the hook process, which waited for the daemon's reply, which is written only after the handler has fully applied the transition (`SocketServer.handleClient` → `semaphore.wait()` → `writeResponse`). Event N was applied before event N+1 was even sent.

After this PR, `async: false` only means the agent waited for the *bytes to be written*. The accept loop still accepts in order, but `SocketServer.swift:124` fans every connection out to `DispatchQueue.global(qos: .userInitiated).async`, each spawning an independent `Task` racing to `MainActor.run` — so accept order no longer implies apply order. The PR makes all 17 events fire-and-forget at the transport layer, which is precisely what those two `asyncEvents` sets were curated to avoid.

Concrete failure: a `Bash` call that needs approval fires `PreToolUse` then `PermissionRequest`. Applied inverted, `PermissionRequest` sets `notification = .set(permission_prompt)` and `.waiting`, then `PreToolUse` (`ClaudeHookSignalSource.swift:23`, `blanketClear` is true for `PreToolUse`) **clears that notification** and sets `newActivityState = .working`. The session card reads "working" with no waiting badge and no pending notification while the agent is actually blocked at the permission prompt — and stays wrong until the next event, which for a blocked prompt may be minutes. The window is small (the two hook processes are milliseconds apart) but it widens under exactly the busy/burst conditions this PR targets: each in-flight hook-event pins a global-queue thread on `semaphore.wait()` while the MainActor is held, and threads unblocking in a burst are where inversion becomes plausible.

Either sequence hook-event application server-side in accept order (a serial hop, or per-session), or — if the race is judged acceptable — say so explicitly and fix the two now-false comments, since they currently describe a guarantee the transport no longer provides.

**🟡 Both `CROW_SOCKET` tests mutate process-global state under a parallel runner** — `Packages/CrowCLI/Tests/CrowCLITests/HookEventTests.swift:59` and `:88`

Swift Testing parallelizes by default, and `make test` (`Makefile:97`) does not pass `--no-parallel` — only CI does (`.github/workflows/ci.yml:66`). Two consequences:

1. Concurrent `setenv`/`unsetenv` against another thread's `ProcessInfo.processInfo.environment` read (which walks `environ`) is a data race, not just a logical one.
2. Functionally it defeats the new gate: if `forwardHookEventSilentWhenAppNotRunning` wins the `setenv` between the new test's `setenv` and its `forwardHookEvent`, the new test connects to the *nonexistent* path, takes `connectionFailed`, returns in ~0ms and passes vacuously — the #903 regression guard silently stops testing anything, in the one configuration (local `make test`) where it isn't serialized. It didn't reproduce in 8 runs here, which is the problem: it fails open.

Wrapping the two in a `@Suite(.serialized)` fixes both, and is cheaper than threading an injectable socket path through `rpcNotify`.

**🟢 The new CLI test is the weaker twin of its CrowIPC counterpart** — `HookEventTests.swift:82`. `postReturnsWithoutWaitingForHandler` asserts *delivery* via the semaphore; this one only asserts elapsed `< 5`, so it can't distinguish "posted" from "silently no-op'd" (see above). Its 60s-sleeping handler also pins a GCD worker for the remainder of the test process, and a regression to `rpc` fails only after the 30s read timeout rather than promptly.

**🟢 `postToNonexistentSocketThrows` is broader than its own docstring** — `SocketRoundTripTests.swift:289`. The comment says the point is that `post` raises the same `connectionFailed` that `forwardHookEvent` catches, but `#expect(throws: SocketError.self)` also passes on `createFailed`. A `do/catch` asserting the specific case would hold the contract the docstring claims.

**🟢 Unconditional error-level log on a per-event path** — `EngineRouter.swift:1305`. Every other hook-event log in this handler is gated behind `hookDebug`. The unresolved-session branch is the common steady state for a global hook config (OpenCode/Antigravity outside a registered worktree), so it emits one error line per event indefinitely. `CrowLog.error` goes to stderr + os_log — under `crow autostart` that's an unrotated launchd log. Worth deduping per `(session_id, cwd)` rather than dropping the line; it's the only diagnostic channel left, which is exactly why it shouldn't drown itself.

**🟢 Stale comment the "honest docs" sweep missed** — `AntigravityHookConfigWriter.swift:39`: "`crow hook-event` is observational and may print a JSON-RPC error on failure". The error print is gone as of this PR. The suppression it justifies is still correct, so this is cosmetic.

**🟢 The residual `write()` block is real for more than the measured case.** `SocketClient.swift:100` documents it well, but "a payload larger than the socket send buffer" isn't exotic — a Unix stream socket's default send space is on the order of 8 KB, and `PreToolUse` for `Write`/`Edit` carries the whole file body in `tool_input`. The doc's "until the daemon drains it" also assumes a drainer: `handleClient` reads the request *before* touching the MainActor, so normally it drains fine — but each blocked connection holds a global-queue thread on the semaphore, and once that pool saturates during a busy window, accepted connections aren't read at all. Strictly better than the status quo either way; noting it so the daemon-side follow-up ticket captures it (and so `SocketServer`'s one-`read()`-per-byte loop gets looked at alongside it).

### Security Review

**Strengths:**
- No change to the trust boundary: same 0600 socket in a 0700 dir, same local-only surface, no new RPC method or parameter.
- Failure modes stay closed — a missing daemon is still a silent no-op (`SocketError.connectionFailed`), and `createFailed`/`writeFailed` still propagate rather than being swallowed.
- No fd leak in the refactor; no secrets, credentials, or user content added to any new output.

**Concerns:**
- Minor, same-trust-boundary: `EngineRouter.swift:1305` interpolates caller-supplied `eventName` and payload `cwd` into a log line unescaped, so an embedded newline can forge a log record. Both values originate from the local user's own agent, so this is hygiene rather than a vulnerability — but it pairs with the rate-limiting note above.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 5 Green] findings. Neither Yellow is an argument against the approach; the ordering one needs either a server-side sequence or an explicit "we accept this" plus a comment fix, and the test one is a three-line `@Suite(.serialized)`.

---

dgershman:
Round 2 addressed in 332ed500.

**🟡 Ordering guarantee.** You're right that fire-and-forget drops the end-to-end apply ordering to arrival-order-only, and that the tight window is `PreToolUse` → `PermissionRequest`. I took the "accept it explicitly + fix the comments" path rather than the server-side sequence, because a real per-session sequencer needs an arrival-order stamp established at accept time and threaded through the generic transport (or a per-session reorder buffer) — that's exactly the daemon-side structural work #903 scopes to a follow-up, and doing it half-way in `SocketServer` is the wrong layer. Kept the sync/async split (non-async still keeps `PreToolUse` *accepted* ahead of `PermissionRequest`, which narrows the window to the MainActor scheduling race; making it async would also race the writes), and rewrote both `ClaudeHookConfigWriter` and `CursorHookConfigWriter` comments to state that arrival order no longer implies apply order and that per-session sequencing is the deferred fix. If you'd rather block merge on the sequencer landing here instead of the follow-up, say the word — but I read your "if the race is judged acceptable, say so explicitly and fix the comments" as sanctioning this.

**🟡 CROW_SOCKET tests.** Wrapped both in `@Suite(.serialized)`. Combined with the delivery assertion below, the vacuous-pass hole is closed from both directions.

**🟢 CLI test is the weaker twin.** Now asserts delivery via a semaphore the handler signals, so it can't pass by silently no-op'ing. Swapped the 60s sleep for a polled `Latch` released in a `defer` — the handler returns within a poll of release, so no socket worker is pinned for the run. (The 30s-to-fail-on-`rpc`-regression is inherent to `SO_RCVTIMEO` and unavoidable without exposing the timeout; it still fails, just slowly.)

**🟢 `postToNonexistentSocketThrows` broader than its docstring.** Now `do`/`catch`es the specific `.connectionFailed` — the case `forwardHookEvent` actually relies on.

**🟢 Unconditional per-event error log + 🔒 log-injection.** Deduped per `(session_id, cwd)` (bounded set, cleared past 256) so an unmapped global hook config logs once, not per event; and newlines in the caller-supplied `eventName`/`cwd` are escaped so an embedded newline can't forge a log record.

**🟢 Stale Antigravity comment.** Fixed — `crow hook-event` prints nothing to stdout now, so the output suppression is noted as belt-and-suspenders rather than "may print a JSON-RPC error."

**🟢 Residual `write()` block / `SocketServer` byte-loop.** Acknowledged for the daemon-side follow-up — the `post` docstring already documents the send-buffer caveat; I'll make sure the follow-up ticket captures both this and the one-`read()`-per-byte loop. No code change this round.

**Verification:** `swift build` clean; CrowIPC 39/39, CrowCLI 342/342 (incl. the serialized suite), CrowEngine hook-event resolution 4/4, and the three hook-config-writer suites (Claude 77, Cursor 67, Antigravity 39) green.

dhilgaertner (changes requested):
## Code & Security Review

Verified locally on `feature/crow-903-hook-event-fire-and-forget`: `swift build` clean; `swift test --package-path Packages/CrowIPC` 39/39; `swift test --package-path Packages/CrowCLI` 342/342 (PR body says 341 — one more now, no failures).

The core mechanism is right. `connectAndWrite` is a correct refactor — the `keepOpen` flag closes the fd on every throw path and hands it off only on success, so neither `send` nor `post` leaks or double-closes, and moving `SO_RCVTIMEO` after the write is a no-op for reads. The AF_UNIX close-after-write reasoning in the doc comment holds: bytes are queued into the peer's receive buffer at `write()` time, so `SocketServer.handleClient`'s byte-at-a-time read drains the full request before it sees EOF, and the late `writeResponse` to the dead fd hits `EPIPE` under the `SIG_IGN` set in `SocketServer.start()` (SocketServer.swift:53) rather than signalling. The residual `write()` block is also bounded better than the doc claims — the server's read loop runs on the global queue *before* the MainActor hop, so draining doesn't wait on the MainActor at all.

### Security Review

**Strengths:**
- `logUnresolvedHookDropOnce` escapes `\n`/`\r` in both `eventName` and `cwd` (EngineRouter.swift:76-78) before they reach `CrowLog.error`. Both are attacker-influenced (payload `cwd` comes from the local agent), and this correctly prevents forging a second log record.
- The dedup set is capacity-bounded rather than growing per distinct cwd.
- No new parsing, no new privileged surface, no credential handling. `post` reuses the same 0600 socket and the same encode path as `send`.

**Concerns:** see Yellow #1 below (log-volume, not confidentiality).

### Findings

**🟡 1. The new `missing event_name` log is the one un-deduped, uncapped log site — the exact thing the sibling site was written to avoid** — `EngineRouter.swift:1238`

Three lines below it, `logUnresolvedHookDropOnce` exists precisely because "an unmapped global hook config re-fires on every event, so without this the only diagnostic channel left after #903 would flood an unrotated launchd log" (EngineRouter.swift:65-69). The `event_name` guard emits one unconditional `CrowLog.error` per request, with no dedup and no cap.

It isn't reachable from `crow hook-event` (`--event` is a required option, HookEventCommand.swift:23), but the CLI is not the only caller: `hook-event` is **not** in `RPCWebSocketHandler.localOnlyDenial` (RPCWebSocketHandler.swift:193-236), so any authenticated `/rpc` peer — or any local process that can write the socket — drives one stderr + `os_log` record per request with nothing throttling it. Route it through the same dedup helper (or drop it, since a params-shaped bug is a caller bug, not a Crow diagnostic), so the two sites don't disagree about whether this channel needs flood control.

**🟡 2. "Any inversion self-heals on the next event" doesn't hold for the pair the sync setting exists to protect** — `ClaudeHookConfigWriter.swift:24-34`

The note is admirably honest that accept order no longer implies apply order, but the mitigation claim is the load-bearing part and it doesn't survive the PreToolUse → PermissionRequest case. If those two invert:

1. `PermissionRequest` applies: `notification = .set(permission_prompt)`, `newActivityState = .waiting` (ClaudeHookSignalSource.swift:72-83).
2. `PreToolUse` applies second. For any tool other than `AskUserQuestion`, `blanketClear` is true (ClaudeHookSignalSource.swift:23-27) so it wipes the pending notification, and the switch sets `newActivityState = .working` (ClaudeHookSignalSource.swift:40-43).

The session then reads **working, no permission badge** while Claude is actually parked at the permission prompt — the precise state Crow exists to surface. And there is no "next event" to heal it: the agent is blocked until the user acts. Recovery depends entirely on Claude Code separately emitting `Notification`/`permission_prompt`, which is a different mechanism than what the comment asserts.

The window is small (a second hook process has to spawn and reach the MainActor queue faster than the first, ~10ms of slack) and it narrows to zero for `AskUserQuestion`, where PreToolUse's `.set("question")` wins either way. But it widens exactly under the busy-daemon load this PR targets. Two acceptable resolutions: qualify the comment to state the concrete failure mode and its real recovery path, or make `PreToolUse`'s `blanketClear` not clobber a `permission_prompt` set by a later-numbered event. Either way this deserves a test — the reordering is currently asserted only in prose.

**🟢 3. The dedup key's `session_id` component is always `none`** — `EngineRouter.swift:1331-1335`

The `else` is only reachable when `providedSessionID` is nil — the preceding `else if let provided = providedSessionID` (line 1321) catches every non-nil case. So `sessionID:` is always nil, the key degenerates to cwd-only, and every emitted line reads `session_id=none`. Harmless, but the parameter and the "deduped per (session_id, cwd)" doc both describe something that can't happen.

**🟢 4. Overflow reset defeats the dedup it's bounding** — `EngineRouter.swift:81`

`removeAll()` at >256 drops the key inserted one line earlier, so it re-logs immediately, and a workload cycling more than 256 distinct cwds re-logs *everything* on each cycle. Holding the cap by refusing new inserts once full (or evicting one entry) keeps suppression intact for the keys already seen.

**🟢 5. Hardcoded `/tmp` path in a new test** — `SocketRoundTripTests.swift:293`

`postToNonexistentSocketThrows` uses a fixed `/tmp/nonexistent-crow-post-test.sock` while every other socket test in the file goes through `tempSocketPath()` (line 13), and the repo convention is `$TMPDIR` over `/tmp`. `/tmp` is world-writable — a leftover or foreign socket at that name turns a `connectionFailed` assertion into a pass-through. `tempSocketPath()` is right here.

### Code Quality

- `@Suite(.serialized)` on `HookEventForwarding` is correctly scoped: `CROW_SOCKET` is mutated in exactly these two tests across the whole tree, so serializing the suite genuinely closes both the `environ` data race and the vacuous-pass hole the comment describes. Good catch, and the reasoning is written down.
- The `Latch` in `HookEventTests.swift:8-15` is the right call over a 60s sleep, and the defer ordering (release before `server.stop()`) is correct.
- Worth noting for the daemon-side follow-up, not this PR: removing the client wait also removes the backpressure that capped in-flight hook connections at ~1 per session. Several can now be outstanding at once, each pinning a global-queue thread in `handleClient`'s `semaphore.wait()` while the MainActor is busy.
- Doc comments across `AntigravityHookConfigWriter`, `CursorHookConfigWriter`, and `SocketClient.post` are accurate about what changed. The "prints nothing to stdout" claim in the Antigravity writer checks out — the only remaining write is `parseHookPayload`'s stderr warning.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 3 Green] findings. The fix itself is sound and well-scoped; #1 is a small consistency fix and #2 is primarily about making the documented tradeoff match its actual failure mode.

dgershman:
Round 3 addressed in 48df67de (rebased onto the main-merge that landed on the branch).

**🟡 #2 — the inversion that doesn't self-heal.** You're right that "self-heals on the next event" is false for `PreToolUse` → `PermissionRequest` — the agent is blocked, there is no next event. Took the stronger fix rather than just qualifying the comment: `ClaudeHookSignalSource` now leaves a pending `permission_prompt` intact (and doesn't downgrade `.waiting` → `.working`) when a generic-tool `PreToolUse` arrives while one is set. Normal in-order `PreToolUse` never sees a pending `permission_prompt`, so only the inverted case is affected, and it's symmetric to the existing `PermissionRequest` `!= "question"` guard. Added `preToolUseKeepsPendingPermissionPrompt` so the reordering is asserted in a test, not just prose, and rewrote the `ClaudeHookConfigWriter` note to describe the guard + the real failure mode it removes.

**🟡 #1 — the un-deduped `event_name` log.** Dropped it. As you note it's unreachable from the CLI but drivable per-request by any `/rpc` peer (it's not in `localOnlyDenial`), so it was exactly the flood the sibling site is deduped against. A params-shaped request is a caller bug and the RPC error already signals it, so there's nothing for a Crow-side log to add.

**🟢 #3 — dedup key `session_id` always `none`.** Correct — that branch is only reached when no `session_id` was provided. Keyed on `cwd` alone and dropped the param; the line now reads `no session_id, cwd=…`.

**🟢 #4 — overflow reset defeated the dedup.** Replaced `removeAll()` with refuse-when-full, so the keys already seen stay suppressed instead of re-logging every cycle.

**🟢 #5 — hardcoded `/tmp`.** `postToNonexistentSocketThrows` now uses `tempSocketPath()` like the rest of the file.

Also folding in your two "for the follow-up, not this PR" notes (removing the client wait removes the ~1-per-session backpressure on in-flight hook connections; the one-`read()`-per-byte loop) — I'll make sure the daemon-side follow-up ticket captures both alongside the per-session sequencing.

**Verification:** `swift build` clean (full rebuild after the merge); CrowIPC 39/39, CrowCLI 342/342, CrowEngine hook-event resolution 4/4, CrowClaude 78/78 (incl. the new guard test).

dhilgaertner (changes requested):
## Code & Security Review

Round 4. Verified locally on `feature/crow-903-hook-event-fire-and-forget` @ 48df67d: `swift build` clean (61.3s); `swift test --package-path Packages/CrowIPC` 39/39; `Packages/CrowCLI` 342/342; `Packages/CrowClaude` 78/78.

Everything from round 3 landed, and landed well:

- The `event_name required` flood site is gone entirely, with a comment explaining why a params-shaped request is a caller bug rather than a Crow diagnostic (`EngineRouter.swift:1261-1266`). Right call — that's the site a remote `/rpc` peer could drive.
- The dedup key is now `cwd` alone and the doc says so (`EngineRouter.swift:74-77`), matching the fact that the branch is only reachable with `providedSessionID == nil`.
- The cap now holds by **refusing** new keys once full (`EngineRouter.swift:85`) instead of `removeAll()`, so suppression survives; the trade-off is stated in the comment.
- `postToNonexistentSocketThrows` uses `tempSocketPath()` and a `do`/`catch` on the specific `connectionFailed` case (`SocketRoundTripTests.swift:590-604`).
- `@Suite(.serialized)` on `HookEventForwarding`, and the CLI twin now asserts *delivery* via `received.wait(...)` rather than elapsed time alone (`HookEventTests.swift:197`) — it can no longer pass by silently no-op'ing. The `Latch` instead of a sleeping handler is a nice touch.
- The `permission_prompt` inversion is guarded at `ClaudeHookSignalSource.swift:39-54` with a matching test (`ClaudeHookSignalSourceTests.swift:273`).

I re-checked the guard's premise ("normal, in-order `PreToolUse` never sees a pending `permission_prompt`") and it holds: in order, the badge set by `PermissionRequest` is cleared by that tool's `PostToolUse` blanket-clear before the next `PreToolUse`, and a new turn's `UserPromptSubmit` clears it too. The `AskUserQuestion` case also resolves correctly in both orders, since `PreToolUse`'s `.set("question")` wins either way.

### Critical Issues

None.

### Security Review

**Strengths:**
- No change to the trust boundary: same 0600 socket in a 0700 directory, same encode path, no new parsing, no credentials.
- The `\n`/`\r` escaping in `logUnresolvedHookDropOnce` (`EngineRouter.swift:87-89`) still covers both attacker-influenced fields before they reach `CrowLog.error`.
- Removing `printJSON(result)` from the hook path remains a real hardening win — `PreToolUse` stdout is the channel Claude Code parses for permission decisions.

**Concerns:** one diagnostic-availability nit, Green 2 below. No confidentiality or integrity issue.

### Findings

**🟡 1. The apply-order guard is Claude-only, but the transport change is global — Codex has the same pair, unguarded** — `Packages/CrowCodex/Sources/CrowCodex/CodexSignalSource.swift:43-48`

`ClaudeHookConfigWriter.swift:31-34` names this failure precisely: *"The one inversion that would NOT self-heal — PreToolUse applied after PermissionRequest, wiping the permission badge while the agent is parked at the prompt."* The fix went into `ClaudeHookSignalSource` only. `CodexSignalSource` still has the unguarded shape:

```swift
case "PreToolUse":                                  // :43
    transition.toolActivity = .set(...)
    transition.newActivityState = .working          // :48   (+ blanketClear wipes the badge)

case "PermissionRequest":                           // :65
    if currentNotificationType != "question" { ...set(permission_prompt) }
    transition.newActivityState = .waiting          // :75
```

Codex registers **both** events (`CodexHookConfigWriter.swift:21-27`), and its `asyncEvents` is empty *by necessity* — "Codex's hook runtime is sync-only as of v0.139.0" (`CodexHookConfigWriter.swift:29-35`). So pre-#903 every Codex event had the end-to-end apply ordering this PR removes; Codex loses strictly more of it than Claude does, and gains none of the compensating guard. Inverted, a Codex session parked at an approval prompt reads **working, no badge**, with no next event to heal it until the user acts.

I checked the other two adapters so the ask stays small — neither is exposed today:
- **Antigravity**: `PreToolUse` is deliberately not registered (`AntigravitySignalSource.swift:64-68`).
- **Cursor**: no Cursor event maps to `PermissionRequest` (`CursorSignalSource.swift:75-78`).
- **OpenCode**: in-order is `permission.ask` → `PermissionRequest` *before* `tool.execute.before` → `PreToolUse` (`OpenCodeHookConfigWriter.swift:248-256`), so an inversion there lands on the *correct* final state (badge + `.waiting`) and self-heals.

So this is one adapter and ~6 lines mirroring `ClaudeHookSignalSource.swift:39-54` (`CodexSignalSource.swift:66-67` already says "Same precedence rule as Claude", so parity is an explicit goal in that file). If Codex's `PreToolUse`/`PermissionRequest` ordering is known to differ from Claude's — I couldn't confirm it from this repo, only that the vocabulary is mirrored from the generated schema — then say so in the `CodexSignalSource` comment instead, so the next reader doesn't have to re-derive it.

**🟢 2. `hook-event` is remotely reachable, so the 256-entry dedup cap can be poisoned** — `EngineRouter.swift:85`, `RPCWebSocketHandler.swift:193-236`

`hook-event` is not in `localOnlyDenial`, so an authenticated `/rpc` peer can send 256 unresolvable events with distinct synthetic `cwd`s and permanently fill `loggedUnresolvedHookDrops`. Because the cap now holds by refusing new keys (the round-3 fix — correct on its own terms), that silences the *only* remaining diagnostic for genuine drops for the daemon's lifetime. Authenticated-only and diagnostics-only, so not a Yellow, and I'm not asking you to swap the eviction policy back. Either gating `hook-event` local-only or noting the poisoning window next to the cap would close it.

**🟢 3. The guard re-sets `toolActivity` that `PermissionRequest` deliberately cleared** — `ClaudeHookSignalSource.swift:52-54` vs `:99`

In the inversion path the late `PreToolUse` sets `.set(toolName, isActive: true)`, but `PermissionRequest` clears tool activity on purpose so the card doesn't show a running tool while parked. The comment calls this out ("still record the tool"), so it's deliberate — the note is only that the same user-visible state now differs depending on apply order, where the rest of the guard is careful to make it not. It heals on the tool's `PostToolUse`.

### Code Quality

- `connectAndWrite`'s `keepOpen` refactor still checks out: fd closed on every throw, handed back once on success, closed exactly once by each caller (`send`'s `defer`, `post`'s direct `close`). No leak, no double close.
- The AF_UNIX reasoning in the `post` docstring (`SocketClient.swift:91-95`) is correct, and the softened send-buffer caveat at `:83-89` is accurate — `handleClient`'s read loop runs on the global queue *before* the MainActor hop, so draining doesn't wait on the MainActor.
- The new `EngineRouter.swift:1274-1286` rationale reads correctly now: "recording the event under the id the hook handed us beats dropping it," with the falsified pre-#903 justification marked as historical rather than deleted.
- The 60s-sleeping handler in `postReturnsWithoutWaitingForHandler` (`SocketRoundTripTests.swift:271`) matches the existing idiom in that file (`:178`, `:309`), so no objection — noting only that the CLI twin's `Latch` is the better shape if that file is ever swept.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The core transport change is correct and I'd take it as-is; the one Yellow is finishing the apply-order defense you already wrote, in the one other adapter that needs it.

---

dgershman:
Round 4 addressed in eee73a46.

**🟡 #1 — Codex had the same unguarded pair.** You're right — the guard was Claude-only while the transport change is global, and Codex loses more ordering (sync-only, empty `asyncEvents`) with no compensation. Mirrored the guard into `CodexSignalSource` with a matching test (`preToolUseKeepsPendingPermissionPrompt`), and noted in-comment that Codex's PreToolUse/PermissionRequest ordering is assumed to match Claude's via the shared hook schema — if that's ever found to differ, that's the spot to record it. Thanks for pre-checking the other three adapters; I confirmed the same (Antigravity no PreToolUse, Cursor no PermissionRequest, OpenCode's in-order is PermissionRequest-first so an inversion self-corrects).

**🟢 #3 — the guard re-set toolActivity PermissionRequest had cleared.** Good catch — that made the inverted parked state differ from in-order (running-tool badge while parked). Reduced the guard to preserving *only* the notification; `newActivityState`/`toolActivity` stay at their `.leave` defaults, so the inverted state is now byte-identical to the in-order one. Applied to both guards; the tests now assert `toolActivity` is left untouched.

**🟢 #2 — remote-reachable dedup cap.** Took the stronger option and gated `hook-event` local-only in `localOnlyDenial`. It's only ever emitted by agent hook processes over the 0600 Unix socket (no `/rpc` or desktop caller — I checked), so this has zero legitimate-use impact and matches the existing trust model for host-local signals (`run-setup`, `open-in-vscode`). That closes the cap-poisoning outright — and retroactively the remote-flood surface that ran through the last two rounds — rather than just documenting it. Added a `hookEventIsLocalOnly` security test and a note by the cap.

An existing Codex test (`nonPermissionRequestClearsPendingNotification`) used `PreToolUse`+`permission_prompt` as its blanket-clear stand-in — exactly the case the guard now special-cases — so I retargeted it to `PostToolUse`, preserving the blanket-clear coverage without colliding.

**Verification:** `swift build` clean; CrowIPC 39/39, CrowCLI 342/342, CrowClaude 78/78, CrowCodex 71/71, CrowDaemon 306/306 (incl. the new local-only test), CrowEngine hook-event resolution 4/4.

dhilgaertner (changes requested):
## Code & Security Review

The core change is right, and the socket-level reasoning is unusually careful: `connectAndWrite`'s `keepOpen` flag makes fd ownership explicit on every throw path, and the AF_UNIX "data already accepted is delivered before the FIN" argument is correct (a stream-socket `write()` queues into the peer's receive queue, so `close()` can't discard it). The tests assert both halves — non-blocking *and* delivery — so they can't pass by silently no-op'ing, and `HookEventForwarding` is `.serialized` for the `CROW_SOCKET` env mutation. Verified locally: `swift build` clean; CrowIPC 39/39, CrowCLI 342/342, CrowClaude 78/78, CrowCodex 71/71, CrowDaemon 306/306, CrowEngine 571/571.

The concerns are all in the *compensating* apply-order guard, not the transport.

### Critical Issues

None.

### Code Quality

**🟡 1. The PreToolUse permission guard suppresses the only healer for `.waiting`, so the reverse inversion strands the card**
`Packages/CrowClaude/Sources/CrowClaude/ClaudeHookSignalSource.swift:39` (and `CodexSignalSource.swift:44`)

The guard fires on *any* `PreToolUse` while a `permission_prompt` is pending — it never checks that the pending prompt belongs to the incoming tool. But `PermissionRequest` sets `.waiting`, and in `ClaudeHookSignalSource` the transitions that return to `.working` are `PreToolUse`, `PermissionDenied`, `UserPromptSubmit`, and `SubagentStart`; `PostToolUse` sets no activity state, and `TaskCreated`/`TaskCompleted`/`SubagentStop` are explicitly gated on `currentActivityState != .waiting`. So after a user approves a permission, the **next `PreToolUse` is the only thing that un-parks the session** — and this guard is what swallows it.

Failure sequence:
1. `PermissionRequest(A)` → badge set, `.waiting`, tool activity cleared.
2. User approves; tool A runs.
3. `PostToolUse(A)` — registered **async** in `ClaudeHookConfigWriter.asyncEvents`, so unordered by construction — is applied *after* `PreToolUse(B)`.
4. `PreToolUse(B)` takes the new guard branch (badge still pending): no `.working`, no tool activity.
5. `PostToolUse(A)` lands and blanket-clears the badge.

Net state: `.waiting` with **no badge**, while the agent is actively executing tool B. It heals only at the next `PreToolUse` or `Stop`. That is the same class of wrong state the guard exists to prevent, inverted, and it gates `IssueTracker.isManagedTerminalIdle` (`IssueTracker.swift:1968`) meanwhile. Parallel tool calls make step 3 a genuine race rather than a theoretical one, and the busy-daemon windows this PR targets are exactly when both requests sit queued together.

The claim in `ClaudeHookConfigWriter.swift:33` that "other inversions self-heal on the next event" doesn't hold for the state this guard leaves behind.

Suggested fix: make the guard conditional on the prompt plausibly belonging to the incoming tool. `StateSignalSource.transition` currently only receives `currentNotificationType: String?` (`StateSignalSource.swift:14`), so this needs widening the protocol to pass the whole `HookNotification?` — then either compare a tool name carried on the notification, or bound the guard with the `timestamp` field it already has (the inversion window is milliseconds; a stale prompt older than a second or two should not suppress the heal).

**🟡 2. The Codex guard ships on an explicitly unverified ordering assumption whose failure mode is permanent, not a race**
`Packages/CrowCodex/Sources/CrowCodex/CodexSignalSource.swift:44-57`

The comment states it outright: *"Codex's PreToolUse/PermissionRequest ordering is assumed to match Claude's (shared hook schema)."* If that assumption is wrong — i.e. Codex emits `PermissionRequest` **before** `PreToolUse` — then every permissioned tool takes the guard branch in the **normal in-order** flow: the badge is preserved, `.working` is never set, and since Codex's `PostToolUse` sets no activity state and Codex has no `Notification` mapping, the card stays "Permission requested / `.waiting`" for the whole tool run until `Stop`. That's an always-on wrong state, not a rare inversion.

The "shared hook schema ⇒ same ordering" reasoning is contradicted inside this repo: `OpenCodeHookConfigWriter.swift:248-258` maps `permission.ask → PermissionRequest` and `tool.execute.before → PreToolUse`, i.e. PermissionRequest-first — which is presumably why OpenCode correctly did *not* get the guard, though the PR doesn't say so. `docs/agent-harness-matrix.md:311-314` already flags Codex hook semantics as a version-pinned empirical re-check target, so the project's own convention is to verify rather than assume here. Either confirm the ordering against a Codex build and record the version, or restrict the guard to Claude in this PR.

**🟡 3. `docs/agent-harness-matrix.md` still asserts the guarantee this PR removes**
`docs/agent-harness-matrix.md:308-310`

> "**Claude:** … `PreToolUse` is intentionally *not* async so it arrives before `PermissionRequest` and keeps state-machine ordering reliable"

The code comments were carefully updated to say accept order no longer implies apply order (`ClaudeHookConfigWriter.swift:24-38`, `CursorHookConfigWriter.swift:61-67`), but the doc CLAUDE.md designates as the harness source of truth was not, so it now overstates the guarantee. The "Hook async delivery" section is the natural place to record the relaxed contract and the daemon-side sequencing follow-up.

### Security Review

**Strengths:**
- Gating `hook-event` local-only (`RPCWebSocketHandler.swift:197-204`) is real hardening, correctly placed behind `localDirect` so the 0600 Unix socket and loopback browsers are unaffected, and covered by `hookEventIsLocalOnly`. It closes both forged per-session hook state and dedup-cap poisoning from an authenticated remote peer.
- The new diagnostic log is bounded (256), deduped per `cwd`, and newline-escaped — an embedded `\n` in a payload `cwd` can't forge a second log record (`EngineRouter.swift:462-480`).
- Not logging the `event_name`-missing branch is the right call; it would be the un-deduped per-request flood the dedup exists to prevent.
- No credential, path, or command-construction surface is touched. `post` writes the same encoder output `send` did.

**Concerns:**
- 🟢 `oneLine()` escapes only `\n`/`\r`. A payload `cwd` containing ANSI/control sequences still reaches the launchd log verbatim, which can mangle a terminal that `cat`s it. Low impact (the source is a local agent process behind a 0600 socket), but stripping C0 controls is a one-liner.
- 🟢 Consider noting the new local-only gate next to `.write("hook-event", cli: "hook-event")` in `ParityLedger.swift:273`, the way other gated methods carry that annotation.

### Additional Notes (non-blocking)

- 🟢 `SocketRoundTripTests.swift:691` still parks its handler on a literal `Task.sleep(60s)`, pinning a `handleClient` GCD worker and its fd for the rest of the run. The CLI-side test introduced the `Latch` type specifically to avoid that (`HookEventTests.swift:107-112`, and its own comment says so) — worth backporting to the IPC test for consistency.
- 🟢 `post` sets no `SO_SNDTIMEO`, so an oversized payload can still block in `write()`. The doc comment (`SocketClient.swift:85-89`) is admirably honest about it and the agent's hook timeout bounds the worst case, but a `setsockopt(SO_SNDTIMEO)` alongside the existing `SO_RCVTIMEO` would close the last residual stall for the price of one line.
- 🟢 Losing the reply also loses back-pressure: nothing now bounds in-flight `hook-event` requests, and `SocketServer.handleClient` blocks a global-queue thread per connection on `semaphore.wait()`. Pre-existing and probably fine at real hook rates, but it's the structural reason the daemon-side follow-up matters.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 3 Yellow, 5 Green] findings.

The transport change itself is sound and I'd take it as-is. What needs another pass is the apply-order guard that compensates for it: it is unconditional on tool identity, which lets the reverse inversion strand a session in `.waiting` (🟡1), it was extended to Codex on a stated-but-unverified ordering assumption that Crow's own OpenCode adapter contradicts (🟡2), and the harness matrix still documents the guarantee that was removed (🟡3).

---

dgershman:
Round 5 addressed in 10293eb0 — and this one reverses direction on the guard, so let me lay out the reasoning.

**🟡 #1 + #2 together.** You're right that the guard is tool-unaware and strands the reverse inversion, and that the Codex mirror rests on an unverified ordering assumption. Digging into the fix you suggested (widen the protocol, match on tool identity or bound by timestamp), I hit a wall the review's own "verify, don't assume" bar makes decisive:

- **Tool identity** needs the pending prompt's tool. `HookNotification` doesn't carry one, and I can find nothing in-repo confirming Claude's `PermissionRequest` payload includes `tool_name` — so I'd be adding a cross-package protocol change (all six `StateSignalSource`s) that silently degrades to "no guard" if the assumption is wrong. That's the same unverified-assumption problem you flag for Codex in #2, just relocated.
- **Timestamp window** avoids the payload assumption but is only heuristic: the reverse case (approve → tool runs → PostToolUse races PreToolUse) can all happen inside a sub-second window under exactly the busy load this PR targets, so a freshness bound still false-fires.
- **Codex (#2)** is worse: I can't confirm its PreToolUse/PermissionRequest ordering, and if it's PermissionRequest-first (like OpenCode) the guard is always-on-wrong in the *normal* flow, not a rare race.

So every client-side form either assumes something I can't verify or stays a heuristic that trades one wrong-state for another. Rather than layer that on, I **removed the guard from both Claude and Codex**. The inversion is real but transient (a few ms, self-corrects on the next state-changing event), and the ordering-correct fix is the per-session server-side sequencing already scoped as the daemon-side follow-up. I documented both inversion directions honestly — no more "self-heals on the next event" overclaim — in `ClaudeHookConfigWriter` and, per **#3**, in `docs/agent-harness-matrix.md`'s "Hook async delivery" section, which did still assert the removed guarantee.

If you'd rather I build the tool-aware guard now — say, if you can confirm from a Claude/Codex build that `PermissionRequest` carries `tool_name` and Codex is PreToolUse-first — I'm happy to do the protocol widening in this PR instead of deferring. I just didn't want to ship it on an assumption I couldn't stand behind.

**🟢 control-char log injection.** `oneLine()` now escapes every C0 control (and DEL) as `\xNN`, not just `\n`/`\r`, so an ESC/ANSI sequence in a payload `cwd` can't mangle a terminal that `cat`s the launchd log.

**🟢 ParityLedger annotation.** Noted the local-only gate next to `.write("hook-event", …)`.

**🟢 non-blocking notes** (IPC-test `Latch` backport, `SO_SNDTIMEO`, back-pressure loss): acknowledged for the daemon-side follow-up. I left the IPC test's `Task.sleep` idiom as-is since it matches that file and you flagged it as no-objection; on `SO_SNDTIMEO` my earlier reasoning still holds (a write timeout would surface as a non-zero `crow hook-event` exit — the hook-error noise this PR removes — unless swallowed, which reintroduces silent loss), so I kept the honest doc + the agent's own timeout as the bound.

**Verification:** `swift build` clean; CrowIPC 39/39, CrowCLI 342/342, CrowClaude 77/77, CrowCodex 70/70, CrowDaemon (LocalOnly 19/19), CrowEngine 571/571, CrowCore 520/520.

dhilgaertner (changes requested):
## Code & Security Review

The core change is right and well-argued: the bimodal fast-or-5s latency profile is the signature of a block, and `SocketClient.post` removes the only wait that could produce it. The `connectAndWrite` extraction is clean — the `keepOpen` flag closes the fd on every throw path after `socket()`, so neither `send` nor `post` can leak a descriptor. Data written into an AF_UNIX stream socket survives the sender's `close()`, and `postReturnsWithoutWaitingForHandler` pins that empirically rather than by assertion. Gating `hook-event` local-only on `/rpc` is a genuine tightening, not a rubber stamp.

Three things to fix before merge, all in the "the code is right but the comments overstate the case" family — which matters here because two of them are the written justification for shipping an accepted regression.

### Critical Issues

None.

### Security Review

**Strengths:**

- `RPCWebSocketHandler.swift:197-204` — `hook-event` gated local-only, pinned by `DaemonSecurityTests.hookEventIsLocalOnly`. I verified no caller loses a surface: `hook-event` appears nowhere under `crow-desktop/`, and there is no exhaustive "these stay un-gated" list test that this contradicts. The `ParityLedger` comment records *why*, which is the right place for it.
- `EngineRouter.swift:353-365` — treating payload `cwd` as untrusted and escaping every C0 control plus DEL (not just `\n`/`\r`) closes both log-record forgery and ANSI mangling of a `cat`'d launchd log. Good instinct.
- `EngineRouter.swift:87` — the dedup cap refuses new keys rather than clearing, so it can't oscillate and re-log the whole set each cycle.
- `post` never reads a reply, so the fire-and-forget path has no response-parsing surface at all. The 1 MB / 30s limits stay on `send`, where they still apply.
- No credentials, paths, or session secrets cross the new code path.

**Concerns:**

- **Yellow — unbounded log line length (`EngineRouter.swift:66-100`).** The dedup bounds the *count* of drop records at 256 but not their *size*. `eventName` and `cwd` are caller-supplied and bounded only by the 1 MB message limit, so the unrotated-launchd-log flood this function exists to prevent can still be written as 256 multi-hundred-KB lines. `oneLine` already treats both as untrusted; length is the same class of hardening and the same one-liner — `String(s.prefix(200))` inside `oneLine` finishes the job.

### Code Quality

- **Yellow — "self-corrects on the next state-changing event" is not true for the case it names (`ClaudeHookConfigWriter.swift:44-45`, `docs/agent-harness-matrix.md:328`).** When `PreToolUse` is applied after `PermissionRequest`, `ClaudeHookSignalSource.swift:30-44` blanket-clears the notification and sets `.working`. The agent is then *parked at the prompt* — no further hook event fires until the user answers, so there is no "next state-changing event" to correct it. The only backstop is `Notification` with `notification_type == "permission_prompt"` (`ClaudeHookSignalSource.swift:61-65`), and that field is read verbatim from the agent's payload (`EngineRouter.swift:1402`), never synthesized. If Claude's `Notification` payload doesn't carry it, the badge stays wiped and the card reads `.working` for the entire duration of the prompt — i.e. Crow silently never raises "agent waiting", which is the app's core signal.

  Scoping the daemon-side sequencing fix out of this PR is defensible; resting that decision on a self-correction that may not exist is not. Either verify the `Notification` backstop and name it explicitly in the comment (and note that it is Claude-specific — the parallel note at `CursorHookConfigWriter.swift:61-70` makes the same reordering claim with no equivalent), or drop the "self-corrects" wording and state the true worst case.

- **Yellow — the description of what the reply was used for is wrong (`HookEventCommand.swift:56`, echoed in `Helpers.swift:24-29`, `SocketClient.swift:80-82`, and the PR body).** "That reply was only ever used to print RPC errors" doesn't hold: `rpc()` throws `ValidationError` whenever `response.error != nil` (`Helpers.swift:16-18`), so the removed `if result["error"] != nil { printJSON(result) }` branch was **dead code** — `result` never contains an `error` key. What actually changed is that an RPC error no longer makes `crow hook-event` **exit non-zero with a stderr message**.

  That change is good — it silences the `session_id required or resolvable from payload cwd` noise for global-hook agents (#897), and it removes a path where a JSON error object could land on a `PreToolUse` hook's stdout. But it is the one part of the old error surface that was load-bearing, and three doc comments now assert it never existed. Please state it accurately; it strengthens the PR rather than weakening it.

**Green (consider):**

- `EngineRouter.swift:79` — `cwd ?? "none"` collides with a literal relative cwd of `none`. Harmless today; a sentinel that can't be a path (or an enum key) removes the ambiguity for free.
- `EngineRouter.swift:359-361` — `\x` escapes aren't zero-padded, so control char `0x09` followed by literal `12` renders as `\x912`. `String(format: "\\x%02x", c.value)` fixes it.
- `loggedUnresolvedHookDrops` is never reset for the daemon's lifetime, so a drop that gets fixed (worktree registered) and later regresses is silent. Fine for a diagnostic — worth one clause in the comment.
- `CrowGrok` / `CrowOpenCode` / `CrowCodex` hook writers didn't get the #903 apply-order note that Claude, Cursor, and Antigravity did. Nothing is *wrong* — their `asyncEvents` are empty — but the harness-matrix caveat is generic and a reader landing in `GrokHookConfigWriter` won't find it.
- `EngineRouter.swift:1276-1283` — the "deliberately not logged" comment on the missing-`event_name` branch is a good call and well justified. No change wanted; noting it so it isn't "fixed" later.

### Verification

| Check | Result |
|---|---|
| `swift build` | clean (14.8s) |
| `swift test --package-path Packages/CrowIPC` | 39/39 pass (incl. both new `post` tests) |
| `swift test --package-path Packages/CrowCLI` | 342/342 pass (incl. `HookEventForwarding` serialized suite) |
| `swift test --package-path Packages/CrowEngine` | 571/571 pass |
| `swift test --package-path Packages/CrowDaemon` | 306/306 pass (incl. `hookEventIsLocalOnly`) |
| `scripts/check-cli-parity.sh` | 94 methods match the ledger |
| `make test` (full sweep) | all packages pass **except** below |

One failure, **pre-existing and unrelated**: `CrowTerminal / TmuxBackend integration / retryReadinessEmitsTimedOutWhenSentinelMissing` (`TmuxBackendTests.swift:395-396`) — reproduces in isolation. `CrowTerminal` depends only on `CrowCore`, and this PR's sole `CrowCore` change is a comment in `ParityLedger`, so it cannot be caused by this branch. Flagging it so it isn't mistaken for fallout.

Also confirmed: the `.serialized` suite is sufficient — `CROW_SOCKET` is written by exactly those two tests process-wide, so no other suite can race their `setenv`.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 3 Yellow, 5 Green] findings.

None of the three are hard to address, and none touch the socket logic, which is the part that had to be right and is. The regression the PR knowingly accepts may well be the correct trade; it just needs its justification to survive contact with the state machine it's about.

---

dgershman:
Round 6 addressed in d5f7b599 — all three were the comments overstating the code, and you're right on each.

**🟡 #3 — "the reply was only used to print RPC errors" was wrong.** Confirmed against `rpc()`: it throws `ValidationError` on `response.error` and returns only the success `result`, which never has an `error` key — so the removed `printJSON(result)` branch was dead code, and what actually changed is that a daemon RPC error no longer surfaces as a non-zero `crow hook-event` exit + stderr. Rewrote `forwardHookEvent`'s comment to say that (and why losing it is a win — the #897 noise, error text off the hook output), and softened `post`'s "a response it wouldn't use anyway."

**🟡 #2 — "self-corrects on the next event" is false for the case it names.** You're right — PreToolUse-after-PermissionRequest wipes the badge and sets `.working` while the agent is *parked*, so nothing else fires until the user answers; the only backstop is a separate `Notification(permission_prompt)`, read verbatim from the payload and never synthesized. I can't verify that a given Claude build emits it, so I dropped the self-correction claim and stated the true worst case (card reads `.working` for the whole prompt, "waiting" never raised), distinguished it from the reverse `.waiting`-while-working inversion that *does* self-heal, and marked it Claude-specific — Cursor keeps a `PermissionRequest` case for parity but doesn't emit it. Updated `ClaudeHookConfigWriter`, the harness-matrix caveat, and the Cursor note; also added the pointer to the Grok/Codex/OpenCode writers that lacked it (OpenCode is `permission.ask`-before-`tool.execute.before`, i.e. PermissionRequest-first, so it self-heals — noted inline).

**🟡 #1 — unbounded line length.** Capped `eventName`/`cwd` to 200 chars in `oneLine`, so the count cap can't be sidestepped as a few hundred-KB lines.

**🟢 nits:** dedup key now distinguishes a `nil` cwd from a literal `none` path (`cwd:` prefix vs `<nil>`); `\x` escapes are zero-padded via `String(format:)`; the never-reset lifetime of the dedup set is documented. Left the "deliberately not logged" `event_name` comment as-is per your note.

On the **pre-existing `TmuxBackend` failure** — agreed it's unrelated (this branch's only `CrowCore` change is a `ParityLedger` comment); thanks for isolating it so it isn't mistaken for fallout.

**Verification:** `swift build` clean; CrowIPC 39/39, CrowCLI 342/342, CrowClaude 77/77, CrowCodex 70/70, CrowCursor 67/67, CrowGrok 51/51, CrowOpenCode 54/54, CrowEngine hook-event resolution 4/4.

dhilgaertner (approved):
## Code & Security Review

Sixth pass. The client-side apply-order guard is gone as agreed in round five, and `d5f7b59` fixed the last inconsistency — the docs now say the permission-badge inversion does **not** self-correct, which matches what `ClaudeHookSignalSource`'s `blanketClear` actually does. Nothing left that blocks.

### Critical Issues

None.

### Security Review

**Strengths:**

- **`hook-event` is now local-only** (`RPCWebSocketHandler.swift:197`, test at `DaemonSecurityTests.swift:816`). This is a net *reduction* in remote surface, not just a fix for something the PR introduced: the method mutates per-session hook state and, as of this PR, feeds a capped in-memory dedup set. Gating it loopback-only means only a caller with access to the 0600 socket in the 0700 dir can reach either. Annotated in `ParityLedger.swift:272` so the ledger explains the asymmetry rather than looking like an oversight. Confirmed nothing in the web assets calls `hook-event`, and the CLI always goes over the Unix socket (`rpcNotify` → `SocketClient()`), so nothing legitimate is cut off.
- **Log-injection hardening is thorough** (`EngineRouter.swift:87-99`). `cwd`/`eventName` are attacker-adjacent (they arrive via the local agent's payload), and `oneLine` escapes every C0 control plus DEL, not just `\n`/`\r` — so an embedded ESC can't mangle a terminal that `cat`s the launchd log, and a newline can't forge a record.
- **The dedup set is bounded two ways** — 256 keys, refusing new entries rather than clearing (clearing would re-log every known key next cycle, which is the flood in a different costume). Distinguishing `nil` cwd (`<nil>`) from a literal `"none"` (`cwd:none`) avoids a key collision that would have suppressed a real drop.
- **Fire-and-forget removes an error channel, not a check.** The old `rpc()` reply was consumed only to print; nothing authenticated or authorized on it. Losing it is strictly less output, not less validation. And `post` is quieter in the failure case that matters: a daemon that dies after `accept` but before replying used to surface a decode failure as agent-visible "hook error" noise; now it's a clean no-op.

**Concerns:** none.

### Code Quality

The `connectAndWrite` extraction is correct where it's easy to get wrong. The `keepOpen`/`defer` inversion hands the fd back on success and closes it on every throw path, so neither caller leaks a descriptor. Moving `SO_RCVTIMEO` from before the write to just before the read in `send` is a no-op by construction — it only governs `read`/`recv`. And the fire-and-forget claim holds at the syscall level: on an `AF_UNIX` `SOCK_STREAM`, `write()` has already copied into the peer's receive buffer by the time it returns, so `close()` just queues the EOF behind data the server will still read; the server's reply to the closed fd hits `EPIPE` and `writeResponse` returns silently under the process-wide `SIG_IGN`. `postReturnsWithoutWaitingForHandler` pins that down empirically by asserting *both* halves — returns early **and** the daemon still received it — so it can't pass by silently no-op'ing.

Verified locally: `swift build` clean (the two `no calls to throwing functions` warnings in `RPCWebSocketHandler.swift:59` are pre-existing on `main`, untouched here). All packages pass — CrowIPC 39/39, CrowCLI 342/342, CrowDaemon 306/306, CrowEngine 571/571, CrowCore 520/520, CrowClaude 77, CrowCursor 67, CrowCodex 70, CrowGrok 51, CrowOpenCode 54, CrowAntigravity 39.

**Green — consider, none blocking:**

1. **`oneLine`'s length cap counts graphemes, not scalars** (`EngineRouter.swift:88`). `String(s.prefix(200))` takes 200 *Characters*; a single grapheme cluster is unbounded in length (arbitrarily many combining marks), so one pathological `cwd` still passes the cap as one Character. The comment claims it "bounds record *size*" — true for every real path, not true in general. `s.unicodeScalars.prefix(200)` would make the comment exact. Zero practical impact (local-only, diagnostic), purely comment-vs-code precision.

2. **The >1 MB drop is now silent on both ends.** `SocketServer.handleClient` answers an oversized request with `"Message too large"` and returns; `post` never reads it, and CrowIPC has no logger, so that event vanishes with no trace anywhere. This is the same reasoning that justified `logUnresolvedHookDropOnce` ("since the client no longer surfaces this, log it") applied to a path that didn't get the treatment — and unlike the `event_name required` case, it isn't a caller bug: a `Write`/`Edit` `PreToolUse` payload carrying a big file body is a legitimate event. It self-heals on the next hook, so it's observability rather than correctness, and the fix means threading a log seam into CrowIPC — reasonable to fold into the daemon-side follow-up rather than reopen this PR.

3. **The CrowIPC twin still parks on a real 60 s sleep** (`SocketRoundTripTests.swift:266`). The CrowCLI version was deliberately rewritten to a poll-latch with the comment "it doesn't pin a socket worker for the rest of the run" — the same argument applies here, where the handler's `Task` and its blocked `handleClient` GCD thread outlive the test by a minute. Harmless at current suite size; mirroring the `Latch` would make the two tests consistent about it.

4. **`loggedUnresolvedHookDrops` is file-scope, not router-scope** (`EngineRouter.swift:73`). Every `makeEngineRouter` in a process shares one set. Correct for the daemon (one router, one lifetime, and the "never reset" behavior is deliberate), but it means the cap and the dedup state leak across routers if a test ever builds two and asserts on logging.

5. **Daemon-side backpressure shifts with this change.** `handleClient` blocks a global-queue thread per connection in `semaphore.wait()` behind `listen(backlog: 5)`. Client-side blocking used to throttle how fast a stalled MainActor could accumulate those; without it, a burst during a stall piles up more concurrent handlers. Pre-existing shape, mildly amplified — worth carrying into the same daemon-side follow-up as the sequencing work, since both are symptoms of "one MainActor, N fanned-out connections".

**On the deferred apply-order limitation:** re-reviewed and still agree it shouldn't block. The `PreToolUse`-after-`PermissionRequest` case is real and genuinely doesn't self-heal, but the two events are separated by a full `crow` process spawn (tens of ms) against a FIFO main queue, so an inversion needs cooperative-pool starvation to beat that gap. Weighed against a *measured* 11% of `PreToolUse:Bash` events hitting the 5 s timeout on `main`, shipping this and fixing ordering properly server-side is the right order of operations. The docs are honest about the failure mode, including that `notification_type` is read verbatim and never synthesized — I confirmed nothing in-repo produces it, so the backstop caveat is stated accurately rather than optimistically.

### Summary Table

---

PR #908, opened 2026-07-29
Boards empty-state: drop the stale 'desktop app required' hint (CROW-907)

### LINKED ISSUE #907: Boards empty-state always says 'Boards require the Crow desktop app to be running' — even when it is running and the board is just empty
The Reviews page (and Tickets / Allowlist) shows **"Boards require the Crow desktop app to be running."** even when the desktop app *is* running and the board simply has no items. It reads as an error/warning when it's really just an empty list.

## Root cause
`boardEmpty(msg)` appends that hint **unconditionally**, after the context message:
```js
// app.js:2865-2869
function boardEmpty(msg) {
  const wrap = el('div', 'board-empty');
  wrap.appendChild(el('div', null, msg));
  wrap.appendChild(el('div', 'board-empty-hint', 'Boards require the Crow desktop app to be running.'));
  return wrap;
}
```
Every empty board routes through it with only a context string:
- Reviews: `boardEmpty('No review requests')` / `'No matching reviews'` — `app.js:3370`
- Tickets: `boardEmpty('No tickets in this view')` / `'No matching tickets'` — `app.js:3056`
- Allowlist: `boardEmpty('No allowlist entries')` / `'No matching entries'` — `app.js:3517`

So with zero pending reviews, `renderReviewBoard` hits `boardEmpty('No review requests')` and the desktop-app hint is tacked on regardless of whether the app is actually running. It conflates **"no items"** with **"app not running."**

## Why it's *sometimes* accurate
Board data is app-sourced: the comments at `app.js:545-549` note board data is "forwarded to the desktop app, so it's empty when the desktop app isn't running." So in a headless `crowd`-only setup the hint is correct; in the common **app-running-but-empty** case it's misleading.

## Fix directions
1. **Gate the hint on a real "desktop app connected/running" signal.** Only append it when the app is actually down; when the app is up and the board is just empty, show only the context message ("No review requests"). If no app-connectivity flag is exposed to the web UI yet, add a lightweight one — the daemon already knows whether the app has registered / pushed board data, so surface it via an existing status/live RPC (`RPCHandlers.swift`) and read it in `boardEmpty` (or pass an `appConnected` arg from each caller).
2. **Reconsider the claim post native→web migration (ADR 0010).** If `crowd` can serve boards without the desktop app at all, the hint is stale and should be reworded or dropped.

## Ambiguity to resolve during implementation
Confirm whether `list-reviews` / `list-tickets` genuinely still require the desktop app today, or whether `crowd` populates them — that decides between "gate the hint" (option 1) and "remove it" (option 2).

## Files
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` — `boardEmpty` (`:2865`) and callers (`:3056`, `:3370`, `:3517`)
- App-connectivity signal source (`RPCHandlers.swift` / the live/status payload) if one needs adding

dgershman:
Closes #907

## Problem
Every empty board — Reviews, Tickets, Allowlist — routed through `boardEmpty(msg)`, which **unconditionally** appended:

> Boards require the Crow desktop app to be running.

So with zero pending reviews (app running, board simply empty), the page read as an error/warning when it was just an empty list. It conflated *"no items"* with *"app not running."*

## Why the hint is wrong (ambiguity resolved → option 2)
The ticket asked whether `crowd` still needs the desktop app for boards. It doesn't:

- `list-tickets` / `list-reviews` / `list-allowlist` answer **locally off the daemon's own `IssueTracker`/`AllowListService`** (`RPCHandlers.swift:648-749`, CROW-581 Milestone C) — "so the boards work with the app down." Production `crowd` always constructs the tracker (`CrowDaemon.swift:132`) and passes it in (`:417`); the `tracker == nil` empty-board path is tests/stripped-builds only.
- `boardEmpty` only renders after the `list-*` RPC **succeeded** and returned zero items (a failed read bails at `app.js:2416` without rendering). So reaching it means the daemon is up and the board is genuinely empty.

The hint was therefore stale post native→web migration (ADR 0010) and unconditionally wrong at the point it rendered.

## Change
- `boardEmpty` no longer appends the hint — empty boards show only their context message (`No review requests`, `No tickets in this view`, `No allowlist entries`).
- Corrected the matching stale comment on the `boardData` declaration ("forwarded to the desktop app…").
- Removed the now-unused `.board-empty-hint` CSS rule.

Web-asset-only change; `node --check` passes.

## Test plan
- Open Reviews / Tickets / Allowlist with the app running and no items → shows only the context message, no "desktop app required" hint.
- Filter to zero matches → shows "No matching …" with no hint.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -639,7 +639,6 @@ html, body {
 }
 
 .board-empty { color: var(--text-muted); text-align: center; padding: 48px 12px; }
-.board-empty-hint { font-size: 11px; margin-top: 8px; opacity: 0.7; }
 
 /* ---- Sidebar: brandmark, Tickets card, nav pills ---- */
 #brand-img { display: block; width: 120px; height: 120px; object-fit: contain; margin: 6px auto 12px; opacity: 0.7; }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -2417,6 +2417,17 @@ async function refreshBoard(key) {
     // A failed read leaves us with no fresh word on the daemon's in-flight
     // flag, and a stale `true` never self-clears (CROW-771).
     if (key === 'tickets') clearDaemonRefreshFlag();
+    // The allowlist never polls or caches (it's manual-refresh-only), so a failed
+    // FIRST read would strand its cold-open "Scanning…" state forever —
+    // renderAllowlist keys that on `boardData.allowlist == null`. Seed an empty,
+    // settled snapshot so it lands on "No allowlist entries" like the pre-CROW-907
+    // board did (the disconnected banner is the real daemon-down signal); a later
+    // manual Refresh still replaces it. A failed LATER read keeps the last-good
+    // snapshot (d already non-null), so only the null case needs this.
+    if (key === 'allowlist' && !boardData.allowlist) {
+      boardData.allowlist = { entries: [], loading: false };
+      if (selectedBoard === 'allowlist') renderBoard();
+    }
     return;
   }
   const changed = JSON.stringify(boardData[key]) !== JSON.stringify(data);
@@ -3477,10 +3488,13 @@ function renderAllowlist(root) {
   const d = boardData.allowlist;
   // The allowlist is never cached (persistSidebarCache stores only tickets/
   // reviews) and never polled — it's manual-refresh-only, auto-scanned once on
-  // open (CROW-593). So a cold open renders before the first `list-allowlist`
-  // read lands (d == null), and `d.loading` marks a scan the daemon reports as
-  // still running. In both, the board isn't known-empty yet — distinguish that
-  // from a genuinely empty allowlist below so it doesn't flash "No entries".
+  // open (CROW-593). So `d == null` means no read has come back yet: the board
+  // isn't known-empty, and rendering "No entries" would flash a false empty
+  // during the cold-open scan. refreshBoard seeds a settled snapshot even on a
+  // failed read (see its catch), so `d == null` can't mean "read failed" and
+  // strand this forever. `d.loading` is defensive only — AllowListService.scan()
+  // is synchronous, so `list-allowlist` never lands mid-scan today; it's the
+  // right shape if that ever goes async, and costs nothing.
   const scanning = !d || d.loading;
   let entries = ((d && d.entries) || []).slice();
   if (allowlistHideGlobal) entries = entries.filter((e) => !e.is_global);
@@ -3528,8 +3542,10 @@ function renderAllowlist(root) {
   root.appendChild(boardFilterInput('allow-filter', allowlistFilter, 'Filter patterns…', (v) => { allowlistFilter = v; }));
 
   if (!entries.length) {
-    // Unfiltered + not yet loaded ⇒ say so rather than claim "empty" mid-scan.
-    if (scanning && !filter) { root.appendChild(boardEmpty('Scanning allowlist…')); return; }
+    // Not loaded yet ⇒ say we're scanning rather than assert empty / no-match —
+    // honest whether or not a filter is active (allowlistFilter survives board
+    // switches, so a cold open can carry one).
+    if (scanning) { root.appendChild(boardEmpty('Scanning allowlist…')); return; }
     root.appendChild(boardEmpty(filter ? 'No matching entries' : 'No allowlist entries'));
     return;
   }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -2862,15 +2862,16 @@ function labelPills(labels, maxVisible) {
   return wrap;
 }
 
-// An empty board is just an empty list, not an error. `crowd` serves the boards
-// off its own IssueTracker/AllowListService (CROW-581 M-C), so we only reach here
-// when the `list-*` RPC succeeded and returned zero items — the daemon is up and
-// the board is genuinely empty. No "desktop app required" hint: it's stale post
-// native→web migration (ADR 0010) and read as a false error/warning (CROW-907).
+// An empty board is just an empty list, not an error — `crowd` serves the boards
+// off its own IssueTracker/AllowListService (CROW-581 M-C), so a "requires the
+// Crow desktop app" hint would be stale post native→web migration (ADR 0010) and
+// read as a false error/warning (CROW-907). Just render the caller's context
+// message ("No review requests", …). NOTE: this is reachable before the first
+// read lands (the pre-refresh paint at selectBoard) — a caller that must not
+// conflate "not loaded yet" with "empty" gates the message itself, as
+// renderAllowlist does with its scanning state.
 function boardEmpty(msg) {
-  const wrap = el('div', 'board-empty');
-  wrap.appendChild(el('div', null, msg));
-  return wrap;
+  return el('div', 'board-empty', msg);
 }
 
 // A spawning action (Start Working / Start Review): disable the button, let the
@@ -3474,6 +3475,13 @@ function reviewCard(r) {
 // -- Allowlist --
 function renderAllowlist(root) {
   const d = boardData.allowlist;
+  // The allowlist is never cached (persistSidebarCache stores only tickets/
+  // reviews) and never polled — it's manual-refresh-only, auto-scanned once on
+  // open (CROW-593). So a cold open renders before the first `list-allowlist`
+  // read lands (d == null), and `d.loading` marks a scan the daemon reports as
+  // still running. In both, the board isn't known-empty yet — distinguish that
+  // from a genuinely empty allowlist below so it doesn't flash "No entries".
+  const scanning = !d || d.loading;
   let entries = ((d && d.entries) || []).slice();
   if (allowlistHideGlobal) entries = entries.filter((e) => !e.is_global);
   // #701: case-insensitive substring filter on pattern (mirrors desktop's
@@ -3490,7 +3498,9 @@ function renderAllowlist(root) {
   for (const p of [...allowlistSelection]) if (!visiblePromotable.has(p)) allowlistSelection.delete(p);
 
   const head = el('div', 'board-head');
-  head.appendChild(el('div', 'board-title', 'Allowlist'));
+  const title = el('div', 'board-title', 'Allowlist');
+  if (scanning) title.appendChild(el('span', 'action-spinner'));
+  head.appendChild(title);
   const hide = el('button', 'action-btn' + (allowlistHideGlobal ? ' active' : ''), allowlistHideGlobal ? 'Show Global' : 'Hide Global');
   hide.onclick = () => { allowlistHideGlobal = !allowlistHideGlobal; renderBoard(); };
   head.appendChild(hide);
@@ -3518,6 +3528,8 @@ function renderAllowlist(root) {
   root.appendChild(boardFilterInput('allow-filter', allowlistFilter, 'Filter patterns…', (v) => { allowlistFilter = v; }));
 
   if (!entries.length) {
+    // Unfiltered + not yet loaded ⇒ say so rather than claim "empty" mid-scan.
+    if (scanning && !filter) { root.appendChild(boardEmpty('Scanning allowlist…')); return; }
     root.appendChild(boardEmpty(filter ? 'No matching entries' : 'No allowlist entries'));
     return;
   }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -546,8 +546,8 @@ let activeTerminal = null; // { id, name, window }
 let liveById = {};
 
 // Boards (Ticket Board / Reviews / Allowlist), mirroring the desktop's
-// full-pane boards. Data is forwarded to the desktop app, so it's empty when
-// the app isn't running.
+// full-pane boards. Served by `crowd` off its own IssueTracker/AllowListService
+// (CROW-581 M-C), so they populate whether or not the desktop app is running.
 let selectedBoard = null; // 'tickets' | 'reviews' | 'allowlist' | 'scorecard' | null
 const boardData = { tickets: null, reviews: null, allowlist: null, scorecard: null };
 
@@ -2862,10 +2862,14 @@ function labelPills(labels, maxVisible) {
   return wrap;
 }
 
+// An empty board is just an empty list, not an error. `crowd` serves the boards
+// off its own IssueTracker/AllowListService (CROW-581 M-C), so we only reach here
+// when the `list-*` RPC succeeded and returned zero items — the daemon is up and
+// the board is genuinely empty. No "desktop app required" hint: it's stale post
+// native→web migration (ADR 0010) and read as a false error/warning (CROW-907).
 function boardEmpty(msg) {
   const wrap = el('div', 'board-empty');
   wrap.appendChild(el('div', null, msg));
-  wrap.appendChild(el('div', 'board-empty-hint', 'Boards require the Crow desktop app to be running.'));
   return wrap;
 }
 
--- a/Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift
+++ b/Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift
@@ -274,4 +274,27 @@ import Testing
             source.contains("navigator.clipboard && navigator.clipboard.readText"),
             "\(asset) must guard readText — it has no fallback to reach for")
     }
+
+    /// CROW-907: an empty board is an empty list, not an error. `boardEmpty` once
+    /// appended "Boards require the Crow desktop app to be running." to every
+    /// empty board — false since `crowd` serves the boards off its own
+    /// IssueTracker/AllowListService (CROW-581 M-C, ADR 0010). Pin the hint out of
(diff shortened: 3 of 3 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Small, well-motivated change. I verified the PR's central premise independently: `RPCHandlers.swift:648-749` answers `list-tickets` / `list-reviews` / `list-allowlist` locally off the daemon's own `IssueTracker`/`AllowListService`, so the "requires the Crow desktop app" hint was indeed stale post-ADR-0010. Removing it is right.

### Critical Issues
None.

### Security Review
**Strengths:**
- Pure deletion of a static string + an unused CSS rule. No user input, no new data flow, no auth/authz surface.
- `boardEmpty` builds nodes via `el()` (`app.js:771-776`), which sets `textContent` — no `innerHTML`, so no XSS surface touched.
- No secrets, no network, no config changes.

**Concerns:** None.

### Code Quality

**🟡 Yellow — the new comment asserts an invariant the code doesn't hold** (`app.js:2865-2869`)

> we only reach here when the `list-*` RPC succeeded and returned zero items — the daemon is up and the board is genuinely empty

`boardEmpty` is reachable in three states that comment says are impossible:

1. **Cold first paint.** `selectBoard` calls `renderBoard()` at `app.js:2400` *before* dispatching the refresh — the comment on that very line says "instant paint (may be stale/empty)".
2. **In-flight scan.** Allowlist is never localStorage-hydrated: `persistSidebarCache` (`app.js:580-591`) caches only `tickets`/`reviews`. So opening Allowlist *always* renders "No allowlist entries" while `refreshAllowlist()` (kicked at `app.js:2404`, CROW-593) is still scanning. `renderAllowlist` (`app.js:3475-3527`) never reads the `loading` flag the RPC does return (`RPCHandlers.swift:744`), so there's no spinner either — unlike the ticket board, which shows one at `app.js:2984`.
3. **Failed read.** `refreshBoard` returns early on a thrown `rpc()` (`app.js:2416-2421`), leaving the pre-RPC empty paint on screen with no re-render.

This isn't a behavioral regression — the removed text was wrong in those states too — but the comment is the load-bearing justification for the deletion, and codifying "empty ⇒ daemon up ⇒ genuinely empty" invites a future maintainer to skip a loading state on exactly the board that most needs one. Please either soften it to describe the successful-read case only, or (better) have `renderAllowlist` read `d.loading` and render a scanning state so the allowlist stops claiming "empty" mid-scan.

**🟢 Green — redundant wrapper.** With the hint gone, `boardEmpty` wraps a single text node. `el()` takes text directly, so `return el('div', 'board-empty', msg);` is equivalent and one line shorter.

**🟢 Green — no regression guard.** The repo already asserts web-asset invariants in `CrowDaemonTests/WebTerminalAssetTests.swift`; a one-line assertion that `app.js` no longer contains the hint string would fit house style. Optional for a pure deletion.

### Verification performed
- `rg` repo-wide: no remaining references to `.board-empty-hint` or "desktop app to be running". Clean removal.
- Only one copy of `app.js` exists — no duplicated bundle to keep in sync.
- `node --check app.js` passes; `swift build` completes (the one warning, `RPCWebSocketHandler.swift:59`, is pre-existing and unrelated).
- Confirmed removing the hint does **not** drop the only daemon-down signal: the status bar flips to `disconnected` at `app.js:1024` when the WS closes.
- All three call sites (`app.js:3060`, `:3374`, `:3521`) keep their context message, including the filtered "No matching …" variants.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The behavior change is correct and I'd happily land it; only the over-claiming comment (and ideally the allowlist loading state it papers over) needs a pass.

---

dgershman:
Thanks for the careful review — addressed in 11d14f54.

**🟡 Yellow — over-claiming comment + allowlist loading state.** You're right on both counts, and I took the "better" path for the allowlist:
- Softened the `boardEmpty` comment to the accurate rationale (boards are daemon-served, so the app-required hint is stale — ADR 0010 / CROW-581 M-C) and dropped the false "empty ⇒ daemon up ⇒ genuinely empty" invariant. It now explicitly notes the state is reachable pre-first-read and defers the "not-loaded-vs-empty" distinction to callers.
- Gave `renderAllowlist` a real scanning state: `scanning = !d || d.loading` → a title spinner + **"Scanning allowlist…"** instead of "No allowlist entries" while the CROW-593 auto-scan is in flight. `!d` covers the cold-open flash you flagged (the allowlist is never `persistSidebarCache`'d); `d.loading` covers a daemon-reported mid-scan. Since `scan()` is synchronous (`AllowListService.swift:60-90`, sets `isLoadingAllowList` back to `false` before returning), `d.loading` can't stick despite the one-shot refresh — and a persistent failed read is already signalled by the `disconnected` status bar (`app.js:1024`), not this empty state.

**🟢 Green — redundant wrapper.** Collapsed `boardEmpty` to `return el('div', 'board-empty', msg);`.

**🟢 Green — regression guard.** Added `emptyBoardsDropTheStaleDesktopAppHint()` to `WebTerminalAssetTests`, pinning the hint out of the executed source, `boardEmpty`'s single-message render, the allowlist scanning state, and the removed `.board-empty-hint` CSS rule.

`swift test --filter WebTerminalAssetTests` passes (8 tests); `node --check app.js` clean.

dhilgaertner (changes requested):
## Code & Security Review

The premise checks out. I verified the claim independently rather than taking the PR body's word for it:

- `list-allowlist` / `list-tickets` / `list-reviews` answer off the daemon's own services with no forwarding path (`RPCHandlers.swift:648-749`), and production `crowd` always constructs both (`CrowDaemon.swift`), so the `nil` empty-board branch is tests/stripped-builds only.
- No stale copy of the hint or the CSS rule survives anywhere in tracked source (only the test's own assertion strings match).
- `swift test --package-path Packages/CrowDaemon --filter WebTerminalAssetTests` → **8 tests, 0 failures**, including the new `emptyBoardsDropTheStaleDesktopAppHint()`. `node --check app.js` passes.

Deleting the hint is right, and the new test pins it out of the *executed* source (`stripComments`) rather than the raw text, so a comment can't smuggle it back. My concern is entirely with the second commit's scanning state.

### Critical Issues

None.

### Security Review

**Strengths:**
- Web-asset-only change with no new sink: `el()` (`app.js:771-776`) assigns via `textContent`, and every string reaching `boardEmpty` is a static literal. The rewrite actually shrinks the constructed DOM.
- No authn/authz, RPC surface, or file-write path is touched; `promoteAllowlist` and the `refresh-allowlist`/`promote-allowlist` handlers are untouched.
- The removed hint carried no information about host, daemon, or session state, so dropping it leaks nothing and hides nothing.

**Concerns:** none.

### Code Quality

**🟡 Yellow — the allowlist can wedge on "Scanning allowlist…" forever after a failed read** (`app.js:3484`, `app.js:3532`)

`scanning = !d || d.loading` treats `d == null` as "still loading", but `d` is also `null` when the one and only read *failed*. The allowlist has no retry:

- `refreshAllowlist()` (`app.js:3590-3593`) fires exactly one `refreshBoard('allowlist')`, 800 ms after the (swallowed) `refresh-allowlist` call.
- `refreshBoard` returns early on a thrown `rpc` without touching `boardData[key]` (`app.js:2416-2421`), leaving `boardData.allowlist === null`.
- Nothing else ever re-reads it — the `changed` nudge (`app.js:105-106`) and the 4s board poll (`app.js:4577-4579`) both cover only `tickets`/`reviews`, and `persistSidebarCache` doesn't store it. The comment at `app.js:3478-3480` says as much.

So one failed read (a daemon restart, a WS drop, a 10s `rpc` timeout, or `sessionDead` short-circuiting every call at `app.js:80`) leaves an animated spinner and "Scanning allowlist…" on screen **permanently** — while the tickets and reviews boards visibly recover on their own polls. The message asserts an in-flight scan that isn't running, and the only escape is the Refresh button. That's the same failure mode this PR set out to remove: a board state that lies about what the system is doing. Ironically, the old text was closer to correct in exactly this case.

The fix is small — record that a read was *attempted*, e.g. on `refreshBoard`'s error path seed `boardData.allowlist = { entries: [], loading: false }` (and re-render) for `key === 'allowlist'`, or gate `scanning` on a module-level `allowlistReadAttempted` flag set in `refreshAllowlist`'s `finally`. Either way the board settles on "No allowlist entries" instead of spinning forever.

**🟢 Green — `d.loading` is never observable as `true`; the comment claims more than the code delivers** (`app.js:3481`)

The comment states `d.loading` "marks a scan the daemon reports as still running." In practice it can't. `AllowListService.scan()` (`AllowListService.swift:60-90`) is `@MainActor` and fully synchronous: it sets `isLoadingAllowList = true`, does all the file I/O inline, and sets it `false` with no suspension point in between. `refresh-allowlist` runs the whole thing inside a single `await MainActor.run { allowList.scan() }` (`RPCHandlers.swift:917-921`), so no concurrent `list-allowlist` hop (`RPCHandlers.swift:744`) can ever land mid-scan.

Keeping the `|| d.loading` term is fine — it's the correct shape if `scan()` ever goes async, and it costs nothing. But the scanning state is carried **entirely** by `d == null` today, which also means the Yellow above isn't a corner case; it's the only case. Worth softening the comment to say the flag is defensive, in the same spirit as the previous round's over-claiming fix.

**🟢 Green — the pre-load paint still claims "No matching entries" when a filter is active** (`app.js:3532-3533`)

The `!filter` guard means a cold open with a non-empty `allowlistFilter` (a module-level var that survives board switches) shows "No matching entries" before any data has arrived. Self-corrects within one rpc and the title spinner is still up, so it's minor — but the scanning message is the honest answer in both cases, and dropping the `!filter` guard would make it uniform.

**Test coverage:** good. The new test pins all four halves — hint gone from executed source, `boardEmpty` renders only `msg`, `renderAllowlist` has the scanning branch, CSS rule deleted — and the docstring explains why each one exists. Nothing to add beyond a case for whichever shape the Yellow lands on.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The hint removal itself is correct, well-tested, and I'd take it as-is; the request is on the scanning state that rode along with it, which can strand the allowlist board on a spinner that never stops.

---

dgershman:
Good catch — the failed-read wedge is real and exactly the failure mode this PR exists to remove. Fixed in ec32937f.

**🟡 Yellow — allowlist strands on "Scanning allowlist…" after a failed read.** Took your first suggested shape: `refreshBoard`'s catch now seeds a settled `{ entries: [], loading: false }` snapshot for a failed **first** allowlist read (`key === 'allowlist' && !boardData.allowlist`), then re-renders if it's the active board. So `d == null` can now only mean "no read has returned yet" — the board settles on "No allowlist entries" (with the `disconnected` banner as the authoritative daemon-down signal), matching the pre-CROW-907 behavior, instead of spinning forever. A failed *later* read still keeps the last-good snapshot (d already non-null), so only the null case needs the seed.

**🟢 Green — `d.loading` never observable as true.** Agreed — softened the comment to say `d.loading` is defensive only (`scan()` is synchronous, so `list-allowlist` can't land mid-scan today; it's the right shape if that ever goes async). `d == null` carries the whole scanning state, which is why the Yellow was the *only* case, not a corner one.

**🟢 Green — `!filter` guard.** Dropped it. A cold open with an active `allowlistFilter` (module-level, survives board switches) now shows the honest "Scanning allowlist…" in both cases rather than a premature "No matching entries."

**Test.** Extended the drift guard to pin the failed-read seed in `refreshBoard`, so the anti-wedge can't be silently removed.

`swift test --filter WebTerminalAssetTests` → 8 pass; `node --check app.js` clean.

dhilgaertner (approved):
## Code & Security Review

Third round. Both prior Yellows are closed, and I re-verified the fix rather than taking the commit message's word for it — I went looking specifically for a surviving path that could strand the new "Scanning…" state, and could not construct one. Details under *Verification performed*.

### Critical Issues

None.

### Security Review

**Strengths:**
- Web-asset-only. `el()` (`app.js:771-776`) assigns via `textContent`, and every string reaching `boardEmpty` is a static literal — the rewrite shrinks the constructed DOM rather than adding a sink.
- No RPC surface, authn/authz path, or file write is touched. `promoteAllowlist` and the `refresh-allowlist` / `promote-allowlist` handlers are untouched; the seeded snapshot is client-side display state only.
- Removing the hint hides nothing: the daemon-down signal is the status bar, which flips to `disconnected` at `app.js:1024` on WS close and is unaffected by `.board-active`.

**Concerns:** none.

### Code Quality

The load-bearing fix (`app.js:2427-2430`) is right, and it's the minimal shape: seed only on the **null** case, only for `allowlist`, and re-render only when that board is actually open. Seeding a settled `{ entries: [], loading: false }` rather than adding a separate `allowlistReadAttempted` flag keeps one source of truth for "has a read landed", which is what `scanning` already keys on.

Worth noting the precedent this lines up with: `restoreSidebarCache` (`app.js:572-576`) already scrubs `loading` off a cached tickets snapshot for exactly this reason — CROW-771's "a cache written mid-fetch would otherwise boot showing a spinner that only a successful `list-tickets` could clear." Same failure class, same resolution. The allowlist isn't cached at all, so it's immune to the boot-time variant.

**🟢 Green — manual Refresh still has no in-flight signal** (`app.js:3516`, `:3521-3522`)

`scanning` is `!d || d.loading`, so it's true only before the first read lands. Once `d` is non-null it never goes back — and `d.loading` is unobservable (below) — so clicking **Refresh** on a populated allowlist board spins nothing through the 800 ms delay plus the rpc round trip. Tickets gates its spinner on `ticketsRefreshing()` (`:2996`) and Reviews on `busy`, which also disables the button (`:3349`, `:3352`). A module-level `allowlistRefreshing` set in `refreshAllowlist` and cleared in the `refreshBoard` callback would close it in ~3 lines. Squarely outside CROW-907, and the current state is not wrong — just quieter than its two siblings.

**🟢 Green — "costs nothing" is a hair optimistic on the `d.loading` term** (`app.js:3495-3497`)

The reachability claim checks out, and I confirmed it end to end: `AllowListService.scan()` (`AllowListService.swift:60-90`) is `@MainActor` and fully synchronous — `isLoadingAllowList = true`, all file I/O inline, `= false`, no suspension point — and `refresh-allowlist` wraps the whole thing in a single `await MainActor.run { allowList.scan() }` (`RPCHandlers.swift:917-921`), so the `list-allowlist` hop at `RPCHandlers.swift:744` can never observe `true`.

The wrinkle is the "if that ever goes async" half. If `loading: true` ever became observable, it would re-open the exact wedge this commit just closed: nothing re-polls the allowlist (the `changed` nudge covers only tickets/reviews at `:105-106`, as does the 20 s fallback at `:4593-4595`), so a `true` would pin the spinner until the user clicked Refresh. The seeding fix doesn't cover that door — it only handles `d == null`. Worth one clause on that comment noting the flag needs a re-poll to ship alongside it, rather than "costs nothing."

**🟢 Green — one narrow false-empty flash survives** (`app.js:2427-2430`)

The seeding is unconditional on board visibility, so if the user opens Allowlist and switches away inside the 800 ms window while that read fails, the seeded snapshot sticks. The next cold open then renders "No allowlist entries" with no spinner for ~800 ms before real data replaces it. Very narrow (requires a failed read *and* a board switch inside the window), self-corrects in under a second, and is strictly better than the pre-PR behavior — noting it only because it's the one residue of the class of flash the scanning state was added to remove.

**Test coverage:** good, and it pins the right four things — hint gone from the *executed* source (`stripComments`, so a comment can't smuggle it back), `boardEmpty` renders only `msg`, `renderAllowlist` has the scanning branch, and the seeding line that keeps the branch from wedging. The docstring explains why each assertion exists, which is what makes these drift guards maintainable. The literal string matches are brittle to reformatting, but that's the established contract for this suite, not a defect.

### Verification performed

- **Premise, re-checked independently:** `list-allowlist` answers off the daemon's own `AllowListService` with no forwarding path (`RPCHandlers.swift:734-749`), same for `list-tickets` / `list-reviews` (`:655-733`). The `nil` empty-board branch is tests/stripped-builds only. The hint was stale post-ADR-0010.
- **Wedge audit (the point of this round):** enumerated every path that can paint the allowlist with `d == null`. `selectedBoard` is assigned only in `selectBoard()` (`:2387`) and reset to null (`:1728`) — `restoreSidebarCache` (`:564-579`) restores neither `selectedBoard` nor `boardData.allowlist`, so there is no boot-time or deep-link route that paints the board without going through `selectBoard`. `selectBoard('allowlist')` always calls `refreshAllowlist()` (`:2404`), which schedules `refreshBoard('allowlist')` unconditionally — the `refresh-allowlist` failure is swallowed and the `setTimeout` is outside the `try` (`:3606-3609`). The await itself is bounded: `rpc()` has a 10 s timeout (`:87`), pending calls reject on WS close (`:55`), a socket that closes before opening rejects (`:53`), and `sessionDead` fails fast (`:80`). Every terminal state lands in `refreshBoard`'s catch, which seeds. **No surviving path strands "Scanning…".**
- **Escape hatch intact:** the scanning branch returns *after* the head and filter input are appended (`:3538-3548`), so Refresh stays on screen during the scan — which matters, since it's the manual recovery for a board with no poll.
- **`renderSidebar()` correctly omitted** from the seeding path: nothing in the sidebar reads `boardData.allowlist` (only the nav pill at `:1090`, keyed on `selectedBoard`).
- **Prior round's filter Green is closed:** the `scanning` check now precedes the `filter ? … : …` choice (`:3548-3549`), so a carried-over `allowlistFilter` no longer yields "No matching entries" before any data has arrived.
- `rg` repo-wide: no `board-empty-hint` or "desktop app to be running" outside the test's own assertion strings and docstring.
- `node --check app.js` passes. `swift test --package-path Packages/CrowDaemon --filter WebTerminalAssetTests` → **8 tests, 0 failures**, including `emptyBoardsDropTheStaleDesktopAppHint()`.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 3 Green] findings. The stale hint is gone, the ticket's ambiguity is resolved with evidence rather than assertion, and the scanning state that rode along now settles in every failure path I could reach. The three Greens are polish (a missing manual-refresh spinner, one over-confident clause, one sub-second flash in a narrow window) and none need to block the merge.

---

PR #910, opened 2026-07-29
Notification center: bell + unread badge with history and click-through (CROW-909)

### LINKED ISSUE #909: Notification center: bell + unread counter with notification history and click-through to context
Add an in-app **notification center**: a bell icon with an unread counter, a panel listing the history of notifications that fired, where clicking an entry navigates contextually back to what triggered it (the session, review, or board). Today notifications are fire-and-forget (a transient browser/Tauri banner + sound) with **no history** — once the banner dismisses, there's no record of what happened.

## Why this is a clean add
Every notification already funnels through **one chokepoint**:
```js
// app.js:438-441
function emitEvent(event, key, detail) {
  playEventSound(event, key);
  showEventNotification(event, key, detail);
}
```
A history append slots in right here, with the exact `(event, key, detail)` tuple to store. All event types flow through it — client-derived (`detectSessionSounds` `app.js:497-521`, `detectReviewSounds` `app.js:523-531`) and server-pushed automation events (`onServerNotify` `app.js:448-470`, from the `notify` frame `EventHub.swift:66-94`).

## Implementation grounding (pieces to reuse)
**History store**
- Append in `emitEvent` (`app.js:438`). Entry schema: `{ id, event, key, title, body, ts, seen }`.
- Title/body come free from `EVENT_LABEL` (`app.js:179-187`) + `EVENT_DESC` (`app.js:188-200`), enriched exactly as `showEventNotification` already does — `sessionNameFor(key)` (`app.js:362-365`) for session events, repo name for reviews (`app.js:400-406`).
- Persist to `localStorage` (same pattern as the sidebar cache at `app.js:783`); cap to the last ~100.

**Bell + counter placement**
- Put it in the sidebar **nav-pill row / tools stack** (`navPillRow` `app.js:1081-1113`, `sidebarToolsStack` `app.js:1065-1078`) — the only region visible in **all** views (sessions *and* boards). The detail-header actions cluster (`app.js:1914-1965`) is the pixel top-right but is per-session and cleared on boards, so it's a worse home for a global bell.
- Reuse the existing unread-badge style `.pill-badge` (`app.css:748`), mirroring the Reviews pill's `unseen` badge (`app.js:1086-1088`).
- Add a `bell` SVG to the `ICONS` registry (`app.js:1189-1211` — **none exists today**) and render via `icon('bell', 14)` (`app.js:1212`).

**Panel**
- Mirror the anchored popover `openNewManagerMenu(anchorEl)` (`app.js:1155-1182`): `.ctx-menu` shell + `.ctx-item` rows (`app.css:768-781`), outside-click close via `closeContextMenu` (`app.js:1483`). Use two-line rows (title + relative timestamp). Reset unread + mark entries `seen` on open; add a "Clear all".

**Click-through (reuse existing routing)**
The needed mapping already lives in the current `Notification.onclick` (`app.js:421-433`):
- session events (`agentWaiting`, `taskComplete`, `changesRequested`, `checksFailing`, and automation events keyed by a session UUID) → `selectSession(key)` (`app.js:1726-1740`)
- `reviewRequested` (key = review id) → resolve via `boardData.reviews.reviews.find(...)` → `selectSession(review_session_id)`, else `selectBoard('reviews')` (`app.js:427-429`, `selectBoard` `app.js:2386`)
- automation keyed by an issue URL / `"config"` → no in-app target: render as an external link (open the URL) or a non-navigable row.

## Scope / decisions to surface
- **Client-side history (localStorage) is sufficient for v1** — no backend change needed (the `notify` frame already carries `{event, key, title, body}`). Caveat: history is then **per-client** (each web tab / the desktop app keep their own), not synced. Call out whether a server-side/synced, persistent history is wanted — that's a larger follow-up, out of scope here.
- Unread counter increments in `emitEvent`, resets on panel open; persist `seen` so a reload doesn't re-inflate the count.
- Keep the existing transient banner/sound behavior; the center is additive.

## Files
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` — `emitEvent` (`:438`), `ICONS` (`:1189`), `navPillRow`/`sidebarToolsStack` (`:1065-1113`), panel (mirror `:1155`), click routing (reuse `:421-433`, `selectSession` `:1726`, `selectBoard` `:2386`)
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css` — reuse `.pill-badge` (`:748`), `.ctx-menu`/`.ctx-item` (`:768-781`); add bell/panel styles

dgershman:
Closes #909

## What

Adds an in-app **notification center**: a bell with an unread counter in the sidebar tools stack, and an anchored panel listing the history of notifications that fired. Clicking an entry navigates back to its context. Today notifications are fire-and-forget (transient banner + sound, no history) — once the banner dismisses there's no record.

The whole feature slots in at the one chokepoint every event already funnels through — `emitEvent(event, key, detail)` (`app.js`) — so client-derived events (`detectSessionSounds`, `detectReviewSounds`) and server-pushed automation events (`onServerNotify`) all land in one place.

## Approach

- **History store** — appended in `emitEvent`, **ungated** (unlike the banner, which self-suppresses on focus/mute), so the panel is a faithful log even for events whose popup you never saw. The detectors' arm window still guards it, so a page load can't flood it. Persisted to `localStorage` (same pattern as the sidebar cache), capped at ~100. Entry schema: `{ id, event, key, title, body, ts, seen, kind, target }`.
- **Bell + counter** — a `tk-tool` button in `sidebarToolsStack` (visible in **all** views — sessions *and* boards), with a `.pill-badge`-style unread badge. Unread count joins `sidebarSignature` so an append actually repaints. Opening the panel marks entries `seen`; `seen` persists so a reload doesn't re-inflate the count.
- **Panel** — reuses the `.ctx-menu` shell (so `closeContextMenu` closes it) with a `.notif-panel` modifier for the wider, scrollable, two-line layout; mirrors `openNewManagerMenu`'s anchored-popover positioning. Two-line rows (title + relative timestamp, plus body), a "Clear all", and bell-click toggle.
- **Click-through** — reuses the existing `Notification.onclick` mapping: session events → `selectSession`, `reviewRequested` → the review's session (resolved live) or the reviews board, automation keyed by an issue URL → open externally, config/other → non-navigable.
- **Shared text** — `eventNotificationText()` factored out and used by both the transient banner and the center, so they read identically.

Existing banner/sound behavior is unchanged — the center is purely additive.

## Scope

Client-side history (localStorage) only for v1, so it's **per-client** (each web tab / the desktop app keep their own) — a server-synced, persistent history is a larger follow-up, out of scope here (the `notify` frame already carries `{event, key, title, body}`, so no backend change was needed).

## Test plan / verification

- `node --check app.js` — passes.
- Focused Node harness exercising the store/routing logic (19 assertions, all pass): session/review/automation-URL/config enrichment + classification, unread count, per-append repaint, persist→restore round-trip, seen-persists-across-reload (no re-inflate), 100-entry cap on both append and restore, `relTime` buckets, and routing (session, review→its session, unknown-review→board, deleted-session→no-op, url→open).
- The panel/bell reuse production-proven primitives (`.ctx-menu` menu with the same positioning/outside-click-close, `tk-tool` buttons, `.pill-badge`), so no new visual machinery.

Manual UI check to run in-app: fire an event (`crowTestSound`/`crowTestNotify` exist, or trigger a real transition) → badge increments → open bell → rows list newest-first, badge clears → click a session/review row navigates → "Clear all" empties → reload keeps history but not the unread count.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -675,6 +675,44 @@ html, body {
 .tk-tool:hover { color: var(--gold); border-color: var(--border-strong); }
 .tk-tool.nav-selecting { color: var(--red); border-color: var(--red); }
 
+/* Notification bell (CROW-909): unread count as a badge over the top-right
+   corner of the bell button. Mirrors the Reviews pill's badge palette. */
+.tk-bell { position: relative; }
+.notif-badge {
+  position: absolute; top: -5px; right: -5px; min-width: 15px; height: 15px;
+  padding: 0 3px; box-sizing: border-box;
+  display: inline-flex; align-items: center; justify-content: center;
+  border-radius: 8px; font-size: 9px; font-weight: 700; line-height: 1;
+  font-variant-numeric: tabular-nums;
+  background: var(--gold); color: var(--bg-deep);
+  border: 1px solid var(--bg-card);
+}
+
+/* Notification center panel: a wider, scrollable variant of the .ctx-menu shell
+   with a header (title + Clear all) and two-line rows. */
+.notif-panel { width: 320px; max-width: calc(100vw - 16px); padding: 0; }
+.notif-header {
+  display: flex; align-items: center; justify-content: space-between;
+  padding: 8px 12px; border-bottom: 1px solid var(--border-subtle);
+}
+.notif-title { font-size: 12px; font-weight: 700; color: var(--gold); }
+.notif-clear {
+  padding: 2px 8px; border-radius: 5px; cursor: pointer;
+  font-size: 11px; font-weight: 600; color: var(--text-secondary);
+  background: transparent; border: 1px solid var(--border-subtle);
+}
+.notif-clear:hover { color: var(--text-primary); border-color: var(--border-strong); }
+.notif-empty { padding: 18px 12px; text-align: center; font-size: 12px; color: var(--text-secondary); }
+.notif-list { max-height: 340px; overflow-y: auto; padding: 4px; }
+.notif-item { padding: 7px 9px; border-radius: 6px; cursor: pointer; }
+.notif-item:hover { background: rgba(221, 196, 130, 0.14); }
+.notif-item.notif-static { cursor: default; }
+.notif-item.notif-static:hover { background: transparent; }
+.notif-row1 { display: flex; align-items: baseline; justify-content: space-between; gap: 8px; }
+.notif-item-title { font-size: 12px; font-weight: 600; color: var(--text-primary); overflow: hidden; text-overflow: ellipsis; white-space: nowrap; }
+.notif-time { flex: 0 0 auto; font-size: 10px; color: var(--text-secondary); font-variant-numeric: tabular-nums; }
+.notif-item-body { margin-top: 2px; font-size: 11px; color: var(--text-secondary); overflow: hidden; text-overflow: ellipsis; white-space: nowrap; }
+
 /* ---- Fixed connection/status bar (bottom-left, CROW-593) ---- */
 #statusbar {
   position: fixed;
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -445,7 +445,8 @@ function showEventNotification(event, key, detail) {
   } catch (_) { /* Notification ctor can throw in restricted contexts */ }
 }
 
-// Both channels fire from one call in the detectors below; each self-gates.
+// All three channels fire from one call in the detectors below; each self-gates
+// (sound + banner on the full toggle cascade, the history on the master levels).
 function emitEvent(event, key, detail) {
   playEventSound(event, key);
   showEventNotification(event, key, detail);
@@ -463,20 +464,22 @@ function emitEvent(event, key, detail) {
 // larger follow-up.
 const NOTIF_HISTORY_KEY = 'crow.notif.history';
 const NOTIF_HISTORY_CAP = 100; // keep the last ~100; older entries drop off
-let notifHistory = [];         // [{ id, event, key, title, body, ts, seen, kind, target }]
-let _notifSeq = 0;             // disambiguates entries appended in the same ms
+let notifHistory = [];         // [{ event, key, title, body, ts, seen, kind, target }]
 
 function restoreNotifHistory() {
   try {
     const raw = localStorage.getItem(NOTIF_HISTORY_KEY);
     if (!raw) return;
     const data = JSON.parse(raw);
-    // Filter to plain objects: a parseable-but-wrong array (`[null, …]`, a
+    // Filter to well-formed entries: a parseable-but-wrong array (`[null, …]`, a
     // foreign schema) would otherwise let a later `e.seen` read throw inside
     // notifUnreadCount → sidebarSignature → every renderSidebar, wedging the
-    // whole sidebar until the key is cleared by hand (review).
+    // whole sidebar until the key is cleared by hand. The `ts` check also keeps
+    // a ts-less entry from rendering "Invalid Date" in relTime (review).
     if (Array.isArray(data)) {
-      notifHistory = data.filter((e) => e && typeof e === 'object').slice(-NOTIF_HISTORY_CAP);
+      notifHistory = data
+        .filter((e) => e && typeof e === 'object' && typeof e.ts === 'number')
+        .slice(-NOTIF_HISTORY_CAP);
     }
   } catch (_) { /* corrupt cache — start empty */ }
 }
@@ -488,6 +491,21 @@ function notifUnreadCount() {
   let n = 0;
   for (const e of notifHistory) if (e && !e.seen) n++;
   return n;
+}
+
+// One coalesced sidebar repaint per task, so a detector tick that records k
+// events triggers a single innerHTML rebuild rather than k of them (the unread
+// count is in sidebarSignature, so the cheap signature short-circuit never
+// applies to an append). A microtask, not rAF, so the badge is current before
+// the next paint and onServerNotify still repaints within the same turn (review).
+let _notifRepaintScheduled = false;
+function scheduleNotifRepaint() {
+  if (_notifRepaintScheduled) return;
+  _notifRepaintScheduled = true;
+  Promise.resolve().then(() => {
+    _notifRepaintScheduled = false;
+    try { renderSidebar(); } catch (_) { /* pre-boot — badge paints on first render */ }
+  });
 }
 
 // Another tab wrote the shared history (append, mark-seen, or clear). Re-read so
@@ -511,13 +529,22 @@ function classifyNotification(event, key) {
   return { kind: 'none', target: null };
 }
 
-// Append one event to the history and bump the unread counter. Called from
-// emitEvent unconditionally: the detectors already gate on the arm window, so
-// this can't be flooded by a page load. A 2s per-(key,event) dedup mirrors the
-// sound/banner channels — a flapping poll or a repeated server push produced
-// duplicate popups there, and would produce duplicate rows here (review).
+// Append one event to the history and bump the unread counter. Gated on only the
+// MASTER levels — globalMute and the per-event `enabled` toggle — so a user who
+// muted everything, or switched this event category off in Settings, doesn't get
+// a growing badge they can't opt out of. The per-event sound/system SUB-toggles
+// and the focus-suppression rule are deliberately NOT applied: the center is a
+// passive log, not an interruption, so it keeps events whose transient banner you
+// chose not to see live. A 2s per-(key,event) dedup mirrors the sound/banner
+// channels, where repeated pushes / a flapping poll produced duplicate popups
+// (review).
 const _lastRecordAt = {};
 function recordNotification(event, key, detail) {
+  const N = uiConfig.notifications;
+  if (!N) return;                          // config not loaded — mirror the sibling channels
+  if (N.globalMute) return;                // master mute silences the log too
+  const cfg = N.events[event] || {};
+  if (cfg.enabled === false) return;       // event category switched off entirely
   const k = (key || '') + '|' + event;
   const now = Date.now();
   if (_lastRecordAt[k] && now - _lastRecordAt[k] < 2000) return;
@@ -527,16 +554,12 @@ function recordNotification(event, key, detail) {
   restoreNotifHistory();
   const { title, body } = eventNotificationText(event, key, detail);
   const { kind, target } = classifyNotification(event, key);
-  notifHistory.push({
-    id: now + '.' + (_notifSeq++),
-    event, key: key || '', title, body, ts: now, seen: false, kind, target,
-  });
+  notifHistory.push({ event, key: key || '', title, body, ts: now, seen: false, kind, target });
   if (notifHistory.length > NOTIF_HISTORY_CAP) {
     notifHistory = notifHistory.slice(-NOTIF_HISTORY_CAP);
   }
   persistNotifHistory();
-  // The unread count is part of sidebarSignature, so this repaints the badge.
-  try { renderSidebar(); } catch (_) { /* pre-boot — badge paints on first render */ }
+  scheduleNotifRepaint();
 }
 
 // Relative "5m ago" timestamp for panel rows; falls back to a locale date past a
@@ -709,6 +732,19 @@ let sidebarCacheHit = false;
 function clearSidebarCache() {
   try { localStorage.removeItem(SIDEBAR_CACHE_KEY); } catch (_) {}
   sidebarCacheHit = false;
+}
+// Drop EVERY localStorage payload that could leak one user's workspace to the
+// next on a shared/kiosk browser, at the two auth boundaries (explicit logout,
+// cookie death). The notification history holds the same class of data as the
+// sidebar cache — session names, server-supplied PR/issue titles and bodies,
+// issue URLs — so it has to be purged alongside it (review, CROW-909). One
+// helper so a future cached store can't be wired into one auth path and missed
+// on the other. NOT called from the boot-catch recovery, which only wants to
+// discard a *corrupt* sidebar cache, not a valid history.
+function purgeSharedBrowserCaches() {
+  clearSidebarCache();
+  notifHistory = [];
+  try { localStorage.removeItem(NOTIF_HISTORY_KEY); } catch (_) {}
 }
 function restoreSidebarCache() {
   try {
@@ -1155,10 +1191,10 @@ async function handleAuthOnDisconnect() {
   try {
     if (await sessionExpired()) {
       sessionDead = true;
-      // Parity with explicit logout: drop cached session/ticket payloads when
-      // the remote web cookie dies (crowd restart) so they don't linger in a
-      // shared browser (CROW-613 review).
-      clearSidebarCache();
+      // Parity with explicit logout: drop cached session/ticket payloads AND the
+      // notification history when the remote web cookie dies (crowd restart) so
+      // they don't linger in a shared browser (CROW-613 review / CROW-909).
+      purgeSharedBrowserCaches();
       renderStatusBar();
     }
   } finally {
@@ -1203,9 +1239,10 @@ function renderStatusBar() {
     out.onclick = async () => {
       if (!await confirmModal('Log out of this web session? You’ll need the web password to sign back in.', { title: 'Log out', okLabel: 'Log out' })) return;
       try { await fetch('/logout', { method: 'POST' }); } catch (_) {}
-      // Drop cached session/ticket payloads so a shared browser can't read them
-      // after logout of a password-protected remote session (CROW-613 review).
-      clearSidebarCache();
+      // Drop cached session/ticket payloads AND the notification history so a
+      // shared browser can't read them after logout of a password-protected
+      // remote session (CROW-613 review / CROW-909).
+      purgeSharedBrowserCaches();
       location.reload();  // now unauthenticated → the auth gate serves the login page
     };
     actions.appendChild(out);
@@ -1339,7 +1376,7 @@ async function openNewManagerMenu(anchorEl) {
   const y = Math.min(rect.bottom + 4, window.innerHeight - menu.offsetHeight - 8);
   menu.style.left = Math.max(4, x) + 'px';
   menu.style.top = Math.max(4, y) + 'px';
-  setTimeout(() => document.addEventListener('click', closeContextMenu, { once: true }), 0);
+  armContextMenuClose();
 }
 
 // Notification center panel (CROW-909): an anchored popover mirroring
@@ -1414,7 +1451,7 @@ function openNotificationPanel(anchorEl) {
   const y = Math.min(rect.bottom + 4, window.innerHeight - menu.offsetHeight - 8);
   menu.style.left = Math.max(4, x) + 'px';
   menu.style.top = Math.max(4, y) + 'px';
-  setTimeout(() => document.addEventListener('click', closeContextMenu, { once: true }), 0);
+  armContextMenuClose();
 }
 
 // Sidebar status/activity indicator, mirroring the desktop: for active
@@ -1717,9 +1754,32 @@ function prBadgeParts(pr, am, autoMergeEnabled) {
 // ---------------------------------------------------------------------------
 // Session right-click context menu (custom — suppresses the browser default).
 // ---------------------------------------------------------------------------
+// The outside-click closer, held module-level so closeContextMenu can remove it.
+// Registered NOT as { once: true } on purpose: a click *inside* a menu that
+// stopPropagations (a notification row, "Clear all") never reaches document, so
+// a once-listener would stay armed with no menu on screen — and the next
+// menu-open's own click would then bubble to it and tear the fresh menu straight
+// back down. Tying the listener's lifetime to closeContextMenu instead of to
+// "some outside click eventually happens" fixes that for all six menu sites
+// (review, CROW-909).
+let _ctxMenuCloser = null;
 function closeContextMenu() {
   const m = document.querySelector('.ctx-menu');
   if (m) m.remove();
+  if (_ctxMenuCloser) {
+    document.removeEventListener('click', _ctxMenuCloser);
+    _ctxMenuCloser = null;
+  }
+}
+// Arm the outside-click close one tick out (so the opening click itself doesn't
+// immediately close the menu). Shared by every menu opener; the deferred arm
+// also means a listener still pending from a prior menu fires against an empty
+// document before this one registers.
+function armContextMenuClose() {
+  setTimeout(() => {
+    _ctxMenuCloser = closeContextMenu;
+    document.addEventListener('click', _ctxMenuCloser);
+  }, 0);
 }
 
 // Right-click a board card → a small menu to copy its link(s). Pass an array of
@@ -1741,7 +1801,7 @@ function showCardMenu(e, items) {
   const y = Math.min(e.clientY, window.innerHeight - menu.offsetHeight - 8);
   menu.style.left = Math.max(4, x) + 'px';
   menu.style.top = Math.max(4, y) + 'px';
-  setTimeout(() => document.addEventListener('click', closeContextMenu, { once: true }), 0);
+  armContextMenuClose();
 }
 
 // Clipboard with a legacy fallback (execCommand) for non-secure contexts where
@@ -1780,7 +1840,7 @@ function showSessionMenu(e, s) {
   const y = Math.min(e.clientY, window.innerHeight - menu.offsetHeight - 8);
   menu.style.left = Math.max(4, x) + 'px';
   menu.style.top = Math.max(4, y) + 'px';
-  setTimeout(() => document.addEventListener('click', closeContextMenu, { once: true }), 0);
+  armContextMenuClose();
 }
 
 // Long-press → context-menu bridge for touch devices. Fires `handler(x, y)` at
@@ -1918,7 +1978,7 @@ async function openHandoffAgentMenu(session, anchorEl) {
   const y = Math.min((rect.bottom || 80) + 4, window.innerHeight - menu.offsetHeight - 8);
   menu.style.left = Math.max(4, x) + 'px';
   menu.style.top = Math.max(4, y) + 'px';
-  setTimeout(() => document.addEventListener('click', closeContextMenu, { once: true }), 0);
+  armContextMenuClose();
 }
 
 // Fire a session RPC from the row menu. A verb can succeed and still not have
@@ -4326,7 +4386,7 @@ function showTerminalMenu(e) {
   const y = Math.min(e.clientY, window.innerHeight - menu.offsetHeight - 8);
   menu.style.left = Math.max(4, x) + 'px';
   menu.style.top = Math.max(4, y) + 'px';
-  setTimeout(() => document.addEventListener('click', closeContextMenu, { once: true }), 0);
+  armContextMenuClose();
 }
 
 function connectTerminalWs() {
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -364,6 +364,27 @@ function sessionNameFor(id) {
   return (s && s.name) || 'Session';
 }
 
+// Derive the { title, body } for an event, enriched with the session name (for
+// session events) or the review's repo (for reviews). A server-supplied
+// title/body (automation events) is used verbatim. Shared by the transient
+// browser/Tauri banner and the notification center so both read identically
+// (CROW-909).
+function eventNotificationText(event, key, detail) {
+  const isSession = !!sessions.find((x) => x.id === key);
+  const title = (detail && detail.title) || EVENT_LABEL[event] || event;
+  let body = (detail && detail.body) || '';
+  if (!body) {
+    body = EVENT_DESC[event] || '';
+    if (isSession) {
+      body = `${sessionNameFor(key)} — ${body}`;
+    } else {
+      const r = ((boardData.reviews && boardData.reviews.reviews) || []).find((x) => x.id === key);
+      if (r && r.repo) body = `${r.repo} — ${body}`;
+    }
+  }
+  return { title, body };
+}
+
 // `detail` (optional) carries a server-supplied { title, body } for automation
 // events, whose text is specific (PR number, issue title) rather than derivable
 // from EVENT_DESC (CROW-768).
@@ -394,17 +415,7 @@ function showEventNotification(event, key, detail) {
 
   // A server-supplied title/body already names the PR / issue / session, so it's
   // used verbatim; derived events get the label plus a subject prefix.
-  const label = (detail && detail.title) || EVENT_LABEL[event] || event;
-  let body = (detail && detail.body) || '';
-  if (!body) {
-    body = EVENT_DESC[event] || '';
-    if (isSession) {
-      body = `${sessionNameFor(key)} — ${body}`;
-    } else {
-      const r = ((boardData.reviews && boardData.reviews.reviews) || []).find((x) => x.id === key);
-      if (r && r.repo) body = `${r.repo} — ${body}`;
-    }
-  }
+  const { title: label, body } = eventNotificationText(event, key, detail);
   try {
     // In the desktop app, WKWebView lacks the Web Notification API — post via the
     // Tauri plugin instead. (Click-to-focus below stays web-only.)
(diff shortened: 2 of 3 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Nice, tight feature — hanging the whole thing off `emitEvent` is the right chokepoint, the `eventNotificationText` extraction is a clean refactor with no behavior change, and reusing the `.ctx-menu`/`tk-tool`/`.pill-badge` primitives keeps the surface small. Security is solid (details below). Four things block, though — one of them breaks the panel's primary open path.

### Critical Issues

**🔴 The panel anchors to the viewport corner whenever the badge is non-zero** — `app.js:1310-1314` vs `app.js:1357`

```js
if (notifUnreadCount()) {
  for (const e of notifHistory) e.seen = true;
  persistNotifHistory();
  renderSidebar();          // ← rebuilds #sidebar
}
...
document.body.appendChild(menu);
const rect = anchorEl.getBoundingClientRect();   // ← anchorEl is now detached
```

`renderSidebar` (`app.js:900-904`) does `root.innerHTML = ''` and rebuilds the tools stack from scratch, so the `bell` passed in as `anchorEl` is detached from the document by the time it's measured. A detached element has no CSS layout box, so `getBoundingClientRect()` returns all zeros — the positioning then computes `x = min(0, …) → 0`, `y = min(4, …) → 4` and clamps to `left: 4px; top: 4px`. The panel lands in the top-left corner of the viewport, over the brandmark, instead of under the bell.

This isn't a rare path: the guard is `if (notifUnreadCount())`, so it fires exactly when the badge is showing — i.e. the normal reason anyone clicks the bell. With zero unread the repaint is skipped and positioning is correct, which is probably why it survived manual checking.

Also note the repaint is guaranteed here, not signature-skipped: `notifUnreadCount()` is in `sidebarSignature` (that's the point of the CROW-909 line), so N→0 always changes the signature.

Fix: measure before mutating, e.g. hoist `const rect = anchorEl.getBoundingClientRect();` above the seen-marking block, or defer the `renderSidebar()` until after the panel is positioned.

### Code Quality

**🟡 `localStorage` is per-origin, not per-tab — concurrent tabs clobber each other's history** — `app.js:467-486`

The PR description says "each web tab / the desktop app keep their own", but `localStorage` is shared across every tab on the origin, and there's no `storage` listener anywhere in `web/` — each tab holds `notifHistory` in memory and `persistNotifHistory()` writes the **whole array**, last-writer-wins:

- Tab A records E1 → writes 11 entries. Tab B (still holding its stale 10) records E2 → writes 10 + E2, permanently dropping E1.
- Worse for `seen`: Tab A opens the panel and marks all seen; Tab B's next `recordNotification` writes its all-unseen copy back, so the badge re-inflates on reload — the exact behavior the test plan claims is covered.
- "Clear all" in one tab is silently undone by the other tab's next append.

`crow.sidebar.cache` has the same last-writer-wins shape, but it's a *derived* cache that the next poll re-syncs from the server, so a lost write is free. This history is the only copy of the data and has no server to recover from, so a lost write is permanent. Either add a `storage` listener to re-read before appending, or at minimum correct the scoping claim in the PR body/comment so the limitation is documented accurately.

**🟡 A poisoned `crow.notif.history` breaks the sidebar permanently** — `app.js:467-474`, `app.js:4718-4726`

`restoreNotifHistory` only checks `Array.isArray(data)` — the entries themselves aren't validated. A parseable-but-wrong array (`[null, …]`, or an older/newer entry schema) makes `notifUnreadCount()`'s `if (!e.seen)` throw, and since that's called from `sidebarSignature()`, every `renderSidebar()` throws with it.

The boot block's fallback is built for exactly this ("A stale-schema cache entry must not abort the rest of boot"), but its recovery only clears the *sidebar* cache:

```js
} catch (_) {
  clearSidebarCache();
  sessions = [];
  lastSidebarSig = null;
  try { renderSidebar(); } catch (_) { /* keep going */ }
}
```

`notifHistory` is left poisoned, so the retry throws too, and so does every subsequent poll-driven render (`refreshLive` at `app.js:863-864` doesn't even swallow it — that path is outside its `try`). The sidebar stays dead until the user clears the key by hand. One line fixes it — filter to plain objects on restore (`data.filter((e) => e && typeof e === 'object')`) and/or reset `notifHistory = []` in the boot `catch`.

**🟡 No committed tests, and the documented manual check can't work**

The diff is 2 files (`app.css`, `app.js`) — the "focused Node harness (19 assertions)" from the description isn't in it, so none of that is reproducible or regression-protected. The repo has an established convention for guarding web-asset behavior: `Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift` asserts against `app.js` source (function bodies, specific literals) precisely because there's no JS test runner. 194 new lines of JS with real state (cap, `seen` persistence, classification, restore) warrants the same treatment.

Separately, the manual plan says "fire an event (`crowTestSound`/`crowTestNotify` exist…) → badge increments". Neither helper reaches `emitEvent` — `crowTestSound` calls `crowSound.play` directly and `crowTestNotify` constructs `new Notification` directly — so neither will ever increment the badge. Anyone following that plan will conclude the feature is broken. Worth either routing a test hook through `emitEvent` or correcting the instructions.

### Security Review

**Strengths:**
- No XSS surface added. Every user/server-supplied string goes through `el(tag, className, text)` (`app.js:874-879`), which assigns `textContent`; no `innerHTML` is used anywhere in the new rendering. Server-controlled `title`/`body` from the `notify` frame are safe by construction.
- `navigateToNotification`'s external open is properly guarded: `classifyNotification` only assigns `kind: 'url'` after `/^https?:\/\//.test(key)` (`app.js:494`), which rejects `javascript:`/`data:` URLs, and `window.open(..., '_blank', 'noopener')` severs the opener reference.
- No new network surface, no new RPC, no credentials touched — purely client-local state.
- Both `localStorage` accesses are wrapped in `try/catch` for quota/private-mode, matching the existing cache pattern.

**Concerns:**
- 🟢 The scheme check happens at *classify* time and `entry.target` is then trusted from `localStorage` at *click* time (`app.js:540`). Re-testing the scheme in `navigateToNotification` would be cheap defense-in-depth against a tampered/legacy stored entry. Not a live vector — writing that key already requires same-origin script execution — hence green.

### Non-blocking (🟢 Consider)

- **No dedup on record.** `playEventSound`/`showEventNotification` both carry a 2s per-`(key, event)` dedup; `recordNotification` has none, so a flapping poll or a repeated server push produces duplicate rows. Arguably consistent with "faithful log", but the sibling channels' dedup exists because dupes were observed.
- **A repaint + full serialize per event.** `recordNotification` calls `persistNotifHistory()` (stringifies up to 100 entries) and `renderSidebar()` (full `innerHTML` teardown) per event. `detectSessionSounds` can emit up to 4 events per session per poll, and since the unread count changes each time the signature guard never short-circuits — a burst rebuilds the sidebar once per event. Batching to one repaint/write per detector pass would be cheap.
- **The open panel doesn't live-update.** A notification arriving while the panel is open re-inflates the badge but leaves the list stale until reopen.
- **The tools stack grows 2 → 3 buttons.** `.sidebar-tools` is `flex-direction: column` inside `.tickets-row { align-items: stretch }`, so it goes from `2×29 + 4 = 62px` to `3×29 + 8 = 95px` and the tickets card stretches to match. Probably fine, but worth an eyeball since it changes the sidebar's resting layout.
- **`entry.id` is written but never read** (`app.js:502`) — dead field, or a missing dedup/React-key use it was intended for.
- **Per-event `enabled: false` still records.** Ungating vs. `globalMute`/focus reads as intentional and is well argued, but a user who explicitly turned an event *off* arguably means "don't tell me", and it still lands in the center.
- **Clicking a row for a deleted session is a silent no-op** (`app.js:532`) — the panel closes and nothing happens, with no feedback that the target is gone.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 3 Yellow, 8 Green] findings. The Red is a small, well-localized fix (measure the anchor rect before repainting); the Yellows are the multi-tab clobber, the non-self-healing cache restore, and the missing tests.

---

dgershman:
Thanks — sharp review. Addressed in d033d277.

## Blocking

**🔴 Panel anchored to the corner when the badge was non-zero** — fixed. Hoisted `const rect = anchorEl.getBoundingClientRect();` above the seen-marking `renderSidebar()`, so the bell is measured while still attached. You nailed the mechanism (the repaint detaches the anchor and a detached node measures all-zeros) and the reason it survived manual checking (guarded on `if (notifUnreadCount())`, so only the zero-unread path was ever exercised).

**🟡 Multi-tab clobber** — fixed, not just documented. `recordNotification` and the seen-marking now `restoreNotifHistory()` (re-read the shared store) before writing, so concurrent tabs *compose* instead of last-writer-wins, and a `storage` listener re-syncs each tab's badge. Your E1/E2 and seen-reinflate sequences both hold now (covered in the harness). Corrected the PR body's "per-tab" claim to per-browser — it's shared localStorage kept in sync, still not server-synced.

**🟡 Poisoned cache wedges the sidebar** — fixed both halves. `restoreNotifHistory` filters to `e && typeof e === 'object'` (so a `[null, …]`/foreign-schema array can't make `notifUnreadCount`'s `e.seen` read throw into every `renderSidebar`), and the boot `catch` now resets `notifHistory = []` so the recovery path can't itself re-throw on the retry render.

**🟡 No tests + manual plan can't work** — added `WebNotificationCenterTests.swift` following the `WebTerminalAssetTests` source-assertion convention you pointed to (9 drift guards: chokepoint append, anchor-before-repaint ordering, object filter, boot reset, re-read + storage sync, 2s dedup, signature wiring, scheme guard, no-innerHTML). And you're right that `crowTestSound`/`crowTestNotify` never reach `emitEvent` — added `crowTestEvent()`, which drives the real path so the badge actually moves.

## Greens

- **Dedup on record** — took it: 2s per-`(key, event)` window mirroring the sibling channels. Kills the flapping-dupe class *and* bounds the repaint burst.
- **Re-test scheme at click time** — took it: `navigateToNotification` re-tests `/^https?:\/\//` on the stored `entry.target` before `window.open`.
- **`entry.id` unused** — kept intentionally as a stable per-entry identity (record dedup keys on `(key,event,time)` instead); harmless and useful if a future keyed dedup/render needs it.
- **Deferred for v1** (called out so they're explicit, not overlooked): open panel doesn't live-update on same-tab arrivals (cross-tab badge now does); per-event `enabled:false` still records (intentional, same "faithful log" rationale as the `globalMute` ungating — the per-event toggle governs the transient banner/sound, not the log); deleted-session row is a silent no-op; per-event repaint isn't batched (the dedup makes bursts rare and they're bounded to ≤4/poll). Happy to fold any of these into follow-ups if you'd rather they land here.
- **Tools stack 2→3 buttons** — eyeballed: `.sidebar-tools` is `justify-content: center`, so the three 29px buttons stay centered against the tickets card, which stretches to match. Resting layout grows ~33px but stays balanced.

Re-requesting review.

dhilgaertner (changes requested):
## Code & Security Review

Nice shape overall — hanging the whole feature off the single `emitEvent` chokepoint is the right call, the `.ctx-menu`/`tk-tool`/`.pill-badge` reuse keeps the visual surface small, and the Swift drift-guards are a good answer to "no JS runner in this repo". Two blockers below, both in the interaction/persistence layer rather than the core store.

### Critical Issues

**🔴 1. The panel silently fails to open after any row click or "Clear all"** — `app.js:1403`, `app.js:1377`, `app.js:1417`

`openNotificationPanel` is the only left-click-opened menu in this file that appends **synchronously**. The other two (`openNewManagerMenu:1316`, `openHandoffAgentMenu:1887`) `await` an RPC before `document.body.appendChild(menu)`, which incidentally pushes the append past the current click's propagation; the three right-click menus are opened from `contextmenu`, which never dispatches a `click`. The bell is neither.

That matters because the row and "Clear all" handlers call `ev.stopPropagation()` *and* `closeContextMenu()`. `stopPropagation` keeps the click from reaching `document`, so the `{ once: true }` outside-click listener armed at line 1417 is **never invoked and therefore never removed**. It stays armed with no menu on screen.

Repro, two clicks:
1. Click the bell — panel opens, listener `L` armed on `document`.
2. Click any navigable row — `stopPropagation` → `L` survives; `closeContextMenu()` removes the panel; nav happens.
3. Click the bell again — `openNotificationPanel` appends the panel, then that same click bubbles to `document`, `L` fires, and `closeContextMenu()` removes the `.notif-panel` it just found. **Nothing appears.**

Step 3 also arms a fresh listener, so the *next* open is killed the same way — it only recovers once the user happens to click elsewhere in between. Same path via "Clear all" (1377).

The right-click variant is worse: right-click a session row (arms the listener at 1783) → click the bell. Lines 1365–1370 run first, so the history is marked seen, persisted, and the badge repainted **before** the panel is destroyed. The unread count is silently discarded and the user never sees the list it was pointing at.

This is guaranteed, not racy: per DOM, `document`'s listener list is cloned when the event *reaches* `document`, so a listener registered in an earlier task always runs here. The `setTimeout(…, 0)` at 1417 (and at the five sibling sites) exists precisely because of this semantic.

Either fix works:
- Drop `ev.stopPropagation()` from the row and Clear-all handlers — the document listener's `closeContextMenu()` is exactly what they already do themselves — **and** defer the panel's `appendChild` by a tick so a listener pending from a right-click menu fires first; or
- give `closeContextMenu()` a module-level handle on the armed listener and `removeEventListener` it, which fixes all six call sites at once.

Worth pinning whichever invariant you choose in `WebNotificationCenterTests`, since the rest of the feature is already guarded there.

**🔴 2. `crow.notif.history` survives logout and session expiry** — `app.js:464` vs. `app.js:1208` / `app.js:1161`

Two paths in this file exist specifically to purge cached payloads from a shared browser, both landed as explicit review requirements:

- `app.js:1206-1208` (logout) — *"Drop cached session/ticket payloads so a shared browser can't read them after logout of a password-protected remote session (CROW-613 review)."*
- `app.js:1158-1161` (cookie death) — *"Parity with explicit logout: drop cached session/ticket payloads when the remote web cookie dies … so they don't linger in a shared browser."*

Both call `clearSidebarCache()`, which removes only `crow.sidebar.cache`. The new store holds up to 100 entries of the same class of data — session names via `sessionNameFor`, server-supplied PR/issue titles and bodies, and issue URLs in `entry.target` — and is never cleared on either path. After logout on a shared or kiosk browser, the next user can open the bell (or just read localStorage) and see the previous user's session and PR titles.

Fix: clear `NOTIF_HISTORY_KEY` and reset `notifHistory` wherever `clearSidebarCache()` is called for the auth reason — simplest is one helper that drops both keys, so the next store to land can't miss it either.

### Security Review

**Strengths:**
- No `innerHTML` anywhere in the new code — every server/user string goes through `el()`'s `textContent`, so a hostile PR title or issue body can't become an XSS vector. Pinned by `panelBuildsRowsWithoutInnerHTML`.
- External navigation is scheme-gated **twice**: at record time in `classifyNotification` (`app.js:510`) and again at click time in `navigateToNotification` (`app.js:571`), so a `javascript:`/`data:` target written by a tampered or older-build localStorage entry can never reach `window.open`. `'_blank', 'noopener'` severs the opener.
- `restoreNotifHistory` filters to plain objects inside a try/catch, so a corrupt or foreign-schema key can't wedge `notifUnreadCount` → `sidebarSignature` → every `renderSidebar`. The boot-catch reset at 4782 closes the retry path too.
- `onServerNotify`'s existing string coercion means `key`/`title`/`body` are always strings by the time they reach the store.

**Concerns:**
- Finding 2 above — persistent, uncleared storage of session/PR/issue metadata across an authenticated session boundary.

### Code Quality

**🟡 3. The center ignores the per-event master toggle** — `app.js:449-452`, `app.js:520`

`recordNotification` runs from `emitEvent` with no config gating at all. But `EventNotificationConfig.enabled` is documented at `NotificationSettings.swift:103` as *"Whether this event category triggers any notification at all"*, and the type header describes a cascade where a notification fires only if the global, category, **and** per-event toggles are on.

Bypassing the *focus* rule and the sound/system sub-toggles is a defensible design choice — a history isn't an interruption, and that's the argument the PR body makes. Bypassing `cfg.enabled === false` isn't the same thing: a user who switches `taskComplete` off in Settings still gets a growing unread badge full of `taskComplete` rows, and there is no bell-level opt-out anywhere — not in the Settings tab, not in `crow notifications set`. Suggest gating on the master toggle (and arguably `globalMute`) while leaving the sub-toggles ungated, and updating the comment at 455-459 to name which levels are deliberately bypassed and why.

**🟡 4. A full sidebar rebuild plus a localStorage round-trip per recorded event, on the poll path** — `app.js:527`, `app.js:537`, `app.js:539`

Each `recordNotification` does a `JSON.parse` of the whole history, a `JSON.stringify` back out, and a `renderSidebar()` — and `renderSidebar` does `root.innerHTML = ''` (`app.js:951`) and rebuilds every row. Because `notifUnreadCount()` is in `sidebarSignature`, the signature always differs on append, so the cheap short-circuit never applies.

`detectSessionSounds` (`app.js:646`) can fire up to 4 events per session per tick and `detectReviewSounds` one per newly-appeared review, so a tick that transitions *k* sessions does *k* full sidebar rebuilds plus *2k* synchronous localStorage ops in one block — on a fleet where many parallel sessions finishing together is the normal case, not the edge case. Both detector call sites already render immediately after (`refreshSessions:899`, `refreshBoard:2663`), so the per-event render is redundant there; only `onServerNotify` (`app.js:584`) actually needs it. A dirty flag drained by the caller, or a coalescing `queueMicrotask`/rAF, removes the quadratic while keeping `onServerNotify` correct.

**🟢 5. An open panel doesn't live-update.** `recordNotification` and the `storage` listener (`app.js:496`) both repaint the sidebar, but a panel already on screen keeps its snapshot. A notification arriving while the bell is open bumps the badge without adding a row, and stays unread because seen-marking only happens at open time. Rebuilding the list when `.notif-panel` is present would close both halves of the gap.

**🟢 6. `entry.id` is dead weight.** Generated at `app.js:531` with a `_notifSeq` tiebreaker, then never read — nothing keys off it (no dedup, no DOM key). Either drop it, or lean into it as a per-row `data-id` for a future mark-one-seen affordance.

**🟢 7. `_lastRecordAt` (`app.js:519`) is never trimmed** — one key per distinct `(key|event)` pair, retained for the page's lifetime even after the session is deleted and its entry has aged past the 100-cap. It's the one part of the feature without a bound. Consistent with the pre-existing `_lastSoundAt`/`_lastNotifyAt`, so noting rather than pushing.

**🟢 8. `relTime` renders "Invalid Date" for a `ts`-less entry.** `restoreNotifHistory` filters non-objects but doesn't validate the schema, so an entry from a foreign or older writer with no `ts` falls through every branch (`Math.max(0, NaN)` is `NaN`, so each comparison is false) into `new Date(NaN).toLocaleDateString()`. Cosmetic, not a throw — a `typeof e.ts === 'number'` in the restore filter would tidy it up alongside the object check that's already there.

### Verification run

- `node --check app.js` — passes.
- `swift test --package-path Packages/CrowDaemon --filter WebNotificationCenterTests` — 9/9 pass.
- `swift test --package-path Packages/CrowDaemon` — 314 tests across 54 suites, all pass.

The drift-guard approach is a good fit for the constraint, and each test naming the bug it guards makes them read well. Worth flagging that they can only pin what is present in the source text — findings 1 and 2 are both live in the current source and no assertion reaches them.

### Summary Table

**Recommendation:** Request Changes — driven by [2 Red, 2 Yellow, 4 Green] findings.

---

dgershman:
Thanks again — the two Reds were both real and I'm glad you caught them. Addressed in fc19e57b.

## Blocking

**🔴 1 — Panel silently fails to open after a row click / "Clear all"** — fixed centrally, which also closes the latent version in the five sibling menus. `closeContextMenu` now holds a module-level handle on the armed outside-click listener and `removeEventListener`s it; every opener arms through one `armContextMenuClose()`. So a click *inside* a menu that `stopPropagation`s no longer strands a live listener — the panel's own `closeContextMenu()` tears it down. That fixes both halves you traced:
- **Row/Clear-all → next open:** the leaked listener is now removed when the row handler calls `closeContextMenu()`, so the next bell click builds a panel with nothing armed to kill it.
- **Right-click-then-bell:** `openNotificationPanel`'s early `closeContextMenu()` removes the pending session-menu listener *before* the panel is built, and the fresh arm is deferred a tick — so the bell click bubbles to an empty document. The seen-marking now can't be discarded under a panel that immediately vanishes.

Pinned in `outsideClickCloserIsRemovableNotOnce` (no site left on `{ once: true }`, `closeContextMenu` removes the handle, all openers arm via the helper).

**🔴 2 — History survives logout / cookie death** — fixed. You're right that it's the same data class as the sidebar cache (session names via `sessionNameFor`, server PR/issue titles+bodies, issue URLs). New `purgeSharedBrowserCaches()` drops **both** keys and resets in-memory state, wired into the logout onclick and `handleAuthOnDisconnect` — the two paths that already carried the CROW-613 purge requirement. Deliberately **not** called from the boot-catch recovery: that path only wants to discard a *corrupt* sidebar cache, not a valid history. Pinned in `authBoundariesPurgeTheNotificationHistory`.

## Should-fix

**🟡 3 — Ignored the per-event master toggle** — fixed. `recordNotification` now gates on `globalMute` and `cfg.enabled === false`, so a user who muted everything or switched an event category off in Settings doesn't get an un-opt-out-able badge. The *sub*-toggles (sound/system) and the focus-suppression rule stay bypassed — that's the "a log isn't an interruption" argument, now spelled out in the comment naming exactly which levels are bypassed and why. Pinned in `recordHonorsGlobalMuteAndThePerEventToggle`.

**🟡 4 — Full rebuild + serialize per event on the poll path** — fixed the quadratic. Appends now coalesce to one sidebar rebuild per task via a microtask (`scheduleNotifRepaint`), so a tick transitioning k sessions does one render, not k. Kept `persistNotifHistory` per-append on purpose — it's the multi-tab compose anchor, and stringifying ≤100 small objects is microsecond-scale next to the `innerHTML` teardown that was the actual cost. Since both detector sites already `renderSidebar()` right after detecting, the microtask short-circuits there (signature unchanged); only `onServerNotify` drives the one render. Pinned in `recordCoalescesTheSidebarRepaint`.

## Greens

- **6 (`entry.id` dead weight)** — dropped it (and `_notifSeq`); nothing keyed off it.
- **8 (`relTime` "Invalid Date")** — `restoreNotifHistory` now also requires `typeof e.ts === 'number'`, dropping a ts-less foreign entry alongside the object check.
- **5 (open panel doesn't live-update)** — deferring: the badge now stays correct (record + storage both repaint), but rebuilding an on-screen panel in place cleanly (without the `innerHTML` the security guard forbids there) is enough surface that I'd rather not fold it in under a fix commit. Happy to take it as a follow-up if you'd prefer it here.
- **7 (`_lastRecordAt` unbounded)** — leaving as-is, consistent with the pre-existing `_lastSoundAt`/`_lastNotifyAt` you noted; worth a shared trim across all three in a separate pass.

## Verification
- `node --check app.js` — clean.
- `swift test --filter WebNotificationCenterTests` — 13/13.
- `swift test --package-path Packages/CrowDaemon` — **318 tests, 54 suites, all green** (your 314 + the 4 new guards).
- Behavioral harness (stubbed globals) covering the round-2 logic: globalMute/`enabled`/no-config gating, microtask coalescing (3 appends → 1 render), purge drops both keys, and the `closeContextMenu` listener lifecycle — all pass.

Re-requesting review.

dhilgaertner (changes requested):
## Code & Security Review

Round-3 verification: both Reds from last round are genuinely fixed, and the fixes are better than the minimum. `closeContextMenu` holding a removable handle closes the listener-leak class at all six menu sites rather than patching the panel — I re-traced the right-click-then-bell sequence and the row-click sequence and both now behave. `purgeSharedBrowserCaches` is the right shape for the auth boundary. The master-toggle gating and the microtask coalescing are both correct.

One new Red, found by walking the panel's render path rather than the diff.

### Critical Issues

**🔴 1. `relTime` collides with a pre-existing function of the same name — every notification row's timestamp renders empty** — `app.js:567` vs. `app.js:3118`, call site `app.js:1436`

The PR adds a top-level `function relTime(ts)` at 567 taking **epoch milliseconds**. `app.js` already has a top-level `function relTime(iso)` at 3118 (the shared card helper, from #594) taking an **ISO string**. Same script scope.

`index.html:73` loads it as a classic script (`<script src="/app.js"></script>`), not a module — so this is not a parse error. Top-level function declarations in a Script are var-scoped and initialized in source order, so the **later** declaration wins unconditionally. Every call to `relTime` in the file resolves to the 3118 version.

The panel then calls it with a number:

```js
line1.appendChild(el('span', 'notif-time', relTime(entry.ts)));   // app.js:1436, entry.ts is Date.now()
```

which lands in:

```js
function relTime(iso) {                 // app.js:3118 — the one that actually runs
  if (!iso) return '';
  const then = Date.parse(iso);         // Date.parse coerces to String → "1753753200000"
  if (isNaN(then)) return '';           // ← always taken
  ...
}
```

`Date.parse` stringifies its argument, and a bare 13-digit epoch is not a Date Time String Format value, so V8 returns `NaN`. Verified against the two declarations lifted verbatim from `app.js` and run in one script scope:

```
found 2 relTime declarations
relTime(Date.now()) => ""
```

So `.notif-time` is an empty span on **every** row, always. `.notif-row1` is `justify-content: space-between` with `.notif-time { flex: 0 0 auto }`, so the row still lays out cleanly — the title just takes the full width and no timestamp appears anywhere. That's why it reads as a styling choice rather than a bug on a screenshot, and it's the one part of the panel the Swift guards don't reach.

Blast radius is one-directional and worth stating: the pre-existing card timestamps (`app.js:3393`, `3400`, `3741`) are **unaffected**, since the version they want is the one that wins. Only the new call site is broken.

Scanned for this systematically — `main` has 149 top-level functions and zero duplicated names; HEAD has 160 and exactly one duplicate, this one. So it's introduced here and it's the only instance.

Fix is a rename (`notifRelTime`, or inline it into the panel). Two notes on guarding it:
- `node --check` cannot catch this — a duplicate function declaration is legal in a Script, which is why the PR's syntax check passes.
- A drift guard written in the existing style would be misleading: `functionBody("relTime", …)` matches on the **first** `function relTime(`, i.e. the dead one at 567, so a test asserting the epoch-ms buckets would pass while the feature stays broken. A guard asserting *no duplicate top-level function names in `app.js`* would catch this class rather than this instance.

### Security Review

**Strengths:**
- No `innerHTML` in the new rendering — every server/user string goes through `el()`'s `textContent`. Pinned by `panelBuildsRowsWithoutInnerHTML`.
- External navigation is scheme-gated twice (record time at `app.js:528`, click time at `app.js:594`), and `window.open(…, '_blank', 'noopener')` severs the opener.
- `restoreNotifHistory`'s object+`ts` filter plus the boot-catch reset at `app.js:4842` close the render-wedge path.
- `purgeSharedBrowserCaches` (`app.js:744-748`) is wired into both auth boundaries and correctly *not* into the boot-catch recovery. The single-helper shape is the right call.

**Concerns:**

**🟡 2. The `storage` listener ignores key *removal*, so a logout in one tab can be undone by a bell click in another** — `app.js:514-518`, `app.js:472`, `app.js:1405`

`purgeSharedBrowserCaches` does `localStorage.removeItem(NOTIF_HISTORY_KEY)`. Other tabs get a `storage` event with `newValue === null`, and the listener re-reads:

```js
window.addEventListener('storage', (e) => {
  if (e.key !== NOTIF_HISTORY_KEY) return;
  restoreNotifHistory();        // ← getItem returns null …
  ...
});

function restoreNotifHistory() {
  const raw = localStorage.getItem(NOTIF_HISTORY_KEY);
  if (!raw) return;             // ← … and this returns without touching notifHistory
```

So tab B keeps its full in-memory history after tab A purges. Then `openNotificationPanel`'s seen-marking writes it straight back:

```js
if (notifUnreadCount()) {
  restoreNotifHistory();
  for (const e of notifHistory) e.seen = true;
  persistNotifHistory();        // app.js:1405 — re-creates the key that was just purged
```

One bell click in tab B restores up to 100 entries of session names, server-supplied PR/issue titles and bodies, and issue URLs to `localStorage`, where they outlive every tab. That's the CROW-613 threat model the purge was added for.

This is not at parity with the sidebar cache, which is why I'd fix rather than note it: `persistSidebarCache()` is only reached from `refreshSessions` (`app.js:933`) and `refreshBoard` (`app.js:2721`), both behind a successful RPC — an unauthenticated tab can't re-persist it. `persistNotifHistory()` needs no server round-trip, just a click on a button this PR adds.

The window is real: tab B only self-heals via `handleAuthOnDisconnect`, which requires the `/rpc` socket to actually drop. Until it does, tab B stays fully interactive (the `sessionDead` scrim at `app.js:1217` is set on the same path that purges).

Smallest fix is in `restoreNotifHistory` — treat a missing key as "empty", not "no change":

```js
const raw = localStorage.getItem(NOTIF_HISTORY_KEY);
if (!raw) { notifHistory = []; return; }
```

That also makes the listener's own comment true for all three write shapes it names (append, mark-seen, clear) plus the fourth it doesn't (purge). Worth confirming this doesn't regress the compose path — it doesn't: `recordNotification` reaches it only with the key already absent, where in-memory is `[]` anyway.

### Code Quality

**🟢 3. The `typeof e.ts === 'number'` restore filter isn't currently doing what its comment says.** Added last round for green 8 ("keeps a ts-less entry from rendering `Invalid Date` in `relTime`"). With finding 1 live, the reachable `relTime` returns `''` for every input, so nothing renders `Invalid Date` today. My round-2 note assumed the 567 declaration was the live one — it isn't. Keep the filter (it's correct and becomes load-bearing the moment finding 1 is fixed), but the comment overstates what it's guarding right now.

**🟢 4. `_ctxMenuCloser` is effectively a boolean.** It's only ever assigned `closeContextMenu` (`app.js:1780`), and `addEventListener` already de-dupes identical `(type, listener, capture)` triples — so a double-arm can't stack, and `removeEventListener('click', closeContextMenu)` would work without the handle. The variable is fine as documentation of intent, just noting it carries no state beyond "armed".

**🟢 5. Open panel still doesn't live-update** and **🟢 6. `_lastRecordAt` is unbounded** — both explicitly deferred last round with reasons I agree with. Carrying them forward only so they don't silently drop off the list.

### Verification run

- `node --check app.js` — passes (and, per finding 1, structurally cannot catch it).
- `swift test --package-path Packages/CrowDaemon` — **318 tests, 54 suites, all pass**, matching the PR comment. The 4 new round-2 guards each pin the invariant they name; I re-read all 13 and they're accurate about what they assert.
- Duplicate top-level function scan across `main` vs. HEAD — 0 → 1, isolating finding 1 to this PR.

The drift-guard suite is doing real work now, and `outsideClickCloserIsRemovableNotOnce` asserting *no site is left on the leaky pattern* (rather than just the panel) is exactly the right generalization. Finding 1 is the reminder that these can only pin what's in the source text — nothing in the suite exercises a call.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 1 Yellow, 4 Green] findings. Both actionable items are small: a rename, and turning one early return into an assignment. Everything from the previous two rounds is genuinely resolved.

---

dgershman:
Great catch on the `relTime` collision — that's the kind of bug a diff read wouldn't surface and `node --check` structurally can't. Fixed in f4226da9.

## Blocking

**🔴 1 — `relTime` name collision, every panel timestamp empty** — fixed by renaming mine to `notifRelTime(ts)` (epoch ms); the pre-existing `relTime(iso)` card helper is untouched. Your mechanism is exactly right: classic `<script>`, both top-level, later declaration wins for every call, and `Date.parse` of a bare 13-digit epoch is `NaN` → `''`.

Verified against the real source (lifted both declarations, ran them in one scope):
```
notifRelTime(now)          => "just now"
notifRelTime(now-5m)       => "5m ago"
relTime(isoString)         => non-empty   (card helper intact)
relTime(epochNumber)       => ""          (the collision you found)
```

On guarding it — took your suggestion exactly. Added `noDuplicateTopLevelFunctionNames`, which scans every column-0 `function`/`async function` declaration and asserts no name repeats. That catches the **class**, not the instance (and, as you noted, a `functionBody("relTime", …)` guard would match the dead 567 declaration and pass green while the feature stayed broken). It's 0 duplicates now; it would have flagged `relTime` before the rename.

**🟡 2 — cross-tab purge could be re-persisted** — fixed. `restoreNotifHistory` now treats a missing key as empty rather than "no change":
```js
if (!raw) { notifHistory = []; return; }
```
So a logout/purge in tab A propagates to tab B via the storage event and clears its in-memory copy, instead of tab B's next bell-click seen-marking writing the stale 100 entries back. Confirmed it doesn't regress the compose path — `recordNotification` only reaches `restoreNotifHistory` with the key already absent (where in-memory is `[]` anyway), so the reset is a no-op there. Pinned in `restoreTreatsMissingKeyAsEmptyAndTimestampUsesTheRenamedHelper`, and I updated the storage-listener comment to name all four write shapes (append, mark-seen, clear, purge).

## Greens
- **3** — you're right the ts-filter comment was overstating things while Red 1 was live (nothing rendered `Invalid Date` because the reachable `relTime` returned `''` for everything). With the rename, `notifRelTime` *is* again the function that would render `Invalid Date` for a ts-less entry, so the filter is load-bearing as documented; updated the comment to name `notifRelTime`.
- **4** (`_ctxMenuCloser` effectively boolean) — leaving as-is per your note; it reads as intent and `removeEventListener` by the shared reference is robust either way.
- **5 / 6** (open-panel live-update, `_lastRecordAt` unbounded) — still deferred, carried forward.

## Verification
- `node --check app.js` — clean (and, per Red 1, can't catch a duplicate declaration).
- `swift test --filter WebNotificationCenterTests` — **15/15** (adds the duplicate-name and purge-propagation guards).
- `swift test --package-path Packages/CrowDaemon` — **320 tests, 54 suites, all green**.
- Behavioral check lifting both `relTime`/`notifRelTime` from the real file — confirms the panel timestamp now renders and the card helper is unaffected.

Re-requesting review.

dhilgaertner (approved):
## Code & Security Review

Round 4. The three prior rounds closed everything substantive; this pass found no Red and no Yellow. Verification below is independent of the PR's own harness.

### Critical Issues

None.

### Security Review

**Strengths:**

- **No injection surface.** Every user/server string in the panel goes through `el()` → `textContent` (`app.js:967`); `innerHTML` appears nowhere in `openNotificationPanel`. A hostile PR title or issue body renders as text. The only `innerHTML` in the feature is `icon()` against the static `ICONS` map.
- **External navigation is fail-closed and double-gated.** `classifyNotification` (`app.js:531-537`) only tags `kind: 'url'` after `/^https?:\/\//`, and `navigateToNotification` (`app.js:605`) re-tests the *stored* target at click time before `window.open(..., '_blank', 'noopener')`. I exercised both: `javascript:`, `data:`, and a tampered localStorage target all fall through to a no-op, and the opener is severed. Re-testing a value that round-tripped through localStorage is the right instinct — the record-time check alone would trust a store an older build (or same-origin script) could have written.
- **The new persistent store is treated as the same data class as the sidebar cache.** The history holds session names, server-supplied PR/issue titles and bodies, and issue URLs. `purgeSharedBrowserCaches` (`app.js:754-758`) drops both keys, and both auth boundaries — explicit logout (`app.js:1253`) and cookie death in `handleAuthOnDisconnect` (`app.js:1205`) — route through it. Folding the two keys into one helper is the part that keeps this correct as more cached stores land.
- **Deliberately *not* called from the boot-catch recovery**, which only wants to discard a corrupt sidebar cache — so a crash-on-boot can't silently eat a valid history. The distinction is called out in the comment and pinned by `authBoundariesPurgeTheNotificationHistory`.
- **Cross-tab purge closes the loop.** `restoreNotifHistory` treating a missing key as *empty* rather than *no change* (`app.js:472`) is what stops tab B from re-persisting the history that tab A's logout just dropped. Easy to get wrong; the reasoning is in the comment.
- No new RPC, route, or server-side surface — the existing `notify` frame already carried `{event, key, title, body}`.

**Concerns:** none.

### Code Quality

**Verification performed**

- `node --check app.js` — passes.
- `swift test --package-path Packages/CrowDaemon` — **320 tests / 54 suites, all pass**, including the 15 new `WebNotificationCenterTests`. No regression from rewiring the five pre-existing menu openers.
- Independent Node harness (mine, not the PR's): extracted the store/classify/route/format functions verbatim from `app.js` and ran them against localStorage + DOM stubs — **40 assertions, all pass**. Covered: session/review/automation enrichment, `kind` classification, scheme rejection at both record and click time, dedup, the full gating cascade (incl. confirming the sub-toggles deliberately *don't* suppress the log), persist→restore with `seen` intact, poisoned/corrupt/non-array/missing-key cache handling, the 100-cap on append and restore, `notifRelTime` buckets incl. a future timestamp, all five routing paths, microtask coalescing (5 appends → 1 repaint), and cross-tab compose (a concurrent tab's entry survives an append).

**Things this round got right**

- `armContextMenuClose` (`app.js:1789-1794`) is the correct fix, not just a local one. Because `_ctxMenuCloser` is always the same function reference, repeated arms collapse via `addEventListener`'s dedup, so a stale pending `setTimeout` from a prior menu can't double-register or orphan the handle — and `closeContextMenu` deterministically removes it. I audited all six sites: each opener calls `closeContextMenu()` first and `armContextMenuClose()` last, and the only `.ctx-menu` `.remove()` in the file is inside `closeContextMenu`. No path leaves a menu on screen with no armed closer, or a closer armed with no menu that isn't self-healing.
- Measuring `getBoundingClientRect` *before* the seen-marking `renderSidebar()` (`app.js:1414`) — and pinning the ordering with an assertion rather than a comment — is the right shape for a bug that only fires on the common path (badge showing).
- `noDuplicateTopLevelFunctionNames` guards the *class* of bug (classic-`<script>` redeclaration silently winning) rather than the `relTime`/`notifRelTime` instance. Good call: a per-function assertion would have matched the dead declaration.
- Gating on the master levels only, with the sub-toggles and focus rule deliberately bypassed, is the right call for a passive log — and the *why* is written down at the call site rather than left to be re-derived.

**Consider (non-blocking)**

- **`app.js:1424` — clicking the panel's own chrome dismisses it.** `.notif-item` and `.notif-clear` `stopPropagation`, but the `.notif-panel` container doesn't, so a click on the "Notifications" header strip, the 4px `.notif-list` padding, or the "No notifications yet" empty state bubbles to the document closer and shuts the panel. Every other `.ctx-menu` is 100% `.ctx-item`, so this is the first menu with non-item chrome. A `menu.onclick = (ev) => ev.stopPropagation();` after construction covers it in one line.
- **`app.js:1435` — an open panel doesn't live-update.** An event recorded while the panel is open bumps the badge (`renderSidebar` runs) but adds no row, and a cross-tab "Clear all" leaves this tab's rows on screen. Nothing is lost — it's correct on the next open — but re-rendering the panel body from the `storage` listener and `scheduleNotifRepaint` when `.notif-panel` is mounted would remove the small inconsistency.
- **`app.js:1447` — deleted-session rows look clickable but are inert.** `navigateToNotification` correctly no-ops when `sessions.find` misses, but the row still renders with `cursor: pointer` and a hover highlight. Resolving `kind === 'session'` liveness at render time and falling back to `.notif-static` would make the dead row read as dead.
- **`app.js:534 / 605` — the scheme test is case-sensitive.** `HTTPS://…` classifies as `none`. Fail-closed, so this is a missed navigation rather than a hole, and the daemon emits lowercase — noting it only so the behaviour is a choice rather than an accident.
- **`app.js:1269` — a11y.** The bell carries a `title` but no `aria-label`/`aria-expanded`, and the unread count isn't announced. Consistent with the neighbouring `tk-tool` buttons, so this is a pre-existing pattern rather than a regression, but the badge is the one tool here that carries state worth announcing.
- **`app.js:551` — `_lastRecordAt` is never pruned.** One key per `(key, event)`, unbounded for the page's lifetime. Identical to the existing `_lastSoundAt` / `_lastNotifyAt`, so it's consistency rather than a new leak — mentioned only for whenever the three get a shared helper.

### Summary Table

The feature is additive, hangs off the one chokepoint every event already funnels through, adds no server surface, and the persistence it introduces is purged at both auth boundaries. All six remaining items are polish.

---

PR #912, opened 2026-07-29
Grok Build: identity-probe the resolved `grok` before marking available (CROW-911)

### LINKED ISSUE #911: Grok Build shown as available/active when a different `grok` is on PATH (binary-name collision false positive)
**Grok Build** appears as available/active in the agent picker even though xAI's Grok Build CLI isn't installed. Root cause: the availability probe matches **any** binary named `grok`, and `grok` collides with a different tool.

## Root cause
Availability is decided purely by whether `findBinary()` resolves the launch token:
```swift
// CrowDaemon.swift:939-946 (registerDiscovered)
if let path = agent.findBinary() {
    AgentRegistry.shared.registerKnown(agent, available: true)   // ← marks it "active"
} else {
    AgentRegistry.shared.registerKnown(agent, available: false)  // greyed
}
...
registerDiscovered(GrokAgent())   // CrowDaemon.swift:970
```
`GrokAgent.launchCommandToken = "grok"` (`GrokAgent.swift:32`) with PATH fallbacks `/opt/homebrew/bin/grok`, `/usr/local/bin/grok`, `~/.local/bin/grok`, `~/.grok/bin/grok` (`GrokAgent.swift:48-52`). The code already documents the collision (`GrokAgent.swift:41-47`): **`grok` collides with the community `superagent-ai/grok-cli`, which also installs a binary named `grok`** (xAI's is `xai-org/grok-build`).

So if the user has `superagent-ai/grok-cli` (or any other `grok`) on PATH but **not** xAI's grok-build, `findBinary()` resolves that foreign `grok`, and Crow marks **Grok Build** available/active. It's a name match, not an identity match — Crow can't tell the two `grok`s apart. The existing escape hatch (`defaults.binaries.grok` pin, per the comment at `CrowDaemon.swift:966-969`) only helps a user who already knows to set it; it doesn't prevent the false positive by default.

## Fix directions
1. **Identity-verify the resolved binary before marking available.** After `findBinary()` resolves a path, run a cheap probe (e.g. `grok --version` / `--help`) and only `registerKnown(available: true)` if the output identifies **xai-org/grok-build**. A foreign `grok` (superagent-ai/grok-cli) fails the check → shown disabled. This distinguishes the two tools that a bare name match cannot.
2. **If a probe is too fragile, don't claim "available" on a bare PATH match for collision-prone tokens.** When `grok` resolves via PATH (not via an explicit `defaults.binaries.grok` override), surface an **ambiguous** state/tooltip — e.g. "a `grok` binary was found but may be a different tool; pin the xAI Grok Build path via `defaults.binaries.grok`" — rather than confidently active. An explicit override still counts as available.
3. Note the **same collision shape applies to Cursor's generic `agent` token** (called out in `GrokAgent.swift:42-44`). Consider making the identity-probe a reusable step on `CodingAgent.findBinary()` / registration for collision-prone tokens, but scope this ticket to Grok Build.

## Files
- `Packages/CrowGrok/Sources/CrowGrok/GrokAgent.swift` — token/fallbacks (`:32`, `:48-52`), add an identity-probe hook
- `Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift` — `registerDiscovered` (`:939-946`, call site `:970`) to consult the probe
- `CodingAgent.findBinary()` (CrowCore) if the probe is generalized

## Verify
With `superagent-ai/grok-cli` installed (a `grok` on PATH) but xAI grok-build absent, Grok Build should render **disabled/greyed** with a "not found / different tool" tooltip — not active. Installing the real `xai-org/grok-build` (or pinning `defaults.binaries.grok`) + restarting `crowd` should flip it to available.

dgershman:
Closes #911

## Problem
`grok` collides with the community `superagent-ai/grok-cli`, which installs a same-named binary. Availability was decided purely by whether `findBinary()` resolved *any* `grok` on PATH (`CrowDaemon.registerDiscovered` → `registerKnown(available: true)`), so a user with grok-cli but not xAI's grok-build saw **Grok Build shown as available/active** — a name match, not an identity match.

## Fix (ticket direction 1 — identity probe)
Registration now identity-probes a bare PATH/fallback match before marking Grok Build available:

- **`CodingAgent.verifyBinaryIdentity(atPath:) async -> Bool`** — new protocol seam, default `true` (unambiguous tokens are trusted, no behavior change for Claude/Codex/Cursor/OpenCode/Antigravity). Reusable; Cursor's generic `agent` token can adopt it later (scoped here to Grok per the ticket).
- **`GrokAgent`** overrides it: runs `grok --version` / `--help` (bounded 3s, tolerates non-zero exit) and matches the output against grok-build-specific **flag markers** — `--prompt-file`, `--prompt-json`, `--permission-mode`, `--always-approve` (the same upstream flag set the launch args are pinned to). The conversational Node `grok-cli` carries none of them → shown **disabled**. Markers are flag names, **not** vendor branding, precisely because grok-cli is itself an xAI Grok API client and may reference "xAI" in its help text (which would false-positive a branding match).
- **`CrowDaemon.registerAgents`** — now async; probes on a PATH/fallback match, **skips the probe for an explicit `defaults.binaries.grok` pin** (authoritative, so a version-pinned probe broken by upstream churn can't grey out a correctly-pinned install), and logs an honest "failed identity probe … pin the real path" line for the foreign case.

`ShellRunner` is injected into `GrokAgent` so the probe is unit-testable without spawning a subprocess.

## Tests
`GrokAgentTests` (5 new): accepts grok-build flags, rejects community grok-cli (the exact false positive), rejects empty output, tolerates non-zero exit. Full suites green — **CrowGrok 55, CrowCore 520, CrowDaemon 322**.

## Docs
`docs/agent-harness-matrix.md`: the collision mitigation is now the identity probe rather than "accept the false positive + manual pin" (token-collision prose, the token-row cell, and the version-pinned re-check row).

## Verify (ticket test plan)
With `superagent-ai/grok-cli` on PATH but xAI grok-build absent, Grok Build renders **disabled** (probe finds no grok-build flags). Installing real `xai-org/grok-build` (or pinning `defaults.binaries.grok`) + restarting `crowd` flips it to available. The probe logic is covered by unit tests; the end-to-end PATH scenario needs a machine with grok-cli installed.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Agent/AgentDiscovery.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Agent/AgentDiscovery.swift
@@ -1,0 +1,40 @@
+import Foundation
+
+/// The outcome of deciding whether a discovered agent is launchable.
+public enum AgentAvailability: Sendable, Equatable {
+    /// The agent's binary didn't resolve on PATH / override / fallback.
+    case unavailableNotFound
+    /// The binary resolved via PATH/fallback but **failed the identity probe** —
+    /// a different tool sharing the launch token (e.g. `superagent-ai/grok-cli`
+    /// standing in for `xai-org/grok-build`). `path` is the impostor, for the
+    /// log line (CROW-911).
+    case unavailableFailedProbe(path: String)
+    /// The binary resolved and is genuinely this agent. `viaOverride` is `true`
+    /// when it came from an explicit `defaults.binaries.<kind>` pin (which is
+    /// authoritative and bypasses the identity probe).
+    case available(path: String, viaOverride: Bool)
+}
+
+/// Pure, host-agnostic availability decision for a discovered agent, extracted
+/// so it's unit-testable without the daemon's `registerAgents` (which writes the
+/// registry and logs). Both the daemon and the desktop mirror wrap this
+/// (CROW-911 review — the registration-layer behavior had no test because the
+/// decision lived inside a `private static` local function).
+public enum AgentDiscovery {
+    /// Resolve `agent`'s binary and decide availability. An explicit
+    /// `.override` pin is trusted without probing; a PATH/`.fallback` match on a
+    /// collision-prone token is identity-probed (`verifyBinaryIdentity`) so a
+    /// foreign same-named binary is reported unavailable rather than launchable.
+    public static func evaluate(_ agent: any CodingAgent) async -> AgentAvailability {
+        guard let resolved = agent.resolveBinary() else {
+            return .unavailableNotFound
+        }
+        if resolved.source == .override {
+            return .available(path: resolved.path, viaOverride: true)
+        }
+        if await agent.verifyBinaryIdentity(atPath: resolved.path) {
+            return .available(path: resolved.path, viaOverride: false)
+        }
+        return .unavailableFailedProbe(path: resolved.path)
+    }
+}
--- a/Packages/CrowCore/Sources/CrowCore/Agent/CodingAgent.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Agent/CodingAgent.swift
@@ -1,5 +1,28 @@
 import Foundation
 
+/// How `CodingAgent.resolveBinary()` located a binary. Lets registration branch
+/// on provenance — an explicit `.override` pin is authoritative and bypasses the
+/// identity probe — from one source of truth instead of re-deriving the
+/// override-first precedence a second time (CROW-911 review).
+public enum BinaryResolutionSource: Sendable, Equatable {
+    /// An explicit `defaults.binaries.<kind>` pin.
+    case override
+    /// A `command -v`-style PATH walk on `launchCommandToken`.
+    case path
+    /// A hardcoded `fallbackCandidates` entry (exotic-PATH last resort).
+    case fallback
+}
+
+/// A resolved agent binary and how it was found.
+public struct ResolvedBinary: Sendable, Equatable {
+    public let path: String
+    public let source: BinaryResolutionSource
+    public init(path: String, source: BinaryResolutionSource) {
+        self.path = path
+        self.source = source
+    }
+}
+
 /// A coding agent that Crow can launch in a terminal and observe via hook
 /// events. Phase A wraps the existing Claude Code integration; later phases
 /// introduce additional conformers.
@@ -41,9 +64,16 @@ public protocol CodingAgent: Sendable {
     /// empty list is fine — most agents will resolve through PATH first.
     var fallbackCandidates: [String] { get }
 
+    /// Resolve this agent's binary on disk **with its provenance**, or `nil` if
+    /// it isn't installed. The single owner of the override → PATH → fallback
+    /// precedence; `findBinary()` and registration both read from it so the two
+    /// can't drift (CROW-911 review). Drives binary-presence gating for the
+    /// per-session picker and the launch-command builder below.
+    func resolveBinary() -> ResolvedBinary?
+
     /// Resolve this agent's binary on disk, or return `nil` if it isn't
-    /// installed. Drives binary-presence gating for the per-session picker
-    /// and the launch-command builder below.
+    /// installed. Convenience over `resolveBinary()` for the many callers that
+    /// only need the path.
     func findBinary() -> String?
 
     /// Confirm the binary resolved at `path` is genuinely *this* agent and not
@@ -163,29 +193,35 @@ public extension CodingAgent {
             sessionID: sessionID, worktreePath: worktreePath, prompt: prompt)
     }
 
-    /// Default binary discovery: explicit `BinaryOverrides` → PATH walk
-    /// (using the agent's `launchCommandToken` as the binary name) →
-    /// hardcoded `fallbackCandidates`. Returns the first resolved absolute
-    /// path, or `nil` if nothing matches.
+    /// Default binary discovery with provenance: explicit `BinaryOverrides`
+    /// (`.override`) → PATH walk on `launchCommandToken` (`.path`) → hardcoded
+    /// `fallbackCandidates` (`.fallback`). Returns the first resolved absolute
+    /// path + how it was found, or `nil` if nothing matches.
     ///
     /// This replaces the per-agent hardcoded-list-only impl that left
     /// nvm/Volta/pnpm/asdf installs invisible to Crow (CROW-484).
-    func findBinary() -> String? {
+    func resolveBinary() -> ResolvedBinary? {
         let fm = FileManager.default
         // 1. Explicit user override from `defaults.binaries.<kind>`. Verify
         //    the path is still executable so a stale override falls through
         //    to discovery rather than breaking registration outright.
         if let configured = BinaryOverrides.shared.path(for: kind),
            fm.isExecutableFile(atPath: configured) {
-            return configured
+            return ResolvedBinary(path: configured, source: .override)
         }
         // 2. Walk the user's resolved PATH the same way `command -v` does.
         if let found = ShellEnvironment.shared.findExecutable(launchCommandToken) {
-            return found
+            return ResolvedBinary(path: found, source: .path)
         }
         // 3. Hardcoded last-resort fallback covers the exotic-PATH case.
-        return fallbackCandidates.first { fm.isExecutableFile(atPath: $0) }
+        if let fb = fallbackCandidates.first(where: { fm.isExecutableFile(atPath: $0) }) {
+            return ResolvedBinary(path: fb, source: .fallback)
+        }
+        return nil
     }
+
+    /// Default `findBinary()`: the resolved path from `resolveBinary()`, or `nil`.
+    func findBinary() -> String? { resolveBinary()?.path }
 
     /// Default identity check: trust a bare match. Agents with an unambiguous
     /// launch token don't need to probe — overridden only by collision-prone
--- a/Packages/CrowCore/Sources/CrowCore/Agent/CodingAgent.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Agent/CodingAgent.swift
@@ -46,6 +46,24 @@ public protocol CodingAgent: Sendable {
     /// and the launch-command builder below.
     func findBinary() -> String?
 
+    /// Confirm the binary resolved at `path` is genuinely *this* agent and not
+    /// a different tool that happens to share the launch token (CROW-911).
+    ///
+    /// Registration runs this after `findBinary()` resolves a path **via the
+    /// PATH walk or `fallbackCandidates`** — never for an explicit
+    /// `defaults.binaries.<kind>` pin, which is authoritative (the user has
+    /// named the exact binary). A `false` return marks the agent unavailable
+    /// even though a same-named binary exists, so a colliding tool is shown
+    /// disabled rather than falsely active.
+    ///
+    /// The default returns `true`: an unambiguous launch token is trusted on a
+    /// bare match, preserving today's behavior for every agent. Only agents
+    /// whose token is known to collide override this to run a cheap
+    /// `--version` / `--help` identity probe — today just Grok Build (`grok`
+    /// collides with the community `superagent-ai/grok-cli`); Cursor's generic
+    /// `agent` token has the same shape and can adopt this seam later.
+    func verifyBinaryIdentity(atPath path: String) async -> Bool
+
     /// Build the full shell command (ending with `\n`) that auto-launches
     /// this agent in `worktreePath`. Returns `nil` when the agent can't be
     /// launched — typically because the binary is missing or the session
@@ -168,6 +186,12 @@ public extension CodingAgent {
         // 3. Hardcoded last-resort fallback covers the exotic-PATH case.
         return fallbackCandidates.first { fm.isExecutableFile(atPath: $0) }
     }
+
+    /// Default identity check: trust a bare match. Agents with an unambiguous
+    /// launch token don't need to probe — overridden only by collision-prone
+    /// tokens (Grok Build today) so a foreign same-named binary is shown
+    /// disabled instead of falsely active (CROW-911).
+    func verifyBinaryIdentity(atPath path: String) async -> Bool { true }
 
     /// Default Manager launch command: invoke the agent's CLI binary by
     /// name with no extra flags. The tmux terminal backend owns
--- a/Packages/CrowCore/Sources/CrowCore/ShellRunner.swift
+++ b/Packages/CrowCore/Sources/CrowCore/ShellRunner.swift
@@ -68,28 +68,45 @@ public struct ProcessShellRunner: ShellRunner {
         if let cwd { process.currentDirectoryURL = URL(fileURLWithPath: cwd) }
         process.standardOutput = pipe
         process.standardError = pipe
+        // No `ShellRunner` caller feeds stdin (the protocol has no stdin knob),
+        // so detach it from the parent's — under a foreground `crowd` that's the
+        // controlling tty. A subprocess that reads stdin then EOFs immediately
+        // instead of blocking, which matters for the boot-time identity probe
+        // that runs an unvetted third-party `grok` (CROW-911).
+        process.standardInput = FileHandle.nullDevice
 
         // Drain the pipe on a background task so a >64KB output can't deadlock waitUntilExit.
-        return try await withCheckedThrowingContinuation { (cont: CheckedContinuation<String, Error>) in
-            Task.detached {
-                do {
-                    try process.run()
-                } catch {
-                    cont.resume(throwing: error)
-                    return
-                }
-                let data = pipe.fileHandleForReading.readDataToEndOfFile()
-                process.waitUntilExit()
-                let output = String(data: data, encoding: .utf8) ?? ""
-                if process.terminationStatus == 0 {
-                    cont.resume(returning: output)
-                } else {
-                    cont.resume(throwing: ShellRunnerError.nonZeroExit(
-                        exitCode: process.terminationStatus,
-                        output: output
-                    ))
+        // Wrapped in a cancellation handler so a cancelled run (e.g. the identity
+        // probe's timeout losing the race) best-effort `terminate()`s the child
+        // rather than leaving it running. `onCancel` only signals the process; the
+        // detached task remains the sole resumer of the continuation, so there's
+        // no double-resume.
+        return try await withTaskCancellationHandler {
+            try await withCheckedThrowingContinuation { (cont: CheckedContinuation<String, Error>) in
+                Task.detached {
+                    do {
+                        try process.run()
+                    } catch {
+                        cont.resume(throwing: error)
+                        return
+                    }
+                    let data = pipe.fileHandleForReading.readDataToEndOfFile()
+                    process.waitUntilExit()
+                    let output = String(data: data, encoding: .utf8) ?? ""
+                    if process.terminationStatus == 0 {
+                        cont.resume(returning: output)
+                    } else {
+                        cont.resume(throwing: ShellRunnerError.nonZeroExit(
+                            exitCode: process.terminationStatus,
+                            output: output
+                        ))
+                    }
                 }
             }
+        } onCancel: {
+            // `isRunning` is false before launch (terminate() would raise) and
+            // after exit, so this is a no-op outside the running window.
+            if process.isRunning { process.terminate() }
         }
     }
 }
--- a/Packages/CrowCore/Tests/CrowCoreTests/AgentDiscoveryTests.swift
+++ b/Packages/CrowCore/Tests/CrowCoreTests/AgentDiscoveryTests.swift
@@ -1,0 +1,97 @@
+import Foundation
+import Testing
+@testable import CrowCore
+
+@Suite("AgentDiscovery")
+struct AgentDiscoveryTests {
+    private final class NoopHookConfigWriter: HookConfigWriter, @unchecked Sendable {
+        func writeHookConfig(worktreePath: String, sessionID: UUID, crowPath: String) throws {}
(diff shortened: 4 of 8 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Verified locally on `feature/crow-911-grok-binary-collision`: `swift build` clean for CrowGrok + CrowDaemon; **CrowGrok 55 / CrowCore 520 / CrowDaemon 322 all green**, matching the PR body. The approach is right — identity-probing on flag markers instead of vendor branding is the correct call, and the reasoning for it in `GrokAgent.identityMarkers` is exactly on point (grok-cli *is* an xAI client, so branding would false-positive). One blocking defect in the probe's timeout.

### Critical Issues

**🔴 The 3s probe timeout does not bound anything — a `grok` that doesn't exit blocks daemon boot indefinitely.**
`GrokAgent.swift:231-250` races `runner.run(...)` against a `Task.sleep` inside `withTaskGroup`. `withTaskGroup` does **not** return when the body returns — it drains all remaining child tasks first. `group.cancelAll()` only helps if the losing child is cancellation-aware, and `ProcessShellRunner.run` (`Packages/CrowCore/Sources/CrowCore/ShellRunner.swift:60-94`) is not: it's a `withCheckedThrowingContinuation` resumed from a `Task.detached` that blocks in `readDataToEndOfFile()` + `waitUntilExit()`, with no cancellation handler. So the group waits for the subprocess no matter what the sleep does.

Reproduced with the exact shape of this code and a non-cancellable runner (a continuation resumed 30s later off a `DispatchQueue`):

```
returned after 32.0s: ''
```

— i.e. the "3s cap" contributed nothing but the empty return value. The doc comments at `GrokAgent.swift:218-220` ("bounded by a short timeout so a binary that blocks on stdin can't stall daemon boot") and `:229-230` ("control returns so registration proceeds") are both incorrect as written.

Failure scenario: a user has *some* `grok` on PATH and no `defaults.binaries.grok` pin — precisely the CROW-911 population. If that binary starts a REPL/TUI on an unrecognized `--help` and reads stdin (the probe inherits `crowd`'s stdin; `ProcessShellRunner` never sets `standardInput`, so under a foreground `crowd` that's the tty), or merely forks a child that inherits the stdout pipe so `readDataToEndOfFile()` never sees EOF, `await registerAgents(...)` at `CrowDaemon.swift:98` never returns. The Unix socket (`CrowDaemon.swift:432`) and the HTTP/WS listener (`:476`) are both downstream of it, so **`crowd` hangs at boot with no listener at all** — every `crow` CLI command fails and the web UI never comes up, with the boot log stopped mid-registration. Before this PR the same binary was only executed when a user actually launched a Grok session.

Suggested fixes (any of):
- Make `ProcessShellRunner.run` cancellation-aware — `withTaskCancellationHandler { … } onCancel: { process.terminate() }` with a resume-once guard. Fixes every other call site too.
- Or give the probe its own `Process` wrapper with a watchdog that `terminate()`s the child at the cap, keeping `ShellRunner` for the injectable-fake seam.
- Independently: set `standardInput = FileHandle.nullDevice` for the probe so a stdin-reading binary EOFs immediately instead of blocking.
- Add a test that `probeArg` returns within the cap when handed a runner that never completes and ignores cancellation (the repro above is one).

### Security Review

**Strengths:**
- No injection surface: `probeArg` passes `[binary, arg]` as an argv array through `/usr/bin/env` — no shell, no interpolation — and `binary` is an absolute path from `findBinary()`. `ShellEnvironment.findExecutable` splits PATH on `:` (empty components dropped), so no cwd entry can sneak in.
- Probe output is only substring-matched and lowercased; it is never logged verbatim or evaluated, so a chatty `--help` can't leak into the boot log.
- The pin escape hatch means a user whose binary can't be probed is never permanently locked out, and the failure log names the exact config key.

**Concerns:**
- Trust-boundary delta worth an explicit line in the matrix: `crowd` now **executes** an unvetted PATH-resolved third-party binary at boot, unconditionally and with the user's full login-shell environment. That's inherent to identity probing and it's the user's own PATH, so it isn't a finding on its own — but it is the mechanism behind the Red above, and it's a new automatic exec that didn't exist before.

### Code Quality

- 🟢 `GrokAgent.swift:212` — `guard !haystack.isEmpty else { return false }` is dead. `probeText` (`:221-225`) returns `version + "\n" + help`, so the result is never empty; the empty-output case is rejected only by the marker mismatch. `identityProbeRejectsEmptyOutput` therefore passes without exercising the branch it's named for. Either trim before the guard or drop it.
- 🟢 `GrokAgent.swift:222-223` — all four markers are `--help` flags, so `--version` can essentially never match, yet it always runs first: an extra spawn on every boot for a genuine install, and (given the Red) the leg that stalls before `--help` is even tried. Probe `--help` first and short-circuit on a match.
- 🟢 `CrowDaemon.swift:954-955` — `pinned` re-derives `findBinary()`'s override-first precedence (`CodingAgent.swift:178-181`) in a second place. If that resolution order ever changes, the two silently drift and the probe runs against a pinned binary (or is skipped for an unpinned one). Consider having the resolver report how it resolved — e.g. `(path, source: .override | .path | .fallback)` — so registration branches on one source of truth.
- 🟢 The registration-layer behavior this PR is *about* — pin bypass, disabled-on-failed-probe — has no test; `registerDiscovered` is a local func inside a `private static` method. The unit tests cover only the pure marker match. Extracting the decision into a testable pure helper would close that gap.
- 🟢 `docs/agent-harness-matrix.md` re-check row is dated `2026-07-29`, a day ahead of today and of every neighboring row (`2026-07-26`).
- 🟢 Nice: default-`true` protocol seam keeps every other adapter's behavior byte-identical, `probeRunner` injection makes the matcher testable without a subprocess, and the docs/matrix updates landed in the same PR per the repo's own rule.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 0 Yellow, 6 Green] findings. The design is sound and the tests pass; the blocker is that the probe's advertised timeout is inoperative, which turns a foreign `grok` from "shown disabled" into "daemon won't start" for the same users the fix targets.

---

dgershman:
Thanks for the thorough review — the Red was a real boot-hang and you're right about the mechanism. Pushed `ae8fe07b`.

**🔴 Timeout bounded nothing** — fixed. `probeArg` is now a resume-once race (run vs. sleep) so the cap is a *hard* bound on the awaiting task: it returns `""` at the cap even against a runner that never completes **and ignores cancellation**. Added `probeArgHardBoundsAgainstNeverCompletingRunner` using exactly your repro shape (a `ShellRunner` whose `run` never resumes), with an injected 50ms cap so it's instant. Separately made `ProcessShellRunner` cancellation-aware (`withTaskCancellationHandler` → `terminate()` on cancel) so the timed-out child is reaped rather than leaked, and set `standardInput = .nullDevice` (no `ShellRunner` caller feeds stdin — verified — so a stdin-reading binary EOFs immediately). The doc comments that were wrong are rewritten to describe the real bound.

**🟢 `--version` runs first / dead empty guard** — fixed together: `verifyBinaryIdentity` now probes `--help` first and short-circuits (every marker is a `--help` flag, so a genuine grok-build matches on one spawn and `--version` never runs). Dropped `probeText` and the dead `!haystack.isEmpty` guard.

**🟢 `pinned` re-derives findBinary's precedence** — fixed. New `resolveBinary() -> ResolvedBinary?` on `CodingAgent` reports provenance (`.override`/`.path`/`.fallback`) as the one source of truth; `findBinary()` delegates to it, and registration branches on `.source == .override`. No second copy of the override-first order.

**🟢 Registration decision untested** — fixed. Extracted it into pure `AgentDiscovery.evaluate` (CrowCore) with 5 unit tests: override→available-without-probing, path/fallback match probed, passing→available, failing→`unavailableFailedProbe`, unresolved→`unavailableNotFound`. `CrowDaemon.registerDiscovered` is now a thin wrapper (registry write + log per outcome).

**🟢 Matrix date** — set to `2026-07-26` (the markers reuse that flag-verification snapshot, so it's the honest date). **Security trust-boundary line** — added: `crowd` now execs the PATH-resolved `grok` at boot, on the user's own PATH, output only substring-matched, hard-bounded.

**Field note (documented, follow-up):** confirmed on a real box that grok-build's installer *also* symlinks `agent → grok` (`~/.grok/bin/agent`), shadowing Cursor's `agent` token — the same bug class. Durable fix is adopting this seam for Cursor with a PATH-walk-that-verifies (the wrong `agent` can be first on PATH); tracked as a follow-up rather than expanded here to keep this PR Grok-scoped.

CrowCore 525 / CrowGrok 56 / CrowDaemon 322 green; root package builds clean.

dhilgaertner (approved):
## Code & Security Review — round 2 (`ae8fe07`)

Re-verified locally on `feature/crow-911-grok-binary-collision`: root package `swift build` **clean** (no warnings in the touched files, Swift 6 language mode), and **CrowGrok 56 / CrowCore 525 / CrowDaemon 322 all green** — matching the numbers in the response comment.

**The Red is fixed, and fixed provably.** `probeArg` (`GrokAgent.swift:241-274`) is now a resume-once race between the run `Task` and the sleep `Task` behind an `OSAllocatedUnfairLock`, with both legs driving a single `withCheckedContinuation`. The awaiting task resumes at the cap regardless of what the runner does, so the bound no longer depends on the runner cooperating. `probeArgHardBoundsAgainstNeverCompletingRunner` (`GrokAgentTests.swift:646-654`) pins it with exactly the repro shape — a `ShellRunner` whose `run` never resumes and ignores cancellation — and passes. The doc comments now describe the real bound rather than the old `withTaskGroup` claim. Every other review point landed too.

### Verification I did beyond the tests

- **`ProcessShellRunner` blast radius is contained.** `.nullDevice` on stdin and the new cancellation handler affect *all* `ShellRunner` callers, not just the probe — so I checked them. `ProcessShellRunner()` is constructed in exactly three places (`ProviderManager`, `AcliAvailability`, `GrokAgent`), i.e. `gh`/`glab`/`acli` only. No call site feeds stdin (no `--body-file -`, no `-F -`, no `--input`), and git does **not** go through `ShellRunner` — `CrowGit/GitManager.swift` owns its own `Process()`. So no `git push`/`git worktree` is now killable by a cancelled task, and no subprocess silently gets an empty body. The stdin change is a net improvement: a provider CLI that would have prompted now EOFs instead of blocking on `crowd`'s tty.
- **`resolveBinary()` is behavior-preserving.** The default impl reproduces the old `findBinary()` order exactly (override-if-executable → PATH walk → first executable fallback), and no agent overrides `findBinary()` anywhere in the tree — so Claude/Codex/Cursor/OpenCode/Antigravity resolution is byte-identical.
- **The launch gate really is closed, not just the picker.** `registerKnown(_:available: false)` (`AgentRegistry.swift:55-58`) drops the kind from the launchable `agents` map, so `agent(for: .grok)` returns nil and `SessionService.handoffAgent` (`:1331`) throws `agentNotRegistered` *before* it reaches its own `findBinary()` check. A probe-failed Grok can't be reached via `crow handoff-agent` or `crow agents set` either — the probe result is authoritative across every surface, not just the boot log.
- **Markers are grounded.** All four (`--prompt-file`, `--prompt-json`, `--permission-mode`, `--always-approve`) are documented top-level flags in `GrokLaunchArgs` — and top-level matters: `launchCommand` emits `grok … --prompt-file <p>` (`GrokLaunchArgs.swift:113`), so they appear in `grok --help` rather than under a subcommand's help.

### Security Review

**Strengths:**
- No injection surface, unchanged: argv array through `/usr/bin/env`, absolute path from `resolveBinary()`, no shell.
- Probe output is lowercased and substring-matched only — never logged verbatim, never evaluated. A chatty or hostile `--help` can't reach the boot log.
- `standardInput = .nullDevice` (`ShellRunner.swift:76`) closes the "probe inherits `crowd`'s controlling tty" vector raised last round, for every `ShellRunner` caller.
- The pin escape hatch is unchanged and the failure log names the exact config key, so a probe broken by upstream churn is never a lockout.
- The trust-boundary delta (`crowd` execs the PATH-resolved `grok` at boot) is now written down in the matrix rather than implicit.

**Concerns:** none outstanding.

### Code Quality

- 🟢 `ShellRunner.swift:106-110` — `onCancel` checks `process.isRunning`, but `process.run()` happens later, inside the `Task.detached`. A cancel landing in that window sees `isRunning == false`, skips `terminate()`, and the child then launches unreaped. Sub-millisecond window against a 3s cap, so effectively unreachable here — but the comment ("no-op outside the running window") reads as if a skip is always harmless, when it's a *missed* kill. A launched-flag, or re-checking `Task.isCancelled` immediately after `process.run()` and terminating, would close it.
- 🟢 `GrokAgent.swift:268-272` — the timeout `Task` is never cancelled when the run wins, so every probe leaves a task sleeping out the full 3s. Harmless today; binding it to a `let timer = Task {…}` and cancelling after the race is one line, and matters more when the seam is adopted for Cursor (per the follow-up note) and this runs per-agent.
- 🟢 `GrokAgent.swift:225-227` — `probeTimeoutNanos` is documented as a "per-leg cap", but two legs run, so the real worst case is 6s added to boot before the socket and HTTP listeners bind (both downstream of `CrowDaemon.swift:98`). Bounded and logged now, so not a defect — worth one line on the constant so nobody adds a third leg casually.
- 🟢 `AgentDiscovery.swift:26` — "Both the daemon and the desktop mirror wrap this" isn't true: `registerKnown` is called only from `CrowDaemon.swift` and tests, and the app is retired (`CrowDaemon.swift:102` — "the retired app's `AppDelegate`"). Inherited phrasing from `AgentRegistry.register`'s doc; better not to propagate it into a new file.
- 🟢 `CodingAgent.swift:77` — `findBinary()` stays a protocol *requirement* alongside `resolveBinary()`, but registration reads only `resolveBinary()`. A future agent that overrides `findBinary()` alone — the older, more obvious hook — would be silently ignored by `AgentDiscovery.evaluate`, reintroducing exactly the drift this round removed. Demoting `findBinary()` to extension-only would make that unrepresentable.
- 🟢 `GrokAgentTests.swift:662-674` — `FakeProbeRunner` doesn't record its calls, so `identityProbeAcceptsGrokBuildFlags` can't assert the `--version` leg was skipped; the short-circuit this round added is only implied by the outputs. A call log plus `#expect(runner.calls == ["--help"])` would pin the behavior the comment claims.
- 🟢 Still no end-to-end run against a real `grok-build` (the PR says so, and the ticket's test plan needs a box with grok-cli). The version-pinned re-check row now carries the markers alongside the flag set they're drawn from, which is the right place for that risk to live.
- 🟢 Nice: `AgentDiscovery.evaluate` is a genuinely pure decision with all four outcomes tested — including `.fallback` being probed rather than trusted, which is the easy one to get wrong — and `registerDiscovered` is now a thin registry-write + log wrapper, so the "greyed-out row ⇒ matching boot log line" contract (#879) is visibly upheld for all three cases.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 8 Green] findings. The boot-hang is bounded by construction rather than by the subprocess behaving, the fix is regression-tested against the exact adversarial runner, and the three structural follow-ups from last round (provenance-aware resolution, an extracted testable decision, cancellation-aware runner) all landed properly rather than as patches. The Greens are polish and comment accuracy; none change behavior.

---

dhilgaertner (approved):
## Code & Security Review — round 3 (`b5b6218`)

Round 2 approved `ae8fe07`; this round is the one commit on top — swapping `probeArg`'s resume-once guard from `os.OSAllocatedUnfairLock` to an `NSLock`-backed `ResumeOnceGate` to unbreak the Linux lane. Re-verified the whole branch: root `swift build` **clean**, **CrowGrok 56 / CrowCore 525 / CrowDaemon 322** green locally, and — the thing that actually matters for this commit — I watched CI to completion: **all four jobs pass**, including `Build & Test` (Linux, `swift:6.1`) in 5m04s. The job log shows `Compiling CrowGrok GrokAgent.swift` followed by `Test run with 56 tests passed`, so the `no such module 'os'` failure is fixed at the exact spot it occurred, not routed around.

**The swap is behavior-identical.** `claim()` is a lock-guarded test-and-set returning `true` exactly once; `OSAllocatedUnfairLock.withLock` returning `shouldResume` was the same predicate. `@unchecked Sendable` is justified — `resumed` is the only mutable state and every access is inside `lock.lock()`/`defer { unlock() }`, and `claim()` doesn't recurse so `NSLock`'s non-reentrancy is irrelevant. The gate is load-bearing rather than cosmetic: a second `cont.resume` on a `withCheckedContinuation` is a `SWIFT TASK CONTINUATION MISUSE` fatal error, i.e. a `crowd` crash, so "resumed exactly once" is the property to preserve and it is preserved.

**Choosing `NSLock` over `#if canImport(Darwin)` is the better call**, and it matches the repo. `NSLock` is already the standard guard here — `JSONStore`, `ConfigStore`, `WebAuth`, `TmuxController`, `Scaffolder`, and ~15 other sites. The alternative, mirroring `CrowLog.swift:2-7`'s guarded `import os`, would have meant carrying two lock implementations for a one-bit flag. `CrowGrok` is in `LINUX_PACKAGES` (`ci.yml:58-59`) — the Apple-only import was a genuine break, not a hypothetical.

### Verification I did beyond the tests

- **`FileHandle.nullDevice` as `standardInput` actually works** (`ShellRunner.swift:76`) — this lands on every `ShellRunner` caller, so I ran it rather than assuming. Against real `Process`: `env echo hello` → `exit=0 out=hello` with and without the null stdin (unaffected), and `env cat` with null stdin → `exit=0`, empty output, **returns immediately** instead of blocking. `FileHandle.nullDevice.fileDescriptor` is `-1`, which `NSTask` special-cases. So the stdin-blocking vector really is closed on the shipping platform, not just described as closed.
- **The Linux lane compiles that line but never executes it.** The only real-runner construction in a Linux package, `GitHubAutoMergeAllowedTests.swift:82`, just builds a `GitHubCodeBackend` and reads `capabilities` — no spawn. Not a gap that matters today (see Green below), but the empirical evidence above is macOS-only, so worth naming rather than implying CI proved it.
- **No other Apple-only surface slipped in.** `import os` / `import Darwin` / `OSAllocatedUnfairLock` now appear in the PR's touched files only inside the `ResumeOnceGate` doc comment. Every other occurrence in the tree is either guarded (`CrowLog.swift`) or in a package excluded from `LINUX_PACKAGES` (`CrowTerminal`, `CrowIPC`'s Darwin bits, `CrowAutostart`).
- **The probe still runs once per boot.** `registerAgents` has exactly one caller (`CrowDaemon.swift:98`), `verifyBinaryIdentity` exactly one (`AgentDiscovery.evaluate`), and the registry holds the constructed `GrokAgent`, so nothing re-probes on a later config read. `defaults set --binary` is `restart_required` (`DefaultsRPCSupport.swift:329-338`), so a changed pin can't leave a stale probe verdict live.

### Security Review

**Strengths:**
- Unchanged and still correct: argv array through `/usr/bin/env` (no shell), absolute path from `resolveBinary()`, probe output lowercased and substring-matched only — never logged verbatim, never evaluated.
- `standardInput = .nullDevice` closes the "probe inherits `crowd`'s controlling tty" vector, now confirmed by execution rather than by reading.
- The `defaults.binaries.grok` pin still bypasses the probe, and the failure log still names the exact config key — no lockout path.
- This commit touches only a lock primitive: no change to the trust boundary, the marker set, or what gets executed.

**Concerns:** none outstanding.

### Code Quality

- 🟢 `GrokAgent.swift:276-286` — `ResumeOnceGate`'s double-resume race has no direct test. `probeArgHardBoundsAgainstNeverCompletingRunner` exercises only the timeout-wins leg and the other probe tests only the run-wins leg; nothing drives both to `finish` concurrently. A genuinely simultaneous race is awkward to force deterministically, but calling `claim()` twice and asserting `true, false` would at least pin the primitive that stands between a slow `grok` and a `crowd` crash.
- 🟢 `ShellRunner.swift:76` — the `.nullDevice` stdin change is exercised at runtime only on macOS. Moot today: `ProcessShellRunner` is constructed in three places (`AcliAvailability`, `ProviderManager`, `GrokAgent`) that all run under `crowd`, and `crowd` is absent from `LINUX_PACKAGES` because it depends on the Darwin-only `CrowTelemetry` (`ci.yml:69-74`). Worth remembering if the daemon ever goes cross-platform, since corelibs-Foundation's `Process` handles a `-1` descriptor by its own rules.
- 🟢 Carried over from round 2, all still open and all still polish — re-listed so they aren't lost, not as new objections: the timeout `Task` is never cancelled when the run wins (`GrokAgent.swift:262-266`); `probeTimeoutNanos` is a per-leg cap so the real boot worst case is 6s (`:224-226`); `onCancel`'s `isRunning` check can miss a kill in the pre-`run()` window (`ShellRunner.swift:106-110`); "Both the daemon and the desktop mirror wrap this" is inaccurate (`AgentDiscovery.swift:20-21`); `findBinary()` remains a protocol requirement that `AgentDiscovery.evaluate` ignores (`CodingAgent.swift:77`); `FakeProbeRunner` doesn't record calls, so the `--help`-first short-circuit is implied rather than asserted (`GrokAgentTests.swift:662-674`).
- 🟢 Still no end-to-end run against a real `grok-build` alongside `grok-cli` — unchanged from round 2, and the PR says so plainly. The version-pinned re-check row carries the markers next to the flag set they're drawn from, which is the right home for that risk.
- 🟢 Nice: the commit message states the failure mode, the mechanism, and why local macOS builds didn't surface it — which is exactly the class of bug that otherwise gets re-introduced. Scoping the fix to the one primitive, with no drive-by changes to the probe logic that round 2 signed off on, kept this round cheap to verify.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 5 Green] findings. The portability fix is minimal, preserves the single-resume property the round-2 Red fix depends on, and is proven by the Linux job that previously failed now compiling `GrokAgent.swift` and passing all 56 CrowGrok tests. The Greens are polish and test-coverage nits carried forward; none change behavior.

---

PR #914, opened 2026-07-29
Sidebar: move Select toggle beside Scorecard; compact the ticket box (CROW-913)

### LINKED ISSUE #913: Sidebar layout: move Select-sessions button next to Scorecard; cap ticket box width to the Notifications+Settings buttons
Two small sidebar-top layout tweaks.

## Current layout
- `renderSidebar` builds `.tickets-row` = the **ticket box** (`ticketsCard`) + the **tools stack** (`sidebarToolsStack`), then the nav pills — `app.js:1007-1011`.
- `sidebarToolsStack` (`.sidebar-tools`) holds three `.tk-tool` icon buttons: **Notifications** (bell), **Settings** (gear), **Select sessions** (checkSquare toggle) — `app.js:1265-1287`.
- `navPillRow` row 1 = Reviews · Allowlist · **Scorecard** pills — `app.js:1294-1300` (Scorecard at `:1300`).
- CSS: `.tickets-row { display:flex; gap:6px }`, `.tickets-card { flex:1 1 auto }` (grows to fill), `.sidebar-tools { display:flex; flex-direction:column; gap:4px }` — `app.css:665-667`.

## Asks
1. **Move the Select-sessions button to the right of the Scorecard button.** Remove `selBtn` (the checkSquare toggle) from `sidebarToolsStack` (`app.js:1281-1285`) and render it in `navPillRow` row 1 immediately after the Scorecard pill (`app.js:1300`). Preserve its behavior: toggles `selectionMode`, clears `selectedSessionIDs` on cancel, drives `bulkActionBar`, and shows its active/cancel state (today the `.tk-tool.nav-selecting` red style, `app.css:675`). It'll need restyling to sit in `.nav-pills-row` next to `.nav-pill`s (as a pill-styled toggle or an aligned icon button).
2. **Cap the ticket box width.** With Select removed, the tools stack is just **Notifications + Settings**. The ticket box (`.tickets-card`) should **not be wider than those two buttons**. Today `.tickets-card { flex:1 1 auto }` makes it grow to fill the row; constrain it so its width matches the Notifications+Settings tools cluster. Adjust `.tickets-row` / `.tickets-card` / `.sidebar-tools` (`app.css:665-667`).

## Note / ambiguity
The tools stack is currently a **vertical** column (`flex-direction: column`, `app.css:667`), so "the width of the two buttons" is a visual target — the implementer should match the rendered intent: the ticket box no wider than the Notifications+Settings cluster (laying those two out to define that width if needed). Confirm against the running UI.

## Files
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` — `sidebarToolsStack` (`:1265`), `navPillRow` (`:1290-1300`), `renderSidebar`/`.tickets-row` (`:1007`)
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css` — `.tickets-row`/`.tickets-card`/`.sidebar-tools` (`:665-667`), `.tk-tool` (`:669-675`), `.nav-pill`/`.nav-pills-row` (`:775-776`)

dgershman:
Closes #913

Two sidebar-top layout tweaks.

## 1. Select-sessions toggle → beside Scorecard
Moved the Select toggle out of the vertical tools stack (`sidebarToolsStack`) and into `navPillRow` row 1, immediately after the **Scorecard** pill, as an aligned icon button (`.nav-select`) that stretches to the pill height. Behavior is preserved exactly — toggles `selectionMode`, clears `selectedSessionIDs` on cancel, drives the bulk action bar, and reads red (`.nav-selecting`) while active (✕ close icon). The now-unused `.tk-tool.nav-selecting` rule is dropped.

## 2. Compact the ticket box
The ticket box no longer grows to fill the row (was `flex:1 1 auto`, which made it dominate the sidebar top). With Select gone, the tools stack is just **Notifications + Settings**, now laid out horizontally to define the target width.

**Width judgment call (per the ticket's "visual target / confirm against the running UI" note):** the literal two-button cluster width (2×30px + gap = 66px) truncates the "Tickets" header — at 66px it renders as "Tick", and even removing the label leaves a tall skinny 5-count strip. The card is capped to **92px**, the tightest width that renders the header + refresh + wrapped counts cleanly. The head switches to a `space-between` flex row with an in-flow refresh (a centred title + absolutely-positioned refresh overlap at this width), and the counts wrap. Result: a compact card sitting beside the tools cluster instead of filling the row — a drop from ~294px to 92px.

@dhilgaertner — flagging the 92px-vs-66px call for your confirmation; happy to go tighter (e.g. drop the "Tickets" word for a literal 66px) if you'd prefer the exact two-button width over the legible header.

## Verification
Rendered the sidebar-top DOM against the real `app.css` in headless Chrome (idle + active/selecting states). Confirmed: Select button aligned beside Scorecard and turning red on toggle; `Reviews · Allowlist · Scorecard · Select` all fit within the 350px sidebar; compact ticket card with intact "Tickets" header + 2-column counts; Manager · + row unchanged. `node --check app.js` passes.

## Files
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` — `sidebarToolsStack`, `navPillRow`
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css` — `.tickets-row`/`.tickets-card`/`.sidebar-tools`/`.tickets-head`/`.tickets-counts`, new `.nav-select`

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -648,10 +648,14 @@ html, body {
 }
 .tickets-card:hover { border-color: var(--border-strong); }
 .tickets-card.selected { border-color: var(--gold); }
-.tickets-head { display: flex; align-items: center; justify-content: center; position: relative; }
-.tickets-title { color: var(--gold); font-weight: 700; font-size: 13px; }
+/* Title left, refresh right (CROW-913): the compact .tickets-card is too narrow for a
+   centred title + absolutely-positioned refresh (they overlap below ~110px), so the head
+   is a space-between row with the refresh in-flow. The title ellipsises only under a
+   heavily raised browser minimum-font-size, where the card also grows toward its cap. */
+.tickets-head { display: flex; align-items: center; justify-content: space-between; gap: 4px; }
+.tickets-title { color: var(--gold); font-weight: 700; font-size: 13px; overflow: hidden; text-overflow: ellipsis; white-space: nowrap; min-width: 0; }
 .tickets-refresh {
-  position: absolute; right: 0; top: 50%; transform: translateY(-50%);
+  flex: 0 0 auto;
   display: inline-flex; align-items: center; justify-content: center;
   width: 22px; height: 22px; padding: 0; border-radius: 6px; cursor: pointer; line-height: 1;
   border: 1px solid var(--border-subtle); background: var(--bg-card); color: var(--text-secondary); font-size: 12px;
@@ -662,18 +666,20 @@ html, body {
    matching the detail-header action-button spinner (CROW-749). */
 .tickets-refresh.spinning { opacity: 0.6; cursor: default; }
 .tickets-row { display: flex; align-items: stretch; gap: 6px; margin: 2px 2px 8px; }
-/* Compact ticket box (CROW-913): was flex:1 1 auto, which grew it to fill the whole row
-   and dominate the sidebar top. Now content-sized with a 160px ceiling instead of a hard
-   px pin. With the five pipeline counts present it sits at that ceiling (~160px) in
-   practice — the cap is what caps it, not a magic fixed number — but a *ceiling* is the
-   point: the header slot scales with the box, so a raised browser minimum-font-size can't
-   clip "Tickets" the way a fixed px width did (review), and the counts wrap (2 rows
-   typically, 3 at large fonts) instead of forcing the card wider. No min-width: the
-   350px grid never shrinks it, and 92px would be below where the centred head clears the
-   refresh anyway. Sits beside the horizontal Notifications+Settings tools cluster. */
+/* Compact ticket box (CROW-913 ask #2): was flex:1 1 auto, which grew it to fill the whole
+   row and dominate the sidebar top. Now content-sized: the counts render in a fixed TWO
+   columns (below), so the card's max-content is a narrow ~91px (two count cells + header)
+   rather than the five-cell-wide ~160px it would be on one line — measured 91px at typical
+   counts, 106px with 3-digit counts, as close to the Notifications+Settings cluster (66px)
+   as "Tickets" + the five counts allow. `max-width` is a far-off CEILING, not the resting
+   width: because the width tracks max-content, the box grows with the font, so a raised
+   browser minimum-font-size can't clip the header (grows to ~158px at a 24px floor) — the
+   failure a fixed px width had (review). Two columns also fix the height: it's a constant
+   ~95px regardless of 1-, 2-, or 3-digit counts, where a single wrapping row grew taller
+   with bigger numbers. Sits beside the horizontal Notifications+Settings tools cluster. */
 .tickets-row .tickets-card { flex: 0 1 auto; width: max-content; max-width: 160px; margin: 0; }
 .sidebar-tools { display: flex; flex-direction: row; align-items: center; gap: 6px; flex: 0 0 auto; }
-.tickets-counts { display: flex; flex-wrap: wrap; justify-content: center; gap: 4px 8px; margin-top: 8px; padding: 0 2px; }
+.tickets-counts { display: grid; grid-template-columns: repeat(2, auto); justify-content: center; gap: 4px 10px; margin-top: 8px; padding: 0 2px; }
 .tk-tool {
   display: inline-flex; align-items: center; justify-content: center;
   width: 30px; height: 29px; border-radius: 6px; cursor: pointer;
@@ -802,7 +808,7 @@ html, body {
    after Scorecard. Sized/coloured like .nav-plus so it stretches to the pill height;
    reads red while a selection is active, matching .nav-selecting elsewhere. */
 .nav-select {
-  flex: 0 0 auto; width: 30px; display: inline-flex; align-items: center; justify-content: center;
+  flex: 0 0 auto; width: 30px; padding: 0; display: inline-flex; align-items: center; justify-content: center;
   border-radius: 6px; cursor: pointer;
   background: var(--bg-surface); border: 1px solid rgba(179, 142, 70, 0.3); color: var(--gold-dark);
 }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -663,13 +663,15 @@ html, body {
 .tickets-refresh.spinning { opacity: 0.6; cursor: default; }
 .tickets-row { display: flex; align-items: stretch; gap: 6px; margin: 2px 2px 8px; }
 /* Compact ticket box (CROW-913): was flex:1 1 auto, which grew it to fill the whole row
-   and dominate the sidebar top. Now content-sized within a range instead of a hard px pin
-   — a fixed width clipped the "Tickets" header (and stacked the counts) once a raised
-   browser minimum-font-size or 3-digit counts made the font-relative text outgrow the
-   fixed box (review). The range keeps it compact at typical counts yet lets it expand into
-   the ~161px of otherwise-dead row width, so the header self-heals and the counts stay
-   ~2 rows. Sits beside the horizontal Notifications+Settings tools cluster. */
-.tickets-row .tickets-card { flex: 0 1 auto; width: max-content; min-width: 92px; max-width: 160px; margin: 0; }
+   and dominate the sidebar top. Now content-sized with a 160px ceiling instead of a hard
+   px pin. With the five pipeline counts present it sits at that ceiling (~160px) in
+   practice — the cap is what caps it, not a magic fixed number — but a *ceiling* is the
+   point: the header slot scales with the box, so a raised browser minimum-font-size can't
+   clip "Tickets" the way a fixed px width did (review), and the counts wrap (2 rows
+   typically, 3 at large fonts) instead of forcing the card wider. No min-width: the
+   350px grid never shrinks it, and 92px would be below where the centred head clears the
+   refresh anyway. Sits beside the horizontal Notifications+Settings tools cluster. */
+.tickets-row .tickets-card { flex: 0 1 auto; width: max-content; max-width: 160px; margin: 0; }
 .sidebar-tools { display: flex; flex-direction: row; align-items: center; gap: 6px; flex: 0 0 auto; }
 .tickets-counts { display: flex; flex-wrap: wrap; justify-content: center; gap: 4px 8px; margin-top: 8px; padding: 0 2px; }
 .tk-tool {
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -649,10 +649,13 @@ html, body {
 }
 .tickets-card:hover { border-color: var(--border-strong); }
 .tickets-card.selected { border-color: var(--gold); }
-.tickets-head { display: flex; align-items: center; justify-content: center; position: relative; }
-.tickets-title { color: var(--gold); font-weight: 700; font-size: 13px; }
+/* Title left, refresh right (CROW-913): the compact .tickets-card is too narrow for a
+   centred title with an absolutely-positioned refresh (they'd overlap), so the head is a
+   space-between flex row and the refresh sits in-flow. */
+.tickets-head { display: flex; align-items: center; justify-content: space-between; gap: 4px; }
+.tickets-title { color: var(--gold); font-weight: 700; font-size: 13px; overflow: hidden; text-overflow: ellipsis; white-space: nowrap; min-width: 0; }
 .tickets-refresh {
-  position: absolute; right: 0; top: 50%; transform: translateY(-50%);
+  flex: 0 0 auto;
   display: inline-flex; align-items: center; justify-content: center;
   width: 22px; height: 22px; padding: 0; border-radius: 6px; cursor: pointer; line-height: 1;
   border: 1px solid var(--border-subtle); background: var(--bg-card); color: var(--text-secondary); font-size: 12px;
@@ -663,16 +666,21 @@ html, body {
    matching the detail-header action-button spinner (CROW-749). */
 .tickets-refresh.spinning { opacity: 0.6; cursor: default; }
 .tickets-row { display: flex; align-items: stretch; gap: 6px; margin: 2px 2px 8px; }
-.tickets-row .tickets-card { flex: 1 1 auto; margin: 0; }
-.sidebar-tools { display: flex; flex-direction: column; justify-content: center; gap: 4px; flex: 0 0 auto; }
-.tickets-counts { display: flex; justify-content: center; gap: 10px; margin-top: 8px; padding: 0 2px; }
+/* Compact ticket box (CROW-913): was flex:1 1 auto, which grew it to fill the whole
+   row and dominate the sidebar top. Now a fixed, compact card sitting beside the
+   Notifications+Settings tools cluster. Target was the two 30px buttons' width (66px),
+   but that truncates the "Tickets" header; 92px is the tightest width that renders the
+   header + refresh + wrapped counts cleanly (verified in-browser). The head switches to
+   space-between and the counts wrap so the card degrades gracefully at this width. */
+.tickets-row .tickets-card { flex: 0 0 92px; width: 92px; margin: 0; }
+.sidebar-tools { display: flex; flex-direction: row; align-items: center; gap: 6px; flex: 0 0 auto; }
+.tickets-counts { display: flex; flex-wrap: wrap; justify-content: center; gap: 4px 8px; margin-top: 8px; padding: 0 2px; }
 .tk-tool {
   display: inline-flex; align-items: center; justify-content: center;
   width: 30px; height: 29px; border-radius: 6px; cursor: pointer;
   border: 1px solid var(--border-subtle); background: var(--bg-card); color: var(--gold-dark);
 }
 .tk-tool:hover { color: var(--gold); border-color: var(--border-strong); }
-.tk-tool.nav-selecting { color: var(--red); border-color: var(--red); }
 
 /* Notification bell (CROW-909): unread count as a badge over the top-right
    corner of the bell button. Mirrors the Reviews pill's badge palette. */
@@ -790,6 +798,17 @@ html, body {
   background: var(--bg-surface); border: 1px solid rgba(179, 142, 70, 0.3); color: var(--gold-dark); font-size: 15px; font-weight: 700;
 }
 .nav-plus:hover { color: var(--gold); border-color: var(--border-strong); }
+
+/* Select-sessions toggle (CROW-913): an aligned icon button living in nav-pills row 1
+   after Scorecard. Sized/coloured like .nav-plus so it stretches to the pill height;
+   reads red while a selection is active, matching .nav-selecting elsewhere. */
+.nav-select {
+  flex: 0 0 auto; width: 30px; display: inline-flex; align-items: center; justify-content: center;
+  border-radius: 6px; cursor: pointer;
+  background: var(--bg-surface); border: 1px solid rgba(179, 142, 70, 0.3); color: var(--gold-dark);
+}
+.nav-select:hover { color: var(--gold); border-color: var(--border-strong); }
+.nav-select.nav-selecting { color: var(--red); border-color: var(--red); }
 
 /* ---- Session row: status-hued left accent, lead/trail split, glowing dot, RC glyph ---- */
 .session-row.status-accent { border-left-width: 3px; padding-left: 8px; }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -66,7 +66,6 @@ html, body {
 }
 
 /* ---- Multi-select (#5) ---- */
-.nav-plus.nav-selecting { color: var(--red); border-color: var(--red); }
 .divider.divider-sel { display: flex; align-items: center; gap: 8px; }
 .divider-label { flex: 1 1 auto; }
 .divider-selall {
@@ -649,13 +648,10 @@ html, body {
 }
 .tickets-card:hover { border-color: var(--border-strong); }
 .tickets-card.selected { border-color: var(--gold); }
-/* Title left, refresh right (CROW-913): the compact .tickets-card is too narrow for a
-   centred title with an absolutely-positioned refresh (they'd overlap), so the head is a
-   space-between flex row and the refresh sits in-flow. */
-.tickets-head { display: flex; align-items: center; justify-content: space-between; gap: 4px; }
-.tickets-title { color: var(--gold); font-weight: 700; font-size: 13px; overflow: hidden; text-overflow: ellipsis; white-space: nowrap; min-width: 0; }
+.tickets-head { display: flex; align-items: center; justify-content: center; position: relative; }
+.tickets-title { color: var(--gold); font-weight: 700; font-size: 13px; }
 .tickets-refresh {
-  flex: 0 0 auto;
+  position: absolute; right: 0; top: 50%; transform: translateY(-50%);
   display: inline-flex; align-items: center; justify-content: center;
   width: 22px; height: 22px; padding: 0; border-radius: 6px; cursor: pointer; line-height: 1;
   border: 1px solid var(--border-subtle); background: var(--bg-card); color: var(--text-secondary); font-size: 12px;
@@ -666,13 +662,14 @@ html, body {
    matching the detail-header action-button spinner (CROW-749). */
 .tickets-refresh.spinning { opacity: 0.6; cursor: default; }
 .tickets-row { display: flex; align-items: stretch; gap: 6px; margin: 2px 2px 8px; }
-/* Compact ticket box (CROW-913): was flex:1 1 auto, which grew it to fill the whole
-   row and dominate the sidebar top. Now a fixed, compact card sitting beside the
-   Notifications+Settings tools cluster. Target was the two 30px buttons' width (66px),
-   but that truncates the "Tickets" header; 92px is the tightest width that renders the
-   header + refresh + wrapped counts cleanly (verified in-browser). The head switches to
-   space-between and the counts wrap so the card degrades gracefully at this width. */
-.tickets-row .tickets-card { flex: 0 0 92px; width: 92px; margin: 0; }
+/* Compact ticket box (CROW-913): was flex:1 1 auto, which grew it to fill the whole row
+   and dominate the sidebar top. Now content-sized within a range instead of a hard px pin
+   — a fixed width clipped the "Tickets" header (and stacked the counts) once a raised
+   browser minimum-font-size or 3-digit counts made the font-relative text outgrow the
+   fixed box (review). The range keeps it compact at typical counts yet lets it expand into
+   the ~161px of otherwise-dead row width, so the header self-heals and the counts stay
+   ~2 rows. Sits beside the horizontal Notifications+Settings tools cluster. */
+.tickets-row .tickets-card { flex: 0 1 auto; width: max-content; min-width: 92px; max-width: 160px; margin: 0; }
 .sidebar-tools { display: flex; flex-direction: row; align-items: center; gap: 6px; flex: 0 0 auto; }
 .tickets-counts { display: flex; flex-wrap: wrap; justify-content: center; gap: 4px 8px; margin-top: 8px; padding: 0 2px; }
 .tk-tool {
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1262,7 +1262,7 @@ function renderStatusBar() {
 
 // Notifications + Settings, side by side to the right of the Tickets card
 // (outside the box). The Select-sessions toggle moved down to navPillRow row 1,
-// next to Scorecard (CROW-913); these two now define the ticket box's width.
+// next to Scorecard (CROW-913).
 function sidebarToolsStack() {
   const stack = el('div', 'sidebar-tools');
   // Notification center (CROW-909): bell + unread badge, first so it's the most
@@ -1299,6 +1299,7 @@ function navPillRow() {
   // clears the selection on cancel, and reads red (.nav-selecting) while active.
   const sel = el('button', 'nav-select' + (selectionMode ? ' nav-selecting' : ''));
   sel.title = selectionMode ? 'Cancel selection' : 'Select sessions';
+  sel.setAttribute('aria-label', sel.title);
   sel.appendChild(icon(selectionMode ? 'close' : 'checkSquare', 14));
   sel.onclick = () => { selectionMode = !selectionMode; if (!selectionMode) selectedSessionIDs.clear(); renderSidebar(); };
   row1.appendChild(sel);
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1260,8 +1260,9 @@ function renderStatusBar() {
   }
 }
 
-// Settings + Select, stacked vertically to the right of the Tickets card
-// (outside the box): Settings on top, Select below.
+// Notifications + Settings, side by side to the right of the Tickets card
+// (outside the box). The Select-sessions toggle moved down to navPillRow row 1,
+// next to Scorecard (CROW-913); these two now define the ticket box's width.
 function sidebarToolsStack() {
   const stack = el('div', 'sidebar-tools');
   // Notification center (CROW-909): bell + unread badge, first so it's the most
@@ -1278,11 +1279,6 @@ function sidebarToolsStack() {
   gear.appendChild(icon('wrench', 14));
   gear.onclick = () => { if (window.openSettings) window.openSettings(); };
   stack.appendChild(gear);
-  const selBtn = el('button', 'tk-tool' + (selectionMode ? ' nav-selecting' : ''));
-  selBtn.title = selectionMode ? 'Cancel selection' : 'Select sessions';
-  selBtn.appendChild(icon(selectionMode ? 'close' : 'checkSquare', 14));
-  selBtn.onclick = () => { selectionMode = !selectionMode; if (!selectionMode) selectedSessionIDs.clear(); renderSidebar(); };
-  stack.appendChild(selBtn);
   return stack;
 }
 
@@ -1290,7 +1286,7 @@ function sidebarToolsStack() {
 function navPillRow() {
   const wrap = el('div', 'nav-pills');
 
-  // Row 1: Reviews · Allowlist · Scorecard (kept together, no mid-group wrap).
+  // Row 1: Reviews · Allowlist · Scorecard · Select (kept together, no mid-group wrap).
   const row1 = el('div', 'nav-pills-row');
   const rev = navPill('Reviews', selectedBoard === 'reviews', () => selectBoard('reviews'));
   const unseen = (boardData.reviews && boardData.reviews.unseen) || 0;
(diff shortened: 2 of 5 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Clean, well-commented change. The JS half is a faithful move — I verified behavior parity rather than assuming it. The CSS half is where the findings are, and both trace to the same root: `92px` is a hard pin tuned against one font rendering.

**How I verified:** rendered the sidebar-top DOM (reconstructed from `ticketsCard()` / `sidebarToolsStack()` / `navPillRow()`) against the real `app.css` in headless Chrome, and against `main:app.css` for a baseline, at sidebar widths 280–430px and with Chrome's minimum-font-size setting on and off. I did not run `swift build` — this is a resources-only change and CI's **Build & Test** is green.

### Security Review

**Strengths:**
- No new attack surface. Pure layout: no network calls, no auth/permission surface, no user-controlled data reaching the DOM.
- The one `innerHTML` in the path (`icon()`, app.js:1501) interpolates only a numeric size and a lookup from the static `ICONS` map — no injection vector, and unchanged by this PR.
- `refresh.onclick`'s `e.stopPropagation()` and the card's `selectBoard('tickets')` handler are preserved, so the nested-button click semantics don't regress.

**Concerns:** none found.

### Code Quality

**Verified correct:**
- `sidebarSignature()` (app.js:977–991) already lists `selectionMode` and `[...selectedSessionIDs]`, so the relocated toggle's `renderSidebar()` still repaints past the signature short-circuit at app.js:995.
- The `onclick` body is identical to the one removed — toggle, clear-on-cancel, re-render. Bulk action bar still keys off `selectionMode` (app.js:1012).
- Dropping `.tk-tool.nav-selecting` is safe: after the move, `nav-selecting` is only ever applied to `.nav-select` (app.js:1300) and `.action-btn` (app.js:3318, 3670). No `.tk-tool` receives it.
- `.nav-select` stretches to exactly the pill height (measured 28px vs `.nav-pill` 28px), and no test references `sidebarToolsStack`/`selectionMode`. `node --check app.js` passes.

---

#### 🟡 1. The 92px pin clips the "Tickets" header under a raised minimum font size

`app.css:675` — `.tickets-row .tickets-card { flex: 0 0 92px; width: 92px; }`

The header slot is a fixed `92 − 2 (borders) − 16 (padding) − 22 (refresh) − 4 (gap) = 48px`, but the text inside it is font-relative. Measured headroom on stock macOS is **0.8px** (`BlinkMacSystemFont`, "Tickets" @ 700 13px = 47.2px in 48px).

Raise the browser's minimum font size and it clips:

| | pre-PR (`main`) | this PR |
|---|---|---|
| `.tickets-title` @ default | fits (289px card) | fits, 0.8px spare |
| `.tickets-title` @ min-font-size 16 | fits (289px card) | **needs 58px, has 48px → "Ticket…"** |
| row-1 pill labels @ min-font-size 16 | Reviews 116 / Allowlist 92 / Scorecard 105, no clip | **all three clip** |

**Failure scenario:** a user with Chrome → Settings → Appearance → *Minimum font size* raised to 16px (or the Android Chrome / Safari "never use font sizes smaller than" equivalents) opens the Crow web UI. The sidebar reads `Ticket…` and `Reviews…` / `Scorecar…`. `px` box widths don't scale with that setting, so the text grows inside a fixed slot. This matters more than usual here because the web UI is explicitly built to be reached from other devices (web password + https proxy), so non-macOS, non-default clients are an expected audience — not an edge case.

Worth flagging specifically because clipping the header to "Tick" is the stated reason the PR rejected 66px. At 92px the same failure still happens, just for a smaller set of clients.

The pill truncation is the same root cause from the other side: the new Select button takes `30 + 6 = 36px` out of row 1, dropping each pill by ~11px (Scorecard 103 → 91px at a 350px sidebar).

#### 🟡 2. The compact card grows taller as ticket counts grow

`app.css:675` + `app.css:677` (`.tickets-counts { flex-wrap: wrap }`)

With 74px of content width, the 5 count cells fit 2-per-row at 1–2 digits but drop to **1-per-row at 3 digits**:

| counts | pre-PR sidebar-top height | this PR |
|---|---|---|
| 1–2 digit (`7 3 12 4 9`) | 95px | 95px (3 count rows) |
| 3-digit (`128 346 999 412 157`) | 95px | **129px** (5 count rows) |
| 3-digit @ min-font-size 16 | 95px | **154px** |

**Failure scenario:** a board with 100+ tickets in any status (`boardData.tickets.counts.Backlog = 128`) makes the sidebar top 36% taller than before and pushes the session list down. Pre-PR the height was a constant 95px regardless of counts.

The trade is the wrong way round for a sidebar: it buys horizontal space that isn't scarce — the row leaves **161px of its 325px unused** at a 350px sidebar — by spending vertical space, which is (the sidebar scrolls on that axis).

#### Suggested fix for both

Let the card size to its content within a range instead of pinning it, so larger fonts and wider counts expand into the dead horizontal space rather than clipping or stacking:

```css
.tickets-row .tickets-card { flex: 0 1 auto; width: max-content; min-width: 92px; max-width: 160px; margin: 0; }
```

That keeps the compact look at typical counts (still ~92px), self-heals the header at larger font sizes, and holds the counts at 2 rows for 3-digit values. Worth re-measuring, but it removes the single-rendering assumption. This is also the concrete data for the 92-vs-66 call you flagged for @dhilgaertner: 66px clips the header on **every** client, 92px clips it on some.

#### 🟢 3. Three copies of the same `nav-selecting` red-state rule

`app.css:69` (`.nav-plus.nav-selecting`), `:532` (`.action-btn.nav-selecting`), `:811` (new `.nav-select.nav-selecting`) all declare the identical `color: var(--red); border-color: var(--red)`. The first is dead — nothing has ever put `nav-selecting` on `.nav-plus`. Since this PR already deleted the dead `.tk-tool.nav-selecting` sibling, collapsing the rest to a bare `.nav-selecting { color: var(--red); border-color: var(--red); }` would finish the job.

#### 🟢 4. No test pinning the move

`WebTerminalAssetTests.swift:305` pins `.board-empty-hint`'s removal from the immediately-preceding CROW-907 PR, so there's precedent for asserting web-asset invariants. A matching one-liner (`.tk-tool.nav-selecting` gone, `.nav-select` present) would be cheap.

### Summary Table

Neither Yellow is a correctness bug and the behavior half of this PR is solid — it's the fixed `92px` I'd like revisited, ideally with the content-sized range above, which also answers the width question you raised in the description.

dgershman:
Thanks for the careful review — the headless-Chrome-at-`minimumFontSize=16` measurements were the convincing part. Pushed 205126d7.

**🟡 1 & 🟡 2 (fixed):** Adopted your content-sized range verbatim:
```css
.tickets-row .tickets-card { flex: 0 1 auto; width: max-content; min-width: 92px; max-width: 160px; margin: 0; }
```
Re-verified in headless Chrome at `minimumFontSize=16` with 3-digit counts (`128 346 999 412 157`): the header renders in full and the counts stay bounded (≤3 rows) — no clip, no runaway height. One calibration note for the record: with all five pipeline counts present, the counts' one-line max-content + card chrome lands the card at **~160px in nearly every case** (not ~92px) — the 5 cells never shrink the card below ~142px, so `min-width:92` only guards the degenerate <5-count case. That's fine — it's compact-enough, bounded, and self-heals — just flagging that "still ~92px at typical counts" reads closer to ~160px in practice.

Because the card is therefore never actually narrow, the **space-between head redesign I'd added was unnecessary churn** — reverted `.tickets-head`/`.tickets-title`/`.tickets-refresh` back to the original centred-title + absolute-refresh, verified no overlap across the whole range.

This also settles the 66-vs-92 question I'd raised: neither hard pin was right — 66 clips on every client, 92 clips on some — so a self-healing range is the correct answer. 👍

**🟢 3 (partially — with a caveat):** Dropped the genuinely-dead `.nav-plus.nav-selecting`. I did **not** collapse the remaining two to a bare `.nav-selecting { … }`, though: a bare `.nav-selecting` is specificity `(0,1,0)` and would lose to `.nav-select:hover` / `.action-btn:hover` `(0,1,1)`, so hovering an *active* Select/select button would flip it from red back to gold. The compound `.nav-select.nav-selecting` `(0,2,0)` is what keeps red winning over hover — so it's load-bearing, not pure duplication. Left the two section-local rules in place for that reason (+ locality); happy to group them into one multi-selector rule if you'd prefer the DRY, but a bare selector would regress the hover state.

**🟢 4 (done):** Added `WebTerminalAssetTests.selectToggleLivesInNavRowNotTheToolsStack` (Select out of `sidebarToolsStack`, into `navPillRow` as `.nav-select`; `.tk-tool.nav-selecting` gone), mirroring the CROW-907 `.board-empty-hint` precedent. `swift test --filter WebTerminalAssetTests` — 9/9 green.

Re-requesting review.

dhilgaertner (changes requested):
## Code & Security Review

Re-review of `205126d`. **Both Yellows from round 1 are genuinely fixed** — I re-measured rather than taking the commit message for it, and the range does what it was supposed to. One new finding, in the test that was added to close Green 4.

**How I verified:** rebuilt the sidebar-top DOM (`ticketsCard()` / `sidebarToolsStack()` / `navPillRow()`) and drove headless Chrome over CDP against the real `app.css`, with `main:app.css` + the pre-PR DOM as the baseline — at sidebar widths 200/280/320/350/430px, with 1–2-digit and 3-digit counts, idle and selecting, and with `--blink-settings=minimumFontSize=16`. Ran `swift test --filter WebTerminalAssetTests` locally (**9/9 pass**) and `node --check app.js` (clean). CI is green on all four jobs. `web-tests/` needs `npm install` for `jsdom`, which the sandbox can't reach — CI covers it, and `sidebar-groups.test.js` only exercises the grouping helper, so nothing there touches the moved button.

### Round 1 findings — verified resolved

**🟡 1 (header clip) — fixed.** At `minimumFontSize=16`, sidebar 350px: `.tickets-title` renders its full **58.1px** with **20.0px** of clearance to the refresh button, `scrollWidth == clientWidth`. No clip at any width I tried.

**🟡 2 (height growth) — fixed, and now better than `main`.** The sidebar-top row is *shorter* than pre-PR and no longer varies with the counts:

| | `main` | this PR |
|---|---|---|
| 1–2 digit counts (`7 3 12 4 9`) | 95px | **72px** (2 count rows) |
| 3-digit counts (`128 346 999 412 157`) | 95px | **72px** (2 count rows) |
| 3-digit @ `minimumFontSize=16` | 95px | 106px (3 count rows) |

The one row where it's taller is the `minFont=16` × 3-digit corner, +11px — a fair trade for the 23px it gives back in the common case. Card width is a stable 160px, tools 66px, row 325px, no overflow at 200–430px.

**🟢 3 — dropped, and your rebuttal is right.** I forced `:hover` on an *active* toggle via `CSS.forcePseudoState` under both selectors:

| | idle | hovered |
|---|---|---|
| as shipped, `.nav-select.nav-selecting` | `rgb(255,69,58)` | `rgb(255,69,58)` ✅ |
| hypothetical bare `.nav-selecting` | `rgb(255,69,58)` | `rgb(221,196,130)` ❌ gold |

So the compound selector is load-bearing exactly as you said — hovering an active Select would have flipped it back to gold. (Both `.nav-select:hover` and `.nav-select.nav-selecting` are specificity `(0,2,0)`; source order at `app.css:807`/`:808` is what breaks the tie. Bare `.nav-selecting` is `(0,1,0)` and loses outright.) Good call not taking the suggestion.

**🟢 4 — test added and passing**, but see the finding below.

### Security Review

**Strengths:**
- Still no new attack surface: pure layout plus one relocated event handler. No network, auth, or permission surface touched; no user-controlled data reaches the DOM.
- `sel.title` is set as a property, not markup. The only `innerHTML` in the path (`icon()`, app.js:1504) interpolates a numeric size and a static `ICONS` lookup — unchanged by this PR.
- The revert to the absolutely-positioned refresh preserves `e.stopPropagation()` on `refresh.onclick`, so the nested-button semantics inside the card's `selectBoard('tickets')` handler still hold.

**Concerns:** none found.

### Code Quality

**Verified correct:**
- `swift test --filter WebTerminalAssetTests` → 9/9, including the new case.
- `.nav-select` measures 30×29px against `.nav-pill` 29px — exact height match; 32px vs 32px at `minFont=16`.
- Selecting state resolves to `rgb(255,69,58)` on the relocated button.
- Nothing else in the repo references `.tk-tool.nav-selecting` or the old location. `board.test.js`'s `nav-selecting` hits are the Reviews board's own `.action-btn` toggle (`reviewSelectionMode`), untouched here.

*Not a finding, just the data:* row-1 pill labels **do** fit at `minimumFontSize=16` in this harness — but `Scorecard` lands at 83px in an 84.3px content box, ~1.3px of headroom versus ~15px on `main`, since the new 30px button plus its gap comes out of the three pills. That's inherent to what CROW-913 asks for and I'm not asking you to change it; noting it so it's on the record if a future pill label gets longer.

---

#### 🟡 1. The new test's `.nav-select` assertion can't fail — `.nav-selecting` contains it as a substring

`WebTerminalAssetTests.swift:328-330`

```swift
#expect(
    css.contains(".nav-select"),
    "the relocated Select toggle needs its .nav-select styles")
```

`".nav-select"` is a prefix of `".nav-selecting"`, and `app.css:531`'s pre-existing `.action-btn.nav-selecting` rule (plus the comment above it on `:530`) is not going anywhere. So the substring is present whether or not the rules this assertion names exist.

**Failure scenario:** someone deletes the whole `.nav-select { … }` / `:hover` / `.nav-selecting` block at `app.css:799-808` — the exact drift this test was added to catch. The Select button renders as an unstyled default `<button>` in the pill row, and `selectToggleLivesInNavRowNotTheToolsStack` still passes green.

I mutation-tested it rather than eyeballing — cut the 660-byte block out of `app.css` and re-ran the predicate:

```
mutated still contains ".nav-select"? -> True
  because of: ['/* Active Select toggle on a board button reads red, matching .nav-selecting. */',
               '.action-btn.nav-selecting { color: var(--red); border-color: var(--red); }']
```

The other three assertions in the test are sound and do bite — this is the one that doesn't. The CROW-907 precedent it mirrors got away with a bare `contains` because `.board-empty-hint` is a unique string; `.nav-select` isn't. Anchoring to the rule opening fixes it:

```swift
#expect(
    css.contains(".nav-select {") && css.contains(".nav-select.nav-selecting"),
    "the relocated Select toggle needs its .nav-select styles")
```

Both halves fail under the same mutation. Worth pinning `.nav-select.nav-selecting` specifically too, given you just established (correctly) that the compound form is what keeps red winning over hover — that's the part most worth freezing.

#### 🟢 2. The negative assertion skips this suite's own `stripComments` convention

`WebTerminalAssetTests.swift:317-319` asserts `!functionBody("sidebarToolsStack").contains("checkSquare")` against the raw source. The suite has a helper for exactly this — `stripComments` (`:45-70`), documented as existing because "the prose around a handler legitimately names the thing the handler must not call … and that explanation is the part most worth keeping" — and both neighbouring negative assertions use it (`:224`, `:288`).

You're safe today only by accident of where the prose sits: `functionBody` starts *after* `function sidebarToolsStack(`, so the block comment at app.js:1263-1265 that explains the move is outside the guard, and the two in-body comments happen not to say `checkSquare`. A future in-body comment along the lines of *"the checkSquare toggle lives in navPillRow now"* fails the test for describing the invariant it's enforcing. `Self.stripComments(String(...))` around it costs nothing.

#### 🟢 3. The declared range is a fixed 160px in practice — and 92px is the one width where the head overlaps

`app.css:665-672`

The comment sells a dynamic that doesn't happen: *"keeps it compact at typical counts yet lets it expand into the ~161px of otherwise-dead row width."* Measured, the card sits at its **160px `max-width` in every configuration** — 1-digit, 2-digit, 3-digit, `minFont=16`, sidebar 200–430px. Your own PR comment says the same ("~160px in nearly every case"), so this is really just the CSS comment not having caught up; the next reader will go looking for a compact state that never renders. Same for *"the counts stay ~2 rows"* — true at default font, 3 rows at `minFont=16`.

`min-width: 92px` is the loose end. It only binds if the card is flex-shrunk, which needs a sidebar under ~256px — unreachable, since `#app` is a hard `grid-template-columns: 350px 1fr` and the ≤720px branch gives the sidebar the full viewport. That matters because 92px is *also* where the reverted centred-title + absolutely-positioned head breaks: forced to a 200px sidebar, the card shrinks to 103px and the title overlaps the refresh button by **3.1px** — the overlap the space-between head you reverted was guarding against.

Nothing is broken today and I'm not asking you to reinstate that head; the revert was the right call. But the CSS currently declares a resting width it can't render cleanly at. Either drop `min-width` (the max-width is what's doing the work) or keep it with a line saying it's a shrink floor that the fixed-350px grid never reaches — so the next person who makes the sidebar resizable sees the tripwire.

### Summary Table

To be clear about the scale: the layout work is done and measurably better than `main`, and the one Yellow is a single string literal in a test. Fix that assertion so the guard actually guards and I'll approve on sight — the two Greens are yours to take or leave.

dgershman:
All three addressed in ea0bff54 — and thanks for mutation-testing the assertion rather than eyeballing it; that's exactly the hole.

**🟡 1 (fixed):** Anchored to `css.contains(".nav-select {") && css.contains(".nav-select.nav-selecting")`. `.nav-select {` matches only the base rule at app.css:802 (not `.nav-select:hover {` / `.nav-select.nav-selecting {`), so cutting the `.nav-select { … }` block at :799-808 now drops both halves and the assertion fails — confirmed the predicate bites. Pinned the compound `.nav-select.nav-selecting` specifically, per your point that it's the specificity that keeps the active toggle red over `:hover`.

**🟢 2 (done):** Wrapped the negative `checkSquare` check in `Self.stripComments(String(...))`, matching `:224`/`:288`. Now a future in-body comment naming the invariant can't trip it.

**🟢 3 (done):** Dropped `min-width: 92px` — you're right it's a false floor: it only binds below ~256px, which the fixed 350px grid never reaches, and 92px is under where the centred head clears the refresh anyway, so it "protected" nothing. The 160px ceiling is what does the work. Rewrote the comment to describe what actually renders (a ~160px card at the cap, header slot scaling so a raised min-font-size can't clip it) instead of the compact→expand story that never happens — my own PR comment already conceded the ~160px reality, so the comment just caught up.

Re-verified: `swift test --filter WebTerminalAssetTests` 9/9, and the 350px layout is byte-identical after the min-width removal (card still 160px). Noted your on-the-record data points (the ~1.3px Scorecard headroom at minFont=16 from the +1 pill, and the +11px minFont×3-digit corner) — both inherent to the ticket's ask, leaving as-is per your call.

Re-requesting.

dhilgaertner (changes requested):
## Code & Security Review

Verified by rendering the **real** sidebar-top DOM (`ticketsCard()` + `sidebarToolsStack()` + `navPillRow()` pulled out of `app.js` via the repo's own jsdom loader shape) against the **real** `app.css` in headless Chrome, then measuring boxes — idle / selecting, counts at 1–3 digits, viewports 320→1400px, and simulated raised browser minimum font sizes.

### Critical Issues

None. No Red findings.

### Security Review

**Strengths:**
- Pure presentation change: no new network, auth, storage, or RPC surface.
- The relocated button is built with `el()` + `textContent` + the shared `icon()` helper — no `innerHTML`, no interpolation of daemon-supplied data. Counts still go through `String(n)`/`textContent` (`app.js:1164`).
- The click handler is a verbatim move; it neither widens nor narrows what `selectionMode` can reach.

**Concerns:** none.

### Code Quality

#### 🟡 1. The ticket box ships at 2.4× the width the ticket asks for — and the number you're being asked to confirm isn't the one in the code

`app.css:674` ships `max-width: 160px`. Measured (350px sidebar, five pipeline counts):

| | measured |
|---|---|
| `.tickets-card` | **160.0px** |
| `.sidebar-tools` (Notifications + Settings) | **66.0px** |

CROW-913 ask #2 is *"The ticket box (`.tickets-card`) should **not be wider than those two buttons**."* The card is 94px wider than the cluster it was supposed to be capped to.

The ticket does license implementer judgment ("visual target… Confirm against the running UI"), and the PR body does flag the call — but it flags **92px**, and 92px is not what's in the tree. Round 2 raised the cap to 160px and the body wasn't updated, so the sign-off being requested is on a design that no longer exists. The body also describes *"The head switches to a `space-between` flex row with an in-flow refresh"* — that isn't in the final diff either; `app.css:651` is still `justify-content: center` and `app.css:653-654` still `position: absolute`. That reverted head change is precisely what made 92px viable.

Sweeping the cap against the **shipped** (centred + absolute refresh) head:

| cap | title clears refresh | count rows | card height |
|---|---|---|---|
| 66px (ticket target) | ❌ overlaps | 5 | 123px |
| 92px (PR body's number) | ❌ overlaps | 3 | 89px |
| 108px | ❌ overlaps | 2 | 72px |
| **110px** | ✅ | **2** | **72px** |
| 160px (shipped) | ✅ | 2 | 72px |

**110px renders identically to 160px** — same 2-row counts, same 72px height, header intact — at 50px narrower and half the distance to the ticket's target. So 160px isn't derived from the "legible header" constraint the PR cites; nothing I can measure explains it over 110px. (Restoring the in-flow head brings the knee down to 92px, matching round 1: head stops overflowing at exactly 92px.)

Worth noting the body's stated reason for rejecting 66px doesn't reproduce: `.tickets-title` has no `overflow`/`text-overflow`, so at 66px it measures `titleClipped: false`, visible fraction `1.0` — it does not render as "Tick". What actually breaks at 66px is the absolutely-positioned refresh **overlapping** the title, which is a different (and more fixable) failure.

Please either tighten the cap toward the ticket (110px as-is, ~92px with the in-flow head restored) or update the PR body so the number under discussion matches `app.css:674`.

#### 🟡 2. The ticket's second ask has no regression coverage, and the new guard tests spelling rather than behavior

`WebTerminalAssetTests.swift:312-334` pins ask #1 (the move) but nothing pins ask #2 (the cap). The width is the value that changed in *every* round of this PR — 92px → self-healing range → 160px — and it's the one thing with zero guard. A `#expect(css.contains("max-width: 160px"))`-style anchor on the `.tickets-row .tickets-card` rule, or better a measured bound, would freeze the decision once it's made.

Separately, the moved handler at `app.js:1303` carries real behavior (`selectedSessionIDs.clear()` on cancel), and the guard only proves the string `checkSquare` changed files — drop the `.clear()` and every assertion still passes. `Packages/CrowDaemon/web-tests/` already drives the real `app.js` under jsdom and already does exactly this for the *board's* `nav-selecting` toggle (`board.test.js:225`); a sidebar case there would cover behavior instead of spelling.

#### 🟢 3. `app.js:1265` comment is now false

> "…these two now define the ticket box's width."

They don't: the cluster is 66px and the card is 160px. `max-width: 160px` defines it. This was true-ish in round 1; round 2 invalidated it. Given how carefully the rest of these comments are maintained, worth correcting.

#### 🟢 4. `app.css:670` — "3 at large fonts" doesn't reproduce

Simulating a raised browser minimum font size (16 / 20 / 24px floors on `.tickets-title`, `.tk-n`, `.pill-label`): counts stay at **2 rows** throughout; only the card height grows (72 → 84 → 99 → 114px). No clipping, no title/refresh overlap at any of them — so the *self-healing* claim the comment is defending holds up well; just the "3 rows" parenthetical is wrong.

#### 🟢 5. The negative guard is icon-name-specific

`WebTerminalAssetTests.swift:317-320` fails only if `checkSquare` reappears in `sidebarToolsStack`. A Select toggle re-added there with any other icon passes the guard whose stated purpose is "must not be rebuilt in the sidebar tools stack". Anchoring on `selectionMode` instead would match the intent.

#### 🟢 6. `.nav-select` leans on the UA button padding

`app.css:804-808` sets `width: 30px` with no `padding`, unlike its stated model `.nav-plus` (`app.css:796`, `padding: 6px 0`). Chrome measures the icon at a clean 14×14 inside the 30×29 box, so this is fine today — it's just inheriting a UA default that varies by engine where the neighbouring rule pins it.

#### 🟢 7. Icon-only control with `title` but no `aria-label`

`app.js:1300-1302`. This matches the existing `.tk-tool`/`.nav-plus` pattern exactly, so it's **not a regression** — but the codebase does reach for `aria-label` on other icon-only controls (`app.js:1659`), and this one is now a top-level nav affordance.

### Verified good

- **Behavior preserved exactly.** Active state measures `class="nav-select nav-selecting"`, `color: rgb(255, 69, 58)`, `border-color: rgb(255, 69, 58)`, `title: "Cancel selection"`, ✕ icon; idle measures gold + "Select sessions" + checkSquare. `selectedSessionIDs.clear()` and the `bulkActionBar()` path are untouched.
- **The repaint guard still sees it.** `selectionMode` is in `sidebarSignature()` (`app.js:980`), so `renderSidebar()` isn't short-circuited by the sig check on toggle.
- **The dropped CSS really is dead.** `nav-selecting` is only ever applied by `app.js:1300` (`.nav-select`) and `app.js:3318`/`:3670` (`.action-btn`). Neither `.nav-plus` nor `.tk-tool` has ever received it, so both deletions are safe — and `.action-btn.nav-selecting` (`app.css:531`) is untouched.
- **Layout holds 320 → 1400px.** `Reviews · Allowlist · Scorecard · Select` all fit with no pill-label ellipsis (`scrollWidth == clientWidth` at every width; narrowest pill 71.9px at a 320px viewport), the Select button never overflows the sidebar, the row never overflows, and the card stays at 160px throughout. No media query touches `.tickets-row`/`.nav-pills-row`, so mobile inherits the same geometry cleanly.
- **Counts are robust to magnitude.** 1-, 2- and 3-digit counts all render 2 rows with no overflow.
- **`.nav-select` aligns to the pill height** (30×29 vs. the 29px pill row) via `align-items: stretch` — the stated goal.

### Tests run

| Command | Result |
|---|---|
| `swift test --filter WebTerminalAssetTests` | ✅ 9/9 pass, including the new `selectToggleLivesInNavRowNotTheToolsStack()` |
| `node --check app.js` | ✅ pass |
| `Packages/CrowDaemon/web-tests` (`npm test`) | ⚠️ 10 failures in `row.test.js` (PR-pill glyphs: ✕/✔/◷/⚠/🏷) — **pre-existing**, reproduced on `origin/main` with the same 43 passed / 10 failed. Not from this PR. |

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 5 Green] findings.

The move itself (ask #1) is clean, correct, and well guarded — no notes. What needs another pass is ask #2: the shipped 160px cap is 2.4× the ticket's target and 50px above what its own rationale needs, the PR body asks for sign-off on a 92px design that isn't in the tree, and the width has no test pinning it.

dgershman:
You were right on all counts — I'd drifted the width up to 160px and left the body describing a design I'd reverted. Fixed properly in 8b4ddf74, and this time the width is genuinely near the ticket's target rather than another point on the bad trade curve.

**🟡 1 — root-caused and fixed.** The reason every round landed at ~160px: with the five counts on one line, the card's `max-content` *is* ~160px, so "content-sized" always hit the cap. The fix is to stop the counts being one line — **render them in a fixed 2-column grid**. Now the card's max-content is two cells wide. Measured against the real DOM + real `app.css`:

| counts | this PR | (shipped 160px before) |
|---|---|---|
| 1–2 digit (`7 3 12 4 9`) | **card 91px, h 95px** | 160px |
| 3-digit (`128 346 999 412 157`) | **card 106px, h 95px** | 160px |
| 3-digit @ minFont 24 | card 158px (grows), header intact | 160px |

91px is 69px tighter than what shipped and as close to the 66px cluster as "Tickets" + five counts allow. `max-width:160px` is now a real *ceiling*, not the resting width.

Crucially, the 2-col grid also resolves the two round-1 Yellows that a naive narrow cap would have revived (I verified both — a plain 92px cap clips to "Tic…" at minFont16 and goes 3 rows/95px on 3-digit counts):
- **Header clip:** width tracks max-content, so the box *grows* with the font (→158px at a 24px floor); "Tickets" never clips. Restored the in-flow (space-between) head, since centred + absolute overlaps below ~110px — that's the head the body always described, so the body is true again.
- **Height growth:** 2 columns = constant ~95px for 1/2/3-digit counts (was your round-1 Yellow 2).

**🟡 2 — pinned two ways.** A swift asset test (`ticketBoxIsCompactAndContentSized`) freezes `width: max-content; max-width: 160px` + the 2-col grid. And per your board.test.js:225 pointer, a new jsdom **behaviour** test (`web-tests/sidebar-select.test.js`) drives the real `.nav-select` toggle and asserts it enters selection mode and **clears `selectedSessionIDs` on cancel** — so dropping the `.clear()` now fails a test instead of passing. `npm test` → sidebar-select 7/7 (I `npm install`ed jsdom locally to verify; the 10 row.test.js failures reproduce on main, as you noted).

**🟢 3** — dropped the false "these two define the ticket box's width" comment. **🟢 4** — removed the wrong "3 rows at large fonts" note (with 2 columns it's always 3). **🟢 5** — the move guard now anchors on `selectionMode`, not `checkSquare`. **🟢 6** — `.nav-select` gets explicit `padding: 0`. **🟢 7** — added `aria-label` to the icon-only Select button (matches app.js:1659).

`swift test WebTerminalAssetTests` 10/10. Re-requesting.

dhilgaertner (approved):
## Code & Security Review

Re-review of `8b4ddf7`. **The round-3 Yellow and all five Greens are genuinely fixed**, and the 2-column grid is the right root-cause fix rather than another point on the width curve — I re-measured everything rather than taking the commit message for it. Approving; the five Greens below are yours to take or leave.

**How I verified:** rebuilt the real sidebar-top DOM (`ticketsCard()` / `sidebarToolsStack()` / `navPillRow()` driven out of the real `app.js` via the repo's own jsdom loader shape) and rendered it against the real `app.css` in headless Chrome, measuring boxes at 1-, 2-, 3- and 4-digit counts, idle and selecting, with `--blink-settings=minimumFontSize` at 0/16/24/32/40 — against the **true merge base** (`5424800`, i.e. with the CROW-909 bell) as the baseline. Ran `swift test --filter WebTerminalAssetTests` (**10/10**), the full `web-tests` suite, `node --check app.js`, and mutation-tested all five new assertions. CI is green on all four jobs.

### Round 3 findings — verified resolved

**🟡 1 (width 2.4× the ticket's target) — fixed, and correctly root-caused.** You identified the real reason every round landed at ~160px — with five counts on one line the card's `max-content` *is* ~160px — and fixed the cause instead of re-tuning the cap. Measured at a 350px sidebar:

| | main (`5424800`) | this PR |
|---|---|---|
| `.tickets-card`, 1–2-digit counts | 289px | **91.2px** |
| `.tickets-card`, 3-digit counts | 289px | **105.8px** |
| `.tickets-card`, 4-digit counts | 289px | **120.4px** |
| sidebar-top row height | 95px | **95px** |

91.2px against a 66px tools cluster, and I can confirm 66px is genuinely unreachable with the header intact: at 1–2-digit counts the binding constraint is the *head*, not the counts — title 47px + 4px gap + 22px refresh + 16px padding + 2px borders = 91px exactly. The counts only take over at 3 digits (105.8px). So the number you flagged for confirmation is now the number that renders, and it's the floor for "Tickets" + refresh.

`max-width: 160px` is finally a real ceiling: it doesn't bind until a ~26px minimum-font-size floor.

**🟡 2 (no coverage for ask #2, guard tested spelling) — fixed both ways.** `ticketBoxIsCompactAndContentSized` pins the mechanism, and `web-tests/sidebar-select.test.js` drives the real button under jsdom. I mutation-tested rather than eyeballing — **all five assertions bite**:

| mutation | result |
|---|---|
| drop `selectedSessionIDs.clear()` from the moved handler | ✅ `sidebar-select.test.js` 6/7, fails exactly that assertion |
| re-pin the card to a hard `width: 110px` | ✅ width assertion fails |
| counts grid `repeat(2, auto)` → `repeat(3, auto)` | ✅ grid assertion fails |
| re-add a Select toggle to `sidebarToolsStack` (different icon) | ✅ negative assertion fails — the `selectionMode` anchor works |
| delete/rename the `.nav-select { … }` block | ✅ both halves fail — **round 2's substring hole is closed** |

**🟢 3, 4, 5, 6, 7 — all done.** The false "these two define the ticket box's width" comment is gone; the wrong "3 rows at large fonts" note is gone; the move guard anchors on `selectionMode`; `.nav-select` has explicit `padding: 0`; the icon-only toggle has `aria-label` (verified live: `"Select sessions"` → `"Cancel selection"`).

### Security Review

**Strengths:**
- Pure presentation plus one relocated handler. No network, auth, storage, or RPC surface touched.
- The button is built with `el()` + `textContent` + the shared `icon()` helper — no `innerHTML`, no daemon-supplied data reaching markup. `sel.title` / `aria-label` are set as properties/attributes, not interpolated into HTML. Counts still go through `String(n)` + `textContent` (`app.js:1164`).
- `refresh.onclick`'s `e.stopPropagation()` survives the in-flow head change, so the nested-button semantics inside the card's `selectBoard('tickets')` handler still hold.

**Concerns:** none.

### Verified good

- **Behavior preserved.** Idle: `class="nav-select"`, `rgb(179,142,70)`, checkSquare, `"Select sessions"`. Active: `class="nav-select nav-selecting"`, `rgb(255,69,58)` on both `color` and `border-color`, ✕, `"Cancel selection"`. `selectedSessionIDs.clear()` on cancel is now test-enforced.
- **Height is count-invariant**, exactly as `app.css:677-679` claims: 95px at 1-, 2-, 3- **and** 4-digit counts (I added the 4-digit case), where main's single-line row grew with magnitude. Reading order stays pipeline order (row-major: Backlog·Ready / In Progress·In Review / Done).
- **The self-heal is real, and beats main where it matters.** At a 24px font floor the card grows to 131–158px with the header fully rendered (`scrollWidth == clientWidth`) and never overflows the sidebar; main's card blows out to **375.9px in a 325px row** (`cardRight` 387.9 vs. a 338px inner edge) at the same setting. The failure mode a fixed px width had is gone.
- **The repaint guard still sees it** — `selectionMode` is in `sidebarSignature()` (`app.js:980`), so the toggle's `renderSidebar()` isn't short-circuited.
- **Both CSS deletions really are dead.** `nav-selecting` is only ever applied at `app.js:1300` (`.nav-select`) and `:3319`/`:3671` (`.action-btn`); neither `.nav-plus` nor `.tk-tool` ever receives it. `.action-btn.nav-selecting` (`app.css:531`) is untouched, and `board.test.js`'s `nav-selecting` hits are that board toggle, not this one.
- **`.nav-select` aligns to the pill row** — 30×29 against `.nav-pill` 29px; 30×32 at a 16px floor, 30×42 at 24px. No pill label ellipsises at any width or font floor I tried.

### Tests run

| Command | Result |
|---|---|
| `swift test --filter WebTerminalAssetTests` | ✅ **10/10**, incl. both new cases |
| `web-tests/sidebar-select.test.js` | ✅ 7/7 |
| `web-tests` full suite | ✅ except 10 pre-existing `row.test.js` PR-glyph failures — **reproduced identically (43 passed / 10 failed) on `5424800`'s `app.js`**, not from this PR |
| `node --check app.js` | ✅ |

---

#### 🟢 1. The comment documents the width self-heal but not its vertical cost

`app.css:669-679` explains that the box grows with the font so the header can't clip (correct — measured 158px at a 24px floor). What it doesn't say is what that costs on the other axis. Measured sidebar-top row height, 1–2-digit counts:

| min-font floor | main | this PR |
|---|---|---|
| default | 95px | 95px |
| 16px | 95px | 110px |
| 24px | 95px | 146px |

Main held 95px there only because its row height was pinned by the fixed-px 3-button vertical tools stack — and it paid for that by overflowing horizontally instead (above). This is the trade round 2 accepted, and the axis it grows on is the one that degrades gracefully, so I'm not asking you to change it. But "constant ~95px" in the comment is count-invariance, not font-invariance, and vertical is the scarce axis in a sidebar — one clause would keep the next reader from over-reading it.

#### 🟢 2. Past the cap, the counts grid escapes the card

At font floors beyond ~26px the card pins at `max-width: 160px`, and the counts stop fitting: grid `auto` tracks can't shrink below min-content (a 12px icon + an unbreakable number), so the cells render **past the card's right border** — 3.7px at a 32px floor, 19.6px at 40px — and the title starts ellipsising at 40px (the `text-overflow` you added handles that part cleanly).

**Failure scenario:** a client with "never use font sizes smaller than 32px" opens the web UI; the ticket counts draw outside the card's border box.

Firmly cosmetic and far outside realistic settings — and *better* than main, which overflows the whole sidebar at a 24px floor. Noting it only because the comment's "far-off CEILING" framing stops being true at the cap, and `overflow: hidden` on `.tickets-card` (or letting `max-width` go) would close it if you ever care.

#### 🟢 3. PR body §2 still describes a mechanism that isn't in the tree

The number is finally right — the body's "92px" is what renders (91.2px measured), so the call you're asking me to confirm is a real one, and my answer is **yes, ship 91px**: 66px is unreachable without dropping the word "Tickets", and I'd rather keep the header.

The prose around it is stale, though: "The card is capped to **92px**" (it's `width: max-content` with a *160px* ceiling — 92px is the resting width, not a cap), "wrapped counts" (a fixed 2-column grid — the Verification section further down already says so), and "the tools stack … now laid out horizontally **to define the target width**" (the 66px cluster doesn't define the 91px card — this is the same claim round 3 flagged in the code comment, which you fixed there but not here).

#### 🟢 4. The two new asset assertions aren't anchored to their rules

`WebTerminalAssetTests.swift:349, 352` match `"width: max-content; max-width: 160px"` and `"grid-template-columns: repeat(2, auto)"` against all of `app.css`. Both are unique today and both bite (mutation-tested above), so this is *not* round 2's can't-fail bug — it's just that neither is tied to `.tickets-row .tickets-card` / `.tickets-counts`, so the same declaration appearing on any unrelated rule would satisfy them. I confirmed the gap: hard-pinning the card to `width: 110px` while leaving the identical declaration on a throwaway selector keeps the assertion green. Same one-line fix shape you used for `.nav-select {` — anchor to the selector.

#### 🟢 5. `.nav-select` is a toggle without `aria-pressed`

`app.js:1300-1302`. The flipping `aria-label` does convey state, and no other control in this codebase uses `aria-pressed`, so consistency argues for leaving it — mentioning it only because this is now a top-level nav affordance rather than a tool-stack button.

### Summary Table

Both asks land: the Select toggle is beside Scorecard with behavior now pinned by a real jsdom test, and the ticket box is 3.2× narrower than `main` at no cost in row height, with the header self-healing instead of clipping. Every Green is prose or test-anchoring — nothing that needs to block the merge.

PR #918, opened 2026-07-29
Sidebar top: two-column layout — wide tickets box over nav pills, icon buttons in a far-right column (CROW-917)

### LINKED ISSUE #917: Sidebar top: two-column layout — wide tickets box over the nav pills, icon buttons stacked in a far-right column
Follow-up to #913, which squished the tickets box too far. Reorganize the sidebar top into a clean two-column grid.

## Target layout
```
LEFT column (= width of the nav-pills group)      RIGHT column (one icon wide)
┌───────────────────────────────────────────┐    ┌─────┐
│              Tickets box                   │    │  🔔  │  Notifications
│         (2 button-rows tall)               │    │  ⚙   │  Settings
├───────────┬───────────┬───────────────────┤    ├─────┤
│  Reviews  │ Allowlist │     Scorecard      │    │  +  │  New Manager
├───────────┴───────────┴───────────────────┤    ├─────┤
│                 Manager                    │    │  ☑  │  Select sessions
└───────────────────────────────────────────┘    └─────┘
```

Specifics from the request:
- **Tickets box** spans the **full width of Reviews + Allowlist + Scorecard combined** (and the same width as the **Manager** row). It should be **only ~2 button-rows tall** — i.e. as tall as the Notifications + Settings buttons stacked.
- **Notifications + Settings** are stacked vertically **to the right of the tickets box**.
- Those two share **one far-right column** with the **"+" (new manager)** and **Select-sessions** buttons — all four icon buttons stacked in that single right-hand column, aligned alongside the left rows (tickets = 2 rows, pills row = 1, Manager row = 1 → 4 icons on the right).

## Current code (what to rearrange)
- `renderSidebar` → `.tickets-row` = `ticketsCard()` + `sidebarToolsStack()`, then `navPillRow()` — `app.js:1007-1011`.
- `sidebarToolsStack` (`.sidebar-tools`) = Notifications bell + Settings gear — `app.js:1265-1287` (select was moved out in #913).
- `navPillRow`: **row 1** = Reviews · Allowlist · Scorecard · **Select** (`.nav-select`, `app.js:1300-1305`); **row 2** = **Manager** pill + **"+"** (`.nav-plus`, `app.js:1308-1324`). `ticketsCard` at `app.js:1130`.
- CSS: `.tickets-row`/`.tickets-card`/`.sidebar-tools` `app.css:665-667`; `.nav-pills`/`.nav-pills-row`/`.nav-pill` `app.css:774-784`; `.tk-tool` `app.css:669-675`.

## Implementation direction
- Restructure the sidebar top into **two columns**: a **left** vertical stack of [`ticketsCard` → nav row1 (Reviews/Allowlist/Scorecard) → nav row2 (Manager, full width)], and a **right** vertical stack of the four icon buttons: **bell, gear, "+", select**.
- Pull **Select** out of nav row 1 (`app.js:1300-1305`) and **"+"** out of nav row 2 (`app.js:1320-1323`) and put both in the right icon column together with bell + gear (from `sidebarToolsStack`). Row 1 becomes Reviews/Allowlist/Scorecard only; row 2 becomes **Manager only**, spanning the full left-column width.
- `ticketsCard`: **full width** of the left column (drop the #913 width cap on `.tickets-card`), **height ≈ 2× `.tk-tool` button height** so it lines up with the Notifications+Settings pair beside it.
- Preserve all behavior/state: Select toggles `selectionMode` + `bulkActionBar`; "+" opens `openNewManagerMenu`; bell opens the notification panel with its unread badge; Manager pill keeps its activity dot / RC glyph.

Supersedes the tickets-box width cap and the Select-next-to-Scorecard placement introduced in #913.

## Files
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` — `renderSidebar` (`:1007`), `sidebarToolsStack` (`:1265`), `navPillRow` (`:1290-1327`), `ticketsCard` (`:1130`)
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css` — `.tickets-row`/`.tickets-card`/`.sidebar-tools` (`:646-667`), `.nav-pills`/`.nav-pills-row`/`.nav-pill` (`:774-784`), `.tk-tool`/`.nav-select`/`.nav-plus`

## Verify
Tickets box renders wide (full nav-pills width, = Manager width) and short (~2 button rows); Notifications, Settings, "+", and Select sit stacked in a single far-right column; Reviews/Allowlist/Scorecard on one row and Manager full-width below; all buttons keep their actions.

dgershman:
Closes #917.

Follow-up to #913, which squished the tickets box too far. Reorganizes the sidebar top into a clean two-column grid.

## Layout
- **Left column** (`.sidebar-left`): the **Tickets** card — now **full width** of the left column and **~2 button-rows tall** — over nav **row 1** (Reviews · Allowlist · Scorecard) and nav **row 2** (the **Manager** pill, spanning the full left-column width).
- **Right column** (`.sidebar-right`): the four global icon buttons — **bell**, **gear**, **"+"** (new manager), **Select** — stacked vertically. `.sidebar-top` uses `align-items: stretch` so the right column matches the left height, and each button is `flex: 1 1 0`, so the four icons divide it evenly and self-align to the tickets(2) + Reviews-row(1) + Manager-row(1) = **4 rows**.

## Changes
- Pulled **Select** out of nav row 1 and **"+"** out of nav row 2 into the right icon column, alongside the bell + gear.
- Tickets counts revert to a **single horizontal row of five** (the #913 two-column count grid only existed to keep the width-capped box narrow; the box is full width now).
- Dropped the now-unused `.tickets-row` / `.sidebar-tools` / `.nav-pills` wrapper classes and the `navPillRow` / `sidebarToolsStack` functions; renamed to `sidebarLeftStack` / `sidebarIconColumn`.
- **Supersedes** the tickets-box width cap and the Select-next-to-Scorecard placement from #913.

## Behavior (preserved verbatim)
Bell → notification panel + unread badge · gear → Settings · "+" → new-manager menu · Select → toggles `selectionMode` (clears selection on cancel, red while active) · Manager pill keeps its activity dot / RC glyph · Reviews unseen badge · tickets refresh spinner.

## Verify
- `node --check app.js` passes; every class the JS emits maps to a live CSS rule (no dangling `tickets-row`/`sidebar-tools`/`.nav-pills`).
- Tickets box renders wide (full nav-pills width, = Manager width) and short (~2 button rows); Notifications, Settings, "+", Select sit stacked in a single far-right column; Reviews/Allowlist/Scorecard on one row and Manager full-width below; all buttons keep their actions.

> Coordinated at merge time with #913 (superseded) and #909 (notification bell), both of which touch this sidebar region.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -683,8 +683,12 @@ html, body {
 .sidebar-right { flex: 0 0 auto; width: 30px; display: flex; flex-direction: column; gap: 6px; }
 /* Every icon button in the right column stretches to the column width and shares
    the height evenly, so bell/gear/+/select read as one aligned stack regardless of
-   each control's intrinsic size. */
-.sidebar-right > button { flex: 1 1 0; width: 100%; height: auto; }
+   each control's intrinsic size. `min-height: 24px` is a hard floor: the four
+   flex:1 buttons divide whatever height `.sidebar-left` resolves to, and a short
+   left column (no Manager pill / cold-start render) would otherwise divide to
+   ~20px — under the WCAG 2.2 §2.5.8 target-size minimum, and the sidebar is the
+   full-width touch view on mobile (CROW-917 review). */
+.sidebar-right > button { flex: 1 1 0; width: 100%; height: auto; min-height: 24px; }
 .tickets-counts { display: flex; align-items: center; justify-content: space-between; gap: 4px; padding: 0 2px; }
 .tk-tool {
   display: inline-flex; align-items: center; justify-content: center;
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -644,7 +644,11 @@ html, body {
 
 .tickets-card {
   background: var(--bg-surface); border: 1px solid var(--border-subtle);
-  border-radius: 8px; padding: 8px; margin: 2px 2px 8px; cursor: pointer;
+  border-radius: 8px; padding: 8px; cursor: pointer;
+  /* Full width of the left column, ~2 button-rows tall so it lines up with the
+     Notifications+Settings pair in the right icon column (CROW-917). */
+  margin: 0; box-sizing: border-box; height: 64px;
+  display: flex; flex-direction: column; justify-content: space-between;
 }
 .tickets-card:hover { border-color: var(--border-strong); }
 .tickets-card.selected { border-color: var(--gold); }
@@ -665,21 +669,18 @@ html, body {
    `.action-spinner` glyph swapped in for the `↻` turns inside it (CROW-797) —
    matching the detail-header action-button spinner (CROW-749). */
 .tickets-refresh.spinning { opacity: 0.6; cursor: default; }
-.tickets-row { display: flex; align-items: stretch; gap: 6px; margin: 2px 2px 8px; }
-/* Compact ticket box (CROW-913 ask #2): was flex:1 1 auto, which grew it to fill the whole
-   row and dominate the sidebar top. Now content-sized: the counts render in a fixed TWO
-   columns (below), so the card's max-content is a narrow ~91px (two count cells + header)
-   rather than the five-cell-wide ~160px it would be on one line — measured 91px at typical
-   counts, 106px with 3-digit counts, as close to the Notifications+Settings cluster (66px)
-   as "Tickets" + the five counts allow. `max-width` is a far-off CEILING, not the resting
-   width: because the width tracks max-content, the box grows with the font, so a raised
-   browser minimum-font-size can't clip the header (grows to ~158px at a 24px floor) — the
-   failure a fixed px width had (review). Two columns also fix the height: it's a constant
-   ~95px regardless of 1-, 2-, or 3-digit counts, where a single wrapping row grew taller
-   with bigger numbers. Sits beside the horizontal Notifications+Settings tools cluster. */
-.tickets-row .tickets-card { flex: 0 1 auto; width: max-content; max-width: 160px; margin: 0; }
-.sidebar-tools { display: flex; flex-direction: row; align-items: center; gap: 6px; flex: 0 0 auto; }
-.tickets-counts { display: grid; grid-template-columns: repeat(2, auto); justify-content: center; gap: 4px 10px; margin-top: 8px; padding: 0 2px; }
+/* Two-column sidebar top (CROW-917): left stack (tickets over the nav rows) beside
+   a one-icon-wide right column. `align-items: stretch` makes the right column match
+   the left column's height so its four icon buttons (flex:1 each) divide it evenly
+   and self-align to the tickets(2) + Reviews-row(1) + Manager-row(1) = 4 rows. */
+.sidebar-top { display: flex; align-items: stretch; gap: 6px; margin: 2px 2px 8px; }
+.sidebar-left { flex: 1 1 auto; min-width: 0; display: flex; flex-direction: column; gap: 6px; }
+.sidebar-right { flex: 0 0 auto; width: 30px; display: flex; flex-direction: column; gap: 6px; }
+/* Every icon button in the right column stretches to the column width and shares
+   the height evenly, so bell/gear/+/select read as one aligned stack regardless of
+   each control's intrinsic size. */
+.sidebar-right > button { flex: 1 1 0; width: 100%; height: auto; }
+.tickets-counts { display: flex; align-items: center; justify-content: space-between; gap: 4px; padding: 0 2px; }
 .tk-tool {
   display: inline-flex; align-items: center; justify-content: center;
   width: 30px; height: 29px; border-radius: 6px; cursor: pointer;
@@ -782,9 +783,9 @@ html, body {
 .tk-dot { width: 7px; height: 7px; border-radius: 50%; }
 .tk-n { font-size: 11px; font-weight: 600; font-variant-numeric: tabular-nums; }
 
-/* Two deterministic rows: Reviews/Allowlist/Scorecard, then Manager + "+".
-   Each .nav-pills-row is its own non-wrapping flex line so groups never reflow. */
-.nav-pills { display: flex; flex-direction: column; gap: 6px; margin: 0 2px 8px; }
+/* Left-column nav rows (CROW-917): row 1 Reviews/Allowlist/Scorecard, row 2 the
+   full-width Manager pill. Each .nav-pills-row is its own non-wrapping flex line so
+   groups never reflow; the "+" and Select buttons live in the right icon column. */
 .nav-pills-row { display: flex; gap: 6px; }
 .nav-pill {
   flex: 1 1 auto; min-width: 0; display: flex; align-items: center; justify-content: center; gap: 5px;
@@ -799,14 +800,15 @@ html, body {
 .pill-dot { width: 8px; height: 8px; border-radius: 50%; flex: 0 0 auto; }
 .pill-dot.pulse { animation: crow-pulse 1.5s ease-in-out infinite; }
 .nav-plus {
-  flex: 0 0 auto; width: 30px; padding: 6px 0; border-radius: 6px; cursor: pointer; line-height: 1;
+  flex: 0 0 auto; width: 30px; padding: 0; border-radius: 6px; cursor: pointer; line-height: 1;
+  display: inline-flex; align-items: center; justify-content: center;
   background: var(--bg-surface); border: 1px solid rgba(179, 142, 70, 0.3); color: var(--gold-dark); font-size: 15px; font-weight: 700;
 }
 .nav-plus:hover { color: var(--gold); border-color: var(--border-strong); }
 
-/* Select-sessions toggle (CROW-913): an aligned icon button living in nav-pills row 1
-   after Scorecard. Sized/coloured like .nav-plus so it stretches to the pill height;
-   reads red while a selection is active, matching .nav-selecting elsewhere. */
+/* Select-sessions toggle (CROW-913 → far-right icon column, CROW-917): sized/coloured
+   like .nav-plus; reads red while a selection is active, matching .nav-selecting elsewhere.
+   In .sidebar-right it flex-fills to align under the bell/gear/+ stack. */
 .nav-select {
   flex: 0 0 auto; width: 30px; padding: 0; display: inline-flex; align-items: center; justify-content: center;
   border-radius: 6px; cursor: pointer;
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -646,8 +646,13 @@ html, body {
   background: var(--bg-surface); border: 1px solid var(--border-subtle);
   border-radius: 8px; padding: 8px; cursor: pointer;
   /* Full width of the left column, ~2 button-rows tall so it lines up with the
-     Notifications+Settings pair in the right icon column (CROW-917). */
-  margin: 0; box-sizing: border-box; height: 64px;
+     Notifications+Settings pair in the right icon column (CROW-917). `min-height`
+     not `height`: a raised browser minimum-font-size grows the title/counts past
+     the resting 46px content box, and a hard height would clip the counts row out
+     the bottom border (the #913 fixed-px failure mode). Growing is safe — the
+     right column stretches to whatever the left resolves to, re-deriving the
+     4-icon alignment for free. */
+  margin: 0; box-sizing: border-box; min-height: 64px;
   display: flex; flex-direction: column; justify-content: space-between;
 }
 .tickets-card:hover { border-color: var(--border-strong); }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -1004,11 +1004,13 @@ function renderSidebar() {
   brand.alt = 'Crow';
   root.appendChild(brand);
 
-  const trow = el('div', 'tickets-row');
-  trow.appendChild(ticketsCard());
-  trow.appendChild(sidebarToolsStack());
-  root.appendChild(trow);
-  root.appendChild(navPillRow());
+  // Two-column sidebar top (CROW-917): a left stack [tickets → Reviews/Allowlist/
+  // Scorecard → Manager] over a far-right column of four stacked icon buttons
+  // [bell, gear, +, select], which flex evenly to align alongside the left rows.
+  const top = el('div', 'sidebar-top');
+  top.appendChild(sidebarLeftStack());
+  top.appendChild(sidebarIconColumn());
+  root.appendChild(top);
   if (selectionMode) root.appendChild(bulkActionBar());
 
   // Cold start only: structured skeleton rows so the left pane isn't blank
@@ -1260,11 +1262,12 @@ function renderStatusBar() {
   }
 }
 
-// Notifications + Settings, side by side to the right of the Tickets card
-// (outside the box). The Select-sessions toggle moved down to navPillRow row 1,
-// next to Scorecard (CROW-913).
-function sidebarToolsStack() {
-  const stack = el('div', 'sidebar-tools');
+// Far-right sidebar column (CROW-917): the four global icon buttons stacked
+// vertically — Notifications bell, Settings gear, "+" new-manager, and the
+// Select-sessions toggle. They flex evenly to align alongside the left stack's
+// rows (tickets = 2 rows, Reviews-row = 1, Manager-row = 1 → 4 icons).
+function sidebarIconColumn() {
+  const col = el('div', 'sidebar-right');
   // Notification center (CROW-909): bell + unread badge, first so it's the most
   // prominent global affordance. Visible in every view (sessions and boards).
   const bell = el('button', 'tk-tool tk-bell');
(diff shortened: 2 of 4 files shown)


dhilgaertner (changes requested):
## Code & Security Review

The layout itself is well-reasoned and the dead-CSS cleanup is thorough — I verified every removed class (`.tickets-row`, `.sidebar-tools`, `.nav-pills`) has no remaining emitter, and every class the new JS emits maps to a live rule. But the two function renames were not propagated to the test suite, which breaks it in two places.

### Critical Issues

**🔴 The `navPillRow` / `sidebarToolsStack` renames break two test suites.** Both are confirmed by running them, not inferred.

**1. `Packages/CrowDaemon/web-tests/sidebar-select.test.js:17,49`** — the jsdom harness reaches into the real `app.js` for `navPillRow`, which no longer exists:

```
$ npm test          # in Packages/CrowDaemon/web-tests
Select toggle — idle → active:
app.js:4929
  navPillRow(){ return navPillRow(); },
                ^
ReferenceError: navPillRow is not defined
    at selBtn (sidebar-select.test.js:49:24)
```

The script dies before its first `check()`, so `npm test` exits non-zero and the six preceding suites' green is the last thing you see.

**2. `Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift:315-338`** — `functionBody` resolves `"function <name>("` via `#require`, so a missing function throws rather than returning empty:

```
$ swift test --filter selectToggleLivesInNavRowNotTheToolsStack
↳ no sidebarToolsStack
✘ Test selectToggleLivesInNavRowNotTheToolsStack() failed after 0.009 seconds with 1 issue.
```

Both `functionBody("sidebarToolsStack", …)` (line 320) and `functionBody("navPillRow", …)` (line 323) are now unresolvable.

Worth flagging explicitly: **the PR CI lane will not catch either of these.** `CrowDaemon` is absent from `LINUX_PACKAGES` in `.github/workflows/ci.yml:58-59` (it depends on the Darwin-only `CrowTelemetry`), and the `macos-15` job there only builds the Tauri desktop. `release.yml:13-35` *does* run `swift test` across every package on macOS — so this lands green on the PR and goes red at release time.

The Swift test is **not** a mechanical rename. It pins the CROW-913 invariant "the Select toggle must live in the nav row and *not* in the tools stack" — an invariant this PR deliberately reverses by moving Select back into the (renamed) icon column. Renaming the two string arguments would invert the assertion into a guaranteed failure. It needs rewriting to pin the CROW-917 invariant instead: Select's `selectionMode` logic lives in `sidebarIconColumn`, and `sidebarLeftStack` contains the nav rows but no selection logic. The `.tk-tool.nav-selecting` / `.nav-select {` / `.nav-select.nav-selecting` CSS assertions (lines 327-337) are still correct as written and should be kept.

The JS test *is* a mechanical rename (`navPillRow` → `sidebarIconColumn`, since that is where `.nav-select` now renders) — but its header comment and the `.nav-select renders in nav-pills row 1` check name (line 57) describe the old placement and should follow.

### Code Quality

**🟡 `app.css:650` — `height: 64px` on `.tickets-card` reintroduces the failure mode #913's review designed the box around.** The comment block this PR deletes said it explicitly: *"a raised browser minimum-font-size can't clip the header … the failure a fixed px width had (review)."* The new rule swaps a fixed cap on the width axis for a hard `height` on the vertical one.

At default sizing the card is fine — content box is `64 − 16 (padding) − 2 (border) = 46px`, holding a 22px `.tickets-head` (the refresh button's height) over a ~14px `.tickets-counts` row, ~10px of slack. Raise the browser's minimum font size to ~24px and `.tickets-title` (13px) and `.tk-n` (11px) both scale up, pushing the two rows past 46px. `.tickets-card` has no `overflow`, and an overflowing `justify-content: space-between` column falls back to `flex-start`, so the counts row spills out the bottom border and collides with the 6px-away nav row 1. `.tickets-title`'s `text-overflow: ellipsis` doesn't help — this overflows on the block axis.

`min-height: 64px` gives the identical resting appearance (the right column stretches to whatever the left column resolves to, so the 4-icon alignment re-derives for free) without the clip.

**🟡 `app.js:1321-1332` — `row2` is appended even when there is no primary manager, which collapses the icon column.** `sidebarLeftStack` builds `row2` unconditionally and only conditionally fills it, so with no manager session it is an empty `.nav-pills-row` — zero height, but `.sidebar-left`'s `gap: 6px` still applies between it and row 1.

Under the old code this could not happen: row 2 always held the `+` button. Now the left column loses a full pill row plus keeps a stray gap, and since `.sidebar-right` stretches to the left column and its children are `flex: 1 1 0`, all four icons shrink with it:

| state | left column | each icon button |
|---|---|---|
| manager present | `64 + 6 + 28 + 6 + 28 = 132px` | `(132 − 18) / 4 ≈ 28.5px` ✅ |
| no manager | `64 + 6 + 28 + 6 + 0 = 104px` | `(104 − 18) / 4 ≈ 21.5px` ❌ |

At 30 × 21.5px the buttons fall under the 24 × 24 CSS px floor of WCAG 2.2 §2.5.8 (Target Size, Minimum) — and on mobile the sidebar is the full-width view (`app.css:455-468`), so that is the touch target. It also breaks the "4 rows ↔ 4 icons" premise the whole PR is built on.

This is not only the empty-workspace case: `sidebar-top` is appended at `app.js:1010-1013`, *before* the cold-start skeleton early-return at `app.js:1019-1023`. So on a cold start with no cache, `sessions` is still empty and the icon column renders squashed, then jumps ~7px per button once `list-sessions` lands. Guarding the append (`if (primaryManager) wrap.appendChild(row2)`) fixes both.

**🟢 `app.js:1273-1290` — inconsistent accessible names across the new stack.** `sel` sets both `title` and `aria-label` (line 1296), while `bell` and `gear` are icon-only with `title` alone, and `plus`'s accessible name is the literal `"+"`. `title` is not a reliable accessible name on all AT. Mostly pre-existing, but these four are now a homogeneous stack where one of four is labelled — cheap to make uniform while you're here.

### Security Review

**Strengths:**
- Purely presentational: no new network calls, RPC verbs, auth paths, or persisted state.
- All text goes through `el(tag, class, text)` → `textContent` (`app.js:967-972`), so ticket counts, unread counts, and titles cannot inject markup. No new `innerHTML` sink; the only assignment is the pre-existing `root.innerHTML = ''` clear.
- The `+` button moving to the far right does not create an off-viewport menu — `openNewManagerMenu` clamps with `Math.min(rect.left, innerWidth − menu.offsetWidth − 8)` and `Math.max(4, …)` (`app.js:1399-1402`).
- The bell's `.notif-badge` overhangs its button by 5px; with `#sidebar`'s `padding: 12px 10px 44px` and `.sidebar-top`'s 2px right margin it stays inside the padding box, so `overflow-y: auto` does not clip it despite the button now sitting at the sidebar's right edge.

**Concerns:** none — no security-relevant surface in this diff.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 2 Yellow, 1 Green] findings.

The Red is the blocker and is mechanical apart from rewriting the Swift test's intent. Please run `swift test --package-path Packages/CrowDaemon` and `npm test` in `Packages/CrowDaemon/web-tests` before the next push — the PR CI lane covers neither.

---

dhilgaertner (changes requested):
## Code & Security Review

Pure front-end layout refactor of the sidebar top. The mechanical part of this PR is clean: I grepped the whole tree for `sidebarToolsStack` / `navPillRow` / `.nav-pills` / `.sidebar-tools` / `.tickets-row` and there are **zero dangling references** — every removed class and function is gone from JS, CSS, and both test layers, and every class the JS still emits maps to a live rule. The four moved controls keep their handlers verbatim, and three of them **gained** `aria-label`s they didn't have before.

### Verification performed

| Check | Result |
|---|---|
| `node --check app.js` | pass |
| `swift test --filter WebTerminalAssetTests` | **10/10 pass** (incl. both rewritten tests) |
| `web-tests/sidebar-select.test.js` | 7/7 pass |
| `board` / `key-handler` / `sidebar-groups` / `touch-scroll` / `wheel-scroll` | 200/200 pass |
| `web-tests/row.test.js` | 43 pass, **10 fail** — see note below |

> **`row.test.js` is not this PR's fault.** I built a detached worktree at `origin/main` and ran it there: the same 10 checks fail identically (PR-status glyph assertions). Pre-existing breakage, flagged only so it isn't mistaken for a regression here — but it does mean this suite is currently not guarding anything.

---

### Yellow — the icon column has no height floor, and the comment that says otherwise is inverted

`app.css:687` sizes all four icon buttons purely by division:

```css
.sidebar-right > button { flex: 1 1 0; width: 100%; height: auto; }
```

With `align-items: stretch`, the right column's height is whatever `.sidebar-left` resolves to, and the four buttons split it. The arithmetic works out exactly as the PR describes **when a Manager pill exists**:

```
left = 64 (card min-height) + 6 + 29 (row1) + 6 + 29 (row2) = 134
btn  = (134 − 3×6) / 4 = 29px      ✅ matches the old .tk-tool height: 29px
```

But `sidebarLeftStack` only appends row 2 `if (primaryManager)` (`app.js:1330`), and row 1 is unconditional, so the no-Manager branch is:

```
left = 64 + 6 + 29 = 99
btn  = (99 − 3×6) / 4 = 20.25px    ❌ below the 24×24 WCAG 2.2 AA target floor
```

Nothing clamps this — `height: auto` overrides `.tk-tool { height: 29px }`, and the flex `min-height: auto` content floor is only ~16–18px (a 14px icon plus border), so 20.25px passes right through it.

The comment at `app.js:1323–1328` claims skipping the row is what *prevents* this:

> *"an empty row would drop a pill's worth of height (plus a stray 6px gap) off the left column, squashing the right icon column's four flex:1 buttons below the 24px WCAG target floor"*

That reasoning is backwards. An empty `.nav-pills-row` has zero height, so appending it costs a pill's height **and** adds 6px of gap → 105px. Skipping it costs the pill's height **and** the gap → 99px. Skipping is the *shorter* of the two by 6px, so it can't be the thing that keeps the buttons above 24px — and at 20.25px they aren't above it either way. Skipping the row is still the right call (the stray gap is real), but the stated justification is wrong, and this codebase's comments are load-bearing enough that a future reader will trust it.

The PR text itself names the reachable trigger: the cold-start no-cache render, where `sessions` is still empty while the first state payload is in flight. It's brief, but it's the exact state the comment was written to cover.

**Fix — one line, and it hardens the layout against any future change to the left column:**

```css
.sidebar-right > button { flex: 1 1 0; width: 100%; height: auto; min-height: 24px; }
```

Then please correct or drop the inverted rationale at `app.js:1323–1328`.

*Caveat on method:* this is derived from the CSS, not measured — `file://` is blocked for the browser extension here and headless Chrome hung on this machine, so I couldn't render it. The derivation reproduces the PR's own 4×29px figure for the Manager case exactly, which is what gives me confidence in the 20.25px for the other branch.

---

### Security Review

**Strengths:**
- No security surface touched. The diff is DOM construction only — no network, auth, storage, or daemon-side code.
- All text goes through `el()` (`app.js:967`), which assigns `textContent`, never `innerHTML`. The unread badge wraps its value in `String(unread)` and caps at `'99+'`. No injection path.
- Moving the bell and "+" to the sidebar's right edge does **not** create off-screen menus: both `openNewManagerMenu` (`app.js:1407`) and `openNotificationPanel` (`app.js:1482`) clamp with `Math.min(rect.left, window.innerWidth - menu.offsetWidth - 8)`.

**Concerns:** none.

### Code Quality

**Green — considered and found fine, no action needed:**
- **Counts row won't overflow.** Reverting `.tickets-counts` to a single flex row of five is safe at the 350px sidebar: ~272px of card content width against ~150–195px of cells even with 3-digit counts. The #913 two-column grid really was only needed for the width-capped box.
- **`min-height` over `height` on `.tickets-card`** is the right call, and the comment explaining why (raised browser min-font-size would clip the counts row past a hard height) is accurate.
- **Badge clipping.** `.notif-badge` sits at `right: -5px` and the bell is now flush right, but `#sidebar` has `padding: 12px 10px` — 10px of room, so no clip.
- **Test string-pinning** (`css.contains(".tickets-counts { display: flex;")`) is brittle against reformatting, and `min-height: 64px` is a whole-file substring that would still match if the declaration migrated to another rule. Both match the existing conventions in `WebTerminalAssetTests.swift`, so this is consistency, not a new problem.
- The `sidebar-select.test.js` rewrite correctly keeps driving real behaviour (toggle + `selectedSessionIDs.clear()`) rather than degrading into a string guard.

### Summary Table

The one Yellow is a two-line change (add `min-height: 24px`; fix the inverted comment). Everything else here is solid work — the dead-class cleanup is thorough and the a11y labels are a genuine improvement.

dhilgaertner (changes requested):
## Code & Security Review

Pure presentation change to the web sidebar (CSS + DOM assembly). No RPC surface, no daemon logic, no auth path, no new data flow. Reviewed the geometry math, the DOM/CSS contract, and the test pinning.

### Critical Issues

None.

### Security Review

**Strengths:**
- No new input handling, no `innerHTML` on any moved node — `el()` sets `textContent`, and the two new labels go through `setAttribute('aria-label', …)`. The bell's label is built from `notifUnreadCount()` (a number), so nothing attacker-controlled reaches the DOM.
- No change to auth, session identity, RPC payloads, or `signedInOverWeb()`. `openNewManagerMenu` / `selectBoard` / `selectSession` are called with the same arguments as before — the buttons only changed parent.
- Accessibility improved: `aria-label` added to the bell and gear, which previously carried `title` only.

**Concerns:** none found.

### Code Quality

Verified in the worktree:
- `node --check app.js` passes; `swift test` on `Packages/CrowDaemon` is **324 tests / 54 suites, all green**, including the two rewritten pins (`ticketBoxIsFullWidthAndShort`, `selectToggleLivesInIconColumnNotTheLeftStack`).
- `sidebar-select.test.js` (jsdom, drives the real handler) passes 7/7 against the relocated `.nav-select`.
- No dangling references: `rg` finds zero remaining uses of `tickets-row`, `sidebar-tools`, `.nav-pills`, `navPillRow`, or `sidebarToolsStack` in the repo. Every class the JS emits maps to a live rule.
- Specificity is correct: `.sidebar-right > button` (0,1,1) beats `.tk-tool` / `.nav-plus` / `.nav-select` (0,1,0), so the column's `flex: 1 1 0; width: 100%; height: auto` wins over their intrinsic `width: 30px; height: 29px` regardless of source order.
- The 64px derivation checks out: with 29px nav rows, left column = 64 + 6 + 29 + 6 + 29 = 134px, so four `flex:1` buttons over three 6px gaps land at exactly 29px each — the same height `.tk-tool` used before. The tickets card then spans exactly 2 buttons + 1 gap = 64px.
- The `row2` guard (app.js:1330) is right — an empty `.nav-pills-row` would still consume a flex-gap slot, so only appending it when a primary manager exists is the correct fix.
- `sidebarSignature()` already carries `selectionMode`, so the toggle's `renderSidebar()` still repaints past the early-return.

**Yellow — the review-mandated WCAG floor and the "+" relocation are both unpinned (app.css:691, app.js:1289-1293).**

This file's convention is to pin layout decisions with string assertions — this PR itself adds three (`min-height: 64px`, `.tickets-counts { display: flex;`, the superseded `max-width: 160px`). Two decisions of equal weight got none:

1. `.sidebar-right > button { … min-height: 24px }` (app.css:691) exists **only** because a prior review round flagged WCAG 2.2 §2.5.8 for the Manager-less / cold-start render, where the four `flex:1` buttons otherwise divide a 99px column into ~20px targets. Nothing pins it.
2. Moving `"+"` out of nav row 2 into the icon column is half of this PR's stated purpose. `selectToggleLivesInIconColumnNotTheLeftStack` only asserts on `.nav-select`, and the jsdom test only queries `.nav-select`.

Confirmed empirically, not inferred — I deleted `min-height: 24px` from app.css:691 **and** deleted `col.appendChild(plus)` from app.js:1293, then re-ran both suites:

```
swift test --filter WebTerminalAssetTests  → 10 tests, 1 suite, all passed
node sidebar-select.test.js                → 7 passed, 0 failed
```

Green on both. So the accessibility floor can be dropped and the new-manager button can vanish from the UI entirely without a single test noticing. Two lines in `WebTerminalAssetTests` close it:

```swift
#expect(css.contains("min-height: 24px"),
        "the right column's icon buttons need the 24px WCAG 2.5.8 target-size floor (CROW-917 review)")
#expect(try Self.functionBody("sidebarIconColumn", in: js).contains("'nav-plus'"),
        "the new-manager \"+\" lives in the icon column, not the nav rows (CROW-917)")
```

**Green — dead declarations in `.nav-plus` / `.nav-select` (app.css:812, 822).** Both still carry `flex: 0 0 auto; width: 30px`, which `.sidebar-right > button` now overrides at their only two call sites. Harmless, but it reads as if the sizing still lives there.

**Green — no-Manager render leaves ~15px of dead space.** With no Manager pill the right column's 24px floor makes it the taller item (4×24 + 3×6 = 114px vs. the left column's 99px), so `align-items: stretch` grows `.sidebar-left` to 114px with its content top-aligned. The tradeoff is documented in the comment and is clearly preferable to sub-24px tap targets; noting it only so the cold-start gap isn't read later as a bug.

**Green — no `CHANGELOG.md` bullet.** `CONTRIBUTING.md:113` asks for one under `[Unreleased]`. Consistency note only: #899, #909, #912, and #914 all merged without one, so this is a repo-wide drift, not a regression this PR introduced.

### Not a finding — pre-existing test failure

`npm test` in `Packages/CrowDaemon/web-tests` reports **43 passed, 10 failed** in `row.test.js` (PR-status glyph assertions). I ran the same test against `origin/main`'s `app.js` in isolation and got an identical 43/10, so it is **pre-existing and unrelated** to this PR — but it does mean the web-test suite exits non-zero on `main` and can no longer catch a real break.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 3 Green] findings. The layout, the geometry math, and the behavior preservation are all sound; the one blocker is that the accessibility floor a previous round asked for, and the `"+"` relocation this PR is half about, can both regress silently. Add the two `#expect`s and this is a clean approve.

dhilgaertner (approved):
## Code & Security Review

Pure client-side sidebar layout change (`app.css` + `app.js`) with both pinning suites updated in step. I read the full diff, checked the CSS cascade by hand, and ran every suite locally.

### Critical Issues

None.

### Security Review

**Strengths:**
- **No new attack surface.** The change is DOM structure + CSS only — no new RPC calls, no new user-controlled input paths, no `innerHTML` sinks. Every node is built through `el()` (`app.js:967`), which assigns `textContent`, so the ticket counts (`String(n)`) and the unread badge (`String(unread)`) cannot inject markup.
- `cell.style.color` still comes from the static `TICKET_STATUS_COLOR` map keyed by a literal `statusKey`, not from server data (`app.js:1164`) — unchanged and not attacker-influenced.
- The relocated `"+"` keeps `openNewManagerMenu(plus)` unchanged, and that menu already clamps to the viewport (`app.js:1407-1410`), so moving the anchor to the sidebar's far-right edge can't push the menu off-screen.
- The notification badge's `-5px` overhang still lands inside `#sidebar`'s 10px right padding (`app.css:49`) plus `.sidebar-top`'s 2px margin, so `overflow-y: auto` won't turn it into a stray scroll axis.
- **Accessibility improves:** four explicit `aria-label`s added (bell/gear/"+"/Select), and `.sidebar-right > button { min-height: 24px }` puts a hard floor under WCAG 2.2 §2.5.8 target size — a real fix, not just a comment.

**Concerns:** none.

### Code Quality

The cascade holds up. `.sidebar-right > button` is specificity (0,1,1) against `.tk-tool` / `.nav-plus` / `.nav-select` at (0,1,0), so the per-control `width: 30px`, `height: 29px`, and `flex: 0 0 auto` are all correctly overridden regardless of rule order. The `flex: 1 1 0` + `min-height: 24px` sizing resolves cleanly: the right column's hypothetical height (4×24 + 3×6 = 114px) participates in `.sidebar-top`'s cross-size max, so the taller of the two columns wins and the icons re-divide — no circular-sizing trap.

Cleanup is complete: `rg` over the whole repo finds **zero** remaining references to `tickets-row`, `sidebar-tools`, `.nav-pills`, `navPillRow`, or `sidebarToolsStack` — no dangling classes, no dead CSS.

**Verification run locally:**
- `node --check app.js` — passes.
- `swift test --filter WebTerminalAssetTests` — **11 tests, all pass**, including the three new/updated pins.
- web-tests: `board` 75/0, `key-handler` 29/0, `sidebar-groups` 20/0, `sidebar-select` 7/0, `touch-scroll` 26/0, `wheel-scroll` 50/0.
- `row.test.js` reports **43 passed / 10 failed** — but these are **pre-existing on `main`, not caused by this PR**: I re-ran it against `origin/main`'s `app.js` and got the identical 43/10. All ten are PR-pill glyph assertions (`✕`/`✔`/`◷`/`⚠`/`🏷`), untouched here.

Green findings, all optional:

1. **`app.css:691` — the icons divide evenly, they don't line up row-for-row.** The PR body says the four icons "self-align to the tickets(2) + Reviews-row(1) + Manager-row(1) = 4 rows." With the Manager pill present the left column resolves to ~132px, which the four `flex: 1 1 0` buttons split into ~33px each, while the left rows are 64 / ~29 / ~29. So the gear's bottom edge sits several px below the tickets card's, and the boundaries drift from there. The CSS comment ("divide it evenly") is the accurate description; it's the PR body that over-claims. Purely cosmetic — worth reconciling the prose so the next reader doesn't hunt for a bug.

2. **`app.js:1326-1332` / `app.css:691` — cold-start and Manager-less renders leave dead space.** With no Manager pill the left column is ~98px, below the right column's 114px floor, so `.sidebar-top` stretches to 114 and the left column carries ~15px of empty space under the Reviews row while the icon stack runs past it. This is the documented WCAG trade-off working as intended, and it's transient (the skeleton path at `app.js:1019` still renders `top` first), so nothing to fix — just flagging that the state is reachable.

3. **`WebTerminalAssetTests.swift:352` — a JS comment-stripper applied to CSS.** `stripComments` is built for JS, and its `//` line pass truncates `app.css:101` at the `http://` inside the checkbox data-URI. It's harmless for `contains("min-height: 24px")` (that rule is nowhere near line 101, and the block-comment pass — the part you actually needed — is correct), but it's a latent trap for whatever CSS assertion gets added on the stripped text next. A one-line note on the helper, or a CSS-only variant, would defuse it.

4. **`app.css:655` — redundant `box-sizing: border-box`.** Already set globally at `app.css:24`. Harmless; drop it if you're touching the rule again.

5. **Neither suite this PR touches runs in CI** — `CrowDaemon` is absent from `ci.yml`'s `LINUX_PACKAGES` (Darwin-only `CrowTelemetry` dep), and the node web-tests aren't wired up at all (`ci.yml:76-77` explains why). Pre-existing and out of scope, but it does mean the three new pins only fire locally or on release tags. I ran both suites by hand for this review, and the mutation-testing rationale in the new test doc comments is exactly the right response to that gap.

### Summary Table

The layout math is sound, the cleanup is complete, the behavior is preserved (proven by the jsdom test, not just a string guard), and accessibility genuinely improved. The Green items are prose/tidiness, not defects.

PR #919, opened 2026-07-29
Web terminal: send CSI-u for Shift+Enter so agents see a newline (CROW-916)

### LINKED ISSUE #916: Shift+Enter does not insert a newline in the web xterm.js terminal (the #599 fix lives on the retired macOS surface)
## Symptom

In Crow's browser terminal, **Shift+Enter submits the prompt instead of inserting a newline** in Claude Code. `/terminal-setup` cannot help — it patches *native* terminal keymaps (iTerm2, Apple Terminal, VS Code, Zed, Alacritty), and Crow's only client is xterm.js in a browser, which it has no notion of. It also refuses to run inside tmux.

## Root cause

The Shift+Enter interception exists, but **only on the surface that no longer ships.**

`#599` (closes `#598`, 2026-07-06) added the handler to the macOS WKWebView host page:

- `Packages/CrowTerminal/Sources/CrowTerminal/Resources/xterm/terminal.html:158-175`

That page is loaded only by `XTermSurfaceView` → `TerminalSurfaceView` → `TmuxBackend.cockpitSurface()` — the AppKit/SwiftUI path deleted by [ADR 0010](docs/adr/0010-retire-the-macos-app.md) ("the web UI is the only client", 2026-07-07). Nothing in `crowd` loads it; `StaticAssets` only serves the xterm *library* files out of that bundle.

Both surviving clients have **no modified-Enter handling at all**:

| File | Custom key handler covers |
|---|---|
| `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js:3878` (`handleTerminalKey`, the real UI) | Cmd/Ctrl+C copy, Cmd+F find — returns `true` for Enter |
| `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/terminal.html:153` (debug page) | copy chord only |

So Shift+Enter falls through to xterm.js's default. From the bundled xterm.js 6.0.0 key table:

```js
case 13: o.key = e.altKey ? s.C0.ESC + s.C0.CR : s.C0.CR, o.cancel = !0; break;
```

There is **no `shiftKey` branch** — Enter and Shift+Enter are byte-identical `\r`. xterm.js 6.0.0 also has zero CSI-u / `modifyOtherKeys` / kitty-keyboard support (`rg -c 'modifyOtherKeys|kitty|;2u' xterm.js` → 0 matches), so it will never emit a distinguishable Shift+Enter on its own and it silently ignores tmux's `CSI > 4 ; 2 m` request. The host page has to synthesize the bytes — which is exactly what `#599` did, on the wrong surface.

### Why CI is green

`BundledResourcesTests.terminalHTMLHandlesModifiedEnter` pins the handler — against the dead macOS page. `WebTerminalAssetTests` parameterizes copy/paste/mouse-mode parity over `["app.js", "terminal.html"]` but has no modified-Enter case, so the surface that regressed is the one nothing asserts on.

## Transport verified end-to-end

Measured with a probe harness driving a real **tmux 3.6a** server started from the bundled `crow-tmux.conf`, injecting bytes at the client PTY exactly as the browser would, with the pane emitting Claude Code's real key negotiation. (Claude Code 2.1.220 emits both `CSI > 1 u` and `CSI > 4 ; 2 m` — present in the binary as `uv(">1u")` and `uv(">4;2m")`.)

| Injected at the client | App negotiates ext. keys (Claude Code) | App does not (vim, shell) |
|---|---|---|
| `\x1b[13;2u` (Shift+Enter, CSI-u) | `\x1b[27;2;13~` ✅ | `\r` ✅ |
| `\x1b\r` (ESC CR) | `\x1b[27;3;13~` | `\x1b\r` |
| `\r` | `\r` | `\r` |

Claude Code parses **both** encodings — its input tokenizer carries `/^\x1b\[(\d+)(?:…u` (CSI-u) and `/^\x1b\[27…~` (modifyOtherKeys), mapping codepoint 13 → `return` and modifier bit 1 → shift.

Two conclusions:

1. **`\x1b[13;2u` is the right sequence.** tmux rewrites it to a proper Shift+Enter for apps that negotiate extended keys, and **downgrades it to a plain `\r` for apps that don't** — so nothing leaks into vim or the shell. The existing `crow-tmux.conf` (`extended-keys on`, `terminal-features ,screen*:extkeys,xterm*:extkeys`) already carries it; no tmux config change is needed.
2. **Plain `ESC CR` would be worse.** It reaches non-negotiating apps as a literal `\x1b\r`, which in vim means "leave insert mode, then Enter".

## Fix

Port the `#599` handler into `handleTerminalKey` in `web/app.js`, ahead of the existing copy/find branches:

```js
// xterm.js emits the same bare \r for Enter and Shift+Enter (its key table has
// no shiftKey branch for keyCode 13), so Claude Code can't tell them apart and
// submits on both. Synthesize a distinct sequence — crow-tmux.conf's
// extended-keys setup carries CSI-u to apps that negotiate extended keys and
// downgrades it to plain \r for apps that don't, so nothing leaks into
// vim/shells. Ported from CrowTerminal's terminal.html (#599), which stopped
// being a live surface when the macOS app was retired (ADR 0010).
if (e.key === 'Enter' && !e.ctrlKey && !e.metaKey) {
  if (e.shiftKey && !e.altKey) {
    sendToPTY('\x1b[13;2u');
    e.preventDefault();
    e.stopPropagation();
    return false;
  }
  if (e.altKey && !e.shiftKey) {
    sendToPTY('\x1b\r');
    e.preventDefault();
    e.stopPropagation();
    return false;
  }
}
```

Notes for whoever picks this up:

- `handleTerminalKey` already early-returns on `e.type !== 'keydown'`, so unlike the macOS version this doesn't need to re-check the event phase — but it *does* need to keep suppressing the keypress/keyup phases, which the early return handles by returning `true`. Confirm no duplicate `\r` reaches the PTY.
- Per `#875`, a branch that shadows a browser default must `preventDefault()` — returning `false` only makes xterm skip its *own* handling. The macOS copy omitted this; include it here, since Enter has a real default on xterm's hidden helper textarea.
- The Option+Enter branch is **redundant** — xterm.js already maps `altKey` + Enter to `ESC CR` natively. Keep it anyway so the two host pages stay literally comparable, or drop it and note why.
- Apply the same change to `web/terminal.html` so the debug page doesn't diverge again.

## Acceptance criteria

- [ ] Shift+Enter inserts a newline in Claude Code in the browser terminal; Enter still submits.
- [ ] Shift+Enter in `vim` (insert mode) and in a bare shell behaves as plain Enter — no stray `^[` and no mode change.
- [ ] Option+Enter still inserts a newline.
- [ ] A regression test in `WebTerminalAssetTests` parameterized over `["app.js", "terminal.html"]` asserts the handler and the `\x1b[13;2u` literal — so the live surfaces are the ones pinned, not just the retired one.
- [ ] Either delete `CrowTerminal/Resources/xterm/terminal.html` + its `terminalHTMLHandlesModifiedEnter` test as dead post-ADR-0010 code, or add a comment marking it unused so the next fix doesn't land there again.

## Workaround until this ships

**Option+Enter** already works today with no Crow code involved (xterm.js maps Alt+Enter to `ESC CR` natively). Backslash + Enter works too.

## Out of scope — separate finding

The tmux client backing the browser attach reports **`TERM=xterm-ghostty`**, inherited from the shell that launched `crowd`:

```
$ tmux list-clients -F '#{client_tty} session=#{client_session} termname=#{client_termname}'
/dev/ttys005 session=crowd-web-2abca994 termname=xterm-ghostty
```

`PTYProcess.swift:111-113` only injects `TERM=xterm-256color` when the inherited environment has none, so whatever terminal started `crowd` leaks into every web attach. This does **not** cause this bug — `xterm*` matches the `extkeys` feature either way, and the probe measured identical results under both `xterm-256color` and `xterm-ghostty` — but it means tmux advertises Ghostty-only capabilities (sixel, RGB, bracketed-paste) to a client that is actually xterm.js 6.0. Worth its own issue.

dhilgaertner:
Closes #916

## What was wrong

xterm.js 6.0's key table has no `shiftKey` branch for keyCode 13 (`case 13: o.key = e.altKey ? ESC+CR : CR`), so Enter and Shift+Enter reach the PTY as the same bare `\r` — and xterm has no CSI-u / `modifyOtherKeys` support of its own to enable. The host page has to synthesize the bytes.

It already did, on the surface that no longer ships. #599 added the handler to CrowTerminal's WKWebView page the day before [ADR 0010](docs/adr/0010-retire-the-macos-app.md) retired the macOS app, and `terminalHTMLHandlesModifiedEnter` has pinned that dead page ever since. `WebTerminalAssetTests` parameterizes copy/paste/mouse-mode parity over the two *live* surfaces but had no modified-Enter case — so the pages that regressed were the ones nothing asserted on, and CI stayed green through it.

## What changed

The branch now lives in both live surfaces — `Resources/web/app.js`'s `handleTerminalKey` and the `terminal.html` debug page — and the guard moved onto them.

```js
if (e.key === 'Enter' && e.shiftKey && !e.altKey && !e.ctrlKey && !e.metaKey) {
  sendToPTY('\x1b[13;2u');
  e.preventDefault();
  e.stopPropagation();
  return false;
}
```

`preventDefault()` is load-bearing, not cosmetic: xterm cancels Enter itself, but returning `false` makes its `_keyDown` return *before* that cancel (#875), and this handler waves the keypress phase straight through — so an uncancelled keydown would write a second `\r`.

No tmux config change: `crow-tmux.conf` already has `extended-keys on` and `xterm*:extkeys`.

## The two calls the ticket left open

**Dropped the Option+Enter branch.** xterm already maps `altKey` + Enter to ESC CR natively, so it would only restate the library; `!e.altKey` keeps Shift+Option+Enter on that native path. A jsdom case pins that the handler leaves Alt+Enter alone, so a future "add it back for symmetry" has to confront the duplication.

**Marked the retired CrowTerminal page unused rather than deleting it.** Deleting means also deleting `BundledResources.terminalHTMLURL`, `XTermSurfaceView`, `TerminalSurfaceView` and `TmuxBackend.cockpitSurface()` — a dead-code sweep that deserves its own change rather than riding along on a key-handling fix. The page and the accessor now carry warnings pointing at `Resources/web/`, and the old test's comment names the live guard that supersedes it. Happy to do the sweep separately if preferred.

## Verification

The daemon isn't buildable to a browser in this environment, so this was verified in two layers that meet in the middle:

**Browser side** — `web-tests/key-handler.test.js` (extended to expose `termWs`) drives the *real* `handleTerminalKey` through the *real* `sendToPTY`: Shift+Enter emits exactly `\x1b[13;2u`, once, and cancels; plain / Option / Ctrl / Cmd + Enter pass through untouched; non-keydown phases emit nothing. 44 assertions pass.

**Transport side** — a probe against real tmux 3.6a started from the bundled `crow-tmux.conf`, injecting bytes at an attached client PTY exactly as the browser would, reading back what the pane process received via `cat -v`:

| Injected | App negotiates extended keys | App does not |
|---|---|---|
| `\x1b[13;2u` | `^[[27;2;13~` — Shift+Enter | bare `\r`, **no `^[`** |
| `\r` | bare `\r` | bare `\r` |

Rows 1-right and 2-right are byte-identical, which is the acceptance criterion for vim and a bare shell. Results were identical under the leaked `TERM=xterm-ghostty` noted in the ticket's out-of-scope section.

**The new guard bites.** Confirmed by stashing only the two source files and re-running: `modifiedEnterIsDistinguishableFromPlainEnter` fails on both assets. It matches `e.key === 'Enter' && e.shiftKey` as one contiguous condition, because the loose form passes on the unfixed `app.js` — its modal dialogs already contain `e.key === 'Enter'`.

`make test` is green (CrowDaemon 325 tests / 54 suites, CrowTerminal all suites, plus the three parity gates). The 10 pre-existing `row.test.js` failures are unrelated and reproduce identically on an unmodified tree.

**Not covered:** the literal browser keydown dispatch into xterm's custom handler — its documented API, already exercised in production by the Cmd+C and Cmd+F branches of this same function.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -4345,13 +4345,41 @@ async function uploadDroppedFiles(files) {
 //
 // Cmd/Ctrl+C copies the selection (falling through to SIGINT when nothing is
 // selected so Ctrl+C still interrupts), Cmd+F opens our find prompt. Lets the
-// browser own copy instead of tmux's copy-mode.
+// browser own copy instead of tmux's copy-mode. Shift+Enter is rewritten to a
+// distinguishable CSI-u sequence (see the branch).
 //
 // A branch that shadows a browser default must preventDefault(): returning
 // false only makes xterm skip its OWN key handling (its _keyDown returns early,
 // before any cancel()), so the browser's default gesture still runs (#875).
 function handleTerminalKey(e) {
   if (e.type !== 'keydown') return true;
+  // Shift+Enter → CSI-u, so Claude Code can tell it from a plain Enter and
+  // inserts a newline instead of submitting (#598/CROW-916). xterm.js's key
+  // table has no shiftKey branch for keyCode 13 — `e.altKey ? ESC+CR : CR` —
+  // so both chords otherwise arrive at the PTY as the same bare \r, and
+  // xterm.js 6.0 has no CSI-u/modifyOtherKeys support of its own to turn on.
+  // crow-tmux.conf's `extended-keys on` + `xterm*:extkeys` carries this to
+  // apps that negotiate extended keys and downgrades it to a plain \r for apps
+  // that don't, so nothing leaks into vim or a shell.
+  //
+  // Option+Enter is deliberately NOT handled: xterm already maps altKey + Enter
+  // to ESC CR natively, so a branch here would only restate the library.
+  // `!e.altKey` keeps Shift+Option+Enter on that native path.
+  //
+  // preventDefault() is load-bearing, not cosmetic. xterm cancels Enter itself
+  // (`o.cancel = true`), but returning false makes its _keyDown return BEFORE
+  // that cancel (#875) — and this handler waves the keypress phase straight
+  // through, so an uncancelled keydown would let keypress fire and write a
+  // second \r.
+  //
+  // Ported from CrowTerminal's terminal.html (#599), which stopped being a live
+  // surface when the macOS app was retired (ADR 0010).
+  if (e.key === 'Enter' && e.shiftKey && !e.altKey && !e.ctrlKey && !e.metaKey) {
+    sendToPTY('\x1b[13;2u');
+    e.preventDefault();
+    e.stopPropagation();
+    return false;
+  }
   const mod = e.metaKey || e.ctrlKey;
   if (mod && (e.key === 'c' || e.key === 'C') && term.hasSelection()) {
     copyToClipboard(term.getSelection());
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/terminal.html
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/terminal.html
@@ -152,6 +152,28 @@
       }
       term.attachCustomKeyEventHandler(function (e) {
         if (e.type !== 'keydown') return true;
+        // Shift+Enter → CSI-u, so Claude Code can tell it from a plain Enter
+        // and inserts a newline instead of submitting (#598/CROW-916). xterm's
+        // key table has no shiftKey branch for keyCode 13, so both chords
+        // otherwise reach the PTY as the same bare \r. crow-tmux.conf's
+        // `extended-keys on` + `xterm*:extkeys` carries this to apps that
+        // negotiate extended keys and downgrades it to a plain \r for apps
+        // that don't, so nothing leaks into vim or a shell.
+        //
+        // Option+Enter is deliberately NOT handled: xterm maps altKey + Enter
+        // to ESC CR natively, so a branch here would only restate the library.
+        // `!e.altKey` keeps Shift+Option+Enter on that native path.
+        //
+        // preventDefault() is required: returning false makes xterm's _keyDown
+        // return before its own cancel (#875), and this handler waves the
+        // keypress phase through — an uncancelled keydown would write a second
+        // \r there.
+        if (e.key === 'Enter' && e.shiftKey && !e.altKey && !e.ctrlKey && !e.metaKey) {
+          sendToPTY('\x1b[13;2u');
+          e.preventDefault();
+          e.stopPropagation();
+          return false;
+        }
         const copyCombo = (e.metaKey && !e.ctrlKey) || (e.ctrlKey && e.shiftKey);
         if (copyCombo && (e.key === 'c' || e.key === 'C')) {
           if (term.hasSelection()) {
--- a/Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift
+++ b/Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift
@@ -249,6 +249,43 @@ import Testing
             "terminal.html's copy branch must cancel the browser default")
     }
 
+    /// CROW-916: both live surfaces must rewrite Shift+Enter to a sequence the
+    /// agent can tell apart from a plain Enter.
+    ///
+    /// xterm.js's key table has no `shiftKey` branch for keyCode 13
+    /// (`e.altKey ? ESC+CR : CR`), so without this the two chords reach the PTY
+    /// as the same bare `\r` and Claude Code submits on both. CSI-u is the right
+    /// sequence because `crow-tmux.conf`'s extended-keys setup carries it to
+    /// apps that negotiate extended keys and downgrades it to a plain `\r` for
+    /// apps that don't — a literal `ESC CR` would reach vim as "leave insert
+    /// mode, then Enter".
+    ///
+    /// This is the guard that was missing. #599 added the handler to
+    /// CrowTerminal's `terminal.html` and pinned it there
+    /// (`BundledResourcesTests.terminalHTMLHandlesModifiedEnter`) — but ADR 0010
+    /// retired that surface, so the passing test covered a page nothing loads
+    /// while both shipping surfaces regressed unasserted. Parameterized over the
+    /// live assets so it can't happen the same way twice.
+    ///
+    /// Asserted on the comment-stripped source: the rationale prose names the
+    /// sequence too, and a guard that a comment can satisfy guards nothing. The
+    /// branch is matched as one contiguous condition rather than as separate
+    /// `'Enter'` / `shiftKey` mentions — `app.js`'s modal dialogs already test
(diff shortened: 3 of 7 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- The handler emits a **constant** control sequence (`\x1b[13;2u`); nothing is user-derived, so there's no injection surface, no auth/authorization impact, and no secrets involved.
- `crow-tmux.conf`'s `extended-keys on` + `xterm*:extkeys` (unchanged) downgrades the sequence to a plain `\r` for apps that don't negotiate extended keys, so nothing leaks into vim or a bare shell. The PR's transport-layer probe (byte-identical to a native terminal) confirms this.

**Concerns:** none.

### Code Quality
- Correct CSI-u encoding: codepoint 13 (CR) + modifier `2` (= 1 + Shift) → `\x1b[13;2u` (`app.js:4377`, `terminal.html:171`).
- Branch is disjoint from the copy/paste/find branches and the modifier exclusions (`!e.altKey && !e.ctrlKey && !e.metaKey`) keep every other Enter chord on its native path — verified by test.
- `preventDefault()` is correctly identified as load-bearing (#875): returning `false` alone leaves the keypress phase to write a second `\r`. The test pins exactly one emitted sequence.
- The two judgment calls (dropping the Option+Enter branch, marking the retired CrowTerminal page unused rather than deleting) are both reasonable and, crucially, **pinned by tests** so a future regression has to confront them.
- The new drift guard is parameterized over both live surfaces (`Resources/web/{app.js,terminal.html}`), directly fixing the class of bug that let #599's fix rot on a dead page — and it asserts on comment-stripped source so prose can't satisfy the guard.

**Verification performed in this review:** installed jsdom and ran `key-handler.test.js` — 44/44 pass, exercising the real `handleTerminalKey` through the real `sendToPTY`. Confirmed `/terminal.html` is a served surface (`StaticAssets.swift:36`) so the app.js↔terminal.html parity is warranted, and `app.js:4114` is the sole `attachCustomKeyEventHandler` attach point. Swift asset assertions confirmed against the actual sources.

- The Swift guard `code.contains("e.key === 'Enter' && e.shiftKey")` is sensitive to conjunct ordering; a purely cosmetic reorder would fail it despite correct behavior. This is an intentional, documented trade-off (fails loud, never passes silently) — flagging only for awareness, not a change request.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 1 Green] findings. Minimal, well-scoped, thoroughly tested, and it repairs the coverage gap that hid the original regression.

---

PR #920, opened 2026-07-29
Hooks: stop a main clone's block from firing in every worktree session (CROW-915)

### LINKED ISSUE #897: Stale hook commands point at a deleted worktree's dev binary — every hook event fails, telemetry silently lost

### LINKED ISSUE #915: Worktree sessions inherit the main clone's .claude/settings.local.json — one stale hook file breaks hooks + telemetry for every session on that repo
## Summary

A stale `.claude/settings.local.json` in a repo's **main clone** is executed by **every git-worktree session of that repo**, not just by sessions whose cwd is the main clone. Because Crow writes its per-worktree hook config assuming that file is the effective config for the session, a single stale file in one main clone breaks hooks for every Crow session on that repo, indefinitely.

This is the blast-radius correction to #897. That issue reported the stale file and assumed it only affected sessions run from the affected clone. It affects all of them.

## Observed

Live `PreToolUse:Bash` failures, repeating on essentially every tool call, in four concurrently running Crow sessions:

```
Ran 2 shell commands
PreToolUse:Bash hook error
Failed with non-blocking status code: /bin/sh: /Users/<user>/Dev2/RadiusMethod/crow-136-session-analytics-otel/.build/arm64-apple-macosx/debug/crow: No such file or directory
```

`crow-136-session-analytics-otel` is a worktree deleted months ago. The path lives in exactly one file on disk:

```
Dev2/RadiusMethod/corveil/.claude/settings.local.json   (mtime Apr 13, 17 hook entries)
```

None of the affected sessions run from that directory. Every one runs from its own worktree, each of which has its own correct, Crow-written `settings.local.json` (17 `hook-event` lines pointing at a valid binary).

## The mechanism

A git worktree's `.git` is a *file*, not a directory:

```
$ cat Dev2/RadiusMethod/corveil-1893-cliclient-decode-diagnostics/.git
gitdir: /Users/<user>/Dev2/RadiusMethod/corveil/.git/worktrees/corveil-1893-cliclient-decode-diagnostics
```

Project-root resolution follows that pointer to the **main clone**, so the main clone's `.claude/settings.local.json` is loaded *in addition to* the worktree's own. Both hook sets run. Crow's is fine; the Apr 13 one fails on every event.

(The inheritance is directly observed — see the evidence below. The gitdir-following explanation is the most plausible mechanism, but the fix does not depend on which resolution rule produces it.)

## Evidence — natural experiment

The `corveil` main clone has the stale file; the `crow` main clone does not. Counting occurrences of the dead path in each session's transcript:

| Session (worktree) | Repo | Hits | What they are |
|---|---|---|---|
| `corveil-1875-nested-extraction-remarshal` | corveil | 373 | runtime hook errors |
| `corveil-1874-fail-on-no-land` | corveil | 279 | runtime hook errors |
| `corveil-1894-worker-run-retry-revive` | corveil | 174 | runtime hook errors |
| `corveil-1895-worker-runs-ops-reporting` | corveil | 171 | runtime hook errors |
| `corveil-docs-marketing-positioning` | corveil | 187 | runtime hook errors |
| `corveil-1815-embeddings-cost-analysis` | corveil | 355 | runtime hook errors |
| `corveil-1893-cliclient-decode-diagnostics` | corveil | 53 | runtime hook errors |
| `crow-806-adr-control-plane-parity` | crow | **0** | — |
| `crow-809-workspaces-cli` | crow | **0** | — |
| `crow-897-stale-hook-binary-path` | crow | 12 | prompt text only (the ticket quotes the path) |

The corveil sessions' prompts never mention `crow-136`, so their hits are entirely runtime. The split is clean along repo lines, which is what makes the main-clone inheritance conclusive rather than circumstantial.

## Why it surfaced now

Nothing changed. The file has been there since April. What changed is *which sessions are running*: all four currently active sessions are `corveil` worktrees, so 100% of live sessions are affected. Earlier in the same evening several `crow` worktree sessions were running alongside them and were completely silent — which made the problem look intermittent or session-specific when it is neither.

## Impact

- **Every hook event fails in every worktree session of an affected repo** — `PreToolUse`, `PostToolUse`, `Stop`, and the rest, on essentially every tool call.
- **Telemetry is silently lost** for those sessions: `hook-event` never reaches `crowd`, so no analytics/scorecard data despite `telemetry.enabled: true`. This is the part that matters more than the visible noise.
- **Crow's per-worktree hook config is not authoritative**, contrary to what the writer assumes. Any stale/hostile/simply-different hook block in a main clone silently composes with it.
- Self-perpetuating: the affected repo is the one people work in most, so the noise scales with usage.

## Root cause, in Crow's terms

1. **The writer assumes isolation it does not have.** Crow writes `<worktree>/.claude/settings.local.json` and treats it as the session's hook configuration. The main clone's file is merged in as well, and Crow neither reads nor reconciles it.
2. **Nothing ever validates or prunes a main clone's file.** Crow writes one into a main clone only if a session was once created there, then never revisits it. The reaper that deleted `crow-136-session-analytics-otel` had no reason to touch a settings file in a different directory.
3. **An old Crow build wrote a `.build/…` path** rather than a stable one. Current Crow writes `{devRoot}/.claude/bin/crow`, which is correct — but see the note below, that symlink is itself fragile.

## Suggested directions

- **Validate the whole effective hook set at launch, not just the file Crow writes.** Before launching an agent in a worktree, also inspect the main clone's `.claude/settings.local.json` (resolve via `git rev-parse --git-common-dir`); if it contains Crow-written hooks whose binary is missing or whose `--session` uuid is unknown, repair or strip them.
- **Prune on reap.** When the retention reaper deletes a worktree, scrub hook blocks referencing that worktree's binary or session uuid from any `settings.local.json` still on disk — including main clones.
- **Never emit a build-product path.** Assert that no Crow-written hook command contains `.build/`; always emit the stable launcher path.
- **Consider making `--session` resolution lazy** — have hooks pass cwd and let `crowd` map it to a live session — so a settings file cannot outlive its session id.
- **One-shot repair on startup** for existing installs: detect dangling Crow hook commands anywhere under devRoot and offer to fix.

## Related fragility worth folding in

`{devRoot}/.claude/bin/` was observed **empty** during this investigation — the `crow` symlink that current Crow writes into every hook command had been removed, and `crow` was off PATH entirely. A rebuild had moved the binary from `.build/arm64-apple-macosx/debug/crow` to `.build/debug/crow`. In the same window, a newly created worktree received hooks pointing at `…/crow/.build/arm64-apple-macosx/debug/crow` — a `.build/` path, already stale on arrival, while `setup.sh` itself resolved a different working path for its own use.

So the "stable symlink" is not reliably stable, and the writer can still emit a build path today. That makes this a live regression, not only an Apr 13 artifact.

## Immediate remediation (for this install)

`Dev2/RadiusMethod/corveil/.claude/settings.local.json` contains **only** the stale hook block — no `env`, no `attribution`, nothing else — so deleting it is lossless and fixes all four running sessions.

## Acceptance ideas

- [ ] A stale hook block in a main clone does not execute in that repo's worktree sessions (repaired, stripped, or refused at launch).
- [ ] No Crow-written hook command contains `.build/` — enforced by a test.
- [ ] Deleting a worktree removes or repairs hook entries elsewhere that reference its binary/session.
- [ ] A session whose hooks were repaired records telemetry normally.
- [ ] Hook-binary resolution survives a rebuild that changes the Swift build output layout.

## Related

- #897 — original report of the stale hook file. This issue supersedes its impact analysis: not "sessions run from an affected clone" but every worktree session of that repo.

dhilgaertner:
Closes #915. Also closes #897 — see "Relationship to #897" below.

## The defect

A linked worktree's `.git` is a *file* pointing at the main clone, and
project-root resolution follows it. So a worktree session loads the **main
clone's** `.claude/settings.local.json` *in addition to* its own, and both hook
blocks fire. Crow writes the per-worktree file and treats it as the session's
hook configuration; it never was. One block in one main clone reaches every
session on that repo.

**Verified on this machine rather than assumed.** Exactly one file under the dev
root contains the dead `crow-136-session-analytics-otel` path — the `corveil`
main clone's settings file — yet the sessions logging those errors all run from
worktrees whose own settings files are clean and name a live binary. The only
possible source is the main clone. `git rev-parse --git-common-dir` from a
worktree lands on the main clone's `.git`, and the `claude` binary's
settings-resolution region references `commondir` / `worktrees` /
`^gitdir:\s*(.+)$` adjacent to `settings.local.json`, which is consistent with
that being the mechanism.

Two failures follow, and #900 addresses neither:

| | Symptom | Why the existing sweep misses it |
|---|---|---|
| **Dangling** block | `/bin/sh` fails before `crow` runs → every hook event errors, no telemetry, for every worktree session of the repo | `ClaudeHookRepair.sweep` is boot-only and walks devRoot at a fixed depth of 2 |
| **Live** block | Events attributed to the main clone's session and fired twice | Its health test is per-directory — good binary + live session reads as *healthy*, so it is deliberately left in place |

The second row is what #915 adds over #897, and nothing addressed it.

## Changes

**1. `cwd` outranks the baked UUID in `hook-event`** (`EngineRouter`).

A `--session` id says which *file* the command was written into; `cwd` says which
*session is running*. A live id that doesn't own the cwd is an inherited copy and
is dropped — the session that owns the cwd reports the same event from its own
block, so nothing is lost. This fixes attribution and the duplicate for every
file already on disk, with no change to what Crow writes.

Carve-outs, each a case where dropping would lose a real event:

- an id that **owns** the directory (two rows on one path is reachable through
  orphan recovery and a `setup.sh` retry, and the pick between them is a
  dictionary coin flip — so an owning id always wins over it);
- the **Manager**, matched by constant and never by path everywhere else;
- a **non-live** id — that is #897's stale uuid, still re-routed by `cwd` rather
  than discarded.

`AppState.sessionIDs(forWorktreePath:)` is new: every owner, stable order, both
sides `standardizingPath`-normalized. Without the normalization the fix goes
silently inert on a spelling mismatch, and routing could disagree with
`repairStaleHooks` (which already normalizes) about who owns a directory.

**2. `ClaudeHookRepair.reconcileMainClone`** — resolves the main clone through the
gitdir chain (`.git` file → `commondir`; no subprocess, no `git` on PATH, no
async hop) and reconciles its block:

| Main clone is… | Action |
|---|---|
| this session's own worktree | leave |
| another live session's worktree | repair in place for that session |
| no live session's worktree | strip Crow entries |

This is the half the daemon cannot do — a dangling binary fails before `crow` is
ever reached. Wired into `prepareWorktreeForAgentLaunch`, the one shared gate
whose call sites are already the canonical enumeration of launch paths, and into
`performDiskCleanup` so a reaped worktree's binary and session id are scrubbed
from the clone immediately rather than at next boot. The reaper's snapshot
excludes the session being deleted — it is still in `AppState` at that point, and
counting it as live is how a block outlives its session in the first place.

Strip removes only entries matching the strict `parseCrowHookCommand` shape, so a
user's own hook under a managed event name survives (covered by a test).

**3. `resolveCrowBinary` never returns a `.build/` path.**

It previously fell back to one with only a log warning. That is not hypothetical:
`{devRoot}/.claude/bin` was empty on this machine while still first on `PATH`, and
the session this PR was written in has `…/crow/.build/arm64-apple-macosx/debug/crow`
baked into all 17 of its hook commands. New order — devRoot link, else a stable
link under `~/.local/share/crow/bin`, else a non-build binary, else `nil`. Writing
no hook block costs that session telemetry; writing a broken one costs every
session on the repo, forever.

## Relationship to #897

#900 fixed #897's dangling-binary case for directories the boot sweep reaches.
This closes the two directions it left open — *validate the effective set at
launch* (2) and *never emit a build-product path* (3) — and extends both to the
main clone that #897's impact analysis did not know was in play. Closing both.

## Test note

`liveSessionIDWins` asserted that a live `session_id` beats a `cwd` resolving
elsewhere ("the fallback must not change routing for anything that works
today"). **That assertion was the bug** — it is the exact misattribution #915
reports, written down as intended behavior. It is inverted here deliberately,
not bent to fit the code.

## Verification

`make build && make test` — full suite green, including CLI/RPC parity.

New coverage: the `reconcileMainClone` suite builds **real** `git init` +
`git worktree add` fixtures (a fabricated `.git` could agree with the code while
both disagree with git) and cross-checks resolution against
`git rev-parse --git-common-dir`. Resolution declines rather than guesses on a
gitdir shape it doesn't recognize — this code deletes hook entries, so a wrong
answer would aim the strip at an unrelated directory.

Not yet run against the live repro on the reporter's machine: that needs a
`crowd` restart, which would interrupt running sessions. Expected there —
`RadiusMethod/corveil/.claude/settings.local.json` is deleted outright (it holds
only the stale block), `Dev2/.claude/bin/crow` is repopulated, and the corveil
worktree sessions stop emitting `PreToolUse:Bash hook error`.

### DIFF
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookConfigWriter.swift
@@ -310,10 +310,40 @@ public struct ClaudeHookConfigWriter: HookConfigWriter {
     /// swallowed — this is best-effort and must never fail a launch.
     @discardableResult
     public static func ensureCrowCLISymlink(devRoot: String, appCrowPath: String? = nil) -> String? {
-        let fm = FileManager.default
         let binDir = (devRoot as NSString).appendingPathComponent(".claude/bin")
         let link = (binDir as NSString).appendingPathComponent("crow")
-        let resolved = appCrowPath ?? appCrowBinary()
+        return materializeCrowSymlink(at: link, target: appCrowPath ?? appCrowBinary())
+    }
+
+    /// `{HOME}/.local/share/crow/bin/crow` — the stable link used when the dev
+    /// root's own cannot be made (unwritable, or no dev root at all).
+    ///
+    /// `~/.local/share/crow/` is already Crow's state directory (the daemon
+    /// socket lives there), so this introduces no new convention. It exists so
+    /// `resolveCrowBinary` always has *some* stable path to emit and never has
+    /// to fall back to a `.build/` product (#915) — the failure that put a dead
+    /// worktree's binary into every hook command in #897.
+    static func stableFallbackLinkPath() -> String {
+        FileManager.default.homeDirectoryForCurrentUser
+            .appendingPathComponent(".local/share/crow/bin/crow").path
+    }
+
+    /// Point `link` at `target`, creating the parent directory as needed.
+    /// Returns the **link** path once it resolves to an executable, `nil`
+    /// otherwise. Idempotent; safe to call on every launch.
+    ///
+    /// Two deliberate behaviors, both load-bearing:
+    /// - A link already pointing at `target` is left alone rather than
+    ///   unlink+relink'd — that window is one where a concurrently firing hook
+    ///   sees ENOENT.
+    /// - A **non-symlink** `crow` at that path is never replaced. We only ever
+    ///   own symlinks here; a real file is the user's.
+    ///
+    /// All errors are logged and swallowed — this is best-effort and must never
+    /// fail a launch.
+    private static func materializeCrowSymlink(at link: String, target resolved: String?) -> String? {
+        let fm = FileManager.default
+        let binDir = (link as NSString).deletingLastPathComponent
 
         guard let target = resolved, fm.isExecutableFile(atPath: target) else {
             // Drop a link we can no longer back, so a broken pointer doesn't
@@ -363,26 +393,53 @@ public struct ClaudeHookConfigWriter: HookConfigWriter {
 
     /// The `crow` path to bake into a hook command (or any other file on disk).
     ///
-    /// Ensures the stable `{devRoot}/.claude/bin/crow` symlink first and returns
-    /// *that*, so an emitted command never contains a `.build/…` product path
-    /// belonging to whichever worktree the running daemon happened to be built
-    /// in — the root cause of #897.
+    /// **Never returns a `.build/…` path.** A build product belongs to whichever
+    /// worktree the running daemon was built in; when that worktree is reaped the
+    /// command dies with it, every hook event fails in `/bin/sh` before `crow`
+    /// runs, and the session silently records no telemetry. That is the root
+    /// cause of #897, and — because a worktree session also loads its *main
+    /// clone's* settings — one such command breaks every session on the repo
+    /// (#915).
     ///
-    /// Falls back to `appCrowBinary()` only when there is no dev root or the
-    /// link could not be created, warning loudly if that fallback is itself a
-    /// build product. Deliberately not a `precondition`: crashing the daemon
-    /// over an unwritable dev root would be worse than the noise it prevents.
-    public static func resolveCrowBinary(devRoot: String?, appCrowPath: String? = nil) -> String? {
+    /// Resolution order, each step a stable path that survives a rebuild:
+    ///   1. `{devRoot}/.claude/bin/crow` (also on the agent's PATH)
+    ///   2. `~/.local/share/crow/bin/crow`, for an unwritable or absent dev root
+    ///   3. `appCrowBinary()`, but only when it is not itself a build product —
+    ///      i.e. a real install on PATH or in a release `.app`
+    ///   4. `nil`
+    ///
+    /// Returning `nil` means the caller writes no hook block at all. That costs
+    /// telemetry for the session, but it is strictly better than writing a
+    /// command that is already broken: a dangling block outlives the session,
+    /// and every worktree of the repo inherits its errors. Deliberately not a
+    /// `precondition` — crashing the daemon would be worse than either.
+    ///
+    /// `appCrowPath` and `stableFallbackLink` are injectable for tests, which
+    /// must not write into the real `~/.local/share` (ADR 0012).
+    public static func resolveCrowBinary(
+        devRoot: String?, appCrowPath: String? = nil, stableFallbackLink: String? = nil
+    ) -> String? {
         if let devRoot, let link = ensureCrowCLISymlink(devRoot: devRoot, appCrowPath: appCrowPath) {
             return link
         }
-        let fallback = appCrowPath ?? appCrowBinary()
-        if let fallback, fallback.contains("/.build/") {
+        let target = appCrowPath ?? appCrowBinary()
+        let fallbackLink = stableFallbackLink ?? stableFallbackLinkPath()
+        if let link = materializeCrowSymlink(at: fallbackLink, target: target) {
             CrowLog.info(
-                "[ClaudeHookConfigWriter] WARNING: falling back to build-product crow path \(fallback)"
-                + " — hook commands written now will break when that build directory is removed."
-                + " Check that \(devRoot ?? "<no dev root>")/.claude/bin is writable.")
+                "[ClaudeHookConfigWriter] \(devRoot ?? "<no dev root>")/.claude/bin is not usable"
+                + " — writing hook commands against \(link) instead.")
+            return link
         }
-        return fallback
+        // Both stable links failed. Only hand back the raw binary if it is one
+        // that will still exist after the next rebuild.
+        if let target, !target.contains("/.build/") {
+            return target
+        }
+        CrowLog.error(
+            "[ClaudeHookConfigWriter] no stable crow path available"
+            + " (candidate \(target ?? "<none>") is a build product); writing no hook config."
+            + " Check that \(devRoot ?? "<no dev root>")/.claude/bin or"
+            + " \(fallbackLink) is writable.")
+        return nil
     }
 }
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookRepair.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookRepair.swift
@@ -265,7 +265,30 @@ public enum ClaudeHookRepair {
     /// cwd-authoritative `hook-event` resolution drops it when it fires inside a
     /// *worktree* session, and accepts it when it fires in the main clone itself.
     ///
-    /// Performs blocking file I/O; call off the MainActor (#892).
+    /// ## Actor context
+    ///
+    /// Performs blocking file I/O, and **the launch path deliberately runs it on
+    /// the MainActor** (`SessionService.prepareWorktreeForAgentLaunch`, which is
+    /// synchronous and MainActor-bound). That is not an oversight and must not be
+    /// "fixed" by detaching: the block has to be gone from disk *before* the agent
+    /// process opens the directory, and the launch paths type the agent's command
+    /// into its pane in the same synchronous block. A detached reconcile would
+    /// race the very launch it exists to protect — the agent would start with the
+    /// inherited block still live, which is the bug.
+    ///
+    /// The cost is bounded so that is affordable: one `stat` for anything that is
+    /// not a linked worktree (main clones, review clones, the Manager's dev root —
+    /// the early-out in `resolveMainClone`), and for a real worktree three more
+    /// small reads (`.git`, `commondir`, the settings file). A *write* happens only
+    /// when a block is actually stale or foreign, which is a once-per-defect event,
+    /// not once per launch. `prepareWorktreeForAgentLaunch` already does MainActor
+    /// file I/O for the trust seed and the Grok/Antigravity strip, so this adds no
+    /// new category of work to that gate (#892).
+    ///
+    /// The reaper reaches it through `Task.detached` instead — there is no launch
+    /// to order against there, and it is already off the MainActor. Taking the
+    /// AppState facts as a value snapshot (`HookOwnership`) is what makes that
+    /// possible: this function needs no actor of its own.
     public static func reconcileMainClone(
         worktreePath: String,
         crowPath: String?,
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookRepair.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeHookRepair.swift
@@ -216,6 +216,259 @@ public enum ClaudeHookRepair {
         return candidates
     }
 
+    // MARK: - Main-clone reconciliation (#915)
+
+    /// What `reconcileMainClone` did, for the launch log.
+    public enum MainCloneOutcome: Equatable, Sendable {
+        /// `worktreePath` is not a linked worktree (its `.git` is a directory),
+        /// so there is no separate main clone to reconcile.
+        case notAWorktree
+        /// The main clone holds no Crow-managed hook block.
+        case noBlock
+        /// The block belongs to the session that owns that directory and is
+        /// healthy — left untouched.
+        case healthy
+        /// Rewritten to the owning session and a good binary.
+        case repaired(mainClone: String)
+        /// Crow entries removed: nothing legitimately runs from that directory.
+        case stripped(mainClone: String)
+        /// Present but not safely actionable (unparseable or unwritable).
+        case skipped(mainClone: String)
+    }
+
+    /// Reconcile the **main clone's** hook block before launching an agent in
+    /// one of its worktrees (#915).
+    ///
+    /// A linked worktree's `.git` is a file pointing at the main clone, and
+    /// project-root resolution follows it — so the main clone's
+    /// `.claude/settings.local.json` is loaded *in addition to* the worktree's
+    /// own, and both hook sets run. Crow writes the per-worktree file and treats
+    /// it as the session's configuration; it is not. One stale block in a main
+    /// clone breaks hooks and telemetry for **every** session on that repo, and
+    /// one merely *foreign* block (naming a different live session) misattributes
+    /// their events and fires each twice.
+    ///
+    /// `sweep` cannot cover this. It is boot-time only, walks devRoot at a fixed
+    /// depth of 2 (so a clone living anywhere else is invisible), and — decisively
+    /// — its health test is per-directory: a block naming a live session *is*
+    /// healthy by that test, which is exactly the case that must still go.
+    ///
+    /// The decision, given the directory the block sits in:
+    ///
+    /// | Main clone is… | Action |
+    /// |---|---|
+    /// | this session's own worktree | leave (it is this session's own block) |
+    /// | another live session's worktree | repair in place for that session |
+    /// | no live session's worktree | strip Crow entries |
+    ///
+    /// Note "repair" keeps a legitimately-hosted block working: the daemon's
+    /// cwd-authoritative `hook-event` resolution drops it when it fires inside a
+    /// *worktree* session, and accepts it when it fires in the main clone itself.
+    ///
+    /// Performs blocking file I/O; call off the MainActor (#892).
+    public static func reconcileMainClone(
+        worktreePath: String,
+        crowPath: String?,
+        liveSessionIDs: Set<UUID>,
+        sessionByWorktreePath: [String: UUID]
+    ) -> MainCloneOutcome {
+        guard let mainClone = resolveMainClone(worktreePath: worktreePath) else {
+            return .notAWorktree
+        }
+        // A worktree whose main clone is itself (shouldn't happen, but a
+        // hand-made `.git` file could say so) needs no reconciliation.
+        let worktree = (worktreePath as NSString).standardizingPath
+        guard mainClone != worktree else { return .notAWorktree }
+
+        return reconcileHookBlock(
+            inDirectory: mainClone, crowPath: crowPath,
+            liveSessionIDs: liveSessionIDs, sessionByWorktreePath: sessionByWorktreePath)
+    }
+
+    /// `reconcileMainClone` for a directory already known to be the main clone.
+    ///
+    /// Split out for the retention reaper, which reconciles *after*
+    /// `git worktree remove` has deleted the worktree — so `resolveMainClone`
+    /// has no `.git` file left to read, but `WorktreeCleanupItem.repoPath`
+    /// already names the clone directly.
+    public static func reconcileHookBlock(
+        inDirectory dir: String,
+        crowPath: String?,
+        liveSessionIDs: Set<UUID>,
+        sessionByWorktreePath: [String: UUID]
+    ) -> MainCloneOutcome {
+        let dir = (dir as NSString).standardizingPath
+        // The session that legitimately owns that directory, if any. Only a
+        // *live* one counts — a row for a deleted session must not keep a block
+        // alive.
+        let owner = sessionByWorktreePath[dir].flatMap {
+            liveSessionIDs.contains($0) ? $0 : nil
+        }
+        return reconcileDirectory(dir, owner: owner, crowPath: crowPath)
+    }
+
+    /// The main clone backing a linked worktree, or `nil` when `worktreePath`
+    /// is not a linked worktree.
+    ///
+    /// Reads the gitdir chain directly rather than shelling out to
+    /// `git rev-parse --git-common-dir`: it is the same resolution, but with no
+    /// subprocess on the launch path, no dependency on `git` being found, and no
+    /// async hop in a `nonisolated static` caller. It also mirrors what the agent
+    /// harness itself does, which is the behavior we are compensating for.
+    ///
+    /// - `.git` a directory → this is a main clone (or a bare repo); no separate
+    ///   main clone exists.
+    /// - `.git` a file → `gitdir: <path>` names the worktree's private gitdir,
+    ///   e.g. `<main>/.git/worktrees/<name>`. The shared dir is that gitdir's
+    ///   `commondir` (git writes it, normally `../..`), and the main clone is its
+    ///   parent.
+    static func resolveMainClone(worktreePath: String) -> String? {
+        let fm = FileManager.default
+        let dotGit = (worktreePath as NSString).appendingPathComponent(".git")
+
+        var isDirectory: ObjCBool = false
+        guard fm.fileExists(atPath: dotGit, isDirectory: &isDirectory) else { return nil }
+        if isDirectory.boolValue { return nil }
+
+        guard let raw = try? String(contentsOfFile: dotGit, encoding: .utf8) else { return nil }
+        // A `.git` file is a single `gitdir: <path>` line. Anything else is not
+        // a shape we understand, and guessing would be worse than doing nothing.
+        let trimmed = raw.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard trimmed.hasPrefix("gitdir:") else { return nil }
+        let pointer = String(trimmed.dropFirst("gitdir:".count))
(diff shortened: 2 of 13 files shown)


dgershman (changes requested):
## Code & Security Review

Reviewed the full diff, built the Swift daemon (`swift build` — clean; the one warning is in untouched `RPCWebSocketHandler.swift`), and ran the changed suites: `ClaudeHookRepair.reconcileMainClone` (10), `ClaudeHookConfigWriter crow-path` (9), `AppState` session-lookup (2), `hook-event session resolution` (8), and the disk-cleanup / grok-strip suites (13) — **all green**.

This is a careful, well-tested fix. The cwd-authoritative routing correctly distinguishes an owning id, the Manager carve-out, an ambiguous two-owner path, a dead owner, a foreign-live inherited block (dropped), and #897's stale non-live id (re-routed) — each with a dedicated test. `resolveMainClone` declines on any gitdir shape it can't confidently interpret (right call, since the strip deletes entries) and is cross-checked against real `git rev-parse` fixtures. `resolveCrowBinary` never emits a `.build/` path, and the reaper correctly excludes the being-deleted session from the ownership snapshot so it can't recreate #897's stale file.

### Security Review
**Strengths:**
- Strip is bounded to the strict `parseCrowHookCommand` shape, so a user's own hook under a managed event name survives (test: `userHookSurvivesStrip`). No path-prefix matching — `sessionIDs(forWorktreePath:)` is exact-match after `standardizingPath` on both sides, so a sibling worktree (`/wt/alpha-1-slug`) can't resolve to `/wt/alpha`.
- `oneLineForLog` caps length and escapes C0/DEL before caller-supplied `cwd`/`eventName` reach the launchd log; both drop-loggers are key-capped (256) and deduped, so an inherited block firing on every tool call can't flood the log.
- Non-atomic settings write is deliberate and documented (preserves the 0600 mode `writeGatewayEnv` sets for a gateway bearer token).

No injection, auth, or secret-exposure concerns found.

### Code Quality

**Yellow — `reconcileMainClone` runs its blocking file I/O *on* the MainActor at every launch site, contradicting its own docs (#892).**

The reconciliation is documented as off-MainActor in two places:
- `ClaudeHookRepair.swift:268` — "Performs blocking file I/O; call off the MainActor (#892)."
- `SessionService.swift:875` — the `HookOwnership` value-snapshot exists "so the reconciliation itself (blocking file I/O) stays off the MainActor (#892)."

But `SessionService` is `@MainActor` (`:13`), and every launch call site invokes `prepareWorktreeForAgentLaunch` — now including `reconcileMainCloneHooks` — synchronously on the MainActor: `launchAgent` (`:1160`), `handoffAgent` (`:1499`), `createManagerTerminal` (`:2021`), `pasteDeferredLaunch` (`:814`), and the `send` RPC (`EngineRouter.swift:1184`, inside `await MainActor.run` at `:1149` — right after `:1148` hoists `resolveCrowBinary` off the MainActor *specifically* to avoid the #892 wedge). Only `performDiskCleanup` (reaper, via `Task.detached`) actually runs off it — the one path the value-snapshot design pays off for.

So for the common case (a Claude/Codex/Cursor worktree launch) this now does `.git` + `commondir` + `settings.local.json` reads — and, on a foreign/stale block, a settings write — on the MainActor on every launch. Practical impact is small (a few tiny local reads, cheap early-out for non-worktrees) and `prepareWorktreeForAgentLaunch` already did MainActor file I/O via `seedTrust`, so this is not a new category. But the docs assert the opposite of what happens, in a codebase where those assertions are load-bearing, and the value-snapshot's stated rationale isn't met on the launch path. Fix either way: dispatch the launch-path reconcile off the MainActor, or correct the two doc claims to say it runs on the MainActor at launch and why that's acceptable.

**Green — consider:**
- The `reconcileMainClone` suite runs ~5 min locally (real `git init`/`git worktree add`, ~15s each, serialized). The fidelity is worth it, but it's a notable CI-time cost worth a glance.
- Theoretical: a live session running in a path registered to a *different* live session (not itself) would have its events dropped as inherited. Requires a pathological orphan-recovery state and routing genuinely can't disambiguate — not actionable, noting for completeness.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The Yellow is a doc/behavior mismatch on a named concurrency invariant with a latent (small) perf cost; the fix can be as light as correcting the two doc claims, or moving the launch-path reconcile off the MainActor.

---

dgershman (approved):
## Code & Security Review

Reviewed the full diff (1262/-122, 13 files), read the surrounding source in `EngineRouter`, `ClaudeHookRepair`, `ClaudeHookConfigWriter`, and `SessionService`, and ran the affected suites locally.

### Critical Issues
None.

### Verification
- `swift build` — clean (12.9s).
- Per-package tests all green: **CrowClaude 92/92** (incl. the new `ClaudeHookRepair.reconcileMainClone` suite), **CrowCore 527/527** (incl. `AppStateSessionLookup`), **CrowEngine 576/576** (incl. `hook-event session resolution`).
- The `reconcileMainClone` suite runs ~300s because it builds **real** `git init` + `git worktree add` fixtures and cross-checks resolution against `git rev-parse --git-common-dir`. Slow, but that's the right call — a fabricated `.git` could agree with the code while both disagree with git.

### What I checked closely

**cwd-authoritative routing (`EngineRouter`).** Traced every branch of the new resolution. The foreign-drop is provably lossless for the case it targets: when Claude loads both the worktree's block and the inherited main-clone block, the owning session reports the *same* event from its own block, so dropping the foreign copy fixes attribution + the duplicate without losing anything. All carve-outs are correct — owning id wins over an ambiguous pick, Manager is matched by constant (never silenced by a stray devRoot worktree row), a non-live provided id is still re-routed by cwd (#897 preserved), and a dead owner never drops a live session's event. Normal one-session-per-worktree, Codex/OpenCode-no-`--session`, and Manager setups are all unaffected.

**`resolveMainClone`.** Strict gitdir-chain resolution (`.git` file → `gitdir:` → `commondir`, else the `../..` derivation only when the gitdir has git's `worktrees/<name>` shape), and it *declines* rather than guesses on any unrecognized shape — the right posture for code that deletes hook entries. Both resolution paths converge and are cross-checked against git.

**`reconcileDirectory`.** The ownership-based health test correctly distinguishes a legitimately-hosted main-clone session (row maps dir→session, live, good binary → left alone / repaired) from an inherited block (no row → stripped), which is exactly the case the boot sweep's per-directory test cannot see. Strip removes only entries matching the strict metachar-rejecting `parseCrowHookCommand` shape, so user hooks under a managed event name survive (tested). Non-atomic in-place write correctly preserves the 0600 mode from `writeGatewayEnv`.

**Reaper path.** Reconcile fires only for genuine worktree items (review clones and main checkouts `continue` earlier), targets the main clone via `item.repoPath`, and uses ownership snapshotted with the deleting session excluded — so a reaped worktree's block is scrubbed immediately instead of at next boot, and a legit main-clone session's block is never counted as gone.

**`resolveCrowBinary` → nil.** Refusing to emit a `.build/` path is a sound tradeoff (a dangling block outlives its session and breaks every worktree of the repo); all production callers handle the optional (`if let` / `guard let` / passed into `HookOwnership.snapshot(crowPath: String?)`), and `LaunchScaffold` forwards it to `sweep`'s optional param. Nil only occurs in the debug-build + unwritable-`~/.local/share` edge case.

### Security Review
**Strengths:**
- The inherited-block cleanup can't be turned into arbitrary-file deletion: `resolveMainClone` is strict and declines on unknown shapes, and strip only touches entries matching the strict Crow-command grammar — every unrelated key survives.
- The hostile-PR vector doesn't apply: a Grok/Antigravity `.review` clone is a standalone `git clone` (its `.git` is a directory), so `resolveMainClone` returns `.notAWorktree` and step 3 no-ops — and `stripGrokConfigFromReviewClone` already strips the clone's committed `.claude/settings*` before the agent opens it. Normal worktree `.git` files are written by Crow's own `git worktree add`, not attacker-controlled.
- Log inputs are length-capped and C0/DEL-escaped (`oneLineForLog`); the foreign-drop logger is deduped and capped.

**Concerns:** None.

### Code Quality
- `parsedCrowHooks` / `writeSettings` factored out and shared with the existing sweep — no duplicated grammar or write path.
- `sessionID(forWorktreePath:)`'s new normalize+deterministic behavior is contained: no production caller other than its own delegation (routing moved to the plural `sessionIDs`).
- Consider (Green, non-blocking): `HookOwnership.sessionByWorktreePath` collapses a duplicate path with `uniquingKeysWith:{first,_ in first}` (dictionary-order pick) while routing's `sessionIDs` sorts — for the very rare two-sessions-sharing-one-main-clone case the reconcile's `owner` pick is nondeterministic (churn, not incorrectness). And the launch gate runs the reconcile's small I/O on the MainActor at the launch/send sites (same as the existing strip/seed), despite the "call off the MainActor" doc — negligible, and pre-existing to this gate.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 2 Green] findings. Careful, well-reasoned, thoroughly tested; the inverted `liveSessionIDWins` assertion is correctly identified as the original bug written down as intent.

---

dhilgaertner:
Thanks — the Yellow was a fair catch. Addressed in 9b075dd, though by fixing the docs rather than the dispatch. Reasoning, plus two notes on the Greens.

## Yellow — docs corrected, call deliberately left on the MainActor

You're right that all three claims were wrong: `prepareWorktreeForAgentLaunch` is synchronous and MainActor-bound, so every launch site runs the reconcile inline.

I went with the doc fix rather than dispatching off the MainActor, because **detaching would be a real bug, not just a different tradeoff.** The launch paths type the agent's command into its pane in the same synchronous block (`launchAgent` → `prepareWorktreeForAgentLaunch` → … → `TerminalRouter.send`). A detached reconcile would race the agent against the strip and let it start with the inherited block still on disk — exactly the failure this PR exists to prevent. The gate's other two steps have the same constraint and do MainActor file I/O for the same reason, so this isn't a new category so much as the same one you already identified.

The `send` RPC site is the ironic one you flagged, and I did look at hoisting just that one: it's genuinely async, but the ownership snapshot needs AppState, so it'd take a MainActor hop → detached reconcile → MainActor hop. That buys a few syscalls at the price of a third actor transition and a launch path whose ordering differs from the other four. Not worth it.

What I did instead is put the cost and the ordering constraint where the reflex to "fix" this would arise — `reconcileMainCloneHooks`, `reconcileMainClone`, and step 3 of the gate's own doc — so the next reader doesn't re-derive it:

- **one `stat`** for anything that isn't a linked worktree (main clones, review clones, the Manager's dev root — the early-out in `resolveMainClone`);
- **three more small reads** for a real worktree (`.git`, `commondir`, the settings file);
- **a write only when a block is actually stale or foreign** — a once-per-defect event, not once per launch.

`HookOwnership`'s rationale is corrected too: the value snapshot is what lets the *reaper* run this inside `Task.detached`, not a claim that the launch path does.

## Green 1 — the ~5 min suite time doesn't reproduce

Measured on this machine, that suite is **0.58 s** wall for all 12 tests (swift-testing runs them in parallel):

```
✔ Test "a row for a dead session does not count as an owner" passed after 0.579 seconds.
✔ Test run with 12 tests in 1 suite passed after 0.579 seconds.
```

The full `git init` + config + commit + `worktree add` fixture is ~0.27 s standalone. I suspect the ~15 s/test you saw was build time attributed to the run, or serialization under a sandbox. Happy to trim the fixtures if you can reproduce it in CI, but on this evidence the fidelity is costing ~0.6 s and I'd rather keep the real git.

## Green 2 — agreed, and it's bounded

Agreed it's not actionable, and worth recording why it's safe to leave: for a live session's events to be dropped, its cwd must be registered to a *different* live session and not to itself. Routing genuinely can't disambiguate that, but it also can't silently lose everything — the drop is logged (deduped per `cwd`+id) naming both session ids, so the state is diagnosable from the daemon log rather than presenting as silence.

dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- The strip/repair boundary is the strict `parseCrowHookCommand` shape, so a user's own hook sharing a managed event name is never removed (verified by `userHookSurvivesStrip`). The metacharacter disqualification (`;|&><\`$()`) keeps a wrapped/pipeline user command out of scope.
- `resolveMainClone` **declines rather than guesses** on any gitdir shape it doesn't recognize (`malformedGitFileResolvesToNil`) — important, since this code path deletes hook entries and a wrong resolution would aim the strip at an unrelated directory. Resolution is cross-checked against real `git rev-parse --git-common-dir`.
- Log values are length-capped and C0/DEL-escaped via the shared `oneLineForLog`, so a caller-supplied `cwd`/`eventName` can't forge or mangle a launchd log line. The foreign-drop dedup key is `(cwd, foreign-id)` — correctly finer-grained than the unresolved logger's `cwd`-only key, since one cwd legitimately sees several foreign ids over a daemon's life.
- `resolveCrowBinary` never emits a `.build/` product path (root cause of the dangling-block class); the stable fallback link is test-injectable so tests don't write into real `~/.local/share` (ADR 0012).

**Concerns:** none.

### Code Quality
- The EngineRouter resolution is correct across every branch: cwd-owner used when no id or a non-live/stale id is provided; a live foreign id dropped (throwing **before** the store write, so no spurious `hookState` row is minted for the main clone — asserted in `foreignLiveSessionIDIsDropped`); Manager and cwd-owning id honored; ambiguous-path prefers the provided owner. `owners[0]` is deterministic because `sessionIDs(forWorktreePath:)` is sorted.
- Reaper path is sound: `HookOwnership.excluding(sessionID:)` treats the deleting session as already gone (so a block naming it is stripped, not preserved — the exact mechanism that created #897's artifact), and reconcile is skipped for review clones and main checkouts.
- `sessionIDs(forWorktreePath:)` normalizes both sides (`standardizingPath`) and is exact-match, not prefix — `sessionIDForWorktreePathNormalizesBothSides` pins that a sibling `repo-1-slug` never resolves to `repo`.
- Shared helpers (`parsedCrowHooks`, `writeSettings`) are factored out and reused by both the sweep and the new reconcile, keeping the non-atomic-write-preserves-0600 rationale in one place.
- Rationale is documented thoroughly, and the inverted `liveSessionIDWins` assertion is called out explicitly as "the assertion was the bug" — a correct, deliberate change, not code-fitting.

Verification: `make build` green; `swift test` green for CrowClaude (92 tests, including the real-`git init`/`worktree add` `reconcileMainClone` fixtures), CrowEngine (586), and the new CrowCore session-lookup tests.

Green (consider), none blocking:
- Launch reconcile runs inline on the MainActor while the reaper reconcile runs detached; both write the same main clone's `settings.local.json` non-atomically. The cross-session window is narrow (a write happens only on a genuinely stale/foreign block, then subsequent reconciles no-op), idempotent, and self-healing at next launch/boot — and it's the same non-atomic pattern `repairDirectory`/`writeGatewayEnv` already use, so no new risk class. Worth a mental note only.
- `.skipped` ("could not reconcile … may keep emitting hook errors") is logged at `info`; a `warning` might surface it better, but it's defensible.
- The `reconcileMainClone` suite is slow (~300s) because each case shells out to real `git` — a deliberate fidelity tradeoff (a fabricated `.git` could agree with the code while both disagree with git), and the reason is documented.

### Summary Table

---

PR #923, opened 2026-07-29
Sidebar polish: center the right icon stack + group the ticket counts (CROW-922)

dgershman:
Closes #922

Follow-up to the two-column sidebar top (#918 / CROW-917). Two spacing/alignment tweaks now that the grid is in place — both in the served `web/app.css` (+ comment/guard updates in `web/app.js`).

## Changes
**Right icon column — group + center (was: spread across 4 rows).**
`.sidebar-right > button` no longer flex-fills the column (`flex: 1 1 0` → `flex: 0 0 auto`, fixed `height: 28px`), and `.sidebar-right` gains `justify-content: center`. The bell/gear/+/select buttons now read as one tight, vertically-centered stack instead of dividing the full column height. 28px stays above the WCAG 2.2 §2.5.8 24px target-size minimum, so the old `min-height: 24px` floor (and its coupling to the left column's height) is dropped. The bell's unread badge is `position: absolute` on `.tk-bell`, so it stays anchored to the bell.

**Ticket counts — group + center (was: edge-to-edge).**
`.tickets-counts` switches `justify-content: space-between` → `center` (gap 4px → 8px) so the 5 status mini-counts sit together, centered, in the wide box. Each `.tk-count` is content-sized, so glyph + number stay together.

Drift-prone comments in `app.css` and the Manager-pill guard note in `app.js` were rewritten to match the new fixed-size/centered behavior.

## Test plan
Visual (per ticket — subjective polish; box height and column width from #918 unchanged):
- Sidebar top right column: bell / gear / + / select sit as a centered block at natural size, no big vertical gaps; bell unread badge still pinned to the bell's top-right corner.
- Tickets box: the 5 status counts (Backlog · Ready · In Progress · In Review · Done·24h) sit grouped and centered, not stretched edge-to-edge.
- Cold-start / Manager-less render: icon buttons keep their 28px size (no shrink below the touch-target floor).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #922, opened 2026-07-29
Sidebar polish: push right-column icons together + center them, and center/group the ticket counts (follow-up to #918)


dgershman:
Follow-up to the two-column sidebar from **#918** (#917). Two spacing/alignment tweaks now that the grid is in place.

## What's off
1. **Right icon column is spread out.** The four icon buttons stretch to divide the whole column height:
   ```css
   /* app.css:683,691 */
   .sidebar-right { flex: 0 0 auto; width: 30px; display: flex; flex-direction: column; gap: 6px; }
   .sidebar-right > button { flex: 1 1 0; width: 100%; height: auto; min-height: 24px; }
   ```
   Because `.sidebar-top { align-items: stretch }` (`app.css:681`) makes the right column match the left height and each button is `flex: 1 1 0`, the bell/gear/"+"/Select spread apart across 4 rows with big gaps.
2. **Ticket counts are spread edge-to-edge.** `.tickets-counts { justify-content: space-between }` (`app.css:692`) pushes the 5 status mini-counts to the full width instead of grouping them (`ticketsCard` builds them at `app.js:1160-1165`; `.tk-count` at `app.css:791`).

## Ask
- **Push the icons together and center them** vertically in the right column (natural button size, grouped as a block) instead of stretching to fill.
- **Center / group the ticket counts** (and any count badges) rather than spreading them across the box.
- Apply the same centered/grouped treatment consistently to the tickets box counts.

## Implementation direction
- Right column: change `.sidebar-right > button` from `flex: 1 1 0` (stretch) to natural/fixed size (e.g. `flex: 0 0 auto`, square ~24-28px), and center the stack with `.sidebar-right { justify-content: center }` — `app.css:683,691`. (With the buttons no longer flex-filling, `justify-content: center` keeps them from bunching at the top under the `align-items: stretch` at `app.css:681`.) Keep the bell's unread badge anchored on the bell.
- Ticket counts: change `.tickets-counts` from `justify-content: space-between` to `center` (keep the gap) so the 5 counts sit together, centered — `app.css:692`. Ensure count glyph + number stay centered within each `.tk-count` (`app.css:791`).
- Subjective visual polish — match the rendered look; the box height and column width from #918 stay as-is.

## Files
[...]

PR #925, opened 2026-07-29
Sidebar: center the "Tickets" title in the tickets box (CROW-924)

### LINKED ISSUE #924: Sidebar: center the "Tickets" title in the tickets box (follow-up to #922)
Follow-up to the sidebar tickets box (#922 / #918 / #917). Small alignment tweak: the **"Tickets"** title should be **centered** in the box rather than left-aligned.

## What's off
The title sits flush-left because the head row splits the title and the refresh (↻) button edge-to-edge:

```css
/* app.css:662-663 */
.tickets-head { display: flex; align-items: center; justify-content: space-between; gap: 4px; }
.tickets-title { color: var(--gold); font-weight: 700; font-size: 13px; overflow: hidden; text-overflow: ellipsis; white-space: nowrap; min-width: 0; }
```

`justify-content: space-between` pins "Tickets" to the left edge and the ↻ refresh button to the right edge (`ticketsCard` builds `head` = title + refresh at `app.js:~1136-1145`).

## Ask
- Center the word **"Tickets"** horizontally in the tickets box.
- Keep the refresh (↻) button reachable and its spinner behavior intact (CROW-797) — it should not push the title off-center.

## Implementation direction
The title needs to center relative to the **whole box**, not just the space left of the refresh button. A couple of options:
- **Absolutely position the refresh button** (e.g. `.tickets-refresh { position: absolute; right: 8px }` with `.tickets-head { position: relative; justify-content: center }`), so the title centers in the full width and the ↻ floats at the right edge. `.tickets-card` is already the positioning context candidate (`position` would go on `.tickets-head`).
- Or a 3-column grid on `.tickets-head` (spacer · centered title · refresh) so the title is centered with the button balanced on the right.

Prefer whichever keeps the ↻ hover/disabled/`.spinning` states (`app.css:664-674`) unchanged. Subjective visual polish — match the rendered look; box size/counts from #922 stay as-is.

## Files
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css` — `.tickets-head` (`:662`), `.tickets-title` (`:663`), `.tickets-refresh` (`:664`)
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` — `ticketsCard` head build (`~:1136-1145`) if any DOM change is needed

dgershman:
Closes #924

Follow-up to the sidebar tickets box chain (#918 → #923). Small alignment tweak: center the **"Tickets"** title in the box rather than pinning it flush-left.

## What changed
`.tickets-head` (`app.css`) goes from a `space-between` flex row to a 3-column grid:

```css
.tickets-head  { display: grid; grid-template-columns: 22px 1fr 22px; align-items: center; gap: 4px; }
.tickets-title { grid-column: 2; text-align: center; … }
.tickets-refresh { grid-column: 3; … }
```

- A **button-width spacer** (col 1, 22px) balances the refresh button (col 3, 22px), so the title in the middle `1fr` track centers relative to the **whole box** — not just the space left of ↻.
- The ↻ stays **in-flow** in its own column, so it never overlaps the title even when the title ellipsises under a raised browser min-font-size. This is the overlap the CROW-913 comment was guarding against on the old compact card; now that the card is full-width (CROW-922) that constraint is gone, which the rewritten comment records.

No DOM change (`ticketsCard` head is still `[title, refresh]` — the refresh is pinned to col 3 explicitly so grid auto-placement doesn't drop it in the empty col 1).

## Untouched
Refresh hover / disabled / `.spinning` states (CROW-797) — unchanged; box size and counts from CROW-922 stay as-is.

## Test plan
- Open the sidebar: **"Tickets"** reads centered in the box; ↻ sits at the right edge.
- Hover ↻ (chrome highlights), click it → spinner turns inside a stationary button; title stays centered while refreshing.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -651,21 +651,28 @@ html, body {
      failure mode). The right icon column no longer divides the left's height
      (CROW-922), so growing here just makes the row taller — nothing to re-derive. */
   margin: 0; box-sizing: border-box; min-height: 64px;
+  /* Refresh-button edge, shared by .tickets-refresh's box and the mirror spacer track in
+     .tickets-head — one source of truth so the two can't drift and silently de-centre the
+     title (CROW-924). At/below the WCAG 2.2 §2.5.8 24px target min today; a future bump to
+     match the sibling icon column (app.css :688 uses 28px) is a one-token change here. */
+  --tk-refresh: 22px;
   display: flex; flex-direction: column; justify-content: space-between;
 }
 .tickets-card:hover { border-color: var(--border-strong); }
 .tickets-card.selected { border-color: var(--gold); }
 /* Centred title, refresh right (CROW-924): now the .tickets-card is full-width (CROW-922)
    the compact-card overlap that forced a left-aligned space-between row (CROW-913) is gone.
-   A 3-column grid — a spacer matching the button width, the title in 1fr, then the refresh —
-   centres the title relative to the whole box while the ↻ stays balanced at the right edge
-   and in-flow (no overlap even when the title ellipsises under a raised min-font-size). */
-.tickets-head { display: grid; grid-template-columns: 22px 1fr 22px; align-items: center; gap: 4px; }
+   A 3-column grid — a left spacer track that mirrors the button width via --tk-refresh, the
+   title in 1fr, then the refresh — centres the title relative to the whole box while the ↻
+   stays balanced at the right edge and in-flow (no overlap even when the title ellipsises
+   under a raised min-font-size). Both spacer and button read the one --tk-refresh so they
+   can't disagree and drift the title off-centre. */
+.tickets-head { display: grid; grid-template-columns: var(--tk-refresh) 1fr var(--tk-refresh); align-items: center; gap: 4px; }
 .tickets-title { grid-column: 2; text-align: center; color: var(--gold); font-weight: 700; font-size: 13px; overflow: hidden; text-overflow: ellipsis; white-space: nowrap; min-width: 0; }
 .tickets-refresh {
   grid-column: 3;
   display: inline-flex; align-items: center; justify-content: center;
-  width: 22px; height: 22px; padding: 0; border-radius: 6px; cursor: pointer; line-height: 1;
+  width: var(--tk-refresh); height: var(--tk-refresh); padding: 0; border-radius: 6px; cursor: pointer; line-height: 1;
   border: 1px solid var(--border-subtle); background: var(--bg-card); color: var(--text-secondary); font-size: 12px;
 }
 .tickets-refresh:hover { color: var(--text-primary); border-color: var(--border-strong); }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -655,14 +655,15 @@ html, body {
 }
 .tickets-card:hover { border-color: var(--border-strong); }
 .tickets-card.selected { border-color: var(--gold); }
-/* Title left, refresh right (CROW-913): the compact .tickets-card is too narrow for a
-   centred title + absolutely-positioned refresh (they overlap below ~110px), so the head
-   is a space-between row with the refresh in-flow. The title ellipsises only under a
-   heavily raised browser minimum-font-size, where the card also grows toward its cap. */
-.tickets-head { display: flex; align-items: center; justify-content: space-between; gap: 4px; }
-.tickets-title { color: var(--gold); font-weight: 700; font-size: 13px; overflow: hidden; text-overflow: ellipsis; white-space: nowrap; min-width: 0; }
+/* Centred title, refresh right (CROW-924): now the .tickets-card is full-width (CROW-922)
+   the compact-card overlap that forced a left-aligned space-between row (CROW-913) is gone.
+   A 3-column grid — a spacer matching the button width, the title in 1fr, then the refresh —
+   centres the title relative to the whole box while the ↻ stays balanced at the right edge
+   and in-flow (no overlap even when the title ellipsises under a raised min-font-size). */
+.tickets-head { display: grid; grid-template-columns: 22px 1fr 22px; align-items: center; gap: 4px; }
+.tickets-title { grid-column: 2; text-align: center; color: var(--gold); font-weight: 700; font-size: 13px; overflow: hidden; text-overflow: ellipsis; white-space: nowrap; min-width: 0; }
 .tickets-refresh {
-  flex: 0 0 auto;
+  grid-column: 3;
   display: inline-flex; align-items: center; justify-content: center;
   width: 22px; height: 22px; padding: 0; border-radius: 6px; cursor: pointer; line-height: 1;
   border: 1px solid var(--border-subtle); background: var(--bg-card); color: var(--text-secondary); font-size: 12px;
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -653,8 +653,9 @@ html, body {
   margin: 0; box-sizing: border-box; min-height: 64px;
   /* Refresh-button edge, shared by .tickets-refresh's box and the mirror spacer track in
      .tickets-head — one source of truth so the two can't drift and silently de-centre the
-     title (CROW-924). At/below the WCAG 2.2 §2.5.8 24px target min today; a future bump to
-     match the sibling icon column (app.css :688 uses 28px) is a one-token change here. */
+     title (CROW-924). 22px is below the WCAG 2.2 §2.5.8 24px target-size minimum; a bump to
+     match the sibling `.sidebar-right > button` (28px) is a one-token change here, deferred
+     to a repo-wide a11y pass (the whole #917→#923 sidebar chain sits under 24px too). */
   --tk-refresh: 22px;
   display: flex; flex-direction: column; justify-content: space-between;
 }
--- a/Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift
+++ b/Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift
@@ -431,4 +431,26 @@ import Testing
             !css.contains("width: max-content; max-width: 160px"),
             "the #913 content-sized width cap is superseded — the card is full-width in the left column now")
     }
+
+    /// CROW-924: the "Tickets" title centres in the full-width box, superseding the
+    /// left-pinning `space-between` head CROW-913 used while the card was compact. The
+    /// centring is a 3-column grid whose two outer tracks both read `--tk-refresh` — the
+    /// same token the refresh button sizes from — so the left spacer can't drift from the
+    /// button's width and silently de-centre the title. That single source of truth is the
+    /// durability fix; the two literal `22px`s it replaces were a standing, untested trap
+    /// (a WCAG target-size bump to one would de-centre with every suite still green). Pin
+    /// the whole grid rule (a revert to `space-between` or to hardcoded outer widths fails
+    /// here) and that the button derives its box from the same token.
+    @Test func ticketTitleCentresViaGrid() throws {
+        // stripComments: `--tk-refresh`, "grid", and "space-between" all appear in the
+        // .tickets-card / .tickets-head doc comments, so a raw-file `contains` would
+        // false-pass off the prose once a declaration is deleted (mutation-checked).
+        let css = Self.stripComments(try Self.webAsset("app.css"))
+        #expect(
+            css.contains(".tickets-head { display: grid; grid-template-columns: var(--tk-refresh) 1fr var(--tk-refresh); align-items: center; gap: 4px; }"),
+            "the title must centre via the var-driven 3-column grid, not a left-pinning space-between row (CROW-924)")
+        #expect(
+            css.contains("width: var(--tk-refresh); height: var(--tk-refresh);"),
+            "the refresh button must size from --tk-refresh so the mirror spacer track can't drift from it and de-centre the title (CROW-924)")
+    }
 }
--- a/Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift
+++ b/Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift
@@ -440,8 +440,9 @@ import Testing
     /// durability fix; the two literal `22px`s it replaces were a standing, untested trap
     /// (a WCAG target-size bump to one would de-centre with every suite still green). Pin
     /// the whole grid rule (a revert to `space-between` or to hardcoded outer widths fails
-    /// here), that the button derives its box from the same token, and that the token is
-    /// actually defined (a dangling var() drops the grid to `none` and the button to `auto`).
+    /// here), the title's own centre-track placement + alignment (the invariant itself),
+    /// that the button derives its box from the same token, and that the token is actually
+    /// defined (a dangling var() drops the grid to `none` and the button to `auto`).
     @Test func ticketTitleCentresViaGrid() throws {
         // stripComments: `--tk-refresh`, "grid", and "space-between" all appear in the
         // .tickets-card / .tickets-head doc comments, so a raw-file `contains` would
@@ -453,14 +454,25 @@ import Testing
         #expect(
             css.contains("width: var(--tk-refresh); height: var(--tk-refresh);"),
             "the refresh button must size from --tk-refresh so the mirror spacer track can't drift from it and de-centre the title (CROW-924)")
+        // Pin the title's own placement + alignment — the invariant this change exists for.
+        // Both declarations are load-bearing and drop silently: without `text-align: center`
+        // the title left-aligns in the 1fr track (the pre-CROW-924 look); without
+        // `grid-column: 2` too, the title (first child) auto-places into the 22px spacer
+        // track and ellipsises to nothing at the left edge. Anchor on the selector so a
+        // revert of either fails here.
+        #expect(
+            css.contains(".tickets-title { grid-column: 2; text-align: center;"),
+            "the title must sit in the centre track and centre its text there, or it auto-places into the 22px spacer (CROW-924)")
         // Pin the token's DEFINITION, not just its two consumers: an unresolvable var()
         // computes to the property's unset value — grid-template-columns → `none` (the
         // explicit grid vanishes, title no longer centred or ellipsising) and the button's
         // width/height → `auto` (collapses to the glyph, losing the CROW-797 stationary box).
-        // stripComments drops the doc-comment mentions of the token, so this matches only the
-        // real declaration and fails if it's deleted (the mutation the review found green).
+        // stripComments drops the doc-comment mentions of the token, and `var(--tk-refresh)`
+        // has no colon, so `--tk-refresh:` matches ONLY the declaration and fails if it's
+        // deleted. Value-agnostic on purpose: the documented WCAG 2.5.8 bump (22px→28px)
+        // stays a genuine one-token edit here rather than a two-place change in this test too.
         #expect(
-            css.contains("--tk-refresh: 22px;"),
+            css.contains("--tk-refresh:"),
             "the --tk-refresh token both the head's spacer track and the button derive from must be defined, or the var() references resolve to `none`/`auto` (CROW-924)")
     }
 }
--- a/Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift
+++ b/Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift
@@ -440,7 +440,8 @@ import Testing
     /// durability fix; the two literal `22px`s it replaces were a standing, untested trap
     /// (a WCAG target-size bump to one would de-centre with every suite still green). Pin
     /// the whole grid rule (a revert to `space-between` or to hardcoded outer widths fails
-    /// here) and that the button derives its box from the same token.
+    /// here), that the button derives its box from the same token, and that the token is
+    /// actually defined (a dangling var() drops the grid to `none` and the button to `auto`).
     @Test func ticketTitleCentresViaGrid() throws {
         // stripComments: `--tk-refresh`, "grid", and "space-between" all appear in the
         // .tickets-card / .tickets-head doc comments, so a raw-file `contains` would
@@ -452,5 +453,14 @@ import Testing
         #expect(
             css.contains("width: var(--tk-refresh); height: var(--tk-refresh);"),
             "the refresh button must size from --tk-refresh so the mirror spacer track can't drift from it and de-centre the title (CROW-924)")
+        // Pin the token's DEFINITION, not just its two consumers: an unresolvable var()
+        // computes to the property's unset value — grid-template-columns → `none` (the
+        // explicit grid vanishes, title no longer centred or ellipsising) and the button's
+        // width/height → `auto` (collapses to the glyph, losing the CROW-797 stationary box).
+        // stripComments drops the doc-comment mentions of the token, so this matches only the
+        // real declaration and fails if it's deleted (the mutation the review found green).
+        #expect(
+            css.contains("--tk-refresh: 22px;"),
+            "the --tk-refresh token both the head's spacer track and the button derive from must be defined, or the var() references resolve to `none`/`auto` (CROW-924)")
     }
 }


dhilgaertner (changes requested):
## Code & Security Review

CSS-only diff (+8 / −7, one file). The implementation matches the ticket's own preferred direction (#924's "3-column grid" option, chosen over the absolute-positioning option specifically because it keeps the ↻ hover / disabled / `.spinning` states untouched — it does).

### Critical Issues
None.

### What I verified (statically — see *Tests* below)

- **The centering is geometrically exact.** `grid-template-columns: 22px 1fr 22px` with `gap: 4px` puts the `1fr` track 26px from each edge of `.tickets-head`, whose content box spans the card's full inner width. `text-align: center` in that symmetric track centers "Tickets" relative to the whole box — the ask in #924, not just "left of ↻". `* { box-sizing: border-box }` (app.css:24) is in force, so the refresh button's 1px border is inside its 22px width and its border box exactly fills the 22px col-3 track — no 2px overflow shifting it past the card's padding edge.
- **No DOM change is needed, and none was made.** Col 1 is an *empty track*, not a spacer element — correct call, nothing extra for a screen reader to skip. Both children are explicitly placed (`grid-column: 2` / `3`), so auto-placement can't drop the ↻ into the empty col 1, as the PR body notes.
- **Ellipsis behavior is preserved.** `.tickets-title` keeps `min-width: 0` + `overflow: hidden`, so the `1fr` track's automatic minimum resolves to 0 and the title ellipsises instead of pushing ↻ out of the row. The comment's claim that the ↻ "stays in-flow, no overlap even when the title ellipsises" holds — fixed tracks can't overlap.
- **`flex: 0 0 auto` → `grid-column: 3` is safe.** `.tickets-refresh` is constructed in exactly one place (`ticketsCard`, app.js:1140-1145) inside `.tickets-head`, so it has no other flex parent to lose sizing behavior in. Button geometry is identical before/after (definite 22×22 either way), so the 12px `.action-spinner` swap-in (CROW-797) and the `.spinning` opacity/cursor states are genuinely untouched.
- **The rewritten comment is accurate,** including its reason for retiring the CROW-913 constraint (the card is full-width since CROW-922, so the compact-card overlap it guarded against no longer exists).

### Security Review

**Strengths:**
- Presentational CSS only — no JS, no DOM construction, no template/HTML interpolation, no `innerHTML`, no event handlers, no network or storage access. Zero attack surface delta.
- No `position: absolute` / `z-index` / overlay introduced, so no clickjacking-adjacent overlap over the ↻ control (the grid keeps it in-flow, which is the *safer* of the two options #924 proposed).
- No secrets, no user-controlled values, no injection vector.

**Concerns:** none.

### Code Quality

**🟡 Yellow — the new centering depends on two hardcoded `22px` values agreeing, and nothing pins the layout.**

`app.css:663` hardcodes the spacer track at `22px` to mirror `.tickets-refresh { width: 22px }` at `app.css:668`. Change one and the title silently de-centers (and the button overflows its track, eating into the card's right padding) with **every suite still green** — no test in the repo mentions `.tickets-head`.

That's not a hypothetical: the immediately preceding PR in this chain (#923 / CROW-922) bumped the *sibling* icon column's buttons from 24px to 28px for WCAG 2.2 §2.5.8, and `.tickets-refresh` at 22px is a standing candidate for the same bump (see Green #1). The next person to make that edit has no signal that a second number must move with it.

This is also the exact case `WebTerminalAssetTests.swift` exists for — its own criterion is "layout decisions with no other pin, verified to regress silently (reverting any leaves every suite green)" (:413-421), and #923 added two whole-rule pins under that rule for changes of precisely this kind. Reverting this PR's `.tickets-head` to `display: flex; justify-content: space-between` currently breaks nothing.

Two ways to close it (either is fine; **a** is the smaller diff, **b** removes the coupling outright):

**(a) One source of truth for the button size:**
```css
.tickets-card   { --tk-refresh: 22px; }
.tickets-head   { display: grid; grid-template-columns: var(--tk-refresh) 1fr var(--tk-refresh); align-items: center; gap: 4px; }
.tickets-refresh{ grid-column: 3; width: var(--tk-refresh); height: var(--tk-refresh); … }
```

**(b) No magic number at all** — symmetric flexible tracks center the title regardless of the button's width:
```css
.tickets-head    { display: grid; grid-template-columns: 1fr auto 1fr; align-items: center; gap: 4px; }
.tickets-title   { grid-column: 2; … }
.tickets-refresh { grid-column: 3; justify-self: end; … }
```
(Col 3's `1fr` floors at the button's 22px min-content contribution, so the ↻ still can't be squeezed or overlapped; below that width the title ellipsises exactly as it does today.)

Plus a pin, so the CROW-924 fix itself has regression cover — a whole-rule anchor in `ticketBoxIsFullWidthAndShort` (or its own `@Test`) in the style #923 established:
```swift
#expect(
    css.contains(".tickets-head { display: grid; grid-template-columns: "),
    "the title must centre via the 3-column grid, not a left-pinning space-between row (CROW-924)")
```
Note that test reads the CSS **without** `stripComments`, and the new doc comment contains the words "space-between" and "3-column grid" — so a negative pin like `!css.contains("space-between")` would false-fail, and a bare `contains("grid-template-columns")` would false-pass off the prose. Anchor on the selector, or use `Self.stripComments(...)` as the CROW-917/922 pin at :391 does.

**🟢 Green — while you're in this rule: the ↻ is 22×22, under the WCAG 2.2 §2.5.8 24px target-size minimum.** Pre-existing (this PR doesn't change the size), so not a blocker — but the codebase holds that line explicitly for the sidebar's other icon buttons (app.css:686-692 cites §2.5.8 and uses a 28px floor), and the sidebar is the full-width touch view under 720px. If you adopt (a) or (b) above, bumping to 24px is a one-token change and the centering follows automatically.

**🟢 Green — no `CHANGELOG.md` `[Unreleased]` bullet.** `CONTRIBUTING.md:113` asks every PR for one. Flagging for consistency only, not as a blocker: the entire #917 → #918 → #922 → #923 chain skipped it too, so this is a repo-wide lapse rather than something this PR introduced.

### Tests

- **No existing test breaks.** I checked every `app.css` assertion in `WebTerminalAssetTests.swift` against the post-change file: `min-height: 64px` (:425) ✓ still at app.css:653, `".tickets-counts { display: flex;"` (:428) ✓ still present verbatim, `!contains("width: max-content; max-width: 160px")` (:431) ✓ still absent, and the two CROW-922 whole-rule pins (:396, :401) are untouched by this diff.
- **Not executed, for disclosure:** `swift test` was not run (no `.build` in the worktree — a full resolve+compile for a CSS-only diff), so the above is a static string-level check, not a green run. The jsdom suite (`Packages/CrowDaemon/web-tests`) couldn't run either — `node_modules` is absent and `npm install` needs network the sandbox blocks — and it wouldn't cover this regardless, since jsdom doesn't load stylesheets. That gap is exactly why the Swift asset pin in the Yellow above is the only available regression cover for this change.
- Centering/ellipsis claims above are from reading the cascade (confirmed `* { box-sizing: border-box }`, confirmed `.tickets-head` has no other rule or media-query override, confirmed `.tickets-refresh` has one construction site), not from a rendered screenshot. Worth one visual pass per the PR's own test plan before merge.

### Summary Table

The change itself is correct and well-reasoned, and the comment rewrite is the good kind of documentation. The one Yellow is about durability, not behavior: de-duplicate the `22px` and add the asset-test pin this repo's own convention calls for, and this is a clean approve.

---

dgershman:
Thanks — addressed the Yellow in e64f2628.

**Yellow (the `22px`↔`22px` coupling, unpinned):** took option **(a)** — hoisted the button edge to a single `--tk-refresh: 22px` custom property on `.tickets-card`. Both the `.tickets-head` left spacer track and `.tickets-refresh`'s `width`/`height` now read it, so they can't drift and silently de-center the title. Kept the exact-symmetry `Xpx 1fr Xpx` geometry you verified rather than `1fr auto 1fr` (b), because the var keeps centering perfect at every width incl. the raised-min-font ellipsis edge case, and — per your Green #1 — it makes the target-size bump a one-token change.

**Pin:** added `ticketTitleCentresViaGrid` to `WebTerminalAssetTests`. Per your note that this test reads the CSS and `--tk-refresh` / "grid" / "space-between" now all appear in the doc comments, it uses `Self.stripComments(...)` (like the :391 pin) rather than raw `css`, and anchors on the full rule text:
- `.tickets-head { display: grid; grid-template-columns: var(--tk-refresh) 1fr var(--tk-refresh); align-items: center; gap: 4px; }` — a revert to `space-between` or to hardcoded outer widths fails here.
- `width: var(--tk-refresh); height: var(--tk-refresh);` — pins that the button sizes from the same token, so the coupling itself is under test.

Reverting `.tickets-head` to `display: flex; justify-content: space-between` now breaks the suite, closing the "reverting leaves every suite green" gap.

**Green #1 (↻ under the 24px §2.5.8 target min):** left at 22px deliberately — it's pre-existing and the ticket scopes this to centering with "box size/counts from #922 stay as-is", and the sibling column's 28px floor is a repo-wide call better made across the whole chain than smuggled in here. The `--tk-refresh` var makes it a one-token follow-up whenever that's taken up (and centering follows automatically, as you noted).

**Green #2 (no CHANGELOG `[Unreleased]` bullet):** as you flagged, the whole #917→#923 chain skipped it, so I've left it consistent with the chain rather than introduce a lone entry — happy to add one if you'd prefer to start the habit here.

**Verification:** same constraint you hit — `swift test` needs a full resolve+compile with no `.build` and the jsdom suite needs blocked network, so I didn't get a green run. I verified both new pins by exact-string (`grep -F`) match against the built asset (the only failure mode for asset-text pins) and confirmed no bare `22px` remains in the tickets rules; `var(--tk-refresh)` resolves to `22px` so the geometry you validated is byte-for-byte unchanged.

dhilgaertner (changes requested):
## Code & Security Review

The layout change itself is correct. I checked the centring math: tracks are `22 + 4 + (W−52) + 4 + 22`, so the `1fr` title track spans `[26, W−26]` and its midpoint is exactly `W/2` — genuinely centred on the whole box, not on the space left of ↻. Both items are explicitly placed (`grid-column: 2` / `3`), so auto-placement never drops the button into the empty spacer track. The title still ellipsises rather than widening the grid (`min-width: 0` zeroes the `minmax(auto, 1fr)` minimum, and `overflow: hidden` would too), and because ↻ is in its own in-flow track the CROW-913 overlap is structurally impossible now — not just unlikely. `justify-self: stretch` blockifies the `<span>` to fill the `1fr` track so `text-align: center` bites. Refresh hover / `:disabled` / `.spinning` states are untouched as claimed.

`swift test --filter WebTerminalAssetTests` → 13/13 pass.

### Security Review
**Strengths:**
- No security surface: CSS-only plus a test. No DOM change, no new user-controlled data path.
- The head is still built through `el()`, which assigns `textContent` (`app.js:970`) — the title and `↻` are static literals, so no injection vector is introduced or widened.
- No new network, auth, or config surface; nothing reads a secret.

**Concerns:** none.

### Code Quality

**🟡 The new test pins both consumers of `--tk-refresh` but not the definition — deleting the token leaves the suite green with a broken head** (`Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift:444`)

I mutation-checked this: removing `--tk-refresh: 22px;` from `.tickets-card` (`app.css:658`) and re-running the suite gives **13/13 passing**, including `ticketTitleCentresViaGrid()`. Both `#expect`s match on literal text that still contains `var(--tk-refresh)`, so neither notices the referent is gone.

What actually renders in that state is badly wrong, not subtly wrong. An unresolvable `var()` is invalid at computed-value time, so each declaration falls back to the property's unset value:

- `grid-template-columns: var(--tk-refresh) 1fr var(--tk-refresh)` → `none`. The explicit grid vanishes; `grid-column: 2` / `3` now create *auto-sized implicit* columns, so the title is neither centred nor in a `1fr` track (and no longer ellipsises — it sizes to content).
- `width` / `height: var(--tk-refresh)` → `auto`. The button collapses to its glyph's intrinsic box, and the `.spinning` state — which swaps the `↻` text for a `.action-spinner` child — loses the fixed box that keeps the chrome stationary (the CROW-797 behaviour).

This matters because it's precisely the invariant the test's own doc comment claims to be protecting: *"That single source of truth is the durability fix … the two literal `22px`s it replaces were a standing, untested trap."* The PR replaces one untested trap (two literals that can drift) with a different untested one (a dangling reference). Given how deliberately mutation-checked the rest of this file is, closing it is one line:

```swift
#expect(
    css.contains("--tk-refresh: 22px;"),
    "the token the head's spacer track and the button both derive from must exist — an unresolvable var() drops grid-template-columns to `none` and the button to auto (CROW-924)")
```

Alternatively, make the consumers self-healing with `var(--tk-refresh, 22px)` — but I'd prefer the assertion, since a silent fallback would hide the deletion rather than surface it.

**🟡 Stale line pointer in the new comment** (`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css:657`)

> `a future bump to match the sibling icon column (app.css :688 uses 28px)`

Line 688 is a comment terminator (`centered block rather than spreading across the height (CROW-922). */`). The `min-height: 28px` declaration is at **699**. It doesn't resolve pre-change either — the hunk adds 8 net lines above it, putting the declaration at 691 on `main`, so the pointer was never correct. Hardcoded line numbers in this file rot on every edit above them; naming the selector survives:

```
a future bump to match the sibling icon column (`.sidebar-right > button` uses 28px) is a one-token change here
```

**🟢 Token is declared a level above both its uses** — `--tk-refresh` sits on `.tickets-card` (`app.css:658`) while both consumers live in `.tickets-head` / `.tickets-refresh`, so the button's box silently depends on inheritance through the card. Harmless today (the button is only ever rendered by `ticketsCard()`), but `.tickets-head` would keep the declaration adjacent to its uses. Stylistic.

**🟢 22px is still under the WCAG 2.2 §2.5.8 24px minimum** — unchanged from the old literal, so not a regression, and the new comment does flag it. Two notes: "At/below the 24px target min" reads as though 22px might satisfy it, and it doesn't — 22 < 24, full stop; and the spacing exception is hard to lean on here since the enclosing card is itself a click target (`card.onclick`, `app.js:1134`). The token now makes this a genuine one-line fix, which is the right setup — worth just doing it, in this PR or a follow-up.

**🟢 `stripComments` truncates at `//`** — `app.css:101` embeds `http://www.w3.org/2000/svg` inside a `url("data:image/svg+xml,…")`, so that line is mangled in the stripped copy. Pre-existing (the CROW-922 test at :391 already relies on this helper) and harmless for both new assertions, which target lines 670/675. Flagging only so nobody later writes an assertion over that region and gets a confusing false negative.

### Summary Table

The CSS is right and I'd take it as-is. What holds this back is that the test guarding it doesn't guard the thing it says it guards — verified by mutation, not inferred — plus a cross-reference that points at the wrong line. Both are small, mechanical fixes.

---

dgershman:
Thanks for the mutation-checked review — both Yellows fixed in 0e782b07.

**Y1 (test pinned the consumers, not the definition):** added a third `#expect` to `ticketTitleCentresViaGrid` pinning `--tk-refresh: 22px;` itself. It runs over the `stripComments`'d CSS, so it matches only the real declaration, not the doc-comment mentions of the token — which is what makes deleting the declaration fail it. I re-ran your exact mutation to confirm:

- Deleted `--tk-refresh: 22px;` → `ticketTitleCentresViaGrid()` **fails** (previously 13/13 green with it gone).
- Restored → **13/13 pass** again.

So the "dangling `var()` → `grid-template-columns: none` / button `auto`" state you described is now caught, not silently green. Updated the test's doc comment to enumerate all three pins.

**Y2 (stale `:688` pointer):** replaced the line number with the selector — `a bump to match the sibling `.sidebar-right > button` (28px)`. You're right it never resolved (688 is a `*/`, decl at 699, and off-by-N pre-change too); naming the selector survives edits above it. Also fixed the misleading wording per your Green — it now reads "22px is below the WCAG 2.2 §2.5.8 24px target-size minimum" (not "at/below"), with the bump explicitly deferred.

**On the Greens I didn't change:**
- **Green (token size still 22px < 24px):** left deferred, and said so in the comment. The ticket scopes this to centering with box size "as-is", and it's a repo-wide gap (the whole #917→#923 chain sits under 24px). Fair point that the card being a click target undercuts the spacing exception — but that argument applies chain-wide, so it belongs in one deliberate a11y pass, not smuggled into this title-centering PR. The `--tk-refresh` var makes it a one-token change (plus updating the pin's `22px`→`24px`) whenever that's taken up.
- **Green (token declared on `.tickets-card`, a level above its uses):** left as-is — you called it harmless/stylistic, and moving it onto `.tickets-head` would prepend it to that rule, forcing the whole-rule pin to a longer, more brittle string for no behavioral gain. Happy to move it if you'd prefer the co-location.
- **Green (`stripComments` truncates at `//`):** noted; neither new assertion targets the `app.css:101` SVG-data-URL region, so no impact.

`swift test --filter WebTerminalAssetTests` → 13/13 pass locally.

dhilgaertner (changes requested):
## Code & Security Review

CSS-only layout change plus a drift-guard test. The grid arithmetic is right: `22px | 4px gap | 1fr | 4px gap | 22px` is symmetric, so the `1fr` track — and the text centred inside it — lands centred on the card's content box (padding is a uniform `8px`). Both title and refresh carry explicit `grid-column`, so nothing auto-places into the spacer. `--tk-refresh` is declared on `.tickets-card` and inherits to `.tickets-head`/`.tickets-refresh`, and `.tickets-refresh` has exactly one call site (`app.js:1140`, inside `ticketsCard()`), so there is no scope where the `var()` dangles. Collapsing the two literal `22px`s into one token is the right call.

Full `CrowDaemon` suite: **327 tests / 54 suites, all passing.**

### Critical Issues
None.

### Security Review
**Strengths:**
- Purely presentational — no new DOM, no data flow, no network or auth surface.
- The title is a static literal (`el('span', 'tickets-title', 'Tickets')`), so the new `overflow`/`text-overflow` path can't be driven by ticket data.
- No secrets, no new dependencies, no config or permission changes.

**Concerns:** none found.

### Code Quality

**🟡 The two declarations that actually centre the title are unpinned — deleting them keeps every suite green**

`app.css:672` — `.tickets-title { grid-column: 2; text-align: center; … }` is not asserted by `ticketTitleCentresViaGrid` or anything else. The test pins the head's grid rule, the button's var-driven box, and the token definition, but not the title's own placement/alignment — which is the invariant the PR exists to establish.

Verified against this branch, restoring the tree after each run:

| Mutation to `.tickets-title` | Suite result | Render |
|---|---|---|
| drop `text-align: center` | ✅ 13/13 green | title left-aligned in the `1fr` track — the exact pre-CROW-924 look |
| drop both `grid-column: 2` and `text-align: center` | ✅ green | title (first child) auto-places into the **22px spacer track** and ellipsises to nothing at the left edge |

The second is a harder break than the `space-between` revert the test *does* catch, and it's the same "standing, untested trap" class the test's own doc-comment (`WebTerminalAssetTests.swift:439-442`) claims to be closing. One line closes it:

```swift
#expect(
    css.contains(".tickets-title { grid-column: 2; text-align: center;"),
    "the title must sit in the centre track and centre its text there, or it auto-places into the 22px spacer (CROW-924)")
```

**🟢 The test pins `--tk-refresh: 22px;` by value, which contradicts the token's own "one-token change" comment**

`app.css:657-658` says a WCAG 2.5.8 bump to 28px "is a one-token change here", but `WebTerminalAssetTests.swift:462` asserts the literal `22px`, making it a two-place change. The failure would be loud rather than silent, so this is fine as-is — but if the intent is to pin *existence* rather than *value*, `css.contains("--tk-refresh:")` matches only the declaration (`stripComments` removes the prose, and `var(--tk-refresh)` has no colon), which would let the deferred a11y bump land in one place as documented.

**🟢 The title's ellipsis threshold tightens by 26px**

The spacer track plus its gap take 26px off the title's available width versus the old `space-between` row. For the static `"Tickets"` in a full-width sidebar card this won't bite in practice, and the doc comment already reasons about the raised-min-font-size case — noted only for completeness.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The layout change itself is correct and I'd take it as-is; the block is solely the one-line assertion so the centring can't silently revert, which is worth landing in this round trip rather than a follow-up.

---

dgershman:
Thanks — mutation-checked again, fixed in 13aa1195.

**Yellow (`.tickets-title` placement/alignment unpinned):** added your suggested pin, anchored on the selector:
```swift
#expect(
    css.contains(".tickets-title { grid-column: 2; text-align: center;"),
    "the title must sit in the centre track and centre its text there, or it auto-places into the 22px spacer (CROW-924)")
```
Re-ran both mutations from your table against this branch:
- drop `text-align: center` → `ticketTitleCentresViaGrid()` **fails** (was green).
- drop `grid-column: 2` **and** `text-align: center` → **fails** (was green).
Restored → 13/13 pass. So the title's own centre-track placement is now guarded, not just the surrounding grid.

**Green (value pin contradicts the "one-token change" comment) — adopted:** switched the token assertion from `--tk-refresh: 22px;` to `css.contains("--tk-refresh:")`. That matches only the declaration (the `var(--tk-refresh)` usages have no colon, and `stripComments` drops the prose), so it still fails if the token is deleted — but it's value-agnostic, so the deferred WCAG 2.5.8 bump (22px→28px) is a genuine single-place edit as the comment promises, not a two-place change that also touches this test. Good catch — that was a self-inflicted contradiction.

**Green (ellipsis threshold −26px):** agreed it won't bite for the static "Tickets" in a full-width card; no change, and the doc comment already reasons about the raised-min-font case.

`swift test --filter WebTerminalAssetTests` → 13/13 pass locally.

dhilgaertner (approved):
## Code & Security Review

CSS-only presentation change plus one asset test. No Swift, no DOM, no data flow, no new inputs — so the security surface here is genuinely nil (no auth path, no user-controlled string reaching the CSS, no new network/IO). Reviewed for layout correctness, comment accuracy, and whether the new test actually holds the invariant.

### Critical Issues
None.

### Verification performed

**Geometry.** `grid-template-columns: 22px 1fr 22px` with `gap: 4px` puts `22+4` on the left of the `1fr` track and `4+22` on the right, and `.tickets-card` has symmetric `padding: 8px` — so the centre track is genuinely centred on the card's content box, and `text-align: center` centres the glyphs inside it. The PR body's claim ("centers relative to the **whole box** — not just the space left of ↻") checks out.

**Token resolution.** `--tk-refresh` is declared on `.tickets-card:659`; both consumers (`.tickets-head:671`, `.tickets-refresh:676`) are descendants, so inheritance resolves. `.tickets-refresh` is constructed in exactly one place (`app.js:1140`, inside `ticketsCard()`), so there is no orphan usage outside the declaring subtree where the `var()` would fall back to `auto`.

**Auto-placement.** `.tickets-refresh` carries an explicit `grid-column: 3` rather than relying on placement order — correct, since without it the two children would land in cols 1–2 and the title would be squeezed into the 22px spacer. The PR body calls this out and the code matches.

**Ellipsis still works.** `min-width: 0` is retained, which matters *more* under grid than under the old flex row (grid items also get an automatic minimum size). With `justify-self: stretch` the title fills the `1fr` track, so `overflow: hidden` + `text-overflow: ellipsis` + `white-space: nowrap` behave as before.

**Headroom for the raised-min-font-size case the comments keep citing.** The `1fr` track resolves to ~222px on the 350px desktop sidebar (350 − 20 sidebar padding − 4 `.sidebar-top` margin − 36 right column+gap − 16 card padding − 52 for the two spacer/button tracks + gaps). "Tickets" at 13px/700 is ~50px, so the 52px the grid takes back from the title versus the old `space-between` row is not close to a practical regression — it would take a ~40px minimum font before the title ellipsised at all, and the ↻ is in-flow so it still can't overlap.

**Spec accuracy of the new comments.** Both non-obvious claims hold: an invalid-at-computed-value-time `var()` resolves to the property's unset value, and both affected properties are non-inherited — `grid-template-columns` → initial `none`, `width`/`height` → initial `auto`. So the doc comment's stated failure mode for a dangling token is right, not hand-waving.

**Tests.** `swift test --package-path Packages/CrowDaemon` → **327 tests in 54 suites passed**. No pre-existing assertion regressed: the only Swift references to `.tickets-head` / `.tickets-title` anywhere in the repo are the new ones, and the neighbouring CROW-917/922 assertions are selector-anchored so the removed `flex: 0 0 auto` / literal `22px` don't touch them.

**Mutation-checked the new test rather than trusting it green:**

| Mutation | Result |
|---|---|
| delete `--tk-refresh: 22px;` from `.tickets-card` | ✘ fails at `:474` (token-definition assertion) |
| drop `grid-column: 2; text-align: center;` from `.tickets-title` | ✘ fails at `:463` (placement assertion) |

Both caught. The test earns its keep — and pinning the *definition* separately from its two consumers is the right call, because that's the one deletion the consumer assertions would sail straight past.

### Security Review

**Strengths:**
- Zero attack surface touched: no Swift, no JS, no request handling, no serialization. `app.css` is a static bundled asset served through the existing `webAsset` path; nothing in the diff is templated or interpolated from state.
- No secrets, no external references (no `url()`, no `@import`, no remote font) — nothing that could turn a stylesheet into an exfiltration or CSP-relaxation vector.
- The title is a static `'Tickets'` literal in `app.js:1136`; the ellipsis/`overflow: hidden` handling is not doing double duty as sanitization of anything user-controlled.

**Concerns:**
- None.

### Code Quality

Genuinely good, and the durability argument is the best part of the change. Replacing the two literal `22px`s with one token removes a real latent trap: previously a target-size bump to the button would have silently de-centred the title with the whole suite still green. The token is scoped to the component root rather than dumped in `:root`, which is the right altitude.

Three notes, all non-blocking:

**1. `--tk-refresh:` is unanchored to a selector** (`WebTerminalAssetTests.swift:475`). `css.contains("--tk-refresh:")` proves the token is *declared somewhere*, not that it's declared on an ancestor of its two consumers. Moving the declaration to a non-ancestor rule (`.tk-count`, say) keeps this assertion green while both `var()`s fall back to `none`/`auto` — the exact failure the surrounding comment is written to prevent. Contrived as a mutation, and I'd rather have this assertion than not; if you want to close it, anchoring on `.tickets-card {` … `--tk-refresh:` (or just asserting the declaration sits between `.tickets-card {` and its closing brace) costs nothing. *Consider.*

**2. The stripComments rationale is overstated** (`WebTerminalAssetTests.swift:447-449`). It says a raw-file `contains` "would false-pass off the prose once a declaration is deleted" — but none of the four needles actually appear in the `app.css` comments today. The three declaration needles are full rule text no prose matches, and the prose mentions `--tk-refresh` only *without* a trailing colon (lines 666, 669), which the assertion's own later comment correctly relies on. Using `stripComments` is still the right defensive default and matches the sibling tests; it's just the stated justification that doesn't hold. Worth trimming to "defensive, consistent with the other asset assertions" so a future reader doesn't go hunting for prose that would false-pass. *Consider.*

**3. WCAG 2.2 §2.5.8 (22px < 24px)** — already disclosed in the new comment at `app.css:656-658`, correctly scoped as a pre-existing condition across the whole #917→#923 chain rather than something this PR introduces, and the token now makes the eventual bump a one-line edit. Flagging only so it stays on the record for that pass; no action here. *Consider.*

### Summary Table

Scope is tight, the geometry is right, the comments are accurate about their own failure modes, the DOM is untouched as claimed, the full suite passes, and the new test survives the two mutations that matter. All three notes are optional polish on comment prose and test anchoring, none of which affects rendered behaviour.

---

PR #927, opened 2026-07-30
Delete dead code left over from ADR 0010 (retired macOS app)

### LINKED ISSUE #926: Delete dead code left over from ADR 0010 (retired macOS app): orphaned CrowUI + Sources/Crow, and vestigial AppKit terminal views
ADR 0010 (`docs/adr/0010-retire-the-macos-app.md`) retired the native macOS app and removed its target, but several source files were left on disk. They're now dead. Clean them up. This finishes the deletion ADR 0010 intended (don't edit the ADR itself).

## Bucket 1 — fully orphaned, safe to delete outright
Verified unreferenced (grep: `import CrowUI` appears nowhere; the root `Package.swift` declares only the `crow`/`CrowCLI` and `crowd` targets — there is no `Crow` target, so `Sources/Crow/*` compiles into nothing):
- **`Packages/CrowUI/`** (whole dir) — no `Package.swift`, referenced by no manifest, imported nowhere. Files: `Sources/CrowUI/ScorecardView.swift`, `Sources/CrowUI/AnalyticsFormatting.swift` (~669 lines). Superseded by the web `renderScorecard` + the Core scorecard model.
- **`Sources/Crow/App/JobRPCSupport.swift`** — its logic already lives in CrowEngine (covered by `CrowEngineTests/JobRPCSupportTests.swift`).
- **`Sources/Crow/Generated/BuildInfo.swift`** — orphaned generated stub. ⚠️ Before deleting, confirm the **daemon's** version/build-info string comes from a *different*, live source (grep for `BuildInfo`), so removing this doesn't break `crowd`'s version reporting. If a live build-info generator writes here, leave it and just remove the rest.

## Bucket 2 — compiled-but-dead AppKit views (audit, then remove)
CrowTerminal still contains AppKit/WKWebView view surfaces that compile into the library on macOS but are **never instantiated** — the daemon renders terminals in-browser via xterm.js over the `/terminal` WebSocket; these compile out entirely on Linux. Remove the dead view code:
- `Packages/CrowTerminal/Sources/CrowTerminal/TerminalSurfaceView.swift`
- `Packages/CrowTerminal/Sources/CrowTerminal/TerminalSearchBar.swift`
- `Packages/CrowTerminal/Sources/CrowTerminal/XTermSurfaceView.swift`

⚠️ **`TmuxBackend.swift` is mixed** — it imports AppKit **but also holds the live tmux/PTY backend the daemon depends on.** Do **not** delete it. Audit it: strip only dead `#if canImport(AppKit)` view code (if any), and preserve the live tmux controller / PTY logic. If the AppKit import is only for a dead view, drop the import; if it's load-bearing, leave it.

## Verify
- Grep to confirm zero references before deleting each file/dir.
- `swift build` for both `crow` and `crowd` passes on macOS.
- Full Swift test suite passes.
- The **Linux CI lane** (`.github/workflows/ci.yml`) stays green.
- `crowd`'s version/build-info string still resolves (guards the BuildInfo deletion).

## Scope
Pure deletion + a small TmuxBackend audit; no behavior change. Bucket 1 is unconditional; bucket 2 is remove-after-confirming-no-references.

dgershman:
Closes #926

Finishes the ADR 0010 macOS-app retirement by deleting source files the target removal left orphaned on disk. **Pure deletion — no behavior change**, plus a small TmuxBackend audit and a comment scrub.

## Bucket 1 — fully orphaned (wired to no SwiftPM target)
- **`Packages/CrowUI/`** (`ScorecardView` + `AnalyticsFormatting`, ~669 lines) — `import CrowUI` appears nowhere; no manifest declares it. Superseded by the web `renderScorecard` + Core `ScorecardDTO`.
- **`Sources/Crow/App/JobRPCSupport.swift`** + the whole **`Tests/CrowTests/`** tree — the retired monolith's source + tests. Root `Package.swift` declares only the `crow`/`crowd` products and **no test target**, so both compile and run on **no lane** (verified via manifest grep + `git ls-files`). `JobRPCSupport`'s logic lives in CrowEngine (`CrowEngineTests/JobRPCSupportTests.swift`); the other `Tests/CrowTests/*` files are the monolith's superseded suites.
- **`Sources/Crow/Generated/BuildInfo.swift`** — **never existed on disk**; it's a build-time stub `scripts/generate-build-info.sh` no longer emits. Guard confirmed: `crowd`'s version resolves from `Resources/web/version.json` and `crow`'s from generated `CLIVersion.swift` — neither is `BuildInfo` — so there was nothing to delete and version reporting is unaffected.

## Bucket 2 — dead AppKit/WKWebView terminal views
Compiled into CrowTerminal on macOS but **never instantiated** — the daemon renders terminals in-browser over the `/terminal` WebSocket.
- Deleted `TerminalSurfaceView.swift`, `TerminalSearchBar.swift`, `XTermSurfaceView.swift`.
- **`TmuxBackend.swift` audit (kept — it holds the live tmux/PTY backend):** stripped only the dead cockpit-surface members — `import AppKit`, `sharedSurface`, `cockpitSurface()`, `existingCockpitSurface`, `onCockpitExit`, and the `shutdown()` destroy-block — and gutted `recycleCockpitSurface()` (still called by `SessionService`) down to its one live effect on the headless daemon: resetting the active-window marker. The tmux controller / PTY / window-binding logic is untouched. TmuxBackend now has **zero** AppKit-gated code, so it's strictly more Linux-safe than before (it previously referenced the AppKit-only `sharedSurface` *unguarded* inside `recycleCockpitSurface`).
- Dropped the now-unused `WebKit` linker framework from `CrowTerminal/Package.swift` (only `XTermSurfaceView` imported WebKit).

## Comment scrub
Removed live-`Sources/` comments that named the deleted symbols/packages (`CrowUI`, `ScorecardView`, `TerminalSurfaceView`, `XTermSurfaceView`) so no live-code comment references removed code (CrowCore, CrowDaemon, CrowEngine, CrowTerminal). Comment-only; historical mentions in ADRs are left as-is.

## Test plan
- ✅ `swift build` — `crow` **and** `crowd` build on macOS.
- ✅ Test suites pass: **CrowTerminal (106)** — real tmux 3.6a integration incl. `recycleCockpitSurfaceIsSafeWithoutSurface` — **CrowEngine (586)**, **CrowCore (527)**, **CrowDaemon (327)**.
- ✅ `crow --version` → `0.1.0`; `crowd`'s `version.json` resolves.
- ✅ Linux CI lane compiles none of the deleted GUI files (`ci.yml` excludes CrowUI/CrowTerminal) and sees only comment-only changes to its packages; the net effect on transitively-compiled TmuxBackend is removal of AppKit-gated code.


dhilgaertner (changes requested):
## Code & Security Review

Verified the deletion claims independently rather than taking the PR body at its word. Most of them hold up well:

- **Root manifest** declares only `crow`/`crowd`, both with explicit `path:` — so `Sources/Crow/` and `Tests/CrowTests/` genuinely compiled on no lane. `Sources/` now contains only `CrowCLI/main.swift` + `crowd/main.swift`.
- **`Packages/CrowUI/`** is fully gone and no manifest ever referenced it.
- **`BuildInfo.swift`**: `scripts/generate-build-info.sh` emits only `CLIVersion.swift` + `web/version.json`. The claim that it no longer emits `BuildInfo` is accurate.
- **WebKit unlink is safe** — no `import WebKit` remains anywhere in CrowTerminal sources.
- **`recycleCockpitSurface()` gutting is correct.** `sharedSurface` was only ever created by `cockpitSurface()`, which nothing called, so it was permanently `nil`; the one live effect (`activeTerminalID = nil`) is preserved and `recycleCockpitSurfaceIsSafeWithoutSurface` still passes. Same reasoning clears the `shutdown()` destroy-block removal.
- **`shouldAutoPermission` no longer exists in live code** (inlined at `SessionService.swift:1247-1248`), so `SessionServiceAutoPermissionTests` really was superseded.
- **`SessionService`'s same-value reassignment comment rewrite is accurate** — `AppState` is `@Observable`, and `withMutation` fires regardless of value equality, so that trick is not dead code.
- **Reproduced locally:** `swift build` succeeds (`crow` + `crowd`); CrowTerminal — **106 tests, all pass**.

TmuxBackend having zero AppKit-gated code is a real net improvement, and the audit-rather-than-delete call on that file was the right one.

### Code Quality

**Yellow — `TmuxBackend.swift:1463`: `shellQuote` is now dead code, in the one file this PR audited**

`private func shellQuote(_:)` has exactly **one** occurrence package-wide — its own definition:

```
$ rg -n 'shellQuote' Packages/CrowTerminal/
Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:1463:    private func shellQuote(_ s: String) -> String {
```

Its only caller was the deleted `cockpitSurface()`, which built the `tmux … attach-session` shell string. Swift doesn't warn on unused `private` methods, so the build stays green and this slipped through. In a PR whose entire purpose is deleting dead code — and which explicitly audited this file member-by-member — the orphaned helper should go with the rest.

**Yellow — `.github/workflows/release.yml:29-34`: the comment now describes deleted files, and the step it guards fails**

```yaml
# Root CrowTests (Tests/CrowTests) exercise root-target business logic
# (IssueTracker, Job, Scaffolder, SessionService, …). The root Crow
# target imports CrowTerminal (AppKit), so it can't run in the Linux PR
# lane — this macOS job is where those tests run. See docs/adr/0007.
echo "==> Testing root package (CrowTests)..."
swift test
```

`Tests/` no longer exists after this PR, and root `swift test` exits **1**:

```
$ swift test; echo $?
error: no tests found; create a target in the 'Tests' directory
1
```

To be fair on attribution: the root manifest is unchanged by this PR (`git diff origin/main...HEAD -- Package.swift` is empty), so this failure is **pre-existing** — the release `test` job has been failing at that step since the root test target went away. But this PR is the one that removes the last reason to keep the step, and it leaves behind a CI comment asserting that meaningful root-target coverage runs on tag builds when it neither runs nor exists. That is precisely the misleading-reference class the PR set out to eliminate; please drop the step and the comment. (`ci.yml:16` and `ci.yml:55` also still list `CrowUI` among the macOS-only packages to avoid.)

**Yellow — `CustomInstructionsSkillTests` was a live-drift guard, not a superseded suite**

The PR body states "the other `Tests/CrowTests/*` files are the monolith's superseded suites." That holds for most of them, but not this one. It was a **markdown drift gate** for #683: it read `skills/crow-workspace/SKILL.md` (+ its bundled `.template`) and asserted the skill still tells the Manager to resolve the *matched* workspace's `customInstructions` and emit a verbatim `## Custom Instructions` block without falling back to `defaults`.

That instruction is still live in `skills/crow-workspace/SKILL.md` (lines ~312-318 and ~472-481), and nothing in the package suites asserts it — `LaunchScaffoldTests` only checks the skill file gets scaffolded and refreshed, never its content. No *live* coverage is lost (the test was dead), but the guard's intent is dropped silently, and #683 can now regress in the skill markdown unnoticed. The `parity` CI job already runs source-level text gates (`check-cli-parity.sh`, `check-notification-events.sh`), so porting this one there is cheap. Failing that, please call it out in the PR body as deliberately dropped rather than superseded.

**Green — `BundledResources.terminalHTMLURL` is now dead public API.** Its only remaining callers are its own two cases in `BundledResourcesTests`; the doc comment added in this PR says as much ("No Swift code reads this any more"). Consider deleting the accessor and those cases. (`Resources/xterm/` itself must stay — `StaticAssets` pulls the xterm *library* files from it.)

**Green — stale references just outside the scrub filter.** The scrub was scoped to Swift under `Sources/`, which leaves these naming deleted symbols: `CrowTerminal/Sources/CrowTerminal/Resources/xterm/terminal.html:6-8` (`XTermSurfaceView`, `TerminalSurfaceView`, `TmuxBackend.cockpitSurface()`), `CrowDaemon/.../web/app.js:2786,2791` (`ScorecardView`, CrowUI's `AnalyticsFormatting`), `web/app.css:1` (`CrowUI/CorveilTheme.swift`), `CrowCoreTests/ScorecardDTOTests.swift:7` (desktop `ScorecardView`). Same class of reference, one filter away.

**Green — #588 detection is now single-signal.** With `onCockpitExit` gone, tmux-server-death detection rests entirely on the lazy `onServerLost`, which only fires on the next tmux command. Pre-existing in practice (the surface was never instantiated on the headless daemon), but the eager signal is now formally gone — worth a follow-up ticket rather than a change here.

### Security Review

**Strengths:**
- Pure deletion: no authn/authz, input-validation, secret-handling, or crypto surface is touched.
- **Attack surface shrinks.** The `WebKit` framework is unlinked and the `WKWebView` surface that loaded a local host page over `file://` is gone.
- **No injection path widened.** `shellQuote`'s only caller assembled a shell command *string*; every remaining tmux invocation passes an argv array with no shell, so quoting was not silently dropped from a live call site.

**Concerns:** none.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 3 Yellow, 3 Green] findings.

The core deletion is sound, well-reasoned, and verified green on build + tests. Requesting changes only so the three Yellows land in this round trip: they are all small and all in-scope for a dead-code PR — one orphaned helper in the audited file, one CI step whose comment now advertises coverage that neither runs nor exists, and one drift guard that needs either a port or an honest "dropped" note.

dgershman:
Thanks for the thorough review — all three Yellows addressed, plus both Greens, in `8171de8c`.

**Yellow — `shellQuote` dead code:** deleted. Confirmed `rg 'shellQuote' Packages/CrowTerminal/` now returns nothing.

**Yellow — `release.yml` root `swift test` step:** dropped the step and its stale comment. The per-package loop above already runs every real suite. Also removed `CrowUI` from the two `ci.yml` macOS-only lists (16, 55).

**Yellow — `CustomInstructionsSkillTests` was a live #683 guard:** ported it rather than dropping it — new `scripts/check-workspace-custom-instructions.sh` (a source-level text gate over `skills/crow-{workspace,batch-workspace}/SKILL.md` **and** their `Resources/*.template` scaffold copies), wired into the `parity` job next to the other text gates. Runs green and `shellcheck scripts/*.sh` is clean.

**Green — dead `terminalHTMLURL`:** deleted the accessor and its two tests. The live modified-Enter guard is `WebTerminalAssetTests.modifiedEnterIsDistinguishableFromPlainEnter` (verified present), and the jumpbottom addon file stays pinned by `jumpToBottomAddonIsBundled` (via `xtermDirectoryURL`). `Resources/xterm/` is untouched.

**Green — stale references outside the Swift-`Sources/` scrub:** scrubbed `terminal.html`, `web/app.js`, `web/app.css`, `ScorecardDTOTests.swift`, and the `WebTerminalAssetTests` doc comment that named the deleted test.

**Green — #588 single-signal detection:** left as-is per your note — worth a follow-up ticket, not a change here. Happy to file one.

Re-verified on macOS: `swift build` (crow + crowd); CrowTerminal (104 — the 2 retired-page tests removed), CrowDaemon (327), CrowCore (527) all pass.

dhilgaertner (approved):
## Code & Security Review

Re-review of `8171de8c`. I re-verified each of the five items from the previous round independently rather than taking the response comment at its word — **all five hold up**, and the port went further than the minimum I asked for.

### Previous findings — verified fixed

- **Yellow `shellQuote`** — gone. `rg 'shellQuote' Packages/CrowTerminal/` returns nothing; the remaining repo-wide hits are the live per-harness `*LaunchArgs.shellQuote` helpers and `CrowDaemon/TerminalCockpit.swift`'s own private copy (which is the *live* cockpit attach path that superseded the deleted `cockpitSurface()`). Nothing was over-deleted.
- **Yellow `release.yml`** — the root `swift test` step and its stale comment are dropped. The surviving `for pkg in Packages/*/` loop still covers every package that has a `Tests/` dir, so no real suite lost a lane. Confirmed the root manifest now declares only `CrowCLI` + `crowd` with explicit `path:`, `Sources/` holds exactly `CrowCLI/main.swift` + `crowd/main.swift`, and `Tests/` is gone — so there was genuinely nothing left for that step to run. `CrowUI` is also out of both `ci.yml` macOS-only lists (16, 55).
- **Yellow `CustomInstructionsSkillTests`** — ported, not dropped. `scripts/check-workspace-custom-instructions.sh` is a faithful 1:1 translation: same four files (both `SKILL.md`s **and** both `Resources/*.template` scaffold copies), same six needles for `/crow-workspace` and same three for `/crow-batch-workspace`, matching the deleted `#expect`s exactly.
  - **I confirmed the gate is not vacuous.** Against a scratch copy with drift injected two different ways — rewriting `do not** read \`defaults\`` in `skills/crow-workspace/SKILL.md`, and renaming the `## Custom Instructions` heading in `Resources/crow-batch-workspace-SKILL.md.template` — it exits 1 and names both offending file/needle pairs. It passes clean on HEAD.
  - Net coverage is a **strict improvement**: the original test lived in `Tests/CrowTests`, which no target compiled, so it ran on *no* lane. The gate now runs in `ci.yml`'s `parity` job on every PR into `main`/`feature/**`.
  - Script itself is sound bash: `set -euo pipefail`, `grep -Fq --` (fixed-string, and `--` guards a needle that could start with `-`), `fail` accumulates across all needles rather than short-circuiting so one run reports every drift, and the two `# shellcheck disable=SC2016` directives are correctly placed and genuinely needed (the backtick-bearing needles are the only SC2016 triggers). I could not run `shellcheck` — not installed on this host — but I found nothing it should flag.
- **Green `terminalHTMLURL`** — accessor and both tests deleted. Verified the replacement guards are real and still present: `WebTerminalAssetTests.modifiedEnterIsDistinguishableFromPlainEnter` is parameterized over the two *live* assets, and `jumpToBottomAddonIsBundled` still pins the addon file via `xtermDirectoryURL`. `Resources/xterm/` stays, correctly — `StaticAssets` serves the xterm library files out of it.
- **Green stale references** — clean. `rg` for `CrowUI|ScorecardView|AnalyticsFormatting|XTermSurfaceView|TerminalSurfaceView|cockpitSurface` across all non-doc files now returns **zero** hits outside ADRs.

### Verification I ran

- **`swift build` passes** (`crow` + `crowd`, exit 0, verified on the real exit code rather than through a pipe).
- **Comment-scrub claims are true.** I read every hunk in `CrowCore`, `CrowDaemon/Sources`, and `CrowEngine`: all of them touch only `///`, `//`, `/* */`, or HTML comment text. **No executable line changed outside `TmuxBackend.swift`** — worth stating explicitly, since "comment-only" is exactly the claim a scrub PR is easiest to hide a real edit behind.
- **No second orphan of the `shellQuote` class.** After the cockpit removal I re-checked every symbol the deleted code touched — `ensureRunningServer`, `CockpitSessionStarter`, `tmuxBinary`, `sessionName`, `onServerLost`, `activeTerminalID`, `recycleCockpitSurface` — and all still have live references. `CockpitSessionStarter` in particular is live: parameter type at `TmuxBackend.swift:1166`, conformed by `TmuxController`, exercised by `EnsureCockpitSessionTests`.

**On the test suites — one correction to the PR body's test plan, which is *not* a fault of this change.** CrowDaemon (327), CrowCore (527), and CrowEngine (586) all pass for me. **CrowTerminal does not fully pass on my host**: `retryReadinessEmitsTimedOutWhenSentinelMissing` fails deterministically (`TmuxBackendTests.swift:393-394` — it observes `.shellReady` where it expects `.timedOut`). I checked before attributing it: on a clean `origin/main` tree the **same single test fails the same way** (main: 106 tests / 2 issues; this branch: 104 tests / 2 issues — the delta is exactly the two intentionally-deleted `terminalHTMLURL` tests). It is a pre-existing, host-timing-dependent flake — the 150ms budget at line 387 assumes the idle shell won't fire another `precmd`, which doesn't hold on a warm tmux server; it passed on my very first cold-server run and failed every run after. **Nothing to fix in this PR**, but the test plan's "CrowTerminal (104) pass" isn't reproducible on all hosts, and that flake is worth its own ticket.

### Code Quality

**Green — the retired `CrowTerminal/Resources/xterm/terminal.html` is now fully dead, and its banner slightly overstates its unreachability.**

With `terminalHTMLURL` deleted, no Swift code references this page at all; it survives only because `Package.swift` does `.copy("Resources/xterm")` on the whole directory. Its header says *"RETIRED SURFACE — NOT SERVED BY `crowd`"* and *"StaticAssets serves the xterm library files out of this directory (`/xterm/:file`) but never this page"* — but `StaticAssets.swift:38-48` is a generic basename lookup over that directory, so `GET /xterm/terminal.html` does serve it. No security impact: it's behind `WebAuthMiddleware` (`isAuthExempt` covers only `/login`, `/logout`, `/health`, `/brand.svg`), and the page is inert in a browser anyway since its transport is `webkit.messageHandlers`.

Deleting the file outright would be the logical completion of this PR and would moot the wording. The one real argument for keeping it is that `app.js:4089`, `app.js:4549`, and `xterm-addon-crow-jumpbottom.js` cite it as the mirror/reference source — so this is a judgment call, not a defect. Either delete it and re-point those three comments, or soften the banner to "not served as a UI surface." Not worth blocking on.

### Security Review

**Strengths:**
- Still a pure deletion — no authn/authz, input-validation, secret-handling, or crypto surface is touched. The only added executable code is a CI text-gate shell script that reads four in-repo markdown files and writes nothing.
- **Attack surface stayed shrunk**: `WebKit` remains unlinked and the `WKWebView`/`file://` surface remains gone.
- **The `shellQuote` deletion is safe, and I checked this specifically** — deleting a shell-quoting helper is the one way a dead-code PR could quietly widen an injection path. Its only caller was the deleted `cockpitSurface()`, and the live cockpit path (`CrowDaemon/TerminalCockpit.swift:99-100`) has its **own** `shellQuote` and still applies it to the tmux binary, socket path, and session group. Every other remaining tmux invocation passes an argv array with no shell. No call site lost quoting.
- New script is injection-free: needles are fixed strings via `grep -F`, no `eval`, no command substitution on file content, no unquoted expansions.

**Concerns:** none.

### Summary Table

Every Yellow from the last round is genuinely resolved, and the `CustomInstructionsSkillTests` port is better than what I asked for — it covers the batch skill and both scaffold templates, and it now runs on a lane the original never did. The remaining Green is optional and the CrowTerminal test failure is pre-existing on `main`, so neither blocks. Nice cleanup: −3792 lines with the one genuinely-live guard preserved rather than dropped.

PR #929, opened 2026-07-30
CI: capture Swift test coverage and upload it as a build artifact (CROW-928)

### LINKED ISSUE #928: CI: capture Swift test coverage and upload it as a build artifact
## Problem

Nothing in the pipeline captures test coverage. There is no `--enable-code-coverage`
anywhere in the `Makefile` or `.github/workflows/`, no codecov/coveralls config, no
badge, and no artifact. Every occurrence of "coverage" in the repo today is the English
word in prose or a comment.

That means we have no idea which code is untested, and no way to tell whether a PR made
things better or worse. Baseline numbers have to be generated by hand each time
(a full local sweep takes ~30 minutes), so in practice nobody does it.

## Baseline (measured 2026-07-30 @ `1441c56`, all 17 suites passing)

Whole repo: **62.7% line / 58.3% region** — 40,502 executable lines, 15,094 missed.

| Package | Line % | Missed |
|---|---:|---:|
| CrowEngine | 42.9% | 7,477 |
| CrowCLI | 59.3% | 988 |
| CrowPersistence | 60.4% | 163 |
| CrowDaemon | 60.5% | 3,518 |
| CrowTelemetry | 63.9% | 537 |
| CrowTerminal | 68.3% | 599 |
| CrowGit | 72.8% | 67 |
| CrowAutostart | 73.9% | 109 |
| CrowOpenCode | 75.2% | 206 |
| CrowCore | 76.7% | 804 |
| CrowClaude | 81.1% | 215 |
| CrowGrok | 82.2% | 105 |
| CrowProvider | 82.3% | 532 |
| CrowCodex | 84.9% | 144 |
| CrowCursor | 85.3% | 110 |
| CrowAntigravity | 88.2% | 53 |
| CrowIPC | 88.7% | 53 |

## Scope

No new dependencies. `swift test --enable-code-coverage` already makes SwiftPM write
`.build/<arch>/debug/codecov/<Package>.json` (llvm-cov export JSON, with per-file
`summary.lines.{count,covered,percent}`). This ticket is collect + merge + upload.

1. **`scripts/coverage-summary.sh`** (or `.py`) — walk `Packages/*/.build/*/debug/codecov/*.json`,
   merge them, and emit two files:
   - `coverage-summary.json` — machine-readable, per-package and per-file
   - `coverage-summary.md` — the human table (same shape as the baseline above)
   Must pass `shellcheck` if written in bash — the `lint` job already gates `scripts/*.sh`.

2. **`ci.yml`** — add `--enable-code-coverage` to the existing `swift test` invocation in
   the "Build & test Linux packages" loop, then run the summary script and upload with
   `actions/upload-artifact@v4` as the **last step of the `test` job**.

3. **`release.yml`** — same treatment on the macOS `test` job, which runs all 17 packages.
   This is the only place the full picture is available (see caveat below).

4. **`make coverage`** — a local target that does the same thing, so the artifact and a
   developer's local run are the same numbers.

## Two gotchas that will bite whoever implements this

1. **Per-package JSON includes dependency sources.** A test binary compiles its
   dependencies, so CrowGit's raw JSON reports ~3,700 lines at 6% instead of its own 246
   at 72.8%. The merge must scope each package's number to files under that package's own
   `Sources/`, then compute the repo total by unioning across packages and taking the
   **max covered per file** (a CrowCore file exercised by CrowEngine's tests is covered,
   even though CrowGit's binary never touches it). Skipping this produces alarming numbers
   that mean nothing.

2. **Filter `*.derived/runner.swift`** out of the file list — SwiftPM's generated test
   runner otherwise lands in the totals.

## Caveat: the PR lane can only ever see 36% of the codebase

Per [ADR-0007](docs/adr/0007-linux-ci-swift.md), the PR lane is Linux and runs an explicit
12-package allowlist. The 5 Darwin-only packages are excluded — and that is where the
coverage is worst:

- **Linux lane (every PR):** 12 packages, 14,718 lines — **36%** of the codebase
- **Darwin-only (release tags only):** CrowEngine, CrowDaemon, CrowTerminal, CrowTelemetry,
  CrowAutostart — 25,784 lines — **64%**, including the two lowest-coverage packages

This is not a bug to fix here. ADR-0007 deliberately rejected per-PR macOS minutes, and
this ticket should not re-litigate that. The consequence to accept knowingly: the per-PR
artifact covers a third of the code, and CrowEngine/CrowDaemon coverage refreshes only at
release tags. If we later want fresher numbers for those, the cheap option is a weekly
scheduled macOS workflow — **out of scope here**, file separately.

While in ADR-0007: line 24 claims CrowTelemetry "has no test target." That is now stale —
it has 5 test files and 1,486 covered-able lines. Worth a one-line correction.

## Non-goals

- No third-party upload (codecov/coveralls) — artifact only.
- No PR gate, no ratchet, no failing the build on a coverage drop. Get measurement in
  first; enforcement is a separate decision once we have trend data.
- No committed `docs/coverage.md`. Artifacts only, per the request. Note the trade-off:
  artifacts expire and are not git-tracked, so there is no long-run trend history and
  consumers must fetch via `gh run download`.
- Do not write any new tests in this ticket.

## Acceptance criteria

- [ ] A PR build produces a downloadable artifact containing `coverage-summary.json` and
      `coverage-summary.md` for the 12 Linux packages.
- [ ] A release build produces the same artifact for all 17 packages.
- [ ] `make coverage` produces byte-comparable output locally.
- [ ] Per-package percentages match the baseline table above within ~1% at `1441c56`
      (this is the check that the dependency-scoping gotcha was handled correctly).
- [ ] No change to PR wall-clock beyond the coverage instrumentation overhead.
- [ ] `shellcheck` passes.

dgershman:
Closes #928

`swift test --enable-code-coverage` already makes SwiftPM drop an llvm-cov export at `Packages/<Pkg>/.build/<triple>/debug/codecov/<Pkg>.json`, so this is collect + merge + upload. No new dependencies, no third-party upload, no gate, no ratchet.

## What landed

- **`scripts/coverage-summary.sh`** — gathers whatever exports are on disk and merges them through **`scripts/coverage-summary.swift`** into `coverage-summary.json` (per-package + per-file) and `coverage-summary.md` (the human table).
- **`ci.yml`** — `--enable-code-coverage` on the Linux package loop, then summarize + `actions/upload-artifact@v4` (`coverage-summary-linux`) as the last step of `test`.
- **`release.yml`** — same on the macOS `test` job (`coverage-summary-macos`), the only lane where all 17 packages are measured.
- **`make coverage`** — same sweep, same script, locally.
- **ADR 0007** — corrected line 24: CrowTelemetry *does* have a test target.

## The two gotchas the ticket flagged

1. **Dependency scoping.** A test binary compiles its dependencies, so a package's raw export describes far more than that package — CrowGit's own JSON reports ~3,700 lines at 6%, which is CrowCore and friends dragged in by the link. Every file is attributed to the package that *owns* its source (`Packages/<Pkg>/Sources/…`) instead.
2. **`*.derived/runner.swift`** is filtered (it also fails the `Sources/` check, but the guard is explicit so the intent survives).

## Two numbers per package

The same file appears in several exports — covered where some suite exercised it, cold where another package merely linked it. Summing double-counts; averaging punishes a file for binaries that never called it. Files are folded by taking the **max covered**, which gives:

- **Own suite** — what that package's own tests reach. The actionable number, and what the table's `Line %` / `Missed` columns report.
- **Any suite** — also credits a file some dependent's tests exercised. This is what the repo total is computed from, which is why the total beats the `Missed` column's sum.

A package linked into a lane but never tested there reads **`—`**, not a fabricated `0%`. On Linux that is CrowAutostart, pulled in via CrowCLI:

| Package | Line % | Region % | Lines | Missed | Any suite % |
|---|---:|---:|---:|---:|---:|
| CrowCLI | 59.3% | 64.5% | 2,427 | 988 | 59.3% |
| … | | | | | |
| CrowAutostart | — | — | 418 | — | 5.7% |

## Why Swift and not Python/jq

The PR lane runs inside the `swift:6.1` container, which ships **neither** — its Python is `libpython3-dev`, headers and stdlib for LLDB, with no interpreter binary. Swift is the one language guaranteed present in both CI lanes and on any machine that can run `make coverage`, so this keeps the "no new dependencies" constraint honest. `coverage-summary.sh` compiles it with `swiftc` into a temp dir rather than relying on interpreter mode.

Output carries no timestamps, hostnames, or absolute paths, and the JSON is emitted by hand rather than through `JSONSerialization` (Darwin escapes `/` as `\/`, corelibs-foundation does not). Verified byte-identical across repeat runs.

## Verification against the baseline (`1441c56`, full local sweep, all 17 suites passing)

**15 of 17 packages match the ticket's baseline exactly.** The other three differ by 1–3 missed lines — CrowTelemetry (540 vs 537), CrowTerminal (601 vs 599), CrowIPC (54 vs 53) — all timing-sensitive suites, so this is test variance rather than a merge error. Every package is well inside the ~1% acceptance bar.

Total executable lines match exactly at **40,502**.

One honest gap: the repo total lands at **61.9%** vs the hand-measured **62.7%**. Per-file max is a *lower bound* on cross-package credit — if suite A covers lines 1–10 of a file and suite B covers 11–20, the true union is 20 and per-file max sees 10. Recovering the difference would mean `llvm-profdata merge` across packages and re-exporting against every test binary, which is a materially bigger job than this ticket scoped ("taking the max covered per file"). Flagging it rather than burying it; happy to file a follow-up if the exact number matters.

Similarly, the Linux artifact reports 15,136 lines rather than the ticket's 14,718 — the delta is exactly CrowAutostart's 418 lines, which the lane really does compile via CrowCLI. Showing it with `—` seemed more truthful than dropping it.

## Test plan

- [x] `shellcheck scripts/*.sh` passes
- [x] Full local sweep (17 packages) → per-package numbers match the baseline table
- [x] Repeat runs produce byte-identical `coverage-summary.{json,md}`
- [x] `coverage-summary.json` parses as valid JSON
- [x] Linux lane simulated against the 12 `LINUX_PACKAGES` exports — correct table, `—` row for CrowAutostart
- [x] `coverage/` is gitignored; nothing generated is committed
- [ ] CI green on this PR, with a downloadable `coverage-summary-linux` artifact

## Known trade-offs (per the ticket, not re-litigated here)

Per [ADR 0007](docs/adr/0007-linux-ci-swift.md) the PR lane is Linux, so the per-PR artifact covers ~36% of the tree; CrowEngine and CrowDaemon coverage refreshes only at release tags. Artifacts also expire and are not git-tracked, so there is no long-run trend history — consumers fetch with `gh run download -n coverage-summary-linux`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,8 +63,24 @@ jobs:
             # --no-parallel: the CrowIPC socket tests spin up real servers with
             # blocking I/O; running them in parallel starves each other on the
             # constrained CI runner and intermittently hits the read timeout.
-            swift test --no-parallel --package-path "Packages/$pkg"
+            # --enable-code-coverage leaves an llvm-cov export under
+            # Packages/$pkg/.build/<triple>/debug/codecov/ for the step below.
+            swift test --no-parallel --enable-code-coverage --package-path "Packages/$pkg"
           done
+
+      # Measurement only — no gate, no ratchet, no third-party upload (CROW-928).
+      # This lane can only ever see the 12 packages above: the 5 Darwin-only ones
+      # (CrowEngine, CrowDaemon, CrowTerminal, CrowTelemetry, CrowAutostart) are
+      # ~64% of the tree and refresh at release tags, per ADR 0007. The full
+      # picture lives in release.yml's macOS test job.
+      - name: Summarize coverage
+        run: bash scripts/coverage-summary.sh
+
+      - name: Upload coverage summary
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-summary-linux
+          path: coverage/
 
   # Source-level drift gates: cheap text checks that need no Swift build, so they
   # cover packages the Linux lane above cannot compile. `check-cli-parity.sh` is
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,28 +59,45 @@ jobs:
                           CrowAntigravity CrowGit CrowOpenCode CrowGrok CrowPersistence CrowProvider CrowCLI)
           for pkg in "${LINUX_PACKAGES[@]}"; do
             echo "==> Building & testing $pkg on Linux..."
-            swift build --package-path "Packages/$pkg"
+            # --enable-code-coverage on BOTH invocations, deliberately: toggling
+            # the flag invalidates SwiftPM's build plan, so an uninstrumented
+            # build here would be thrown away and recompiled from scratch by the
+            # instrumented `swift test` below. That is a full second compile of
+            # all 12 packages on a lane that keeps no compiled products between
+            # runs (ADR 0007) — sharing the flag makes the build reusable.
+            swift build --enable-code-coverage --package-path "Packages/$pkg"
             # --no-parallel: the CrowIPC socket tests spin up real servers with
             # blocking I/O; running them in parallel starves each other on the
             # constrained CI runner and intermittently hits the read timeout.
-            # --enable-code-coverage leaves an llvm-cov export under
-            # Packages/$pkg/.build/<triple>/debug/codecov/ for the step below.
+            # The export lands at Packages/$pkg/.build/<triple>/debug/codecov/.
             swift test --no-parallel --enable-code-coverage --package-path "Packages/$pkg"
           done
 
       # Measurement only — no gate, no ratchet, no third-party upload (CROW-928).
-      # This lane can only ever see the 12 packages above: the 5 Darwin-only ones
-      # (CrowEngine, CrowDaemon, CrowTerminal, CrowTelemetry, CrowAutostart) are
-      # ~64% of the tree and refresh at release tags, per ADR 0007. The full
-      # picture lives in release.yml's macOS test job.
+      # This lane measures only what LINUX_PACKAGES lists; the rest of the tree is
+      # ~64% of the lines and refreshes at release tags, per ADR 0007.
+      #
+      # "Not in the allow-list" is not the same as "cannot build here":
+      # CrowAutostart is #if canImport-guarded and CrowCLI links it, so it does
+      # compile on Linux and shows up in the report as a "—" row (linked, never
+      # tested in this lane) rather than as a measured number. Adding it to
+      # LINUX_PACKAGES would turn that into a real figure — deliberately left as
+      # a follow-up, since ADR 0007 wants allow-list changes made consciously.
       - name: Summarize coverage
-        run: bash scripts/coverage-summary.sh
+        if: always()
+        run: bash scripts/coverage-summary.sh --allow-empty
 
+      # always(): a red suite is exactly when the covered/uncovered split is worth
+      # having. The summary describes only the exports that exist, so a lane that
+      # died partway yields a smaller table, never a wrong one.
       - name: Upload coverage summary
+        if: always()
         uses: actions/upload-artifact@v4
         with:
           name: coverage-summary-linux
           path: coverage/
+          if-no-files-found: ignore
+          retention-days: 30
 
   # Source-level drift gates: cheap text checks that need no Swift build, so they
   # cover packages the Linux lane above cannot compile. `check-cli-parity.sh` is
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,9 +24,21 @@ jobs:
           for pkg in Packages/*/; do
             if [ -d "$pkg/Tests" ]; then
               echo "==> Testing $(basename "$pkg")..."
-              swift test --package-path "$pkg"
+              swift test --enable-code-coverage --package-path "$pkg"
             fi
           done
+
+      # The only lane that measures all 17 packages — the PR lane is Linux and
+      # cannot compile the 5 Darwin-only ones, which are ~64% of the tree and
+      # include the two worst-covered (ADR 0007, CROW-928).
+      - name: Summarize coverage
+        run: bash scripts/coverage-summary.sh
+
+      - name: Upload coverage summary
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-summary-macos
+          path: coverage/
 
   shellcheck:
     name: Shell Lint
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,17 +28,21 @@ jobs:
             fi
           done
 
-      # The only lane that measures all 17 packages — the PR lane is Linux and
-      # cannot compile the 5 Darwin-only ones, which are ~64% of the tree and
-      # include the two worst-covered (ADR 0007, CROW-928).
+      # The only lane that measures every package with a test target — the PR
+      # lane runs an explicit 12-package allow-list (ADR 0007), which leaves out
+      # ~64% of the lines, including the two worst-covered packages (CROW-928).
       - name: Summarize coverage
-        run: bash scripts/coverage-summary.sh
+        if: always()
+        run: bash scripts/coverage-summary.sh --allow-empty
 
       - name: Upload coverage summary
+        if: always()
         uses: actions/upload-artifact@v4
         with:
           name: coverage-summary-macos
           path: coverage/
+          if-no-files-found: ignore
+          retention-days: 30
 
   shellcheck:
     name: Shell Lint
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@ DerivedData/
 *.xcworkspace
 xcuserdata/
 
+# Coverage summaries (`make coverage`) — uploaded as a CI artifact, never committed
+coverage/
+
 # Generated sources
 Sources/Crow/Generated/
 Sources/CrowCLI/Generated/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,15 +42,15 @@ swift test        # Also works (runs root package tests)
 make coverage     # Full sweep with instrumentation (~30 min) → coverage/coverage-summary.{json,md}
 ```
 
-Tests use the Swift Testing framework (`@Test` macros). Currently, tests exist
-in `CrowCore`. When adding new functionality to any package, include a test
+Tests use the Swift Testing framework (`@Test` macros) and live under
+`Packages/*/Tests/`. When adding new functionality to any package, include a test
 target in that package's `Package.swift` and add tests.
 
-`make coverage` produces the same two files CI attaches to every run as the
-`coverage-summary-*` artifact (`gh run download -n coverage-summary-linux`).
-Nothing gates on the numbers — they are there so a change can be seen, not
-enforced. Note that the PR artifact only covers the 12 Linux-buildable
-packages; all 17 are measured only at release tags
+`make coverage` produces the same two files CI attaches as the `coverage-summary-*`
+artifact (`gh run download -n coverage-summary-linux`). Nothing gates on the
+numbers — they are there so a change can be seen, not enforced. Note that the
+per-PR artifact only covers the 12 Linux-buildable packages; every package with a
+test target is measured only on release tags
 (see [ADR 0007](docs/adr/0007-linux-ci-swift.md)).
 
 ## Code Style
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,11 +39,19 @@ make build
 ```bash
 make test         # Preferred — runs all package tests
 swift test        # Also works (runs root package tests)
+make coverage     # Full sweep with instrumentation (~30 min) → coverage/coverage-summary.{json,md}
 ```
 
 Tests use the Swift Testing framework (`@Test` macros). Currently, tests exist
 in `CrowCore`. When adding new functionality to any package, include a test
 target in that package's `Package.swift` and add tests.
+
+`make coverage` produces the same two files CI attaches to every run as the
+`coverage-summary-*` artifact (`gh run download -n coverage-summary-linux`).
+Nothing gates on the numbers — they are there so a change can be seen, not
+enforced. Note that the PR artifact only covers the 12 Linux-buildable
+packages; all 17 are measured only at release tags
+(see [ADR 0007](docs/adr/0007-linux-ci-swift.md)).
 
 ## Code Style
 
--- a/Makefile
+++ b/Makefile
@@ -106,12 +106,17 @@ test:
 # byte-comparable over the same package set. This is a full sweep — budget ~30
 # minutes. All 17 packages build here; the Linux PR lane sees only 12, so its
 # artifact reports a different (smaller) tree. See docs/adr/0007-linux-ci-swift.md.
+#
+# `|| exit 1` matters: a shell for-loop does not abort on a failing iteration and
+# a recipe's status is the last iteration's, so without it a suite failing at
+# package 4 of 17 would still print an authoritative-looking table — built partly
+# from the *previous* run's export for the package that just failed — and exit 0.
 coverage:
 	@bash scripts/generate-build-info.sh
 	@for pkg in Packages/*/; do \
 		if [ -d "$$pkg/Tests" ]; then \
 			echo "==> Testing $$(basename $$pkg) with coverage..."; \
-			swift test --enable-code-coverage --package-path "$$pkg"; \
+			swift test --enable-code-coverage --package-path "$$pkg" || exit 1; \
 		fi; \
 	done
 	@bash scripts/coverage-summary.sh
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build daemon app run setup install uninstall clean check test parity help daemon-run docs
+.PHONY: build daemon app run setup install uninstall clean check test coverage parity help daemon-run docs
 
 # Install destination and build config (override on the command line, e.g.
 # `make install BINDIR=/usr/local/bin` or `make build CONFIG=release`).
@@ -53,6 +53,7 @@ help:
 	@echo "  setup      Check build prerequisites"
 	@echo "  check      Verify all build and runtime prerequisites"
 	@echo "  test       Run all package tests, then the parity gates"
+	@echo "  coverage   Run all package tests with coverage → coverage/coverage-summary.{json,md}"
 	@echo "  parity     Run just the source-level drift gates (catalogs + CLI/RPC parity)"
 	@echo "  docs       Regenerate docs/cli.md from the CLI's ArgumentParser metadata"
 	@echo "  install    Symlink crow + crowd into ~/.local/bin (override BINDIR=, CONFIG=release)"
@@ -98,6 +99,22 @@ test:
 		fi; \
 	done
 	@$(MAKE) --no-print-directory parity
+
+# Local twin of the CI coverage artifact (CROW-928). Runs the same suites as
+# `test` with instrumentation, then merges the per-package llvm-cov exports
+# through the same script CI uses, so a local run and a downloaded artifact are
+# byte-comparable over the same package set. This is a full sweep — budget ~30
+# minutes. All 17 packages build here; the Linux PR lane sees only 12, so its
+# artifact reports a different (smaller) tree. See docs/adr/0007-linux-ci-swift.md.
+coverage:
+	@bash scripts/generate-build-info.sh
+	@for pkg in Packages/*/; do \
+		if [ -d "$$pkg/Tests" ]; then \
+			echo "==> Testing $$(basename $$pkg) with coverage..."; \
+			swift test --enable-code-coverage --package-path "$$pkg"; \
+		fi; \
+	done
+	@bash scripts/coverage-summary.sh
 
 # Source-level drift gates. Cheap, no build required, and both also run in CI —
 # see the `parity` job in .github/workflows/ci.yml.
--- a/docs/adr/0007-linux-ci-swift.md
+++ b/docs/adr/0007-linux-ci-swift.md
@@ -21,7 +21,7 @@ PR CI and cache-warm run on `ubuntu-latest` inside the official `swift:6` contai
 - **The root `CrowTests` suite moves from every-PR to release-tag time.** Those tests `@testable import Crow`, and the root `Crow` target imports `CrowTerminal` (AppKit), so they cannot run in the Linux PR lane. They exercise root-target business logic (IssueTracker, Job decisions, Scaffolder, SessionService, …), not just GUI, so to avoid dropping that coverage entirely they now run in `release.yml`'s macOS `test` job at tag time — not on every PR. A logic regression there is caught at release, not on the PR that introduces it.
 - The Linux allow-list in `ci.yml`/`cache-warm.yml` must be updated by hand when a new Linux-compilable package is added. This is deliberate: a glob would silently try to build a new macOS-only package on Linux and turn CI red.
 - Only the SwiftPM dependency cache (`~/.cache/org.swift.swiftpm`) is retained, not compiled `.build` products, so each Linux run recompiles the packages from scratch. Acceptable given each package builds under its own `Packages/$pkg/.build`.
-- `CrowTelemetry` (Apple `Network.framework`) is excluded; it has no test target and nothing in the allow-list depends on it, so excluding it costs no test coverage.
+- `CrowTelemetry` (Apple `Network.framework`) is excluded because it does not compile on Linux, and nothing in the allow-list depends on it. It *does* have a test target — 5 test files, ~1,500 executable lines — so unlike the note originally recorded here, excluding it does cost test coverage on PRs; those tests run only in `release.yml`. (Corrected 2026-07-30 while adding coverage measurement, CROW-928.)
 
 ## Alternatives considered
 
--- a/scripts/coverage-summary.sh
+++ b/scripts/coverage-summary.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+# Collect Swift test coverage into one repo-wide summary (CROW-928).
+#
+# `swift test --enable-code-coverage` leaves an llvm-cov export at
+# Packages/<Pkg>/.build/<triple>/debug/codecov/<Pkg>.json. This gathers whatever
+# exports are on disk, merges them (scripts/coverage-summary.swift does the
+# arithmetic — read its header for the two ways this can silently go wrong), and
+# writes coverage-summary.json + coverage-summary.md into the output directory.
+#
+# It reports on exactly the packages that were tested, so the numbers differ by
+# lane on purpose: the Linux PR lane covers the 12 packages in ci.yml's
+# LINUX_PACKAGES allow-list (~36% of the tree), while release.yml's macOS job
+# is the only place all 17 are measured — see docs/adr/0007-linux-ci-swift.md.
+# It does not run the tests itself; `make coverage` is the local front door.
+#
+#   bash scripts/coverage-summary.sh [OUT_DIR]     # default: coverage/
+set -euo pipefail
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$ROOT"
+
+OUT_DIR="${1:-coverage}"
+
+# Pin the depth so this can only match Packages/<Pkg>/.build/<triple>/debug/
+# codecov/<file>.json. `find` does not follow the .build/debug symlink, so each
+# export is seen once. LC_ALL=C keeps the ordering (and therefore the output)
+# identical across machines.
+REPORTS=()
+while IFS= read -r report; do
+  REPORTS+=("$report")
+done < <(find Packages -mindepth 6 -maxdepth 6 -type f \
+              -path 'Packages/*/.build/*/debug/codecov/*.json' | LC_ALL=C sort)
+
+if [ ${#REPORTS[@]} -eq 0 ]; then
+  echo "ERROR: no coverage exports found under Packages/*/.build/*/debug/codecov/." >&2
+  echo "       Run the suites with coverage instrumentation first, e.g. 'make coverage'" >&2
+  echo "       or 'swift test --enable-code-coverage --package-path Packages/<Pkg>'." >&2
+  exit 1
+fi
+
+echo "==> Merging ${#REPORTS[@]} coverage export(s) into $OUT_DIR/"
+
+# Compile the merger rather than running it through `swift`'s interpreter mode:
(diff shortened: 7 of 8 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Reviewed the two new scripts, both workflow edits, the Makefile target, and the ADR correction. I compiled `scripts/coverage-summary.swift` with `swiftc -O` (clean), exercised the merger against hand-built llvm-cov exports, and ran the real `scripts/coverage-summary.sh` end-to-end against a genuine `swift test --enable-code-coverage` export.

**What I verified holds:**
- The export-path assumption is real — `swift test --enable-code-coverage --package-path Packages/CrowPersistence` produced `Packages/CrowPersistence/.build/arm64-apple-macosx/debug/codecov/CrowPersistence.json`, exactly the shape `find -mindepth 6 -maxdepth 6` targets. The sibling `default.profdata` / `*.profraw` are correctly excluded by the `*.json` glob, and the `.build/debug` symlink sits at depth 5 so it can't double-count.
- **Dependency scoping works.** A single CrowPersistence run attributed 412 lines to CrowPersistence and CrowCore's 3,454 to CrowCore, with a `—` row for CrowCore — gotcha 1 from the ticket is genuinely handled, not just described.
- `.derived/runner.swift` and non-repo paths (`/usr/lib/swift/…`) are dropped; `covered ≤ count` holds under `absorb`, since `max(covered) ≤ max(count)`.
- **Determinism confirmed** — repeat runs are byte-identical, and the emitted JSON contained zero absolute paths.
- `swiftc` compiles clean; the JSON parses under `python3 -c json.load`.
- ADR 0007's correction is accurate: CrowTelemetry has 5 test files against ~1,900 source lines.

### Security Review

**Strengths:**
- No secrets, no third-party upload service, no new dependencies. `actions/upload-artifact@v4` is first-party and matches the repo's existing major-tag pinning convention.
- **No path disclosure.** `repoRelative` returns `nil` for anything it can't re-root, and those files are dropped rather than emitted, so absolute build paths cannot reach the artifact. Confirmed empirically against a real export.
- Hand-rolled `jsonString` correctly escapes `"`, `\`, control chars via `\u%04x`. Nothing attacker-controlled reaches it anyway (keys are package/path names).
- `mktemp -d` + `trap … EXIT` for the compile scratch dir; `set -euo pipefail` throughout.
- The trigger stays `pull_request` (not `pull_request_target`), so a fork PR compiling and running `coverage-summary.swift` gets no secret access — no new privilege surface over the `swift test` this lane already runs.

**Concerns:** none found.

### Code Quality

**🟡 1. `.github/workflows/ci.yml:62,68` — every Linux package now compiles twice per PR.**

Toggling `--enable-code-coverage` invalidates SwiftPM's incremental build. Measured on CrowIPC:

```
swift build                          → [8/8] compiled, Build complete (2.40s)
swift build --enable-code-coverage   → [7/7] RECOMPILED everything
swift build --enable-code-coverage   → Build complete (0.07s)   # no-op, as expected
swift build                          → [7/7] RECOMPILED everything again
```

So line 62's uninstrumented `swift build` is now pure waste: line 68 throws it away and rebuilds the whole target graph instrumented. Across 12 packages on a lane that (per ADR 0007) retains no `.build` cache and recompiles cold every run, that is roughly a doubling of the non-test compile work on every PR. One-word fix — add `--enable-code-coverage` to line 62 so both invocations share a build dir. `release.yml` doesn't have this problem (single `swift test`, no separate build).

**🟡 2. `Makefile:109-117` — `make coverage` prints a summary and exits 0 even when suites fail.**

A failing iteration doesn't abort a shell `for` loop, and the recipe's status is the last iteration's. Reproduced with a minimal Makefile of the same shape: a failing middle iteration → loop exits 0 → the next recipe line runs anyway. So if `swift test` fails on package 4 of 17, the sweep continues, `coverage-summary.sh` still runs on whatever exports are on disk (stale or partial for the failed package), and `make coverage` exits 0 having printed an authoritative-looking table.

`make test:94-100` has the same shape, so this is pre-existing style — but there a failure at least leaves visible red in the log, whereas here it silently yields *a number*, which is precisely what this PR is trying to make trustworthy. `swift test --enable-code-coverage --package-path "$$pkg" || exit 1;` inside the loop.

**🟡 3. `scripts/coverage-summary.swift:368` — the headline says "Whole repo" but reports only what the lane measured.**

From my real single-package run, the artifact's first line read:

> **Whole repo: 22.2% line / 20.2% region** — 3,866 executable lines, 3,006 missed.

against a repo of ~40,500 executable lines. On the `coverage-summary-linux` artifact the same line will assert a repo-wide figure computed over ~36% of the tree. The `—` legend below the table explains per-package gaps, and the PR body and both workflow comments are scrupulous about the lane distinction — but the artifact is downloaded and read out of context, and its single most quotable line is the one that isn't hedged. `Measured packages:` or `This lane:` would fix it.

**🟢 4. `CONTRIBUTING.md:46-47`** — "Currently, tests exist in `CrowCore`" is stale, and now sits three lines above a new paragraph documenting 12 measured packages. Pre-existing, but this PR is already editing the block.

**🟢 5. `ci.yml:79-83`, `release.yml:37-41`** — no `if: always()` on the upload steps, so a single failing suite loses the artifact for a run where you'd arguably most want it; and no `retention-days`, so these accumulate at the 90-day default.

**🟢 6. `coverage-summary.swift:341-365`** — a package with `ownSuiteMeasured: false` still emits `ownSuite: {covered: 0, percent: 0.00}`. The markdown renders `—`, but a machine consumer that skips the flag reads a false 0%. `null` would make it unmissable.

**🟢 7. `coverage-summary.sh:28-39`** — merges whatever exports are on disk with no provenance in the output. Deliberate (byte-determinism precludes a timestamp), but a deleted package's or removed test target's leftover `codecov/*.json` keeps contributing to local runs indefinitely. Recording the merged export list in the JSON would cost nothing deterministic.

**Not run:** `shellcheck` (not installed locally — CI's `shellcheck scripts/*.sh` job covers it), and the full 17-package sweep.

### Summary Table

The core merger is careful, well-reasoned work — the two gotchas are handled for real, determinism holds, and the honest accounting of the 61.9% vs 62.7% gap in the PR body is exactly the right call. The findings are all in the surrounding plumbing, and #1 and #2 are one-line fixes.

**Recommendation:** Request Changes — driven by 0 Red, 3 Yellow, 4 Green findings.

---

dgershman:
CI green on all four checks, and the `coverage-summary-linux` artifact downloads with both files (131 KB JSON + 1.5 KB Markdown). That also settles the one thing I could not verify locally: `swiftc` compiling the merger inside the `swift:6.1` container works.

The `—` row behaved as predicted — CrowAutostart is compiled on Linux via CrowCLI but its own suite never runs there, so it reports `—` with 5.7% any-suite rather than a false `0.0%`.

One observation worth recording. The CI (Linux) numbers differ slightly from my local (macOS) run over the same 12 packages:

| | Lines | Missed | Line % |
|---|---:|---:|---:|
| Local macOS | 15,136 | 3,736 | 75.3% |
| CI Linux | 15,126 | 3,729 | 75.3% |

That is expected, not drift: `#if canImport(Darwin)` / `Glibc` branches mean the two platforms genuinely compile a slightly different set of executable lines (CrowCore is 3,454 on macOS vs 3,444 on Linux). Determinism holds where it is claimed — repeat runs on the same platform over the same package set are byte-identical, which is what `make coverage` matching the artifact means in practice. Comparing a Linux artifact against a macOS one is not apples-to-apples and never can be.

Test plan status:

- [x] `shellcheck` passes
- [x] Full local sweep matches the baseline table (15/17 exact, 3 off by 1–3 missed lines from timing-sensitive suites)
- [x] Repeat runs byte-identical
- [x] `coverage-summary.json` parses
- [x] Linux lane produces a downloadable artifact with both files for the 12 `LINUX_PACKAGES`
- [x] `coverage/` gitignored
- [ ] Release lane (all 17 packages) — fires only on a `v*` tag, so it cannot be exercised on this PR. It reuses the same two steps on macOS, which is the configuration the full local sweep validated.

dhilgaertner (changes requested):
## Code & Security Review

Verified locally: compiled `scripts/coverage-summary.swift` with `swiftc -O` (clean), exercised the merger against synthetic llvm-cov exports, and reproduced the `find` pattern and both error paths.

**What holds up.** The merge arithmetic does what the header claims. On a fixture with `A.json` (covering `A/Sources`, `B/Sources`, `C/Sources`, plus `A/Tests/…` and a `*.derived/runner.swift`) and `B.json` (covering `B/Sources` and a cold view of `A/Sources`):

- Dependency scoping works — every file lands on the package owning `Packages/<Pkg>/Sources/…`; `Tests/` and `.derived/` are dropped.
- Max-fold works — `B` reports its own 90%, not the 20% `A`'s binary saw; `A` reports 80%, not the 0% `B`'s binary saw.
- `C` (linked only) gets `ownSuiteMeasured: false` and a `—` row, not a fabricated 0%.
- Output is byte-identical across three runs (`shasum` match on both files) despite `merged`/`packages` being `Dictionary` — every order-dependent path is sorted with a name tiebreak, and the roll-ups are integer sums. The sort comparator is a valid strict weak ordering.
- Emitted JSON parses; the hand-rolled `jsonString`/`grouped`/`fixed` helpers are locale-independent as documented, and the artifact carries no absolute paths.
- Error paths are loud and correct: no exports → exit 1 with a usable message; export with no `Packages/*/Sources` files → exit 1 naming the repo root; malformed JSON → exit 1 naming the file.
- `find -mindepth 6 -maxdepth 6` matches exactly the intended depth and does not double-count through the `.build/debug` symlink (confirmed against a tree with that symlink present).
- `${#REPORTS[@]}` under `set -u` is safe on bash 3.2 (the macOS lane) — checked, returns `0` rather than erroring.

### Security Review

**Strengths:**
- No secrets, no network egress, no third-party upload service. `actions/upload-artifact@v4` is first-party and matches the repo's existing `@v4` convention for `checkout`/`cache`.
- No new token scopes. `release.yml`'s pre-existing top-level `contents: write` is untouched; the artifact upload needs nothing beyond the default.
- No injection surface: `coverage-summary.sh` takes no untrusted input, `mktemp -d` is quoted, and the `trap … EXIT` cleanup is correct. `swiftc` compiles a repo-tracked source, not anything derived from the environment.
- Artifact contents are repo-relative paths and integers — no timestamps, hostnames, absolute paths, or environment leakage. Confirmed in the generated output.

**Concerns:** none found.

### Code Quality

**🟡 1. The Linux lane now compiles every package twice** — `.github/workflows/ci.yml:62` + `:68`

`swift build --package-path "Packages/$pkg"` builds uninstrumented; `swift test --no-parallel --enable-code-coverage` then rebuilds the same library targets with `-profile-generate -profile-coverage-mapping`. Toggling that flag invalidates the build plan, so nothing is reused.

Measured on `Packages/CrowIPC` with a shared scratch path:

```
1. swift build                      → cold, 8/8 tasks
2. swift build                      → no-op, 0.09s  (build products reused)
3. swift build --enable-code-coverage → 7/7 tasks recompiled
```

Per ADR 0007 this lane retains only the SwiftPM dependency cache, not compiled products, so every PR pays a full extra compile of all 12 packages and the plain `swift build` is now pure waste. Adding the flag to line 62 fixes it — I confirmed `swift build` accepts `--enable-code-coverage`:

```yaml
swift build --enable-code-coverage --package-path "Packages/$pkg"
```

**🟡 2. The new comments call CrowAutostart "Darwin-only", which this PR's own artifact disproves** — `.github/workflows/ci.yml:72`, `.github/workflows/release.yml:32`

Both new comments group CrowAutostart with the packages the Linux lane "cannot compile". It compiles there today:

- `Packages/CrowCLI/Package.swift:14,27` — CrowCLI depends on it directly, and CrowCLI is in `LINUX_PACKAGES` and green.
- `DaemonProbe.swift:2-5` / `LaunchdAutostart.swift:2-5` — `#if canImport(Darwin) … #else import Glibc`, i.e. deliberately portable.
- The PR body's own Linux table reports 418 executable lines for it.

The reason it shows `—` is that it is absent from `LINUX_PACKAGES`, not that it fails to build. Worth getting right given this PR's stated purpose includes correcting exactly this class of stale claim in ADR 0007 (`docs/adr/0007-linux-ci-swift.md:24`) — it would be a shame to fix one and add another.

Separately, and optionally: `Packages/CrowAutostart/Tests/` exists and imports only `Testing` + `Foundation`, so adding `CrowAutostart` to `LINUX_PACKAGES` would likely turn the single `—` row on Linux into a real number. I can't run the Linux lane from here to confirm the suite passes (`LaunchdAutostartTests` may assert launchd paths), so treat that as a follow-up, not a blocker. Correcting the two comments is the ask.

**🟡 3. `make coverage` can publish a summary from a suite that failed** — `Makefile:111-117`

The `for` loop has no `|| exit 1`, and make recipes run under plain `sh`. A package whose tests fail mid-sweep does not abort the loop; `coverage-summary.sh` then merges whatever is on disk and `make coverage` exits 0 with an authoritative-looking table. Because `.build/*/debug/codecov/<Pkg>.json` persists between runs, the failed package silently contributes a *previous* run's export rather than being visibly absent.

This mirrors the existing `test:` target (`Makefile:95-100`), so it is a convention, not a regression — but the consequence differs. `test` fails loudly in the log; `coverage` emits a wrong number that reads as fact, and the PR description leans on this target as the verification method. One token fixes it:

```make
swift test --enable-code-coverage --package-path "$$pkg" || exit 1; \
```

**🟢 4. Whole-export decode holds the full JSON in memory** — `scripts/coverage-summary.swift:230-238`

`FileManager.contents(atPath:)` plus `JSONDecoder` parses each export in full, including the `segments`/`expansions` arrays that dwarf the `summary` fields actually read. Peak is one export at a time so nothing is at risk today; noting it because these files grow with the tree.

**🟢 5. `repoRelative`'s `/Packages/` fallback could mint a phantom package** — `scripts/coverage-summary.swift:194-196`

A dependency checked out under a path containing `/Packages/<Name>/Sources/` would be re-rooted and attributed to a package that does not exist in this repo. No current dependency has that layout, and the primary `hasPrefix(repoRoot)` branch handles the real cases; purely defensive.

**🟢 6. Wording nit** — `CONTRIBUTING.md:47`

"the same two files CI attaches to every run" — the macOS artifact only exists at release tags. The following sentence clarifies the 12-vs-17 split, so this is a half-sentence away from being exactly right.

### Verified claims

- `docs/adr/0007-linux-ci-swift.md:24` correction is accurate — `Packages/CrowTelemetry/Tests/` holds 5 test files, matching the new text.
- `LINUX_PACKAGES` is 12 of the 17 package directories; the remaining 5 are named correctly in the comment (setwise), even though the "Darwin-only" characterisation is wrong for one of them.
- `coverage/` is gitignored and nothing generated is committed.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 3 Yellow, 3 Green] findings.

The design is sound and the honesty about the 61.9%-vs-62.7% gap is the right call — none of this is architectural. Findings 1 and 2 are both one-line edits in `ci.yml`, and 3 is one token in the `Makefile`.

dgershman:
Thanks — both reviews were unusually concrete, and the two measured findings (double compile, silent-pass sweep) were things I would not have caught by reading. Pushed as `7a50742e`. Every finding from both reviews is addressed in code except one, explained at the bottom.

## Yellow

**1. Double compile (both reviews).** Fixed — `--enable-code-coverage` on the `swift build` too. I verified `swift build` accepts the flag before applying it. The comment now says *why* both invocations carry it, so nobody "cleans up" the apparent redundancy later.

**2. `make coverage` publishing from a failed sweep.** Fixed with `|| exit 1`. I reproduced both shapes rather than taking it on faith:

```
broken (no || exit 1):  ok a / ok c / SUMMARY RAN   → exit 0
fixed  (|| exit 1):     ok a / make: *** Error 1    → exit 2
```

Your point about *why* this one matters more than the same shape in `test:` is the part I'd missed — the stale `codecov/<Pkg>.json` from a previous run means the failed package doesn't go absent, it silently contributes yesterday's number. That's a wrong figure presented as fact, which is the exact failure mode this PR exists to prevent.

**3. "Whole repo" headline.** Fixed. Now:

> **Measured packages: 75.3% line / 70.5% region** — 15,136 executable lines, 3,736 missed.
>
> That total covers the 13 packages in *this* run, not necessarily the whole repository — CI's Linux pull-request lane builds an explicit subset of the tree (see `docs/adr/0007-linux-ci-swift.md`).

You're right that this was the weak point: I was careful about the lane distinction everywhere the artifact *isn't* — PR body, workflow comments — and left it off the one line that travels.

**4. "Darwin-only" mischaracterisation of CrowAutostart.** Fixed, and thank you for catching it — adding a stale claim in the same PR that fixes one in ADR 0007 would have been a poor look. The comments now say it compiles on Linux and shows `—` because it is absent from `LINUX_PACKAGES`, not because it fails to build. I've left adding it to the allow-list as a follow-up per your framing and noted that in the comment, since ADR 0007 explicitly wants allow-list changes made consciously. Happy to file that issue if you want it tracked.

## Green — all implemented

- **`if: always()` + `retention-days: 30`** on both upload steps. This needed one extra piece: with `always()`, a lane that dies before any suite finishes would hit the summarize step with zero exports and go red a second time, pointing at the wrong thing. So `coverage-summary.sh` takes `--allow-empty`, which CI passes and `make coverage` does not — a developer running against an uninstrumented tree still gets the loud error.
- **`ownSuite: null`** when `ownSuiteMeasured` is false, at package *and* file level. Verified: `CrowAutostart ownSuite= None`.
- **Export provenance** — the JSON now lists the exports it merged, repo-relative and sorted, so a stale leftover is visible rather than silently folded in.
- **`repoRelative` fallback** now stats the re-rooted path before trusting it, so an external checkout containing `/Packages/<Name>/Sources/` can't mint a phantom package. The legitimate `/private/var` case still resolves, since that file does exist under the root.
- **CONTRIBUTING.md** — dropped "tests exist in CrowCore" and the "every run" implication about the macOS artifact.

## Not changed: 🟢4, whole-export decode

Leaving this one as-is, with numbers rather than a shrug. Current exports total 24 MB across 17 packages; the largest is CrowDaemon at 15 MB, and peak really is one at a time. Streaming would mean hand-writing an incremental JSON parser in Swift — the container has no `jq` or `python3`, which is what put us in Swift to begin with — so it is a substantial, bug-prone chunk of code against no present problem. The trigger to revisit is a single export approaching runner memory, which is two orders of magnitude away. Flagging rather than silently declining; say the word if you'd rather have it now.

## Re-verified after the changes

- `shellcheck scripts/*.sh` clean
- Repeat runs byte-identical
- JSON parses, no absolute paths
- Per-package numbers still match the baseline (CrowCore 76.7% / 804 missed, unchanged)
- `--allow-empty` exits 0 with a notice; without it, exit 1; unknown flag exits 2 with usage

dhilgaertner (approved):
## Code & Security Review

Third pass. Every finding from both prior reviews is addressed in code; the one deferral (Green — whole-export `JSONDecoder` memory) is correctly judged not-a-problem-today.

**How I verified this round.** Compiled `scripts/coverage-summary.swift` with `swiftc -O` (clean), ran the real `scripts/coverage-summary.sh` and the merger against a synthetic fixture built to trip every filter rule, and then **downloaded the actual `coverage-summary-linux` artifact** from run `30645927201` and cross-checked it independently.

### Prior findings — confirmed fixed

| # | Finding | Status |
|---|---|---|
| 🟡 | Double compile on the Linux lane | Fixed — `ci.yml:68` now carries `--enable-code-coverage`, with a comment explaining why both invocations share it |
| 🟡 | `make coverage` publishing from a failed sweep | Fixed — `Makefile:117` `\|\| exit 1`; I reproduced both shapes in a minimal Makefile (broken → `SUMMARY RAN`, exit 0; fixed → `Error 1`, no summary) |
| 🟡 | "Whole repo" headline | Fixed — `coverage-summary.swift:401` now reads `Measured packages:` |
| 🟡 | CrowAutostart mis-described as Darwin-only | Fixed — `ci.yml:80-85` states the opposite, correctly. Verified: `Packages/CrowCLI/Package.swift:14,27` does depend on it |
| 🟢 | Fabricated `0%` for unmeasured packages | Fixed — emits `"ownSuite": null`; confirmed in the real artifact (`CrowAutostart` → `None`) |
| 🟢 | No provenance for merged exports | Fixed — `exports` array; the real artifact lists all 12, repo-relative |
| 🟢 | `/Packages/` fallback could mint a phantom package | Fixed — `coverage-summary.swift:201` `fileExists` guard. My fixture fed it `/opt/deps/Packages/Ghost/Sources/Ghost/G.swift` (4,242 lines); dropped, as intended |
| 🟢 | Missing `if: always()` / `retention-days` on upload | Fixed, both lanes |
| 🟢 | Stale `CONTRIBUTING.md` claims | Fixed |

### Verification against the shipped artifact

Downloaded `coverage-summary-linux` (131 KB JSON + 1.7 KB Markdown) and checked it standalone:

- **No path disclosure.** Zero absolute paths in either file — no `/__w/crow/crow`, no `/home`, no `/Users`. This is the security property that matters here and it holds on the real output, not just a fixture.
- **Arithmetic is internally consistent.** Per-file sums equal every package rollup (155 files, 0 mismatches); package rollups sum exactly to `totals` (15,126 lines / 11,397 covered); `covered ≤ count` on every metric; no `percent > 100` anywhere; per-package file lists sorted.
- **`ownSuite.count == anySuite.count`** for all 12 measured packages, i.e. the shared-denominator normalisation (`:281-286`) does what its comment claims.
- **Determinism** re-confirmed: three consecutive runs, byte-identical (`shasum` match on both files), and reordering the export arguments leaves the Markdown identical.

Merge semantics re-checked on a fixture with A (own 80%, seen cold at 0% by B), B (own 90%, seen cold at 20% by A), C (linked only), plus a `Tests/` file, a `*.derived/runner.swift`, a `/usr/lib/swift/` path, a phantom `/Packages/Ghost/`, and a `/private/var/…` resolved-path view of A. Result: A→80%, B→90%, C→`—` with `null`, everything else dropped, total exactly 250 lines. Both ticket gotchas hold.

Error paths all exit non-zero with usable messages: malformed JSON names the file, no-repo-sources names the repo root, missing file, too-few-args, unknown option → exit 2. `--allow-empty` downgrades only the no-exports case. `find -mindepth 6 -maxdepth 6` does not double-count through a `.build/debug` symlink (tested with one present).

### CI cost — the double-compile fix is real and measured

| Run | Build & Test |
|---|---:|
| Baseline, no coverage (`30642888584`, today) | 5m12s |
| This PR pre-fix, double compile (`5eac0c9`) | 6m44s |
| This PR now (`7a50742`) | **4m57s** |

The instrumented lane is at or below the pre-PR baseline. Whatever coverage costs is inside the noise, and the extra compile is gone.

### Security Review

**Strengths:**
- Trigger stays `pull_request`, not `pull_request_target` — a fork PR gets no secret access, and this adds no privilege surface over the `swift test` the lane already ran.
- **No `${{ }}` interpolation into any `run:` block** in the new steps — no script-injection vector from a PR title or branch name.
- No new dependencies, no network egress, no third-party coverage service. `actions/upload-artifact@v4` is first-party and matches the repo's existing `@v4` convention.
- Artifact contents are repo-relative paths and integers only — verified empirically on the shipped artifact, not just asserted.
- `mktemp -d` + `trap … EXIT`, `set -euo pipefail`, all expansions quoted; `swiftc` compiles a repo-tracked source with no environment-derived input.
- `retention-days: 30` bounds artifact storage rather than taking the 90-day default.

**Concerns:** none found.

### Code Quality

**🟢 1. `coverage-summary.swift:401,407` — "measured" carries two meanings four lines apart.**

The headline reads `Measured packages: 75.3%`, but that total spans all 13 packages in the run; the line below says `12 packages measured`. Measured-only would be 11,373/14,708 = **77.3%** — a 2.0pp gap, entirely CrowAutostart's 418 lines. The number and its derivation are correct and the "Any suite %" paragraph explains it; only the overloaded word is loose. `Packages in this run:` would settle it in one word.

**🟢 2. `ci.yml:87,94` / `release.yml:35,39` — `always()` also fires on cancellation.**

With `cancel-in-progress: true` (`ci.yml:10`), a superseded run still compiles the merger and uploads a partial artifact. `${{ !cancelled() }}` keeps the failed-suite case the comment argues for — which is the right call — without the cancelled-run noise.

**🟢 3. The Markdown half carries no provenance the JSON has.**

`exports` lands in the JSON only. On the partially-failed lane `always()` deliberately supports, the table is simply smaller with nothing saying so — accurate, as the comment claims, but a reader can't tell a clean 12-package run from a run that died at package 7. Echoing the export count into the `.md` would close it for free.

**🟢 4. `coverage-summary.sh:50-51` merges every export on disk.**

Fine in CI (fresh checkout). Locally, a package whose test target was deleted keeps contributing its last export and still reads `ownSuiteMeasured: true`. The `exports` list makes it diagnosable after the fact; pruning before the sweep would make it impossible. Narrow and local-only.

**🟢 5. `release.yml` — the new upload inherits top-level `contents: write`.**

Pre-existing grant (`:7-8`), not introduced here, but the `test` job now does something with it. A job-level `permissions: contents: read` would scope the artifact upload to what it actually needs.

### Verified claims

- ADR 0007's correction is accurate: `Packages/CrowTelemetry/Tests/` holds exactly 5 test files against 1,939 raw source lines, consistent with "~1,500 executable".
- No `Package.swift` uses a custom target `path:`, so the `Packages/<Pkg>/Sources/` ownership rule captures every package source. The only sources it skips are the root package's two `main.swift` shims — 9 lines total.
- `coverage/` is gitignored; nothing generated is committed.

**Not run locally:** `shellcheck` (not installed here — CI's Shell Lint job is green on `7a50742`), and the full 17-package macOS sweep.

### Summary Table

Both measured findings from the last round are fixed and I confirmed the fixes with numbers rather than by reading: the compile saving is 1m47s and lands the instrumented lane below the uncovered baseline, and the `|| exit 1` genuinely changes the failure shape. The five remaining items are wording and hygiene, none worth another round trip — take them or leave them.

---

dgershman:
CI green on `7a50742e`, and the artifact carries all the review changes — `ownSuite: null` for CrowAutostart, 12 exports recorded, no absolute paths, and the headline now reads "Measured packages" with the qualifier attached.

Your finding #1 also shows up in the wall-clock. Same job, same lane, only the build flag changed:

| | Build & Test |
|---|---:|
| `5eac0c9a` (uninstrumented `swift build`, then instrumented `swift test`) | 6m44s |
| `7a50742e` (both instrumented) | 4m57s |

**~27% off the Linux lane.** So the double compile was not just theoretically wasteful — sharing the flag put the lane meaningfully *below* where it sat before, which also settles the ticket's "no change to PR wall-clock beyond the instrumentation overhead" criterion more comfortably than I had it.

PR #231, opened 2026-05-01
Add quick action buttons to session detail header

dhilgaertner:
## Summary

- Adds PR-status-aware quick action buttons to the session detail header so users can act on a session card without switching into it and typing a prompt by hand.
- Each button maps 1:1 to an existing status badge: **Rebase & Fix Conflicts** (conflicts), **Address Review** (changes requested), **Fix Checks** (failing CI), **Merge PR** (ready to merge).
- Click → routes through a new `dispatchManual` on `AutoRespondCoordinator` → injects a deterministic single-line prompt into the session's managed Claude Code terminal via `TerminalManager.shared.send`.
- The `addressChanges` and `fixChecks` prompts share text with the existing auto-respond pipeline (`AutoRespondPrompts`) so manual and auto paths stay consistent. New prompts added for `fixConflicts` and `mergePR`, including GitLab variants.
- Buttons hide when the PR is merged or no status is known, and disable (with a tooltip) when the session has no managed terminal.

## Test plan

- [ ] `make build` passes
- [ ] `make test` passes (CrowCore + Tests/)
- [ ] Open a session whose PR has merge conflicts → Rebase button appears, click types the rebase prompt into the terminal
- [ ] Open a session with changes requested → Address Review button appears
- [ ] Open a session with failing checks → Fix Checks button appears
- [ ] Open a session that is ready to merge (passing + approved + mergeable) → Merge PR button appears as the prominent green action
- [ ] Open a session whose PR is merged → no quick action buttons render
- [ ] Open a session with no managed terminal → buttons render disabled with tooltip
- [ ] Existing detail header buttons (Open VS Code, Open Terminal, In Review, Mark Completed, Delete) are unchanged

Closes #230

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #838, opened 2026-07-24
crow:merge label added via right-click has no effect — no merge icon + auto-merge skips 'no-crow-merge-label' despite label present on the PR


dhilgaertner:
## Symptom

Right-clicking a session → **"Add label crow:merge to PR"** appears to do nothing:
1. The session never shows the merge icon.
2. Auto-merge never fires, even when the PR is green and mergeable.

## What's actually happening (live diagnosis)

Reproduced against the live `crow-835-wheel-scroll-speed` session (`1BAACE1B-A5BC-4C08-B059-ADC486AD66C1`, status `inReview`, linked to **PR #836**):

- **The label *was* added to GitHub.** The right-click action reached the API — `crow:merge` is on PR #836 (added `2026-07-24T17:02:35Z` by `dhilgaertner`, confirmed via REST `gh pr view` and the GitHub **timeline**). So the action isn't fully dead; the *outcome* (icon + merge) is what's missing.
- **PR #836 meets every auto-merge precondition:** OPEN, not draft, author = the viewer (`dhilgaertner`), all checks `SUCCESS ×3`, mergeable.
- **But Crow's auto-merge watcher reports `#836:no-crow-merge-label` on every poll** after the label was added — 17:03:15, 17:04:23, 17:05:27, … (still doing so 6+ min later). Each of those poll lines is emitted from inside a full `refresh()` cycle, so the GraphQL is being re-fetched each time.
- **GitHub *does* return the label** via the same GraphQL shape the app uses: `... pullRequests { nodes { labels(first: 20) { nodes { name } } } }` → `#836 → [{name:"crow:merge"}]` (checked live at 17:08).

So this is **not** eventual consistency, **not** the `labels(first:)` window, and **not** a failed label write. Crow's in-memory `ViewerPR.labels` for #836 simply doesn't reflect the `crow:merge` GitHub has — and **both broken symptoms read from that same field**:
- Merge **icon**: the per-session PR runtime state `hasMergeLabel: pr.labels.contains { … "crow:merge" }` (`IssueTracker.swift` ~:2867).
- **Auto-merge**: `autoMergeSkipReason` gates on `pr.labels.contains { … "crow:merge" }` → returns `.noMergeLabel` (`IssueTracker.swift:1932`).

## Root cause (where the stale label view comes from)

The label is available to the fetch (raw GraphQL returns it) but doesn't survive into the `ViewerPR` the consumers read. Candidate loci to confirm/fix:

[...]

PR #930, opened 2026-07-31
Auto re-request review when a changes-requested PR has been fixed (CROW-921)

### LINKED ISSUE #921: Auto-respond loop dead-ends: a changes-requested PR that was fixed outside the addressChanges prompt never gets review re-requested
A PR whose `CHANGES_REQUESTED` review has been **fully addressed** can park in
`changesRequested` forever. Both halves of the automation loop go idle — each
correct by its own rule — because the only thing that re-requests review is a
sentence inside a prompt that, by construction, cannot fire once the fixes land.

Observed live on `corveil/corveil#1898` (Crow session `6A132C6D-D7A3-48E1-8A57-C698272B3574`).

## Reproduction / evidence

| Time (UTC) | Event |
|---|---|
| 17:43:56 | `auto-rebase: #1898 rebased & force-pushed (was BEHIND)` |
| 17:47:45 | `CHANGES_REQUESTED` #1 — 1 Yellow (bulk-revive `skipped` conflates cap-truncation with dedup-blocking) |
| 17:49:40 | `CHANGES_REQUESTED` #2 — 1 Red (`cliclient_test.go` no longer compiles) |
| 17:54:56 | commit `af7a4de` *"fix(cliclient): update #1893's test to the WorkerRunListParams signature"* — fixes the Red |
| 18:33:56 | `auto-rebase: #1898 conflicts — delegating to agent` |
| 18:40:30 | rebase lands (3 commits restamped) |
| 18:44:52 | commit `e0e49fc` — rebase reconciliation |
| 19:01:20 | commit `819c474` *"Review finding (dgershman, Yellow)"* — fixes the Yellow |
| 19:02:11 | agent idle (`activityState = done`) |

Both findings are genuinely fixed; `819c474`'s message walks through each one and
states which Green it deliberately leaves. `needs-refine fired` never appears in
`crowd-automation.log` for this session — correctly, since
`PRStatus.needsRefine` (`Packages/CrowCore/Sources/CrowCore/Models/PRStatus.swift:150`)
requires `lastSubstantiveCommitAt < lastChangesRequestedAt` and here it is
`19:01:20 > 17:49:40`.

End state, from `crow get-state`:

```json
"reviewStatus": "changesRequested",
"lastChangesRequestedAt": "2026-07-29T17:49:40Z",
"lastSubstantiveCommitAt": "2026-07-29T19:01:20Z"
```

`reviewRequests` on the PR is empty — GitHub cleared it when the review was submitted.

## Root cause

The only re-request-review mechanism in the codebase is prose inside the
`addressChanges` auto-respond prompt,
`Packages/CrowEngine/Sources/CrowEngine/AutoRespondCoordinator.swift:164`:

> "After pushing, re-request review from each reviewer who requested changes by
> running `gh pr edit … --add-reviewer <login>` for each one…"

`rg 'add-reviewer|requestReviews'` finds nothing else. That prompt is reachable
only when `needsRefine` is true — a condition that goes false the instant the
fixes land. So whenever the agent addresses a review by **any path other than
the auto-respond prompt** (its own in-flight work, an auto-rebase conflict
delegation re-engaging it, a manual nudge, the user typing in the terminal), the
re-request never happens.

The reviewer side then can't recover either: `IssueTracker.reviewKickoffAction`
(`Packages/CrowEngine/Sources/CrowEngine/IssueTracker.swift:3145`) — including
its `.reReview` head-advance path — only sees PRs matched by
`review-requested:@me` (`Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubCodeBackend.swift:67`).
With the request consumed and never re-added, the PR is permanently invisible to it.

Note the auto-rebase is **not** the cause. It is what re-engaged the agent, but
the dead-end occurs with or without it.

## Proposed fix

Today there are two states for a changes-requested PR; the loop needs three:

1. `changesRequested` ∧ no commits since review → **needs refine** — today's `needsRefine`, unchanged
2. `changesRequested` ∧ commits since review ∧ **no pending review request** → **awaiting re-request** ← missing
3. `changesRequested` ∧ pending review request → **awaiting reviewer** — nothing to do

State 2 is one field away. Add `reviewRequests(first:10){totalCount}` to the PR
query (`GitHubCodeBackend.swift:385`), carry `hasPendingReviewRequest` on
`PRStatus`, and add an `applyAutoReRequestReview` watcher alongside
`applyAutoMerge` / `applyAutoRebase` (`IssueTracker.swift:1935`), behind a new
`autoRespond.autoReRequestReview` toggle.

Make it **deterministic rather than prompt-driven** — the daemon runs
`gh pr edit <url> --add-reviewer <login>` for each reviewer whose latest review
is `CHANGES_REQUESTED`. Routing a one-line API call through an agent turn burns
tokens and re-inherits the `terminalIdle` gate that made this fragile in the
first place; `applyAutoMerge` already writes to PRs directly. Keep the prompt
hint as belt-and-braces.

## Two supporting fixes

**1. The documented rebase gap becomes load-bearing.**
`GitHubCodeBackend.swift:535-541` knowingly lets a rebase restamp `committedDate`
and advance `lastSubstantiveCommitAt` (it did exactly that to three commits at
18:40:30 above — harmless only because a real fix had already closed the window).
With a re-request watcher gated on "commits since review", a rebase-only advance
would fire a **no-op review round** at the reviewer. Switching the rule to
`authoredDate` — already in the GraphQL response, zero extra cost, preserved
across rebase — removes the false negative in `needsRefine` *and* the false
positive in the new watcher, and is cheaper than the tree-equality check CROW-508
deferred.

**2. Observability.** `IssueTracker` logs only when needs-refine *fires*
(`IssueTracker.swift:1905`), never when it is evaluated and suppressed.
Diagnosing this required pulling `prStatus` out of `crow get-state` and
hand-converting Apple reference-date timestamps. One line per poll per
changes-requested PR —
`needs-refine: #1898 gated (lastCR=…, lastCommit=…, idle=…)` — makes it a
30-second read.

## Acceptance

- [ ] A PR in `changesRequested` with commits after the review and no pending
      review request gets its reviewers re-requested automatically.
- [ ] No re-request fires when the only post-review commit advance came from a
      rebase restamp (`authoredDate` unchanged).
- [ ] No re-request fires while a review request is already pending.
- [ ] `needsRefine` behaviour is unchanged for states 1 and 3; existing
      `PRStatusNeedsRefineTests` still pass.
- [ ] The new watcher is gated by `autoRespond.autoReRequestReview`, readable and
      writable via `crow automation get` / `set`, and covered in the CLI parity ledger.
- [ ] Suppressed needs-refine evaluations are logged.

## Workaround

`gh pr edit https://github.com/corveil/corveil/pull/1898 --add-reviewer dgershman`

The "Address changes" button also unsticks it, but sends the agent to re-fix work
it has already done.

dhilgaertner:
Closes #921

A PR whose `CHANGES_REQUESTED` findings were genuinely addressed could park in `changesRequested` forever. Both halves of the loop went idle, each correct by its own rule:

- `PRStatus.needsRefine` goes false the moment the fix lands.
- The **only** re-request mechanism in the codebase was a sentence *inside* the `addressChanges` prompt — which that same rule gates.

So whenever the agent addressed a review by any other path (its own in-flight work, an auto-rebase conflict delegation, a human nudge), the re-request never happened — and with the host's review request already consumed, the PR was invisible to the reviewer's queue and to `review-requested:@me` alike.

## The fix: name the third state

```swift
public enum ChangesRequestedState {
    case notApplicable      // not changes-requested / closed / no anchor
    case needsRefine        // no work since the review     → prompt the agent
    case awaitingReRequest  // fix landed, no request open  → re-request  ← NEW
    case awaitingReviewer   // request pending              → wait
}
```

`needsRefine` and the new `applyAutoReRequestReview` watcher both derive from this one function, so they cannot drift into overlapping or contradictory conditions. `needsRefine` keeps its signature and every call site.

**Deterministic, not prompt-driven.** The watcher calls `gh pr edit --add-reviewer` from the daemon. Routing a one-line API call through an agent turn costs a full turn's tokens, can't be verified, and would re-inherit the very gate that caused the dead-end; `applyAutoMerge` already writes to PRs directly. The prompt hint stays as belt-and-braces — re-requesting an already-requested reviewer is a no-op.

**It waits for the agent, with an escape hatch.** A session with no launched agent terminal counts as settled. Unlike `needsRefine`, `awaitingReRequest` is a *stable* condition — it holds until somebody adds the request — so waiting only delays by a poll or two, while avoiding a reviewer ping (or a spawned review session, in auto-review repos) while the agent is still on finding 2 of 3.

Gated by `autoRespond.autoReRequestReview`, **on by default** alongside `respondToChangesRequested` — the prompt has always instructed the agent to do this, so the toggle completes expected behaviour rather than adding new behaviour.

## Two supporting fixes

**`authoredDate` instead of `committedDate`.** A rebase replays feature commits with the *committer* date rewritten to ~now, which read as "the agent pushed a fix". This closes the documented CROW-508 gap in both directions: a rebase restamp no longer suppresses a genuine needs-refine, nor fires a no-op review round at a reviewer.

Not exotic — of eight live viewer PRs checked while writing this, three carried committer dates hours to days ahead of their author dates from ordinary rebases:

| PR | max authored | max committed |
|---|---|---|
| #1898 | Jul 29 19:40 | **Jul 31 14:46** |
| #1974 | Jul 31 14:25 | Jul 31 14:45 |
| #231 | May 27 | **Jun 1** |

*Accepted gap, documented in `parsePRNode`:* `--amend` and `cherry-pick` also preserve author dates, so an amend-delivered fix is indistinguishable from a rebase by dates alone (telling them apart needs the tree check CROW-508 deferred). Both failure directions are the safe one — Crow nudges the agent rather than pinging a human.

**Suppressed evaluations are logged.** Previously only *firing* was logged, so a stuck PR left no trace at all; diagnosing #921 meant reading raw `crow get-state` and hand-converting Apple reference dates. Now:

```
needs-refine: #1898 gated (reason=awaiting-re-request, state=awaitingReRequest,
  lastCR=…, lastCommit=…, pendingRequest=no, idle=yes, cooldown=ok)
```

Rate-limited to changes plus an hourly heartbeat, shared with the watcher's skip lines — a line every 60s poll (~1440/day/PR) would bury the signal as effectively as silence.

## Note on ordering

The pending-request check sits **above** the `lastChangesRequestedAt` guard deliberately. Verified live: GitHub drops a review from `latestReviews` as soon as its author is re-requested, which nils the timestamp. Checked the other way round, a PR genuinely awaiting its reviewer reports `notApplicable` and the log says "no CR timestamp" instead of naming the real state.

## Also in here

Both noticed while touching the surrounding code, called out so they don't read as scope creep:

- `IssueTracker.withURL` silently dropped `mergeCommitOid` and `repoAutoMergeAllowed`. Inert today (its only caller is the GitLab path where both are nil), but a "copy with one field changed" helper that loses fields is a trap that worsens every time `PRRecord` grows. It now copies everything.
- The automation-boolean count in docs and comments said "six of these eleven default to on" when only five did. Adding this toggle makes "six of twelve" genuinely correct, so only the total changed.

## Verification

- `make test` — **0 failures**, 2,300+ tests across 18 packages.
- All three parity gates green (notification catalogs, CLI/RPC control-plane parity, no-NSLog). `docs/cli.md` regenerated via `make docs`.
- The modified GraphQL query was run against the live API to confirm it is valid and that the two new fields populate as expected.
- New coverage: `IssueTrackerAutoReReviewTests` (skip reasons, round-key re-arming, dispatch bookkeeping, the no-terminal escape hatch, log volume), `PRStatusChangesRequestedStateTests` including a cross-product sweep asserting the states partition so the two actions can never both fire, and backend fixtures pinning that a rebase restamp does not advance `lastSubstantiveCommitAt`.
- All ten existing `PRStatusNeedsRefineTests` pass unchanged.


dgershman (changes requested):
## Code & Security Review

Excellent design work — naming the third state and deriving both actions from one classifier is the right shape, and the `authoredDate` switch is well-argued with its residual gap honestly documented. Verification reproduces cleanly: `swift build` green, CrowEngine 612 / CrowCore 539 / CrowDaemon 327 / CrowProvider 58 tests pass, all three parity gates green.

Three findings, one blocking.

### Critical Issues

**🔴 `hasPendingReviewRequest` is PR-wide, so a second pending reviewer silently disables needs-refine** — `GitHubCodeBackend.swift:442`, `PRStatus.swift:188`

The GraphQL selection is `reviewRequests(first: 1) { totalCount }`, which counts *every* pending review request on the PR — not just requests to the reviewers whose latest review is `CHANGES_REQUESTED`. `changesRequestedState` then returns `.awaitingReviewer` on any pending request, ahead of the needs-refine branch.

Concrete case — a PR with two requested reviewers:

1. Reviewers A and B are requested. A submits `CHANGES_REQUESTED`; GitHub clears **only A's** request. B never reviews.
2. `reviewDecision == CHANGES_REQUESTED`, `latestReviews` still holds A's review (so `lastChangesRequestedAt` is populated), and `reviewRequests.totalCount == 1` (B).
3. → `.awaitingReviewer` → `needsRefine == false`, **and** the watcher skips with `.notAwaitingReRequest`.

The agent is never prompted to address A's findings and nothing re-requests: the same permanent dead-end CROW-921 exists to eliminate, reintroduced for multi-reviewer PRs. It is also a regression against shipped CROW-508 behaviour — pre-PR `needsRefine` had no pending-request term and would have fired here. CODEOWNERS-driven team requests hit the same path (a team request stays pending until a team member reviews).

`PRStatusNeedsRefineTests.swift:158` documents the gate as intentional, but its justification — *"Unreachable on GitHub today (it empties `latestReviews` when it re-requests, which nils the anchor)"* — only holds when the pending request and the CR review belong to the same person, i.e. a single-reviewer PR. The PR description's "Note on ordering" verification appears to have been done against that shape.

Suggested fix, no extra API call — make the field reviewer-specific:

```graphql
reviewRequests(first: 20) { nodes { requestedReviewer { ... on User { login } } } }
```

then set `hasPendingReviewRequest` from the intersection with `changesRequestedReviewerLogins`. That preserves the deliberate above-the-anchor ordering (a re-requested CR author still reads as `.awaitingReviewer` even once the anchor is nil) while leaving an unrelated pending reviewer out of the decision.

### Code Quality

**🟡 The failure path retries unbounded, with unrate-limited logging** — `IssueTracker.swift:3693-3722`

On error, `attemptReRequestReview` clears the round key so the next poll re-dispatches. Both the dispatch line (`:3666`) and the failure line (`:3720`) call `CrowLog.automation` directly rather than going through `logSteadyState`, so a permanently-failing PR emits **two** un-deduped lines every 60s poll — ~2,880/day/PR. That is precisely the flood the PR's own rate limiter was introduced to prevent.

Reproduced by the new suite itself: `swift test --package-path Packages/CrowEngine` emits 14 of these lines (7 `dispatched` + 7 `failed:no-code-backend`) for a single PR across repeated polls.

Two conditions are permanently failing, not transient:

- `codeBackend(for:) == nil` → `failed:no-code-backend`, retried forever. Unlike the adjacent `provider-unsupported` guard (`:3700`), which correctly latches the round key, this one clears it (`:3697`).
- `requestReviewers` throwing on a malformed login — the deliberate all-or-nothing throw at `GitHubCodeBackend.swift:369`. The login can't change without a new head or review, so the retry can never succeed.

`applyAutoRebase` already has the idiom to reuse: `autoRebaseDeferralBackoff` with capped exponential doublings (`IssueTracker.swift:3446`). At minimum, route the failure log through `logSteadyState` and latch the round key for `no-code-backend`. No test currently covers any failure path.

**🟡 The partition test's key assertion is vacuous** — `PRStatusNeedsRefineTests.swift:288`

```swift
#expect(!(state == .needsRefine && state == .awaitingReRequest))
```

`state` is a single enum value, so this holds for any implementation — including a broken one. The suite's docstring names this as "the invariant the whole design rests on", so it reads as proving something it cannot. The line above it (`needsRefine ⟺ state == .needsRefine`) is the meaningful assertion. Consider replacing the tautology with the invariant that actually has teeth: that `needsRefineGate` and `autoReReviewSkipReason` never both return "go" for the same `PRStatus`.

**🟢 Nits**

- `docs/automation.md:87` — "Both toggles read from `AutoRespondSettings`" now sits under a three-item list.
- `IssueTracker.swift:3673` — hardcoded `3600` where `Self.steadyStateLogHeartbeat` is the same value and is what the sibling rate-limiter uses.

### Security Review

**Strengths:**

- `requestReviewers` uses direct argv with no `sh -c` and a `$TMPDIR` cwd, matching `addMergeLabel` — nothing in `prURL` or a login reaches a shell.
- `isSafeReviewerLogin` (`GitHubCodeBackend.swift:381`) closes flag injection, which direct argv does *not* cover — a login of `--add-label` would otherwise be parsed by `gh` as an option. Stricter-than-GitHub validation with a hard reject is the right call, and the all-or-nothing throw correctly avoids recording a success that re-requested a short list. `testIsSafeReviewerLogin` covers the flag, space, slash, empty and length cases.
- Capability gating is correct end to end, and the `CodeBackend` default throws rather than silently no-oping.
- The default-on toggle is defensible: idempotent, reversible, and matches what the `addressChanges` prompt already instructed.

**Concerns:** none beyond the correctness findings above. No secrets, no injection surface, no new network trust.

### Also Verified

- `withURL` field-loss fix (`IssueTracker.swift:1207-1214`) is a real latent-bug fix, correctly restoring `mergeCommitOid` / `repoAutoMergeAllowed`.
- The `needsRefineGate` refactor is behaviour-preserving: `cooldownElapsed` (`:2195`) and `isManagedTerminalIdle` are both pure reads, so losing Swift's short-circuit costs nothing.
- Round-key pruning is guarded by the non-empty `viewerPRs` check, so a failed fetch can't wipe live state.
- Control-plane plumbing is complete and consistent: CLI flag, RPC patch, parity ledger entry, web toggle, `docs/cli.md` regenerated, CHANGELOG.

### Summary Table

**Recommendation:** Request Changes — driven by 1 Red, 2 Yellow, 2 Green findings. The Red is the one that matters: as written, this PR closes the single-reviewer dead-end while opening the same dead-end for multi-reviewer PRs, and regresses CROW-508 there. The reviewer-scoped `reviewRequests` selection should resolve it without changing the design.

---

dhilgaertner:
Thanks — the Red is correct and was a real dead-end, not a nitpick. All five findings addressed in 17c3423.

## 🔴 The multi-reviewer regression

Your diagnosis is exactly right, and so is the part I'd have missed on my own: **my justification was wrong, not just my code.** The test comment claimed the gate was "unreachable on GitHub today," and that only holds when the pending request and the CR review belong to the same person. I verified the ordering against precisely that single-reviewer shape and generalised from it.

The fix keeps the design and changes the signal. `PRStatus.changesRequestedReviewerIsPending` now derives from two lists rather than a count, and is true in two cases:

1. A blocking reviewer is explicitly pending (the intersection is non-empty).
2. Something is pending and **no** blocking reviewer is visible — the host hid the review because it re-requested its author.

Your A/B case matches neither: clause 1 is false (B isn't a blocker), clause 2 is false (A is visible). So the PR stays in `.needsRefine`, the agent gets prompted, and the loop keeps working.

I took clause 2 rather than the plain intersection you suggested because the plain version loses state 3. Once A is re-requested, `latestReviews` empties, so `changesRequestedReviewerLogins` is empty and the intersection is empty too — the PR would read `.notApplicable` instead of `.awaitingReviewer`. Behaviour is identical (neither action fires either way, the nil anchor already blocks both), but the gated log would name the wrong state, which is the whole reason that check sits above the anchor guard. Clause 2 keeps the ordering rationale intact instead of trading it away.

Query is `reviewRequests(first: 20) { totalCount nodes { ... on User { login } } }` — no extra call, as you suggested. Team requests carry no login and stay visible only through `totalCount`: folding a team slug into the login list would risk colliding with a real user, and the intersection is against review *authors*, who are always Users.

Re-ran the modified query live. Both shapes are present right now, which is a nice accident:

| PR | totalCount | pending | crAuthors | → |
|---|---|---|---|---|
| #1898 | 1 | `[dgershman]` | `[]` | clause 2 → `awaitingReviewer` |
| #930 | 0 | `[]` | `[dgershman]` | neither → falls through to the anchor |

## 🟡 Failure path

Both parts fixed, and thank you for counting the actual lines from my own suite's output — that's the kind of evidence that makes a finding impossible to argue with.

- `no-code-backend` now **latches** like the adjacent capability gate. It's terminal for the same reason: the provider is a property of the session and can't change under a live round.
- Retries capped at 3 per round (`maxAutoReReviewFailureRetries`), counted per round key and pruned with it. I went with a bounded count rather than `autoRebaseDeferralBackoff` because the inputs to a re-request are *fixed* for the life of a round — same PR, same logins — so backoff would only slow down an outcome that can't change. A new push or reviewer submission re-arms it, which is the natural retry boundary.
- Dispatch and failure lines now go through `logSteadyState`. The dispatch line gained the head sha so a genuinely new round still logs while a retry within one round dedupes.

Failure-path tests added — you were right that there were none.

## 🟡 The vacuous assertion

Guilty, and worse than it looks: CrowCore *cannot* express that invariant, because only the classifier's single enum value is in scope there. Any check written at that layer is a tautology.

Moved the real one to `IssueTrackerAutoReReviewTests`, where both gate functions are visible — a cross-product sweep asserting `needsRefineGate` and `autoReReviewSkipReason` never both return "go". It also counts how often each arm opens and asserts both are non-zero, so it can't pass vacuously either. The CrowCore sweep keeps the assertion that does have teeth (`needsRefine ⟺ state == .needsRefine`) and now also verifies it reached all four states.

## 🟢 Nits

Both applied — the "Both toggles" wording, and `Self.steadyStateLogHeartbeat` in place of the literal.

## Unrelated flake, reported rather than absorbed

`ManagerMigrationTests.hydratePreservesManagerTmuxBinding` fails intermittently under the parallel suite. I measured it before claiming it wasn't mine: **1/8 on the pre-change baseline, 2/8 on this branch** — same test, comparable rate, passes 5/5 in isolation. It asserts on manager command rebuild and `remoteControlActiveTerminals`, i.e. SessionService hydration, which nothing here touches. Out of scope for this PR, but it will randomly redden CI and probably deserves its own ticket.

## Verification

Full `make test` green — 0 failures across 25 runs — plus all three parity gates.

dgershman (changes requested):
## Code & Security Review

Re-review of `17c3423`. All five findings from the previous round are genuinely addressed, and two of the responses are better than what I suggested:

- **The Red.** `changesRequestedReviewerIsPending` (`PRStatus.swift:171`) derives from two lists instead of a count, and clause 2 (`anyPendingRequest && changesRequestedReviewers.isEmpty`) preserves the above-the-anchor ordering that a plain intersection would have traded away. I worked the cross-product: A-blocks-visible + B-pending → falls through to the anchor ✅; A-re-requested-hidden + B-pending → `.awaitingReviewer` ✅; A hidden **and** B still blocking-and-visible → falls through and re-requests **only B**, which is the correct subset ✅; CODEOWNERS team pending alongside a visible blocker → falls through ✅. Keeping Teams out of the login list and visible only through `totalCount` is right — the intersection is against review *authors*, who are always Users.
- **The failure path.** Reproduced the count: `swift test --package-path Packages/CrowEngine` now emits **1** `auto-re-request` line where it emitted 14. Latching `no-code-backend` and capping retries at 3 per round is the right call over backoff, for the reason given — the inputs are fixed for the life of a round.
- **The vacuous assertion.** `theTwoActionsNeverBothFireForTheSamePR` (`IssueTrackerAutoReReviewTests.swift:181`) is a real sweep over both gate functions, and the `refineFired > 0 / reRequestFired > 0` counters mean it can't pass vacuously either. Correct diagnosis that CrowCore structurally cannot host it.

Verification reproduces: `swift build` green; CrowEngine 618 / CrowCore 544 / CrowProvider 58 / CrowDaemon 327 tests pass; all three parity gates green (11 notification events, 94 CLI/RPC methods, no-NSLog).

One new finding, blocking. It is in the same family as the `withURL` fix this PR ships — and, I think, the more consequential half of it.

### Critical Issues

**🔴 `mergePRRecords` silently drops all three new `PRRecord` fields** — `IssueTracker.swift:924-952`

`withURL` was fixed in this PR because "a helper that silently loses fields is a trap that gets worse every time `PRRecord` grows" (`:1195`). `mergePRRecords` is the sibling helper ~270 lines up, it constructs `PRRecord(...)` from an explicit argument list, and it does not carry `changesRequestedReviewerLogins`, `pendingReviewerLogins`, or `hasPendingReviewRequest`. All three have defaults (`[]`, `[]`, `false`), so the omission compiles silently.

Confirmed empirically with a throwaway probe against this branch (viewer record carrying `crAuthors=["dgershman"]`, `pending=["someoneelse"]`, `anyPending=true`; stale record with the defaults):

```
PROBE crAuthors=[] pending=[] anyPending=false      ← mergePRRecords
PROBE reviewerIsPending=false                        ← buildPRStatus of the merged record
PROBE dedupedCrAuthors=[]                            ← dedupedByURL([viewer, stale])
```

Both halves of the loop are affected, in opposite directions:

1. `changesRequestedReviewerLogins == []` → `autoReReviewSkipReason` returns `.noReviewers` (`:3594`) → **the PR is never re-requested.** That is the CROW-921 dead-end this PR exists to close.
2. `changesRequestedReviewerIsPending == false` → a PR that is genuinely `.awaitingReviewer` classifies as `.needsRefine` instead → a spurious `addressChanges` prompt while the ball is already with the reviewer.

**Reachability.** The merge site is `dedupedByURL(ghResult.viewerPRs + stalePRs)` (`:692`). Stale candidates are excluded from the open set by **exact string** compare (`collectStalePRURLs`, `:1082`), but `prStates` returns the host's **canonical** URL — a divergence the code explicitly anticipates two hundred lines later:

> `// Keying by PRRef means we don't lose records when the API returns a canonical URL different from the stored one.` (`:1150-1153`)

So any session whose `.pr` link URL is not byte-identical to GitHub's canonical form produces two records under the canonical URL and trips the merge. Owner/repo are case-insensitive on GitHub, and links also arrive from `crow add-link --url` and the workspace skill, not only from the viewer payload; a repo rename redirect does it too. This is not a one-poll blip — the exclusion compare fails on *every* poll for that link, so the feature is silently off for the life of that session.

Diagnosability is partially preserved (`skipped:no-changes-requested-reviewers` does get logged, rate-limited), which is a real mitigation — but the symptom points away from the cause.

This is also the exact bug of shipped **#838**, in this same function: `mergePRRecords` dropped `labels` and blinded the auto-merge watcher. Its regression test — `IssueTrackerDedupTests.swift:223`, `dedupedByURLUnionsLabelsAcrossViewerAndStaleRecords`, an "end-to-end assembly guard" for precisely the viewer+stale overlap — is three fields short of catching this one.

Fix is three lines, matching the adjacent idioms:

```swift
changesRequestedReviewerLogins: winner.changesRequestedReviewerLogins.isEmpty
    ? loser.changesRequestedReviewerLogins : winner.changesRequestedReviewerLogins,
pendingReviewerLogins: winner.pendingReviewerLogins.isEmpty
    ? loser.pendingReviewerLogins : winner.pendingReviewerLogins,
// `false` means "not fetched" (stale query / GitLab), so OR rather than pick.
hasPendingReviewRequest: winner.hasPendingReviewRequest || loser.hasPendingReviewRequest,
```

Worth extending the `:223` test with the reviewer fields so the guard covers the whole record rather than one field at a time.

### Code Quality

**🟢 Nits**

- `CHANGELOG.md:21` still documents the gated line as `…, pendingRequest=…, idle=…`. The code emits `reviewerReRequested=` (`IssueTracker.swift:2157`) and `docs/automation.md:108` was updated; the CHANGELOG wasn't. These strings are explicitly grep-stable, so a stale one in the release notes is a wrong instruction to a reader.
- `IssueTracker.swift:3679` logs `auto-re-request: dispatched #N → …` *before* `attemptReRequestReview` runs, so the two terminal skips (`no-code-backend`, `provider-unsupported`) always follow a line claiming a dispatch that never happened. Harmless, but it makes `grep dispatched` an overcount.

### Security Review

**Strengths:**

- `requestReviewers` (`GitHubCodeBackend.swift:369`) uses direct argv with a `$TMPDIR` cwd and no `sh -c`; `isSafeReviewerLogin` (`:391`) closes the flag-injection gap that direct argv does *not* cover, by requiring an ASCII-alnum first character — so `--add-label` is rejected rather than parsed as an option. All-or-nothing throw avoids recording a success that re-requested a short list.
- `prURL` is not login-validated, but it is unreachable as a flag: a record only exists here if it came back from the host API or parsed as a PR URL via `parseTicketURLComponents`. Same shape as the existing `addMergeLabel` path.
- `IssueTracker` is `@MainActor` (`:17`), so the `Task { await self.attemptReRequestReview(…) }` mutations of `autoReReviewFailureCounts` / `autoReRequestAttempted` hop back to the actor — no race on the new bookkeeping.
- Capability gating is end to end, and the `CodeBackend` default throws rather than silently no-oping. Default-on is defensible: idempotent, reversible, and matches what the `addressChanges` prompt already instructed.

**Concerns:** none beyond the correctness finding above. No secrets, no injection surface, no new network trust.

### Also Verified

- Round-key pruning stays guarded by the non-empty `viewerPRs` check, and `autoReReviewFailureCounts` is pruned alongside it (`:3635`) — a degraded fetch can't reset the retry budget.
- Retry logging is bounded: dispatch and failure share one `(channel, url)` dedupe slot and alternate, so a permanently-failing round costs ≤6 lines total, then latches.
- Control-plane plumbing is complete and consistent: CLI flag, RPC patch, `ParityLedger` entry, web toggle, daemon provider wiring, `SettingsRPCSupport`, `docs/cli.md` regenerated, and every "eleven → twelve" count updated across four docs.
- The `PRStatus` coding-key rename (`hasPendingReviewRequest` → `changesRequestedReviewerIsPending`) decodes absent-as-`false`, and the old key never shipped on `main`, so there is no persisted state to migrate.

### Summary Table

**Recommendation:** Request Changes — driven by 1 Red, 0 Yellow, 2 Green findings. The design is right and the previous round's fixes are solid; the remaining Red is a three-line omission in a merge helper that happens to defeat the exact feature this PR adds, in the same function and the same way as shipped #838.

---

dhilgaertner:
Caught cleanly — and the sting is that I fixed `withURL`'s sibling *in this very PR* for exactly this reason, then walked past `mergePRRecords` 280 lines up. All three findings addressed in 6099294.

## 🔴 `mergePRRecords`

Confirmed by reading, and then confirmed properly: I reverted just the three merge lines and re-ran the new tests. They fail on every field, reproducing your probe output exactly —

```
crAuthors=[] pending=[] anyPending=false
```

— so the guard has teeth rather than passing because the code happens to work today.

Fixed with the adjacent idioms, including the `||` for the boolean: `false` means "not fetched" on every path that doesn't select `reviewRequests`, so picking a side could only lose information. Your reachability analysis is the part I'd have underweighted — I'd have filed this as a rare mid-refresh overlap. The exact-string-vs-canonical-URL divergence means it repeats on *every* poll for an affected session, so the feature is silently off for that session's whole life, not for one cycle. That's a different severity.

Two things beyond the three lines:

- Renamed `dedupedByURLUnionsLabelsAcrossViewerAndStaleRecords` → `…UnionsLabelsAndReviewerFields` and made it assert the whole record. Your framing — that #838's guard was "three fields short" — is really a statement that a per-field guard was always going to lose this race. Added the loser-side backfill case (the direction that actually broke, since the stale record wins the state rank) and the reverse guard that backfill can't clobber a populated winner.
- Put the "every field must be carried" rule in `mergePRRecords`' doc comment, naming both shipped bugs, so the next person adding a `PRRecord` field reads the reason at the point of change rather than in a changelog.

I also audited the other three `PRRecord(...)` sites: `parsePRNode` is the source and populates them; the two GitLab ones legitimately have no such data, which is what the field docs already say `false`/empty means there. So the two copy/merge helpers were the whole exposure, and both are now closed.

## 🟢 Nits

Both real, both fixed.

- The CHANGELOG string. You're right that this one isn't cosmetic: those strings are grep-stable *by contract*, so a stale one in the release notes is a wrong instruction to someone debugging at 2am.
- The `dispatched` line moved into `attemptReRequestReview` after both terminal guards. It no longer stands in front of a skip that contradicts it, and `grep dispatched` now counts attempts that actually reached the host.

## Verification

Full `make test` green — 0 failures — plus all three parity gates. Dedup suite is 15 tests, up from 12.

Thanks for working the cross-product on clause 2 rather than taking my word for it — particularly the `A hidden and B still blocking-and-visible → re-requests only B` case, which I hadn't enumerated explicitly and which is the one where "correct subset" is doing real work.

dgershman (approved):
## Code & Security Review

Re-review of `6099294`. The round-2 Red is fixed correctly, and both Green nits are applied. No new blocking findings — approving.

Verification reproduces: `swift build` green; CrowEngine **620** / CrowCore **544** / CrowProvider **58** / CrowDaemon **327** tests pass (CrowEngine +2 for the new dedup guards); all three parity gates green (11 notification events, 94 CLI/RPC methods, no-NSLog).

### The Red is fixed

`mergePRRecords` (`IssueTracker.swift:954-969`) now carries all three fields, with the idioms the surrounding lines already use — empty-means-not-fetched backfill for the two lists, and **OR** for the boolean, correctly reasoned: `false` also means "not fetched" on every path that doesn't select `reviewRequests`, so a `true` from either side is the only informative value. Picking a side would have re-lost it whenever the stale record won the state rank.

The guard is real this time, not assumed. Three tests cover the three directions that matter:

- `dedupedByURLUnionsLabelsAndReviewerFields` (`:229`) — the end-to-end assembly path, now asserting the whole record rather than `labels` alone, which is what let #838's guard sit three fields short of catching this.
- `mergePRRecordsCarriesReviewerFieldsFromEitherSide` (`:264`) — **loser-side** backfill in both argument orders. This is the direction that actually broke (the stale record wins on `MERGED > OPEN`), and the one a winner-only test would have missed.
- `aPopulatedWinnerIsNotOverwrittenByAnEmptyLoser` (`:286`) — the reverse guard, so the fix can't trade one silent-loss bug for another.

Moving the rule into `mergePRRecords`' doc comment (`:925-934`) and naming both shipped bugs at the point of change is the right durable fix — the next person adding a field reads the reason where they're working, not in a PR description.

### Both Greens applied

- **CHANGELOG** now says `reviewerReRequested=` — matching what the code emits (`:2157`). These strings are explicitly grep-stable, so the release notes no longer hand a reader a string that isn't in the log.
- **The `dispatched` line** moved into `attemptReRequestReview` (`:3767`) *after* both terminal guards, so it no longer stands in front of a skip that contradicts it and `grep dispatched` counts attempts that reached the host.

### Code Quality

**🟢 `updatedAt` is the one field `mergePRRecords` still drops** — `IssueTracker.swift:938-976`

Mechanically diffing the call's argument list against `PRRecord.init` leaves exactly one parameter unpassed:

```
$ comm -13 <merge args> <init params>
updatedAt
```

Not a regression and not reachable: it is absent on `main` too, and the sole production call site (`:692`) feeds `applyPRStatuses` / `updatePRAttributions`, neither of which reads it — `buildPRStatus` doesn't project it, and the reconcile tie-break that does consume it builds `ReconcileBranchMatch` from the separate `prsForKeys` / `prsForBranches` results. So this is inert exactly the way `withURL`'s dropped `mergeCommitOid` was inert before this PR fixed it.

Raising it only because the new doc comment directly above now reads "**Every field must be carried here.**" and closes with "extend it when `PRRecord` grows" — a comment written specifically to stop the next person from doing what the function still does, one line from being true. Note `withURL` (`:1244`) *does* carry it, so the two sibling helpers disagree. Worth the line, and worth a `#expect(merged.updatedAt != nil)` in the assembly guard.

**🟢 `theRetryBudgetIsBounded` asserts the constant, not the behaviour** — `IssueTrackerAutoReReviewTests.swift:411-418`

`#expect(IssueTracker.maxAutoReReviewFailureRetries == 3)` pins the value but never exercises the loop it bounds. The counter path — `requestReviewers` throwing, `failures` incrementing, the round key latching on the third — has no behavioural coverage: `aPermanentlyUnservableSessionIsNotRetriedEveryPoll` (`:385`) covers the `no-code-backend` latch, which returns *before* the `do`/`catch` and so never touches `autoReReviewFailureCounts`. The logic reads correctly (`:3780-3791`) and the bookkeeping is pruned (`:3659`, covered at `:372`), so this is coverage rather than doubt — a throwing stub backend would close it.

### Also Verified

- **The round-2 log flood is gone.** The four `auto-re-request` lines in a full CrowEngine run are one apiece from four independent test instances, not the 7-per-PR repeat the finding was about; the terminal conditions latch (`:3737-3756`) so a permanently-failing round costs a bounded number of lines rather than ~2,880/day.
- **The invariant test has teeth.** `theTwoActionsNeverBothFireForTheSamePR` (`:423`) sweeps 2,048 combinations through both real gate functions and asserts `refineFired > 0` / `reRequestFired > 0` / `bothQuiet > 0`, so it cannot pass vacuously — a genuine replacement for the round-1 tautology.
- **The multi-reviewer Red stays fixed.** `changesRequestedReviewerIsPending` (`PRStatus.swift:171`) still derives from the two lists, with clause 2 preserving the above-the-anchor ordering; `anUnrelatedPendingReviewerDoesNotSilenceTheLoop` (`:486`) pins it.
- **GraphQL parsing is sound.** Team requests carry no `login`, so `compactMap` drops them from `pendingReviewerLogins` while `totalCount` still answers "anyone at all" — which is exactly the split the field docs describe. A `>20`-pending truncation could only under-report clause 1, and re-requesting an already-pending reviewer is a host-side no-op.
- **`agentSettled` escape hatch** (`:2216`) treats a session with no `.agentLaunched` managed terminal as settled, so a PR whose session never launched an agent still gets rescued.
- **Doc counts are consistent.** "six of … twelve" now agrees across `CLAUDE.md:284`, `docs/cli-reference.md:725`, `SettingsRPCSupport.swift:103`, `ConfigStoreTests.swift:326`, and `AutomationHandlerTests.swift:95`.

### Security Review

**Strengths:**

- `requestReviewers` (`GitHubCodeBackend.swift:369`) uses direct argv with a `$TMPDIR` cwd and no `sh -c`; `isSafeReviewerLogin` (`:391`) closes the flag-injection gap direct argv does *not* cover by requiring an ASCII-alnum first character, so `--add-label` is rejected rather than parsed as an option. The all-or-nothing throw avoids recording a success that re-requested a short list.
- `prURL` is not login-validated but is unreachable as a flag — a record only exists here if it came back from the host API or parsed via `parseTicketURLComponents`. Same shape as the existing `addMergeLabel` path.
- `IssueTracker` is `@MainActor`, so the `Task { await self.attemptReRequestReview(…) }` mutations of `autoReReviewFailureCounts` / `autoReRequestAttempted` hop back to the actor — no race on the retry bookkeeping.
- Round-key and failure-count pruning stay behind the non-empty `viewerPRs` check, so a degraded fetch can't wipe live state or reset the retry budget.
- Capability gating is end to end, and the `CodeBackend` default throws rather than silently no-oping. Default-on remains defensible: idempotent, reversible, and matching what the `addressChanges` prompt already instructed.

**Concerns:** none. No secrets, no injection surface, no new network trust.

### Summary Table

**Recommendation:** Approve — driven by 0 Red, 0 Yellow, 2 Green findings. Three rounds in, the design has held up unchanged while every finding against it was a field or a gate quietly going missing at an assembly seam; the response this round fixed the seam and then documented the rule at it. Both Greens are optional and neither is a regression — the `updatedAt` line predates this PR and reaches no consumer.

---

PR #932, opened 2026-07-31
Stop chained session-menu actions from timing out client-side (CROW-931)

### LINKED ISSUE #931: Chained session-menu actions time out client-side: serialized /rpc socket + a full board poll inside add-merge-label + a flat 10s timeout
Chaining several session-menu actions from the sidebar right-click menu (set
`crow:merge` on a few PRs, mark several In Review, …) produces a cascade of
`rpc timeout: <method>` alert modals.

**The actions actually succeed.** Only the client gives up. The user is told N
operations failed when every one of them completed server-side — and the natural
response, retrying, makes the underlying problem worse.

## Cause

Three things compound. None is a bug on its own.

**1. One WebSocket, strictly one RPC in flight.**
`RPCWebSocketHandler.swift:71-93` reads messages in a single loop and `await`s
the dispatch **inline** before reading the next one:

```swift
for try await message in inbound.messages(maxSize: 1 << 20) {
    …
    response = await commandRouter.handle(request: request)   // :88 — blocks the read loop
    …
}
```

The whole browser shares one `/rpc` socket (`rpcState.ready`, `app.js:81`), so
every queued request waits for all of its predecessors. Head-of-line blocking.

**2. `add-merge-label` is expensive — dominated by a full board poll.**
`IssueTracker.addMergeLabel` (`IssueTracker.swift:4077`) does:

| Step | Measured |
|---|---|
| `ensureMergeLabel` → `gh label create` (`GitHubCodeBackend.swift:53`) | ~0.4 s |
| `addMergeLabel` → `gh pr edit --add-label` (`GitHubCodeBackend.swift:315`) | ~0.5 s |
| **`await refresh()` — full board poll** (`IssueTracker.swift:4120`) | **4.2–5.0 s** |

`refresh()` (`IssueTracker.swift:475`) is the complete board read: consolidated
GitHub GraphQL, every GitLab host, Jira REST (may shell `op read`), Corveil CLI.
Timed on this machine via `crow refresh-tickets`: **4.22 / 4.97 / 4.49 s**,
against ~0.02 s for a trivial RPC like `list-sessions`.

Total ≈ **5–6 s per `add-merge-label`**.

**3. A flat 10 s client timeout, started at send time.**
`app.js:87`:
```js
setTimeout(() => { if (rpcState.pending.delete(id)) reject(new Error('rpc timeout: ' + method)); }, 10000);
```
`sessionAction` (`app.js:2025`) is fire-and-forget from each menu item, so every
click starts its own timer immediately. The timer knows nothing about queue
position.

### The arithmetic

Action *k* completes at ≈ *k* × 6 s:

| Action | Completes | 10 s timer |
|---|---|---|
| 1 | ~6 s | ✅ |
| 2 | ~12 s | ❌ timeout |
| 3 | ~18 s | ❌ timeout |
| 4 | ~24 s | ❌ timeout |

The **second** chained action already fails. Hence the cascade of popups.

### Why the failure is silent-but-wrong

When the real response finally arrives, its id is gone from `rpcState.pending`,
so `app.js:43-44` drops it on the floor:

```js
const waiter = rpcState.pending.get(msg.id);
if (!waiter) return;
```

No success path, no correction to the already-shown modal.

Note the CLI does **not** have this problem — it passes `timeoutSeconds: 60` for
exactly these two verbs (`SessionLifecycleCommands.swift:95,123`). The web client
hardcodes 10 s for every method. The knowledge that these are slow already exists
in the codebase; it just never reached the browser.

## Secondary finding: N actions ⇒ N full board polls

Each chained `add-merge-label` runs its own complete `refresh()`. The
`guard !isRefreshing else { return }` at `IssueTracker.swift:476` does not help,
because the serialized queue means they are never concurrent — they are
back-to-back. Ten actions ⇒ ten consolidated GitHub GraphQL queries in a minute,
purely to recompute a warning string. That is gratuitous rate-limit pressure on
top of the latency.

## Proposed fix

In rough order of payoff:

**A. Stop doing a full board refresh inside a per-session mutation.**
`refresh()` is called at `IssueTracker.swift:4120` only so `autoMergeWarning`
(`:4138`) can read a freshly populated `autoMergeState`. Options: fetch just that
one PR; or coalesce/debounce the refresh across a burst; or return without the
warning and let the existing `changed` push deliver it. This alone takes the
action from ~6 s to ~1 s and largely dissolves the problem.

**B. Stop head-of-line blocking the reader loop.**
Dispatch each request into a child task instead of awaiting inline, with a
bounded concurrency cap so a burst can't spawn unbounded work. Safe as-is:
responses already funnel through the single writer via `outCont.yield`, and
JSON-RPC ids mean out-of-order responses are correct by construction — the client
already maps by id at `app.js:43`. Worth pairing with a **per-session** serializer
so two writes to the same session keep their order.

**C. Per-method client timeouts.**
Mirror the CLI's map rather than a blanket 10 s — `add-merge-label` and
`mark-issue-done` get 60 s, trivial reads keep something short.

**D. A batch verb, following the precedent already in the repo.**
CROW-865 solved exactly this shape for the Reviews board: `batch-start-review`
takes every selected URL in one call, queues on the daemon's serializer, and acks
immediately, with results arriving via the poll (`app.js:3726-3730`). A
`batch-session-action` for the sidebar menu is the same pattern and the real
answer for "I selected twelve sessions and set them all In Review."

**E. Minor.** `ensureMergeLabel` unconditionally shells `gh label create` and
catches "already exists" as control flow — a guaranteed wasted round trip per
call. Cache per repo.

## Acceptance

- [ ] Chaining ≥5 session-menu actions produces no `rpc timeout` modal.
- [ ] A slow RPC on the `/rpc` socket no longer delays unrelated RPCs behind it.
- [ ] Ordering is preserved for multiple writes to the same session.
- [ ] A single `add-merge-label` no longer triggers a full multi-provider board poll.
- [ ] A late response can never leave a spurious failure modal on screen.
- [ ] Web client timeouts match the CLI's per-method budgets.

## Reproduction

Right-click several sessions in the sidebar in quick succession, choosing
"Add label crow:merge to PR" (or "Mark as In Review") each time. From the second
action onward, each raises `rpc timeout: add-merge-label` — then confirm on
GitHub that every label was in fact applied.

dhilgaertner:
Closes #931

Chaining session-menu actions from the sidebar produced a cascade of `rpc timeout: <method>` modals. **The actions all succeeded** — only the client gave up, and the natural response (retry) made it worse.

I verified every line number in the ticket against `ca07f1b` before building on it: **zero drift**. Taking **A, B, C, E**; skipping **D** (see below).

## A — `add-merge-label` no longer runs a full board poll

`IssueTracker.addMergeLabel` ended with `await refresh()` — the complete multi-provider read (GitHub GraphQL, every GitLab host, Jira REST which may shell `op read`, Corveil CLI), 4–5s, run *solely* so `autoMergeWarning` could read a freshly populated `autoMergeState`. It now does a targeted per-PR fetch through the existing `fetchStalePRStates` and evaluates that one session.

**That call was also a live correctness bug, not just latency.** `refresh()` opens with `guard !isRefreshing` and `guard shouldPoll()`, so a scheduled poll already in flight made it a silent no-op — and the `autoMergeWarning` read below it then reported the *previous* poll's verdict, or nothing at all on a freshly linked PR, while claiming to have re-read. The new path has no such guard: it always fetches.

To keep **one** code path rather than a second hand-rolled copy that drifts, `applyAutoMerge`'s per-session loop body is lifted verbatim into `evaluateAutoMerge(session:byURL:)` and its watcher-off branch into `publishWatcherOffVerdict(session:byURL:)`. `applyAutoMerge` is now a thin aggregator over both, preserving its exact per-poll summary line. **The existing 221 `IssueTracker` tests pass unchanged** — that is the evidence the extraction was behaviour-preserving, and it was landed and verified before the targeted tail was added.

A read-your-write union (new `withLabels` helper, sibling to the existing `withURL`) adds the label to the fetched record when the provider's read side lags the write we just made. Without it the targeted pass reads "no merge label", publishes nothing, dispatches nothing, and the user gets a bare success for a label the watcher won't look at until the next poll — #888's exact shape. `withLabels` preserves *every* field, unlike `withURL`, because `repoAutoMergeAllowed` decides a branch in the evaluation.

## B — the `/rpc` reader loop no longer blocks head-of-line

`RPCWebSocketHandler.swift:88` awaited `commandRouter.handle(request:)` **inline** inside the read loop, so one slow method stalled every request queued behind it on the browser's single shared socket. The reader now hands each request to a per-connection `RPCDispatcher` and reads the next frame immediately.

**`RPCDispatcher`** (new) — an actor with three invariants:
- **Lane ordering** via a `Task` chain per lane, generalizing `ReviewKickoffSerializer`'s single global chain. Tails are evicted on completion, so the map is bounded by *live* lanes, not by every session id a long-lived connection ever addressed.
- **Bounded work** — permits gate execution, an admission ceiling gates accepted work. **The permit is taken *after* the lane wait, never before**, so the classic deadlock (every permit held by a task waiting on a predecessor that can never get one) is structurally impossible.
- **Explicit lifecycle** — `drain()` returns only once every accepted operation has yielded its response; `shutdown()` releases every waiter so an abnormal close never parks the reader.

**`RPCLanePolicy`** (new) — states each method's lane rather than inferring one. A "does it carry a `session_id`" heuristic would silently hand full concurrency to `launch-agent` (keys on `terminal_id`) and `job-run` (`job_id`). It's gated against `ParityLedger.rpcMethods`' write half, so **a new write RPC fails the build until someone decides its lane**. Three non-session families need one: the `config.json` set (ordering, but the load-bearing reason is that `ConfigStore.withConfigLock` is an `NSLock` held across disk I/O — the #874 shape), the Manager surface, and the review kickoffs (permit conservation, not ordering: `start-review` awaits its serializer task, so unlaned kickoffs would reproduce #931 one level down).

Three things worth a careful look in review:
1. **`Task {}`, not `group.addTask`** — `_ = try? await group.next()` returns on the *first* child to finish, and that return *is* the teardown trigger; adding dispatch tasks to that group would make the first completed RPC look like a socket close.
2. **`shutdown()` must precede `group.cancelAll()`** — `Task<Void, Never>.value` and `withCheckedContinuation` don't resume on cancellation, so cancelling first would leave the reader suspended in `drain()` and leak a connection per disconnect. The most fragile line in the diff.
3. **The hoisted `JSONEncoder` had to go** — it isn't `Sendable` and dispatched operations now encode concurrently. Encoding moved into a per-call `emit` helper. The `JSONDecoder` stays hoisted; it's still single-threaded on the reader task.

**Risk note:** MainActor reentrancy here is not new. `SocketServer` already dispatches each CLI connection to the global concurrent queue, and every browser tab is its own upgrade — the router has always been reentrant in production. What changes is that *one* client can now trigger it. Cross-lane ordering is gone and cannot be recovered (`set-config` then `set-status` can land in either order); every `app.js` call site awaits its response before issuing a dependent one.

## C — per-method client timeouts, and no spurious failure modal

`app.js` hardcoded 10s for every method while the CLI already passed `timeoutSeconds: 60` for these exact verbs. The knowledge existed in the repo; it just never reached the browser. The web now mirrors the CLI's table (30s default). **Keys are wire method names, not CLI verbs** — `crow job run` sends `job-run`, Settings' "Run now" sends `run-job`, so both are listed.

A timeout no longer *deletes* the pending entry — which is why the late response vanished at `onmessage`'s `!waiter` guard and the "failed" modal could never be taken back down. It now flags the entry `settled`, rejects with a tagged `err.rpcTimeout`, and routes the eventual reply to an `onLate` hook (with a GC sweep so the map stays bounded).

`sessionAction` shows an honest **"Still running"** advisory rather than "failed": a timeout means *we stopped waiting*, not that the daemon stopped working, and saying otherwise invites a retry that duplicates a write already in flight. Its `onLate` dismisses that advisory on late success, surfaces an additive `warning` (#888) if one comes back, or replaces it with the real error on late failure. A small `dismissModalDialog(token)` helper scopes the retraction by token, so it can never yank a dialog the user opened afterwards.

## E — `ensureMergeLabel` memoized

It unconditionally shelled `gh label create` and caught "already exists" as control flow — a guaranteed wasted round trip per call. `GitHubCodeBackend` is a struct rebuilt by every `ProviderManager.codeBackend(for:)` call, so the memo lives on the long-lived `@MainActor IssueTracker` that both callers already route through. Populated **only on success**, so a repo the token couldn't reach retries rather than latching a failure as done (CROW-816). The throwing form is the primitive and the watcher's best-effort behaviour is a `do`/`catch` on top — not the other way round — so `addMergeLabel` still reports a real label-creation failure.

## Deliberately out of scope

- **Fix D** (`batch-session-action`, following the CROW-865 `batch-start-review` precedent). A+B+C already make chained actions work; D answers "I selected twelve sessions", which needs sidebar multi-select the UI doesn't have yet. Worth its own ticket with its own UI.
- **`run-setup`** stays on the default timeout. It has a separate pre-existing bug — the daemon re-execs, `ws.onclose` rejects, and `completeSetup` aborts into `state.error` before its reconnect loop runs. Not #931; papering over it with a longer timeout would hide it.
- **`start-review`** (`RPCHandlers.swift:837`) awaits its serializer task where `batch-start-review` doesn't — same exposure class, out of scope here.

## Acceptance

| | |
|---|---|
| Chaining ≥5 session-menu actions produces no `rpc timeout` modal | A (~6s → ~1s) + B + C |
| A slow RPC no longer delays unrelated RPCs behind it | B |
| Ordering preserved for multiple writes to the same session | B (`RPCLanePolicy` + lane chains) |
| A single `add-merge-label` no longer triggers a full board poll | A |
| A late response can never leave a spurious failure modal | C |
| Web client timeouts match the CLI's per-method budgets | C |

## Testing

- `make build` and `make test` — pass, **0 failures**, CLI/RPC parity gates green (94 methods).
- New Swift suites: `RPCDispatcherTests` (12) and `RPCLanePolicyTests` (10) — including the ledger gate, the deadlock case (a lane burst deeper than the concurrency cap), lane-tail eviction, and `drain()`/`shutdown()` lifecycle.
- `IssueTrackerTargetedAutoMergeTests` (7) — the targeted path reaches the same verdict as the poll path, the read-your-write union works, and a failed targeted fetch stalls rather than raising a false alarm.
- `web-tests/rpc-timeout.test.js` (47) — drives the **real** `app.js` in jsdom against a fake socket and a controllable timer queue. Firing deadlines *by delay* is what proves each method was armed from the table rather than the default.
- `npm test` in `web-tests`: all suites pass except `row.test.js` (43 passed, 10 failed) — **pre-existing on `main`, verified identical on a clean tree**, unrelated to this change.

**Not verified live:** timing `crow add-merge-label` end-to-end and the browser repro would require restarting the daemon that hosts this session. Notably, `crow refresh-tickets` against the running daemon returned in 0.03s — the `isRefreshing`/`shouldPoll` early-return path, which is itself the silent no-op that fix A removes from this code path.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/RPCDispatcher.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/RPCDispatcher.swift
@@ -1,0 +1,259 @@
+import Foundation
+
+/// Per-connection dispatcher for `/rpc` requests (#931).
+///
+/// The reader task used to `await commandRouter.handle(request:)` inline, so one
+/// slow method held the socket's read loop and everything queued behind it hit
+/// the browser's timeout — a `refresh-tickets` behind a `gh` call stalled the
+/// unrelated `set-status` the user clicked next. The reader now hands each
+/// request here and immediately reads the next frame.
+///
+/// Three invariants, in the order they matter:
+///
+/// 1. **Lane ordering.** Requests in the same ``Lane`` run one at a time in
+///    arrival order, via a `Task` chain per lane — the same shape as
+///    ``ReviewKickoffSerializer``, whose single global chain this generalizes to
+///    one chain per lane. `set-status` then `complete-session` on one session
+///    still cannot reorder.
+/// 2. **Bounded work.** At most `maxConcurrent` operations *execute* at once,
+///    and at most `maxInFlight` are *accepted* before the dispatcher refuses.
+///    A queued lane task waits for its predecessor **before** taking a permit,
+///    never while holding one — see ``run(id:lane:predecessor:operation:)``.
+/// 3. **Drain before finish.** ``drain()`` returns only once every accepted
+///    operation has yielded its response, so the reader can end the writer's
+///    stream without truncating a reply. ``shutdown()`` releases every waiter so
+///    an abnormal close never parks the reader.
+///
+/// One instance per connection, created in `onUpgrade`. Nothing is shared
+/// between connections — the daemon already served those concurrently
+/// (`SocketServer` dispatches each CLI connection to the global concurrent
+/// queue, and every browser tab is its own upgrade), so this change makes
+/// intra-connection concurrency match the inter-connection concurrency that
+/// already existed rather than introducing a new kind.
+actor RPCDispatcher {
+
+    /// A serialization lane. Same lane → strict arrival order. Different lanes,
+    /// and lane-less requests, run concurrently.
+    ///
+    /// Ordering is **not** preserved *across* lanes: a `set-config` and a
+    /// `set-status` sent back to back can land in either order. That is the
+    /// deliberate cost of the fix, and it matches what two open browser tabs
+    /// could already produce. Every `app.js` call site awaits its response
+    /// before issuing a dependent one.
+    enum Lane: Hashable, Sendable {
+        /// Serialize on one request parameter's value — `session_id`,
+        /// `terminal_id`, `job_id`. The parameter *name* is part of the key so a
+        /// session UUID and a terminal UUID can never collide into one lane.
+        case param(name: String, value: String)
+
+        /// The `config.json` read-modify-write family.
+        ///
+        /// Two reasons, and the second is the load-bearing one:
+        /// `ConfigStore.withConfigLock` already makes each mutation atomic, so
+        /// the lane is what keeps two *pipelined* mutations from landing out of
+        /// order — but that lock is an `NSLock` held across synchronous disk
+        /// I/O, so concurrent config writers would **block cooperative-pool
+        /// threads**, the shape that wedged the daemon in #874. One lane means
+        /// `/rpc` never contends for that lock with itself.
+        case config
+
+        /// The Manager window and the tmux server it lives in. Typing into the
+        /// Manager (`work-on-issue`) must not race creating or restarting it.
+        case manager
+
+        /// `start-review` / `batch-start-review`. They already funnel through
+        /// ``ReviewKickoffSerializer`` — and `start-review` *awaits* its chained
+        /// task, so without a lane N pipelined kickoffs would each hold a
+        /// concurrency permit for the whole serialized clone-and-spawn run and
+        /// re-create #931 one level down.
+        case reviewKickoff
+
+        /// The global `~/.claude/settings.json` allowlist: promote, then scan.
+        case allowlist
+
+        /// Whole-board provider sweeps (`resync-jira`), which walk every session
+        /// and transition its ticket. Two at once would double every provider
+        /// call.
+        case tracker
+    }
+
+    // MARK: - Configuration
+
+    /// Ceiling on operations executing at once.
+    ///
+    /// Deliberately small. Most handlers hop to `@MainActor`, which is the real
+    /// bottleneck — raising this buys queueing, not throughput — and several
+    /// take `ConfigStore.withConfigLock`, an `NSLock` that blocks a cooperative
+    /// thread while held. 8 keeps blocked threads well under the pool width on
+    /// any host `crowd` runs on, and the ``Lane/config`` lane keeps `/rpc`'s own
+    /// contribution to that lock at one.
+    private let maxConcurrent: Int
+
+    /// Ceiling on operations *accepted* per connection before ``dispatch(lane:operation:)``
+    /// starts refusing.
+    ///
+    /// Non-blocking rejection rather than reader backpressure, matching
+    /// `TerminalConnectionLimiter`: a client holding 64 requests open on one
+    /// socket is not one the daemon should quietly queue for. The web UI never
+    /// exceeds a handful, so this only fires on a runaway caller — which learns
+    /// immediately instead of timing out 64 times.
+    private let maxInFlight: Int
+
+    // MARK: - State
+
+    private struct LaneTail {
+        let id: UUID
+        let task: Task<Void, Never>
+    }
+
+    private var lanes: [Lane: LaneTail] = [:]
+    private var pending: [UUID: Task<Void, Never>] = [:]
+    private var inFlight = 0
+    private var running = 0
+    private var permitWaiters: [CheckedContinuation<Bool, Never>] = []
+    private var drainWaiters: [CheckedContinuation<Void, Never>] = []
+    private var isShutDown = false
+
+    init(maxConcurrent: Int = 8, maxInFlight: Int = 64) {
+        self.maxConcurrent = maxConcurrent
+        self.maxInFlight = maxInFlight
+    }
+
+    // MARK: - Introspection (tests)
+
+    /// Live lane count. Zero once every chain has drained — the assertion that
+    /// the lane map is bounded by *live* lanes rather than by every session id a
+    /// long-lived connection has ever addressed.
+    var laneCount: Int { lanes.count }
+    var inFlightCount: Int { inFlight }
+    var runningCount: Int { running }
+
+    // MARK: - Dispatch
+
+    /// Accept `operation` for execution, ordered behind anything already queued
+    /// in `lane`. Returns immediately — the operation runs on its own task.
+    ///
+    /// Returns `false` when the connection is at its in-flight ceiling or the
+    /// dispatcher has shut down; the caller answers the request itself rather
+    /// than dropping it, since a dropped request would sit until the client's
+    /// deadline — the exact symptom this change exists to remove.
+    @discardableResult
+    func dispatch(lane: Lane?, operation: @escaping @Sendable () async -> Void) -> Bool {
+        guard !isShutDown, inFlight < maxInFlight else { return false }
+        inFlight += 1
+
+        let id = UUID()
+        let predecessor = lane.flatMap { lanes[$0]?.task }
+        let task = Task {
+            await self.run(id: id, lane: lane, predecessor: predecessor, operation: operation)
+        }
+        pending[id] = task
+        if let lane { lanes[lane] = LaneTail(id: id, task: task) }
+        return true
+    }
+
+    /// Return once every accepted operation has yielded its response.
+    ///
+    /// Called by the reader after inbound closes, so ending the writer's stream
+    /// cannot truncate a reply that is still being computed. ``shutdown()`` also
+    /// releases this, so a reader parked here when the *writer* is what died
+    /// unwinds instead of hanging the connection's task group.
+    func drain() async {
+        guard inFlight > 0, !isShutDown else { return }
+        await withCheckedContinuation { (c: CheckedContinuation<Void, Never>) in
+            drainWaiters.append(c)
+        }
+    }
+
+    /// Stop accepting, cancel every chain, and release every waiter.
+    ///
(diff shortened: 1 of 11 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- The `localOnlyDenial` security gate remains on the sequential reader path (`RPCWebSocketHandler.swift:100-110`), so privileged-field checks (e.g. `set-config` binaries, `gateway-get` with `reveal`) are not evaluated against a config another concurrent dispatch may be mid-writing.
- `RPCLanePolicy` is build-gated against `ParityLedger.rpcMethods` — every write RPC must declare a lane before the build passes, preventing silent concurrency for new handlers like `launch-agent` (`terminal_id`) or `job-run` (`job_id`).
- Bounded execution (`maxConcurrent: 8`, `maxInFlight: 64`) with immediate rejection (`RPCWebSocketHandler.swift:120-127`) instead of silent queueing limits resource exhaustion from a runaway client on one socket.
- `ensureMergeLabelOnce` populates the memo only on success (`IssueTracker.swift:3557-3575`), preserving CROW-816's guarantee that unreachable repos retry rather than latch a false success.
- `dismissModalDialog(token)` scopes modal retraction by identity (`app.js`), preventing a late RPC callback from dismissing a dialog the user opened afterward.

**Concerns:**
- No new security concerns identified. Cross-lane write reordering is a deliberate, documented trade-off; the web client already awaits each response before issuing dependents, and same-session / config / manager / job lanes preserve ordering where it matters.

### Code Quality
- **Fix A correctness win:** Replacing `await refresh()` in `addMergeLabel` with targeted `reevaluateAutoMergeAfterLabel` fixes a live bug where `refresh()`'s `guard !isRefreshing` made the `autoMergeWarning` read stale or empty while claiming success.
- **Fix B architecture:** `RPCDispatcher` is well-structured — permit acquired after lane wait (deadlock-safe), lane-tail eviction bounds memory, and `shutdown()` before `group.cancelAll()` correctly unparks `drain()` waiters. The 12 `RPCDispatcherTests` and 10 `RPCLanePolicyTests` cover the critical invariants.
- **Fix C UX:** Per-method timeouts mirror the CLI table; settled-entry retention + `onLate` correctly prevents spurious failure modals. The 47 `rpc-timeout.test.js` cases exercise the real `app.js` transport layer.
- **Fix E:** `ensureMergeLabelTasks` deduplication prevents N concurrent identical `gh label create` shell-outs on first poll — clean `@MainActor` join pattern.
- **Tests:** `make test` passes (0 failures, CLI/RPC parity green at 94 methods). All new suites pass locally.

### Summary Table

| # | Finding | Color |
|---|---------|-------|
| 1 | Security boundary (`localOnlyDenial`) kept sequential; lane policy build-gated | Green |
| 2 | Targeted auto-merge re-eval fixes silent `refresh()` no-op + removes 4-5s board poll from `add-merge-label` | Green |
| 3 | `RPCDispatcher` concurrency model is deadlock-safe with strong test coverage | Green |
| 4 | Client timeout/`onLate`/`dismissModalDialog` design prevents false failure UX and duplicate retries | Green |
| 5 | `ensureMergeLabel` memoization is success-only with in-flight dedup | Green |

---

Issue #790, opened 2026-07-20
In Progress status via crow:in-progress label fallback when issue has no GitHub Project (mirror #706)


dhilgaertner:
## Summary

When a session is opened for an issue, the issue often doesn't move to **In Progress** —
because we rely on GitHub **Project** status, and these issues aren't on a project. We already
solved the same problem for **In Review** with a fallback **label** (`crow:in-review`, #706).
Do the same for **In Progress**: when an issue has no project board, represent In Progress with
a `crow:in-progress` label.

## Background — the existing In Review fallback (the pattern to mirror)

#706 added a no-project fallback for In Review, all in
`Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubTaskBackend.swift`:

- **Constant:** `TicketStatus.inReviewFallbackLabel = "crow:in-review"`
  (`Packages/CrowCore/Sources/CrowCore/Models/Enums.swift`).
- **Write path — `setTaskStatus(url:status:)` (~L112–176):** when the issue has no project items
  (`!hasProjectItems(...)`), it calls `setInReviewFallbackLabel(url:repo:applied: status == .inReview)`
  instead of a project mutation (~L158–168). `ensureInReviewFallbackLabel` creates the label on
  demand (`gh label create crow:in-review … --description "Crow: in review (no-project status
  fallback)"`, ~L308–316).
- **Read path (~L605–611):** when the board yields no status and the issue carries
  `crow:in-review`, `projectStatus` is set to `.inReview`.

## Why In Progress is missing today

The write path's own comment spells it out (`GitHubTaskBackend.swift:162–164`):

> The label only encodes In Review — any other target status just clears it (Done is signaled by
> closing the issue; **In Progress by the active session**).

So In Progress was intentionally left to "there's a linked session" — which is invisible on the
issue itself when there's no project. Concretely, on session setup `setup.sh` runs
`crow transition-ticket --session … --to inProgress` (`setup.sh:1142`, routes through
`setTaskStatus`), and its legacy direct-project path logs **"Issue not on any project, skipping
status update"** (`setup.sh:1179`) — the issue never shows In Progress.

## Scope — mirror #706 for In Progress

- [ ] Add `TicketStatus.inProgressFallbackLabel = "crow:in-progress"` in `Enums.swift`.
- [ ] **Write path:** in `setTaskStatus`'s no-project branch, apply `crow:in-progress` when
      `status == .inProgress` (and clear it otherwise), alongside the existing in-review handling.
[...]

Issue #706, opened 2026-07-13
Label-based ticket status fallback (crow:in-review) when an issue isn't on a GitHub Project


dhilgaertner:
## Summary

When a GitHub issue is **not on a Project (v2)**, it has no Status field, so Crow can't move it through the pipeline (In Progress → In Review → Done) — the status update is simply skipped. Add a **label-based status fallback**: when an issue isn't on a project, represent its pipeline status with a `crow:` label (e.g. **`crow:in-review`**) instead of a Project Status field. The label then drives the in-review state (board section, badges, session ↔ ticket linkage) with no GitHub Project required.

## Problem

Crow's ticket pipeline status is entirely **GitHub Projects-v2-backed**:
- `AssignedIssue.projectStatus: TicketStatus` (`Packages/CrowCore/Sources/CrowCore/Models/AssignedIssue.swift:19`), with pipeline states `.backlog / .ready / .inProgress / .inReview / .done` (`Enums.swift:95`).
- **Read:** `GitHubTaskBackend` pulls the Project item's single-select "Status" field → `TicketStatus(projectBoardName:)` (`GitHubTaskBackend.swift:~531-538`). No project item ⇒ `projectStatus` stays `.unknown` (→ `.backlog`, `AppState.swift:657`).
- **Write:** `updateProjectV2ItemFieldValue` mutation (`GitHubTaskBackend.swift:~170`); driven from `IssueTracker.markInReview` (`:2539`), `setTicketStatus` (`:2639-2661`), the `transition-ticket` CLI (`MetadataCommands.swift:66`, `inProgress|inReview|done`), and gated by `AppState.canSetProjectStatus(for:)` (`:502-508`).
- **setup.sh:** `set_project_status()` queries the issue's `projectItems`; **when there are none it logs `Issue not on any project, skipping status update`** and does nothing (`setup.sh:~1146-1179`).

So for any repo/issue that isn't wired to a Project board (e.g. the Corveil `crow` repo — every `/crow-workspace` there prints "Issue not on any project, skipping status update"), the whole In-Progress/In-Review pipeline is a no-op. There's no way to signal "this issue is in review" without a Project.

## Proposal

Fall back to a **status label** when `canSetProjectStatus(for:)` is false / the issue has no Project item. The label is the source of truth for pipeline status in the no-project case; when the issue *is* on a Project, the existing Projects-v2 path stays authoritative (label fallback is only used when there's no board).

[...]

PR #933, opened 2026-07-31
Make mark-in-review move the provider board; delete EngineRouter's dead routes (CROW-876)

### LINKED ISSUE #876: Session-lifecycle: mark-in-review does no provider board transition; EngineRouter copies still return unconditional ok
Two related leftovers from #816 / PR #868, both consequences of the app retirement (ADR 0010). Neither is reachable-broken today, but both are code that says one thing and does another.

## 1. `mark-in-review` performs no provider board transition

`IssueTracker.markInReview` exists and does the real work — capability-gates on `.projectBoardStatus`, calls `backend.setTaskStatus(url:status:.inReview)`, handles `insufficientScope`. **Nothing calls it.**

The daemon handler (`RPCHandlers.swift`) only writes session status. It was written as "forward to the app when running, degrade to a local write when it is down"; the forwarding branch went away with the app, so the degraded branch is now the only branch. The result is that `mark-in-review` — from the web menu *and* `crow mark-in-review` — is functionally `set-status inReview`, while the UI still gates the menu item on `can_set_project_status`, implying a board move that never happens.

Workaround today is `crow transition-ticket --to inReview`, which does hit the provider. PR #868 documents that distinction in `docs/cli-reference.md` rather than silently leaving the trap.

Fix: have the handler call `tracker.markInReview` (reporting failures, as `mark-issue-done` / `add-merge-label` now do via `SessionActionError`), or delete `markInReview` and make the split explicit.

## 2. `EngineRouter` copies of the five lifecycle methods

`EngineRouter.swift` registers `mark-in-review`, `mark-issue-done`, `complete-session`, `set-session-active`, `add-merge-label` as fire-and-forget callbacks that return unconditional `{"ok":true}`. Four of those callbacks are never wired by the daemon.

Unreachable today — the daemon router registers all five and shadows the engine copies — but they are exactly the false-success shape #868 removed from the daemon path. If routing ever changes, the old lie comes back.

Fix: delete them, or align them with the daemon handlers' typed errors.

---
Context: PR #868 (https://github.com/corveil/crow/pull/868), raised in review.

dhilgaertner:
Closes #876

Two leftovers from the app retirement (ADR 0010), both code that says one thing and does another. This is a **decision ticket** — each item ended in "fix X, or delete it," and I went one way on each.

## Item 1 — `mark-in-review` now moves the board (fix, not delete)

`IssueTracker.markInReview` did the real work but had **zero callers**: its only entry point was `AppState.onMarkInReview`, which nothing assigns since the app was retired. The verb silently degraded to `set-status inReview` while the web menu still gated it on `can_set_project_status`.

**Why fix rather than delete.** `setTaskStatus(…, .inReview)` appears in exactly two places repo-wide — the dead `markInReview` and the generic `transitionTicket`. The board pipeline has three legs and two were wired:

| Leg | Path | Before |
|---|---|---|
| session start → In Progress | `setup.sh` → `crow transition-ticket` | wired |
| issue done → Done | `markIssueDone` → `closeTask` | wired |
| session → In Review | — | **missing** |

`resyncJira` maps `inReview` → In Review — it exists to *repair* the state this gap creates. Deleting would have ratified the hole and left `crow resync-jira` as the only route.

**Shape.** Follows `mark-issue-done`: board first (a failed transition exits non-zero rather than leaving the session marked), then the session write. Preconditions and provider failures are typed `SessionActionError`s.

**The load-bearing call: `.unimplemented` warns, it does not throw.** `GitHubTaskBackend` throws `.unimplemented` for any board whose Status column isn't named "In Review"/"Review", and for any item whose Status is unset. Throwing there would have made `crow mark-in-review` a hard error on a large slice of real boards — a breakage dressed as a fix. Instead the session still transitions and the response carries an additive `warning` (#888's shape). The same path covers GitLab.

**Two defects fixed in passing:**
- `markInReview` resolved its backend from the bare provider enum, dropping the host for self-hosted GitLab/Corveil. Now `taskBackend(forURL:)`, like its sibling.
- `ProviderError` is not `LocalizedError`, so copying `markIssueDone`'s generic catch would have reported `"The operation couldn't be completed. (error 1.)"` instead of the real `gh` error. Added `providerFailureDetail` to destructure it.

`SessionLifecycleRPC.requireTicketURL` retired — `mark-in-review` was its last caller, and keeping it would have meant two sources of truth for the ticket check in the very file whose doc comment forbids that.

⚠️ **New outward-facing side effect:** on a GitHub repo with **no** project board, `mark-in-review` now applies the `crow:in-review` fallback label (#706, #790) — a repo write this verb never made. It follows from calling `setTaskStatus` at all (`transition-ticket --to inReview` already does it), but it's new for this verb, so it's called out in the CHANGELOG rather than shipped quietly.

## Item 2 — EngineRouter's dead routes (delete)

Deleted **thirteen** routes returning unconditional `{"ok":true}`: the five lifecycle verbs plus `quick-action`, `work-on-issue`, `start-review`, `promote-allowlist`, `refresh-tickets`, `refresh-allowlist`, `create-manager`, `run-job`. Verified individually — every one is shadowed by a daemon handler (`CommandRouter` reaches a fallback only for an *un*registered method), and **12 of 13 callbacks are never assigned anywhere**. Aligning dead code with typed errors makes no sense, so: deleted.

The 12 orphaned `AppState` callbacks went too. `onCompleteSession` stays — `IssueTracker` uses it for auto-complete-on-merge. `EngineRouter.swift` and `AppState.swift` are **pure deletions** (122 and 47 lines; one doc-comment line changed between them). `transition-ticket` / `resync-jira` are engine-only and do real work — untouched.

Parity unaffected: the daemon registers all thirteen, so the union both `RPCLedgerParityTests` and `check-cli-parity.sh` assert on is unchanged (94 methods).

## Docs

`docs/automation.md` claimed a **"PR opened → mapped In Review"** transition that no code performs. Corrected to describe reality rather than implementing the missing trigger. Also updated the four surfaces documenting the old split (`CLAUDE.md` — which `Scaffolder` ships as the dev-root Manager context — `docs/cli-reference.md`, the ArgumentParser `discussion` literal and its regenerated `docs/cli.md`), plus `docs/troubleshooting.md`, which blamed `markInReview` for the session-*start* transition that goes through `transitionTicket`.

The web gate stays on `can_set_project_status`: the button's name now matches what it does on board-capable providers. Relaxing it for GitLab is a separate UX call.

`Resources/CLAUDE.md.template` is listed by ADR 0016 but is stale/orphaned — no Session Lifecycle section, nothing to update. Left alone; worth a follow-up.

## Verification

`make build` clean · `make docs` regenerated · `check-cli-parity.sh` green (94 methods)

| Package | Tests |
|---|---|
| CrowEngine | 624 ✅ |
| CrowCore | 544 ✅ |
| CrowCLI | 342 ✅ (incl. `CLIDocsTests` drift guard) |
| CrowDaemon | 55 ✅ (targeted: lifecycle + parity suites) |

`BoardForwarderTests`' fixture needed real surgery, not an assertion tweak: `seededRouter` passed no tracker, and the missing-session-id case had to leave the shared loop because the tracker guard now precedes param decoding — matching how `LocalLiveActionTests` already treats `mark-issue-done`.

## Known limits (documented, not expanded into)

- **Jira** treats an unreachable transition as a logged no-op and returns cleanly (`JiraTaskBackend.swift:223-225`) — reports success with no warning. #876's shape one layer down; follow-up candidate.
- **Corveil** maps `.inReview` → `in_progress` (ADR 0005). Lossy and silent; follow-up candidate.
- Sessions with a ticket but an unresolvable provider now get `.noProvider` where they previously got a silent local write. `mark-issue-done` already accepted this trade.

## Note on #932

PR #932 (CROW-931) is still open and touches `IssueTracker.swift` / `RPCHandlers.swift` too, but in different regions (`addMergeLabel` / `ensureMergeLabel` vs `markInReview`). Nothing to rebase onto yet; whichever lands second should expect a trivial merge at most.


dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Removes thirteen `EngineRouter` routes that returned unconditional `{"ok": true}` without performing work — eliminates false-success paths that could mislead callers if routing ever fell through to the engine fallback.
- `mark-in-review` now follows the same board-first, typed-error contract as `mark-issue-done`: provider failures surface as `SessionActionError` via `mapRPCError` rather than silently succeeding with a half-action.
- Web UI warning display uses `textContent` via `el()` (`app.js:970`), so server-returned warning strings are not interpreted as HTML.
- Tracker guard requires a provider-configured daemon before `mark-in-review` runs, preventing the prior session-only write that contradicted the UI's `can_set_project_status` gate.
- Self-hosted GitLab/Corveil backend resolution fixed by using `taskBackend(forURL:)` instead of bare provider enum (`IssueTracker.swift:4396-4401`).

**Concerns:**
- None blocking. The new `crow:in-review` fallback-label side effect on boardless GitHub repos is intentional, documented in CHANGELOG, and consistent with `transition-ticket --to inReview`.

### Code Quality
- Board-first ordering is correct: `tracker.markInReview` runs before `applySessionStatus`, so a failed provider transition leaves the session unchanged (`RPCHandlers.swift:1345-1359`).
- The `.unimplemented` → warning (not throw) asymmetry is well-reasoned and tested (`SessionActionReportingTests.markInReviewWarnsRatherThanThrowsWhenBoardCannotMove`).
- `providerFailureDetail` fixes opaque `ProviderError` messages for CLI users (`IssueTracker.swift:4457-4468`).
- Dual session write (`onSetSessionInReview` + `applySessionStatus`) mirrors the established `mark-issue-done` pattern and is documented as idempotent for no-tmux hosts.
- Dead `AppState` callbacks and `requireTicketURL` removed cleanly; precondition checks consolidated in `IssueTracker`.
- Test coverage spans engine preconditions, daemon router guards, and response-shape tests. `swift build` clean; targeted test suites pass locally.

### Summary Table

**Green (non-blocking):**
1. Consider reusing `providerFailureDetail` in `markIssueDone` for consistent `ProviderError` messaging (same file, pre-existing gap partially addressed).
2. Jira unreachable-transition silent no-op and Corveil `.inReview` → `in_progress` lossy mapping are documented known limits — reasonable follow-up tickets.
3. Daemon-layer test could assert the `warning` field on `mark-in-review` RPC responses (engine tests already cover the tracker return value).
4. `Resources/CLAUDE.md.template` staleness noted in PR description — worth a separate cleanup.

---

PR #935, opened 2026-08-03
Re-sync shell scrollback from tmux history at top-of-buffer (CROW-934)

### LINKED ISSUE #934: Shell tab scrollback is truncated — can't scroll back far
## Problem

Opening a **shell tab** on a session (`crow new-terminal`, i.e. a non-agent terminal) gives very limited scrollback. Scrolling up hits the top of the buffer after only a short distance, so output that scrolled past is unrecoverable.

## Expected

A shell tab should retain a large scrollback buffer — comparable to what agent terminals retain — so you can scroll back through prior command output for the life of the tab.

## Suspected causes (verify, don't assume)

Two independent limits stack here; check both:

1. **xterm.js `scrollback` option** — if the shell-tab terminal is constructed without an explicit `scrollback`, it falls back to the xterm.js default (1000 lines), which is small for build/test output. Confirm whether agent terminals set this and shell tabs don't.
2. **tmux `history-limit`** — the bundled tmux config governs how much the pane itself keeps. If the pane's history is short, no front-end setting can recover the lost lines.

Also worth checking: whether `AppConfig.terminal` (`wheelScrollLines`, `agentWheelNotches`) affects only agent panes and leaves shell tabs on a different path.

## Important: three front-ends

Per the terminal architecture, this behavior is hand-mirrored across **desktop `terminal.html`**, **web `terminal.html`**, and **web `app.js`** — these drift. Fix must land in all three (or, better, be factored into shared terminal-construction config rather than duplicated a fourth time). Transport stays host-specific; scrollback config is shared behavior.

## Acceptance criteria

- [ ] Open a shell tab, produce several thousand lines of output (e.g. `seq 1 20000`), scroll to the top — the earliest lines are still reachable.
- [ ] Same behavior verified in the desktop app **and** the web UI.
- [ ] Agent terminals are not regressed (scrollback unchanged or improved).
- [ ] If tmux `history-limit` is the binding constraint, it's raised in the bundled config and the change is picked up by `crow reload-tmux-config` / new panes.

dgershman:
Closes #934

## What was actually wrong

**Neither limit named in the issue was the cause — both were already correct.** Verified before changing anything:

- tmux `history-limit` is **50000**, and every live window — including freshly created `Shell` tabs — is born with it (`history_limit=50000, alternate_on=0`).
- All xterm.js constructions already pass `scrollback: 50000`.
- The daemon's `capture-pane -pe -S -50000` replay delivers **all 20 000 lines** over the live `/terminal` socket.

The real cause is that **the local xterm buffer is not a faithful copy of the pane**. tmux collapses pane redraws when the attached client can't drain them fast enough, and the browser *is* that slow client. Measured against the bundled config with `seq 1 20000` (pane history 19 989 lines):

| client | bytes delivered | lines reaching the front-end |
|---|---|---|
| raw reader | 210 KB | 19 768 / 20 000 |
| browser-paced reader | 7 KB | **337 / 20 000** |

crowd's `select-window` reply carries the authoritative copy and rebuilds the buffer in place — but it only ran on connect and on a tab **change** (`attachWindow` → `reloadTerminal`). That's why switching away and back "fixed" the history, while the tab you were sitting on never re-synced and scroll-up hit the top after a few screens.

## The fix

Re-request the replay when a plain-shell surface reaches the top of its local buffer. That's exactly when the user is asking for older lines, so it costs one capture per output burst rather than polling.

- **Agent surfaces are excluded** — they keep their transcript in the scrollback-less alt buffer and repaint it themselves (ADR-0013), so a capture there returns only the viewport.
- **`baseY > 0` is load-bearing** — on a tab that has printed less than one screen `viewportY` is permanently 0, so testing "at the top" alone would fire a tmux capture on every output burst for the life of the tab.
- The settle is **debounced, not `term.write`'s callback**, because live output can interleave with the replay; it's self-healing, so a replay that never arrives can't wedge the surface.

Verified that a **repeat** `select-window` on the same window returns 20 000/20 000 lines and is byte-identical to the first — the `ESC[H ESC[2J ESC[3J` rebuild means re-requesting can't stack duplicates.

## On "must land in all three front-ends"

The issue asks for the fix in all three. Only **one** is a live front-end with tab switching:

- `web/app.js` — production. Fixed here.
- `web/terminal.html` — the standalone debug page; single-pane, has no `select-window` protocol at all, and uses xterm's native wheel. Nothing to sync.
- `CrowTerminal/Resources/xterm/terminal.html` — **dead**. Its `crowWrite`/`crowInput`/`crowReady`/`crowResize` bridge has zero callers anywhere in the repo; it lost its last consumer when ADR-0010 retired the macOS app. app.js's comment claiming to mirror it is corrected here. (Deleting the file is left as a separate cleanup, with precedent in #927.)

No shared terminal-construction config was introduced because the construction values were never the problem — they are already identical and correct in both live pages. This change adds behavior that only makes sense where tab switching exists.

## Acceptance criteria

- [x] Shell tab, several thousand lines, scroll to top → earliest lines reachable (the re-sync restores all 20 000 and lands the viewport on line 1).
- [x] Same code path serves the desktop app — `crow-desktop` is a shell that loads this same `app.js`, so there is no second implementation to keep in step.
- [x] Agent terminals not regressed — excluded by an explicit `agent_surface` guard, pinned by a test.
- [x] tmux `history-limit` was **not** the binding constraint (verified 50000 on new panes), so no config change was needed.

## Testing

New `web-tests/scrollback-hydrate.test.js` pins the gating — 10 cases covering the shell-at-top trigger, the agent exclusion, the short-buffer guard, mid-buffer, already-synced, repeat-scroll idempotence, closed socket, and no active terminal.

Full web suite passes. `row.test.js` reports 10 failures, which are **pre-existing on main** — confirmed identical (43 passed / 10 failed) with these changes stashed.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -4328,8 +4328,14 @@ function ensureTerminal() {
 
 // The one guarded writer to the PTY socket, shared by term.onData and the touch
 // scroll shim below (mirrors terminal.html's sendToPTY of the same name).
+// CROW-934: bumped on every keystroke so a hydrate settling in the background
+// can tell whether the user has typed since it armed — `scrollOnUserInput`
+// deliberately takes a typing user to the bottom, and a restore that yanks them
+// back to the top mid-keystroke is worse than no restore.
+let userInputSeq = 0;
 function sendToPTY(text) {
   if (!text) return;
+  userInputSeq++;
   if (termWs && termWs.readyState === WebSocket.OPEN) termWs.send(new TextEncoder().encode(text));
 }
 
@@ -4691,10 +4697,11 @@ function connectTerminalWs() {
     if (event.data instanceof ArrayBuffer) {
       term.write(new Uint8Array(event.data));
       // CROW-934: while a re-sync is in flight these bytes are (or accompany)
-      // the replay — keep pushing the settle out so scrollToTop runs once the
-      // rebuild is quiet. Otherwise they're live output, which tmux may again
-      // have thinned, so the next visit to the top should re-sync.
-      if (hydratingScrollback) scheduleHydrateSettle(150);
+      // the replay — record that one actually landed and keep pushing the settle
+      // out so the restore runs once the rebuild is quiet. Otherwise they're
+      // live output, which tmux may again have thinned, so the next visit to the
+      // top should re-sync.
+      if (hydratingScrollback) { hydrateSawReplay = true; scheduleHydrateSettle(150); }
       else scrollbackDirty = true;
       // Fade the skeleton out once the first real content has landed — rAF so
       // the hide follows the paint, not precedes it.
@@ -4711,6 +4718,10 @@ function connectTerminalWs() {
     // #679 expired scrim.
     clearTimeout(termSkelTimer);
     termSkelTimer = null;
+    // CROW-934: this path reconnects WITHOUT going through reloadTerminal, so it
+    // needs the same teardown — a sync left in flight across the drop would have
+    // the fresh attach's own replay consumed as its replay (#935 review).
+    resetScrollbackSync();
     if (sessionDead) { hideTerminalSkeleton(); return; }
     setTerminalReconnecting(true);
     showTerminalSkeleton();
@@ -4800,13 +4811,35 @@ function selectWindow(win) {
 // "fixed" the history — but the tab you sit on never re-synced.
 //
 // Trigger on reaching the top of the local buffer: that is exactly when the
-// user is asking for older lines, and it costs one capture per output burst
-// rather than polling. Shell surfaces only — an agent keeps its transcript in
-// the scrollback-less alt buffer and repaints it itself (ADR-0013), so a
-// capture there returns just the viewport and would buy nothing.
+// user is asking for older lines. Shell surfaces only — an agent keeps its
+// transcript in the scrollback-less alt buffer and repaints it itself
+// (ADR-0013), so a capture there returns just the viewport and would buy
+// nothing.
+//
+// Two independent brakes, because `onScroll` is NOT a user-scroll event: xterm's
+// BufferService.scroll ends in an unconditional `this._onScroll.fire(ydisp)`, so
+// it fires for every line pushed at the bottom, and a viewport the user parked
+// at the top stays pinned at `ydisp === 0` (`Math.max(ydisp - 1, 0)` once the
+// buffer is full). Reading early output while a build keeps printing therefore
+// satisfies every guard on every output line. Without a brake each capture
+// re-armed the next one — two tmux spawns and the whole 50k history per round
+// trip, indefinitely (#935 review).
+//
+//   * `lastHydrateAt` bounds the cost while output is still streaming.
+//   * `scrollbackFullySynced` latches off entirely once a capture comes back
+//     with nothing new, until the viewport leaves the top. That is also what
+//     keeps the one speculative capture per tab (the connect replay already left
+//     the buffer authoritative, but we can't tell that without asking) from
+//     repeating.
+const HYDRATE_MIN_INTERVAL_MS = 5000;
 let scrollbackDirty = true; // live output may have dropped lines since the last sync
 let hydratingScrollback = false;
 let hydrateSettleTimer = null;
+let hydrateBaseY = 0; // scrollback depth when the capture was armed
+let hydrateSawReplay = false; // did any bytes actually land during the flight?
+let hydrateInputSeq = 0; // userInputSeq when the capture was armed
+let scrollbackFullySynced = false; // last capture returned nothing new
+let lastHydrateAt = 0;
 
 // The replay rebuilds from the top (`ESC[H ESC[2J ESC[3J`), which parks the
 // viewport at the BOTTOM — so once the bytes stop landing, put the user back on
@@ -4819,23 +4852,54 @@ function scheduleHydrateSettle(ms) {
   clearTimeout(hydrateSettleTimer);
   hydrateSettleTimer = setTimeout(() => {
     hydratingScrollback = false;
-    try { term.scrollToTop(); } catch (_) {}
+    const buf = term && term.buffer && term.buffer.active;
+    // Nothing deeper came back → the local buffer already matches the pane, so
+    // stop asking until the viewport moves off the top.
+    if (buf && buf.baseY <= hydrateBaseY) scrollbackFullySynced = true;
+    // Only move the viewport if a replay actually landed AND the user hasn't
+    // typed since we armed. Both matter: the 2000 ms "no replay ever arrived"
+    // path would otherwise jump them to the top of an unchanged buffer, and
+    // `scrollOnUserInput` deliberately takes a typing user to the bottom —
+    // yanking them back mid-keystroke is worse than not restoring at all.
+    if (hydrateSawReplay && userInputSeq === hydrateInputSeq) {
+      try { term.scrollToTop(); } catch (_) {}
+    }
   }, ms);
 }
 
+// Shared by reloadTerminal and the onclose auto-reconnect: both tear the socket
+// down, and a sync left in flight across one would have the fresh attach's own
+// replay consumed as its replay, with a pending scrollToTop fighting it.
+function resetScrollbackSync() {
+  clearTimeout(hydrateSettleTimer);
+  hydrateSettleTimer = null;
+  hydratingScrollback = false;
+  hydrateSawReplay = false;
+  scrollbackFullySynced = false;
+  scrollbackDirty = true;
+}
+
 function maybeHydrateScrollback() {
-  if (!term || hydratingScrollback || !scrollbackDirty) return;
+  if (!term || hydratingScrollback) return;
+  const buf = term.buffer.active;
+  // Leaving the top re-arms the latch — there is new ground to cover next time.
+  if (buf.viewportY !== 0) { scrollbackFullySynced = false; return; }
+  if (scrollbackFullySynced || !scrollbackDirty) return;
+  // `baseY > 0` is load-bearing: on a tab that has printed less than one screen
+  // viewportY is permanently 0, so testing the top alone would fire a capture on
+  // every output burst for the whole life of the tab.
+  if (buf.baseY === 0) return;
   if (!activeTerminal || activeTerminal.agent_surface) return;
   if (activeTerminal.window == null) return;
   if (!termWs || termWs.readyState !== WebSocket.OPEN) return;
-  // At the top of a buffer that HAS scrollback. `baseY > 0` is load-bearing: on
-  // a short buffer (a tab that has printed less than one screen) viewportY is
-  // permanently 0, so testing the top alone would fire a capture on every
-  // output burst for the whole life of the tab.
-  const buf = term.buffer.active;
-  if (buf.baseY === 0 || buf.viewportY !== 0) return;
+  const now = Date.now();
+  if (now - lastHydrateAt < HYDRATE_MIN_INTERVAL_MS) return;
+  lastHydrateAt = now;
   hydratingScrollback = true;
   scrollbackDirty = false;
+  hydrateBaseY = buf.baseY;
+  hydrateSawReplay = false;
+  hydrateInputSeq = userInputSeq;
   scheduleHydrateSettle(2000);
   selectWindow(activeTerminal.window);
 }
@@ -4870,11 +4934,9 @@ function reloadTerminal() {
   lastTermCols = 0;
   lastTermRows = 0;
   // CROW-934: the reconnect re-selects the window, so crowd replays the pane
-  // anyway — drop any in-flight re-sync (its scrollToTop would fight the fresh
+  // anyway — drop any in-flight re-sync (its restore would fight the fresh
   // attach) and treat the rebuilt buffer as needing one again.
-  clearTimeout(hydrateSettleTimer);
-  hydratingScrollback = false;
-  scrollbackDirty = true;
+  resetScrollbackSync();
   if (termWs) {
     const old = termWs;
     old.onopen = old.onmessage = old.onclose = old.onerror = null;
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -4328,14 +4328,8 @@ function ensureTerminal() {
 
 // The one guarded writer to the PTY socket, shared by term.onData and the touch
 // scroll shim below (mirrors terminal.html's sendToPTY of the same name).
-// CROW-934: bumped on every keystroke so a hydrate settling in the background
-// can tell whether the user has typed since it armed — `scrollOnUserInput`
-// deliberately takes a typing user to the bottom, and a restore that yanks them
-// back to the top mid-keystroke is worse than no restore.
-let userInputSeq = 0;
 function sendToPTY(text) {
   if (!text) return;
-  userInputSeq++;
   if (termWs && termWs.readyState === WebSocket.OPEN) termWs.send(new TextEncoder().encode(text));
 }
 
@@ -4844,33 +4838,38 @@ let hydrateSettleTimer = null;
 let hydrateArmedAt = 0; // start of the current flight, for the absolute deadline
 let hydrateBaseY = 0; // scrollback depth when the capture was armed
 let hydrateSawReplay = false; // did any bytes actually land during the flight?
-let hydrateInputSeq = 0; // userInputSeq when the capture was armed
 let scrollbackFullySynced = false; // last capture returned nothing new
 let lastHydrateAt = 0;
 let lastViewportY = 0; // previous scroll position, to detect arrival at the top
 
-// The replay rebuilds from the top (`ESC[H ESC[2J ESC[3J`). Once the bytes stop
-// landing, put the user back on the oldest lines they scrolled up to ask for.
-// Debounced rather than done in term.write's callback because live output can
-// interleave with the replay — but clamped to the arm-time deadline so a busy
-// shell cannot extend it indefinitely. Self-healing in both directions: no
-// replay at all still clears the flag, and neither does a replay that never
-// stops.
+// The flight ends this long after the last frame, but never later than the
+// arm-time deadline. There is deliberately NO viewport restore here: the replay
+// rebuilds the buffer while `isUserScrolling` is true (which the trigger's
+// `viewportY === 0 && baseY > 0` guards imply), and every write then takes
+// `isUserScrolling || i.ydisp++` / `isUserScrolling && (ydisp = max(ydisp-1,0))`,
+// so `ydisp` stays pinned at 0 — the rebuild leaves the user on the oldest lines
+// by itself. A `scrollToTop()` here would be `Viewport.scrollLines(-0)`, the same
+// scrollTop, hence a no-op on that path; the ONLY way to reach it non-inert is
+// the user leaving the top mid-flight (the #668 pill calls `scrollToBottom()`,
+// clearing `isUserScrolling` so `ydisp` tracks the bottom), where it would yank
+// them off the live prompt to line 1 of a 20k buffer. It never helps and acts
+// only when acting is wrong, so it is gone (#935 review round 3).
 function scheduleHydrateSettle(ms) {
   clearTimeout(hydrateSettleTimer);
   const remaining = hydrateArmedAt + HYDRATE_MAX_FLIGHT_MS - Date.now();
   hydrateSettleTimer = setTimeout(() => {
     hydratingScrollback = false;
     const buf = term && term.buffer && term.buffer.active;
-    // Nothing deeper came back → the local buffer already matches the pane, so
-    // stop asking until the viewport moves off the top. Only trust that when a
-    // replay actually landed; an empty flight proves nothing.
-    if (hydrateSawReplay && buf && buf.baseY <= hydrateBaseY) scrollbackFullySynced = true;
-    // Only move the viewport if a replay landed AND the user hasn't typed since
-    // we armed — `scrollOnUserInput` deliberately takes a typing user to the
-    // bottom, and yanking them back mid-keystroke is worse than not restoring.
-    if (hydrateSawReplay && userInputSeq === hydrateInputSeq) {
-      try { term.scrollToTop(); } catch (_) {}
+    if (hydrateSawReplay) {
+      // Nothing deeper came back → the local buffer already matches the pane, so
+      // stop asking until the viewport moves off the top.
+      if (buf && buf.baseY <= hydrateBaseY) scrollbackFullySynced = true;
+    } else {
+      // An empty flight proves nothing — neither that we synced (so no latch)
+      // nor that the buffer is clean. Arming cleared `scrollbackDirty` on the
+      // premise of a sync that never happened; put it back so a retry is
+      // possible without waiting for live output.
+      scrollbackDirty = true;
     }
   }, Math.max(0, Math.min(ms, remaining)));
 }
@@ -4889,7 +4888,7 @@ function noteTerminalFrame() {
 
 // Shared by reloadTerminal and the onclose auto-reconnect: both tear the socket
 // down, and a sync left in flight across one would have the fresh attach's own
-// replay consumed as its replay, with a pending scrollToTop fighting it.
+// replay consumed as its replay.
 function resetScrollbackSync() {
   clearTimeout(hydrateSettleTimer);
   hydrateSettleTimer = null;
@@ -4898,6 +4897,7 @@ function resetScrollbackSync() {
   scrollbackFullySynced = false;
   scrollbackDirty = true;
   lastHydrateAt = 0; // the new surface gets its own budget, not the old one's
+  lastViewportY = 0; // ...and its own scroll history, not the old one's position
 }
 
 function maybeHydrateScrollback() {
@@ -4926,7 +4926,6 @@ function maybeHydrateScrollback() {
   hydrateArmedAt = now;
   hydrateBaseY = buf.baseY;
   hydrateSawReplay = false;
-  hydrateInputSeq = userInputSeq;
   scheduleHydrateSettle(2000);
   selectWindow(activeTerminal.window);
 }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -4255,7 +4255,10 @@ function ensureTerminal() {
   const imageAddon = new ImageAddon.ImageAddon({ sixelSupport: true, iipSupport: true, kittySupport: true });
   searchAddon = new SearchAddon.SearchAddon();
   const webLinksAddon = new WebLinksAddon.WebLinksAddon();
-  // Config block mirrors CrowTerminal/Resources/xterm/terminal.html.
+  // Config block mirrors web/terminal.html (the standalone debug page). It also
+  // named CrowTerminal/Resources/xterm/terminal.html, but that page lost its
+  // last consumer when ADR-0010 retired the macOS app — its crowWrite/crowInput
+  // bridge has no caller — so it is no longer a front-end to stay in sync with.
   term = new Terminal({
     cursorBlink: true,
     fontSize: 14,
@@ -4295,6 +4298,9 @@ function ensureTerminal() {
     term.loadAddon(webglAddon);
   } catch (_) { /* WebGL unavailable → canvas/DOM renderer */ }
   term.onData(sendToPTY);
+  // CROW-934: hydrate the shell scrollback from tmux when the user reaches the
+  // top of what the live stream actually delivered.
+  term.onScroll(maybeHydrateScrollback);
   term.attachCustomKeyEventHandler(handleTerminalKey);
   enableTouchScroll(document.getElementById('terminal'));
   enableWheelScroll(document.getElementById('terminal'));
@@ -4684,6 +4690,12 @@ function connectTerminalWs() {
   termWs.onmessage = (event) => {
     if (event.data instanceof ArrayBuffer) {
       term.write(new Uint8Array(event.data));
+      // CROW-934: while a re-sync is in flight these bytes are (or accompany)
+      // the replay — keep pushing the settle out so scrollToTop runs once the
+      // rebuild is quiet. Otherwise they're live output, which tmux may again
+      // have thinned, so the next visit to the top should re-sync.
+      if (hydratingScrollback) scheduleHydrateSettle(150);
+      else scrollbackDirty = true;
       // Fade the skeleton out once the first real content has landed — rAF so
       // the hide follows the paint, not precedes it.
       if (!painted) { painted = true; requestAnimationFrame(hideTerminalSkeleton); }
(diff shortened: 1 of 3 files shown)


dhilgaertner (changes requested):
## Code & Security Review

The diagnosis in the PR body is excellent and I verified the load-bearing claims independently:

- `CrowTerminal/Resources/xterm/terminal.html` really is dead — `crowWrite`/`crowInput`/`crowReady`/`crowResize` have zero callers outside that file. The comment correction is right.
- `replayFrame`'s `ESC[H ESC[2J ESC[3J` prefix (`TerminalCockpit.swift:145`) does make a repeat `select-window` idempotent, so re-requesting cannot stack duplicate history.
- `ensureTerminal` is `if (term) return` guarded, so `term.onScroll` is registered exactly once.
- `baseY === 0` is genuinely load-bearing, and `viewportY`/`baseY` are used with the right semantics.
- The `agent_surface` exclusion is consistent with ADR-0013 and with how `appOwnsScroll`/`activeSurfaceIsAgent` already treat daemon-supplied metadata as authoritative.
- `scrollback-hydrate.test.js` passes 10/10. `row.test.js`'s 43 passed / 10 failed is **confirmed pre-existing** — I ran that same test against `origin/main`'s `app.js` and got byte-identical results.

The mechanism is right. The problem is that the trigger has no rate limit, and `onScroll` fires far more often than "the user scrolled."

### Critical Issues

**🔴 `app.js:4826` — the re-sync has no cooldown, so a shell with ongoing output re-captures the full 50 000-line history every ~RTT + 150 ms, indefinitely.**

The comment says the trigger "costs one capture per output burst rather than polling." That is accurate, but an output burst is a far lower bar than it reads as, because `onScroll` is not a user-scroll event. From the bundled `@xterm/xterm 6.0.0` (`xterm.js`, `BufferService.scroll`):

```js
… this.isUserScrolling || (i.ydisp = i.ybase), this._onScroll.fire(i.ydisp)
```

That fire is unconditional and `scroll()` runs for **every line pushed at the bottom**, not just viewport moves. And in the same function, a user-scrolled viewport keeps `ydisp` pinned (`i.ybase++, this.isUserScrolling || i.ydisp++`; and `i.ydisp = Math.max(i.ydisp - 1, 0)` once the buffer is full). So a viewport parked at the top stays at `viewportY === 0` while output streams, and every output line fires `maybeHydrateScrollback` with the guards already satisfied.

The loop closes on itself:

1. Live frame → `scrollbackDirty = true` (`app.js:4698`); its `onScroll` → hydrate: `hydrating = true`, `settle(2000)`, `select-window` sent.
2. Daemon runs `select-window` + `capture-pane -pe -S -50000` (`TerminalWebSocket.swift:92-104`) — two tmux spawns, the whole pane history down the socket.
3. Replay frames arrive → `scheduleHydrateSettle(150)` each; interleaved live output only extends the settle.
4. 150 ms of quiet → `hydrating = false`, `term.scrollToTop()` → `ydisp` is 0 (whether or not the rebuild parked at the bottom).
5. Next live frame → `scrollbackDirty = true` → its `onScroll` → **step 1 again.**

Nothing in the chain latches off. On a local daemon that is roughly 5 full-history captures per second: 10 tmux process spawns/sec, ~200 KB × 5/s over the WebSocket (your own measurement for a 20 k-line pane), and a 20 k-line xterm buffer teardown-and-rebuild five times a second. The browser was already the slow client this PR exists to compensate for; this is likely to wedge the tab rather than help it.

The repro is an ordinary workflow, not a corner case: scroll to the top of a shell to read early output while a build or `tail -f` keeps printing. Note also that `term.scrollToTop()` in step 4 is what re-arms the loop — if the rebuild left the viewport at the bottom the loop would break on its own.

Suggested shape for a fix (any one of these closes it):
- A minimum interval between re-syncs (a `lastHydrateAt` timestamp, refuse inside ~5 s), **or**
- latch off when a sync yields nothing: record `buf.baseY` before the capture and, if it hasn't grown after the settle, set a "fully synced" flag that only clears when `viewportY` moves away from 0.

The second also fixes the redundant capture noted in Green #1 below.

### Code Quality

**🟡 `app.js:4818` — `scheduleHydrateSettle`'s `term.scrollToTop()` is unconditional and can yank a viewport the user has since moved.**

The timer fires on both paths, including the 2000 ms "no replay ever arrived" one — so a failed sync doesn't just clear the flag (which the comment correctly calls self-healing), it also jumps the user to the top of the *unchanged* buffer for no reason. And `scrollOnUserInput` defaults to `true` in this config block, so typing scrolls to the bottom: trigger a hydrate at the top, start typing a command, and ~150 ms after the last replay byte the viewport is pulled back to the top mid-keystroke. Same for hitting the #668 jump-to-bottom pill inside the window. Worth gating the `scrollToTop()` on the viewport still being where the sync left it (or on a replay having actually landed).

**🟡 `app.js:4874-4879` — the flag reset added to `reloadTerminal()` is missing from the auto-reconnect path.**

`termWs.onclose` reconnects via `setTimeout(connectTerminalWs, termReconnectDelay)` (`app.js:4718`), which does **not** go through `reloadTerminal()`. So a socket that drops mid-hydrate keeps `hydratingScrollback === true` and `scrollbackDirty === false` across the reconnect; if the reconnect lands inside the settle window, the fresh attach's own replay is consumed as the hydrate replay and the pending `scrollToTop()` fights it — exactly the hazard the `reloadTerminal` reset was added to prevent, per its own comment. Clearing the same three pieces of state in `onclose` would make the two teardown paths agree.

### Security Review

**Strengths:**
- No new input handling, no new message type, no new parsing. `select-window` is an existing frame and `activeTerminal.window` is a daemon-supplied integer, not user text.
- The `/terminal` upgrade is still gated by `WebSocketOriginGuard` + `WebAuthGuard`, untouched here.
- No secrets, no new storage, no new URL surface.

**Concerns:**
- Only the self-inflicted amplification above. `select-window` has no server-side rate limit — `TerminalWebSocket.swift:92` runs a 50 k-line `capture-pane` per frame with no throttle. That is pre-existing and reachable only post-auth, but this change is the first thing to drive it automatically in a loop, which makes a server-side floor worth considering as a follow-up.

### Tests

**🟢 The new suite pins the gating well, but the timers are stubbed out, so the behavior the Red finding lives in is untested.** `window.setTimeout = () => 0` (`scrollback-hydrate.test.js:41`) means `scheduleHydrateSettle` never runs — nothing exercises the settle, the `scrollToTop()`, the `onmessage` dirty/settle branch, or the `reloadTerminal` reset. "Repeated scrolls at the top issue exactly one capture" only covers repeats *within one in-flight hydrate*; the claim that matters — one capture per burst, not one per output line, across settle boundaries — isn't pinned. A fake timer queue (collect callbacks, run them explicitly) would let the same harness cover the full cycle and would have surfaced the loop.

**🟢 `app.js:4698` — first connect leaves `scrollbackDirty === true` on a buffer that just came from an authoritative replay.** `onopen` → `selectWindow` → the replay frames land with `hydratingScrollback === false`, so they set the dirty flag. The first scroll-to-top of every tab therefore spends a full 50 k capture re-fetching history the buffer already has.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 2 Yellow, 2 Green] findings.

---

dgershman:
Thanks — the Red finding is correct, and I reproduced the mechanism in the bundled xterm.js rather than taking it on trust. `BufferService.scroll` ends in an unconditional `this._onScroll.fire(i.ydisp)`, and the full-buffer branch is `e ? this.isUserScrolling && (i.ydisp = Math.max(i.ydisp - 1, 0))` — so a top-parked viewport stays pinned at 0 while every output line fires the handler with all guards satisfied. Your step-by-step was exactly right, including that `scrollToTop()` is what re-armed it.

All five findings addressed in 38e778b1.

**🔴 The loop — two brakes, because either alone leaves a hole.** I took your second suggestion (the latch) but had to add the first as well: the latch trips on "capture came back no deeper," which never happens while output is still streaming, since each capture legitimately returns more. So:

- **5 s floor** (`HYDRATE_MIN_INTERVAL_MS`) bounds the streaming case.
- **Latch** (`scrollbackFullySynced`) takes a quiet shell to zero captures, cleared when the viewport leaves the top.

You were right that the latch also subsumes **🟢 #1**: the connect replay leaves the buffer authoritative, we just can't know that without asking once — so it now costs exactly one capture per visit to the top instead of one per burst.

**🟡 restore yanking the viewport.** Now gated on a replay having actually landed *and* `userInputSeq` being unchanged since the capture armed. That closes both cases you named: the 2000 ms self-heal path no longer jumps to the top of an unchanged buffer, and typing mid-hydrate no longer fights `scrollOnUserInput`. The jump-to-bottom pill is not covered — it isn't distinguishable from the rebuild's own parking without hooking the addon — so a pill click inside the settle window can still be overridden. Happy to take that further if you want it airtight.

**🟡 reconnect path.** Factored into `resetScrollbackSync()`, called from `onclose` (before the `sessionDead` return) as well as `reloadTerminal`, so the two teardown paths agree.

**🟢 tests.** You were right that the stubbed timer is why I missed this. The suite now drives a fake timer queue, so the settle, restore, latch and teardown are all exercised — 24 cases, up from 10. The pin for this finding is a 500-output-line case at the top asserting exactly one capture; it also asserts the latch is *off* in that case, so it can't quietly start passing via the wrong brake.

One residual cost worth naming: parked at the top with output still streaming, you now get one full-history rebuild every 5 s. That's bounded and each one returns real new lines, but it is a repaint. If you'd rather trade freshness for quiet there, raising the floor is a one-constant change.

Not addressed, and I don't think it should be here: the server-side floor on `select-window`. Agreed it's worth having, but it's pre-existing and post-auth, and adding throttling to `TerminalWebSocket` is a different blast radius than this client change — better as its own ticket.

dhilgaertner (changes requested):
## Code & Security Review

The two brakes are the right answer to the previous round's Red, and I verified them rather than taking them on trust:

- The 5 s floor and the latch both hold. `scrollback-hydrate.test.js` is 25/25, and the streaming case genuinely pins the cooldown — the `scrollbackFullySynced === false` assertion at line 171 stops it from passing via the wrong brake, which is exactly the right instinct.
- `resetScrollbackSync()` is called from `onclose` (`app.js:4724`) **before** the `sessionDead` return, so the reconnect path and `reloadTerminal` now agree. Yellow 2 closed.
- The latch's re-arm is correct: leaving the top clears it (`app.js:4886`), and a quiet shell returning to the top costs zero captures because `scrollbackDirty` is still false. Green 1 is genuinely subsumed.
- `row.test.js` 43 passed / 10 failed is **confirmed pre-existing** — I ran that same suite against `origin/main`'s `app.js` in an isolated tree and got byte-identical results. Every other suite passes.
- Daemon side re-verified: `TerminalWebSocket.swift:92-104` → `replayData` → `replayFrame`'s `ESC[H ESC[2J ESC[3J` prefix (`TerminalCockpit.swift:141-148`), so a repeat `select-window` still cannot stack duplicates.

Two findings remain, and they are linked — the second is why the first exists.

### Code Quality

**🟡 `app.js:4704` — the settle has no deadline, so live output re-arms it indefinitely and `hydratingScrollback` latches true on a busy shell.**

```js
if (hydratingScrollback) { hydrateSawReplay = true; scheduleHydrateSettle(150); }
else scrollbackDirty = true;
```

Every binary frame during a flight pushes the settle out another 150 ms, and `scheduleHydrateSettle` is `clearTimeout` + `setTimeout` with no absolute cap (`app.js:4852-4853`). Frames are not batched anywhere: `pty.onOutput` yields per read chunk and `outputTask` writes one WS binary frame per chunk (`TerminalWebSocket.swift:56, 67-70`). A shell producing output more often than every 150 ms — a build, `npm test`, `tail -f` on a busy log — therefore never lets the settle fire. Then:

- `hydratingScrollback` stays `true`, so `maybeHydrateScrollback` early-returns at `app.js:4883` for the life of the socket;
- `scrollbackDirty` stays `false`, because every frame takes the `if` branch;
- neither brake is ever consulted again — the surface simply stops re-syncing until a tab switch or a socket drop calls `resetScrollbackSync()`.

The first hydrate does land, so the user gets their history once; after that the feature is silently off on exactly the workload the PR describes ("read early output while a build keeps printing"). The comment at `app.js:4849` — "Self-healing: if no replay ever arrives the timer still clears the flag, so a failed sync can't wedge the surface" — holds only for *zero* bytes, not for the far more common case of bytes that aren't the replay.

The suite can't see this because it never drives `termWs.onmessage`. The streaming case models a live frame as `T.scrollbackDirty = true` (`scrollback-hydrate.test.js:161`), which is the `else` branch — the one that does **not** run during a flight. Driving the real `onmessage` (or a helper that mirrors it) would have surfaced this the same way the fake timer queue surfaced the last one.

Fix is small: an absolute deadline on the settle (record an arm timestamp, refuse to extend past it), or stop extending once `hydrateSawReplay` is set and the frame is plausibly post-replay.

**🟡 `app.js:4845-4868` — the restore is a no-op in the only path that can reach it, and its test passes only because the fake `term` has no scroll semantics.**

The comment says the replay "parks the viewport at the BOTTOM". In the bundled xterm 6.0.0 that is not what happens on the trigger path, because the trigger *implies* `isUserScrolling === true`:

- `BufferService.scroll` ends in `this.isUserScrolling || (i.ydisp = i.ybase)`, so with `isUserScrolling` false `ydisp` always equals `ybase`. The guards require `viewportY === 0 && baseY > 0` (`app.js:4886, 4891`), which is unreachable unless `isUserScrolling` is true.
- `ESC[3J` trims scrollback via `ybase = max(ybase-e,0); ydisp = max(ydisp-e,0)` and never touches `isUserScrolling`.
- Rewriting the history then hits `(i.ybase++, this.isUserScrolling || i.ydisp++)` and, once full, `this.isUserScrolling && (i.ydisp = Math.max(i.ydisp-1, 0))` — `ydisp` stays 0 throughout.
- So at settle, `scrollToTop()` is `scrollLines(-0)`, and `scrollLines` moves nothing when `ydisp` is already 0.

The rebuild already leaves the user on the oldest lines, which is the good news — the feature works. But it means `hydrateSawReplay`, `hydrateInputSeq`, the `userInputSeq` counter (`app.js:4335, 4338`) and the guarded `scrollToTop()` are machinery for a condition that doesn't arise, and the 150 ms settle that exists to schedule them is what makes the finding above reachable. `scrollback-hydrate.test.js:192-200` asserts `fake.scrolledToTop === 1`, but `fake.scrollToTop()` unconditionally increments a counter (line 88) — the fake has no `ydisp`/`isUserScrolling`, so the assertion cannot distinguish "restored" from "called a no-op".

Worth confirming against a real xterm before deciding what to keep. If it holds, dropping the restore collapses the settle to a single bounded timer and dissolves the first finding with it.

### Security Review

**Strengths:**
- No new frame type, no new parsing, no new storage. `select-window` is pre-existing and `activeTerminal.window` is a daemon-supplied integer.
- `/terminal` is still gated by `WebSocketOriginGuard` + `WebAuthGuard` (`TerminalWebSocket.swift:22-37`), untouched.
- The previous round's amplification concern is properly closed client-side: `HYDRATE_MIN_INTERVAL_MS` caps a client at one 50 k capture per 5 s.

**Concerns:** none new. The absent server-side floor on `select-window` remains pre-existing and post-auth; agreed it belongs in its own ticket.

### Consider

**🟢 Gate on a real scroll gesture rather than `onScroll`.** `enableWheelScroll` (`app.js:4462-4472`) and `enableTouchScroll` (`app.js:4394-4414`) are the only paths a user reaches the top by. A seq counter bumped there — the same shape as `userInputSeq` — would let `maybeHydrateScrollback` require an actual gesture, which is what the comment already claims the trigger means. That removes the residual you named (one full capture per 5 s, ~40 KB/s plus two tmux spawns, indefinitely, on a tab merely parked at the top of a sparsely-printing shell) rather than bounding it, and it makes the cooldown a backstop instead of the load-bearing brake.

**🟢 `resetScrollbackSync()` doesn't clear `lastHydrateAt`.** A tab switch inherits up to 5 s of cooldown from the previous surface, and since the only retry is another `onScroll`, a quiet new tab can silently skip its first scroll-to-top. Harmless in practice — the fresh attach's replay already made that buffer authoritative — but the omission looks accidental next to the other five fields.

### Summary Table

---

dgershman:
Round 2 addressed in 976bf02c. Three of the four actioned; **Yellow 2 I could not confirm**, and I'd rather say so than delete working code on an unverified premise.

**🟡 settle deadline — fixed, and your diagnosis of why the suite couldn't see it was exact.** Confirmed by reading it back: every frame extended by 150 ms with no cap, frames aren't batched, so a shell printing faster than the quiet window left `hydratingScrollback` true and `scrollbackDirty` false for the life of the socket — the feature silently off on its own use case. `scheduleHydrateSettle` now clamps every extension to `HYDRATE_MAX_FLIGHT_MS` from the arm. I also extracted the per-frame bookkeeping as `noteTerminalFrame()` and drive it directly from the suite, since modelling a frame as `scrollbackDirty = true` took the `else` branch — the one that never runs during a flight.

**🟢 gesture gate — taken, with one change.** I used a *transition* (`lastViewportY !== 0 → 0`) rather than a seq counter on the wheel/touch handlers. Same effect for those two, but it also covers the scrollbar and Shift+PageUp, which xterm scrolls internally and which a counter in `enableWheelScroll`/`enableTouchScroll` would have missed — that would have quietly broken hydration for keyboard users. This removes the residual I flagged rather than bounding it, and the cooldown is now a backstop as you wanted.

**🟢 `lastHydrateAt`** — cleared in `resetScrollbackSync()`. It was an oversight, not deliberate.

Also tightened the latch while I was in there: it now only trips when a replay actually landed. An empty flight proves nothing about whether the pane has more history, and latching on it would have suppressed the next legitimate sync.

---

**🟡 "the restore is a no-op" — not actioned, because I couldn't verify it.**

I tried to settle it properly rather than by re-reading the minified source: loaded the bundled `xterm.js` into jsdom, wrote 300 lines, `scrollToTop()`, then wrote `replayFrame`'s exact prefix plus a deeper history, and watched `viewportY`. **The harness isn't faithful.** In it, even the plain user-level `scrollToTop()` at step 2 failed to move the viewport (`viewportY` stayed at 291) — which is not real-browser behaviour, so nothing it reports about `ydisp` after a rebuild can be trusted. jsdom has no layout, and xterm's Viewport syncs `ydisp` against a scrollable element's `scrollTop`. I've deleted the probe rather than commit a misleading test.

Your reasoning is plausible and may well be right — `scrollToTop()` is `scrollLines(-ydisp)`, and `scrollLines` returns early when `ydisp === 0`, so *if* the rebuild leaves `ydisp` at 0 the restore is indeed inert. What I can't confirm from here is the premise, and the failure mode of being wrong is asymmetric: if `ydisp` is ever non-zero at settle on a real surface, deleting the restore drops the user at the bottom of a 20k-line buffer they just scrolled up to read, which is the original bug wearing a different hat.

So I've left the machinery and **corrected the dishonest part instead** — you were right that `fake.scrolledToTop === 1` couldn't distinguish "restored" from "called a no-op". The fake now records `restoreCalls` and the assertions claim only what it can witness: the restore is *invoked*, and typing suppresses it. No assertion pretends to know the viewport moved.

If you can confirm the no-op against a real browser, I'll happily drop the restore, `hydrateSawReplay`'s restore role, `hydrateInputSeq` and `userInputSeq` in a follow-up — it's a real simplification. I just don't want to make it on my own unverified say-so. Note the deadline fix above stands on its own either way, so the two findings are no longer coupled.

Suite is 30 cases, up from 24. `row.test.js` unchanged and still pre-existing.

dhilgaertner (changes requested):
## Code & Security Review

Round 2's three actioned findings all hold up, and I verified them rather than taking them on trust:

- **Settle deadline.** `scheduleHydrateSettle` now clamps every extension to `hydrateArmedAt + HYDRATE_MAX_FLIGHT_MS` (`app.js:4861, 4875`), and `hydrateArmedAt` is set *before* the arming call (`app.js:4926, 4930`), so the first settle is a real 2000 ms and every later one is bounded. The 200-frame case at `scrollback-hydrate.test.js:210` drives the real `noteTerminalFrame` and pins it — extracting that function was the right call.
- **Gesture gate.** The transition (`app.js:4906`) is a better answer than the seq counter I suggested, and for the reason you gave. I confirmed the scrollbar path is covered: `Viewport._handleScroll` fires `_onRequestScrollLines.fire(i)` with a bare number, wired to `super.scrollLines(e, !1)` — `suppressScrollEvent` false, so `BufferService.scrollLines` reaches `t || this._onScroll.fire(i.ydisp)`. A counter in the wheel/touch handlers really would have missed it.
- **`lastHydrateAt`** cleared in `resetScrollbackSync()` (`app.js:4900`), pinned at `scrollback-hydrate.test.js:266`.
- The tightened latch (`hydrateSawReplay &&` at `app.js:4868`) is a correct catch that wasn't in my findings.
- `scrollback-hydrate.test.js` is 30/30; every other suite passes. `row.test.js` 43 passed / 10 failed is **confirmed pre-existing** — I ran it against `origin/main`'s `app.js` in an isolated tree and got identical counts.

One finding remains: the one I couldn't substantiate last round. **I can now, and it resolves against keeping the restore.**

### Code Quality

**🟡 `app.js:4872-4874` — the restore is inert on the path that reaches it, and the only way to reach it non-inert is a case where it does the wrong thing.**

You were right to refuse the jsdom probe — it isn't faithful. I traced the shipped bundle (`CrowTerminal/Resources/xterm/xterm.js`, which `StaticAssets.swift:38` serves as `/xterm/xterm.js`) instead. Every step is a branch you can read:

1. `scrollToTop(){this._core.scrollToTop()}` → `scrollToTop(){this.scrollLines(-this._bufferService.buffer.ydisp)}`.
2. `scrollLines(e,t){this._viewport?this._viewport.scrollLines(e):super.scrollLines(e,t)…}` — a rendered terminal always has a viewport, so this never reaches `BufferService`.
3. `Viewport.scrollLines(e)` is `setScrollPosition({scrollTop: t.scrollTop + e*cellHeight})`. With `e === -0` that is the *same* scrollTop → no DOM scroll event → `_handleScroll` never runs → `ydisp` untouched. **No-op.**
4. And `ydisp === 0` is guaranteed at settle on the trigger path. `scrollLines` sets `isUserScrolling=!0` on the scroll-up that lands at 0 (`if(e<0){if(0===i.ydisp)return;this.isUserScrolling=!0}`), which is what your `viewportY === 0 && baseY > 0` guards require. `isUserScrolling=!1` occurs in exactly two places — `BufferService.reset()` and `scrollLines` scrolling *down* to the bottom — and the replay does neither: `ESC[3J` is `trimStart(e),ybase=Math.max(ybase-e,0),ydisp=Math.max(ydisp-e,0),this._onScroll.fire(0)`, which never touches the flag. So through the whole rebuild `scroll()` takes `isUserScrolling||i.ydisp++` (no increment) and `isUserScrolling&&(i.ydisp=Math.max(i.ydisp-1,0))` — pinned at 0.

So on the intended path `scrollToTop()` does nothing, which is the good news you already suspected: the rebuild leaves the user on the oldest lines by itself.

The part that changes the recommendation is what happens when `ydisp` *isn't* 0 at settle. The only way there is the user leaving the top inside the ≤3 s flight, and the cleanest route is the #668 pill — `xterm-addon-crow-jumpbottom.js` calls `term.scrollToBottom()` and nothing else. That sets `isUserScrolling=!1` and `ydisp=ybase`; the remaining replay lines then track the bottom via `isUserScrolling||(i.ydisp=i.ybase)`. At settle `ydisp === ybase ≠ 0`, so `scrollToTop()` fires *for real* and drops the user from the live prompt to line 1 of a 20 000-line buffer. `userInputSeq` doesn't cover it — the pill is a click, not a keystroke.

That's the residual you flagged as "not covered" last round, and the no-op result reframes it: the restore never helps, and acts only when acting is wrong. Dropping it is strictly better than gating it. `hydrateInputSeq` (`app.js:4847, 4929`) and `userInputSeq` (`app.js:4335, 4338`) go with it; `hydrateSawReplay` stays, since the latch at `app.js:4868` still needs it, and so does the settle. If you'd rather keep the machinery, adding `buf.viewportY === 0` to the `app.js:4872` condition closes the harm case — but I'd take the deletion.

I should be straight about the limit of this: it's an exhaustive read of the shipped bundle, not a live-browser observation. It's the same class of evidence as the `BufferService.scroll` argument from round 1 that did hold up. If you want a browser check before deleting, that's reasonable — but the pill case is a bug either way.

### Security Review

**Strengths:**
- No new frame type, no new parsing, no new storage; `select-window` is pre-existing and `activeTerminal.window` is a daemon-supplied integer.
- `/terminal` still gated by `WebSocketOriginGuard` + `WebAuthGuard`, untouched.
- Round 1's amplification is now closed twice over: the transition gate means a top-parked viewport stops re-capturing entirely (`lastViewportY` is 0 once parked, so `arrivedAtTop` is permanently false), with `HYDRATE_MIN_INTERVAL_MS` as the backstop you intended.

**Concerns:** none new. The absent server-side floor on `select-window` stays pre-existing and post-auth; agreed it's its own ticket.

### Consider

**🟢 `app.js:4925` — an empty flight leaves `scrollbackDirty` false, which contradicts the latch's own reasoning two lines up.** Arming sets `scrollbackDirty = false` on the premise that we've synced. At settle you correctly withhold the latch when no replay landed — "an empty flight proves nothing" (`app.js:4867`) — but `scrollbackDirty` still asserts the opposite, so `app.js:4914` blocks every retry, including scrolling away and back. A quiet shell whose `select-window` went unanswered on a still-open socket stays un-synced until live output hits `app.js:4887` or a teardown calls `resetScrollbackSync()`. Rare and self-healing, but restoring `scrollbackDirty = true` when `!hydrateSawReplay` would make the two settle-time decisions agree.

**🟢 `app.js:4850` — `lastViewportY` isn't cleared by `resetScrollbackSync()`.** A new surface inherits the previous one's position as its "came from". It happens to be harmless: after `term.reset()` a fresh attach has `isUserScrolling === false`, so `ydisp === ybase`, and the one moment `viewportY` is 0 is also the moment `baseY` is 0, which `app.js:4917` blocks. But that safety rests on an xterm invariant nothing here states, and it's the same omission shape as `lastHydrateAt` last round — one line in the reset makes it uniform.

**🟢 `scrollback-hydrate.test.js:212` — `T.scrollbackDirty === true || T.scrollbackFullySynced === false` can't fail for the reason it's named.** Either disjunct alone passes it. It is meaningful today (the latch does trip in that case, so the first disjunct carries it), but as two separate checks it would say what it means — the same instinct as the `scrollbackFullySynced === false` assertion at line 174, which is doing exactly that job well.

### Summary Table

---

dgershman:
All four actioned in a5293612. You convinced me on the restore — the pill case is what did it.

**🟡 restore deleted.** I verified each branch against the shipped bundle before acting, rather than taking the trace on trust:

- `scrollLines(e,t){this._viewport?this._viewport.scrollLines(e):super.scrollLines(e,t)…}` — confirmed, a rendered terminal never reaches `BufferService`.
- `Viewport.scrollLines(e)` is `setScrollPosition({scrollTop: t.scrollTop + e*cellHeight})` — confirmed, so `-0` is the same scrollTop and no scroll event fires.
- The jump-bottom addon calls `term.scrollToBottom()` and nothing else — confirmed.

What changed my mind wasn't the no-op half (I'd suspected that and it's harmless on its own) but your point that **the only reachable-and-non-inert path is the one where it does damage**. The pill is a click, so `userInputSeq` never saw it — that's exactly the residual I flagged last round as "not covered", and you've shown it isn't a gap in the guard, it's the whole of what the code does. Gating it with `viewportY === 0` would have preserved machinery whose entire remaining behaviour is a bug. Deleted, along with `hydrateInputSeq` and `userInputSeq`. `hydrateSawReplay` stays for the latch.

The two restore tests are replaced by two that pin the new contract: a landed replay does not move the viewport, and the pill mid-flight is not overridden.

**🟢 empty flight leaves `scrollbackDirty` false.** Good catch, and worse than "rare" — the two settle-time decisions were in direct contradiction. Withholding the latch says "we don't know if we synced"; leaving the flag false asserts "we did". Fixed by restoring `scrollbackDirty = true` when `!hydrateSawReplay`, and pinned with a case that scrolls away and back to prove the retry actually gets through.

**🟢 `lastViewportY` not cleared in the reset.** Fixed. You're right that the existing safety rested on an unstated xterm invariant — and it's the same omission shape as `lastHydrateAt`, which is twice now, so the reset is the thing to keep uniform.

**🟢 the `A || B` assertion.** Fair — it couldn't fail for its stated reason. Split into the two things it was trying to say: the latch tripped (the replay did land), and later live output can still mark the buffer dirty.

On your caveat about evidence class: agreed it's a bundle read rather than a live-browser observation, but I don't think the decision turns on it any more. Even if `ydisp` were somehow non-zero at settle on the normal path, the pill case is still a bug, and the rebuild leaving the user on the oldest lines is exactly what the original acceptance criterion asks for. Deleting is correct under both readings, which is why I stopped holding out for a browser check.

32 cases. `row.test.js` unchanged and still pre-existing.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues

None.

### What I verified independently

The PR rests on a few load-bearing claims; I checked each against the source rather than the description.

- **A repeat `select-window` really does re-deliver the capture.** `TerminalWebSocket.swift:92-106` yields `cockpit.replayData(...)` unconditionally on every `select-window` frame — it does not compare against the currently selected window. So re-requesting on the same window is a real re-sync, not a no-op.
- **Re-requesting cannot stack duplicates.** `TerminalCockpit.replayFrame` (`TerminalCockpit.swift:141-147`) prepends `ESC[H ESC[2J ESC[3J`, so the reply rebuilds the buffer rather than appending to it.
- **The arrival-transition guard is grounded in the bundled xterm.** In `xterm.js`, `BufferService.scroll()` ends in an unconditional `this._onScroll.fire(i.ydisp)` and takes `isUserScrolling && (i.ydisp = Math.max(i.ydisp - 1, 0))` on a full buffer — so a top-parked viewport does fire `onScroll` on every output line while staying at `ydisp === 0`. `viewportY === 0 && lastViewportY !== 0` is the right shape for that, and `scrollLines` fires only on an actual `ydisp` change, so `lastViewportY` reliably holds the previous position.
- **The scrollbar path reaches the handler.** `Viewport.onRequestScrollLines` → `super.scrollLines(e, !1)` passes `suppressScrollEvent = false`, so a scrollbar drag to the top triggers the re-sync, not just wheel/touch (which go through `term.scrollLines` in `enableWheelScroll`/`enableTouchScroll`).
- **The corrected comment at app.js:4258-4261 is accurate.** `crowWrite`/`crowInput`/`crowReady`/`crowResize` appear only inside `CrowTerminal/Resources/xterm/terminal.html` itself — no caller anywhere in the repo.
- **The scoping decision holds.** `web/terminal.html` contains no `select-window` at all, so there is genuinely nothing to keep in sync there.
- **No handler leak.** `ensureTerminal` is `if (term) return` guarded, so `term.onScroll(maybeHydrateScrollback)` registers exactly once.
- **The state machine cannot wedge.** Every `scheduleHydrateSettle` clamps to `min(ms, armedAt + HYDRATE_MAX_FLIGHT_MS - now)`, so the flight always ends within 3 s of arming; `resetScrollbackSync` is the only other exit and clears the flag directly. There is no path that leaves `hydratingScrollback` true.

### Security Review

**Strengths:**
- No new endpoint, no new message type, no new parsing. The re-sync reuses the existing `select-window` control frame, which already sits behind `WebSocketOriginGuard` + `WebAuthGuard` on the `/terminal` upgrade.
- `window` is an `Int` taken from the daemon's own terminal record and decoded as `Int?` server-side, then reaches tmux through `controller.run([...])` — an argv array, not a shell string. No injection surface.
- No DOM string interpolation, no clipboard/storage access, no credential handling added.

**Concerns:**
- None. The one thing worth naming is server work per gesture (a `select-window` + `capture-pane` spawn), and that is bounded client-side by `HYDRATE_MIN_INTERVAL_MS`, the arrival transition, the dirty flag, and the `scrollbackFullySynced` latch. An authenticated client could already send unlimited `select-window` frames directly, so this does not widen anything.

### Code Quality

Green — consider, none blocking:

1. **The settle's `baseY` comparison races xterm's parser.** `term.write()` is asynchronous: `noteTerminalFrame()` (app.js:4693) runs at enqueue time, but xterm's `WriteBuffer` parses a multi-megabyte replay in ~12 ms slices across many event-loop turns. So at settle time `buf.baseY` may not yet reflect the rebuilt buffer, and `buf.baseY <= hydrateBaseY` (app.js:4866) is closer to advisory than to a measurement. Both outcomes are safe — a false latch suppresses redundant captures for a sync that *did* land, and a false non-latch just defers to `scrollbackDirty` — so this is not a defect. Worth a word in the comment, which currently reads as though the check observes a settled buffer.

2. **`hydrateSawReplay` is set by any frame, not only the replay.** On a shell already at the 50 000-line cap, `baseY` stops growing, so live output arriving during a flight can satisfy both `hydrateSawReplay` and `baseY <= hydrateBaseY` and latch `scrollbackFullySynced` on a flight that never received a replay. It self-heals as soon as the viewport leaves the top, so the blast radius is one missed re-sync — but the name promises more than the check delivers.

3. **One bare timing literal.** `scheduleHydrateSettle(2000)` (app.js:4929) is the only one of the four timings not named, next to `HYDRATE_MIN_INTERVAL_MS` / `HYDRATE_QUIET_MS` / `HYDRATE_MAX_FLIGHT_MS`. A `HYDRATE_FIRST_SETTLE_MS` would keep the whole timer story readable in one place.

4. **Follow-up, not this PR: the re-sync is all-or-nothing.** Every hydrate ships the full `capture-pane -S -50000` blob and re-parses the entire buffer, and the rebuild drops the user on absolute line 1 — past the ~19 500 lines they had not seen. That is the correct scope here (reusing the existing replay rather than inventing a protocol), and it does satisfy the ticket: all content becomes reachable. But on a remote or mobile client the repeat cost is a multi-MB transfer plus a full re-parse per scroll-to-top, and being dropped at the very beginning is a surprising landing spot. An incremental capture from the known depth would fix both; worth a ticket.

### Testing

- `scrollback-hydrate.test.js`: **32/32 pass.** Good shape — it drives the real `maybeHydrateScrollback`/`noteTerminalFrame`/`resetScrollbackSync` against a faked clock and faked timers rather than stand-ins, which is exactly what let rounds 1–2 surface the re-capture loop and the extend-forever settle. The `fake.moves` counter pinning "the settle never touches the viewport" is a nice way to make round 3's removal a regression test rather than a deletion.
- Full web suite: every file green except `row.test.js` at 43 passed / 10 failed. **I confirmed these are pre-existing** by checking out `main`'s `app.js`, re-running `row.test.js` (identical 43/10), and restoring — they are session-row glyph assertions, untouched by this diff.
- Not covered, and reasonably so: the async-`write` ordering in item 1 and real-xterm buffer behavior, neither of which a jsdom harness can reach.

### Summary Table

The diagnosis is the strongest part of this PR: the issue named two limits, both were already correct, and the author measured the actual cause (a browser-paced reader receiving 337/20 000 lines) before writing code. The three review rounds each removed a way the mechanism could misbehave — the re-capture loop, the unbounded flight, and a viewport restore that only fired when firing was wrong — and the guard stack that survived is tight: an arrival transition, a dirty flag, a cooldown, a sync latch, an absolute deadline, and teardown parity across both socket paths. Scoping to `app.js` alone is correctly justified, not a shortcut.

---

PR #937, opened 2026-08-04
Give the web client real URLs: hash-based routing (CROW-936)

### LINKED ISSUE #936: Web UI: URL doesn't change on navigation — no deeplinking or bookmarking
## Problem

Navigating around Crow's web UI — selecting a session, switching terminals/tabs, opening the Ticket Board, Reviews, Settings, Jobs — never changes the browser URL. Everything renders from in-memory state at a single static URL.

Consequences:

- **No deeplinking.** You can't send someone (or yourself) a link to a specific session or screen.
- **No bookmarking.** A bookmark always drops you at the default view.
- **Back/forward are broken.** Browser history doesn't reflect in-app navigation; Back exits the app instead of going back a screen.
- **Reload loses your place.** Refreshing returns to the default view rather than where you were.

## Ask

Add real URL routing to the web client so app state is reflected in the address bar and is restorable from a cold load.

## Suggested route shape (open to alternatives)

```
/                                  → default / session list
/sessions/:sessionId               → a session
/sessions/:sessionId/t/:terminalId → a specific terminal within a session
/tickets                           → Ticket Board
/reviews                           → Reviews board
/jobs                              → Jobs
/settings/:tab                     → Settings, per tab
```

Session and terminal IDs are already UUIDs, so they're stable identifiers to route on.

## Requirements

- Navigation **pushes** history entries; Back/Forward move between views rather than leaving the app.
- Loading a URL cold (fresh tab, no prior state) resolves to the right view once state loads.
- **Unknown or stale IDs degrade gracefully** — a session that's been reaped by the cleanup retention window, or an ID from another machine, should land on a clear "not found" state or the default view, not a blank screen or a crash. Sessions are routinely deleted by the retention reaper, so dead links are the normal case, not the edge case.
- Transient UI state (scroll position, open menus) does **not** belong in the URL — only the addressable view.
- Decide and document whether this is History API or hash-based routing; if the web client is served in a way that can't rewrite server-side routes, hash routing may be the pragmatic choice.

## Scope note

This is the **web client** (`web/`). The desktop app is native and doesn't have an address bar, so it's out of scope for URL semantics — but if session/terminal selection logic is shared or mirrored between the two front-ends, keep the change from drifting them further apart.

## Acceptance criteria

- [ ] Selecting a session changes the URL; copying that URL into a new tab lands on the same session.
- [ ] Selecting a terminal within a session is reflected in the URL and restores on load.
- [ ] Board/Settings/Jobs screens each have their own addressable URL.
- [ ] Browser Back/Forward navigate within the app correctly.
- [ ] Reloading the page keeps you on the current view.
- [ ] A URL for a deleted/unknown session shows a graceful not-found state, not a broken UI.

dgershman:
Closes #936

The web UI rendered every view from in-memory state at one static URL — nothing was linkable or bookmarkable, Back left the app, and reload dumped you at "Select a session".

## The ticket's open decision: hash routing

The ticket asked us to decide and document History API vs hash. **Hash**, documented in [ADR 0018](docs/adr/0018-web-client-hash-routing.md). `crowd` serves the client from a table of **exact literal paths with no catch-all**, so under History routing:

- an **authenticated** cold load of `/sessions/<id>` 404s in the Hummingbird router;
- an **unauthenticated** one renders the login page, because `WebAuthMiddleware` answers any navigational GET with a 200 — the deep link works only when you're logged out;
- the Tauri shell's `window.location.reload()` (`crow-desktop/src-tauri/src/lib.rs`) hits the same 404.

A fragment never reaches the server, so `/` always resolves. **No Swift file changes.**

## Routes

```
#/                                   home / empty state
#/sessions/:sessionId
#/sessions/:sessionId/t/:terminalId
#/tickets  #/reviews  #/allowlist  #/scorecard
#/settings/:tab
```

Two corrections to the ticket's suggested shape:
- **Four boards, not two** — `allowlist` and `scorecard` are boards too.
- **`/jobs` isn't a screen.** Jobs is a *Settings tab* → `#/settings/jobs`. (The "Jobs" in the sidebar is a session-group heading.)

## Approach

Routed navigation goes **through** the existing `selectSession` / `selectBoard` / `switchTerminal` / `openSettings` entry points rather than beside them, so a URL and a click can't disagree. `navigate()` no-ops when the hash already matches — that one guard is what stops *applying* a route from pushing a second history entry, which is what makes Back work in one press.

The terminal segment is written only by `switchTerminal`. Opening a session leaves `#/sessions/<id>`, which reloads onto `terminals[0]` — the same tab — so nothing is lost and clicking a session doesn't bury history under an id the user never picked.

## Dead links are the normal case

The retention reaper deletes completed sessions, so a shared URL has a short shelf life by design. Previously a stale id left a **blank pane** — `renderHeader(undefined)` returns early while `has-selection` hides the empty state. Now it's an explicit not-found card with a way back.

The same guard also catches a session reaped **while you're viewing it**, which nothing checked before (blank header + orphan `+` tab + frozen terminal). Its helper replaces the near-identical cleanup block that `deleteSession` and `bulkDeleteSelected` each carried.

Deep links also survive login: the middleware serves the login page *at the requested URL*, so the fragment is still in the address bar and `login.html` hands it back.

## Tests

`web-tests/router.test.js` — 70 assertions driving the **real** `app.js`: route shapes and the malformed ones that must degrade not throw; history semantics (push / no-op / replace); cold-load deferral until `sessionsLoaded` can tell "deleted" from "not loaded yet"; not-found on both the cold and live paths; terminal restore and URL correction for a dead one; Back applying without duplicates; plus drift gates pinning the two lists the router has to duplicate (no modules, two classic scripts).

Two bugs a self-review caught, both now covered: `app.js` boots *before* `settings.js` defines `openSettings`, so a cold `#/settings/*` link was a silent no-op (now deferred to `DOMContentLoaded`); and `selectSession` wrote the hash before `refreshTerminals` consumed the routed terminal, rewriting `#/sessions/x/t/y` down to `#/sessions/x`.

Also verified out-of-band by loading the real `index.html` + `app.js` + `settings.js` together: both evaluate clean, `#/tickets` applies `board-active` from the URL, and the real `settings.js` opens on the Jobs tab.

## ⚠️ CI wiring — please read

The jsdom suite **ran in no CI job at all** until now; it's added to the `parity` job, which already needs node.

`test:ci` deliberately omits **`row.test.js`, which has 10 failures already on `main`** — CROW-802 drift that the file's own comment describes (PR status parts became SVG `.ico` spans; the `r.glyphs` assertions still read text-only `.pr-ico`). Gating on the full suite would fail every PR. Fixing it means deciding what those icons *should* render, which is a change to `row`, not to this ticket — so I left it alone rather than quietly rewriting someone else's assertions. `npm test` still runs everything; the exclusion is documented in `web-tests/README.md` so it can't quietly become permanent.

## Not done

No browser click-through — there's no built `crowd` in this worktree and no browser automation here. Everything above is verified under jsdom against the real files. To check by hand:

```sh
make daemon-run   # then click a session, switch tabs, Back/Forward, reload
                  # and paste #/sessions/00000000-0000-0000-0000-000000000000
```

I also skipped the ADR-0013-style Swift drift guard in `WebTerminalAssetTests`: `CrowDaemon` is Darwin-only and absent from the PR lane, so it would never run on a PR — whereas the jsdom test now does.


dhilgaertner (changes requested):
## Code & Security Review

Strong PR. The hash-vs-History analysis is correct and load-bearing — I checked `StaticAssets.swift` and `WebAuthMiddleware`, and the described inversion (authenticated deep load 404s, unauthenticated one renders login with a 200) is real, so the decision is well grounded. Routing *through* `selectSession`/`selectBoard`/`switchTerminal`/`openSettings` rather than beside them is the right shape, and the `navigate()` no-op-on-match guard is exactly what makes Back a single press. ADR 0018 is one of the better ones in the repo.

I verified the PR's own claims rather than taking them on trust:
- `npm run test:ci` → **exit 0**, router suite **70/70**.
- `npm test` → exit 1, and I confirmed the cause is genuinely pre-existing: running `row.test.js` against `git show main:…/app.js` gives the **same 43 passed, 10 failed**. The exclusion is honest and correctly documented.
- No Swift changes, as claimed.

### Critical Issues

None. No Red findings.

### Security Review

**Strengths:**
- **No open redirect** in the new login hop. `login.html:56` builds `'/' + (location.hash || '')`; `location.hash` is always `#`-prefixed or empty, so the result can never become protocol-relative (`//evil`) or absolute. The `location.pathname === '/'` → `reload()` branch is the correct handling of the assign-to-same-URL no-op.
- **No attacker-controlled id reaches an RPC.** `applyRoute` (`app.js:5442`) checks membership in `sessions` before calling `selectSession`, and `pendingTerminalId` is only ever matched against the `list-terminals` response — a crafted `#/sessions/<anything>` can't turn into a `session_id` on the wire. That's a deliberate-looking defensive ordering and it holds.
- **No XSS surface.** The routed id is rendered via `textContent` in `showSessionNotFound`; `routeToHash` uses `encodeURIComponent`; `parseRoute` wraps `decodeURIComponent` in try/catch and returns `null` on a malformed `%`-escape (covered by a test).
- CSP untouched — the router lives in `app.js`, no inline bootstrapping, consistent with the ADR's stated reason for not splitting it into its own file.

**Concerns:** none found.

### Code Quality

**🟡 Yellow — Back between Settings tabs silently discards unsaved edits.**
`applyRoute` routes *every* `#/settings/*` hash through `window.openSettings` (`app.js:5431`), and `openSettings` re-entry (`settings.js:139`) unconditionally re-runs `get-config`, overwrites `cfg`, and sets `dirty = false` / `subForm = null` (`settings.js:163`).

Repro: open Settings → Workspaces, edit a field, click the Automation tab (pushes an entry), press Back. `openSettings('workspaces')` re-fetches, the working copy is replaced, Save goes disabled — no prompt. Clicking between tabs preserves `cfg`; Back destroys it. Same intent, two different outcomes, and the destructive one is silent.

The related `closeSettings(true)` on route-away is at least a *documented* choice, but it compounds this: Back is now the most natural way to dismiss the modal, and it's the one path that never prompts while Close/Esc does.

Cheapest fix that removes the surprising half: when `settingsIsOpen()` is already true and only the tab differs, switch the tab in place instead of re-entering `openSettings` — no re-fetch, no `dirty` reset.

**🟡 Yellow — the `route-missing` state shows a hamburger that can't do anything (mobile).**
`app.css:490` deliberately reveals `#back-to-sidebar` in the not-found state, but nothing can act on it: the only rule that un-hides the sidebar under `mobile-show-sidebar` is `#app.has-selection.mobile-show-sidebar #sidebar` (`app.css:496`), and `showEmptyDetail` removes `has-selection`. So `#app.route-missing #sidebar { display: none }` (`app.css:489`, specificity 2-1-0) keeps winning. Tapping the button adds the class and reveals only the status bar.

Users aren't trapped — the "Back to sessions" button works — but this is a control the new CSS goes out of its way to display, in the exact scenario the comment calls out as most likely on mobile. Either add `#app.route-missing.mobile-show-sidebar #sidebar { display: block; }` (plus hiding `#detail`), or drop line 490 and let the in-card button be the only affordance.

**🟢 Green — `applyBootRoute` re-implements a check `applyRoute` already does.**
`app.js:5501` sets `pendingRoute = route` unconditionally for a session route, but `applyRoute` already defers on `!sessionsLoaded` (`app.js:5443`). Since `refreshSessions()` is kicked off synchronously at boot while `applyBootRoute` may be deferred to `DOMContentLoaded`, a first `list-sessions` that resolves before the parser finishes `settings.js` leaves `pendingRoute` set *after* `sessionsLoaded` is already true — the deep link then does nothing until the 10s poll, and lands on top of whatever the user selected in the meantime. In practice `settings.js` is a local static file and almost always wins the race, so this is latent rather than live — but `else applyRoute(route)` for the session case too is both shorter and correct by construction. (The jsdom harness can't catch it: `readyState` is already `complete`, so the deferred path is never exercised.)

**🟢 Green — every Settings tab click pushes a history entry.**
Open Settings and browse six tabs and it's eight Back presses to leave. ADR 0018 argues against turning Back into "an undo stack for UI noise"; tab switches inside an already-open modal are arguably that. `{ replace: true }` for in-modal tab changes would keep each tab addressable without the depth.

**🟢 Green — CI now depends on an unpinned dependency.** `package-lock.json` is git-ignored, so the added `npm install` step resolves `jsdom: ^24.0.0` fresh on every run. That was harmless while the suite was local-only; now a jsdom 24.x patch can turn an unrelated PR red. Consider pinning to an exact version or un-ignoring the lockfile.

**Praise, briefly:** the not-found guard catching a session reaped *while open* is a genuine pre-existing bug fixed in passing, and folding `deleteSession`/`bulkDeleteSelected`'s duplicated cleanup into `showEmptyDetail` is the right call. The two drift gates (settings `TABS`, the `selectedBoard` union) are a good answer to "two classic scripts, no modules". The two self-caught bugs called out in the description — cold `#/settings/*` firing before `settings.js` loads, and `selectSession` flattening `/t/<id>` before `refreshTerminals` consumes it — are both the kind that would have shipped, and both now have tests.

### Summary Table

Both Yellows are small and local. The Settings/Back data loss is the one I'd genuinely like fixed before merge; the mobile hamburger is a two-line CSS call (fix it or stop showing the button).

---

dgershman:
Thanks — the verification of the PR's own claims (re-running `row.test.js` against `git show main:…/app.js` to confirm the 43/10 baseline) is exactly the right way to check that exclusion, and I appreciate it.

All five findings addressed in 362695c5. Nothing pushed back on.

### 🟡 Back between Settings tabs discarded unsaved edits

You're right, and the asymmetry was the worst part — clicking a tab preserved `cfg`, arriving at the same tab via Back destroyed it, and only the destructive path was silent.

Took the fix you suggested. `settings.js` gains `setSettingsTab()`, which moves the tab in place — no `get-config`, no `dirty`/`cfg` reset — and `applyRoute` uses it whenever `settingsIsOpen()` is already true, falling back to `openSettings` only for a genuine cold open.

This one needed a better test than I had: `router.test.js` was stubbing `window.openSettings`, which can't observe a re-fetch. It now has a loader that runs the **real** `settings.js` alongside `app.js` the way `index.html` does, counts `get-config` calls, and asserts the second one never happens.

### 🟡 `route-missing` hamburger revealed nothing on mobile

Confirmed your specificity analysis — `#app.route-missing #sidebar { display: none }` (0,2,1,0) beat everything, since the only un-hiding rule is scoped to `.has-selection`, which `showEmptyDetail` removes.

Chose "make it work" over "stop showing it": on mobile the hamburger is *the* way back to the list everywhere else in the app, so removing it here would be the inconsistency. Added the `.mobile-show-sidebar` trio mirroring the `has-selection` block. Verified by evaluating the real `@media` block under jsdom rather than by eye:

```
route-missing:                        sidebar none  | detail flex | hamburger flex
route-missing + mobile-show-sidebar:  sidebar block | detail none | hamburger none
```

### 🟢 `applyBootRoute` re-implementing `applyRoute`'s check

Fixed as suggested — every non-home route now goes through `applyRoute`, which owns the `!sessionsLoaded` decision. Correct by construction whichever side of the race wins, and shorter.

You're also right that the harness can't reach it (`readyState` is `complete` under jsdom), so I tested the property it now leans on instead: `applyRoute` with sessions already loaded resolves immediately and leaves `pendingRoute` null.

### 🟢 Every Settings tab click pushed a history entry

Fair — and it contradicted ADR 0018's own "not an undo stack for UI noise" line, which is the more embarrassing part. In-modal tab changes now use `{ replace: true }`; each tab stays addressable, browsing six costs one Back press. Asserted.

### 🟢 CI depending on an unpinned dependency

Real regression in risk that I introduced by wiring the suite into CI. Un-ignored `package-lock.json` and switched the step to `npm ci` — that pins transitives too, which an exact version in `package.json` wouldn't.

---

`npm run test:ci` → exit 0, router suite **84/84**, deterministic across repeated runs. Still no Swift changes.

dhilgaertner (changes requested):
## Code & Security Review

Verified the ADR's premises directly rather than taking them on trust: `StaticAssets.mount` really is a table of exact literal paths with no catch-all and no `FileMiddleware` (`StaticAssets.swift:16-48`), and `serveLoginPageForUnauthorized` really is `method == .get && accept.contains("text/html")` (`WebAuthRoutes.swift:74-92`) — so the "authenticated deep link 404s, unauthenticated one renders login" inversion is real and hash routing is the right call here. The `row.test.js` claim also checks out: neither `row.js` nor `row.test.js` is touched by this PR and it fails 10/53 on this branch, so the `test:ci` exclusion is documenting a pre-existing state, not hiding a regression introduced here. `npm ci && npm run test:ci` → **84 passed, 0 failed**.

The routing design is sound — going *through* `selectSession`/`selectBoard`/`switchTerminal`/`openSettings` instead of beside them, and the `navigate()` no-op guard as the single thing that keeps Back to one press, is the right shape. Two behavioural gaps below, both reproduced against the real `app.js` under the same jsdom harness.

### Critical Issues

None.

### Code Quality

**🟡 `pendingRoute` is never invalidated, so a deferred deep link overrides a choice the user made after it** — `app.js:5448` (set) / `app.js:1056-1061` (consumed)

`applyRoute` stashes a session route while `!sessionsLoaded`, but nothing clears it if the user navigates in the meantime. The sidebar is deliberately painted early from the localStorage cache (CROW-613), so its board pills and session rows are clickable during exactly that window — and on a remote/proxied session the window is not short.

Reproduced by driving the real `app.js`:

```
boot: pendingRoute = {"view":"session","sessionId":"sess-1"} hash = #/sessions/sess-1
after click: selectedBoard = tickets   hash = #/tickets
after list-sessions: selectedBoard = null | selectedId = sess-1 | hash = #/sessions/sess-1
>>> user yanked off the board they chose
```

Cheapest fix is one line in `navigate()` — when it actually writes a hash, `pendingRoute = null`; the user has superseded it. Alternatively gate the consumption on `routeToHash(route) === (location.hash || '#/')`.

**🟡 Back out of Settings silently discards unsaved edits; Esc / ✕ / backdrop-click prompt** — `app.js:5442`, `settings.js:183-184`

`applyRoute` closes with `closeSettings(true)`, and `force` is precisely the flag that skips `confirmModal('Discard unsaved changes?')`. Every other exit from the modal (`settings.js:229, 236, 247, 277`) goes through the unforced path and prompts. So the same destination reached two ways has two outcomes, and the destructive one is the one with no prompt — which is the exact bug class this PR already fixed for `setSettingsTab` (and described that way in `settings.js:171-174`).

Before this PR, Back left the site entirely, so losing the edits at least came with an obvious page transition. Routing turns Back into a cheap in-app gesture, which is what makes the silent discard newly costly. The comment ("a route change is not the moment to prompt") is a defensible position, but it should at least be symmetric with ✕.

Workable fix: make `closeSettings` return `false` when the user cancels, call it unforced from `applyRoute`, and on refusal restore the URL with `navigate({ view: 'settings', tab: activeTab }, { replace: true })` and return without applying the route.

**🟢 Re-clicking the already-open session row drops `/t/<id>` while that tab stays active** — `app.js:2287`

`selectSession` navigates with `pendingTerminalId || null`, which is `null` for an ordinary click, so re-selecting the current session rewrites `#/sessions/A/t/t2` → `#/sessions/A` *and* pushes a history entry, while `refreshTerminals` keeps `t2` active (confirmed: `hash = #/sessions/A`, `activeTerminal = t2`, +1 history entry). The URL you'd copy at that moment no longer names the tab you're looking at. Guarding with `id === selectedId ? (activeTerminal && activeTerminal.id) : pendingTerminalId` — or returning early on re-select — covers it. Same latent shape in `navPill('Manager', …)` at `app.js:1454`.

**🟢 A terminal that disappears mid-session leaves a stale `/t/<id>` behind** — `app.js:2834`

The URL correction is inside the `if (pendingTerminalId)` block, so it only ever fires on the routed pass. If the tab you're on is closed from another client, `activeTerminal` falls back to `terminals[0]` but the URL keeps naming the dead id until something else rewrites it. Correcting whenever the hash names a terminal absent from `terminals` would close it.

**🟢 `switchTerminal` with no selection would route home** — `app.js:2905`

`routeToHash({ view: 'session', sessionId: null, … })` falls through to `'#/'`, so the tab switch would navigate to home. Looks unreachable today (tabs only render under a selection), but an early return on `!selectedId` is a cheap guard.

**🟢 Unawaited `applyRoute` in `refreshSessions`** — `app.js:1059`

`applyRoute` is `async` and a rejection from `selectSession` here becomes an unhandled rejection instead of being swallowed by the surrounding `catch` (which has already been passed by that point). A `.catch(() => {})` matches the "transient — next poll retries" intent of the enclosing block.

**🟢 The `parity` job relies on the runner's ambient Node** — `.github/workflows/ci.yml:147-152`

Committing the lockfile and using `npm ci` is the right instinct for a step that now gates every PR, but the Node version is still whatever `ubuntu-latest` ships that month. An `actions/setup-node` step with a pinned major finishes the job the lockfile started.

### Security Review

**Strengths:**

- No injection surface from route data. The session id reaches the DOM only via `textContent` (`app.js` `showEmptyDetail`), never `innerHTML`, and `applyRoute` validates the id against `sessions` before it reaches `selectSession` / `rpc`.
- `routeToHash` uses `encodeURIComponent` on both id segments, `parseRoute` wraps `decodeURIComponent` in try/catch so a malformed `%`-escape returns `null` rather than throwing (covered at `router.test.js:182`), and board/tab values are checked against `ROUTE_BOARDS` / `ROUTE_SETTINGS_TABS` allowlists with an unknown tab degrading to `general`.
- No open redirect in the login hop. `location.hash` is always `#`-prefixed, so `'/' + hash` (`login.html`) and `'/login' + hash` (`app.js:1363`, `app.js:2372`) stay same-origin; a crafted `#//evil.com` parses to an unknown route and is normalized to `#/` at boot. The `location.pathname === '/'` branch is also correct — a fragment deep link keeps `pathname === '/'`, so `reload()` is the right call and preserves the fragment.
- Zero Swift changes means the auth gate, the `index.html`-keyed CSP, and `isSafeAssetName` are all untouched — the claim in the description holds.
- The committed `package-lock.json` + `npm ci` narrows supply-chain drift on a step that now gates every PR.

**Concerns:** none found.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 6 Green] findings.

Both Yellows are small, local fixes, and the architecture underneath them is right — this is a good change with two rough edges at the boundaries where routing meets state that predates it.

---

dgershman:
All eight addressed in 5c2c526a — no pushback. The two Yellows were both real and both reproducible; thanks for driving them against the real `app.js` rather than describing them, the traces made them unambiguous.

### 🟡 `pendingRoute` never invalidated

Confirmed. Took the one-line fix: `navigate()` clears `pendingRoute` whenever it *actually writes*. A real navigation supersedes a link still waiting on `list-sessions`.

Deliberately **not** cleared on the no-op path — if the hash already matches, nothing was superseded and the deferred route should still stand (boot at `#/sessions/A`, user clicks the same row: `navigate` no-ops, the drain later applies the same thing).

Test drives your exact repro: defer → click a board → `list-sessions` lands → still on the board, `selectSession` never called.

### 🟡 Back out of Settings discarded silently while ✕ prompts

You're right, and the sting is that it's the same asymmetry I'd just fixed for `setSettingsTab` and then described in a comment three lines away.

Took your suggested shape exactly: `closeSettings` returns `false` on cancel, `applyRoute` calls it **unforced**, and on refusal restores the URL with `{ replace: true }` and abandons the route. Verified both branches — cancel keeps the modal up with `#/settings/automation` restored and the board not applied; accept dismisses and applies.

The test flips a real toggle and fires its own `onchange` rather than reaching into settings.js's closure for `dirty`, so it exercises the shipped `markDirty` path.

### 🟢 Re-select drops `/t/<id>`

Fixed — `selectSession` carries the active terminal when `id === selectedId`, which also covers the `navPill('Manager', …)` shape you flagged. Asserted, including that it pushes no entry.

### 🟢 Terminal vanishing mid-session leaves a stale `/t/<id>`

Fixed by moving the correction out of the `if (pendingTerminalId)` branch: it now runs on every pass, keyed on the hash naming a terminal absent from `terminals`.

One deviation worth flagging — it re-points at the tab actually shown rather than dropping the segment, so a URL copied at that moment still names what's on screen. That changed one existing expectation (`#/sessions/A` → `#/sessions/A/t/term-1`); I updated it rather than preserving the weaker behaviour. Say the word if you'd rather it drop to the bare session.

### 🟢 `switchTerminal` with no selection

Guarded. Kept it to the `navigate` call rather than an early return, so the tab still switches if it's ever reached — asserted both halves.

### 🟢 Unawaited `applyRoute`

`.catch(() => {})` on all three call sites, not just the one in `refreshSessions`.

### 🟢 Ambient Node

`actions/setup-node@v4` pinned to 22, placed immediately before the web-tests step so the shell gates above keep the environment they have today.

---

`npm run test:ci` → exit 0, router suite **101/101**, deterministic across repeated runs. Still no Swift changes.

dhilgaertner (changes requested):
## Code & Security Review

Third pass. Both prior rounds' findings are genuinely fixed — I re-ran the earlier repros and they're all closed: `pendingRoute` is now invalidated by `navigate()`'s write path (and correctly *not* on the no-op path), the Settings discard prompt is symmetric, `setSettingsTab` moves the tab without re-running `get-config`, and the `route-missing` mobile trio makes the hamburger actually reveal the sidebar.

Verified rather than assumed:
- `npm ci && npm run test:ci` → **101 passed, 0 failed** (84 last round).
- `row.test.js` → 43 passed, 10 failed, and `git diff main...HEAD` touches neither `row.js` nor `row.test.js` — the `test:ci` exclusion is still documenting a pre-existing state, not hiding anything.
- No Swift changes, as claimed.
- `list-sessions` (`RPCHandlers.swift:213`) maps `appState.sessions` with no status filter, so the new mid-session reap guard can't misfire on a session the user merely archived. I checked this specifically because it would have been a nasty false positive.

One blocker below. It's on the primary path and it breaks the feature the ticket is named for.

### Critical Issues

**🔴 Back cannot leave a session once you've clicked a terminal tab — the bare `#/sessions/:id` entry is a permanent wall** — `app.js:2292-2294`

`selectSession` carries the active terminal into the URL:

```js
const routedTerminal = pendingTerminalId
  || (id === selectedId && activeTerminal ? activeTerminal.id : null);
navigate({ view: 'session', sessionId: id, terminalId: routedTerminal });
```

That's right for a *click* (it's the round-2 fix for re-selecting the open row). But `applyRoute` reaches the same line when it's **applying** a route, and then the URL is authoritative — synthesizing a `/t/<id>` the route didn't ask for makes `navigate()` write instead of no-op, which is exactly the invariant ADR 0018 rests on ("`navigate()` is a no-op when the computed hash already matches, which is what keeps applying a route from pushing a second history entry").

Clicking a session always pushes bare `#/sessions/A`; clicking any tab then pushes `#/sessions/A/t/T2`. Back onto the bare entry calls `selectSession(A)` while `selectedId === A` and `activeTerminal` is still `T2`, so it immediately pushes `#/sessions/A/t/T2` again — truncating the forward entry and landing you where you started. The push means the wall never moves.

Driven against the real `app.js` under the `router.test.js` harness, with a faithful history model (Back moves the index left; a push truncates and appends):

```
history: ["#/tickets","#/sessions/sess-1","#/sessions/sess-1/t/term-2"] idx = 2

Back press #1:  landed on #/sessions/sess-1/t/term-2  idx = 2
Back press #2:  landed on #/sessions/sess-1/t/term-2  idx = 2
Back press #3:  landed on #/sessions/sess-1/t/term-2  idx = 2

Can Back reach #/tickets? NO — trapped in the session
```

`refreshTerminals`'s correction block can't rescue it: the hash names `term-2`, which *is* in `terminals`, so no replace fires.

Same trap on the way out of Settings — Back from `#/settings/general` to a `#/sessions/A` entry closes the modal, then bounces the URL to `/t/<active>`. And pasting a bare `#/sessions/A` link while already in that session on tab 2 silently rewrites itself.

The distinction the code needs is *applying* a route vs *initiating* one. Smallest shape that keeps the round-2 fix intact:

```js
// applyRoute
pendingTerminalId = route.terminalId || null;
await selectSession(route.sessionId, { fromRoute: true });

// selectSession(id, opts)
const routedTerminal = pendingTerminalId
  || ((opts && opts.fromRoute) ? null
      : (id === selectedId && activeTerminal ? activeTerminal.id : null));
```

Worth deciding alongside it whether applying a bare `#/sessions/A` should also snap the visible tab to `terminals[0]`. Otherwise Back leaves the URL bare while `T2` stays on screen, and a reload of that same URL lands on `T1` — URL and view disagreeing is the thing this PR exists to prevent, and "Back undoes the tab switch" is probably what a user expects anyway.

`router.test.js` has the right primitives for a regression test already; the gap is that the existing Back assertions never exercise *bare-session-preceded-by-a-terminal-entry*, which is the one sequence every session visit produces.

### Security Review

**Strengths:** unchanged from the last pass and still holding — `login.html:56`'s `'/' + (location.hash || '')` can't go protocol-relative (a non-empty `location.hash` is always `#`-prefixed), `applyRoute` checks session membership before any id reaches an RPC, `pendingTerminalId` is only ever matched against the `list-terminals` response, the not-found id is rendered via `textContent`, and `parseRoute` catches malformed `%`-escapes. The new surface since round 2 (`setSettingsTab`, `closeSettings`'s return value, the `route-missing` CSS) carries no security weight. CSP untouched.

**Concerns:** none found.

### Code Quality

Non-blocking, all four are "consider":

**🟢 A cancelled discard prompt destroys the entry Back had just moved to** — `app.js:5468`. On refusal `applyRoute` does `navigate({view:'settings'}, {replace:true})`, but the browser has already moved the index, so `replaceState` overwrites the *destination* entry rather than restoring the one you came from. Cancelling a discard at `#/settings/general` with `#/tickets` behind it leaves `#/tickets` gone from history. `routeBeforeSettings` gets the user back on eventual close, so it's recoverable — `history.forward()` is the alternative and has its own races, so this may just be worth a comment saying the trade-off was taken knowingly.

**🟢 `#/settings/<unknown-tab>` leaves a dead history entry** — `app.js:5455`. `parseRoute` degrades an unknown tab to `general` (good), but `openSettings` then `navigate`s to `#/settings/general`, which differs from the bogus hash and therefore *pushes*. Back returns to the bogus entry, `setSettingsTab` accepts it, and the address bar sits on a shape the app silently reinterprets. Normalizing with `{replace: true}` when the parsed tab differs from the raw segment would close it.

**🟢 Deleting the open session pushes `#/`** — `app.js:2809`, `app.js:1245`. Back then lands on the not-found card for the session you just deleted. `{replace: true}` reads better for a destination the user can never return to.

**🟢 Comment slightly overstates reach** — `app.js:2836-2839`. The terminal-correction block says it covers "the tab you were on closed from another client", but `refreshTerminals` isn't on any interval (only `refreshSessions`/`refreshLive`/`refreshArtifacts` are) — it runs on this client's own actions, so that case is corrected on the next local action rather than live. The code is right; the comment promises a bit more than it delivers.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 0 Yellow, 4 Green] findings.

Everything else here is ready. The ADR is accurate down to the `serveLoginPageForUnauthorized` predicate, the CI wiring is the right call (`npm ci` against a now-committed lockfile, node pinned, placed so the existing shell gates keep their environment), and the round-2 fixes are all real. It's the one line where a click's convenience leaks into a route application.

---

dgershman:
The 🔴 is real and it's mine — the round-2 fix for re-selecting the open row leaked click convenience into route application. Fixed in 77c18fd3, along with all four Greens.

### 🔴 Back walled inside a session

Your diagnosis is exactly right, including *why*: `applyRoute` reaches the same line as a click, and when applying a route the URL is authoritative. Synthesizing a `/t/<id>` it didn't ask for turns `navigate()` into a push — which is the invariant ADR 0018's Back guarantee rests on, so this wasn't a new trade-off, it was me breaking a stated one.

Took your suggested shape verbatim: `selectSession(id, { fromRoute })`, set only by `applyRoute`, synthesizing nothing when it's on.

**I wrote the test first and it reproduced your trace before the fix:**

```
✗ Back #1 lands on the bare session   got "#/sessions/sess-1/t/term-2"
✗ Back #2 escapes to the board        got "#/sessions/sess-1/t/term-2"
```

After: Back #1 → `#/sessions/sess-1`, Back #2 → `#/tickets`, session actually left.

It drives the real `app.js` through the sequence every session visit produces, with the history model you described (push truncates and appends, replace overwrites, Back moves the index left) — jsdom doesn't model Back against a vm context. You were right that the gap was *bare-session-preceded-by-a-terminal-entry*; none of my Back assertions had that shape.

**On your open question — yes, applying a bare route now snaps the view to `terminals[0]`.** Your reasoning decided it: leaving T2 on screen under a bare URL means reloading that same URL shows T1, and URL-and-view disagreeing is the thing this PR exists to prevent. "Back undoes the tab switch" is also what I'd expect as a user. Asserted both halves.

### 🟢 Unknown settings tab left a dead entry

Fixed — `applyRoute` normalizes the bogus hash in place before opening, so Back can't return to a shape the app reinterprets. Asserted it doesn't push.

### 🟢 Deleting the open session pushed `#/`

Fixed in both `deleteSession` and `bulkDeleteSelected`. Asserted.

### 🟢 Cancelled discard destroys the destination entry

Confirmed, and I kept `replaceState` — but added a comment saying so knowingly, which is what you suggested. `history.forward()` preserves the entry but no-ops when there's nothing forward (a hand-edited URL), leaving the address bar wrong while the modal is still up; that's the worse failure, and `routeBeforeSettings` makes the current behaviour recoverable.

### 🟢 Comment overstated reach

You're right — `refreshTerminals` appears in no `setInterval` (only `refreshSessions` / `refreshLive` / `refreshArtifacts` / the board poll), so a tab closed from another client is corrected on this client's next action, not live. Comment corrected; code unchanged.

---

Also added the applying-vs-initiating distinction to ADR 0018's Decision section — the Red was a violation of a guarantee the ADR already made, so the rule that protects it belongs next to it.

`npm run test:ci` → exit 0, router suite **112/112**, deterministic. Still no Swift changes.

dhilgaertner (changes requested):
## Code & Security Review

Third pass. The two prior rounds' findings all landed, and the reasoning in ADR 0018 checks out against the source it cites — I verified `StaticAssets.swift` really is a literal-path table with no catch-all (`/`, `/index.html`, `/login`, `/app.js`, `/app.css`, `/settings.js`, `/settings.css`, `/brand.svg`, `/version.json`, `/terminal.html`, `/xterm/:file`), and `WebAuthRoutes.serveLoginPageForUnauthorized` really does answer any `GET` + `Accept: text/html` with the login page at 200. The "History routing works only when logged out" inversion is real, so hash routing is the right call for a web-assets-only change.

One Yellow. Everything else is Green.

### Critical Issues

None.

### Security Review

**Strengths:**

- **No open redirect in the login hop.** `login.html:60` builds `'/' + (location.hash || '')`. `location.hash` is always `#`-prefixed or empty per spec, so the result can never become protocol-relative (`//evil.example`) — the classic `'/' + userInput` bug does not apply here. `location.replace` (not `href`) also keeps `/login` out of history.
- **No HTML injection in the not-found card.** The detail string goes through `textContent` (`app.js:2249`, `2253`). Verified with a markup payload under the PR's own jsdom loader: `innerHTML.indexOf('<') === -1`.
- **Malformed fragments degrade rather than throw.** `parseRoute` wraps `decodeURIComponent` in try/catch (`app.js:5416`) and returns `null`, which `applyRoute` normalizes to home with `replace`. `#/sessions/a/b/c/d`, `#/settings/bogus`, `#/nonsense` all land safely.
- **Supply chain pinned in the right direction.** The new CI step uses `npm ci` against a newly committed `package-lock.json` with `actions/setup-node@v4` at an explicit `node-version: '22'`. Since this step now gates every PR, floating deps or a floating runtime would have been the failure mode; both are closed.
- **No Swift changes**, so no server-side attack surface moved.

**Concerns:**

🟡 **`showSessionNotFound` echoes the whole fragment segment back unbounded** — `app.js:2278-2284`

The id is interpolated into the detail line with no length or shape check:

```js
detail: 'It may have been deleted — Crow removes completed sessions automatically.'
  + (id ? ' (' + id + ')' : ''),
```

Driving the real `app.js` under the suite's own loader with `#/sessions/<435 chars>`:

```
payload chars in URL : 435
rendered chars       : 511
markup escaped       : true
attacker text shown  : true
sample : "It may have been deleted — Crow removes completed sessions automatically.
          (Your Crow session expired. Re-enter your password at http://evil.example/ to continue. …"
```

Not XSS — `textContent` holds. But it is unbounded attacker-chosen prose rendered inside Crow's own empty-state chrome (CWE-451, content spoofing), and the value is *supposed* to be a UUID, so nothing is lost by bounding it. Two consequences:

1. A link to the victim's own Crow host can display arbitrary text under Crow's branding. Low value on its own, but the card's whole purpose is to be the thing a *shared, stale* link lands on — the one place a user is primed to read an explanation rather than distrust it.
2. `.empty-sub` caps width (`max-width: 380px`) but not height, and `#detail-empty` is a centered column flex container. Long enough input grows the block until `.empty-back` — the card's only recovery affordance, and the one the mobile CSS work in the last round existed to make reachable — is pushed out of view.

One line fixes both, e.g. `id.slice(0, 64)` with an ellipsis, or only echo when `/^[0-9a-f-]{36}$/i.test(id)` and drop the parenthetical otherwise.

### Code Quality

🟢 **`navigate()`'s deep-link invalidation is placed after the no-op early return** — `app.js:5444-5452`

```js
const next = routeToHash(route);
if (next === (location.hash || '#/')) return;   // ← returns first
pendingRoute = null;                            // ← intent: "a real navigation supersedes a deferred deep link"
```

So the guard is skipped in exactly one case: the click targets the session the pending route already names. Cold-load `#/sessions/A`, sidebar paints from the CROW-613 cache, user clicks A's row — `selectedId` is still `null`, so `routedTerminal` is `null`, so the computed hash equals the current one and `pendingRoute` survives. When `list-sessions` lands it re-applies, running `selectSession(A, { fromRoute: true })` a second time and nulling `activeTerminal`.

Harmless in practice — it lands on the same session and `attachWindow` no-ops on an unchanged window — so this is not the socket teardown the `selfWroteHash` comment guards against. But hoisting `pendingRoute = null` above the early return costs nothing and makes the invalidation match its stated intent unconditionally.

🟢 **`openSettings`' failure path leaves the URL claiming a modal that isn't open** — `settings.js:139-149`

On a cold `#/settings/jobs` where `get-config` rejects (daemon restarting, rpc timeout), the function `alertModal`s and returns *after* capturing `routeBeforeSettings` but *before* `navigate`. The hash is already `#/settings/jobs` (the deep link or Back put it there), so the user sits on the previous view under a Settings URL. Secondary effect: `before.view === 'settings'` at capture time, so `routeBeforeSettings` falls back to `{ view: 'home' }` — a later successful open→close then dumps the user home instead of back where they were. Reload recovers. Error-path polish only.

🟢 **Settings open/close pushes twice per visit; terminal tabs push per click**

`openSettings` pushes and `closeSettings` pushes, so five open/close cycles cost ten Back presses — the "undo stack for UI noise" ADR 0018 rejects, and the reason the *tab* switches inside the modal correctly use `replace` (`settings.js:264`). `switchTerminal` (`app.js:2929`) likewise pushes on every tab click, so toggling between two tabs five times costs five presses to leave the session. Both are defensible — a modal and a terminal tab are more "content" than a settings sub-tab — and the ADR is explicit that the terminal segment is deliberate. Flagging only so the asymmetry with the `replace` used two files over is a decision rather than an accident.

🟢 **`test:ci` drops `row.test.js` wholesale, so its 43 green assertions stop gating too**

I re-confirmed the baseline locally: `43 passed, 10 failed`, and this PR touches no row-rendering code, so the failures are pre-existing CROW-802 drift exactly as described. The exclusion is documented in `ci.yml`, `README.md`, and the PR body — that's the right handling, and it is strictly better than `main`, where the entire suite gated nothing. Worth noting only so the eventual follow-up restores the 43 rather than just repairing the 10.

### Tests

- New suite: **112 passed, 0 failed**.
- Full `npm run test:ci` set: green.
- The two drift gates are the right ones and are real assertions, not string greps — `ROUTE_SETTINGS_TABS` is parsed out of `settings.js`'s actual `TABS` and `ROUTE_BOARDS` out of `selectedBoard`'s declared union, so the two lists `app.js` is forced to duplicate (no modules, two classic scripts) can't silently diverge.
- `loadWithSettings` running the *real* `settings.js` alongside `app.js` and counting `get-config` calls is what makes the tab-routing assertion meaningful; a stub could not have caught the `dirty`-reset regression it guards.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 4 Green] findings. The Yellow is a one-line bound on the echoed session id; the four Greens are optional. Everything else here is solid work, and the ADR is unusually good at recording *why* rather than just *what*.

---

dgershman:
All five addressed in 0c28539f.

### 🟡 Unbounded id echoed into the not-found card

Agreed, and the second consequence is the one that convinced me: `.empty-sub` caps width but not height, so a long enough payload pushes `.empty-back` off screen — the affordance the *previous* round's mobile CSS work existed to make reachable. Breaking that with attacker-controlled length would have undone that fix from a different direction.

Gated on shape rather than length: session ids are UUIDs (the ticket says so, and `crow --session` enforces it), so `/^[0-9a-f]{8}-…-[0-9a-f]{12}$/i` costs nothing legitimate. A non-matching id isn't truncated, it simply isn't repeated — a 64-char slice of hostile prose is still hostile prose, just shorter. The value stays in the address bar for diagnosis.

Test drives your payload and asserts both halves — the sentence is dropped, and separately that no markup survives (checking for `onerror` rather than `<img`, since `#detail-empty` legitimately contains the brand image; my first assertion was wrong about that and failed for the wrong reason).

### 🟢 `pendingRoute = null` below the early return

Right — the one case it missed is clicking the very row the pending route names, where the computed hash matches and `navigate()` returns first. Hoisted above the return. Asserted directly: no-op write, `pendingRoute` cleared anyway, `list-sessions` doesn't re-apply.

### 🟢 `openSettings` failure leaves the URL claiming a modal

Fixed — rewinds to `routeBeforeSettings` with `replace`, since an open that never happened shouldn't be a history entry. That also closes the secondary effect you noted, where `routeBeforeSettings` had already degraded to `{ view: 'home' }`. Tested with `get-config` rejecting.

### 🟢 Push/replace asymmetry

Made it a decision instead of leaving it to look accidental — ADR 0018 now states the split and the test behind it: views the user chose (Settings open/close, terminal tabs, session/board selection) push; sub-tab switches and every correction the app makes on the user's behalf replace. The question is whether Back returning there would be useful.

I kept both current behaviours. A modal and a terminal tab are content the user picked and should be able to Back out of; a sub-tab is UI noise. Happy to flip either if you read it differently — it's now written down either way.

### 🟢 `test:ci` drops row's 43 green assertions too

Good catch, and worth being explicit about since the exclusion otherwise looks like it only costs the 10 failures. Noted in `web-tests/README.md`: whoever repairs the 10 should re-add `row` to `test:ci` in the same change, because the win is restoring 43, not fixing 10.

---

`npm run test:ci` → exit 0, router suite **125/125**, deterministic. Still no Swift changes.

dhilgaertner (approved):
## Code & Security Review

Fifth pass. The round-4 Yellow and all four Greens are closed, and I re-derived the claims rather than trusting the commit message.

Verified:

- `npm ci && npm run test:ci` → **426 assertions across 10 files, 0 failed**. `router.test.js` alone: **125 passed, 0 failed** (112 last round).
- `row.test.js` → 43 passed, 10 failed, and `git diff origin/main...HEAD` touches neither `row.js` nor `row.test.js`. The `test:ci` exclusion still documents a pre-existing state rather than hiding anything this PR caused.
- **Lockfile audit** (it now gates every PR, so it's worth doing once): all 63 `resolved` entries point at `registry.npmjs.org`, all 63 carry an `integrity` hash, `lockfileVersion: 3`, and `npm ci` reproduces cleanly. No alternate registries, no git or tarball URLs.
- **ADR 0018's Swift claims, against the source it cites.** `StaticAssets.swift` really is a literal-path table with no catch-all (`/`, `/index.html`, `/login`, `/app.css`, `/app.js`, `/settings.css`, `/settings.js`, `/brand.svg`, `/version.json`, `/auth/check`, `/terminal.html`, `/xterm/:file`); `WebAuthRoutes.serveLoginPageForUnauthorized` is exactly `method == .get && (accept ?? "").contains("text/html")`; and `crow-desktop/src-tauri/src/lib.rs:195` really does `win.eval("window.location.reload()")`. The "deep link works only when you're logged out" inversion is real, so hash routing is the right call for a web-assets-only change.
- No Swift changes, as claimed.

### Critical Issues

None.

### Security Review

**Strengths:**

- **The round-4 Yellow is properly closed.** `SESSION_ID_SHAPE` (`app.js:2288`) gates the echo on the UUID shape a session id actually has, so a crafted `#/sessions/<prose>` renders the generic explanation and nothing else. The test drives a real hostile payload and asserts both that the prose is dropped *and* that no markup survives — bounding the length would have fixed the layout half while leaving 64 chars of attacker text under Crow's branding, so gating on shape is the stronger of the two fixes that were on the table. `.empty-back` stays reachable.
- **No HTML injection anywhere on the route path.** The id reaches the DOM only through `textContent` (`app.js:2251`, `2255`), never `innerHTML`.
- **No open redirect in the login hop.** `login.html` builds `'/' + (location.hash || '')` and `app.js` builds `'/login' + (location.hash || '')`. `location.hash` is `#`-prefixed or empty per spec, so neither can go protocol-relative. `location.replace` also keeps `/login` out of history, and the `location.pathname === '/'` branch is correct — a fragment deep link keeps `pathname === '/'`, where assignment wouldn't reload but `reload()` preserves the fragment.
- **Malformed fragments degrade rather than throw.** `parseRoute` wraps `decodeURIComponent` in try/catch and returns `null`; `applyRoute` normalizes that to home with `replace`. Board and settings segments are checked against `ROUTE_BOARDS` / `ROUTE_SETTINGS_TABS` allowlists, with an unknown tab degrading to `general` *and* the bogus hash rewritten in place so the address bar can't sit on a shape the app silently reinterprets.
- **Ids are validated before they reach an RPC.** `applyRoute` checks session membership against `sessions` before `selectSession`; `pendingTerminalId` is only ever matched against the `list-terminals` response.
- **Supply chain pinned in both directions** — `npm ci` against the now-committed lockfile, `actions/setup-node@v4` at an explicit `node-version: '22'`. Since this step gates every PR, a floating dep or a floating runtime were the two failure modes; both are closed.
- Zero Swift changes means the auth gate, the `index.html`-keyed CSP, and `isSafeAssetName` are untouched.

**Concerns:** none found.

### Code Quality

Four Greens, all "consider". None block.

**🟢 The boot branch that actually ships is the one the tests don't take** — `app.js:5586-5590`

`index.html` loads `/app.js` then `/settings.js`, both at the end of `<body>` with no `defer`, so `document.readyState === 'loading'` when app.js evaluates and the live path is always `document.addEventListener('DOMContentLoaded', applyBootRoute, { once: true })`. The jsdom loader evaluates app.js against a fully parsed document, so every cold-load assertion in `router.test.js` goes through the `else { applyBootRoute(); }` branch instead. The comment at `app.js:5584` says this outright, which I appreciate — but it means the deferral that exists specifically to fix the `#/settings/*` cold-link bug is the untested half. I traced it and it's correct either way (`applyRoute` re-checks `sessionsLoaded` itself, so both orderings against the boot `refreshSessions()` resolve), so this is coverage, not a defect. Dispatching a synthetic `DOMContentLoaded` in one test would close it.

**🟢 `navigate()`'s `replace` path leaves `selfWroteHash` stale** — `app.js:5467`

The push path sets `selfWroteHash = next`; the `replace` path returns early without clearing it. If a `replace` lands between a push and that push's asynchronous `hashchange`, the event fires with `location.hash` already at the replaced value, the `selfWroteHash === cur` guard misses, and `onHashChange` re-applies the route it just applied — the `selectSession → refreshTerminals → attachWindow` teardown the guard's own comment exists to prevent. I walked every ordered pair of `navigate` calls in the file and could not reach it today: the one push-then-replace sequence (`selectSession`'s push, then `refreshTerminals`' dead-terminal correction) has `await rpc('list-terminals')` between them, and the queued `hashchange` task always runs before a WebSocket response. So it's latent, not live — but it's latent on an ordering assumption nothing states, and `selfWroteHash = null` on the replace path removes the dependency for one line.

**🟢 The mid-session reap guard trusts a single `list-sessions` response** — `app.js:1068`

`if (selectedId && !sessions.some(...)) showSessionNotFound(...)` is the right handling for the reaper, and I checked the obvious false positive (`list-sessions` applies no status filter, so an archived session can't trip it). The remaining exposure is a transient *empty* response — the client polls a reconnecting daemon after the `run-setup` re-exec, for instance. Because `showEmptyDetail` nulls `selectedId`, a single bad response is unrecoverable by polling: the user is parked on "Session not found" until they re-click. I can't demonstrate it — `list-sessions` maps `appState.sessions`, which is populated at daemon startup — so this is speculative. Requiring two consecutive misses, or skipping the guard when `sessions.length === 0`, would make it robust for about as much code as the comment above it.

**🟢 Nothing detaches the terminal socket on the not-found path** — `app.js:2237-2238`

`showEmptyDetail` clears `terminals` and `activeTerminal`, and the CSS hides `#terminal-wrap`, but no `detachWindow` runs. Not a regression — the `deleteSession` cleanup this replaces didn't detach either — except that the not-found path is now reachable on a *live* session via the reap guard, which the old code path never was. Worth a glance at whether `attachWindow`'s subscription to a now-dead tmux window matters; I couldn't confirm without a running daemon.

### Tests

The suite is the strongest part of this PR. The two drift gates are real parsing, not string greps — `ROUTE_SETTINGS_TABS` is pulled out of `settings.js`'s actual `TABS` array and `ROUTE_BOARDS` out of `selectedBoard`'s declared union — so the two lists `app.js` is forced to duplicate (no modules, two classic scripts) can't silently diverge. `loadWithSettings` running the *real* `settings.js` alongside `app.js` and counting `get-config` calls is what makes the tab-routing assertion mean anything; a stub could not have caught the `dirty`-reset regression it guards. The hand-rolled history model is the right call given jsdom doesn't model Back against a vm context, and it's what caught the round-3 Back-wall.

One coupling worth knowing about: the `selectedBoard` drift gate parses a source *comment*. It fails closed, which is the right direction, but reformatting that one line breaks CI with a message about the gate rather than about the change.

### Notes on this review

The review worktree was deleted by the retention reaper partway through — fittingly, the exact scenario this PR now handles. All verification above completed before that; only the `closeSettings` caller check was re-fetched from GitHub afterwards (five call sites, all fire-and-forget, so the new `false`/`true` return is purely additive for the router). As with previous rounds and as the PR body states, there was no browser click-through: everything here is source reading plus the jsdom suite against the real files.

### Summary Table

Five rounds is a lot, but each one moved something real, and the residue is four "consider"s I couldn't turn into a demonstrable failure. The architecture is right — routing goes *through* the selection functions rather than beside them, the applying-vs-initiating distinction is now explicit in both the code and the ADR, and the push/replace split is recorded as a decision with a stated test. ADR 0018 is unusually good at recording *why*, and it holds up line-by-line against the Swift it reasons about. Getting this suite into CI at all is a bigger win than the feature.

---

PR #940, opened 2026-08-05
Fix release.yml to ship universal crow + crowd binaries

### LINKED ISSUE #939: release.yml is broken: calls bundle.sh and sign-and-notarize.sh, which no longer exist
## Problem

`.github/workflows/release.yml` cannot succeed. Its `release` job runs:

```yaml
- name: Bundle Crow.app
  run: |
    bash scripts/generate-build-info.sh
    bash scripts/bundle.sh

- name: Sign, create DMG, and notarize
  run: bash scripts/sign-and-notarize.sh
```

**Neither `scripts/bundle.sh` nor `scripts/sign-and-notarize.sh` exists.** The current contents of `scripts/` are:

```
check-cli-parity.sh   check-no-nslog.sh   check-notification-events.sh
check-workspace-custom-instructions.sh   coverage-summary.sh
coverage-summary.swift   daemon-run.sh   generate-build-info.sh
```

Both were deleted when the macOS app was retired ([ADR 0010](docs/adr/0010-retire-the-macos-app.md)) — that ADR explicitly removed "the `.app` packaging/signing scripts" — but `release.yml` was never updated to match. The workflow still bundles, signs, and notarizes an app that no longer exists, and still publishes `Crow-${VERSION}.dmg` as a release asset.

## Impact

Any `v*` tag push fails at "Bundle Crow.app". Consequences visible today:

- **Zero git tags and zero GitHub Releases** on `corveil/crow` (`gh api repos/corveil/crow/releases` → `0`). Same on `radiusmethod/crow`.
- `VERSION` frozen at `0.1.0` while ~310 commits landed on `main`.
- `CHANGELOG.md` has every one of those commits under `[Unreleased]`, described as a "backfill of merged PRs since the 0.1.0 release".
- The certificate-import, notarization, and keychain-cleanup steps carry real Apple secrets (`DEVELOPER_CERTIFICATE_BASE64`, `APPLE_APP_SPECIFIC_PASSWORD`, …) for a signing path that no longer has anything to sign.

Because it's tag-triggered, this is invisible on PRs and on `main` — nothing runs it, so nothing reports it broken.

## What a release should produce now

Post-ADR-0010 the build produces **two binaries, no app bundle**: `crow` (CLI) and `crowd` (daemon). The workflow already builds the CLI correctly:

```yaml
swift build -c release --arch arm64 --arch x86_64 --product crow
```

…and [ADR 0006](docs/adr/0006-universal-macos-binary.md) commits to shipping universal `arm64` + `x86_64` artifacts, which that command satisfies.

## Suggested fix

1. Delete the "Bundle Crow.app" and "Sign, create DMG, and notarize" steps, plus the certificate-import and keychain-cleanup steps that exist only to serve them.
2. Build **both** products universal (`crow` and `crowd`), not just `crow`.
3. Replace the `.dmg` release asset with a tarball of the two binaries; keep `generate_release_notes: true`.
4. Decide and document whether the shipped binaries are signed/notarized at all. If yes, sign the two binaries directly (`codesign` + `notarytool` on the archive) and keep the Apple secrets; if no, remove the secrets from the workflow rather than leaving them wired to dead steps. **Unsigned binaries downloaded from a browser hit Gatekeeper quarantine**, so this needs an explicit answer and an install note, not a default.
5. Verify end-to-end by pushing a prerelease tag (`v0.2.0-rc.1` — the workflow already sets `prerelease: ${{ contains(github.ref_name, '-') }}`) before cutting a real one.
6. Roll `CHANGELOG.md`'s `[Unreleased]` into a real version heading as part of the first successful release.

## Acceptance criteria

- [ ] `release.yml` references no nonexistent script.
- [ ] A prerelease tag push completes green and publishes downloadable assets.
- [ ] Published assets contain universal (`arm64` + `x86_64`) `crow` **and** `crowd` — verified with `lipo -info`.
- [ ] The signing/notarization decision is explicit: either it happens and the secrets are used, or the steps and secrets are gone.
- [ ] Downloading and running the released `crow` on a clean machine works, with any Gatekeeper step documented in the release notes or README.
- [ ] `VERSION` and `CHANGELOG.md` reflect the released version.

## Related

Blocks the "prefer real releases" option in the version-check work (#938), which currently has to compare against the branch head because no release has ever existed.

dgershman:
Closes #939

## Summary

- Remove the dead `.app` bundle, DMG, and notarization steps from `release.yml` (scripts deleted in ADR 0010)
- Build and publish universal (`arm64` + `x86_64`) `crow` and `crowd` binaries in a tarball via new `scripts/package-release.sh`
- Ship **unsigned** release artifacts — signing secrets and keychain steps are removed; install/Gatekeeper notes are in the release body and `docs/troubleshooting.md`
- Bump `VERSION` to `0.2.0` and roll `[Unreleased]` changelog entries into `[0.2.0]`

## Test plan

- [ ] Merge and push a prerelease tag (`v0.2.0-rc.1`) to verify the workflow completes green
- [ ] Confirm the published tarball contains universal `crow` and `crowd` (`lipo -info` on both)
- [ ] Download and run the released `crow` on a clean machine (clear quarantine with `xattr -d` if needed)


dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues

**🔴 `scripts/package-release.sh:20` — the build produces only `crowd`; `crow` is never built, and the release job dies on the first tag push.**

```bash
swift build -c release --arch arm64 --arch x86_64 --product crow --product crowd
```

SwiftPM's `--product` is a **single-value** option (`var product: String?`), unlike `--arch`, which is an array. Repeating it doesn't accumulate — the last value wins and the earlier one is discarded. Verified two ways in this worktree:

```
$ swift build --product crow --product bogusproduct
error: no product named 'bogusproduct'          # 'crow' was dropped; only the last value was parsed

$ swift build --scratch-path "$TMPDIR/pr940-build" --product crow --product crowd
[1196/1198] Linking crowd
Build of product 'crowd' complete! (34.37s)
$ ls "$TMPDIR/pr940-build/debug/" | grep -w crow
                                                 # crow MISSING, crowd EXISTS
```

CrowCLI is never even compiled. The failure then cascades through the rest of the script:

- `package-release.sh:22` probes `[ -f .build/apple/Products/Release/crow ]` — that is precisely the file that won't exist — so `BUILD_DIR` silently falls back to `.build/release`.
- `package-release.sh:28` `cp "$BUILD_DIR/crow" …` then fails under `set -euo pipefail`, exiting non-zero.

Net effect: pushing `v0.2.0` runs `test` + `shellcheck` green, then the `release` job fails at "Build release binaries" and **no release is published**. The PR's test plan (`Merge and push a prerelease tag (v0.2.0-rc.1)`) would surface this, but only after merge.

The repo already has the correct incantation in two places — `docs/getting-started.md:49-50` uses two separate invocations, and [ADR 0006](docs/adr/0006-universal-macos-binary.md) specifies `swift build --arch arm64 --arch x86_64` with no `--product` at all. Either works:

```bash
# Option A — build all products (crow + crowd are the only two)
swift build -c release --arch arm64 --arch x86_64

# Option B — two invocations; the second is nearly free (shared build cache)
swift build -c release --arch arm64 --arch x86_64 --product crow
swift build -c release --arch arm64 --arch x86_64 --product crowd
```

### Security Review

**Strengths:**

- Removing the signing steps removes the whole keychain-handling surface with them: no more base64 P12 decode to `$RUNNER_TEMP`, no `security create-keychain` with a generated password, no `set-key-partition-list`, and no cleanup step that could leak a keychain on early failure. Strictly less credential material touches the runner.
- Six Apple secrets drop out of the workflow's blast radius (`DEVELOPER_CERTIFICATE_BASE64`, `DEVELOPER_CERTIFICATE_PASSWORD`, `DEVELOPER_ID_APPLICATION`, `APPLE_ID`, `APPLE_APP_SPECIFIC_PASSWORD`, `APPLE_TEAM_ID`).
- No secrets are introduced, echoed, or interpolated into the new script or the release body. `CROW_VERSION` is the only injected value and it's tag-derived.
- The removal is complete and consistent — no dangling references to `Crow-*.dmg`, `crow-cli-macos-universal`, `scripts/bundle.sh`, or `scripts/sign-and-notarize.sh` survive anywhere in the repo.
- Dropping signing causes no *functional* regression: the codebase has no `SecItem`/Keychain usage, so there's no keychain-ACL identity churn from unsigned rebuilds, and `docs/troubleshooting.md` already documented unsigned source builds as the status quo.

**Concerns:**

**🟡 The tarball ships with no integrity metadata, while the release body instructs users to strip Gatekeeper.** Signing and notarization were the authenticity guarantee for downloaded artifacts, and nothing replaces them. `release.yml:98` then tells every user to run:

```bash
xattr -d com.apple.quarantine ~/.local/bin/crow ~/.local/bin/crowd
```

That's the last OS-level check on a binary the user has no way to verify. Source builds being unsigned is fine — the user compiled the code. A *downloaded* binary is a different trust position. Publish checksums so the de-quarantine step is a verified action rather than blind trust:

```bash
# in package-release.sh, alongside the tarball
shasum -a 256 "$ARCHIVE" > "$ARCHIVE.sha256"
```

```yaml
# release.yml
files: |
  crow-${{ env.CROW_VERSION }}-macos-universal.tar.gz
  crow-${{ env.CROW_VERSION }}-macos-universal.tar.gz.sha256
```

…and add a `shasum -a 256 -c` line to the install snippet. Note the CLI binaries could still be Developer ID signed + notarized even without an `.app` (standalone Mach-O signing and zip-based notarization both work) — but if that's deliberately out of scope, checksums are the cheap floor.

**🟢 The release pipeline is now the sole integrity guarantee, and it runs unpinned third-party actions.** `softprops/action-gh-release@v2` and `maxim-lobanov/setup-xcode@v1` are mutable tags, under a workflow-level `permissions: contents: write` that also applies to the `test` and `shellcheck` jobs. This is pre-existing and not introduced here — but signing was the countermeasure that would have made a tampered artifact detectable, and this PR removes it. Worth SHA-pinning now that nothing downstream would catch a compromise.

**🟢 Retire the now-unused Apple secrets.** The six secrets above no longer have a consumer in any workflow. Dormant credentials with no consumer are pure downside — delete them from repo settings (and revoke the Developer ID cert if it isn't used elsewhere).

### Code Quality

**🟡 `scripts/package-release.sh:31` — `lipo -info` prints but never asserts.** The script's entire purpose is producing universal binaries, yet nothing fails if they aren't. If the `.build/release` fallback branch at line 25 is ever taken, the workflow packages arch-thin binaries into a file named `…-macos-universal.tar.gz` and goes green. The PR's own test plan defers this to a manual post-release check (`Confirm the published tarball contains universal crow and crowd (lipo -info on both)`) — make CI do it instead:

```bash
for binary in crow crowd; do
  cp "$BUILD_DIR/$binary" "$STAGING_DIR/$binary"
  archs="$(lipo -archs "$STAGING_DIR/$binary")"
  for arch in arm64 x86_64; do
    case " $archs " in
      *" $arch "*) ;;
      *) echo "ERROR: $binary is missing $arch (got: $archs)" >&2; exit 1 ;;
    esac
  done
done
```

**🟢 `.gitignore` doesn't cover the script's outputs.** Neither `release-staging/` nor `crow-*.tar.gz` is ignored, so running `scripts/package-release.sh` locally leaves an untracked multi-MB tarball in the repo root. `rm -rf "$STAGING_DIR"` only runs on the success path with no `trap`, so a mid-run failure also leaves `release-staging/` behind.

**🟢 `release.yml:92` — `install -m 755 crow crowd ~/.local/bin/` fails when the directory doesn't exist.** `install` won't create the destination. Prepend `mkdir -p ~/.local/bin` to the snippet.

**Verified clean:**

- `shellcheck scripts/*.sh` passes (exit 0) — mirrors the CI `shellcheck` job, including the new script.
- `generate_release_notes: true` alongside a custom `body` is valid: GitHub prepends the supplied body to the generated notes, so the install instructions land above the changelog rather than replacing it.
- `CHANGELOG.md` uses inline links throughout with no reference-link definition block, so the new `## [0.2.0]` heading needs no matching `[0.2.0]: <url>` definition. The `2026-08-05` date is today's.
- `VERSION` → `0.2.0` can't mislabel a release: `scripts/generate-build-info.sh:16` prefers `CROW_VERSION` (tag-derived) over the file, so tag/VERSION drift is harmless.
- `STAGING_DIR` is rooted at `$ROOT_DIR` derived from `BASH_SOURCE`, so the `rm -rf` can never expand to an empty or unexpected path.

### Summary Table

| # | Finding | Location |
|---|---------|----------|
| 🔴 | `--product` is single-valued — only `crowd` is built, `cp` of `crow` then fails the job | `scripts/package-release.sh:20` |
| 🟡 | No checksums for unsigned artifacts, while the body tells users to strip quarantine | `release.yml:88-100`, `package-release.sh` |
| 🟡 | `lipo -info` is informational — a thin build ships as "universal" and CI stays green | `scripts/package-release.sh:31` |
| 🟢 | Unpinned third-party actions are now the only integrity guarantee | `.github/workflows/release.yml` |
| 🟢 | Six Apple secrets are now unused — retire them | repo settings |
| 🟢 | `release-staging/` and `crow-*.tar.gz` not gitignored; no `trap` cleanup | `.gitignore`, `package-release.sh` |
| 🟢 | `install` fails without `mkdir -p ~/.local/bin` | `release.yml:92` |

The direction here is right — the dead `.app`/DMG/notarization steps genuinely needed to go after ADR 0010, the tarball shape is a clear improvement over two loose assets, and the docs/changelog/version work is thorough. The blocker is a one-line SwiftPM argument-arity bug that happens to be fatal to the exact thing the PR sets out to fix.

**Recommendation:** Request Changes — driven by [1 Red, 2 Yellow, 4 Green] findings.

---

dhilgaertner (changes requested):
## Code & Security Review

Removing the dead `.app`/DMG/notarization path is clearly right, and dropping the certificate-import + keychain steps is a genuine security win (four fewer secrets on the runner, no `.p12` on disk, no `set-key-partition-list`). Two defects in the new packaging path block the release, though — both fail the very first tagged run or ship a broken `crowd`.

### Critical Issues

#### 🔴 1. `swift build --product crow --product crowd` builds **only `crowd`** — the release job dies at the `cp`

`scripts/package-release.sh:20`

SwiftPM's `--product` is a **singular** option (`swift build --help`: `--product <product>   Build the specified product.` — unlike `--arch`, which documents "use the option multiple times"). Repeating it doesn't accumulate; the **last** value wins and the earlier one is silently ignored. Verified empirically in this worktree (Swift 6.3.2):

```
$ swift build --product crow --product nonexistentXYZ
error: no product named 'nonexistentXYZ'        # last value is the one used

$ swift build --product nonexistentXYZ --product crow
[183/184] Applying crow
Build of product 'crow' complete! (13.12s)      # first value silently dropped
```

So line 20 builds `crowd` only. On a clean CI runner `.build/apple/Products/Release/crow` then does not exist, the probe at `:22` falls through to `BUILD_DIR=".build/release"` (which a `--arch`-universal build never populates), and `cp "$BUILD_DIR/crow"` fails under `set -e`. **Every `v*` tag fails the release job.** The old workflow was safe here because it passed a single `--product crow`.

Fix — two invocations (or drop `--product` entirely and build all products):

```bash
swift build -c release --arch arm64 --arch x86_64 --product crow
swift build -c release --arch arm64 --arch x86_64 --product crowd
```

#### 🔴 2. The tarball ships a `crowd` that will `fatalError` — its SwiftPM resource bundles are left behind

`scripts/package-release.sh:32-38`

`crowd` reaches its resources through `Bundle.module`, from two targets that declare `resources:`:

- `Packages/CrowDaemon/Package.swift:64` → `.copy("Resources/web")`, read by `CrowDaemon/StaticAssets.swift:75` (`index.html`, `login.html`, `version.json` — the entire web UI, which is the only client per ADR 0010)
- `Packages/CrowTerminal/Package.swift:23` → `crow-shell-wrapper.sh`, `crow-tmux.conf`, `xterm/`, read by `CrowTerminal/BundledResources.swift:14,19,27` (every terminal + the xterm assets `crowd` serves over HTTP)

SwiftPM emits those as sibling directories of the executable — confirmed by building `crowd` here:

```
.build/arm64-apple-macosx/debug/CrowDaemon_CrowDaemon.bundle
.build/arm64-apple-macosx/debug/CrowTerminal_CrowTerminal.bundle
```

and the generated accessor (`.build/.../CrowDaemon.build/DerivedSources/resource_bundle_accessor.swift`) is fatal when neither location resolves:

```swift
let mainPath = Bundle.main.bundleURL.appendingPathComponent("CrowDaemon_CrowDaemon.bundle").path
let buildPath = "/Users/.../.build/arm64-apple-macosx/debug/CrowDaemon_CrowDaemon.bundle"
guard let bundle = Bundle(path: mainPath) ?? Bundle(path: buildPath) else {
    Swift.fatalError("could not load resource bundle: from \(mainPath) or \(buildPath)")
}
```

The loop at `:32-35` copies **only the two Mach-O files**, so a user who follows the release body (`install -m 755 crow crowd ~/.local/bin/`) gets a `crowd` with no sibling bundle and a baked-in `buildPath` pointing at a GitHub runner directory that doesn't exist on their machine. The first browser hit (`StaticAssets`) or first terminal spawn (`BundledResources`) crashes the daemon. This is exactly what `make install`'s **symlinks** hide — the symlink target sits next to its bundles, so nobody has hit it locally.

`crow` is unaffected: `Packages/CrowCLI/Package.swift` pulls in `CrowIPC`/`CrowCore`/`CrowCodex`/`CrowAutostart`/`ArgumentParser`, none of which declare resources. This is a `crowd`-only break.

Fix — stage the bundles alongside the binaries and tar the directory, e.g.:

```bash
cp -R "$BUILD_DIR"/*.bundle "$STAGING_DIR"/
```

then adjust the install instructions (a bare-binary `PATH` drop no longer works — ship a directory, or add a wrapper/install step that keeps the bundles adjacent). Please also add a smoke step that runs the **staged** `crowd` (e.g. start it, `curl` `/` and `/version.json`, stop it) from a directory other than the build tree — that would have caught this and #1 in CI.

### Security Review

**Strengths:**

- Removes `DEVELOPER_CERTIFICATE_BASE64`, `DEVELOPER_CERTIFICATE_PASSWORD`, `DEVELOPER_ID_APPLICATION`, `APPLE_ID`, `APPLE_APP_SPECIFIC_PASSWORD`, `APPLE_TEAM_ID` from the workflow — no more base64 secret written to `$RUNNER_TEMP/certificate.p12`, no ad-hoc keychain, no `security set-key-partition-list`. Net reduction in secret handling, and the `if: always()` keychain cleanup it obviated can't leak on a mid-job failure any more.
- `permissions: contents: write` is unchanged and still the minimum for `action-gh-release`.
- `${{ github.repository }}` in the release body is a trusted context, not attacker-controlled input; no `${{ }}` interpolation lands inside a `run:` shell block.
- `scripts/package-release.sh` handles no credentials, and `rm -rf "$STAGING_DIR"` is anchored to `$ROOT_DIR/release-staging` (derived from `BASH_SOURCE`), never an unbounded path.
- Shipping unsigned is documented rather than silent, in both the release body and `docs/troubleshooting.md`. `swift build` still ad-hoc-signs the merged universal binary, so the artifact remains executable on Apple Silicon.

**Concerns:**

- None beyond the above. Worth noting for the record that unsigned + un-notarized artifacts mean users have no provenance check on the download; if that's a long-term concern, GitHub artifact attestations (`actions/attest-build-provenance`) are a cheap, secret-free substitute for Developer ID. Not blocking.

### Code Quality

#### 🟡 3. `lipo -info` can't fail — a thin binary would ship labelled `macos-universal`

`scripts/package-release.sh:22-26,34`

`lipo -info` only prints; it exits 0 for a thin binary. Combined with the silent `BUILD_DIR` fallback at `:22-26` (which, once #1 is fixed, is the only thing standing between a failed universal build and a `.build/release` thin one), nothing gates the central promise of the PR title. Assert instead:

```bash
archs="$(lipo -archs "$STAGING_DIR/$binary")"
for want in arm64 x86_64; do
  case " $archs " in *" $want "*) ;; *) echo "ERROR: $binary missing $want (got: $archs)" >&2; exit 1 ;; esac
done
```

Same argument for the `BUILD_DIR` probe: since the script always passes `--arch`, `.build/apple/Products/Release` is the only correct output path — failing loudly when it's absent beats silently packaging a thin fallback.

#### 🟡 4. The release body's install snippet fails on a machine without `~/.local/bin`

`.github/workflows/release.yml:94,100`

This is the primary install path for people who explicitly are *not* building from source, so `~/.local/bin` frequently won't exist — `install -m 755 crow crowd ~/.local/bin/` then errors with `No such file or directory`. Add `mkdir -p ~/.local/bin` ahead of it. Likewise `xattr -d com.apple.quarantine …` (`:100`, and `docs/troubleshooting.md:82`) exits non-zero with `No such xattr` when the attribute isn't set — `xattr -c` or `xattr -d … 2>/dev/null || true` reads better in copy-pasted instructions. Both snippets will also need rewording once #2 is fixed and the tarball contains more than two files.

#### 🟡 5. Nothing checks the tag against `VERSION`

`.github/workflows/release.yml:74`, `VERSION:1`

`CROW_VERSION` comes from the tag and `generate-build-info.sh` prefers it over the `VERSION` file, so the two can silently diverge — pushing `v0.3.0` against a `VERSION` of `0.2.0` produces a green release with mismatched provenance. The bump to `0.2.0` is correct here; a one-line guard in the "Extract version from tag" step would keep it that way:

```bash
[ "$CROW_VERSION" = "$(tr -d '[:space:]' < VERSION)" ] || { echo "tag/VERSION mismatch"; exit 1; }
```

#### 🟢 6. Nits

- `release-staging/` and `crow-*.tar.gz` aren't in `.gitignore`, so a local run of the script leaves committable artifacts in the repo root (and the script leaves `release-staging/` behind if it fails between `:30` and `:39`). The `.gitignore` still carries `*.dmg` / `*.p12` for the paths this PR deletes — fine to leave, but they're now dead entries.
- `tar -czf` on macOS can emit `._*` AppleDouble members for files carrying extended attributes; `COPYFILE_DISABLE=1 tar …` makes the archive deterministic and avoids the stray entries.
- `scripts/package-release.sh` passes `shellcheck` clean (verified locally), so the `shellcheck` gate stays green.
- CHANGELOG restructuring is correct: `[Unreleased]` is left empty and the backfill paragraph moves under `[0.2.0]`, matching Keep-a-Changelog. No link-reference block exists in this file, so none is missing.

### Summary Table

**Recommendation:** Request Changes — driven by [2 Red, 3 Yellow, 1 Green] findings.

Both Reds are pre-merge blockers rather than "fix in the next release": #1 means the workflow can never produce an artifact, and #2 means that once #1 is fixed the artifact it does produce is a `crowd` that crashes on first use. The test plan's second and third boxes (`lipo -info` on both binaries; run on a clean machine) are precisely the checks that would surface them — please run the `v0.2.0-rc.1` prerelease before merging rather than after.

dhilgaertner (changes requested):
## Code & Security Review

The three fixes in `912bd02` land, and I verified each: the `--product` arity bug is gone (both binaries build fat), `lipo` now asserts instead of printing, and checksums, `.gitignore`, `trap` cleanup, and `mkdir -p ~/.local/bin` are all in. One blocker from the previous round is still open, though — and it's the one that makes the published `crowd` unusable.

### Critical Issues

#### 🔴 1. The tarball ships `crowd` stripped of its SwiftPM resource bundles — it `fatalError`s on the first web request or terminal spawn

`scripts/package-release.sh:39-52`

`crowd` reaches its resources through `Bundle.module`, from two targets that declare `resources:`:

| Target | Declaration | Read by |
|---|---|---|
| `CrowDaemon` | `Packages/CrowDaemon/Package.swift:65` — `.copy("Resources/web")` | `StaticAssets.swift:75` — every `/`, `/app.js`, `/login`, `/version.json` |
| `CrowTerminal` | `Packages/CrowTerminal/Package.swift:24-26` — wrapper script, tmux conf, `xterm/` | `BundledResources.swift:14,19,27` — every terminal spawn, and the `/xterm/*` route |

SwiftPM emits those as **sibling directories of the executable**. Confirmed by running the exact build this script runs (`swift build -c release --arch arm64 --arch x86_64`) in this worktree:

```
$ ls .build/apple/Products/Release/ | grep bundle
CrowDaemon_CrowDaemon.bundle
CrowTerminal_CrowTerminal.bundle
swift-crypto_Crypto.bundle
swift-nio__NIOFileSystem.bundle
swift-nio_NIOPosix.bundle
```

The loop at `:39-49` copies only the two Mach-O files. Running the script itself:

```
$ CROW_VERSION=0.2.0-reviewtest bash scripts/package-release.sh
crow: x86_64 arm64
crowd: x86_64 arm64
Created crow-0.2.0-reviewtest-macos-universal.tar.gz and ….sha256

$ tar -tzf crow-0.2.0-reviewtest-macos-universal.tar.gz
crow
crowd
```

The generated accessor is fatal when neither location resolves (`.build/…/CrowDaemon.build/DerivedSources/resource_bundle_accessor.swift`):

```swift
let mainPath = Bundle.main.bundleURL.appendingPathComponent("CrowDaemon_CrowDaemon.bundle").path
let buildPath = "/Users/…/.build/…/CrowDaemon_CrowDaemon.bundle"
guard let bundle = Bundle(path: mainPath) ?? Bundle(path: buildPath) else {
    Swift.fatalError("could not load resource bundle: from \(mainPath) or \(buildPath)")
}
```

A user following the release body puts `crowd` at `~/.local/bin/crowd`, so `mainPath` is `~/.local/bin/CrowDaemon_CrowDaemon.bundle` (absent) and `buildPath` is a GitHub runner directory (absent). `Bundle.module` is a lazy `static let`, so the daemon **starts fine and dies on first use** — the first browser hit through `StaticAssets.webResponse`, or the first terminal through `BundledResources`. Per [ADR 0010](docs/adr/0010-retire-the-macos-app.md) the web UI is the *only* client, so that is the entire product.

`make install` hides this because it **symlinks** to the binary sitting next to its bundles — which is why nobody has hit it locally.

`crow` is unaffected: `Packages/CrowCLI/Package.swift:11-15` pulls `CrowIPC`/`CrowCore`/`CrowCodex`/`CrowAutostart`/`ArgumentParser`, and the only targets in the repo declaring `resources:` are `CrowDaemon`, `CrowTerminal`, and `CrowTelemetry` (test fixtures). This is a `crowd`-only break.

Fix — stage the bundles alongside the binaries:

```bash
cp -R "$BUILD_DIR"/*.bundle "$STAGING_DIR"/
tar -czf "$ARCHIVE" -C "$STAGING_DIR" .
```

…and rework the install snippet, because `install -m 755 crow crowd ~/.local/bin/` no longer suffices — the bundles have to stay adjacent to `crowd`. The usual shape is a versioned directory (`crow-0.2.0/` holding binaries + bundles) that the user copies whole, then symlinks or adds to `PATH`.

### 🟡 Should Fix

#### 🟡 2. Nothing in CI ever executes the artifact it publishes

`.github/workflows/release.yml:76-79`

The release job builds, asserts arch, tars, and publishes. No step runs the staged binaries from a directory other than the build tree — which is exactly the check that separates "the file exists and is fat" from "the binary works once installed." Both this round's Red and the previous round's `--product` Red are invisible to every gate in the workflow, and the test plan defers all three verifications to manual steps *after* a tag push.

A few lines against `$STAGING_DIR` would close the gap permanently:

```bash
( cd "$STAGING_DIR" && ./crow --version )
# and, once #1 is fixed: start ./crowd on a scratch --socket-path/--port,
# curl / and /version.json, then stop it
```

The `crowd` half is the one that matters — it is the only check that exercises `Bundle.module` outside the build tree.

### Security Review

**Strengths:**

- Removing the signing steps removes the whole credential surface with them: no base64 `.p12` written to `$RUNNER_TEMP`, no ad-hoc keychain with a generated password, no `security set-key-partition-list`, and no `if: always()` cleanup that could leave a keychain behind on early failure. Verified no workflow still references `DEVELOPER_CERTIFICATE_BASE64`, `DEVELOPER_CERTIFICATE_PASSWORD`, `DEVELOPER_ID_APPLICATION`, `APPLE_ID`, `APPLE_APP_SPECIFIC_PASSWORD`, or `APPLE_TEAM_ID`.
- Checksums are now published and the `shasum -a 256 -c` verification leads the install snippet — last round's Yellow is properly closed.
- `permissions: contents: write` is unchanged and still the minimum for `action-gh-release`.
- `${{ github.repository }}` is trusted context, and no `${{ }}` interpolation lands inside a `run:` shell block.
- `rm -rf "$STAGING_DIR"` is anchored to `$ROOT_DIR/release-staging`, derived from `BASH_SOURCE` — it can never expand to an empty or unbounded path. The added `trap cleanup EXIT` also fixes the leftover-on-failure case.
- The removal is complete and consistent: no `Crow-*.dmg`, `crow-cli-macos-universal`, `scripts/bundle.sh`, or `scripts/sign-and-notarize.sh` references survive anywhere in the repo.
- Shipping unsigned is documented rather than silent, in both the release body and `docs/troubleshooting.md`.

**Concerns:**

- **🟢 The checksum is same-origin.** The `.sha256` is published on the same release, by the same job, as the tarball it attests. It catches a corrupted or truncated download — not a compromised pipeline or anyone with repo write access. That is the right cheap floor, but `actions/attest-build-provenance` is secret-free and would restore the verifiable provenance that Developer ID signing used to carry. Worth adding now that nothing downstream would notice a tampered artifact.
- **🟢 Unpinned third-party actions.** `softprops/action-gh-release@v2`, `maxim-lobanov/setup-xcode@v1`, `actions/checkout@v4`, and `actions/upload-artifact@v4` are all mutable tags, under a workflow-level `permissions: contents: write` that also covers `test` and `shellcheck` (neither needs write). Pre-existing and not introduced here — but signing was the countermeasure that would have made a tampered artifact detectable. SHA-pinning, and scoping the write permission to the `release` job, are both cheap now.
- **🟢 Retire the six orphaned Apple secrets.** They no longer have a consumer in any workflow. Dormant credentials with no consumer are pure downside — delete them from repo settings and revoke the Developer ID cert if it isn't used elsewhere.

### Code Quality

- **🟢 `make build CONFIG=release` in the release body needs Rust.** `build: daemon app` (`Makefile:20`) and the `app` target hard-errors without cargo (`Makefile:31-33`). Because `daemon` runs first the Swift binaries *do* get built, but the `&&` short-circuits and `make install` never runs — the user sees a cargo error and has to work out that the binaries are actually there. `make daemon CONFIG=release && make install CONFIG=release` is the Swift-only path. `README.md:130` and `docs/getting-started.md:172` carry the same phrasing, so this is a pre-existing docs wart now surfacing on the "I don't want unsigned binaries" path — worth fixing in all three.
- **🟢 `xattr -d com.apple.quarantine` in the release body (`:104`) and `docs/troubleshooting.md:82`.** Both are correctly hedged with "may need to", so this is minor — but worth knowing that `tar` (unlike Archive Utility) does **not** propagate `com.apple.quarantine` to extracted members. The common case is that the attribute sits on the `.tar.gz` and not on the installed binaries, where the command exits non-zero with `No such xattr`. `xattr -d … 2>/dev/null || true` reads better in copy-pasted instructions.
- **🟢 Tag / `VERSION` drift is still unguarded** (`release.yml:74`, `VERSION:1`). Low harm: `generate-build-info.sh:14-17` prefers `CROW_VERSION`, so the *released* binary is always tag-labelled and can't mislabel itself. The drift only shows up for someone building from a checked-out tag. A one-line equality check in "Extract version from tag" would keep the two honest.
- **🟢 `COPYFILE_DISABLE=1 tar …`** avoids stray `._*` AppleDouble members on macOS and makes the archive deterministic.

**Verified clean:**

- `shellcheck scripts/*.sh` — exit 0, mirroring the CI gate including the new script.
- The universal build produces genuinely fat binaries — `crow: x86_64 arm64`, `crowd: x86_64 arm64` — so last round's `--product` Red is fixed and the new `lipo -archs` assertion works as intended.
- `.gitignore` now covers `release-staging/`, `crow-*.tar.gz`, and `crow-*.tar.gz.sha256`; `trap cleanup EXIT` handles mid-run failures.
- `generate_release_notes: true` alongside a custom `body` is valid — GitHub appends the generated notes to the supplied body, so the install instructions sit above the changelog.
- The `.sha256` records a bare basename (the script `cd`s to `$ROOT_DIR` first), so `shasum -a 256 -c` works in the user's download directory as written.
- `CHANGELOG.md`: `[Unreleased]` is left empty and the backfill moves under `[0.2.0] - 2026-08-05` (today), matching Keep-a-Changelog. The file uses inline links with no reference-definition block, so no `[0.2.0]: <url>` is missing.
- Dropping `bundle.sh` didn't drop a web build step — the web assets are committed under `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/`, not generated.
- `fetch-depth: 0` is retained, which `generate-build-info.sh:26` needs for `git rev-list --count HEAD`.

I did not run `crowd` locally to demonstrate the crash in #1: `ConfigStore` resolves its data directory via `FileManager.default.homeDirectoryForCurrentUser`, which ignores a `HOME` override, so an isolated run would have written to live Crow state. The build output, the tarball listing, and the generated accessor above are the evidence instead.

### Summary Table

| # | Finding | Location |
|---|---------|----------|
| 🔴 1 | Tarball omits `CrowDaemon`/`CrowTerminal` resource bundles — shipped `crowd` `fatalError`s on first web request or terminal spawn | `scripts/package-release.sh:39-52` |
| 🟡 2 | No CI step ever executes the packaged artifact outside the build tree | `.github/workflows/release.yml:76-79` |
| 🟢 3 | Checksum is same-origin; `actions/attest-build-provenance` would restore provenance | `.github/workflows/release.yml` |
| 🟢 4 | Unpinned third-party actions; `contents: write` broader than the release job | `.github/workflows/release.yml` |
| 🟢 5 | Six Apple secrets now orphaned — revoke and remove | repo settings |
| 🟢 6 | `make build CONFIG=release` needs Rust; `make daemon` is the Swift-only target | `release.yml:107`, `README.md:130`, `docs/getting-started.md:172` |
| 🟢 7 | `xattr -d` exits non-zero when the attribute is absent, which is the common case after `tar` | `release.yml:104`, `docs/troubleshooting.md:82` |
| 🟢 8 | Tag / `VERSION` drift unguarded | `release.yml:74`, `VERSION:1` |
| 🟢 9 | `COPYFILE_DISABLE=1` for a deterministic archive | `scripts/package-release.sh:52` |

The direction remains right and this round genuinely closed four of the previous findings — the universal build, the arch assertion, checksums, and the local-artifact hygiene are all correct now. What's left is the one defect that survives every gate the workflow has: the artifact is fat, checksummed, and broken. Fixing #1 and adding the smoke step in #2 together mean the next tag push is a verification rather than an experiment.

**Recommendation:** Request Changes — driven by [1 Red, 1 Yellow, 7 Green] findings.

---

dhilgaertner (changes requested):
## Code & Security Review

The bundle-packaging fix from round 3 lands correctly, and the smoke test is the right instinct — it turns "did we ship a `crowd` that can't find `Bundle.module`?" into a build-time failure instead of a user-time `fatalError`. I verified the bundle path end to end: `/version.json` and `/` both resolve through `Bundle.module` (`StaticAssets.swift:73-77`), so a missing `.bundle` really does fail the build now.

Two blockers remain. One makes the first tagged run impossible; the other makes the script unsafe to run on a developer machine.

### Critical Issues

#### 🔴 1. The new version guard rejects every prerelease tag — including the one the test plan calls for

`.github/workflows/release.yml:73-81`

```bash
CROW_VERSION="${GITHUB_REF_NAME#v}"
FILE_VERSION="$(tr -d '[:space:]' < VERSION)"
if [ "$CROW_VERSION" != "$FILE_VERSION" ]; then
  exit 1
fi
```

Push `v0.2.0-rc.1` → `CROW_VERSION=0.2.0-rc.1`, `FILE_VERSION=0.2.0` → mismatch → the job dies at step 3.

This contradicts the same file eleven lines later:

```yaml
prerelease: ${{ contains(github.ref_name, '-') }}    # release.yml:95
```

That expression can only ever be `true` for a tag containing `-`, and no such tag can now reach it — so the prerelease path is dead code. It also makes this PR's own test plan unexecutable: *"Merge and push a prerelease tag (`v0.2.0-rc.1`) to verify the workflow completes green."* It cannot complete green.

Compare the release core instead, so `-rc.1` / `-beta.2` still validate against `VERSION`:

```bash
if [ "${CROW_VERSION%%-*}" != "$FILE_VERSION" ]; then
  echo "ERROR: tag version (${CROW_VERSION%%-*}) does not match VERSION file ($FILE_VERSION)" >&2
  exit 1
fi
```

The guard itself is a good idea — a tag/VERSION mismatch would otherwise ship an archive whose directory name lies about the binary inside it. It just needs to admit prereleases.

#### 🔴 2. The smoke test boots `crowd` against the developer's **real** dev root, store, and telemetry DB — and leaves `{devRoot}/.claude/bin/crow` dangling

`scripts/package-release.sh:71-89`

```bash
./crowd --socket-path "$SMOKE_SOCKET" --host 127.0.0.1 --http-port "$SMOKE_PORT" &
```

`--socket-path` and the randomized port isolate the two things that would have failed *loudly*. The state that corrupts *silently* is untouched.

**The dev root.** No `--dev-root` and no `CROW_DEV_ROOT`, so `DaemonOptions.parse` falls through to the App Support pointer (`CrowDaemon.swift:1315-1322`):

```swift
if !devRootExplicit {
    if let configured = ConfigStore.loadDevRoot() {
        options.devRoot = configured
        options.devRootConfigured = true      // ← gate opens
    }
}
```

On CI that pointer is absent, so `devRootConfigured` stays `false` and nothing happens. On any machine that has ever run Crow it resolves to the user's real dev root and opens the `LaunchScaffold.run` gate (`CrowDaemon.swift:108-109`), which rewrites that root's `.claude/skills/`, `CLAUDE.md`, and `settings.local.json`.

**The dangling symlink.** Inside that scaffold, `ClaudeHookConfigWriter.ensureCrowCLISymlink` re-points `{devRoot}/.claude/bin/crow` at `appCrowBinary()`, which is `dirname(argv[0]) + "/crow"` (`ClaudeHookConfigWriter.swift:249-255`). `crowd` was launched from `$STAGING_DIR`, where a `crow` was copied moments earlier — so the link now targets `$STAGING_DIR/crow`, and the `cleanup` trap at line 27 `rm -rf`s exactly that directory on exit. `Scaffolder.swift:206-208` says that symlink is "what makes a written hook command survive the worktree its build came from being deleted (#897)", and `LaunchScaffold.repairStaleHooks` spells out the fallout: a `crow` that no longer exists makes "every hook event in that directory fail, which is noisy *and* silently drops all telemetry for sessions run there." It self-heals on the next real `crowd` start, but until then the developer's hooks are broken by a packaging script.

**The store and telemetry DB.** Both live at fixed paths — `AppSupportDirectory.swift:10-25` and `TelemetryService.swift:95-99` resolve `~/Library/Application Support/crow/{store.json,telemetry.db}` with **no flag or env override**. So `--dev-root` alone will not fix this: the smoke-test daemon opens the user's live store as a second writer, which is the precise hazard `JSONStore`'s own `EXTERNAL WRITE DETECTED … a second writer is touching store.json` guard (`JSONStore.swift:256-266`) and CLAUDE.md's #728 note exist to catch. `startBoardPoll`'s first tick also calls `ensureManagerSession`, so a slow-enough boot can spawn a Manager session against real state before the `kill`.

Isolating `HOME` is what actually contains this, since that is what the App Support lookups key off:

```bash
SMOKE_HOME="$(mktemp -d)"
trap 'rm -rf "$SMOKE_HOME"' EXIT      # inside the subshell, alongside the kill
HOME="$SMOKE_HOME" ./crowd \
  --dev-root "$SMOKE_HOME/devroot" \
  --socket-path "$SMOKE_SOCKET" --host 127.0.0.1 --http-port "$SMOKE_PORT" &
```

Point `--dev-root` at scratch, **not** at `$STAGING_DIR` — passing it sets `devRootConfigured = true`, which *enables* scaffolding, and `$STAGING_DIR` is tarred four lines later (line 92). Scaffolding into the staging dir would ship `.claude/` into the release archive.

### Code Quality

#### 🟡 3. The readiness probe can't distinguish "still booting" from "the bundle is broken" — blunting the exact signal this test was added for

`scripts/package-release.sh:79-88`

`curl -sf` treats a 404 as failure, and `webResponse` returns 404 rather than crashing when a resource is missing (`StaticAssets.swift:73-77`). So a `version.json` that didn't make it into the bundle spins the full 30 s and reports:

```
ERROR: crowd smoke test timed out waiting for http://127.0.0.1:18437
```

which points at daemon startup, not at the missing asset. Since catching a missing-resource regression is the whole reason this block exists, it should say so. Use a bundle-independent readiness signal (e.g. poll until the port accepts a connection, or until `/auth/check` returns 204 — `StaticAssets.swift:34` serves it without touching `Bundle.module`), then assert the bundle-backed routes once with a distinct error message.

#### 🟡 4. `docs/troubleshooting.md:77` still says `make build` — the wording this PR fixes everywhere else

The PR correctly replaces `make build` with `make daemon` in `README.md:130` and `docs/getting-started.md:164,172`, because `make build` is `daemon app` and the `app` target hard-errors without the Rust arm64 toolchain (`Makefile:22,30-34`). The sentence directly above the new paragraph still reads:

> Crow builds two command-line binaries — `crow` and `crowd`. … `make build` produces unsigned but fully functional binaries

Same fix, same file, one line up.

### Security Review

**Strengths:**

- Deleting the signing path is a real supply-chain win: five secrets leave the runner (`DEVELOPER_CERTIFICATE_BASE64`, `DEVELOPER_CERTIFICATE_PASSWORD`, `APPLE_ID`, `APPLE_APP_SPECIFIC_PASSWORD`, `APPLE_TEAM_ID`), along with the decoded `.p12` on disk, the ad-hoc keychain, and `security set-key-partition-list -S apple-tool:,apple:`.
- Publishing a SHA-256 alongside the tarball and documenting `shasum -c` is the right compensating control. The limitation is disclosed honestly rather than oversold — the checksum ships next to the artifact, so it detects corruption, not a compromised release. That is inherent to unsigned distribution, and `docs/troubleshooting.md` says so.
- The tag ↔ `VERSION` guard is a genuine integrity check on the artifact's identity (once issue 1 is fixed).
- `lipo -archs` is asserted per binary per arch (`package-release.sh:44-52`) rather than printed, so a thin binary fails the build instead of shipping as "universal".
- `COPYFILE_DISABLE=1` on the `tar` keeps AppleDouble `._*` files out of the archive.
- `permissions: contents: write` remains correctly scoped, and no new secret is referenced by the workflow or the script. `CROW_VERSION` is the only external input and is checked non-empty at `package-release.sh:16-19`.
- `shellcheck scripts/*.sh` passes clean on the new script (verified locally); the `shellcheck` job already gates it.

**Concerns:**

- Issue 2 above is the only one, and it is a local-developer-environment integrity problem rather than a released-artifact one.

### Minor (non-blocking)

- 🟢 `package-release.sh:26-29` — the `cleanup` trap `rm -rf`s `$STAGING_ROOT` on *any* exit, including a failed smoke test, deleting the artifact you'd most want to inspect. Consider retaining it on failure.
- 🟢 `package-release.sh:70` — `SMOKE_PORT=$((18000 + RANDOM % 1000))` is a single attempt; a bound port surfaces as the same generic 30 s timeout. A short retry loop, or letting the kernel assign, would be more robust.
- 🟢 `package-release.sh:69` — the socket path under macOS's `/var/folders/...` `TMPDIR` lands around 80 bytes against the 104-byte `sun_path` limit. Fine today, thin margin.

### Summary Table

**Recommendation:** Request Changes — driven by [2 Red, 2 Yellow, 3 Green] findings.

---

dgershman:
Re: smoke-test isolation (#2) — implemented `HOME` + `--dev-root` scratch as suggested, but verified on this machine that macOS `FileManager.urls(for: .applicationSupportDirectory, …)` still resolves to the real `~/Library/Application Support/crow` even with `HOME` overridden, so a local `crowd` smoke test would still contend for the store writer lock / touch live state. The `crowd` smoke therefore runs only when `GITHUB_ACTIONS` is set (clean CI runners); `crow --version` from the staging dir still runs locally.

dhilgaertner (approved):
## Code & Security Review

Round 5. Both blockers from the previous round are fixed, and I verified each:

**1. Prerelease version guard** (`.github/workflows/release.yml:73-82`) — `RELEASE_VERSION="${CROW_VERSION%%-*}"` strips the prerelease suffix before comparison, so `v0.2.0-rc.1` → `0.2.0` matches `VERSION`, while `CROW_VERSION` keeps the full `0.2.0-rc.1` for artifact naming and `CLIVersion`. The test plan's own tag now passes the guard.

**2. Developer-machine safety** (`scripts/package-release.sh:84-87`) — the `crowd` half of the smoke test is gated on `GITHUB_ACTIONS`, and the local path exits *before* the cleanup trap is installed. The inline comment records why (App Support isn't relocatable via `HOME`), matching your PR comment.

### What I verified

The packaging path has no test coverage, so I reproduced it end to end with a synthetic package mirroring Crow's layout (root executable → `Packages/Lib` with resources → `Packages/Deep` with resources):

- **The universal build emits nested-package bundles into the same products dir.** `swift build -c release --arch arm64 --arch x86_64` put *both* `Lib_Lib.bundle` and `Deep_Deep.bundle` in `.build/apple/Products/Release/`, so the `"$BUILD_DIR"/*.bundle` glob (`package-release.sh:63`) really does pick up `CrowTerminal`'s bundle alongside `CrowDaemon`'s — not just the one the smoke test exercises.
- **The documented install works.** tarball → extract → `ln -s` into a PATH dir → build tree deleted → both `Bundle.module` lookups resolved. Worth recording *why*, because it is not obvious: I measured `Bundle.main.bundleURL` and `Bundle.main.resourceURL` both reporting the **symlink's** directory (`~/.local/bin`), so the only candidate that rescues a symlinked install is `Bundle(for: BundleFinder.self).resourceURL`, which resolves via the dyld image path. The xcbuild-generated accessor includes it, so `ln -sf ~/.local/lib/crow-<v>/crowd ~/.local/bin/crowd` is safe.
- `shasum -a 256 -c` round-trips from the download directory, and `COPYFILE_DISABLE=1` keeps AppleDouble `._*` entries out of the archive.
- `shellcheck scripts/*.sh` passes locally (the gating CI job).
- Nothing in the repo references the old `crow-cli-macos-universal` asset name, and no test pins `0.1.0`, so the asset rename and the `VERSION` bump break nothing in-tree.

I did not run the Swift test suite — this diff touches no Swift sources.

### Security Review

**Strengths:**
- Five secrets (`DEVELOPER_CERTIFICATE_BASE64`, `DEVELOPER_CERTIFICATE_PASSWORD`, `DEVELOPER_ID_APPLICATION`, `APPLE_ID`, `APPLE_APP_SPECIFIC_PASSWORD`, `APPLE_TEAM_ID`) no longer reach the runner. No `.p12` written to disk, no `security create-keychain` / `set-key-partition-list`, and no cleanup step that can strand a keychain on a cancelled job.
- `permissions: contents: write` stays minimal and workflow-scoped.
- Publishing a `.sha256` next to the tarball is new and correct.
- The version guard is itself a supply-chain check: a tag disagreeing with `VERSION` can no longer produce a mislabelled release.

**Observations (not blocking):**
- Unsigned artifacts plus the `xattr -d com.apple.quarantine` guidance is a deliberate, already-settled trade-off (ADR 0010 removed the bundling scripts; the PR body and `docs/troubleshooting.md#unsigned-builds` both say so). Since the checksum is served from the same release page, it defends against transfer corruption rather than a compromised release — worth stating plainly in the troubleshooting section.
- `softprops/action-gh-release@v2` is a floating major tag on a job holding `contents: write`. Pre-existing and untouched here; pinning to a commit SHA would be a reasonable follow-up.

### Code Quality — all non-blocking

1. **`scripts/package-release.sh:62-67`** — the guard asserts *at least one* bundle, while the file header names two specific ones. I confirmed both land today, so this is drift protection rather than a live bug: if `CrowTerminal`'s `resources:` were later renamed or dropped, the glob would still find `CrowDaemon`'s bundle, the guard would pass, and `crowd` would `fatalError` at first terminal spawn. Asserting the set by name is about three lines.

2. **`scripts/package-release.sh:113-121`** — the smoke test proves `CrowDaemon`'s bundle (`/version.json`, `/`) but never `CrowTerminal`'s tmux/xterm resources, which back the daemon's primary feature. A `crow new-terminal`-shaped probe would close the gap the rest of the script is built to close.

3. **`scripts/package-release.sh:102`** — `/auth/check` is explicitly a **gated** path; `WebAuthRoutes.swift:83-85` lists the exempt set as `/login`, `/logout`, `/health`, `/brand.svg`. It answers 200 here only because a loopback peer with no `X-Forwarded-For` is auto-authorized (`WebAuth.swift:174-176`). `/health` (`CrowDaemon.swift:456`) is the purpose-built exempt probe and won't start failing if that local-direct rule is ever tightened.

4. **`scripts/package-release.sh:91`** — `wait "$crowd_pid" 2>/dev/null` runs under `set -e`. The happy path is fine: Hummingbird's `runService(gracefulShutdownSignals: [.sigterm, .sigint])` handles SIGTERM, so `crowd` exits 0. But if `crowd` died during startup, the ready-loop's `exit 1` fires the trap, `kill`/`wait` can return non-zero, and errexit aborts the trap *before* `rm -rf` — leaking `$SMOKE_HOME` on the runner and replacing the intended exit status. `|| true` on both calls.

5. **`scripts/package-release.sh:26-34`** — the EXIT trap is installed before `$STAGING_ROOT` is created, so a failure in `generate-build-info.sh` or `swift build` prints "Keeping … for inspection" for a directory that never existed. Cosmetic.

6. **Docs** — swapping `make build CONFIG=release` → `make daemon CONFIG=release` is right, since `make build` is `daemon app` and drags in the Tauri/Rust toolchain for binaries `make install` only symlinks. `make build` survives in `README.md:87,143`, `docs/getting-started.md:17,56`, and `docs/troubleshooting.md:7-8` — correct in those contexts, but a consistency pass would be welcome eventually.

### Summary Table

The test plan's open boxes (push `v0.2.0-rc.1`, `lipo -info` both binaries, run on a clean machine) remain the real proof, and nothing I found should block getting there.

PR #941, opened 2026-08-05
Detect when a newer Crow build is available (CROW-938)

### LINKED ISSUE #938: Detect and surface when a newer Crow version is available
## Problem

Crow can't tell you it's out of date. Running an old build looks identical to running the newest one.

The local half of this already exists — `scripts/generate-build-info.sh` stamps `VERSION` into `CLIVersion.swift` and writes `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/version.json` (`version`, `gitSha`, `buildDate`), which **Settings → About already renders**. What's missing is any notion of *latest*, and therefore any comparison.

## Why the obvious approach doesn't work yet

Comparing `VERSION` against the newest GitHub Release would compare against nothing:

- `corveil/crow` has **zero git tags and zero GitHub Releases** (`gh api repos/corveil/crow/releases` → `0`). Same for `radiusmethod/crow`.
- `VERSION` has been pinned at `0.1.0` while ~310 commits landed; `CHANGELOG.md` has all of them sitting under `[Unreleased]`.
- `.github/workflows/release.yml` triggers on `v*` tags and calls `scripts/bundle.sh` and `scripts/sign-and-notarize.sh` — **neither script exists.** They were deleted when the macOS app was retired ([ADR 0010](docs/adr/0010-retire-the-macos-app.md)), but the workflow still references them, so a `v*` tag push fails at the "Bundle Crow.app" step.

Crow is also installed by building from source, not by downloading a release artifact, so semver-vs-latest-release is the wrong primitive today regardless.

## Proposed approach — compare against the branch head, not a release

Define "newer version available" as **the local build's `gitSha` being behind the default branch on GitHub**. That's checkable today with zero release-pipeline work and matches how Crow is actually installed.

```
crow version --check

  crow 0.1.0 (f56939d, built 2026-08-05)
  origin/main is at a1b2c3d (2026-08-05)
  → 42 commits behind

  Update:  git -C <clone> pull && make install
```

Mechanism: `GET /repos/corveil/crow/compare/<localSha>...main` returns `behind_by` / `ahead_by` / `status` in a single call — no cloning, no local git required, and it works from a binary installed outside its source tree.

**Alternative, if you'd rather do it properly first:** repair `release.yml`, cut a real `v0.2.0`, and compare `VERSION` against the latest release tag. Better long-term primitive, but it blocks this feature behind a CI repair. A middle path is to write the check so it *prefers* releases when any exist and falls back to branch-head comparison when none do — works now, needs no rewrite after the first real release.

**Implementer: pick one and say which in the PR description.**

## Surfaces

Per [ADR 0016](docs/adr/0016-cli-control-plane-parity.md) (every user-mutable capability has a CLI path or a written exemption):

- **CLI verb** — `crow version --check` (or `crow update check`). Non-negotiable: the Manager agent drives Crow through the CLI, so a browser-only check is a capability no agent can reach.
- **Settings → About tab** — already renders `/version.json`. Add an inline up-to-date / N-behind indicator plus a "check now" action.
- **Passive indicator** — a dismissible "update available" hint in the web UI when the daemon's periodic check finds the build behind.

## Design constraints to respect

- **`crowd` is the authority ([ADR 0009](docs/adr/0009-crowd-sole-authority-clients-only.md)).** The network call and any cached result live in `crowd`; the web UI renders what it's told and does not call GitHub itself.
- **Rate limits.** Unauthenticated GitHub API is 60 req/hr/IP. Check at most once per daemon start plus a long interval (≥6h), cache the result, and fail *silently* — an update check must never produce a startup error or a red banner because GitHub was unreachable. Reuse whatever auth the board poller already holds if available, but don't make the check *depend* on a configured token.
- **Offline / detached builds.** A build whose `gitSha` is `dev` (the generator's fallback outside a git repo), or a SHA GitHub doesn't recognize (local-only commits, a fork), must report **"can't determine"** — never "up to date", never an error.
- **ADR 0016 file set.** A new verb (and config group, if added) requires: shared handler in `Packages/CrowEngine/.../*RPCSupport.swift`, RPC registration in `RPCHandlers.swift`, the verb under `Packages/CrowCLI/.../Commands/`, registration in `CrowCommand.swift` `subcommands`, a `ParityLedger.swift` row, tests, `docs/cli.md` via `make docs`, a worked example in `docs/cli-reference.md`, `CLAUDE.md` + `Resources/CLAUDE.md.template`, and a `CHANGELOG.md` entry. `ParityGateTests` / `check-cli-parity.sh` / `CLIDocsTests` go red otherwise. If a config group is added, populate it in `parityProbeConfig()` too — and **never write a null** into it, since one null makes the whole `config.json` undecodable.
- **Opt-out.** Whatever settings group this adds needs an enable/disable toggle. A dev tool that phones home on every start with no off switch is a bug.

## Out of scope

- **Self-update / auto-install.** This detects and reports only. It does not pull, rebuild, or restart anything.
- **Repairing `release.yml`.** The missing `bundle.sh` / `sign-and-notarize.sh` are a real but separate defect — filed on its own. This feature must not depend on it.

## Acceptance criteria

- [ ] `crow version --check` reports the local version + SHA, the remote head, and whether the build is behind.
- [ ] An up-to-date build says so and exits `0`; a behind build says how far behind and how to update.
- [ ] GitHub unreachable / rate-limited / unknown SHA → a clear "couldn't check" result. Never a crash, never a false "up to date".
- [ ] Settings → About shows the same state without a manual CLI run.
- [ ] The periodic check is off-able, and its interval can't threaten the 60 req/hr unauthenticated limit.
- [ ] `ParityGateTests`, `RPCLedgerParityTests`, `check-cli-parity.sh`, and `CLIDocsTests` all pass.

dgershman:
Closes #938

## Approach

**Branch-head comparison** against `corveil/crow` `main` via GitHub's compare API (`GET /repos/corveil/crow/compare/{localSha}...main`). This works today with zero release-pipeline work and matches how Crow is actually installed (build from source). Release-tag comparison can be layered on later when real GitHub Releases exist.

## Summary

- `crowd` owns the network call, caches the result, and checks once after startup plus on a configurable interval (default 6h, minimum 6h, opt-out via `versionUpdate.enabled`)
- Reuses `gh auth token` when available but does not depend on it
- `dev` / unknown / fork SHAs → `unknown` (never false up-to-date); GitHub errors fail silently
- **CLI:** `crow version`, `crow version --check` / `crow version check`, `crow version get|set`
- **Settings → About:** inline status, opt-out toggle, interval picker, Check now
- **Web UI:** dismissible banner when behind (keyed on remote SHA so new commits re-show)

## Test plan

- [x] `swift test --package-path Packages/CrowEngine --filter VersionUpdate`
- [x] `swift test --package-path Packages/CrowCLI --filter 'ParityGate|CLIDocs'`
- [x] `swift test --package-path Packages/CrowDaemon --filter 'VersionUpdate|RPCLedger'`
- [x] `scripts/check-cli-parity.sh`
- [ ] Manual: `crow version --check` on an up-to-date build → exit 0
- [ ] Manual: Settings → About shows matching status without a CLI run
- [ ] Manual: dismiss banner, confirm it stays dismissed until remote SHA changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- CROW-938 — Detect when a newer Crow build is available by comparing the stamped `gitSha` against `corveil/crow` `main` (no release tags required). `crowd` owns the GitHub compare call, caches the result, and checks at most once per daemon start plus every ≥6 hours (configurable, opt-out via `versionUpdate.enabled`). Surfaces: `crow version --check` / `crow version check` (human summary with exit codes), `crow version get|set`, Settings → About (status + Check now), and a dismissible web banner when behind. Offline, rate-limited, `dev`, or unknown SHAs report `unknown` — never a false up-to-date.
+
 ## [0.2.0] - 2026-08-05
 
 Backfill of merged PRs since the 0.1.0 release, grouped by theme.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -85,6 +85,11 @@ crow cleanup get                                                        → {"cl
 crow cleanup set [--enabled true|false] [--retention-hours N]           → patch; live within ~1 board poll. Deletes completed/archived sessions incl. worktree + branch
 crow ui get                                                             → {"ui":{"sidebar":{"hide_session_details":…}}}
 crow ui set --hide-session-details true|false                           → patch; connected browsers repaint within ~2s
+crow version                                                            → prints the stamped build version
+crow version --check                                                    → compare against corveil/crow main; human summary, exit 0/1/2
+crow version check                                                      → same as --check
+crow version get                                                        → {"version_update":{…},"status":{…}}
+crow version set [--enabled true|false] [--interval-hours N]            → patch; interval floored at 6h
 ```
 
 ### Defaults
--- a/Packages/CrowCLI/Sources/CrowCLILib/Commands/VersionCommands.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Commands/VersionCommands.swift
@@ -45,7 +45,13 @@ public struct VersionCheckCmd: ParsableCommand {
 /// Shared implementation for `crow version --check` and `crow version check`.
 enum VersionCheck {
     static func run() throws {
-        let result = try rpc("version-update-check", params: ["force": .bool(true)])
+        let result: [String: JSONValue]
+        do {
+            result = try rpc("version-update-check", params: ["force": .bool(true)])
+        } catch {
+            warn("Could not check for updates — is crowd running?")
+            throw ExitCode(2)
+        }
         guard let status = result["status"], status != .null else {
             warn("Could not check for updates — is crowd running?")
             throw ExitCode(2)
--- a/Packages/CrowCLI/Sources/CrowCLILib/Commands/VersionCommands.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Commands/VersionCommands.swift
@@ -1,0 +1,154 @@
+import ArgumentParser
+import CrowIPC
+import Foundation
+
+// MARK: - crow version
+
+/// `crow version` — local build info and upstream comparison (CROW-938).
+public struct Version: ParsableCommand {
+    public static let configuration = CommandConfiguration(
+        commandName: "version",
+        abstract: "Show the local build and check for upstream updates",
+        discussion: """
+        Bare `crow version` prints the stamped build version. Pass `--check` or run \
+        `crow version check` to compare against corveil/crow main via the daemon.
+        """,
+        subcommands: [VersionCheckCmd.self, VersionGet.self, VersionSet.self]
+    )
+
+    @Flag(name: .long, help: "Compare this build against corveil/crow main")
+    var check: Bool = false
+
+    public init() {}
+
+    public func run() throws {
+        if check {
+            try VersionCheck.run()
+            return
+        }
+        print(CLIVersion.version)
+    }
+}
+
+public struct VersionCheckCmd: ParsableCommand {
+    public static let configuration = CommandConfiguration(
+        commandName: "check",
+        abstract: "Compare this build against corveil/crow main")
+
+    public init() {}
+
+    public func run() throws {
+        try VersionCheck.run()
+    }
+}
+
+/// Shared implementation for `crow version --check` and `crow version check`.
+enum VersionCheck {
+    static func run() throws {
+        let result = try rpc("version-update-check", params: ["force": .bool(true)])
+        guard let status = result["status"], status != .null else {
+            fputs("Could not check for updates — is crowd running?\n", stderr)
+            throw ExitCode(2)
+        }
+        printHuman(status)
+        let state = status.objectValue?["state"]?.stringValue ?? "unknown"
+        switch state {
+        case "up_to_date": throw ExitCode(0)
+        case "behind": throw ExitCode(1)
+        default: throw ExitCode(2)
+        }
+    }
+
+    private static func printHuman(_ status: JSONValue) {
+        let obj = status.objectValue ?? [:]
+        let version = obj["local_version"]?.stringValue ?? "?"
+        let sha = obj["local_sha"]?.stringValue ?? "?"
+        let buildDate = obj["build_date"]?.stringValue ?? ""
+        var localLine = "crow \(version) (\(sha)"
+        if !buildDate.isEmpty { localLine += ", built \(buildDate)" }
+        localLine += ")"
+        print(localLine)
+
+        let state = obj["state"]?.stringValue ?? "unknown"
+        switch state {
+        case "up_to_date":
+            if let remoteSha = obj["remote_sha"]?.stringValue {
+                var remoteLine = "origin/main is at \(remoteSha)"
+                if let remoteDate = obj["remote_date"]?.stringValue, !remoteDate.isEmpty {
+                    remoteLine += " (\(remoteDate))"
+                }
+                print(remoteLine)
+            }
+            print("Up to date.")
+        case "behind":
+            if let remoteSha = obj["remote_sha"]?.stringValue {
+                var remoteLine = "origin/main is at \(remoteSha)"
+                if let remoteDate = obj["remote_date"]?.stringValue, !remoteDate.isEmpty {
+                    remoteLine += " (\(remoteDate))"
+                }
+                print(remoteLine)
+            }
+            if let behind = obj["behind_by"]?.intValue {
+                print("→ \(behind) commit\(behind == 1 ? "" : "s") behind")
+            }
+            if let cmd = obj["update_command"]?.stringValue {
+                print("")
+                print("Update:  \(cmd)")
+            }
+        default:
+            let reason = obj["reason"]?.stringValue ?? "Could not determine update status"
+            print(reason)
+        }
+    }
+}
+
+public struct VersionGet: ParsableCommand {
+    public static let configuration = CommandConfiguration(
+        commandName: "get",
+        abstract: "Show version-update settings and the cached check result")
+
+    public init() {}
+
+    public func run() throws {
+        printJSON(try rpc("version-update-get"))
+    }
+}
+
+public struct VersionSet: ParsableCommand {
+    public static let configuration = CommandConfiguration(
+        commandName: "set",
+        abstract: "Change version-update settings",
+        discussion: """
+        Only the flags you pass change; at least one is required. \
+        --interval-hours is floored at 6 so unauthenticated GitHub checks cannot \
+        exhaust the 60 req/hr limit.
+        """
+    )
+
+    @Option(name: .long, help: "Enable periodic upstream checks (true or false)")
+    var enabled: Bool?
+
+    @Option(
+        name: .customLong("interval-hours"),
+        help: "Hours between automatic checks (minimum 6, default 6)")
+    var intervalHours: Int?
+
+    public init() {}
+
+    public func validate() throws {
+        guard enabled != nil || intervalHours != nil else {
+            throw ValidationError(
+                "Nothing to set — provide at least one of --enabled, --interval-hours.")
+        }
+        if let intervalHours, intervalHours < 6 {
+            throw ValidationError("--interval-hours must be at least 6.")
+        }
+    }
+
+    public func run() throws {
+        var params: [String: JSONValue] = [:]
+        if let enabled { params["enabled"] = .bool(enabled) }
+        if let intervalHours { params["interval_hours"] = .int(intervalHours) }
+        printJSON(try rpc("version-update-set", params: params))
+    }
+}
--- a/Packages/CrowCLI/Sources/CrowCLILib/Commands/VersionCommands.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Commands/VersionCommands.swift
@@ -47,7 +47,7 @@ enum VersionCheck {
     static func run() throws {
         let result = try rpc("version-update-check", params: ["force": .bool(true)])
         guard let status = result["status"], status != .null else {
-            fputs("Could not check for updates — is crowd running?\n", stderr)
+            warn("Could not check for updates — is crowd running?")
             throw ExitCode(2)
         }
         printHuman(status)
--- a/Packages/CrowCLI/Sources/CrowCLILib/CrowCommand.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/CrowCommand.swift
@@ -79,6 +79,7 @@ public struct CrowCommand: ParsableCommand {
             HookEventCmd.self,
             CodexNotify.self,
             GenerateDocs.self,
+            Version.self,
         ]
     )
 
--- a/Packages/CrowCLI/Tests/CrowCLITests/ParityGateTests.swift
+++ b/Packages/CrowCLI/Tests/CrowCLITests/ParityGateTests.swift
@@ -174,6 +174,7 @@ struct ParityGateTests {
             terminal: TerminalSettings(),
             autoRespond: AutoRespondSettings(),
             cleanup: CleanupConfig(),
+            versionUpdate: VersionUpdateConfig(),
             jobs: [
                 JobConfig(
                     name: "probe",
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -50,6 +50,9 @@ public struct AppConfig: Codable, Sendable, Equatable {
     /// can still pick up previously-labeled issues.
     public var autoCreateWatcherEnabled: Bool
     public var cleanup: CleanupConfig
+    /// Periodic check against `corveil/crow` `main` to surface when this build
+    /// is behind upstream (CROW-938). Off-able; interval floored at 6h.
+    public var versionUpdate: VersionUpdateConfig
     /// Scheduled jobs: named sets of prompts that fire automatically on a
     /// schedule, scoped to a repo. Driven by `JobScheduler` (CROW-317).
     public var jobs: [JobConfig]
@@ -144,6 +147,7 @@ public struct AppConfig: Codable, Sendable, Equatable {
         autoMergeWatcherEnabled: Bool = false,
         autoCreateWatcherEnabled: Bool = false,
         cleanup: CleanupConfig = CleanupConfig(),
+        versionUpdate: VersionUpdateConfig = VersionUpdateConfig(),
         jobs: [JobConfig] = [],
         defaultAgentKind: AgentKind = .claudeCode,
         agentsByKind: [String: AgentKind] = [:],
@@ -167,6 +171,7 @@ public struct AppConfig: Codable, Sendable, Equatable {
         self.autoMergeWatcherEnabled = autoMergeWatcherEnabled
         self.autoCreateWatcherEnabled = autoCreateWatcherEnabled
         self.cleanup = cleanup
+        self.versionUpdate = versionUpdate
         self.jobs = jobs
         self.defaultAgentKind = defaultAgentKind
         self.agentsByKind = agentsByKind
@@ -202,6 +207,8 @@ public struct AppConfig: Codable, Sendable, Equatable {
             autoRespond.autoRebaseAndResolveConflicts = true
         }
         cleanup = try container.decodeIfPresent(CleanupConfig.self, forKey: .cleanup) ?? CleanupConfig()
+        versionUpdate = try container.decodeIfPresent(VersionUpdateConfig.self, forKey: .versionUpdate)
+            ?? VersionUpdateConfig()
         jobs = try container.decodeIfPresent([JobConfig].self, forKey: .jobs) ?? []
         defaultAgentKind = try container.decodeIfPresent(AgentKind.self, forKey: .defaultAgentKind) ?? .claudeCode
         agentsByKind = try container.decodeIfPresent([String: AgentKind].self, forKey: .agentsByKind) ?? [:]
@@ -245,7 +252,7 @@ public struct AppConfig: Codable, Sendable, Equatable {
     }
 
     private enum CodingKeys: String, CodingKey {
-        case workspaces, defaults, notifications, sidebar, remoteControlEnabled, managerAutoPermissionMode, jobsAutoPermissionMode, reviewAutoPermissionMode, coderViewAutoPermissionMode, telemetry, terminal, autoRespond, attributionTrailers, autoMergeWatcherEnabled, autoCreateWatcherEnabled, cleanup, jobs, defaultAgentKind, agentsByKind, managerGateway, jiraCredential, webAuth
+        case workspaces, defaults, notifications, sidebar, remoteControlEnabled, managerAutoPermissionMode, jobsAutoPermissionMode, reviewAutoPermissionMode, coderViewAutoPermissionMode, telemetry, terminal, autoRespond, attributionTrailers, autoMergeWatcherEnabled, autoCreateWatcherEnabled, cleanup, versionUpdate, jobs, defaultAgentKind, agentsByKind, managerGateway, jiraCredential, webAuth
     }
 
     /// Resolve the agent that should drive a newly-created session of the
--- a/Packages/CrowCore/Sources/CrowCore/Models/VersionUpdate.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/VersionUpdate.swift
@@ -1,0 +1,114 @@
+import Foundation
+
+/// Settings for the periodic upstream version check (CROW-938).
+public struct VersionUpdateConfig: Codable, Sendable, Equatable {
+    /// When false, `crowd` does not phone GitHub on a schedule. Defaults to
+    /// true — the check is on unless the user opts out.
+    public var enabled: Bool
+    /// Minimum hours between automatic checks. Floored at 6 so an
+    /// unauthenticated GitHub client (60 req/hr/IP) cannot be exhausted by a
+    /// misconfigured interval.
+    public var intervalHours: Int
+
+    public static let minimumIntervalHours = 6
+
+    public init(enabled: Bool = true, intervalHours: Int = 6) {
+        self.enabled = enabled
+        self.intervalHours = max(Self.minimumIntervalHours, intervalHours)
+    }
+
+    public init(from decoder: Decoder) throws {
+        let c = try decoder.container(keyedBy: CodingKeys.self)
+        enabled = try c.decodeIfPresent(Bool.self, forKey: .enabled) ?? true
+        let raw = try c.decodeIfPresent(Int.self, forKey: .intervalHours) ?? Self.minimumIntervalHours
+        intervalHours = max(Self.minimumIntervalHours, raw)
+    }
+
+    enum CodingKeys: String, CodingKey { case enabled, intervalHours }
+}
+
+/// Build metadata stamped by `scripts/generate-build-info.sh` into
+/// `version.json` and served at `/version.json`.
+public struct BuildInfo: Codable, Sendable, Equatable {
+    public var version: String
+    /// Short SHA for display (`git rev-parse --short`).
+    public var gitSha: String
+    /// Full SHA for GitHub compare; falls back to `gitSha` when absent.
+    public var gitShaFull: String?
+    public var buildDate: String
+
+    public init(version: String, gitSha: String, gitShaFull: String? = nil, buildDate: String) {
+        self.version = version
+        self.gitSha = gitSha
+        self.gitShaFull = gitShaFull
+        self.buildDate = buildDate
+    }
+
+    public init(from decoder: Decoder) throws {
+        let c = try decoder.container(keyedBy: CodingKeys.self)
+        version = try c.decodeIfPresent(String.self, forKey: .version) ?? "?"
+        gitSha = try c.decodeIfPresent(String.self, forKey: .gitSha) ?? "dev"
+        gitShaFull = try c.decodeIfPresent(String.self, forKey: .gitShaFull)
+        buildDate = try c.decodeIfPresent(String.self, forKey: .buildDate) ?? ""
+    }
+
+    /// SHA sent to GitHub's compare API — full when stamped, else short.
+    public var compareSha: String { gitShaFull ?? gitSha }
+
+    /// Whether this build's SHA is too ambiguous for an upstream comparison.
+    public var isIndeterminate: Bool {
+        let sha = compareSha.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
+        return sha.isEmpty || sha == "dev"
+    }
+
+    enum CodingKeys: String, CodingKey { case version, gitSha, gitShaFull, buildDate }
+}
+
+/// Cached result of comparing the local build against `origin/main`.
+public struct VersionUpdateStatus: Codable, Sendable, Equatable {
+    public enum State: String, Codable, Sendable {
+        case upToDate = "up_to_date"
+        case behind
+        case unknown
+    }
+
+    public var state: State
+    public var localVersion: String
+    public var localSha: String
+    public var buildDate: String
+    public var remoteSha: String?
+    public var remoteDate: String?
+    public var behindBy: Int?
+    public var aheadBy: Int?
+    public var updateCommand: String?
+    /// Human-readable reason when `state == .unknown`.
+    public var reason: String?
+    /// Epoch milliseconds when this result was produced.
+    public var checkedAtMs: Int64?
+
+    public init(
+        state: State,
+        localVersion: String,
+        localSha: String,
+        buildDate: String,
+        remoteSha: String? = nil,
+        remoteDate: String? = nil,
+        behindBy: Int? = nil,
+        aheadBy: Int? = nil,
+        updateCommand: String? = nil,
+        reason: String? = nil,
+        checkedAtMs: Int64? = nil
+    ) {
+        self.state = state
+        self.localVersion = localVersion
+        self.localSha = localSha
+        self.buildDate = buildDate
+        self.remoteSha = remoteSha
+        self.remoteDate = remoteDate
+        self.behindBy = behindBy
+        self.aheadBy = aheadBy
+        self.updateCommand = updateCommand
+        self.reason = reason
+        self.checkedAtMs = checkedAtMs
+    }
+}
--- a/Packages/CrowCore/Sources/CrowCore/Parity/ParityLedger.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Parity/ParityLedger.swift
@@ -216,6 +216,9 @@ public enum ParityLedger {
         .write("cleanup-set", cli: "cleanup set"),
         .read("ui-get", cli: "ui get"),
         .write("ui-set", cli: "ui set"),
+        .read("version-update-get", cli: "version get"),
+        .write("version-update-set", cli: "version set"),
+        .write("version-update-check", cli: "version check"),
         .read("notifications-get", cli: "notifications get"),
         .write("notifications-set", cli: "notifications set"),
         .read(
@@ -332,6 +335,9 @@ public enum ParityLedger {
 
         .field("cleanup.enabled", read: "cleanup get", write: "cleanup set"),
         .field("cleanup.retentionHours", read: "cleanup get", write: "cleanup set"),
+
+        .field("versionUpdate.enabled", read: "version get", write: "version set"),
+        .field("versionUpdate.intervalHours", read: "version get", write: "version set"),
 
         .field("sidebar.hideSessionDetails", read: "ui get", write: "ui set"),
 
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
@@ -386,6 +386,17 @@ public enum CrowDaemon {
         }
         if let jobScheduler { startJobPoll(scheduler: jobScheduler) }
 
+        // Upstream version check (CROW-938): compare the stamped build SHA against
+        // corveil/crow main. Runs on a long interval and caches the result in
+        // memory — failures are silent and never block startup.
+        let buildInfo = BuildInfoLoader.load(webDir: options.webDir)
+            ?? BuildInfo(version: "?", gitSha: "dev", buildDate: "")
+        let versionUpdateService = await MainActor.run {
+            VersionUpdateService(buildInfo: buildInfo, devRoot: options.devRoot)
+        }
+        startVersionUpdatePoll(
+            service: versionUpdateService, devRoot: options.devRoot, eventHub: eventHub)
+
         // Delegate any method the daemon's curated router doesn't explicitly own
         // to the app's FULL engine router (hook-event, send, link/ticket ops,
         // resync-jira, get-session, list-worktrees, …). This makes a "missing
@@ -416,7 +427,8 @@ public enum CrowDaemon {
             appState: appState, store: store, git: git, devRoot: options.devRoot,
             cockpit: cockpit, tracker: tracker, allowList: allowList,
             sessionService: sessionService, autoRespond: autoRespond, jobScheduler: jobScheduler,
-            rebuildScorecard: rebuildScorecard, fallback: engineFallback)
+            rebuildScorecard: rebuildScorecard, versionUpdateService: versionUpdateService,
+            fallback: engineFallback)
 
         // Unix socket — lets the existing `crow` CLI talk to the daemon. By
         // default this IS the app's well-known `crow.sock` (the daemon owns it in
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
@@ -392,7 +392,7 @@ public enum CrowDaemon {
         let buildInfo = BuildInfoLoader.load(webDir: options.webDir)
             ?? BuildInfo(version: "?", gitSha: "dev", buildDate: "")
         let versionUpdateService = await MainActor.run {
-            VersionUpdateService(buildInfo: buildInfo, devRoot: options.devRoot)
+            VersionUpdateService(buildInfo: buildInfo)
         }
         startVersionUpdatePoll(
             service: versionUpdateService, devRoot: options.devRoot, eventHub: eventHub)
--- a/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
@@ -1715,7 +1715,7 @@ func makeCommandRouter(
             }
         },
         "version-update-check": { params in
-            let force = params["force"]?.boolValue ?? true
+            let force = params["force"]?.boolValue ?? false
             guard let versionUpdateService else {
                 return ["status": VersionUpdateRPC.statusJSON(nil)]
             }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
@@ -162,6 +162,7 @@ func makeCommandRouter(
     // SessionService and the telemetry receiver are in scope; nil when telemetry
     // is off (there'd be no DB to rebuild from).
     rebuildScorecard: (@MainActor @Sendable () async -> Void)? = nil,
+    versionUpdateService: VersionUpdateService? = nil,
     fallback: CommandRouter? = nil
 ) -> CommandRouter {
     // Serializes review kickoffs (see start-review) — one per router instance.
@@ -1680,6 +1681,55 @@ func makeCommandRouter(
                     "restart_required": .bool(false),
                 ]
             }
+        },
+
+        // Version update check (CROW-938) — `crow version get|set|--check`.
+        "version-update-get": { _ in
+            let config = ConfigStore.loadConfig(devRoot: devRoot) ?? AppConfig()
+            let status = await versionUpdateService?.cachedStatus
+            return [
+                "version_update": VersionUpdateRPC.configJSON(config.versionUpdate),
+                "status": VersionUpdateRPC.statusJSON(status),
+            ]
+        },
+        "version-update-set": { params in
+            try await mapRPCError {
+                let enabled = try SettingsRPC.patchBool(params, "enabled")
+                let intervalHours = try VersionUpdateRPC.patchIntervalHours(params)
+                guard enabled != nil || intervalHours != nil else {
+                    throw RPCError.invalidParams(
+                        "Nothing to set — provide at least one of enabled, interval_hours.")
+                }
+                let versionUpdate = try mutateConfig(devRoot: devRoot) { config -> VersionUpdateConfig in
+                    if let enabled { config.versionUpdate.enabled = enabled }
+                    if let intervalHours {
+                        config.versionUpdate.intervalHours = max(
+                            VersionUpdateConfig.minimumIntervalHours, intervalHours)
+                    }
+                    return config.versionUpdate
+                }
+                return [
+                    "version_update": VersionUpdateRPC.configJSON(versionUpdate),
+                    "saved": .bool(true),
+                ]
+            }
+        },
+        "version-update-check": { params in
+            let force = params["force"]?.boolValue ?? true
+            guard let versionUpdateService else {
+                return ["status": VersionUpdateRPC.statusJSON(nil)]
+            }
+            let config = ConfigStore.loadConfig(devRoot: devRoot) ?? AppConfig()
+            let status: VersionUpdateStatus?
+            if force {
+                status = await versionUpdateService.runCheck()
+            } else {
+                status = await versionUpdateService.checkIfDue(
+                    enabled: config.versionUpdate.enabled,
+                    intervalHours: config.versionUpdate.intervalHours,
+                    force: false)
+            }
+            return ["status": VersionUpdateRPC.statusJSON(status)]
         },
 
         // Settings → Automation for `crow automation` (CROW-812). Same shape and
--- a/Packages/CrowDaemon/Sources/CrowDaemon/RPCLanePolicy.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/RPCLanePolicy.swift
@@ -131,6 +131,10 @@ enum RPCLanePolicy {
         "job-disable": .fixed(.config),
         "job-delete": .fixed(.config),
         "job-duplicate": .fixed(.config),
+        "version-update-set": .fixed(.config),
+        // Already single-flight: a second caller awaits the in-flight compare
+        // instead of spawning another `gh auth token` subprocess + GitHub call.
+        "version-update-check": .concurrent,
 
         // MARK: Job runs
         // Not `.config`: these await a full worktree + tmux spawn, and parking
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -32,6 +32,30 @@ html, body {
   overflow: hidden;
 }
 
+.update-banner {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  gap: 12px;
+  padding: 8px 14px;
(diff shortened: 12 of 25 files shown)


dhilgaertner (changes requested):
## Code & Security Review

Solid, well-scoped feature: branch-head comparison is the right call given Crow ships from source, the `unknown`-over-false-`up_to_date` bias is correct throughout, and the CLI/RPC/parity/docs surfaces are all wired consistently. One CI-breaking gate and two behavioural issues below.

### Critical Issues

**🔴 The lane-policy gate fails — `version-update-set` / `version-update-check` are ledgered writes with no lane.**

`ParityLedger.swift:219-221` registers both as `.write`, but `RPCLanePolicy.rules` was not updated, and `RPCLanePolicyTests.everyWriteMethodIsClassified` is exactly the gate for that (#931 — "lane classification is a decision, not a default"). Reproduced on this branch:

```
✘ Test "every ledgered write declares a lane, and nothing else does"
  ↳ unclassified write RPCs: ["version-update-check", "version-update-set"]
✘ Test run with 10 tests in 1 suite failed
```

`version-update-set` goes through `mutateConfig` like every other settings verb, so it belongs with them:

```swift
"version-update-set": .fixed(.config),
```

`version-update-check` writes no config and wants `.concurrent`, but per the table's own convention that needs a comment saying why — and it is only safe once the single-flight below actually holds (`rebuild-scorecard` carries precisely that justification: *"already single-flight: a second caller awaits the in-flight run"*).

The rest of the suite passes: `VersionUpdateTests` (8/8), `VersionUpdateHandlerTests` (3/3), `RPCLedger` parity, `CrowCLI ParityGate|CLIDocs` (13/13), `scripts/check-cli-parity.sh` (97 methods), `swift build` clean.

### Security Review

**Strengths:**
- 6h floor is enforced in three independent places (`VersionUpdateConfig.init`, the `init(from:)` decoder, `patchIntervalHours`), so a hand-edited `config.json` can't undercut the rate-limit budget.
- No secret exposure: the `gh auth token` value is used only as an `Authorization` header and is never cached, logged, or returned over RPC.
- Fails closed everywhere — malformed compare URL, non-HTTP response, 403+`X-RateLimit-Remaining: 0`, 404, decode failure, and transport errors all land on `.unknown` with a reason rather than a false `up_to_date`.
- Banner and About text are built with `textContent`, not `innerHTML`, so the GitHub-sourced `update_command` / `reason` strings can't inject markup.
- Read-only outbound call to a hardcoded `corveil/crow` endpoint; no user-supplied URL reaches the network.

**Concerns:**

**🟡 `runCheck()`'s single-flight guard doesn't hold, and `version-update-check` is remote-reachable with `force` defaulting to `true`.**

`VersionUpdateService.swift:65-67`:

```swift
if inFlight, let cachedStatus { return cachedStatus }
inFlight = true
defer { inFlight = false }
```

Two problems. On the *first* check `cachedStatus` is `nil`, so a concurrent caller falls straight through and starts a second `gh auth token` subprocess plus a second GitHub request — the guard is inert exactly when it matters. And `defer { inFlight = false }` clears the flag when whichever call finishes first returns, while the other is still in flight. (Once a cache does exist the guard has the opposite failure: it returns a *stale* value to a caller that asked to force a fresh compare.)

That matters because `RPCHandlers.swift:1718` defaults `force` to `true` and `version-update-check` is not listed in `RPCWebSocketHandler.localOnlyDenial`, so it is reachable from remote `/rpc`. Each call spawns a `gh` subprocess and hits GitHub with no throttle at all — the 6h floor guards only the scheduled path. An authenticated remote peer, or just a client stuck in a retry loop, can burn the 60 req/hr unauthenticated budget and spawn processes at request rate.

Suggest holding the in-flight `Task` and awaiting it (`if let task = inFlightTask { return await task.value }`), and adding a short minimum spacing for forced checks so "Check now" can't be leaned on.

### Code Quality

**🟡 `remoteHeadCommit` reads `commits.last`, which is not `main`'s head past 250 commits** — `VersionUpdateRPCSupport.swift:165`.

GitHub's compare endpoint returns at most 250 entries in `commits`. When the local build is further behind than that, `.last` is the 250th-oldest commit rather than upstream head, so:
- `crow version --check` prints a wrong `origin/main is at <sha>` (and its date),
- the web banner's dismissal key is that same `remote_sha` (`app.js:236`), so a dismissal is pinned to a commit that isn't head and re-shows on changes that have nothing to do with new upstream work.

`behind_by` stays correct, so the *state* is right — only the displayed sha/date and the dismissal key are wrong. Worth either capping the claim ("250+ commits behind") or fetching head separately via `GET /repos/{repo}/commits/main`.

**🟢 Nits**

- `VersionUpdateRPCSupport.swift:204` — `var object` is never mutated; make it `let`, or inline it into the `return`.
- `VersionUpdateService.swift:105` — the poll sleeps the *previous* interval before re-reading config, so lowering 168h → 6h doesn't take effect until the current sleep expires. The `Task` also isn't retained, so `while !Task.isCancelled` can never be false.
- `VersionUpdateService.swift:38-43` — `devRoot` is accepted and immediately discarded (`_ = devRoot`). Fine as a documented placeholder, but it's currently dead surface on the initializer.
- `VersionUpdateService.swift:53` — disabling the toggle sets `cachedStatus = nil`, so Settings → About goes blank rather than keeping the last known result. Worth confirming that's intended vs. just stopping future checks.
- `VersionUpdateHandlerTests.checkReturnsCachedStatus` shells out to the real `gh auth token` (visible as the PATH-resolution log line and a 0.286s runtime). `VersionUpdateService` already injects `ShellRunner` — passing a stub would keep the test hermetic.
- `VersionUpdateRPCSupport.swift:55` — `localSha` is interpolated into the compare URL unencoded. Not exploitable today (the value comes from `generate-build-info.sh`, and `URL(string:)` fails closed into `"Invalid compare URL"`), but percent-encoding it would make that a property rather than a coincidence.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 2 Yellow, 6 Green] findings.

---

dhilgaertner (changes requested):
## Code & Security Review

Nice shape overall: injectable transport keeps the parser testable, the web UI uses `textContent` throughout (no XSS surface), the config decodes defensively, and the docs/CHANGELOG/parity ledger work is thorough — `ParityGate`, `CLIDocs`, and `RPCLedger` all pass locally. Two blocking issues below, though: the core comparison is inverted, and a CI gate fails.

### Critical Issues

**🔴 1. `ahead_by` / `behind_by` are inverted — the feature never fires in its main case, and false-fires in the dev case.** `Packages/CrowEngine/Sources/CrowEngine/VersionUpdateRPCSupport.swift:101-128`

For `compare/{base}...{head}`, GitHub reports the relationship of **head relative to base**. This code calls it with `base = localSha`, `head = main`, so `ahead_by` = commits on `main` that the local build lacks = *how far behind the local build is*. `behind_by` is the opposite: commits the local build has that `main` lacks.

Verified against the live API using this repo — `5075fba` is a real commit on `main`, two commits stale:

```
$ gh api "repos/corveil/crow/compare/5075fba...main" \
    --jq '{status, ahead_by, behind_by, total_commits, commits_last: .commits[-1].sha, base_commit_sha: .base_commit.sha}'
{"ahead_by":2,"behind_by":0,"status":"ahead","total_commits":2,
 "commits_last":"f56939d…","base_commit_sha":"5075fba…"}
```

Tracing that payload through `check()`: `behindBy = 0`, `aheadBy = 2` → the guard at line 101 (`aheadBy > 0 && behindBy == 0`) matches first → returns `.unknown` with reason *"Local build is not on upstream main."*

So a genuinely stale build reports `unknown`: the banner never renders, `crow version --check` exits `2`, and Settings → About shows the italic "not on upstream main" text. Conversely, a build made from any feature branch (commits not on `main`) returns `behind_by > 0` → `.behind`, telling the developer to `git pull && make install` when they are in fact ahead. The two states are swapped.

Note the parsing half of the same function already assumes the correct semantics — `remoteHeadCommit` reads the upstream tip out of `commits`, and `commits`/`total_commits` is precisely the `ahead_by` set. So line 93 reads `behind_by` for a distance whose commit list it takes from the ahead side.

Roughly:

```swift
// main has commits we lack → we are behind by `aheadBy`.
if aheadBy > 0 && behindBy == 0 { /* .behind, behindBy: aheadBy */ }
// we have commits main lacks (feature branch / fork) → not comparable.
if behindBy > 0 { /* .unknown, "Local build is not on upstream main" */ }
// identical → .upToDate
```

`status` needs the same treatment: `status == "ahead"` means *you are behind*; `"behind"`/`"diverged"` should map to `unknown`.

The four client tests pass only because the fixtures encode the same inversion — `fixture(status: "behind", behindBy: 2, aheadBy: 0)` (`VersionUpdateTests.swift:32-50`) is a payload the real API will not produce for a stale build, and `aheadOnlyIsUnknown` (line 64) asserts the broken behavior. Please re-fixture both from a real compare response so they'd catch a future re-inversion.

**🔴 2. `RPCLanePolicyTests` fails — two write RPCs ship without a lane.** `Packages/CrowCore/Sources/CrowCore/Parity/ParityLedger.swift:219-221`

The PR adds `.write("version-update-set", …)` and `.write("version-update-check", …)` to the ledger but adds nothing to `RPCLanePolicy.rules`. The gate asserts both directions:

```
✘ Test "every ledgered write declares a lane, and nothing else does"
  recorded an issue at RPCLanePolicyTests.swift:23:9:
  Expectation failed: (writes.subtracting(declared) →
    ["version-update-check", "version-update-set"]).isEmpty → false
  ↳ unclassified write RPCs: ["version-update-check", "version-update-set"]
```

This is a real CI failure, not a flake. The test plan's filter — `--filter 'VersionUpdate|RPCLedger'` — doesn't match `RPCLanePolicyTests`, which is why it went unnoticed.

`version-update-set` genuinely needs `.fixed(.config)`: it goes through `mutateConfig` → `ConfigStore.withConfigLock`, an NSLock held across disk I/O, which is exactly the contention `.config` exists to serialize (#931). `version-update-check` does network I/O and spawns a subprocess, so `.config` would be wrong for it — `.concurrent` is declarable explicitly (as `refresh-tickets` and `rebuild-scorecard` are), which satisfies the gate while keeping the classification a deliberate decision.

### Security Review

**Strengths:**
- No XSS: `app.js:238-240` and `settings.js:889-899` build all update text with `textContent` / `classList`, never `innerHTML`, so the GitHub-sourced `reason`, `remote_sha`, and `update_command` can't inject markup.
- The `gh auth token` value is only ever sent as a `Bearer` header to `api.github.com` (`VersionUpdateRPCSupport.swift:63-65`), is never persisted into the cached status, and never reaches the RPC payload or the UI.
- Failure modes are closed correctly — network errors, non-2xx, rate limits, and `dev` SHAs all resolve to `unknown` rather than a false `up_to_date`.
- The 6h floor is enforced in three independent places (`VersionUpdateConfig.init`, the decoder, and `patchIntervalHours`), so a hand-edited `config.json` can't drop below it.

**Concerns:**

**🟡 3. `force` defaults to `true`, which bypasses the rate-limit floor the whole design rests on.** `Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift:1718`

```swift
let force = params["force"]?.boolValue ?? true
```

Any caller that omits `force` gets a live GitHub call plus a `gh auth token` subprocess spawn. `version-update-check` isn't listed in `RPCWebSocketHandler.localOnlyDenial`, so it's reachable over remote `/rpc`; a loop of `crow version check` or a scripted web client can exhaust the 60 req/hr unauthenticated budget and spawn one subprocess per call — the exact exhaustion the 6h minimum is documented as preventing. Both real callers (`VersionCommands.swift:48` and the "Check now" button in `settings.js:906`) already pass `force: true` explicitly, so defaulting to `false` costs nothing.

Related, same area: the in-flight guard in `VersionUpdateService.runCheck` (`VersionUpdateService.swift:65`) is `if inFlight, let cachedStatus { return cachedStatus }`. Before the first successful check there's no cached status, so concurrent callers all fall through and each issues its own request — which is precisely the daemon-start window.

**🟢 4. The build SHA is interpolated into the request path unvalidated.** `VersionUpdateRPCSupport.swift:54-56`

`localSha` comes from `version.json` and goes straight into the compare URL. It's locally generated by `generate-build-info.sh`, so this is hardening rather than a live bug, but a value containing `../` would re-path a request that carries the user's GitHub token. A hex/length check on `compareSha` before building the URL is a cheap guard.

### Code Quality

**🟡 5. `remote_sha` falls back to the *local* SHA.** `VersionUpdateRPCSupport.swift:168`

```swift
if let head = json["base_commit"] as? [String: Any], let parsed = parseCommitNode(head) {
```

In `compare/{base}...{head}` with `base = localSha`, `base_commit` is the **local** commit, not the upstream head — confirmed in the API output above, where `base_commit_sha` is `5075fba` (the local one). The binding is named `head`, which is what makes it read as correct. Whenever `commits` is empty or unparseable, `remote_sha` reports the local build's own SHA as the remote tip; it only looks right in the `identical` case, where the two coincide. Since `app.js:236` keys banner dismissal on `remote_sha`, a wrong value makes the dismissal stick to the wrong commit.

**🟡 6. `commits.last` isn't the upstream head past 250 commits.** `VersionUpdateRPCSupport.swift:165-167`

GitHub's compare response caps `commits` at 250 entries. A build more than 250 commits behind — plausible for exactly the stale installs this feature targets — yields a mid-history commit as `remote_sha`, so the banner's dismissal key is a commit that was never the tip and the displayed "origin/main is at …" is wrong. `ahead_by` / `total_commits` stay accurate. Resolving the tip via `GET /repos/{repo}/commits/main` (or the `Link: rel="last"` page) would be exact. Worth a comment either way, since the 250 cap isn't obvious at the call site.

**🟢 7. Minor cleanups.**
- `VersionUpdateRPCSupport.swift:204` — `var object` is never mutated; should be `let` (likely an unused-mutation warning).
- `VersionUpdateRPCSupport.swift:35-41` — `base` is built as a full `VersionUpdateStatus` but only `base.checkedAtMs` is read, while three branches re-spell the struct by hand. A plain `let checkedAt = currentTimeMs()` would be clearer and drop the duplication.
- `VersionUpdateService.swift:42` — `_ = devRoot` stores nothing; consider dropping the parameter until the "future smarter clone hint" actually lands.
- `VersionUpdateService.swift:104-105` — the poll sleeps for the *old* interval, so changing 24h → 6h in Settings doesn't take effect until the current sleep expires (up to a day later). Same for `enabled: false`, which only clears `cachedStatus` on the next wake. A shorter tick that re-reads config, or waking on config change, would match the "live within ~1 board poll" behavior the other settings advertise.

### Summary Table

| # | Finding | Severity |
|---|---------|----------|
| 1 | `ahead_by`/`behind_by` inverted — feature never reports `behind` when stale, false-reports on feature branches | 🔴 Red |
| 2 | `RPCLanePolicyTests` fails: `version-update-set` / `version-update-check` declare no lane | 🔴 Red |
| 3 | `force` defaults to `true`, bypassing the 6h rate-limit floor; in-flight guard ineffective pre-cache | 🟡 Yellow |
| 5 | `remote_sha` falls back to `base_commit` = the local SHA | 🟡 Yellow |
| 6 | `commits.last` isn't the head past GitHub's 250-commit cap | 🟡 Yellow |
| 4 | Build SHA interpolated into the token-bearing request path unvalidated | 🟢 Green |
| 7 | `var`→`let`, unused `base`/`devRoot`, interval change not picked up until next wake | 🟢 Green |

**Recommendation:** Request Changes — driven by [2 Red, 3 Yellow, 2 Green] findings.

Finding 1 is the one to fix first — with it in place the feature can't report `behind` on a stale build at all, so the manual test-plan items ("dismiss banner", "Settings → About shows matching status") can't pass as written. A test built from a captured real compare payload would pin it.

---

dhilgaertner (changes requested):
## Code & Security Review

Solid, well-scoped feature. The design decisions are good and the comments explain *why* rather than *what* — the separate `/commits/main` fetch to dodge the compare API's 250-commit cap (`VersionUpdateRPCSupport.swift:173-176`) and the single-flight coalescing modeled on `ScorecardRebuilder` are both the right calls. Verified locally: `swift build --package-path Packages/CrowDaemon` clean, `VersionUpdate` suites green in CrowEngine (9 tests) and CrowDaemon (5 tests), `ParityGate|CLIDocs` green (13 tests), `scripts/check-cli-parity.sh` → 97 methods matched.

Two Yellow findings below; no blockers.

### Critical Issues

None.

### Security Review

**Strengths:**
- API host is a hardcoded constant (`api.github.com`) — the `gh auth token` Bearer header cannot be steered to a third-party host.
- `gh auth token` failure is a soft `try?` → unauthenticated request, never a crash or a hang (`VersionUpdateService.swift:99-103`).
- All web rendering uses `textContent` / `.title` property assignment, never `innerHTML` — no XSS surface from the GitHub-sourced `reason` / `update_command` / `remote_sha` strings.
- `version-update-set` correctly routed to `.fixed(.config)` so it shares the config write lock; `version-update-check` correctly left `.concurrent` given the single-flight.
- Failure modes all collapse to `unknown` with a reason — never a false `up_to_date`. `aheadOnlyIsUnknown` and `rateLimitIsUnknown` pin this.

**Concerns:**

**🟡 Yellow — the forced-check path allows ~120 GitHub req/hr, double the budget the 6h floor exists to protect.**
`VersionUpdate.swift:8-10` justifies `minimumIntervalHours = 6` as: *"so an unauthenticated GitHub client (60 req/hr/IP) cannot be exhausted by a misconfigured interval."* But each check makes **two** requests — the compare (`VersionUpdateRPCSupport.swift:56-66`) plus `fetchBranchHead` (`:176-192`) — and `minimumForcedSpacing = 60` seconds (`VersionUpdateService.swift:36`) permits 60 forced checks/hr. 60 × 2 = 120 req/hr against a 60 req/hr budget.

`version-update-check` is not in `RPCWebSocketHandler.localOnlyDenial`, so an authenticated remote `/rpc` peer can drive this; so can a shell loop of `crow version --check`, which always sends `force: true` (`VersionCommands.swift:48`).

*Failure scenario:* on a host where `gh auth token` returns nothing, a 60-second-cadence forced check exhausts the unauthenticated budget in ~30 minutes. Every subsequent check — including the scheduled one — returns `unknown` ("GitHub rate limit exceeded") for the rest of the hour, and the banner silently stops working.

Cheapest fix: skip `fetchBranchHead` when the compare returns `status == "identical"` — in that case `main`'s head *is* the local SHA, so the second call is pure waste in the common steady state. That halves the cost of the case you'll hit 99% of the time. Raising `minimumForcedSpacing` to 120s would also close it.

### Code Quality

**🟡 Yellow — the update banner un-dismisses itself when `remote_sha` is absent.**
`app.js:244-247` persists the dismissal only `if (remoteSha)`, and `app.js:234-236` re-shows the banner whenever `remoteSha` is empty. `fetchBranchHead` returns `nil` on any non-2xx or transport error (`VersionUpdateRPCSupport.swift:185-190`), independently of the compare call succeeding.

*Failure scenario:* the compare request succeeds but the branch-head request 403s (rate-limit boundary — reachable via the finding above, since the two calls are counted separately). `status` comes back `behind` with `remote_sha: null`. The user clicks ×, `banner.hidden = true`, nothing is written to localStorage — and the next `onServerChanged` broadcast (`app.js:217`, which fires on any session mutation) re-runs `renderVersionUpdateBanner` and sets `hidden = false`. The banner reappears within seconds and cannot be dismissed until GitHub recovers. An in-memory dismissal fallback for the no-SHA case would fix it.

**🟢 Green — stale tooltip on the About status line.** `settings.js:886-903` resets `className` at the top of `renderUpdateStatus` but never clears `updateStatus.title`, which is only ever *set* (`:898`). After Check now flips `behind` → `up_to_date`, the "Update: git -C <clone> pull && make install" tooltip stays attached to the "Up to date with origin/main." text.

**🟢 Green — `checkedAtMs` narrowing.** `VersionUpdateRPCSupport.swift:235` does `.int(Int($0))` on an `Int64`. Latent only (all Crow targets are 64-bit), but it's an unconditional trap rather than a clamp.

**🟢 Green — `gitShaFull` isn't validated before URL interpolation.** `VersionUpdateRPCSupport.swift:55` percent-encodes with `.urlPathAllowed`, which permits `/`. A crafted `version.json` (e.g. under an operator-supplied `--web-dir`, `VersionUpdateService.swift:9-12`) could steer the request to a different `api.github.com` path while carrying the Bearer token. The hardcoded host means the token can't leave GitHub, so this is defense-in-depth only — but a `[0-9a-f]{7,40}` guard is one line and removes the class.

**🟢 Green — extra RPC round-trip per broadcast.** `refreshVersionUpdateBanner()` in `onServerChanged` (`app.js:217`) issues a `version-update-get` on every server change, and the handler does a full `ConfigStore.loadConfig` disk read + `AppConfig` decode on the MainActor, per connected browser. The banner's content can only change once per ≥6 hours. The `cachedStatus` fast path at `app.js:250-254` already exists — folding the status into the `changed` payload would avoid the round-trip entirely.

**🟢 Green — interval-lowering math is imprecise.** `VersionUpdateService.swift:132-139` replaces `remaining` with a *full* fresh interval rather than `newInterval - elapsed`. Lowering 168h → 6h one hour into the sleep schedules the next wake at T+7h, not T+6h. Harmless because `checkIfDue`'s own `lastCheckAt` guard is the real authority, but the comment at `:125-126` promises tighter behavior than the code delivers.

**🟢 Green — the interval picker can misrepresent a CLI-set value.** `crow version set --interval-hours 10` passes validation (`VersionCommands.swift:143-145` only enforces ≥6), but `settings.js:928-930` offers 6/12/24/168 — with 10 stored, no `<option>` matches and the browser displays "6 hours" while config says 10. Same pattern as the existing `cleanup.retentionHours` picker, so this is consistent with precedent rather than a regression.

### Test Coverage

Good coverage of the interesting decision points — the 250-commit-cap regression is pinned by name (`branchHeadComesFromCommitsEndpointNotComparePayload`), and `devShaIsIndeterminate` / `aheadOnlyIsUnknown` / `rateLimitIsUnknown` cover the never-false-positive contract.

**🟢 Green — two daemon tests don't test what they're named for.** `forcedChecksCoalesce` and `checkReturnsCachedStatus` (`VersionUpdateHandlerTests.swift:70-106`) both use `gitSha: "dev"`, which short-circuits at `BuildInfo.isIndeterminate` before any transport call. Both assertions hold trivially whether or not coalescing works — a counting stub transport would make `forcedChecksCoalesce` meaningful. Also uncovered: the 404 ("Local SHA is not on GitHub") and generic-HTTP-error branches, and the `fetchBranchHead`-fails-but-compare-succeeds combination that produces the banner bug above.

### Docs & Parity

Complete and consistent: `CLAUDE.md`, `Resources/CLAUDE.md.template`, `docs/cli.md` (generated), `docs/cli-reference.md` (hand-written), the `ParityLedger` method + field rows, and the `CHANGELOG` entry all landed together. `check-cli-parity.sh` and the `CLIDocs` suite confirm it.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 7 Green] findings. Both Yellows are small, localized fixes; the Greens are yours to take or leave.

---

dhilgaertner (changes requested):
## Code & Security Review

Round 4. The compare-semantics fix from the last round is **correct** — I verified the `base...head` orientation against GitHub's documented fields and the new tests pin it precisely: with `base = localSha, head = main`, `ahead_by` is how far the local build is behind and `behind_by` is what the local build carries that `main` lacks. `staleBuildReportsBehind`, `localAheadOfMainIsUnknown`, and `identicalIsUpToDateAndSkipsBranchHeadFetch` lock all three arms down, and `branchHeadComesFromCommitsEndpointNotComparePayload` covers the >250-commit cap that made the old inferred-sha approach wrong.

Verified locally: `swift build --package-path Packages/CrowDaemon` clean; `VersionUpdate` (CrowEngine, 10 tests), `VersionUpdate|RPCLedger` (CrowDaemon, 7 tests), `ParityGate|CLIDocs` (CrowCLI, 13 tests) all pass; `scripts/check-cli-parity.sh` reports 97 methods in sync.

### Critical Issues

None.

### Security Review

**Strengths:**
- `gh auth token` is invoked in argv form (`shell.run("gh", "auth", "token")`) — no shell string, no injection surface. The token is only ever set as a `Bearer` header and is never logged, cached to disk, or echoed back in `VersionUpdateStatus`.
- Request host and path template are compile-time literals; only the SHA is interpolated, and it is percent-encoded (`VersionUpdateRPCSupport.swift:55`). Nothing user-supplied reaches the URL.
- Failure modes are uniformly silent and non-fatal: transport errors, non-2xx, 403+`X-RateLimit-Remaining: 0`, and undecodable bodies all land on `.unknown` with a reason. There is no path that reports a false `up_to_date` — including `dev`/empty SHAs, which short-circuit via `isIndeterminate` before any network call.
- Web surfaces use `textContent`/`title` exclusively — no `innerHTML`, so no XSS even though the banner string is assembled from RPC fields. Banner state in `localStorage` is a bare SHA, nothing sensitive.
- Rate-limit defence is layered properly: the 6h floor is enforced in `init`, in `init(from:)`, in the RPC patch validator, in the CLI's `validate()`, and again in the poll loop — a hand-edited `config.json` can't get under it.

**Concerns:**
- None blocking. See the forced-check budget note under Code Quality.

### Code Quality

**🟡 Yellow — the update banner overflows the viewport and clips the bottom of the app**

`index.html:13` inserts `#update-banner` as a normal-flow sibling *before* `#app`, but `#app` is viewport-sized rather than flex-sized:

```css
html, body { height: 100%; overflow: hidden; }   /* app.css:25-33 */
#app { height: 100vh; height: 100dvh; }          /* app.css:59-64 */
```

When the banner is visible, body content becomes `bannerHeight + 100dvh`. Because `body` has `overflow: hidden`, the excess cannot be scrolled to — the bottom ~35px of `#app` (8px padding ×2 + line box + 1px border) is clipped and unreachable, taking the terminal's last rows with it. This fires exactly on the feature's payload path: the only time the banner renders is the only time the layout breaks.

Secondary symptom on mobile: `#back-to-sidebar` is `position: fixed; top: calc(12px + env(safe-area-inset-top))` (`app.css:477-482`). Fixed positioning is viewport-anchored and ignores the banner's flow box, so the ☰ button renders on top of the banner text.

Either make the pair a flex column (`body { display: flex; flex-direction: column }` + `#app { flex: 1; min-height: 0 }` in place of `100dvh`), or make the banner `position: fixed` and add a compensating top offset on `#app` while it is shown.

**🟢 Green — `identical` short-circuit reports the build date as the remote commit date**

`VersionUpdateRPCSupport.swift:100` fills the skipped branch-head fetch with `RemoteCommit(sha: build.gitSha, date: build.buildDate)`. The SHA is right, but `buildDate` is when the binary was compiled, not when the commit landed — so an up-to-date `crow version --check` prints `origin/main is at abc1234 (<build date>)` and labels it a commit date. Leaving `date: nil` there costs nothing (the CLI and Settings both guard on empty) and avoids stating something untrue.

**🟢 Green — stale tooltip on the Settings → About status line**

`renderUpdateStatus` (`settings.js:887-889`) resets `textContent` and `className` on every render but not `title`. The `behind` arm sets `updateStatus.title = 'Update: ' + …`; after a "Check now" that flips to `up_to_date`, the stale "Update: …" tooltip is still attached. One line alongside the className reset: `updateStatus.title = '';`.

**🟢 Green — forced-check spacing sits exactly on the unauthenticated budget**

`minimumForcedSpacing = 120` (`VersionUpdateService.swift:36`) permits 30 forced checks/hr, and each non-identical check costs **two** GitHub requests (compare + branch head) — 60 req/hr, precisely GitHub's unauthenticated per-IP cap, before the scheduled check spends anything. A `/rpc` peer leaning on "Check now" can therefore starve the periodic check into `unknown`. Self-healing and silent, and `gh auth token` lifts the ceiling to 5000/hr, so this is a tuning note rather than a defect — but the constant was presumably chosen against a one-request-per-check model, and the second call landed after it.

**🟢 Green — `forcedChecksCoalesce` doesn't exercise the path it names**

The test seeds `gitSha: "dev"`, which returns from `VersionUpdateClient.check` at the `isIndeterminate` guard before any transport call. Both results are then trivially equal whether or not single-flight worked, so the `inFlight` coalescing is effectively unasserted. `VersionUpdateService` hardcodes `VersionUpdateClient.check` with the default `URLSession.shared` transport and has no seam — threading through the same injectable transport `VersionUpdateClient` already exposes would let this test (and the 120s throttle, also uncovered) assert something real. The client-level tests are genuinely thorough, so this is the one soft spot in otherwise good coverage.

### Notes (no action needed)

- `startVersionUpdatePoll` discarding its `Task` handle matches `startJobPoll` (`CrowDaemon.swift:907`) — consistent with the codebase, and the loop is daemon-lifetime by design.
- The hourly re-read in the sleep loop compares the *full* refreshed interval against `remaining` rather than the elapsed-adjusted remainder, so lowering the interval mid-sleep is slightly conservative (waits elapsed + new interval). Harmless given the 6h floor.
- `version-update-check` honours `force` even when `versionUpdate.enabled` is false. That reads as intentional — the opt-out governs the schedule, and an explicit `crow version --check` is a user action — and the docs describe it that way.
- Single-flight in `runCheck` is race-free: the `if let inFlight` guard and the `inFlight = task` assignment are both on `MainActor` with no suspension between them, so no second task can be created while one is live.

### Summary Table

The Yellow is the only thing standing between this and an approve, and it is a contained CSS fix. The Swift side — compare semantics, rate-limit floors, failure handling, single-flight, parity ledger, docs — is in good shape.

---

dhilgaertner (changes requested):
## Code & Security Review

Verified locally: `swift build` (CrowDaemon, CrowCLI) clean; `swift test --filter VersionUpdate` (CrowEngine 10/10, CrowDaemon 17/17 incl. `RPCLedger`/`RPCLanePolicy`), `--filter 'ParityGate|CLIDocs'` (13/13), and `scripts/check-cli-parity.sh` (97 methods) all pass.

The compare direction is correct, and worth stating explicitly since it's the easiest thing to get backwards: with `compare/{local}...main`, GitHub's `ahead_by` is how far `main` has moved past the local build and `behind_by` is local-only commits, and `VersionUpdateRPCSupport.swift:108-136` maps both the right way round (feature branch / fork → `unknown`, never a false `up_to_date`). Docs, CHANGELOG, `CLAUDE.md`, the template, and the parity ledger are all updated in lockstep.

### Security Review

**Strengths:**
- The `gh auth token` value is confined to `Authorization: Bearer` on two hardcoded `api.github.com` URLs; it is never logged, cached, or echoed into the RPC payload (`VersionUpdateService.swift:99-103`, `VersionUpdateRPCSupport.swift:151-161`). `crow version get` returns config + status only.
- Both web surfaces write via `textContent` / `el(...)`, so GitHub-derived strings (`remote_sha`, `remote_date`) can't inject markup (`app.js:59-62`, `settings.js:884-901`).
- Interval is floored at 6h in three independent places — `VersionUpdateConfig.init`, `init(from:)`, and `patchIntervalHours` — so the `set-config` path can't smuggle a smaller value past `crow version set`.
- Failure modes are closed correctly: no HTTP response, 404, rate limit, decode failure, and transport errors all land on `unknown` rather than `up_to_date`.

**Concerns:**
- No injection or credential-exposure issue found. One hardening note under Green below.

### Yellow — should fix

**1. `app.css` — the `100dvh` mobile fix is silently dropped.** `#app` lost `height: 100vh; height: 100dvh; /* exclude mobile browser chrome so the last row is reachable */` and became `flex: 1; min-height: 0` inside the new `body { display: flex; flex-direction: column }`. But `html, body` is still `height: 100%` (`app.css:25-32`), and `100%` resolves against the *layout* viewport — i.e. `100vh` semantics, the large viewport on mobile. `dvh` no longer appears anywhere in `app.css`. So the bottom row is back under mobile browser chrome, which is exactly the bug the deleted comment documented. This is a whole-app regression, not banner-scoped. Give `body` the dynamic height instead:

```css
html, body { height: 100%; }
body { height: 100dvh; display: flex; flex-direction: column; }
```

**2. `VersionCommands.swift:47-59` — documented exit codes are unreachable in the common failure case.** `docs/cli-reference.md` promises exit `0` up to date, `1` behind, `2` could not check. When `crowd` isn't running, `rpc()` throws `SocketError.connectionFailed` straight out of `run()`, and ArgumentParser exits **1** — which the docs define as "behind". A script guard like `crow version --check || notify "update available"` fires a false update alert every time the daemon is down. An RPC-level error is worse: `rpc()` wraps it in `ValidationError`, which exits **64**. The null-status branch already has the right behaviour; the throw just needs to reach it, and `HookEventCommand.swift:79` shows the repo's own convention for catching this:

```swift
let result: [String: JSONValue]
do {
    result = try rpc("version-update-check", params: ["force": .bool(true)])
} catch {
    warn("Could not check for updates — is crowd running?")
    throw ExitCode(2)
}
```

**3. The GitHub-quota guards have no test coverage.** `minimumForcedSpacing` is introduced specifically as a defense — its own comment says "so a remote `/rpc` peer can't burn the GitHub budget" (`VersionUpdateService.swift:35-36`) — and neither it nor the single-flight `inFlight` path is exercised. `forcedChecksCoalesce` (`VersionUpdateHandlerTests.swift:96-106`) looks like it covers the latter, but it builds the service with `gitSha: "dev"`, which short-circuits in `check()` at the `isIndeterminate` guard before any transport work; the two calls need never overlap, so `results[0] == results[1]` would hold with the single-flight removed entirely. `VersionUpdateClient.check` already takes an injectable `transport`, but `VersionUpdateService` never forwards one, so the service can't be driven off the network. Threading a `transport` stored property through the initializer would make both guards testable with a counting stub (assert one transport call for N concurrent forced checks; assert the second forced check within 120s returns cache without a transport hit).

### Green — consider

- **`VersionUpdateRPCSupport.swift:102-104` — the branch-head fetch is spent on results that discard it.** `fetchBranchHead` runs for every non-`identical` compare, including the `behind`/`diverged` case that returns `.unknown` two lines later. Every developer on a feature branch therefore burns two API calls per interval for a status the UI renders as `reason` text only. Moving the fetch below the `.unknown` early-return halves the quota cost for that (common) population.
- **`VersionUpdateRPCSupport.swift:55` — validate the SHA before path interpolation.** `.urlPathAllowed` leaves `/` unescaped, so a malformed `gitShaFull` could reshape the API path. Not exploitable in practice (the host is fixed, so the bearer token can't be redirected off `api.github.com`, and the value comes from a locally generated `version.json`), but a `compareSha.allSatisfy(\.isHexDigit)` guard is one line and would let `isIndeterminate` reject garbage rather than only `""`/`dev`.
- **`app.js:218` — `refreshVersionUpdateBanner()` on every `changed` push.** Each nudge costs an RPC plus a daemon-side `ConfigStore.loadConfig` disk read, and `syncUpdateBannerLayout` reads `banner.offsetHeight`, forcing a synchronous layout. All for state that moves at most every 6 hours. The boot call plus a slow fallback poll would cover it.
- **`app.js:36-39` — `--update-banner-height` is measured once per render and never on resize**, so `#back-to-sidebar` keeps a stale offset if the banner text rewraps to two lines.
- **`app.js:47-56` — dismissal isn't durable when the head fetch fails.** If `fetchBranchHead` returns nil while the state is still `behind`, `remote_sha` is null and dismissal falls back to the in-memory `versionBannerDismissedWithoutSha`, so the banner returns on every page reload until a head fetch succeeds.
- **`VersionUpdateService.swift:53-55` — `replaceBuildInfo(_:)` is dead code**; nothing in the tree calls it.
- **`RPCHandlers.swift:1725` — `force: true` bypasses `versionUpdate.enabled`.** Defensible (the doc comment scopes the opt-out to "on a schedule", and both callers are explicit user actions), but Settings → About leaves "Check now" fully enabled with no hint that it still reaches the network after opt-out. Worth a word in the toggle's help text.
- **`VersionUpdateRPCSupport.swift:245` — `.int(Int($0))` on an `Int64` epoch traps on a 32-bit platform.** Moot on macOS/Linux arm64/x86_64; noting only because it's a silent trap rather than a clamp.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 3 Yellow, 7 Green] findings. The feature itself is well-shaped: the compare direction is right, the failure modes all close to `unknown`, and the credential handling is clean. The blockers are the dropped `100dvh` (an app-wide mobile regression riding along with a banner layout change) and the exit-code contract that `crow version --check` documents but doesn't honour when `crowd` is down.

---

dhilgaertner (approved):
## Code & Security Review

Round 6. Every blocker from the previous rounds is resolved, and I re-verified the two that were easiest to regress:

- **Compare orientation is correct.** With `compare/{local}...main`, GitHub reports head-relative-to-base, so `ahead_by` is how far `main` has moved past the local build and `behind_by` is local-only commits. `VersionUpdateRPCSupport.swift:108-136` maps both the right way round and swaps them into local-relative terms for the payload (`behindBy: aheadBy`). `diverged` is caught by the first guard, so a fork can never read as `behind`.
- **The `100dvh` mobile fix is back.** `app.css:35-39` restores it on `body` with the original comment, and `#app` fills the remainder via `flex: 1; min-height: 0` — the banner is a sibling flex item, so the last row stays reachable.
- **The lane-policy gate passes.** `version-update-set` → `.fixed(.config)`, `version-update-check` → `.concurrent` with the single-flight justification the table's convention requires.

### Critical Issues

None.

### Verification

| Gate | Result |
| --- | --- |
| `swift build --package-path Packages/CrowDaemon` | clean |
| `CrowEngine` (full) | 641/641 |
| `CrowCore` (full) | 544/544 |
| `CrowCLI` (full, incl. `ParityGate`/`CLIDocs`) | 342/342 |
| `CrowDaemon --filter 'VersionUpdate\|RPCLedger\|RPCLanePolicy'` | 19/19 |
| `scripts/check-cli-parity.sh` | 97 methods in sync |
| `shellcheck scripts/generate-build-info.sh` | clean |

**Pre-existing failures, not from this PR:** the full `CrowDaemon` suite reports 2 failures — `WebTerminalAssetTests.refreshTerminalsRebindsTheActiveTerminalToTheFreshRow` and `WebNotificationCenterTests.bootCatchResetsTheHistory`. I reproduced both on a clean `origin/main` worktree at `f56939d` (this PR's merge base), and the `app.js` regions they scrape are byte-identical between `main` and this branch. Worth a separate issue — `refreshTerminals` no longer contains `activeTerminal = terminals.find(` at all, so that drift guard is currently asserting against code that has moved.

### Security Review

**Strengths:**
- The `gh auth token` value is fetched in argv form (`shell.run("gh", "auth", "token")` — no shell string), used only as an `Authorization: Bearer` header on two compile-time-literal `api.github.com` URLs, and never logged, cached to disk, or echoed into `VersionUpdateStatus`. `crow version get` returns config + status only.
- Failure modes are uniformly closed: transport error, non-2xx, 404, `403` + `X-RateLimit-Remaining: 0`, and undecodable bodies all land on `.unknown` with a reason. `dev`/empty SHAs short-circuit via `isIndeterminate` before any network call. There is no path that yields a false `up_to_date`, and `staleBuildReportsBehind` / `localAheadOfMainIsUnknown` / `rateLimitIsUnknown` pin it.
- Both web surfaces render via `textContent` / `.title` / `el(...)` — never `innerHTML` — so GitHub-derived `remote_sha`, `remote_date`, and `reason` carry no XSS surface. The `localStorage` dismiss key is a bare SHA.
- The 6h floor is enforced in four independent places (`VersionUpdateConfig.init`, `init(from:)`, `patchIntervalHours`, and the `mutateConfig` clamp in the handler), so neither a hand-edited `config.json` nor the `set-config` path can undercut it.
- Default-on network egress is properly disclosed and opt-out: Settings → About toggle, `crow version set --enabled false`, CHANGELOG, `CLAUDE.md`, the template, and both docs. The only data sent is the build SHA and `User-Agent: Crow/<version>`, to a host Crow already talks to constantly.
- `version.json` is generated, not committed, so a fresh clone falls back to `gitSha: "dev"` → `unknown` with no network call. The safe default is the un-stamped one.

**Concerns:** none blocking.

### Green — consider

**1. `.urlPathAllowed` is the wrong character set for a single path segment.** `VersionUpdateRPCSupport.swift:55` percent-encodes the SHA with `.urlPathAllowed`, which permits `/` and `.` — so a `gitShaFull` containing a slash would extend the path rather than be escaped. Not exploitable in practice (the value comes from `git rev-parse HEAD` at build time, the host is pinned, the request is GET-only, and anyone who can write `version.json` in the served web dir already controls `app.js`), but a `CharacterSet.alphanumerics` check — or just rejecting a non-hex SHA in `isIndeterminate` — would make the guarantee structural rather than circumstantial.

**2. The forced-check budget now sits exactly at 100% of the unauthenticated quota.** `minimumForcedSpacing = 120` permits 30 forced checks/hr, and each check is up to 2 requests (compare + `fetchBranchHead`) — precisely the 60 req/hr/IP that `VersionUpdate.swift:8-10` cites as the thing to protect. Better than the previous round's 2×, and the steady up-to-date case is 1 request thanks to the `identical` short-circuit, but it leaves zero headroom for `gh` or anything else sharing the IP. 180s would restore some margin.

**3. No jsdom coverage for the banner dismiss state machine.** `app.js:220-290` adds a non-trivial bit of state — the `localStorage` key on `remote_sha` plus the `versionBannerDismissedWithoutSha` session fallback for when `fetchBranchHead` fails. The repo has a jsdom harness built for exactly this kind of extraction test, and the PR's own test plan leaves "dismiss banner, confirm it stays dismissed until remote SHA changes" as a manual unchecked box. The Swift side is thoroughly covered; this is the one gap.

**4. `--update-banner-height` is measured once and never re-measured.** `syncUpdateBannerLayout` reads `banner.offsetHeight` at render time only. On a phone rotation that rewraps the banner text to two lines, the cached value goes stale and `#back-to-sidebar` (`app.css:505-507`) sits at the wrong offset until the next render. A `ResizeObserver` on the banner would close it; cosmetic and mobile-only.

### Code Quality

Genuinely good. The comments explain *why* rather than *what* — the separate `/commits/main` fetch to dodge the compare API's 250-commit cap (`:183-185`), the `.concurrent` lane justification, and the hourly wake in the poll loop so a lowered interval takes effect without waiting out a long prior sleep. The injectable `transport` keeps the parser testable without a network, `staleBuildReportsBehind` is pinned against a real captured API response, and `identicalIsUpToDateAndSkipsBranchHeadFetch` proves the short-circuit by making the second call throw. Docs, CHANGELOG, `CLAUDE.md`, the template, the generated CLI reference, and the parity ledger all move in lockstep.

Two small things I checked and am satisfied with: `throw ExitCode(0)` is the correct ArgumentParser idiom and exits silently through `CrowCommand.main()`, and forcing a check while `enabled: false` is right — the toggle governs the *schedule*, and "Check now" / `--check` are explicit user consent.

### Summary Table

| Color  | Meaning      | Count | Verdict effect  |
|--------|--------------|-------|-----------------|
| Red    | Must fix     | 0     | Request changes |
| Yellow | Should fix   | 0     | Request changes |
| Green  | Consider     | 4     | Approve allowed |

---

dhilgaertner (changes requested):
## Code & Security Review

Round 6. All three round-5 Yellows are **fixed and verified**:

1. **`100dvh` restored** — `body { height: 100dvh; display: flex; flex-direction: column }` (`app.css:35-39`) with `#app { flex: 1; min-height: 0 }`. I checked the blast radius of making `body` a flex container: its other children (`#session-scrim`, `#statusbar`, `#lightbox`) are all `position: fixed`, so they're out of flow and unaffected, and `<script>` generates no box. The change is correctly scoped.
2. **Exit codes** — `VersionCommands.swift:48-58` now wraps `rpc()` in `do/catch` and throws `ExitCode(2)`, so a down `crowd` no longer reports as "behind".
3. **Quota coverage is real now** — `CountingTransport` with a 50ms delay makes `forcedChecksCoalesceOntoOneTransportRound` assert `requestCount == 2` (one compare + one head, not two of each), and `forcedChecksThrottleWithinMinimumSpacing` pins the 120s guard. Both round-4 Greens also landed (`date: nil` on the `identical` short-circuit, `updateStatus.title = ''`).

Verified locally: `swift build --package-path Packages/CrowDaemon` clean; CrowEngine `VersionUpdate` 10/10, CrowDaemon `VersionUpdate|RPCLedger|RPCLanePolicy` 19/19, CrowCLI `ParityGate|CLIDocs` 13/13, `scripts/check-cli-parity.sh` 97 methods in sync.

One new defect, which is why this isn't an approve.

### Critical Issues

**🔴 Red — `.update-banner` overrides the `hidden` attribute, so the banner is permanently visible and cannot be dismissed**

`index.html:13` ships the banner with the `hidden` attribute, and `app.js` toggles visibility exclusively through `banner.hidden = true/false` (lines 243, 250, 255, 263, 271, 286). But `app.css:41-51` sets `display` on it:

```css
.update-banner {
  display: flex;
  ...
}
```

The `hidden` attribute is implemented as a **UA-stylesheet** rule (`[hidden] { display: none }`). Author-origin normal declarations beat UA-origin ones in the cascade regardless of specificity, so `.update-banner { display: flex }` wins and `hidden` becomes inert. Consequences on the default path:

- **Every page load, for every user**, renders a ~34px empty gold-tinted bar (8px padding ×2 + the 18px `×` glyph + 1px border) above `#app` — including users who are up to date, on a `dev` build, or offline, since `renderVersionUpdateBanner` hides via `banner.hidden = true` in all four of those arms.
- **The dismiss button does nothing visually.** `dismiss.onclick` writes the SHA to `localStorage` and sets `banner.hidden = true`; the bar stays on screen. The PR description's "dismissible banner" is inoperative.
- The mobile symptom from round 4 returns in a new form: `syncUpdateBannerLayout` correctly *removes* `body.update-banner-visible` when hidden, so `--update-banner-height` is dropped and `#back-to-sidebar` (`position: fixed; top: calc(12px + …)`, `app.css:485`) sits back at 12px — now on top of a bar that is still occupying space. The offset compensation and the actual rendering disagree precisely because the hide didn't take.

This codebase already knows about this cascade interaction and works around it twice, in this same file, for the same reason:

```css
#session-scrim[hidden] { display: none; }   /* app.css:1023 */
#lightbox[hidden]      { display: none; }   /* app.css:1055 */
```

`#lightbox` is the only other `hidden`-toggled element in `app.js` (lines 2460-2467), and it has its companion rule. `.update-banner` is the one that's missing it. One line next to the block at `app.css:41` closes it:

```css
.update-banner[hidden] { display: none; }
```

Worth noting the PR's own unchecked manual item — *"dismiss banner, confirm it stays dismissed until remote SHA changes"* — is exactly the check that surfaces this. The three checked automated items can't: there is no test over the web assets, so the entire banner path is unexercised by CI.

*(Verified by code inspection and the in-repo precedent. I attempted a live render to confirm, but the browser extension couldn't reach a localhost server from this environment, so this rests on the cascade rule and the two existing workarounds rather than a screenshot.)*

### Security Review

**Strengths:**
- No change to the credential path since round 5, and it remains clean: `gh auth token` is invoked in argv form (`shell.run("gh", "auth", "token")`) — no shell string, no injection surface. The token is only ever a `Bearer` header on two compile-time-literal `api.github.com` URLs, and never reaches disk, logs, or `VersionUpdateStatus`.
- `version-update-get`/`-check` return config + status only; `crow version get` cannot leak the token.
- Both web surfaces write through `textContent` / `el(...)`, so GitHub-derived `remote_sha` / `remote_date` / `reason` can't inject markup. The `localStorage` dismissal key holds a bare short SHA.
- Every failure mode closes to `unknown`, never a false `up_to_date`: transport error, non-2xx, 404, 403 + `X-RateLimit-Remaining: 0`, undecodable body, and `dev`/empty SHAs (short-circuited at `isIndeterminate` before any network call).
- The 6h floor is enforced in five independent places — `VersionUpdateConfig.init`, `init(from:)`, `VersionUpdateRPC.patchIntervalHours`, `VersionSet.validate()`, and the poll loop — so neither a hand-edited `config.json` nor the `set-config` path can get under it.

**Concerns:**
- None. The Red above is a rendering defect, not a security issue.

### Code Quality

Green — carried forward from round 5, all still open and all still non-blocking:

- **`VersionUpdateRPCSupport.swift:101-104` — the branch-head fetch is spent on results that discard it.** `fetchBranchHead` runs for every non-`identical` compare, including the `behind`/`diverged` arm that returns `.unknown` four lines later. Moving it below that early-return halves the quota cost for everyone on a feature branch.
- **`VersionUpdateRPCSupport.swift:55` — validate the SHA before path interpolation.** `.urlPathAllowed` leaves `/` unescaped. Not exploitable (fixed host, locally generated `version.json`), but `compareSha.allSatisfy(\.isHexDigit)` folded into `isIndeterminate` would reject garbage rather than only `""`/`dev`.
- **`app.js:217` — `refreshVersionUpdateBanner()` on every `changed` push**, costing an RPC plus a daemon-side `ConfigStore.loadConfig` disk read plus a forced layout (`banner.offsetHeight`) for state that moves at most every 6 hours.
- **`app.js:37-39` — `--update-banner-height` is measured once per render, never on resize**, so the offset goes stale if the banner text rewraps.
- **`app.js:47-56` — dismissal isn't durable when the head fetch fails** (`remote_sha` null → in-memory `versionBannerDismissedWithoutSha` only).
- **`VersionUpdateService.swift:62-64` — `replaceBuildInfo(_:)` is dead code.**
- **`RPCHandlers.swift:1726` — `force: true` bypasses `versionUpdate.enabled`.** Defensible and documented, but Settings → About leaves "Check now" fully enabled with no hint that it still reaches the network after opt-out.
- **`VersionUpdateRPCSupport.swift:245` — `.int(Int($0))` on an `Int64` epoch traps rather than clamps on a 32-bit platform.** Moot on the supported targets.

New this round, both minor:

- **🟢 `CountingTransport` is duplicated** — a nested copy in `VersionUpdateServiceTests.swift:14-42` and a file-private one in `VersionUpdateHandlerTests.swift:118-140`, with divergent stub payloads (`ahead` vs `identical`). Fine as-is, but a shared test helper would keep the two from drifting.
- **🟢 `generate-build-info.sh` ignores a dirty working tree.** `GIT_FULL_SHA` is `rev-parse HEAD`, so a binary built with uncommitted changes stamps the last commit and can legitimately report `up_to_date` while running modified code. Inherent to SHA-based comparison rather than a defect in this change; a `git diff --quiet || GIT_FULL_SHA="dev"` guard would make it honest if you care.

### Notes (no action needed)

- Single-flight in `runCheck` remains race-free: `if let inFlight` and `inFlight = task` are both `@MainActor` with no suspension between them, so the child task can't start before the assignment lands.
- The parity ledger, `RPCLanePolicy`, `docs/cli.md`, `docs/cli-reference.md`, `CHANGELOG.md`, `CLAUDE.md`, and `Resources/CLAUDE.md.template` are all updated in lockstep — `check-cli-parity.sh` and `CLIDocs` both confirm.
- `AppConfig.versionUpdate` decodes via `decodeIfPresent` with a default, so existing `config.json` files upgrade cleanly, and the Settings modal's `set-config` round-trip stays floored by `VersionUpdateConfig.init(from:)`.

### Summary Table

**Recommendation:** Request Changes — driven by [1 Red, 0 Yellow, 10 Green] findings.

The Swift side is done: compare semantics, rate-limit floors, failure handling, single-flight, credential hygiene, parity, and docs are all in good shape, and the round-5 fixes are correct. The one blocker is a single missing CSS line that the repo already writes twice elsewhere for the same reason.

---

PR #943, opened 2026-08-05
fix(web): hide update banner when [hidden] is set

### LINKED ISSUE #942: Update banner renders as an empty bar when up to date — .update-banner CSS overrides [hidden]
## Problem

The update banner is visible on every page load as an **empty** bar, even when the build is up to date.

Reproduced on `main` @ `0fbce3d` with the daemon reporting no update available:

```
$ crow version --check
crow 0.2.0 (0fbce3d, built 2026-08-05)
origin/main is at 0fbce3d
Up to date.
```

The banner still renders — an empty gold-tinted strip above `#app` with a `×` on the right and no text.

## Root cause

`.update-banner` sets `display: flex`, which overrides the `hidden` attribute.

`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css:41`

```css
.update-banner {
  display: flex;
  ...
}
```

`hidden` only hides an element via the **user-agent** stylesheet's `[hidden] { display: none }`. An author-origin `display` declaration beats it regardless of specificity, so every `banner.hidden = true` in `app.js` is a no-op and the element is permanently visible.

`index.html:13` ships the banner with `hidden` set:

```html
<div id="update-banner" class="update-banner" hidden role="status">
  <span class="update-banner-text"></span>
  ...
```

…so it's visible from first paint, before any RPC has run.

The bar is *empty* because `renderVersionUpdateBanner` (`app.js:235`) early-returns on every non-`behind` state **without** touching `.update-banner-text`:

```js
if (!status || status.state !== 'behind') {
  banner.hidden = true;      // no-op — CSS wins
  syncUpdateBannerLayout(banner);
  return;                    // .update-banner-text left empty
}
```

`textContent` is only assigned in the `behind` branch at `app.js:260`.

### Secondary defect: the dismiss button is inert

`dismiss.onclick` is bound only inside the `behind` branch (`app.js:265`). On the early-return paths the `×` is rendered but has no handler, so the empty banner **cannot be dismissed** — clicking does nothing.

### Layout is not affected

`syncUpdateBannerLayout` correctly removes `update-banner-visible` from `<body>` and clears `--update-banner-height` when `banner.hidden` is true, so no space is reserved. That's why the bar overlays/pushes rather than sitting in a reserved slot.

## Fix

Add the `[hidden]` guard, matching the convention this file already uses twice for elements that set their own `display`:

```css
.update-banner[hidden] { display: none; }
```

Precedent in the same stylesheet:

- `app.css:1023` — `#session-scrim[hidden] { display: none; }`
- `app.css:1055` — `#lightbox[hidden] { display: none; }`

`.update-banner` was simply missed.

Worth doing alongside:

- Bind `dismiss.onclick` unconditionally (or clear `textContent` on the early-return paths) so a stale/empty banner can never be both visible and inert.
- Grep the web assets for any other class that sets `display` on an element also driven by `hidden` — this is the third instance of the same trap in one stylesheet, so a lint rule or a shared `.is-hidden` utility may be worth more than a third one-off.

## Acceptance criteria

- [ ] With `crow version --check` reporting **Up to date**, no banner is visible — on first paint and after the check completes.
- [ ] With the build behind, the banner appears **with** its text and a working dismiss button.
- [ ] Dismissing hides it; it does not reappear until the remote SHA changes (existing `crow.updateBannerDismissedSha` behavior is unchanged).
- [ ] `version-update-get` failing or returning `unknown` leaves no visible banner.
- [ ] No empty banner during the window between page load and the first `version-update-get` response.

## Related

Introduced by #941 (CROW-938).

dgershman:
Closes #942

## Summary

- Add `.update-banner[hidden] { display: none; }` so the flex layout no longer overrides the `hidden` attribute (matching `#session-scrim` and `#lightbox`).
- Bind the dismiss handler unconditionally and clear banner text on all hide paths so a stale/empty bar is never both visible and inert.

## Test plan

- [ ] With `crow version --check` reporting **Up to date**, no banner is visible on first paint or after the check completes.
- [ ] With the build behind, the banner appears with text and a working dismiss button.
- [ ] Dismissing hides the banner until the remote SHA changes.
- [ ] `version-update-get` failing or returning `unknown` leaves no visible banner.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -49,6 +49,7 @@ body {
   color: var(--gold);
   font-size: 13px;
 }
+.update-banner[hidden] { display: none; }
 .update-banner-text { flex: 1 1 auto; }
 .update-banner-dismiss {
   flex: 0 0 auto;
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -220,6 +220,15 @@ function onServerChanged() {
 
 const VERSION_BANNER_DISMISS_KEY = 'crow.updateBannerDismissedSha';
 let versionBannerDismissedWithoutSha = false;
+let versionBannerDismissedSha = '';
+
+function readDismissedSha() {
+  try {
+    return localStorage.getItem(VERSION_BANNER_DISMISS_KEY) || '';
+  } catch (_) {
+    return versionBannerDismissedSha;
+  }
+}
 
 function syncUpdateBannerLayout(banner) {
   if (!banner || banner.hidden) {
@@ -252,10 +261,17 @@ function renderVersionUpdateBanner(status) {
     try {
       if (dismissRemoteSha) {
         localStorage.setItem(VERSION_BANNER_DISMISS_KEY, dismissRemoteSha);
+        versionBannerDismissedSha = dismissRemoteSha;
       } else if (dismissArmed) {
         versionBannerDismissedWithoutSha = true;
       }
-    } catch (_) { /* session-only when storage is unavailable */ }
+    } catch (_) {
+      if (dismissRemoteSha) {
+        versionBannerDismissedSha = dismissRemoteSha;
+      } else if (dismissArmed) {
+        versionBannerDismissedWithoutSha = true;
+      }
+    }
   };
 
   if (!status || status.state !== 'behind') {
@@ -264,7 +280,7 @@ function renderVersionUpdateBanner(status) {
   }
   const remoteSha = status.remote_sha || '';
   if (remoteSha) versionBannerDismissedWithoutSha = false;
-  if (remoteSha && localStorage.getItem(VERSION_BANNER_DISMISS_KEY) === remoteSha) {
+  if (remoteSha && readDismissedSha() === remoteSha) {
     hideVersionUpdateBanner(banner, text);
     return;
   }
@@ -276,9 +292,10 @@ function renderVersionUpdateBanner(status) {
   dismissRemoteSha = remoteSha;
   dismissArmed = !remoteSha;
   banner.hidden = false;
-  text.textContent = 'A newer Crow build is available — '
+  const msg = 'A newer Crow build is available — '
     + n + ' commit' + (n === 1 ? '' : 's') + ' behind origin/main.'
     + (status.update_command ? (' Update: ' + status.update_command) : '');
+  if (text.textContent !== msg) text.textContent = msg;
   syncUpdateBannerLayout(banner);
 }
 
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -222,11 +222,13 @@ const VERSION_BANNER_DISMISS_KEY = 'crow.updateBannerDismissedSha';
 let versionBannerDismissedWithoutSha = false;
 let versionBannerDismissedSha = '';
 
-function readDismissedSha() {
+function isDismissedSha(sha) {
+  if (!sha) return false;
+  if (versionBannerDismissedSha === sha) return true;
   try {
-    return localStorage.getItem(VERSION_BANNER_DISMISS_KEY) || versionBannerDismissedSha;
+    return localStorage.getItem(VERSION_BANNER_DISMISS_KEY) === sha;
   } catch (_) {
-    return versionBannerDismissedSha;
+    return false;
   }
 }
 
@@ -275,7 +277,7 @@ function renderVersionUpdateBanner(status) {
   }
   const remoteSha = status.remote_sha || '';
   if (remoteSha) versionBannerDismissedWithoutSha = false;
-  if (remoteSha && readDismissedSha() === remoteSha) {
+  if (remoteSha && isDismissedSha(remoteSha)) {
     hideVersionUpdateBanner(banner, text);
     return;
   }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -232,6 +232,12 @@ function syncUpdateBannerLayout(banner) {
     '--update-banner-height', banner.offsetHeight + 'px');
 }
 
+function hideVersionUpdateBanner(banner, text) {
+  if (text) text.textContent = '';
+  banner.hidden = true;
+  syncUpdateBannerLayout(banner);
+}
+
 function renderVersionUpdateBanner(status) {
   const banner = document.getElementById('update-banner');
   if (!banner) return;
@@ -239,38 +245,39 @@ function renderVersionUpdateBanner(status) {
   const dismiss = banner.querySelector('.update-banner-dismiss');
   if (!text || !dismiss) return;
 
+  let dismissRemoteSha = '';
+  let dismissWithoutSha = false;
+  dismiss.onclick = () => {
+    if (dismissRemoteSha) {
+      localStorage.setItem(VERSION_BANNER_DISMISS_KEY, dismissRemoteSha);
+    } else if (dismissWithoutSha) {
+      versionBannerDismissedWithoutSha = true;
+    }
+    hideVersionUpdateBanner(banner, text);
+  };
+
   if (!status || status.state !== 'behind') {
-    banner.hidden = true;
-    syncUpdateBannerLayout(banner);
+    hideVersionUpdateBanner(banner, text);
     return;
   }
   const remoteSha = status.remote_sha || '';
   if (remoteSha) versionBannerDismissedWithoutSha = false;
   if (remoteSha && localStorage.getItem(VERSION_BANNER_DISMISS_KEY) === remoteSha) {
-    banner.hidden = true;
-    syncUpdateBannerLayout(banner);
+    hideVersionUpdateBanner(banner, text);
     return;
   }
   if (!remoteSha && versionBannerDismissedWithoutSha) {
-    banner.hidden = true;
-    syncUpdateBannerLayout(banner);
+    hideVersionUpdateBanner(banner, text);
     return;
   }
   const n = status.behind_by || 0;
   text.textContent = 'A newer Crow build is available — '
     + n + ' commit' + (n === 1 ? '' : 's') + ' behind origin/main.'
     + (status.update_command ? (' Update: ' + status.update_command) : '');
+  dismissRemoteSha = remoteSha;
+  dismissWithoutSha = !remoteSha;
   banner.hidden = false;
   syncUpdateBannerLayout(banner);
-  dismiss.onclick = () => {
-    if (remoteSha) {
-      localStorage.setItem(VERSION_BANNER_DISMISS_KEY, remoteSha);
-    } else {
-      versionBannerDismissedWithoutSha = true;
-    }
-    banner.hidden = true;
-    syncUpdateBannerLayout(banner);
-  };
 }
 
 async function refreshVersionUpdateBanner(cachedStatus) {
@@ -283,7 +290,9 @@ async function refreshVersionUpdateBanner(cachedStatus) {
     renderVersionUpdateBanner(res && res.status);
   } catch (_) {
     const banner = document.getElementById('update-banner');
-    if (banner) banner.hidden = true;
+    if (banner) {
+      hideVersionUpdateBanner(banner, banner.querySelector('.update-banner-text'));
+    }
   }
 }
 window.refreshVersionUpdateBanner = refreshVersionUpdateBanner;
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -246,14 +246,16 @@ function renderVersionUpdateBanner(status) {
   if (!text || !dismiss) return;
 
   let dismissRemoteSha = '';
-  let dismissWithoutSha = false;
+  let dismissArmed = false;
   dismiss.onclick = () => {
-    if (dismissRemoteSha) {
-      localStorage.setItem(VERSION_BANNER_DISMISS_KEY, dismissRemoteSha);
-    } else if (dismissWithoutSha) {
-      versionBannerDismissedWithoutSha = true;
-    }
     hideVersionUpdateBanner(banner, text);
+    try {
+      if (dismissRemoteSha) {
+        localStorage.setItem(VERSION_BANNER_DISMISS_KEY, dismissRemoteSha);
+      } else if (dismissArmed) {
+        versionBannerDismissedWithoutSha = true;
+      }
+    } catch (_) { /* session-only when storage is unavailable */ }
   };
 
   if (!status || status.state !== 'behind') {
@@ -271,12 +273,12 @@ function renderVersionUpdateBanner(status) {
     return;
   }
   const n = status.behind_by || 0;
+  dismissRemoteSha = remoteSha;
+  dismissArmed = !remoteSha;
+  banner.hidden = false;
   text.textContent = 'A newer Crow build is available — '
     + n + ' commit' + (n === 1 ? '' : 's') + ' behind origin/main.'
     + (status.update_command ? (' Update: ' + status.update_command) : '');
-  dismissRemoteSha = remoteSha;
-  dismissWithoutSha = !remoteSha;
-  banner.hidden = false;
   syncUpdateBannerLayout(banner);
 }
 
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -224,7 +224,7 @@ let versionBannerDismissedSha = '';
 
 function readDismissedSha() {
   try {
-    return localStorage.getItem(VERSION_BANNER_DISMISS_KEY) || '';
+    return localStorage.getItem(VERSION_BANNER_DISMISS_KEY) || versionBannerDismissedSha;
   } catch (_) {
     return versionBannerDismissedSha;
   }
@@ -242,6 +242,7 @@ function syncUpdateBannerLayout(banner) {
 }
 
 function hideVersionUpdateBanner(banner, text) {
+  if (!banner) return;
   if (text) text.textContent = '';
   banner.hidden = true;
   syncUpdateBannerLayout(banner);
@@ -255,22 +256,16 @@ function renderVersionUpdateBanner(status) {
   if (!text || !dismiss) return;
 
   let dismissRemoteSha = '';
-  let dismissArmed = false;
+  let dismissArmedWithoutSha = false;
   dismiss.onclick = () => {
     hideVersionUpdateBanner(banner, text);
-    try {
-      if (dismissRemoteSha) {
+    if (dismissRemoteSha) {
+      versionBannerDismissedSha = dismissRemoteSha;
+      try {
         localStorage.setItem(VERSION_BANNER_DISMISS_KEY, dismissRemoteSha);
-        versionBannerDismissedSha = dismissRemoteSha;
-      } else if (dismissArmed) {
-        versionBannerDismissedWithoutSha = true;
-      }
-    } catch (_) {
-      if (dismissRemoteSha) {
-        versionBannerDismissedSha = dismissRemoteSha;
-      } else if (dismissArmed) {
-        versionBannerDismissedWithoutSha = true;
-      }
+      } catch (_) {}
+    } else if (dismissArmedWithoutSha) {
+      versionBannerDismissedWithoutSha = true;
     }
   };
 
@@ -290,7 +285,7 @@ function renderVersionUpdateBanner(status) {
   }
   const n = status.behind_by || 0;
   dismissRemoteSha = remoteSha;
-  dismissArmed = !remoteSha;
+  dismissArmedWithoutSha = !remoteSha;
   banner.hidden = false;
   const msg = 'A newer Crow build is available — '
     + n + ' commit' + (n === 1 ? '' : 's') + ' behind origin/main.'
--- a/Packages/CrowDaemon/web-tests/README.md
+++ b/Packages/CrowDaemon/web-tests/README.md
@@ -109,6 +109,16 @@ way `index.html` does, so `openSettings` / `setSettingsTab` are the shipped
 implementations rather than stubs — the tab-routing bug they guard (a re-entry
 that silently reset `dirty`) is invisible to a stub.
 
+## `version-banner.test.js` — version-update banner (#942)
+
+Drives `renderVersionUpdateBanner` / `refreshVersionUpdateBanner` against
+mock `localStorage` and `rpc`, with the real `app.css` injected so assertions
+use computed style (not just the `hidden` property). Coverage: `[hidden]`
+genuinely hides the flex banner, behind/up-to-date toggling, SHA dismiss
+surviving a poll, a new SHA re-showing, the `setItem`-throws path with stale
+storage, no-SHA session dismiss resetting when a SHA arrives, `getItem` throws
+not blocking render or in-memory dismiss, and RPC failure hiding the banner.
+
 ## Run
 
 ```sh
--- a/Packages/CrowDaemon/web-tests/package.json
+++ b/Packages/CrowDaemon/web-tests/package.json
@@ -4,8 +4,8 @@
   "private": true,
   "description": "Headless jsdom regression tests for the web UI (Ticket Board; terminal touch- and wheel-scroll; terminal scrollback re-sync; terminal key handling; sidebar session rows; sidebar grouping/collapse; sidebar select toggle; JSON-RPC deadlines and late responses; hash URL routing). Drives the real app.js from Resources/web against mocks.",
   "scripts": {
-    "test": "fail=0; for f in board touch-scroll wheel-scroll scrollback-hydrate key-handler row sidebar-groups sidebar-select rpc-timeout router; do node \"$f.test.js\" || fail=1; done; exit $fail",
-    "test:ci": "fail=0; for f in board touch-scroll wheel-scroll scrollback-hydrate key-handler sidebar-groups sidebar-select rpc-timeout router; do node \"$f.test.js\" || fail=1; done; exit $fail"
+    "test": "fail=0; for f in board touch-scroll wheel-scroll scrollback-hydrate key-handler row sidebar-groups sidebar-select rpc-timeout router version-banner; do node \"$f.test.js\" || fail=1; done; exit $fail",
+    "test:ci": "fail=0; for f in board touch-scroll wheel-scroll scrollback-hydrate key-handler sidebar-groups sidebar-select rpc-timeout router version-banner; do node \"$f.test.js\" || fail=1; done; exit $fail"
   },
   "devDependencies": {
     "jsdom": "^24.0.0"
--- a/Packages/CrowDaemon/web-tests/package.json
+++ b/Packages/CrowDaemon/web-tests/package.json
@@ -2,7 +2,7 @@
   "name": "crow-web-tests",
   "version": "1.0.0",
   "private": true,
-  "description": "Headless jsdom regression tests for the web UI (Ticket Board; terminal touch- and wheel-scroll; terminal scrollback re-sync; terminal key handling; sidebar session rows; sidebar grouping/collapse; sidebar select toggle; JSON-RPC deadlines and late responses; hash URL routing). Drives the real app.js from Resources/web against mocks.",
+  "description": "Headless jsdom regression tests for the web UI (Ticket Board; terminal touch- and wheel-scroll; terminal scrollback re-sync; terminal key handling; sidebar session rows; sidebar grouping/collapse; sidebar select toggle; JSON-RPC deadlines and late responses; hash URL routing; version-update banner). Drives the real app.js from Resources/web against mocks.",
   "scripts": {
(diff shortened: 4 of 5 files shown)


dhilgaertner (changes requested):
## Code & Security Review

The diagnosis is right and the CSS fix is the correct, idiomatic one. Two follow-ups below — one of them defeats the invariant this PR sets out to establish.

### Critical Issues

None.

### Security Review

**Strengths:**
- All daemon-supplied strings (`status.update_command`, `status.behind_by`) go through `textContent`, never `innerHTML` — `update_command` is operator-controlled config that lands in the DOM, and this keeps it inert. Unchanged by the PR, but worth stating since the PR moves that line.
- `dismiss.onclick =` (property assignment) rather than `addEventListener` is the right call for the new unconditional binding. `renderVersionUpdateBanner` runs on every `changed` nudge (`app.js:217`), so `addEventListener` here would have accumulated one handler per poll. Property assignment makes rebinding idempotent.
- `localStorage` holds only a git SHA under a namespaced key (`crow.updateBannerDismissedSha`). No secrets, no PII.
- No new network calls, no auth/authz surface, no injection vectors introduced.

**Concerns:**

None security-relevant.

### Code Quality

**🟡 `localStorage.setItem` throwing leaves the banner permanently undismissable — `app.js:250-257`**

```js
dismiss.onclick = () => {
  if (dismissRemoteSha) {
    localStorage.setItem(VERSION_BANNER_DISMISS_KEY, dismissRemoteSha);  // ← can throw
  } else if (dismissWithoutSha) {
    versionBannerDismissedWithoutSha = true;
  }
  hideVersionUpdateBanner(banner, text);  // ← never reached if it does
};
```

`localStorage.setItem` throws `SecurityError` when site data is blocked and `QuotaExceededError` when storage is full. Crow's web UI is typically reached over LAN at a bare IP, which is exactly the origin shape a "block all cookies and site data" browser setting refuses storage for. When it throws, the exception escapes the click handler and `hideVersionUpdateBanner` never runs — the user clicks × and the bar stays put, forever, across every click.

This failure mode predates the PR, but the PR's stated contract is "a stale/empty bar is never both visible and inert," and this is the one remaining path to precisely that state. It's also inside lines this PR already rewrites. Hiding first makes the dismissal work even when persistence doesn't:

```js
dismiss.onclick = () => {
  hideVersionUpdateBanner(banner, text);
  try {
    if (dismissRemoteSha) {
      localStorage.setItem(VERSION_BANNER_DISMISS_KEY, dismissRemoteSha);
    } else if (dismissWithoutSha) {
      versionBannerDismissedWithoutSha = true;
    }
  } catch (_) { /* dismissal is session-only when storage is unavailable */ }
};
```

Worth checking the read side too — `localStorage.getItem` at `app.js:265` throws in the same environments, though there it's already covered by the `try` in `refreshVersionUpdateBanner`.

**🟢 The `role="status"` announcement may now be dropped — `app.js:274-280`**

The text is written while the banner is still `hidden`, then the banner is unhidden:

```js
text.textContent = 'A newer Crow build is available — ' + …;
dismissRemoteSha = remoteSha;
dismissWithoutSha = !remoteSha;
banner.hidden = false;
```

Before this PR the region was never actually `display: none` — that's the bug being fixed — so every mutation happened inside the accessibility tree and got announced. Now the mutation lands while the region is outside the tree, and whether the content is announced on re-show is AT-dependent. Swapping to `banner.hidden = false` → set `textContent` → `syncUpdateBannerLayout(banner)` makes the polite announcement reliable and costs nothing. (`syncUpdateBannerLayout` reads `offsetHeight`, so it must stay last either way.)

**🟢 `dismissWithoutSha` is derivable — `app.js:248-249, 277-278`**

The two flags encode one bit: whenever the banner is shown, `dismissWithoutSha === !dismissRemoteSha`, and both are falsy on every hide path. A single `dismissArmed` boolean set alongside `dismissRemoteSha` at line 277 would carry the same meaning with less to keep in sync. Non-blocking.

### What I verified

- `.update-banner[hidden]` has specificity (0,2,0) vs `.update-banner`'s (0,1,0), so it wins on specificity rather than source order — robust against future reordering.
- It matches the existing idiom exactly: `#session-scrim[hidden]` (`app.css:1024`) and `#lightbox[hidden]` (`app.css:1056`). Those plus `#update-banner` are the **only** three elements in `index.html` carrying the `hidden` attribute, so this completes the set — no other element in the app has the same latent bug.
- `app.css` is a render-blocking `<link>` in `<head>` (`index.html:9`), so the fix applies at first paint, which is the case the issue reports.
- The new closure captures `dismissRemoteSha`/`dismissWithoutSha` as `let` bindings mutated later in the same invocation — correct, since closures capture variables rather than values. Each render replaces the handler wholesale, so there's no stale-sha window.
- `node --check app.js` passes; `swift build --package-path Packages/CrowDaemon` succeeds (the lone warning is pre-existing, in `RPCWebSocketHandler.swift`, untouched here).
- No JS test harness exists in the repo (no `package.json`, no lint config), so the manual test plan is the right level of verification for this change — not raising missing tests as a finding.

### Summary Table

The Yellow is a one-line reorder. Fold in the a11y ordering while you're in there and this is good to go.

---

dhilgaertner (changes requested):
## Code & Security Review

The root-cause diagnosis is right and the fix is correct: `.update-banner[hidden]` is specificity `(0,2,0)` against `.update-banner`'s `(0,1,0)`, there is no `!important` on the base rule, and placement matches the existing precedent at `app.css:1024` / `app.css:1056`. Hoisting the `dismiss.onclick` binding above the early returns and clearing `textContent` on every hide path closes the "visible but inert" secondary defect cleanly.

I verified the fix is **exhaustive**, which the issue explicitly asked for: `update-banner`, `session-scrim`, and `lightbox` are the only three elements driven by the `hidden` attribute anywhere in the web assets (no `setAttribute('hidden')` / `toggleAttribute('hidden')` paths exist), and all three now carry a `[hidden]` guard. There is no fourth instance of the trap.

All five acceptance criteria on #942 are met by this diff.

### Critical Issues

None.

### Security Review

**Strengths:**
- Banner copy is written with `textContent`, never `innerHTML` (`app.js:279`). `status.update_command` and `status.behind_by` arrive from the daemon and are inserted as text, so there is no XSS sink even if the RPC payload were attacker-influenced.
- The `localStorage` value is a git SHA used only for an equality comparison (`app.js:267`) — never interpolated into markup, a URL, or a command.
- No new network calls, no auth/authz surface, no secrets, no new configuration. The CSS change is a static rule with no injection surface.

**Concerns:** none.

### Code Quality

**🟡 Yellow — storage-unavailable handling is half-done, and the comment documents behavior the code doesn't have** (`app.js:252-258`, `app.js:267`)

Commit 525715c added a `try`/`catch` around the dismiss write, annotated:

```js
} catch (_) { /* session-only when storage is unavailable */ }
```

Two problems with that:

1. **The catch is only reachable in the arm where the fallback does not happen.** The `else if` arm is a plain assignment that cannot throw, so `setItem` is the only statement that can land in the catch — and in that case nothing is recorded at all: not `localStorage`, not `versionBannerDismissedWithoutSha`. The dismissal is forgotten immediately, and `renderVersionUpdateBanner` re-runs on every `onServerChanged` nudge (`app.js:217`, fired on each board poll and session edit), so the banner reappears within seconds. "Session-only" is precisely what does *not* occur.

   Note that simply setting `versionBannerDismissedWithoutSha` in the catch would not fix it either — `app.js:266` clears that flag whenever `remoteSha` is truthy, and the guard at `app.js:271` requires `!remoteSha`. It needs a separate in-memory sha.

2. **The read three lines below is still unguarded.** `localStorage.getItem` at `app.js:267` is outside any `try`. In the same browser configurations that make `setItem` throw (site data blocked in Chrome/Firefox, sandboxed iframe without `allow-same-origin`), `getItem` throws first — `renderVersionUpdateBanner` unwinds into `refreshVersionUpdateBanner`'s catch (`app.js:293`), which hides the banner. Net effect: those users are **never told their build is behind**, silently.

To be clear on regression status: the unguarded `getItem` is pre-existing and unchanged, and the write path is strictly *better* than `main` (where the uncaught throw meant `banner.hidden = true` never ran at all). I'm flagging it because this PR deliberately opened the storage-hardening door and left it half-open, with a comment asserting it's closed. Finishing it is a few lines:

```js
let versionBannerDismissedSha = '';           // module scope, mirrors the stored sha

function readDismissedSha() {
  try { return localStorage.getItem(VERSION_BANNER_DISMISS_KEY) || ''; }
  catch (_) { return versionBannerDismissedSha; }
}
```

then `app.js:267` becomes `if (remoteSha && readDismissedSha() === remoteSha)`, and the catch becomes `catch (_) { versionBannerDismissedSha = dismissRemoteSha; }` — which delivers exactly the session-only fallback the comment promises. Alternatively, drop the comment and state that dismissal is not persisted when storage is unavailable.

**🟢 Green — `role="status"` re-announcement is now idempotent-unsafe** (`app.js:278-281`)

Swapping the order so `banner.hidden = false` precedes the `textContent` write is a genuine improvement: the mutation now lands while the live region is rendered, so assistive tech actually announces it (a mutation made while the region is hidden generally isn't).

The flip side is that `renderVersionUpdateBanner` re-runs on every `onServerChanged` nudge and unconditionally reassigns the *same* string. Replacing a text node with identical content is still a live-region mutation, and several AT/browser pairs re-announce it — so a screen-reader user may hear the notice repeatedly for as long as the banner is up. A cheap guard:

```js
const msg = 'A newer Crow build is available — ' + …;
if (text.textContent !== msg) text.textContent = msg;
```

This behavior also existed on `main`, so it isn't a regression — but the reorder is what makes the region's live-ness meaningful, so it's worth handling here.

**🟢 Green — shared hidden utility deferred**

The issue floated a `.is-hidden` utility or a lint rule over a third one-off. Skipping that is the right call at three call sites, all now covered — but it's worth a note somewhere that any future `display`-setting class toggled via `hidden` needs the guard, since nothing enforces it.

**Tests:** none added, correctly — the repo ships no JS lint config or JS test harness, and the Swift suites (`VersionUpdateTests`, `VersionUpdateServiceTests`, `VersionUpdateHandlerTests`) cover the daemon side, which is untouched. `node --check` on `app.js` passes. The changed files are verbatim-copied bundle resources, so there is no build impact.

### Summary Table

The `[hidden]` fix itself is correct, complete, and ready. The Yellow is the inaccurate `catch` comment plus the unguarded read beside it; both are small and land naturally in this round rather than as a follow-up.

---

dhilgaertner (changes requested):
## Code & Security Review

The CSS fix is correct and the JS hardening is thoughtful. One real gap in the storage-unavailable fallback keeps this from an approve.

### Critical Issues

None.

### Security Review

**Strengths:**
- All banner content goes through `textContent`, never `innerHTML` — `status.update_command` and `behind_by` come from the daemon and are inserted as text, so there's no injection path (`app.js:295-298`).
- The only persisted value is a commit SHA under a constant key; it is compared, never evaluated or reflected into the DOM.
- No auth, network, or privilege surface is touched. `localStorage` failures are contained rather than allowed to break the render path.

**Concerns:**
- None.

### Code Quality

**🟡 Yellow — `readDismissedSha()` can't reach its own in-memory fallback (`app.js:225-231`)**

The fallback only fires when the *read* throws. The realistic failure mode is the opposite: `getItem` works, `setItem` throws (quota exceeded, or Safari private browsing). Trace it:

1. Banner shows, user dismisses → `hideVersionUpdateBanner` hides it.
2. `localStorage.setItem(...)` throws, so the `versionBannerDismissedSha = dismissRemoteSha` on the *next* line never runs (`app.js:263-264`).
3. `catch` records `versionBannerDismissedSha = 'abc123'` (`app.js:270`). ✅
4. Next render: `readDismissedSha()` calls `getItem`, which **succeeds** and returns `null` → `|| ''` → `''`.
5. `'' === 'abc123'` is false → banner reappears.

Because `refreshVersionUpdateBanner()` runs from `onServerChanged` (`app.js:217`), that's the next board poll — seconds later. So the dismiss visibly fails to stick, which is the same "banner is on screen when it shouldn't be" class this PR exists to fix, and `689dc7c` ("complete storage-unavailable handling") is the commit that meant to close it.

One-token fix:

```js
return localStorage.getItem(VERSION_BANNER_DISMISS_KEY) || versionBannerDismissedSha;
```

**🟢 Green — collapse the duplicated try/catch fallback (`app.js:259-275`)**

Both arms assign the same in-memory state; the `catch` is the `try` minus `setItem`. Setting the in-memory state *first* and then attempting the write removes the duplication and makes the fallback load-bearing alongside the fix above:

```js
dismiss.onclick = () => {
  hideVersionUpdateBanner(banner, text);
  if (dismissRemoteSha) {
    versionBannerDismissedSha = dismissRemoteSha;
    try { localStorage.setItem(VERSION_BANNER_DISMISS_KEY, dismissRemoteSha); } catch (_) {}
  } else if (dismissArmed) {
    versionBannerDismissedWithoutSha = true;
  }
};
```

Minor naming note while you're in here: `dismissArmed` reads as "the dismiss button is live," but it's false in the (also dismissable) remote-SHA case. Something like `dismissArmedWithoutSha` says what it gates.

**🟢 Green — asymmetric null guard in `hideVersionUpdateBanner` (`app.js:244-248`)**

It guards `text` but dereferences `banner` unguarded. Not a live crash — `renderVersionUpdateBanner` early-returns on `!banner` and the `catch` site checks `if (banner)` — but the mismatched guards invite a future unguarded call site. An `if (!banner) return;` costs nothing and matches how `syncUpdateBannerLayout` already handles it.

### What's Right Here

- **The CSS fix is correct and complete.** `.update-banner[hidden]` has specificity (0,2,0) against `.update-banner`'s (0,1,0) and is declared after it, so it wins cleanly. It also matches the established `#session-scrim[hidden]` / `#lightbox[hidden]` convention (`app.css:1024`, `app.css:1056`) — and since those three are the *only* elements in `index.html` using the `hidden` attribute, this closes the bug class rather than one instance of it.
- **`if (text.textContent !== msg)` is a genuine accessibility win** (`app.js:298`). The banner is `role="status"`, so the old unconditional assignment re-announced the same message to screen readers on every board poll. Worth keeping.
- **Clearing content on hide mirrors the existing lightbox idiom** (`app.js:2493` does `img.src = ''`), so the pattern is consistent with the file.
- **Binding `dismiss.onclick` unconditionally plus the `dismissArmed` gate is the right call** — it prevents a click on a hidden/inert banner from poisoning `versionBannerDismissedWithoutSha` and suppressing a future legitimate no-SHA banner.

### Testing

No automated coverage was added, but the daemon's web assets have no JS test harness (`crow-desktop/package.json` exposes only a `tauri` script), so this matches the repo's status quo rather than being a gap this PR introduced. Verified locally:

- `swift build` — **Build complete!** (resources bundle cleanly)
- `swift test --package-path Packages/CrowEngine --filter VersionUpdate` — **10 tests, 2 suites, all passed**
- `node --check app.js` — syntax OK

The manual test plan in the description covers the right paths; please add the quota-exceeded case (dismiss with `localStorage.setItem` stubbed to throw, then confirm the banner stays hidden across a poll) once the Yellow above is addressed.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The Yellow is a one-line change; everything else is optional polish.

---

dhilgaertner (changes requested):
## Code & Security Review

Verified behaviourally: I ran the real `app.js` under the repo's own jsdom harness (`Packages/CrowDaemon/web-tests`) with the real `app.css` injected, driving `renderVersionUpdateBanner` and asserting on **computed style** rather than the `hidden` property. The core fix works. Two things should be tightened before merge.

### Critical Issues

None. No Red findings.

### Security Review

**Strengths:**
- No XSS surface introduced. Every write goes through `textContent` (`app.js:293`, `app.js:246`); there is no `innerHTML`, no `insertAdjacentHTML`, no template-string DOM construction anywhere in the changed region.
- The only externally-sourced values — `status.update_command` and `status.remote_sha`, both from the local daemon's `version-update-get` — are rendered as text and compared for equality respectively. `remote_sha` reaches `localStorage.setItem` as a *value* under a constant key (`app.js:221`), never as a key, selector, or URL fragment, so a hostile sha string is inert.
- `localStorage` access is fully guarded (`app.js:226-231`, `app.js:264-266`). A `SecurityError` from a partitioned/blocked storage context can no longer take down the render path — this was worth doing and is done correctly.

**Concerns:** none.

### Code Quality

#### 🟡 1. The in-memory dismiss fallback is shadowed by any stale `localStorage` value

`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js:225-231`

```js
return localStorage.getItem(VERSION_BANNER_DISMISS_KEY) || versionBannerDismissedSha;
```

`versionBannerDismissedSha` is consulted only when `getItem` returns **falsy**. But `getItem` returns falsy only when the key was *never successfully written* — and any user who has dismissed the banner once before has a value there. So the fallback that commit 08d989a ("fall back to in-memory dismissed sha when setItem fails") exists to provide is bypassed in the most likely instance of the very scenario it targets.

Failure scenario, reproduced:

1. Dismiss at sha `aaa` while storage is writable → `localStorage = 'aaa'`, memory = `'aaa'`. ✅
2. Storage becomes read-only (quota exhausted by another feature, private-mode switch). Remote advances to `bbb`; banner appears; user dismisses → memory = `'bbb'`, `setItem` throws and is swallowed, storage still holds `'aaa'`.
3. Next poll re-renders the same status: `readDismissedSha()` returns `'aaa'` (truthy, so memory is never reached) ≠ `'bbb'` → **the banner comes back, and comes back on every poll for the rest of the session.**

My harness run, all other assertions passing:

```
SAME, but storage already holds an older sha from a prior dismiss
  ✓ aaa persisted
  ✓ banner shown for bbb
  ✓ dismiss hid it
  ✓ memory holds bbb
  ✓ storage still holds stale aaa
  ✗ STAYS hidden on re-render        <-- the claim under test
```

A dismiss that un-dismisses itself is a worse end state than the pre-PR behaviour, where the throw at least propagated. Rather than reordering the `||` (which just moves the staleness question to a second tab that dismissed a different sha), make it a membership test so either store hiding is sufficient — order-independent, no "which is fresher" question:

```js
function isDismissedSha(sha) {
  if (!sha) return false;
  if (versionBannerDismissedSha === sha) return true;
  try {
    return localStorage.getItem(VERSION_BANNER_DISMISS_KEY) === sha;
  } catch (_) {
    return false;
  }
}
```

and at `app.js:278`:

```js
if (remoteSha && isDismissedSha(remoteSha)) {
```

#### 🟡 2. No regression test, for the fourth consecutive fix to this banner

This branch is four commits deep on one component (`a3ca699`, `525715c`, `689dc7c`, `08d989a`), each fixing a defect the previous one missed. That is the exact churn signature a regression test ends.

The harness for it already exists and **already gates every PR** — `.github/workflows/ci.yml:153-157` runs `npm run test:ci`, and `board.test.js` / `router.test.js` drive the real `app.js` under jsdom against mocks. The banner shipped in #941 with Swift-side coverage only (`VersionUpdateHandlerTests.swift`), which is why a pure-CSS specificity bug reached users.

The original bug is cheap to pin, because jsdom computes styles from an injected `<style>` — asserting on `getComputedStyle(banner).display` rather than `banner.hidden` is what distinguishes this fix from the bug:

```js
const T = load(makeStorage());               // real app.css in a <style> tag
T.render({ state: 'behind', behind_by: 3, remote_sha: 'aaa' });
ok('shown when behind', T.visible());
T.render({ state: 'up_to_date' });
ok('hidden again', !T.visible());            // fails on main, passes here
```

Four cases cover the whole surface: `[hidden]` genuinely hides, dismiss survives a re-render, a *new* sha re-shows, and the `setItem`-throws path. One note if you add it — `test:ci` enumerates its files explicitly in `web-tests/package.json`, so a new `version-banner.test.js` must be added to that list or CI will silently never run it.

#### 🟢 3. Things that are right, for the record

- **The CSS fix is correct.** `.update-banner[hidden]` is specificity (0,2,0) against `.update-banner`'s (0,1,0), and no `#update-banner` rule exists to outrank it. It matches the established pattern at `app.css:1024` / `app.css:1056`. Confirmed via computed style, not by reading the selector.
- **`hideVersionUpdateBanner` is a genuine consolidation**, not just deduplication: all four hide paths now necessarily call `syncUpdateBannerLayout`, so `body.update-banner-visible` and `--update-banner-height` (which `#back-to-sidebar` positions against, `app.css:506-507`) can no longer be left stale by a path that forgot the call. The `catch` in `refreshVersionUpdateBanner` (`app.js:305-310`) was previously one such path.
- **Reordering `banner.hidden = false` ahead of the text write is safe** — it is all one synchronous task, so no paint intervenes and there is no empty-bar flash.
- **`if (text.textContent !== msg)` (`app.js:293`) earns its keep** beyond avoiding a redundant write: reassigning `textContent` destroys the text node, which would drop a user's selection mid-copy of the update command on every poll.
- **Unconditional handler binding is sound.** Each render creates fresh `dismissRemoteSha` / `dismissArmedWithoutSha` bindings and replaces `onclick`, so no handler outlives its sha. I verified a click on the hidden banner writes nothing and does not poison a later show.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 1 Green] findings.

The CSS fix at the heart of this PR is correct and I'd take it as-is. Finding 1 is a one-function change; finding 2 is what stops a fifth commit on this banner.

---

dhilgaertner (changes requested):
## Code & Security Review

The core fix is correct and I verified it empirically. `app.css:52` (`.update-banner[hidden] { display: none; }`) has specificity `(0,2,0)` vs the `.update-banner` flex rule's `(0,1,0)`, so it wins cleanly, and it matches the existing precedent in the same file (`#session-scrim[hidden]` at `app.css:1024`, `#lightbox[hidden]` at `app.css:1056`).

**Mutation check:** I deleted `app.css:52` and re-ran the new suite — 6 of 12 assertions fail. The regression test is real, not vacuous, and it would have caught #942.

**Test results:** full suite green — `125 passed, 0 failed` for the pre-existing tests plus `12 passed, 0 failed` for `version-banner`. It's wired into both `test` and `test:ci`, and CI runs `test:ci` (`.github/workflows/ci.yml:157`), so it will actually gate.

### Critical Issues

None. No Red findings — I found no defect, regression, or security issue. Everything below is test coverage.

### Security Review

**Strengths:**
- `app.js:292-295` builds the message and assigns via `textContent`, never `innerHTML`. The daemon-supplied `status.update_command` is interpolated into that string, so `textContent` is what keeps this from being an injection sink — the right call.
- Both `localStorage` accesses are inside `try`/`catch` (`app.js:228-232`, `app.js:266-268`). This matters beyond quota errors: in a cookies-blocked context, touching the `localStorage` property itself throws `SecurityError`, and both call sites have that access inside the guarded block.
- The persisted value is an opaque SHA compared by equality; it never reaches a selector, URL, or HTML sink.
- `role="status"` and the dismiss button's `aria-label` are preserved in `index.html`; the test fixture markup matches `index.html:13-15` exactly, so the DOM under test is the real one.

**Concerns:** none found.

### Code Quality

Several things here are done well and worth calling out, because they're the non-obvious choices:
- `dismiss.onclick =` (not `addEventListener`) — the handler is rebound on every render, so `onclick` replaces rather than accumulating listeners. Correct choice.
- `dismissArmedWithoutSha` (`app.js:290`) is a sound guard: a stray click on a hidden banner can't permanently suppress the no-SHA banner.
- `if (text.textContent !== msg)` (`app.js:295`) avoids re-mutating a `role="status"` live region on every poll, which prevents repeat announcements.
- The in-memory `versionBannerDismissedSha` fallback means a dismiss sticks even when `setItem` fails — and the test proves it.

I traced all three restructured paths (SHA dismiss, no-SHA dismiss, RPC failure) and believe each is correct. My findings are that three branches this PR **introduced or touched** ship with no coverage, in a PR whose entire purpose is regression-proofing this component against exactly this class of silent breakage.

**Yellow 1 — new no-SHA dismiss logic is untested.** `dismissArmedWithoutSha` (`app.js:261`, `269-271`, `290`) is genuinely new conditional logic, not a refactor, and it changes *when* `versionBannerDismissedWithoutSha` gets set. No test renders `{ state: 'behind', behind_by: N }` without `remote_sha`. Suggested: render with no `remote_sha` → visible; dismiss → hidden; re-render with no `remote_sha` → still hidden; render with a `remote_sha` → visible again (asserting the `app.js:279` reset).

**Yellow 2 — the `getItem` failure path is untested, though the helper is 90% there.** `makeStorage` already has `setSetItemThrows` (`version-banner.test.js:35`) but no `setGetItemThrows`, so the catch at `app.js:230-232` never executes under test. This is the load-bearing half: if a `getItem` throw escaped, `renderVersionUpdateBanner` would abort and the banner would never render *at all* in a cookies-blocked browser — silent, and invisible to every existing assertion. Adding a `setGetItemThrows` toggle mirroring line 35 is ~3 lines.

**Yellow 3 — test-plan checkbox 4 is unverified.** "`version-update-get` failing … leaves no visible banner" is in the PR description, and `app.js:307-312` is a line this PR changed, but no test calls `refreshVersionUpdateBanner` — the suite only drives `renderVersionUpdateBanner` through `__t.render`. Stubbing `rpc` to reject would close it.

**Green — suite inventory not updated.** `web-tests/package.json`'s `description` enumerates every test area and is currently a complete inventory of all 10 files; `version-banner` is the first addition not folded in. `web-tests/README.md` also keeps a per-file section for most files — though `sidebar-groups`/`sidebar-select` already lack one, so that half is a pre-existing gap, not yours.

### Summary Table

**Recommendation:** Request Changes — driven by [0 Red, 3 Yellow, 1 Green] findings.

To be clear about what this verdict is and isn't: the fix itself is correct, minimal, and well-targeted, and I'd have approved it on the production diff alone. The three Yellows are all "add the test," each a few lines against scaffolding that already exists in your own helper. Flagging them now lands them in this round trip rather than a follow-up.

---

dhilgaertner (approved):
## Code & Security Review

### Critical Issues

None.

### Security Review

**Strengths:**

- **No injection surface.** All daemon-supplied values (`behind_by`, `update_command`, `remote_sha`) reach the DOM through `textContent`, never `innerHTML` (`app.js:292-295`). Markup in `update_command` cannot execute. `remote_sha` is only ever written to / compared against `localStorage` under a constant key with `===` — no dynamic key construction.
- **This PR closes a real availability bug in blocked-storage contexts.** Pre-PR, `renderVersionUpdateBanner` called `localStorage.getItem` unguarded. In a context where storage access throws `SecurityError` (Safari Lockdown, third-party iframe, some private modes), that exception propagated out of the render, was swallowed by `refreshVersionUpdateBanner`'s `catch`, and the banner was hidden *permanently* — users in those contexts would never learn their build was stale. `isDismissedSha` (`app.js:225-233`) now contains the throw and degrades to "not dismissed", which is the correct fail-open direction for a notice. Directly covered by the `getItem throws but banner still renders when behind` case.
- **Failure modes of the `catch (_) {}` blocks are benign and deliberate.** A `setItem` throw (`app.js:266-268`) degrades dismiss from "persists across reload" to "persists for this tab" via the `versionBannerDismissedSha` fallback — it does not crash the render or lose the dismiss within the session.
- No new RPC methods, no auth/authz surface, no network calls, no secrets touched.

**Concerns:**

None found.

### Code Quality

**The CSS fix is correct and idiomatic.** `.update-banner[hidden] { display: none; }` (`app.css:52`) carries specificity `(0,2,0)` against `.update-banner`'s `(0,1,0)`, so it wins on specificity rather than source order — and it is placed immediately after regardless. It matches the existing precedent in the same file exactly: `#session-scrim[hidden]` (`app.css:1024`) and `#lightbox[hidden]` (`app.css:1056`). I confirmed `app.css` is the only stylesheet touching `.update-banner`, `index.html:13` carries no inline `style`, and no media query re-declares `display` on it.

**The JS change fixes the genuinely more subtle half of the bug.** Pre-PR, `dismiss.onclick` was assigned only on the *show* path, at the very end of the function. That means the exact state the CSS bug produced — a banner visible when it should not be — was also a banner whose × did nothing. Hoisting the binding above the early returns (`app.js:262-272`) and arming it via `dismissRemoteSha` / `dismissArmedWithoutSha` is the right shape: the handler always exists, and what it *records* is scoped per-render.

**Undocumented accessibility improvement worth noting.** The banner is `role="status"`, an implicit `aria-live="polite"` region. Reordering to `banner.hidden = false` *before* the text write (`app.js:291-295`) is the correct sequence — a live region must be in the accessibility tree before its content changes for the change to be announced. The old order (write text while `hidden`, then reveal) is a pattern most screen readers do not announce at all. The `if (text.textContent !== msg)` guard then prevents re-announcing an unchanged message on every poll. Both are right; neither is mentioned in the PR description.

**`hideVersionUpdateBanner` collapses four duplicated 2-line hide blocks into one**, which is what makes "clear text on *all* hide paths" enforceable rather than something a fifth path could quietly skip.

**Tests are real regression tests, and I verified that.** Reverting only the one CSS line and re-running `version-banner.test.js` fails **10 of 20** assertions. That is the property that matters here: the suite asserts on `window.getComputedStyle(b).display !== 'none'` rather than on the `hidden` *property*, so it tests the user-visible outcome — precisely what a property-only assertion would have missed for this bug. It loads the real `app.css` and real `app.js` (not stubs), and covers both storage-throw directions, cross-poll dismiss persistence, no-SHA session dismiss and its reset when a SHA arrives, and the RPC-failure path.

Full suite passes locally: **125 existing + 20 new = 145/145**. CI shows Catalog & Parity Gates, Desktop (Tauri) Build, and Shell Lint green; Build & Test was still pending at review time. Both `test` and `test:ci` scripts pick up the new file, and the README entry (`web-tests/README.md`) matches house documentation style.

**Consider (non-blocking, no change requested):**

- `isDismissedSha`'s `if (!sha) return false` (`app.js:226`) is unreachable via its only call site, which already guards with `remoteSha &&` (`app.js:280`). Harmless defensiveness; worth keeping if the helper may gain callers.
- `versionBannerDismissedSha` is never cleared for the tab's lifetime, so an in-memory dismiss survives an external `localStorage.clear()`. Almost certainly the behaviour you want for a dismissible notice — flagging only so it is a deliberate choice rather than an accident.

### Summary Table

Tight, well-scoped fix with tests that genuinely fail without it, plus two unadvertised improvements (storage-throw containment and live-region announcement ordering) that are correct on their own merits.

---

PR #949, opened 2026-08-08
Close review rounds so a re-request starts round 2 (CROW-945)

### LINKED ISSUE #945: A second review request never starts a second review session: viewerLastReviewedAt is always nil (fractional-seconds ISO8601 formatter), so the round never closes
A second review request on the same PR never starts a second review session. Round 1
finishes (Crow posts `CHANGES_REQUESTED`), the author pushes fixes and re-requests
review, and the Reviews board renders **"Go to Session"** pointing at the dead round-1
session instead of **"Start Review"**. The only way forward is to delete the old
session by hand.

The design *intends* to handle this — `decideReviewCompletions` rule 1
(`Packages/CrowEngine/Sources/CrowEngine/IssueTracker.swift:4250-4269`) exists to close
the round so the next request lands with no linked session. It has never been able to
fire in production.

## Root cause 1 — `viewerLastReviewedAt` is always `nil`

Rule 1's entire input is `ReviewRequest.viewerLastReviewedAt`:

```swift
// IssueTracker.swift:4264
if let request = reviewRequestsByPRURL[prLink.url],
   let reviewedAt = request.viewerLastReviewedAt,
   reviewedAt > session.createdAt {
    decisions.append(CompletionDecision(sessionID: session.id, reason: "viewer submitted review"))
```

That field is parsed in `GitHubCodeBackend.parseReviewRequests`
(`Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubCodeBackend.swift:752-764`)
with the formatter built by its caller at
`GitHubCodeBackend.swift:482-483`:

```swift
let dateFmt = ISO8601DateFormatter()
dateFmt.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
```

GitHub's GraphQL `DateTime` scalar has **no** fractional seconds:

```
$ gh api graphql -f query='… reviews(last:5){nodes{state submittedAt}} …'
{"state":"CHANGES_REQUESTED","submittedAt":"2026-08-06T06:14:47Z"}
```

An `ISO8601DateFormatter` configured `.withFractionalSeconds` returns `nil` for that
shape:

```swift
let f = ISO8601DateFormatter()
f.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
f.date(from: "2026-08-06T06:14:47Z")   // nil
f.formatOptions = [.withInternetDateTime]
f.date(from: "2026-08-06T06:14:47Z")   // 2026-08-06 06:14:47 +0000
```

This is the **exact trap already documented in this file**. `parseGitHubDateTime`
(`GitHubCodeBackend.swift:704-726`) was added to fix it for `PRStatus`, and its
docstring names the remaining offenders:

> "a strict formatter silently disables features that depend on parsed timestamps
> (CROW-508 `needsRefine` was inert this way; see PR #509 review). … Other call sites
> in this file still use the brittle pattern (see **`parseReviewRequests`**,
> `parseStaleMRResponse`, `parseStaleStateResponse`). They're out of scope for this
> PR but should migrate to this helper in a follow-up — same trap, different fields."

The follow-up never happened, so `parseReviewRequests` is still inert.

**Why the tests pass.** The only parser coverage,
`BackendsTests.testParseReviewRequestsSkipsNullNodes`
(`Packages/CrowProvider/Tests/CrowProviderTests/BackendsTests.swift:441-468`), uses a
hand-written fixture *with* fractional seconds:

```json
"updatedAt":"2026-06-07T10:00:00.000Z",
"reviews":{"nodes":[…{"state":"APPROVED","submittedAt":"2026-06-06T10:00:00.000Z"}…]}
```

`XCTAssertNotNil(req.viewerLastReviewedAt)` therefore passes against a payload GitHub
never sends. `IssueTrackerCompletionTests` builds `ReviewRequest` values directly, so
it never exercises the parser either.

**Two more symptoms of the same line**, both visible on the Reviews board today:
`requestedAt: updatedAt` (`GitHubCodeBackend.swift:776`) is nil for every request, so
the `relTime(r.requested_at)` chip in `reviewCard`
(`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js:3957`) never renders, and
the sort at `GitHubCodeBackend.swift:783` degrades to `.distantPast` for every element —
the board's "newest first" ordering is arbitrary.

## Root cause 2 — the round-closing signal is read from a search that no longer contains the PR

Fixing the formatter is necessary but **not sufficient**. Rule 1 reads
`reviewRequestsByPRURL`, which is built from `reviews` — the
`review-requested:@me state:open type:pr` search
(`GitHubCodeBackend.swift:71`, wired at `IssueTracker.swift:774-779`).

GitHub clears the pending review request the moment the requested reviewer submits a
review, so the PR **drops out of that search** at exactly the instant rule 1 wants to
fire. (Same mechanism #921 documented: "`reviewRequests` on the PR is empty — GitHub
cleared it when the review was submitted.") Confirmed live: `search(review-requested:@me)`
returns `issueCount: 0` and `crow list-reviews` is empty, despite reviews submitted on
`corveil/corveil#2121` earlier the same day.

So even with the date parsed, rule 1 would only fire if a poll happened to land in the
gap between the review being submitted and GitHub's search index dropping the row — a
race, not a mechanism. **The round-closing signal has to come from somewhere that
survives the request being consumed.**

## How that becomes the observed symptom

1. Round 1 review session posts `CHANGES_REQUESTED`. Rule 1 can't fire → session stays
   `.active`.
2. `AppState.reviewSessions` (`Packages/CrowCore/Sources/CrowCore/AppState.swift:482-484`)
   excludes only `.completed`/`.archived`, so the dead session is still live.
3. Author pushes fixes, re-requests review. The PR reappears in the search;
   `IssueTracker.swift:735-741` re-attaches `reviewSessionID` = the stale session.
4. `reviewCard` (`app.js:3931`, `3967-3970`) sees `review_session_id`, marks the card
   not-selectable and renders **"Go to Session"** — no way to start round 2.
5. `crow start-review --url …` doesn't help either: `SessionService.createReviewSession`
   (`Packages/CrowEngine/Sources/CrowEngine/SessionService.swift:2853-2857`)
   short-circuits on `existingReviewSession(forPRURL:)`
   (`AppState.swift:692`) and returns the stale session's ID.
6. Deleting the stale session is the only unblock.

The `.reReview` head-advance fallback in `reviewKickoffAction`
(`IssueTracker.swift:4214-4235`) *would* catch this — it completes the stale session
and spins up a fresh one — but it is only ever called from the daemon's auto-kickoff
hook (`Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift:776-818`), which returns
immediately unless some workspace sets `autoReviewRepos`:

```
$ crow workspace list | jq '.workspaces[] | {name, auto_review_repos}'
{"name":"RadiusMethod","auto_review_repos":[]}
{"name":"Personal",    "auto_review_repos":[]}
{"name":"Corveil",     "auto_review_repos":[]}
```

With no auto-review repos configured, nothing in the manual/board path ever consults
`reviewKickoffAction`. The board is gated purely on "does a non-completed review
session link to this PR URL", which is a much weaker question than "has this review
round been answered".

## Proposed fix

**1. Migrate `parseReviewRequests` to `parseGitHubDateTime`** (`GitHubCodeBackend.swift:719`)
for both `submittedAt` and `updatedAt`, and take `parseStaleMRResponse` /
`parseStaleStateResponse` with it — the docstring's own follow-up list. Re-fixture the
existing tests with GitHub's real non-fractional shape, and add a regression test
asserting a plain `…T10:00:00Z` timestamp parses. Consider making the brittle
construction unavailable (a single shared formatter helper) so the trap can't recur a
fourth time.

**2. Source the round-closing signal from the PR, not from the request search.**
An active review session's PR is someone else's, so it is never in `viewerPRs` and
already lands in `collectStalePRURLs` (`IssueTracker.swift:1124-1140`) → the batched
`prStates` query (`GitHubCodeBackend.swift:103-132`) already fetches it every poll.
That query does not currently select reviews; adding

```graphql
reviews(last: 20) { nodes { author { login } state submittedAt } }
reviewRequests(first: 20) { totalCount nodes { requestedReviewer { … on User { login } } } }
```

gives `viewerLastReviewedAt` (and pending-request state) from a source that survives
the request being consumed, at **zero extra API calls**. Rule 1 then becomes reliable
rather than racy.

**3. Make the round, not the PR URL, the dedup key.** Once a review session's round is
closed, it must stop shadowing the PR. Either complete it (rule 1, now working) or
teach the board/`createReviewSession` to compare `session.lastReviewedHeadSha`
(`Packages/CrowCore/Sources/CrowCore/Models/Session.swift:31`) against the request's
current `headRefOid` and treat a mismatch as "not covered". Routing the board's
"Start Review" button and `crow start-review` through `reviewKickoffAction` — so the
manual path gets the same `.create` / `.reReview` / `.skip` decision the auto-kickoff
path already has — is the smaller change and removes the second, divergent dedup rule.

**4. Decide what a re-review should do to the previous session.** The auto-kickoff path
completes the stale session and clones a fresh one (`CrowDaemon.swift:796-816`).
The `reReview` quick action instead re-prompts the *existing* session with a diff
scoped by `lastReviewedHeadSha` (`AutoRespondCoordinator.swift:263-287`), which is
cheaper and keeps the reviewer's context. Whichever is chosen, the board must offer
it — today it offers neither.

## Acceptance

- [ ] `parseReviewRequests` parses GitHub's real non-fractional timestamps;
      `viewerLastReviewedAt` and `requestedAt` are non-nil against a live payload.
- [ ] The Reviews board shows relative "requested" times again and sorts newest-first.
- [ ] A review session is auto-completed once the viewer submits its verdict, without
      depending on the PR still being in `review-requested:@me`.
- [ ] Author pushes fixes + re-requests review ⇒ the board offers a new review round
      (Start Review or Re-review) with no manual session deletion.
- [ ] `crow start-review` on a PR whose previous round is closed starts a new round
      instead of returning the stale session ID.
- [ ] Re-requesting review with **no** new commits does not spawn a duplicate round.
- [ ] The manual/board path and the `autoReviewRepos` auto-kickoff path share one
      decision function; no second dedup rule.
- [ ] Test fixtures use GitHub's actual timestamp shape; a regression test would fail
      if the fractional-seconds formatter came back.

## Workaround

Delete the stale review session, then press Start Review:

```bash
crow delete-session --session <stale-review-session-uuid>
crow start-review --url https://github.com/owner/repo/pull/123
```

Or re-prompt the existing session in place:

```bash
crow quick-action --session <stale-review-session-uuid> --action reReview
```

dhilgaertner:
Closes #945

A second review request on the same PR never started a second review session. Round 1 posted `CHANGES_REQUESTED`, the author pushed fixes and re-requested, and the Reviews board rendered **"Go to Session"** pointing at the dead round-1 session. Deleting that session by hand was the only way forward.

`decideReviewCompletions` rule 1 exists to close the round. It had never fired in production, for two independent reasons.

## Root cause 1 — the date never parsed

`parseMonitoredPRsResponse` built an `ISO8601DateFormatter` with `.withFractionalSeconds` and handed it to `parseReviewRequests`. GitHub's GraphQL `DateTime` scalar carries no fraction, and that formatter returns `nil` for that shape — so `viewerLastReviewedAt`, rule 1's entire input, was always nil. So was `requestedAt`, which is why the board's "requested" chip never rendered and its newest-first sort was arbitrary.

This is the trap `parseGitHubDateTime` was written to fix; its docstring named `parseReviewRequests` as an unmigrated call site. The tests missed it because the fixture used `.000Z` timestamps GitHub has never sent.

## Root cause 2 — the signal came from a search that no longer contains the PR

Rule 1 read the `review-requested:@me` payload, but GitHub clears the pending request the instant the viewer submits a review — the PR leaves that search at exactly the moment the round should close. Sourced only from there, rule 1 could fire only by winning a race against GitHub's search index.

## What changed

**One tolerant parser.** `IssueDate.parse` absorbs its byte-identical twin `parseGitHubDateTime` — two copies of the fix is how a third call site still got it wrong — and the five remaining hand-rolled formatters in the GitHub/GitLab backends migrate to it.

**Rule 1 gains a second source that survives the request being consumed.** `reviews(last: 1, author: $viewer, states: [APPROVED, CHANGES_REQUESTED, DISMISSED])` on the batched `prStates` query, which already fetches every review session's PR each poll — **zero extra API calls**.

Not `viewerLatestReview`, which the ticket suggested: schema introspection confirms it takes no arguments, so it returns the viewer's latest review of *any* state. A follow-up `--comment` review (COMMENTED) or an unsubmitted draft (PENDING, null `submittedAt`) would mask the real verdict and reproduce this exact bug through a new field. `reviews` accepts `author:` and `states:`, so the filter happens server-side: one node per alias instead of twenty, and immune to the >20-review truncation the search path has.

The new branch is presence-based and so is deliberately **not** gated on `prDataComplete` — that flag exists to stop *absence* being read as evidence, and gating presence would make the fix silently miss whenever an unrelated provider errored.

**One dedup rule.** `createReviewSession` now runs the same `IssueTracker.reviewKickoffAction` the auto-review hook used, against the head it fetches itself, so `crow start-review` starts a new round instead of returning the stale session id. The daemon hook drops its duplicate decision — two callers deciding from two different heads is the divergence that let a round keep shadowing its PR.

**The board offers the next round.** `list-reviews` gains `head_ref_oid` and a server-computed `kickoff_action`, and the two byte-identical copies of that handler collapse into one shared builder (the CLI and the web agreed only by accident of copy-paste). Cards render Start Review / Re-review / Go to Session accordingly. The action is advisory — it's computed from a poll-stale head — so the client never suppresses the RPC on it.

## Pre-existing bugs found on the same path

Fixed here because the new field would have joined them:

- `autoCompleteFinishedReviews` considered only `.active` sessions, so a `.paused` or `.inReview` review session blocked its PR forever with no rule anywhere able to complete it.
- `mergePRRecords` dropped `updatedAt`, despite its own *"Every field must be carried here"* banner and two named prior incidents.
- `withLabels` dropped the three CROW-921 reviewer fields, despite claiming to preserve every field; its caller feeds `evaluateAutoMerge`, which reads them.

## `lastReviewedHeadSha` split

It had two readers wanting opposite things. It stays the **round boundary** the kickoff guard reads; the new `lastReRequestedHeadSha` is the **diff anchor** the in-place re-review prompt scopes from. Repeat passes stop replaying round 1's delta without disarming the head-advance fallback for an agent that stalls without posting a verdict.

## Acceptance

- [x] `parseReviewRequests` parses GitHub's real non-fractional timestamps
- [x] The Reviews board shows relative "requested" times again and sorts newest-first
- [x] A review session is auto-completed once the viewer submits its verdict, without depending on the PR still being in `review-requested:@me`
- [x] Author pushes fixes + re-requests ⇒ the board offers a new review round
- [x] `crow start-review` on a PR whose previous round is closed starts a new round
- [ ] **Deferred:** re-requesting with no new commits does not spawn a duplicate round
- [x] The manual/board path and the auto-kickoff path share one decision function
- [x] Test fixtures use GitHub's actual timestamp shape; a regression test would fail if the fractional-seconds formatter came back

**On the deferred box:** the only re-kick loop is the `autoReviewRepos` auto path, which already dedups via an in-memory SHA fingerprint. That loop is self-terminating (GitHub clears the request once a verdict posts), the in-flight clone window is covered by `existingByPRSessionID`, and `autoReviewRepos` is empty for every workspace here — so the gap is a daemon restart on a path that isn't live. Making it durable means persisting the fingerprint with a pruning policy; putting the guard in the shared decision function instead would gate the *manual* Start Review button, so "I want to look at this again" would fail with an error dialog.

## Verification

- `make daemon` clean; `make parity` passes all three gates.
- CrowProvider 163, CrowEngine 655, CrowCore, CrowCLI — all green.
- Web (jsdom) suite green, including 7 new `kickoff_action` board assertions.
- CrowDaemon has 2 failing tests (`refreshTerminalsRebindsTheActiveTerminalToTheFreshRow`, `bootCatchResetsTheHistory`). **Pre-existing** — verified by swapping `HEAD`'s `app.js` back in and re-running: they fail identically. Both are source-shape assertions on `app.js` regions this PR doesn't touch, and one anchors on a string absent from `HEAD` too.
- The new query shape was run live against GitHub, returning `"submittedAt":"2026-08-05T22:22:04Z"` — exactly the non-fractional form the old formatter returned nil for.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Models/Session.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/Session.swift
@@ -22,13 +22,29 @@ public struct Session: Identifiable, Codable, Sendable {
     // prompt dispatched. Gates the launchClaude prompt-vs-`--continue`
     // branch so completed reviews don't restart on app relaunch.
     public var reviewPromptDispatched: Bool
-    // Head SHA of the PR at the time the review session was created or
-    // last re-launched. Used by the kickoff guard as a fallback re-kick
-    // signal when a PR's head advances without an explicit re-request
-    // (e.g. force-push) or before the viewer-submitted-review signal is
-    // observed. Nil for non-review sessions and for legacy persisted
-    // sessions predating this field (CROW-290).
+    // Head SHA of the PR at the time the review session was created. Used by
+    // the kickoff guard as a fallback re-kick signal when a PR's head advances
+    // without an explicit re-request (e.g. force-push) or before the
+    // viewer-submitted-review signal is observed. Nil for non-review sessions
+    // and for legacy persisted sessions predating this field (CROW-290).
+    //
+    // Deliberately NOT updated when a round is re-reviewed in place. This is a
+    // *round boundary*, and the kickoff guard's whole job is to notice "the
+    // head moved and nobody opened a new round". Advancing it on re-review
+    // would disarm that fallback for the new head, so an agent that then
+    // stalled without posting a verdict would leave the PR with nothing to
+    // re-fire. For the *diff anchor* — "what should `git diff …HEAD` cover on
+    // the next pass" — use `lastReRequestedHeadSha`, which is what the two
+    // readers actually needed separately (CROW-945).
     public var lastReviewedHeadSha: String?
+    // Head SHA the last in-place re-review was scoped from (CROW-945). Written
+    // by the `reReview` quick action, which re-prompts the *existing* session
+    // rather than starting a new round, so without it every repeat pass hands
+    // the agent an ever-widening diff from round 1's head — more tokens, and
+    // findings the agent already reported. Read only for prompt scoping, never
+    // by the kickoff guard; see `lastReviewedHeadSha`. Nil until the first
+    // in-place re-review, and for every session predating this field.
+    public var lastReRequestedHeadSha: String?
     // Timestamp at which Crow enabled GitHub native auto-merge on the
     // linked PR (CROW-299). Non-nil means the one-shot enable has already
     // run; the auto-merge watcher skips this session on subsequent polls.
@@ -143,6 +159,7 @@ public struct Session: Identifiable, Codable, Sendable {
         updatedAt: Date = Date(),
         reviewPromptDispatched: Bool = false,
         lastReviewedHeadSha: String? = nil,
+        lastReRequestedHeadSha: String? = nil,
         autoMergeEnabledAt: Date? = nil,
         locked: Bool = false,
         agentSessionStartedAt: Date? = nil,
@@ -165,6 +182,7 @@ public struct Session: Identifiable, Codable, Sendable {
         self.updatedAt = updatedAt
         self.reviewPromptDispatched = reviewPromptDispatched
         self.lastReviewedHeadSha = lastReviewedHeadSha
+        self.lastReRequestedHeadSha = lastReRequestedHeadSha
         self.autoMergeEnabledAt = autoMergeEnabledAt
         self.locked = locked
         self.agentSessionStartedAt = agentSessionStartedAt
@@ -207,6 +225,7 @@ public struct Session: Identifiable, Codable, Sendable {
         updatedAt = try container.decode(Date.self, forKey: .updatedAt)
         reviewPromptDispatched = try container.decodeIfPresent(Bool.self, forKey: .reviewPromptDispatched) ?? true
         lastReviewedHeadSha = try container.decodeIfPresent(String.self, forKey: .lastReviewedHeadSha)
+        lastReRequestedHeadSha = try container.decodeIfPresent(String.self, forKey: .lastReRequestedHeadSha)
         autoMergeEnabledAt = try container.decodeIfPresent(Date.self, forKey: .autoMergeEnabledAt)
         agentSessionStartedAt = try container.decodeIfPresent(Date.self, forKey: .agentSessionStartedAt)
         agentSessionEndedAt = try container.decodeIfPresent(Date.self, forKey: .agentSessionEndedAt)
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift
@@ -784,46 +784,28 @@ public enum CrowDaemon {
         // (force-push, or round-2 commits landing before Signal A —
         // `decideReviewCompletions` rule 1 — completed round 1). The SHA-keyed
         // fingerprint makes each head its own round so B isn't blocked by A.
+        //
+        // Both decisions now live in `SessionService.createReviewSession`, which
+        // runs the same `IssueTracker.reviewKickoffAction` against the head it
+        // fetches itself (CROW-945). This hook deliberately does NOT pre-decide:
+        // it would have to use `request.headRefOid` — a board snapshot up to a
+        // poll old — and two callers deciding from two different heads is
+        // exactly the divergence that let a round go on shadowing its PR. So
+        // this stays a filter + a rate guard, and the service owns the verdict.
         var autoReviewed: Set<String> = []
         tracker.onReviewRequestsRefreshed = { requests in
             let patterns = (config()?.workspaces ?? []).flatMap { $0.autoReviewRepos }
             guard !patterns.isEmpty else { return }
             for request in requests {
                 guard repoMatchesPatterns(request.repo, patterns: patterns) else { continue }
-                let linkedSession = request.reviewSessionID.flatMap { id in
-                    appState.sessions.first(where: { $0.id == id })
-                }
-                let action = IssueTracker.reviewKickoffAction(
-                    reviewSessionID: request.reviewSessionID,
-                    headRefOid: request.headRefOid,
-                    linkedSession: linkedSession,
-                    existingByPRSessionID: appState.existingReviewSession(forPRURL: request.url)?.id
-                )
-                guard action != .skip else { continue }
                 // SHA-keyed dedup: a new head is a fresh round; an unchanged head
                 // never re-kicks (also guards the in-flight clone window before
-                // the session's link/reviewSessionID land).
+                // the session's link/reviewSessionID land). In-memory only, so a
+                // daemon restart can re-ask — `createReviewSession` still refuses
+                // to duplicate a live round, so the worst case is one redundant
+                // metadata fetch, not a duplicate session.
                 let fingerprint = "\(request.id)\n\(request.headRefOid ?? "")"
                 guard autoReviewed.insert(fingerprint).inserted else { continue }
-                // B-fallback: retire the stale round's review session before
-                // enqueuing so `reviewSessions` doesn't double up for this PR.
-                // Complete it (not delete): completing writes its end-of-round
-                // `SessionAnalyticsSnapshot` and keeps its telemetry, whereas
-                // `deleteSession` transitions no status — so no snapshot is
-                // written — and then drops the raw rows, silently erasing round
-                // 1's tokens/cost from the scorecard on Crow's own auto-review
-                // flow (CROW-877 review). Completing also makes it invisible to
-                // `existingReviewSession(forPRURL:)` (that dedup guard excludes
-                // completed/archived), so the create below starts the new round
-                // instead of reusing the stale session. The identical completed
-                // `review-<repo>-<pr>` rows that used to pile up in Completed are
-                // now folded into one by the sidebar's within-section collapse
-                // (`groupSessions` in app.js), which also cleans up pre-existing
-                // pile-ups — so the visible-duplicate half no longer needs a
-                // destructive teardown here.
-                if case let .reReview(staleID) = action {
-                    appState.onCompleteSession?(staleID)
-                }
                 let url = request.url
                 Task { _ = await reviewSerializer.enqueue { await sessionService.createReviewSession(prURL: url, selectAfterCreate: false) } }
             }
--- a/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
@@ -703,31 +703,7 @@ func makeCommandRouter(
         },
         "list-reviews": { _ in
             if tracker != nil {
-                return await MainActor.run {
-                    let fmt = ISO8601DateFormatter()
-                    let reviews: [JSONValue] = appState.filteredReviewRequests.map { r in
-                        .object([
-                            "id": .string(r.id),
-                            "pr_number": .int(r.prNumber),
-                            "title": .string(r.title),
-                            "url": .string(r.url),
-                            "repo": .string(r.repo),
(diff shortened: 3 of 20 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- `viewerLogin` is sourced from GitHub's authenticated GraphQL response and passed as a typed GraphQL variable (`-f viewer=…`, not `-F`), avoiding shell/GraphQL injection and the all-digit-login type-inference trap (`GitHubCodeBackend.swift:1313-1317`).
- `roundClosingReviewStates` is a hardcoded set embedded in the query — no user-controlled GraphQL fragments.
- `viewerLastReviewedAt` uses a strict nil-is-"not-fetched" convention; rule 1 is presence-based and gated by `> session.createdAt`, preventing round-1 verdicts from closing round-2 sessions.
- `SessionService.createReviewSession` calls `completeSession(id:)` directly instead of the optional `onCompleteSession` callback, closing the double-session hole when the daemon callback is unwired.

**Concerns:**
- No security issues identified. `head_ref_oid` in the `list-reviews` payload is a public commit SHA — acceptable exposure for board UX.

### Code Quality
- **Root-cause fix is sound:** consolidating date parsing into `IssueDate.parse` and migrating all hand-rolled formatters eliminates the silent-nil trap that broke `requestedAt`, `viewerLastReviewedAt`, and round closure (`BackendTypes.swift:235-244`, `GitHubCodeBackend.swift:777-790`).
- **Dual-source rule 1:** reading `viewerLastReviewedAt` from both `reviewRequestsByPRURL` and `prsByURL` (via `reviews(last:1, author:, states:)`) correctly survives GitHub clearing the `review-requested:@me` search at verdict time (`IssueTracker.swift:4311-4316`).
- **Unified kickoff decision:** `SessionService.createReviewSession` and `ReviewsPayload.build` both call `IssueTracker.reviewKickoffAction` after fetching/linking, eliminating the divergent dedup that left dead round-1 sessions shadowing re-requests.
- **Pre-existing bugs fixed on-path:** `mergePRRecords` now carries `viewerLastReviewedAt`/`updatedAt`; `withLabels` preserves CROW-921 reviewer fields; `autoCompleteFinishedReviews` includes `.paused`/`.inReview` sessions.
- **`lastReRequestedHeadSha` split** cleanly separates round-boundary (`lastReviewedHeadSha`) from diff-anchor (`lastReRequestedHeadSha`) for in-place re-reviews (`AutoRespondCoordinator.swift:107-127`).
- **Test coverage is excellent:** regression tests for non-fractional GitHub timestamps, PR-sourced round closure without review-request search, `kickoff_action` board rendering, and merge-field preservation. Swift build and targeted test suites pass locally.

### Summary Table

Green notes (non-blocking):
1. Re-request without new commits remains deferred per PR scope; the in-memory SHA fingerprint mitigates the auto-review path, and manual Start Review is intentionally not gated.
2. GitLab workspaces will not populate `viewerLastReviewedAt` on PR records; round closure there still depends on the review-request search path — acceptable given this ticket's GitHub focus.
3. `kickoff_action` is correctly treated as advisory (poll-stale head) while the server fetches fresh metadata before deciding.

---

PR #950, opened 2026-08-08
Measure auto-rebase behind-ness with git, not mergeStateStatus (CROW-944)

### LINKED ISSUE #944: Auto-rebase misses behind-but-clean PRs: behind-ness is read only from mergeStateStatus == BEHIND, which BLOCKED/DIRTY/DRAFT/UNKNOWN mask
A PR that is **behind its base with no conflicts** — the state GitHub renders as a
lone "Update branch" button — is not reliably picked up by the auto-rebase watcher.
On repos with `strict_required_status_checks_policy` (branches must be up to date
before merging) that stalls the PR short of merge until a human presses the button.

The watcher itself works — `crowd-automation.log` has plenty of
`auto-rebase: #NNNN rebased & force-pushed (was BEHIND)`. The problem is that
Crow's only definition of "behind" is one GitHub field that frequently reports
something else, plus three dead-ends in the surrounding wiring.

## Root cause 1 — behind-ness is read from a field that masks it

`IssueTracker.shouldAttemptAutoRebase`
(`Packages/CrowEngine/Sources/CrowEngine/IssueTracker.swift:3599`) is the whole
decision:

```swift
guard pr.state == "OPEN" else { return false }
return pr.mergeStateStatus == "BEHIND" || pr.mergeable == "CONFLICTING"
```

`mergeStateStatus` is GitHub's **single-valued** summary of the merge button, not a
set of flags. It reports the highest-priority reason the button isn't green, so
`BEHIND` only surfaces when nothing above it applies. A branch that is behind base
**and** any of the following reports the other value instead:

| Also true | `mergeStateStatus` | Behind-ness visible to Crow? |
|---|---|---|
| Required review not yet in / required check failing or missing | `BLOCKED` | ✗ |
| Merge conflict | `DIRTY` | ✗ (caught by the `CONFLICTING` clause, by luck) |
| Draft | `DRAFT` | ✗ |
| Base moved and GitHub hasn't recomputed mergeability yet | `UNKNOWN` | ✗ |

`BLOCKED` is the expensive one, because it is the *normal* state of a PR waiting on
its reviewer — precisely the window in which a busy base branch drifts ahead.
`corveil/corveil`'s ruleset makes this the default shape of every PR:

```jsonc
// gh api repos/corveil/corveil/rules/branches/main
{"type":"pull_request",          "parameters":{"required_approving_review_count":1,
                                               "require_code_owner_review":true, …}}
{"type":"required_status_checks","parameters":{"strict_required_status_checks_policy":true, …}}
```

So the sequence is: PR opens → base moves → PR is behind but `BLOCKED` → auto-rebase
never sees it → reviewer approves → *now* it flips to `BEHIND` → auto-rebase fires →
force-push restarts the full CI cycle → base has moved again. The rebase Crow could
have done for free while waiting on review is instead serialized behind approval, and
on a fast-moving base the loop can outrun itself.

Behind-ness and `mergeStateStatus` are independent facts, and Crow measures only the
latter. Live example right now:

```
$ gh api repos/corveil/corveil-cloud-terraform/pulls/231 --jq .mergeable_state
dirty
$ gh api repos/corveil/corveil-cloud-terraform/compare/main...feature/codename-spotlight-230-stripe-live-cutover \
    --jq '{behind: .behind_by, ahead: .ahead_by}'
{"behind":38,"ahead":1}
```

38 commits behind, and the field Crow keys on says nothing about it.

## Root cause 2 — `outOfSyncWithRemote(.diverged)` is an unattended dead-end

`GitManager.rebaseOntoBase` (`Packages/CrowGit/Sources/CrowGit/GitManager.swift:160-175`)
correctly refuses to force-push a worktree that has diverged from `origin/<branch>`.
`IssueTracker.deferRebase` (`IssueTracker.swift:3788`) then re-arms the head key with
exponential backoff capped at `autoRebaseDeferralMaxDelay` = 900 s — **forever**.
Nothing reconciles the worktree, no notification fires (`onAutoRebaseConflicts` is
`.conflicts`-only, `IssueTracker.swift:3745-3750`), and there is no `autoRebaseState`
published to the UI the way `AppState.autoMergeState` (`Packages/CrowCore/Sources/CrowCore/AppState.swift:251`)
is for auto-merge. The PR silently sits behind base while the log fills with:

```
$ rg -o "auto-rebase: #[0-9]+ (skipped|deferred):[a-z-]+" ~/Library/Logs/crow/crowd-automation.log \
    | sed 's/#[0-9]*//' | sort | uniq -c | sort -rn
  64 auto-rebase:  deferred:out-of-sync-diverged
  23 auto-rebase:  deferred:dirty-worktree
   4 auto-rebase:  skipped:no-crow-session-trailer
   3 auto-rebase:  deferred:out-of-sync-ahead
```

The two watchers **create** this state for each other: `gh pr update-branch`
(`Packages/CrowProvider/Sources/CrowProvider/Backends/GitHubCodeBackend.swift:351-354`)
adds a merge commit on the *remote*, so the local worktree diverges the instant
auto-merge's update-branch path runs — and auto-rebase then refuses to touch it.

## Root cause 3 — two leaks in the update-branch path

`IssueTracker.attemptUpdateBranch` (`IssueTracker.swift:3501-3521`):

```swift
private func attemptUpdateBranch(session: Session, pr: ViewerPR) async {
    guard let backend = codeBackend(for: session) else { return }        // ← leaks
    guard await prHasCrowAuthoredCommit(pr: pr, backend: backend) else {
        CrowLog.automation(…)
        return                                                            // ← leaks
    }
    defer { autoMergeInFlight.remove(pr.url) }                            // registered too late
```

Both early returns sit **above** the `defer`, so the PR URL stays in
`autoMergeInFlight` for the rest of the process lifetime. `applyAutoMerge`'s
`guard !autoMergeInFlight.contains(prLink.url)` (`IssueTracker.swift:3134`) then skips
that PR on every subsequent poll — no log line, no chip, nothing.

Separately, `autoUpdateBranchAttempted` (`IssueTracker.swift:202`) is never pruned and
never cleared on failure, and a *failed* `gh pr update-branch` leaves `headRefOid`
unchanged — so the key can never clear. Auto-rebase can't rescue it either, because
it deliberately `continue`s past crow:merge BEHIND candidates
(`IssueTracker.swift:3644-3647`) to avoid fighting auto-merge over the same branch.
One failed update ⇒ the PR is stuck until a human pushes a commit. Contrast
`autoRebaseFailureCounts`, which does implement bounded retries
(`shouldRetryFailedRebase`, `IssueTracker.swift:3678`).

Not yet observed in this log (`rg "updateBranch failed"` is empty), but the code path
is unconditional.

## Proposed fix

**1. Stop inferring behind-ness from `mergeStateStatus`; measure it.**
`rebaseOntoBase` already runs `git fetch origin <base> <branch>`
(`GitManager.swift:147`), so the authoritative answer is one command away:

```
git -C <worktree> rev-list --count HEAD..origin/<base>
```

Turn `shouldAttemptAutoRebase` into a *candidate* filter — OPEN and
`mergeStateStatus != "CLEAN"` — and let git decide. Add a
`RebaseOutcome.alreadyUpToDate` case (`GitManager.swift:304`) returned when the range
is empty, so the widened candidate set is a no-op for PRs that don't need anything.
The existing per-head `autoRebaseAttempted` key already caps this at one fetch per
head state, so the extra polling cost is bounded.

For PRs with no usable worktree, `compare(base...head).behind_by` is the API fallback,
but it's one REST call per PR per poll — prefer the local check and keep the API path
for the "no worktree" branch of `attemptRebase` (`IssueTracker.swift:3713-3721`) if it's
wanted at all.

**2. Reconcile a diverged worktree instead of deferring forever.** Rather than backing
off indefinitely, after N consecutive `out-of-sync-*` deferrals either (a) hard-reset
the worktree to `origin/<branch>` when the tree is clean and the local commits are all
present upstream, or (b) escalate — fire a notification and/or delegate to the agent
the way `.conflicts` already does. Silence is the worst of the three.

**3. Publish an `autoRebaseState`** per session mirroring `AutoMergeState`
(`AppState.swift:251`) so `deferred:out-of-sync-diverged` is visible in the session
card, not only in `crowd-automation.log`.

**4. Fix the update-branch path.** Move the `defer` above both early returns; clear
`autoUpdateBranchAttempted` on failure and apply bounded retries as
`shouldRetryFailedRebase` does; and consider `gh pr update-branch --rebase` so the
operation matches what auto-rebase does and stops manufacturing a diverged local
worktree (root cause 2). If auto-rebase becomes reliable for BEHIND, the precedence
`continue` at `IssueTracker.swift:3644-3647` can be inverted — let auto-rebase own
*every* behind branch and drop `shouldUpdateBranchBeforeMerge` (`IssueTracker.swift:2559`)
to a fallback for sessions with no worktree.

## Acceptance

- [ ] A PR that is behind base and `BLOCKED` (required review pending / required
      check not yet green) is rebased and force-pushed without waiting for approval.
- [ ] A PR that is behind base and `UNKNOWN` (GitHub mid-recompute) is picked up on a
      later poll rather than being permanently skipped for that head.
- [ ] A PR that is already up to date is a cheap no-op (`alreadyUpToDate`), logged
      once per head, with no force-push.
- [ ] Draft PRs behind base are still not force-pushed unless that's the intent —
      state the decision either way in the code.
- [ ] `attemptUpdateBranch` cannot leak `autoMergeInFlight` on any return path;
      a regression test covers the no-backend and no-trailer returns.
- [ ] A failed `gh pr update-branch` is retried a bounded number of times rather than
      latching until the head commit changes.
- [ ] A worktree stuck in `out-of-sync-diverged` either self-reconciles or escalates
      visibly within a bounded number of deferrals.
- [ ] Auto-rebase state is visible on the session card, not just in the log.
- [ ] `shouldAttemptAutoRebase` / the new git-side check are unit-tested against the
      `BLOCKED`, `DIRTY`, `DRAFT`, `UNKNOWN` and `CLEAN` shapes.

## Workaround

Press "Update branch" on GitHub, or from the session's worktree:

```bash
git -C <worktree> fetch origin main && git -C <worktree> rebase origin/main \
  && git -C <worktree> push --force-with-lease
```

dhilgaertner:
Closes #944

## The bug

Auto-rebase decided "is this branch behind its base?" from one GitHub field, `mergeStateStatus == "BEHIND"` (`IssueTracker.swift:3599`). That field is **single-valued** — it reports the highest-priority reason the merge button isn't green, not a set of flags — so a PR that is behind base *and* anything else reports the other value:

| Also true | `mergeStateStatus` | Behind-ness visible? |
|---|---|---|
| Required review pending | `BLOCKED` | ✗ |
| Merge conflict | `DIRTY` | ✗ (caught by the `CONFLICTING` clause, by luck) |
| Draft | `DRAFT` | ✗ |
| Base moved, GitHub mid-recompute | `UNKNOWN` | ✗ |

`BLOCKED` is the expensive one: it is the *normal* state of a PR waiting on its reviewer — exactly the window in which a busy base drifts ahead. On a repo with `strict_required_status_checks_policy` the PR then stalled short of merge until a human pressed **Update branch**, and the rebase Crow could have done for free during review was serialized behind approval instead.

## The fix

`shouldAttemptAutoRebase` becomes a *candidate* filter (OPEN and not `CLEAN`); git answers the real question.

- **`GitManager.behindBase`** — one fetch + `rev-list --count origin/<branch>..origin/<base>`. It compares the two **remote** refs, not `HEAD`: behind-ness is a property of the PR head, and a `HEAD`-based probe would read a locally-rebased-but-unpushed branch as up to date and swallow the `out-of-sync-ahead` deferral that tells you to push. Three-valued (`.upToDate` / `.behind(n)` / `.unknown`) so a fetch flake falls through to a full attempt rather than silently skipping.
- It runs **before** `prHasCrowAuthoredCommit`, so the widened candidate set costs a `git fetch` per head state rather than a provider API call.
- **`RebaseOutcome.alreadyUpToDate`**, checked after the fast-forward reconcile. Without it a no-op rebase (exit 0) plus a no-op force-push ("Everything up-to-date", exit 0) returned `.rebasedAndPushed` and fired a phantom "Branch rebased" notification.

### The trap this created, and how it's handled

Widening + the per-head latch means a PR probed while merely `BLOCKED`-and-not-yet-behind burns its one attempt — and the base moving afterwards is **invisible** in `headRefOid`, so it would never be looked at again. Worse than the original bug. `autoRebaseUpToDateHeads` re-arms the latch when GitHub's view of that same head *changes*, and only then, so a persistent git/GitHub disagreement can't hot-loop.

## Three fixes around it

**A wedged branch now says so.** `out-of-sync-diverged` backed off at a 15-minute cap **forever**, in silence — 64 such lines in my `crowd-automation.log`. After 5 consecutive deferrals (where `autoRebaseDeferralBackoff` saturates; a test pins the two together) it publishes `AutoRebaseState(.blocked)` and fires a new `autoRebaseStuck` notification. Escalating is **not** giving up — Crow keeps retrying at the cap, so a human fix lands next cycle. It deliberately does **not** self-heal a diverged worktree (by definition that holds commits `origin` lacks; a reset would destroy them) and does **not** hand off to the agent the way the conflict path does, for the same reason.

**`attemptUpdateBranch` stops latching and lying.** Its no-backend and no-trailer guards returned *above* the `defer`, so the PR stayed in `autoMergeInFlight` for the process lifetime. Worse than "silent": nothing wrote `autoMergePermanentSkips`, so `evaluateAutoMerge`'s guard fell back to `.inFlight` — the UI said *"Crow is working on this PR's auto-merge right now."* Forever. The defer now covers every path; suppression is `autoUpdateBranchAttempted`'s job, which returns before any dispatch, so this costs no extra backend calls. A failed `gh pr update-branch` gets bounded retries rather than latching until a head commit that a failed update is precisely what didn't change, and both maps are now pruned (`autoUpdateBranchAttempted` never was).

**The precedence `continue` yields only while auto-merge can still act.** It was unconditional, so once auto-merge spent its one-shot per-head attempt and gave up, *nobody* fixed the branch. The `autoMergeInFlight` disjunct is load-bearing: the key is inserted *before* the async attempt and `applyAutoMerge` runs synchronously right before `applyAutoRebase` in the same poll, so `!contains` alone is already false in the very poll auto-merge dispatched.

## Visibility

`auto_rebase_state` rides `list-sessions-live` beside `auto_merge_state`, rendered as a ⟲ chip (orange waiting / red stuck), inserted **before** the ⛙ auto-merge chip — chronological, and it keeps `autoIco` (last icon) meaning "auto-merge" for every existing #888 assertion. The justification is sharper than auto-merge's: `prStatusJSON` never ships `mergeStateStatus`, so a wedged branch rendered as a **fully green pill**.

## Decisions stated in the code

- **Drafts stay eligible**, as since CROW-318 — a rebase only rewrites the session's own branch and can never merge (`shouldAttemptAutoMerge` keeps its own draft guard). `DRAFT` masked behind-ness *permanently*, so this is where the change bites most.
- **Every open GitLab MR becomes a candidate** — that backend never populates `mergeStateStatus`. Wanted (MRs do fall behind, nothing else handles them), bounded to one probe per head, but a deliberate scope expansion rather than a side effect.
- **`shouldUpdateBranchBeforeMerge` is NOT widened.** So a `crow:merge` PR that is `BLOCKED`-but-behind now routes to auto-rebase (force-push) rather than auto-merge (merge commit). On a repo with "dismiss stale reviews on push", that dismisses approvals. Flagging it rather than letting someone discover it via a dismissed review.

Out of scope: `gh pr update-branch --rebase`, and inverting precedence wholesale.

## Testing

`make test` green except CrowDaemon's 2 pre-existing `app.js`-source-shape failures (`refreshTerminalsRebindsTheActiveTerminalToTheFreshRow`, `bootCatchResetsTheHistory`) — verified identical with `app.js` stashed to `HEAD`. `make parity` passes at 12 events; `make docs` regenerated and committed.

⚠️ **`CrowEngine` is not in CI's Linux allow-list and `row.test.js` is omitted from `test:ci`** (10 pre-existing CROW-802 failures), so most of the new tests won't run in CI. Verified locally:
- `CrowGit` 27 tests (the CI-covered ones): `alreadyUpToDate`, `behindBase` × 4 incl. the remote-vs-HEAD distinction and `.unknown`-not-`.upToDate` on a bad ref.
- `CrowEngine` 664 tests. `shouldAttemptAutoRebase` against BLOCKED/DIRTY/DRAFT/UNKNOWN/UNSTABLE/HAS_HOOKS/CLEAN; `rejectsUnknownState` **inverted** with a comment explaining why rather than deleted; escalation threshold pinned to the backoff curve; verdict `permanent ⟺ blocked` invariant. New `IssueTrackerBehindPRHandoffTests` drives real dispatch bookkeeping for the precedence hand-off, both `autoMergeInFlight` leak paths, pruning, and the up-to-date re-check.
- `row.test.js`: **57 passed / 10 failed**, baseline was **43 / 10** — 14 new assertions, same 10 pre-existing failures. (Found and fixed a real bug doing this: the harness's own `T.prStatusInline` shim dropped the new 4th argument.)

Not yet exercised end-to-end against a live BLOCKED-and-behind PR — that's the one acceptance criterion no test reaches.


dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Git operations use `Process` argument arrays (no shell interpolation), so branch/base ref names from the provider are not a command-injection vector (`GitManager.swift:335-341`).
- Auto-rebase only force-pushes after `prHasCrowAuthoredCommit` and excludes review/manager sessions (`IssueTracker.swift:3836-3838`, `4035-4038`).
- `rebaseOntoBase` refuses dirty worktrees, verifies HEAD branch, fast-forwards only when strictly behind remote, and pushes with `--force-with-lease` (`GitManager.swift:171-289`).
- UI surfaces daemon `message` via `textContent` / `title` attributes, not `innerHTML` (`app.js:1192-1196`, `2748-2756`); git stderr in verdicts is truncated to 200 chars (`IssueTracker.swift:358`).

**Concerns:**
- No material security issues identified. Residual risk is operational: force-push on Crow-authored branches can dismiss approvals on repos with “dismiss stale reviews on push” — documented in the PR and an accepted tradeoff vs. `shouldUpdateBranchBeforeMerge` staying `BEHIND`-only.

### Code Quality
- **Yellow — re-arm signal may miss pure base drift:** `autoRebaseUpToDateHeads` re-arms the per-head latch only when `mergeStateStatus` changes (`IssueTracker.swift:3912-3915`). GitHub keeps `mergeStateStatus` at `BLOCKED` while a review is pending even when the base moves ahead, so a PR probed while `BLOCKED` and git-up-to-date can burn its one latch, then miss a later base drift with no `headRefOid` change. The test at `IssueTrackerBehindPRHandoffTests.swift:204-219` covers `BLOCKED → BEHIND` flips but not stable-`BLOCKED` base drift. Consider a secondary re-arm keyed on fetched `origin/<base>` SHA or a bounded periodic re-probe.
- **Green:** `behindBase` correctly compares remote refs (not `HEAD`), returns `.unknown` on failure rather than `.upToDate`, and `alreadyUpToDate` prevents phantom notifications (`GitManager.swift:100-135`, `232-252`).
- **Green:** `attemptUpdateBranch` in-flight leak fix and bounded retry mirror the rebase path (`IssueTracker.swift:3663-3718`); precedence hand-off tests exercise real dispatch bookkeeping (`IssueTrackerBehindPRHandoffTests.swift`).
- **Green:** `AutoRebaseState` UI chip, RPC payload, and `autoRebaseStuck` notification close the visibility gap for wedged branches (`app.js:1982-2006`, `RPCHandlers.swift:977-983`, `CrowDaemon.swift:685-688`).
- **Green:** `swift build` succeeds; `CrowGit` (27 tests) and `IssueTrackerAutoRebaseTests` (33 tests) pass locally.

### Summary Table

---

dhilgaertner:
Thanks — the Yellow finding is correct, and tracing it, it's worse than "may miss": that re-arm could **never** fire for the case CROW-944 is about.

`mergeStateStatus` is single-valued, which is the whole premise of the ticket — `BLOCKED` outranks `BEHIND`. So when the base drifts under a review-pending PR the field doesn't flip to `BEHIND`, it just *stays* `BLOCKED`. `headRefOid` doesn't move either. A status-change re-arm therefore only fires when BLOCKED finally clears, i.e. once the PR is approved — precisely the serialization this ticket exists to remove. My fix only ever helped PRs that were already behind the first time they were seen, which is the minority path.

Fixed in 25cb27b.

I went with your second suggestion (bounded periodic re-probe) rather than the `origin/<base>` SHA, because the SHA is circular here: you need a `git fetch` to learn the base SHA, and the fetch is the cost the latch exists to avoid. `PRRecord` carries `baseRefName` but no base OID, so there's no free signal from the poll either.

`AutoRebaseUpToDateHead` now records a `recheckAt` alongside the status, and `shouldRecheckUpToDateHead` re-arms on either signal:

```swift
nonisolated static func shouldRecheckUpToDateHead(
    _ seen: AutoRebaseUpToDateHead, currentStatus: String, now: Date
) -> Bool {
    seen.mergeStateStatus != currentStatus || now >= seen.recheckAt
}
```

Interval is 900s, matching `autoRebaseDeferralMaxDelay` — four local `git fetch`es an hour per open PR with a worktree, comfortably inside how long CI takes so the detection lag never becomes the critical path, and still bounded enough that a persistent git/GitHub disagreement re-probes on the interval rather than every poll (the hot-loop the status gate was guarding against).

Two tests: `aStableBlockedHeadIsRecheckedOnTheIntervalSoBaseDriftIsCaught` for the exact scenario you described, and `recheckPolicyIsGatedByBothStatusAndClock` for the pure policy. I confirmed both **fail** with the clock condition removed, so they're real regression cover rather than tests that pass either way. `docs/automation.md` updated — the old prose described the insufficient behavior.

`make test` is green except CrowDaemon's 2 pre-existing `app.js`-source-shape failures (verified identical with `app.js` stashed to `HEAD`); `make parity` passes.

dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- `GitManager.run` passes argv arrays to `Process` (no shell interpolation), and `behindBase`/`rebaseOntoBase` inherit the existing `--force-with-lease` + dirty-tree guards.
- `BaseBehindState.unknown` correctly falls through to a full attempt instead of silently skipping work — avoids a class of "optimization drops real rebases" bugs.
- `AutoRebaseVerdict.gaveUp` truncates git stderr to 200 chars before it reaches RPC/UI; `prStatusPart` renders via `textContent`/`title`, not `innerHTML` (`app.js:2748-2756`, `IssueTracker.swift:387-396`).
- Crow-authorship gate still applies before any rewrite; review/manager sessions remain excluded.

**Concerns:**
- None blocking. Operational note (not a vulnerability): `crow:merge` PRs that are `BLOCKED` but git-behind now route through auto-rebase (force-push) rather than `gh pr update-branch`, which can dismiss approvals on repos with "dismiss stale reviews on push." This is documented and intentional — worth awareness for operators.

### Code Quality
- Root-cause fix is sound: widen `shouldAttemptAutoRebase` to a cheap candidate filter, then let `GitManager.behindBase` answer the real question using **remote** refs (not `HEAD`), with `RebaseOutcome.alreadyUpToDate` preventing phantom notifications.
- The `autoRebaseUpToDateHeads` + 900s recheck interval correctly handles base drift invisible in both `headRefOid` and `mergeStateStatus` (`IssueTracker.swift:3938-3953`) — without this, widening the candidate set would be worse than the original bug.
- `attemptUpdateBranch` defer fix (`IssueTracker.swift:3706-3751`) closes a real `autoMergeInFlight` leak; bounded retry on failed `update-branch` mirrors the rebase path.
- Precedence hand-off only yields while auto-merge can still act (`IssueTracker.swift:3929-3935`), with thorough coverage in `IssueTrackerBehindPRHandoffTests`.
- UI/RPC wiring (`AutoRebaseState`, `list-sessions-live`, ⟲ chip) mirrors `AutoMergeState` conventions; stuck escalation + `autoRebaseStuck` notification address the prior silent 15-minute log loop.
- Tests are extensive: `CrowGit` behindBase/alreadyUpToDate (13/13), `IssueTrackerAutoRebaseTests` (33/33), `IssueTrackerBehindPRHandoffTests` (14/14), `AutoRebaseStatePayloadTests` (5/5). `swift build` clean.

### Summary Table

| # | Finding | Color |
|---|---------|-------|
| 1 | Remote-ref `behindBase` probe + three-valued result prevents both false skips and false positives | Green |
| 2 | `autoRebaseUpToDateHeads` time-based re-arm closes the widened-candidate latch trap | Green |
| 3 | `attemptUpdateBranch` in-flight leak + bounded failure retries | Green |
| 4 | `crow:merge` + `BLOCKED`-but-behind may force-push (dismiss reviews on some repos) — documented trade-off | Green |
| 5 | New `CrowEngine`/web tests won't run in Linux CI (pre-existing CROW-802); verified locally | Green |

---

Issue #902, opened 2026-07-28
Wire review support for Antigravity (Phase B): dispatch /crow-review-pr via inline SKILL body


dgershman:
Antigravity currently returns `nil` for `.review` sessions (`AntigravityAgent.swift:108-115`), so any review session assigned to it can't launch — Crow instead echoes the "agent "Antigravity" cannot launch a review session" warning into the terminal (`SessionService.swift:985-1001`). This was a deliberate Phase-A deferral (#860/#862), not a hard blocker. Wire it up.

Antigravity has **no Crow slash-command engine**, so it takes the same path as Cursor/Codex/OpenCode (**inline SKILL body**), not Claude's slash-command path.

## How review dispatch works today (for reference)
- `prepareReviewClone` writes `.crow-review-prompt.md` into the review clone (`SessionService.swift:2793-2795`); its content is built by `buildReviewPrompt` (`:3074-3096`).
  - **Claude** gets the terse one-liner `/crow-review-pr <URL>`, resolved by its slash/SKILL engine against the copied `.claude/skills/crow-review-pr/SKILL.md` (`:2804-2808`).
  - **Cursor / OpenCode / Codex** have no slash engine, so they get the **whole SKILL body inlined** into the prompt file (``→PR URL substituted), self-contained (`:3076-3089`).
- The agent's `autoLaunchCommand(.review)` reads that file via `$(cat .crow-review-prompt.md)` on first launch, then resumes on restart.

## Changes to make
1. **`AntigravityAgent.autoLaunchCommand(.review)`** (`AntigravityAgent.swift:108-115`): replace the `nil` return with a real dispatch that mirrors its own `.job` path (`:100-107`):
   - first launch (`!session.reviewPromptDispatched`): `agy <autoArgs> -p "$(cat .crow-review-prompt.md)"` (quote the path like the job path does);
   - restart: `agy <autoArgs> -c` (continue most-recent conversation).
2. **`buildReviewPrompt`** (`SessionService.swift:3076`): add `.antigravity` to the `.cursor, .openCode, .codex` inline-SKILL branch so Antigravity gets the self-contained SKILL body via `cursorReviewPrompt(skillBody:prURL:agentKind:)` — NOT the bare slash line. `agentKind` is already threaded so the posted-review footer names Antigravity correctly.
[...]

PR #861, opened 2026-07-26
CrowGrok adapter — Grok Build core loop (#859)

dgershman:
Closes #859

Adds a `CrowGrok` coding-agent adapter for **Grok Build** (`xai-org/grok-build`, binary `grok`) — xAI's official Rust coding-agent harness — per [ADR 0014](https://github.com/corveil/crow/blob/main/docs/adr/0014-pluggable-coding-agent-adapter.md) / [0015](https://github.com/corveil/crow/blob/main/docs/adr/0015-harness-capability-tiers.md). Grok's hook schema is byte-compatible with Claude's, so this reaches the **good tier**: per-worktree, per-session-UUID hook scope (like Claude/Cursor), not the `cwd`-match fallback Codex/OpenCode use.

## What's in it

**New package `Packages/CrowGrok/`** (mirrors `CrowClaude`'s hook writer + `CrowOpenCode`'s lean shape):

| File | Role |
|---|---|
| `GrokAgent` | `CodingAgent` conformance. Token `grok` + `fallbackCandidates`/`defaults.binaries.grok` for the `superagent-ai/grok-cli` collision. `supportsRemoteControl=true` (faked via `crow send`; native ACP deferred). `/rename` passthrough. |
| `GrokHookConfigWriter` | Per-worktree `.grok/hooks/crow.json`, `hook-event --session <UUID>` — Claude-compatible `{"hooks":{…}}`, Crow-owned file. |
| `GrokSignalSource` | Event→state over Grok's verified vocabulary (real `Stop`, `Notification`, `SessionStart/End`, `Pre/PostToolUse`, `UserPromptSubmit`, `StopFailure`). |
| `GrokLauncher` / `GrokLaunchArgs` | Run-then-continue prompt injection (`grok -p …; grok -c`), bounded `.job`-only auto-permission (`--permission-mode auto` + hard `--deny`, **never `--yolo`**), resume `-c`. |
| `GrokTrustSeeder` | Seeds `~/.grok/trusted_folders.toml` (`[folders."…"] trusted = true`), mirroring `CodexTrustSeeder`. |

**Wiring:** `AgentKind.grok` (CrowCore); gated registration in `CrowDaemon.registerAgents`; `SessionService` `.grok` trust-seed on all three launch paths (gated `!= .review`) + committed-`.grok/` strip in `prepareReviewClone`.

**Docs:** verified Grok column added to `docs/agent-harness-matrix.md` with the maturity/churn caveat and the `grok` collision recorded in the version-pinned table.

[...]

PR #955, opened 2026-08-10
Launch Cursor review sessions pre-trusted (CROW-954)

### LINKED ISSUE #954: Cursor review sessions block on the workspace-trust prompt instead of launching pre-trusted
Choosing Cursor as the review agent leaves review sessions stuck on Cursor's folder-trust prompt instead of starting work:

```
⚠ Workspace Trust Required
Cursor Agent can execute code and access files in this directory.
Do you trust the contents of this directory?

  /Users/…/crow-reviews/corveil-pr-2152

▶ [a] Trust this workspace
  [q] Quit
```

The session sits there until a human presses `a`, which defeats unattended review dispatch.

## Cause

`CursorLaunchArgs.launchSuffix` is called with `seedTrust: session.kind != .review`, so `.review` launches omit `--trust`. That carve-out is from CROW-890 (review, Red 1), justified as: a review tree is a `gh` clone at the PR author's head, so let Cursor's folder-trust dialog serve as the human gate.

Both halves of that rationale fail in practice:

1. **The dialog gates nothing actionable.** Review launches already carry `--force --approve-mcps` (`reviewAutoPermissionMode` defaults on), so the only useful answer at the prompt is "trust" — after which every tool call runs unapproved anyway. It is a keypress, not a decision.
2. **`--force` does not suppress the prompt.** CROW-890's own comment flagged this as "unverified"; it is now confirmed on `agent 2026.08.04` — a review session launched with `--force --approve-mcps` still stops on "Workspace Trust Required".

So the carve-out bought no security while costing exactly the unattended dispatch the trust seed exists to provide.

Cursor is also the outlier here: `SessionService.shouldSeedFolderTrust` returns `true` for Claude on **every** kind, `.review` included, so Claude-driven reviews have always launched pre-trusted.

## Expected

A Cursor review session comes up already trusted **and** in Run Everything mode — no prompt, no keypress. ("Run Everything" is `--force`; `agent --help` documents `--yolo` as a pure alias for it, so reviews are already in that mode.)

## Notes

- `--trust` is documented as ["headless mode only"](https://cursor.com/docs/cli/reference/parameters), which would make the fix a no-op. That doc is stale: verified on `agent 2026.08.04` that running `agent --trust` under a pty with **no** `--print` writes `~/.cursor/projects/<slug>/.workspace-trusted` containing `"trustMethod": "cli-flag"` — the same saved decision the dialog records on accept.
- Dropping the carve-out means the review clone's committed `.cursor/` (`hooks.json`, `mcp.json`) is no longer gated by a dialog, so `stripCursorConfigFromReviewClone` becomes the sole defense and must be promoted from a creation-time + handoff strip to run on **every** launch path via `prepareWorktreeForAgentLaunch` — the review skill's `gh pr checkout` can restore a hostile `.cursor/` from the PR head, and a warm `crowd` restart reopens the clone through neither existing arm. Same strip-not-trust posture Grok (#861) and Antigravity (#902) reviews already use.
- Codex and Grok should **keep** their `!= .review` trust guards: they write durable global trust records (`~/.codex/config.toml`, `~/.grok/trusted_folders.toml`), a materially larger blast radius than Cursor's per-workspace marker.

dhilgaertner:
Closes #954.

Choosing Cursor as the review agent left review sessions parked on Cursor's "Workspace Trust Required" prompt instead of starting work, defeating unattended review dispatch.

## Cause

`CursorLaunchArgs.launchSuffix` was called with `seedTrust: session.kind != .review` — a CROW-890 carve-out (review, Red 1) justified as "review falls back to Cursor's folder-trust dialog as its human gate." Both halves of that fail in practice:

1. **The dialog gated nothing actionable.** Review launches already carry `--force --approve-mcps` (`reviewAutoPermissionMode` defaults on), so the only useful answer at the prompt is "trust" — after which every tool call runs unapproved anyway. It was a keypress, not a decision.
2. **`--force` does not suppress the prompt.** CROW-890's own comment flagged this as *"unverified"*. Now settled: on `agent 2026.08.04`, a review session launched with `--force --approve-mcps` still stops on "Workspace Trust Required" and waits.

So the carve-out bought no security while costing exactly the unattended dispatch the trust seed exists to provide. Cursor was also the outlier — `SessionService.shouldSeedFolderTrust` returns `true` for Claude on **every** kind, `.review` included, so Claude reviews have always launched pre-trusted.

## Change

- **`CursorAgent` seeds `--trust` on every session kind**, `.review` included. Reviews now launch as `'agent' --trust --force --approve-mcps \"$(cat …/.crow-review-prompt.md)\"` — already trusted, already in Run Everything mode.
- **Dropped Cursor's kind-aware `launchCommand(…sessionKind:)` override** — it existed solely to withhold the seed. Nothing else implements it, so the protocol default now serves every caller; the protocol requirement is kept as the seam a future harness would need.
- **Promoted `stripCursorConfigFromReviewClone` to every launch path** via a new `shouldStripCursorReviewClone` gate in `prepareWorktreeForAgentLaunch`.

That last one is the load-bearing part. The clone's real defense was never the dialog — it's the strip that removes committed `.cursor/hooks.json` / `mcp.json` before Cursor opens the tree. It previously ran only at clone creation and on handoff. With no dialog behind it, it must also re-fire after the review skill's `gh pr checkout` (or a head-advancing re-review) restores a hostile `.cursor/` from the PR head, and after a warm `crowd` restart or `crow send` reopens the clone through neither existing arm. This is the same **strip-not-trust** posture Grok (#861) and Antigravity (#902) reviews already rely on.

**Codex and Grok deliberately keep their `!= .review` trust guards.** They write durable *global* trust records (`~/.codex/config.toml`, `~/.grok/trusted_folders.toml`) that outlive the session — a materially larger blast radius than Cursor's per-workspace marker under `~/.cursor/projects/`. Nothing here argues for touching them.

## On the security tradeoff

This is a real reduction in defense depth, stated plainly: a hostile PR's `.cursor/` now has one mechanism standing in its way instead of two. The judgment is that the second mechanism was not actually a mechanism — a prompt whose only answer is "yes", in front of a session already running `--force`, is not a gate. The strip is now covered on every launch path and by tests that fail if the gate is removed.

## Verification

`--trust` is documented as ["headless mode only"](https://cursor.com/docs/cli/reference/parameters), which would have made this fix a silent no-op. **That doc is stale.** Verified empirically on `agent 2026.08.04` — running `agent --trust` under a pty in a fresh repo, with no `--print`, writes `~/.cursor/projects/<slug>/.workspace-trusted`:

```json
{
  "trustedAt": "2026-08-10T15:47:02.341Z",
  "workspacePath": "/private/tmp/…/repo",
  "trustMethod": "cli-flag"
}
```

That is the same saved decision the dialog records on accept. (Test artifacts were cleaned up afterward.)

Also confirmed from `agent --help` that **"Run Everything" is `--force`** — `--yolo  Alias for --force (Run Everything)` — so reviews were already in that mode; only trust was missing.

## Tests

- CrowCore **544** pass, CrowCursor **67** pass, CrowEngine **685** pass.
- Rewrote the CrowCursor assertions that pinned the old carve-out; `autoLaunchCommandReviewSeedsTrustAndRunEverything` is the regression pin for the reported symptom (asserts both `--trust` and `--force --approve-mcps` under the shipped default).
- Added 5 CrowEngine tests for the new strip gate, including `prepareStripsCursorConfigWhenGateFires` — delete the new arm of `prepareWorktreeForAgentLaunch` and only that test fails.
- CrowDaemon has 2 failures (`WebTerminalAssetTests`, `WebNotificationCenterTests`); verified identical on a clean `origin/main` worktree, so **pre-existing and unrelated**.

## Docs

`docs/agent-harness-matrix.md` (grid row, auto-permission bullet, trust-seed bullet, version re-check row) and an ADR 0015 amendment marking the #890 `.review` carve-out superseded.

## Note

The fix takes effect only after a rebuild + `crowd` restart; review sessions launched by the old binary still need one keypress.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/Agent/CodingAgent.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Agent/CodingAgent.swift
@@ -132,13 +132,19 @@ public protocol CodingAgent: Sendable {
         prompt: String
     ) async throws -> String
 
-    /// Kind-aware variant of `launchCommand`. Agents whose launch depends on the
-    /// session kind override this — e.g. Cursor withholds its workspace-trust
-    /// seed (`--trust`) from attacker-controlled `.review` handoff clones,
-    /// mirroring the `session.kind != .review` guard on `CodexTrustSeeder`. The
-    /// protocol-extension default ignores `sessionKind` and delegates to the
-    /// three-argument `launchCommand`, so agents with no kind-dependent launch
-    /// (Claude, Codex, OpenCode, Antigravity) need not implement it.
+    /// Kind-aware variant of `launchCommand`, for agents whose handoff launch
+    /// depends on the session kind. The protocol-extension default ignores
+    /// `sessionKind` and delegates to the three-argument `launchCommand`, so an
+    /// agent with no kind-dependent launch need not implement it.
+    ///
+    /// **No agent overrides this today.** Cursor was the only one — it withheld
+    /// its `--trust` workspace-trust seed from `.review` handoff clones — and
+    /// CROW-954 dropped that carve-out (review clones now launch pre-trusted,
+    /// defended by the launch-path `.cursor/` strip instead of a folder-trust
+    /// dialog). The requirement is kept because it is the seam a future harness
+    /// needs to treat a hostile review checkout differently from a `.work`
+    /// worktree; `SessionService.handoffAgent` already calls through it with the
+    /// live `session.kind`, so adding one is a single override.
     func launchCommand(
         sessionID: UUID,
         worktreePath: String,
--- a/Packages/CrowCursor/Sources/CrowCursor/CursorAgent.swift
+++ b/Packages/CrowCursor/Sources/CrowCursor/CursorAgent.swift
@@ -75,20 +75,19 @@ public struct CursorAgent: CodingAgent {
         // unavailable kinds.
         // The `--trust` workspace-trust seed (skips the folder-trust dialog on a
         // fresh worktree — the per-launch analogue of `ClaudeTrustSeeder`,
-        // CROW-890) rides every launch path EXCEPT `.review`: a review working
-        // tree is an attacker-controlled `gh` clone at the PR author's head, so —
-        // mirroring the `session.kind != .review` guard on `CodexTrustSeeder` in
-        // `SessionService` — Crow never auto-trusts it (the intent is that review
-        // falls back to Cursor's folder-trust dialog; whether `--force` still
-        // surfaces that dialog is unverified — see `CursorLaunchArgs.launchSuffix`
-        // — but withholding `--trust` is never worse). The auto-permission
-        // flags (`--force --approve-mcps`) still apply per the caller's opt-in,
-        // including on `.review` (unchanged). See `CursorLaunchArgs` for why
-        // `--sandbox` is left unset (#829). On the non-review paths the seed also
-        // rides `--continue` resume — harmless: the first launch already recorded
-        // the saved trust decision.
+        // CROW-890) rides EVERY launch path, `.review` included as of CROW-954.
+        // The original review carve-out left unattended reviews stuck on
+        // "Workspace Trust Required" while already carrying `--force`, so the
+        // dialog gated nothing a reviewer could act on — and the clone's real
+        // defense is `stripCursorConfigFromReviewClone`, which now re-runs on
+        // every launch via `prepareWorktreeForAgentLaunch`. See
+        // `CursorLaunchArgs.trustSuffix` for the full rationale. The
+        // auto-permission flags (`--force --approve-mcps`) still apply per the
+        // caller's opt-in (unchanged). See `CursorLaunchArgs` for why `--sandbox`
+        // is left unset (#829). The seed also rides `--continue` resume —
+        // harmless: the first launch already recorded the saved trust decision.
         let launchArgs = CursorLaunchArgs.launchSuffix(
-            seedTrust: session.kind != .review,
+            seedTrust: true,
             autoPermissionMode: autoPermissionMode)
 
         switch session.kind {
@@ -162,41 +161,18 @@ public struct CursorAgent: CodingAgent {
         worktreePath: String,
         prompt: String
     ) async throws -> String {
-        // Fail closed (CROW-890 review, Yellow 2): the kindless three-argument
-        // requirement can't prove a worktree is non-`.review`, so it never seeds
-        // trust. It has no production caller — `SessionService.handoffAgent`
-        // always calls the `sessionKind:` overload below — so this can't
-        // under-trust a real launch; it just guarantees a hypothetical future
-        // caller of the bare form gets the safe default rather than reopening the
-        // review-clone hole. The kind-aware overload is the seeding path.
+        // Seeds trust for every kind (CROW-954). The CROW-890 "fail closed"
+        // default — never seed, because a kindless signature can't prove the
+        // worktree isn't `.review` — is moot now that `.review` seeds too, so
+        // Cursor no longer needs the kind-aware overload at all and the
+        // `CodingAgent` protocol default (which drops `sessionKind` and lands
+        // here) is correct for every caller.
         try await launcher.launchCommand(
             sessionID: sessionID,
             worktreePath: worktreePath,
             prompt: prompt,
             binary: findBinary() ?? "agent",
-            seedTrust: false
-        )
-    }
-
-    /// Kind-aware handoff launch. Overrides the `CodingAgent` default so a
-    /// `.review` handoff to Cursor does **not** carry the `--trust` seed — the
-    /// review clone is attacker-controlled (a `gh` checkout at the PR author's
-    /// head), so Crow does not auto-trust it, mirroring the `session.kind !=
-    /// .review` guard on `CodexTrustSeeder` (CROW-890 review, Red 1).
-    /// `SessionService.handoffAgent` calls this with the live `session.kind`;
-    /// every other kind seeds trust as before.
-    public func launchCommand(
-        sessionID: UUID,
-        worktreePath: String,
-        prompt: String,
-        sessionKind: SessionKind
-    ) async throws -> String {
-        try await launcher.launchCommand(
-            sessionID: sessionID,
-            worktreePath: worktreePath,
-            prompt: prompt,
-            binary: findBinary() ?? "agent",
-            seedTrust: sessionKind != .review
+            seedTrust: true
         )
     }
 
--- a/Packages/CrowCursor/Sources/CrowCursor/CursorLaunchArgs.swift
+++ b/Packages/CrowCursor/Sources/CrowCursor/CursorLaunchArgs.swift
@@ -10,8 +10,8 @@ public enum CursorLaunchArgs {
         "'" + s.replacingOccurrences(of: "'", with: "'\\''") + "'"
     }
 
-    /// Workspace-trust seed appended to every Crow-launched Cursor command
-    /// **except `.review`** (auto-launch, Manager, and the handoff one-shot). A
+    /// Workspace-trust seed appended to **every** Crow-launched Cursor command
+    /// (auto-launch, Manager, and the handoff one-shot) — `.review` included. A
     /// leading-space suffix so callers concatenate it onto the shell-quoted
     /// binary path.
     ///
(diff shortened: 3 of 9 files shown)


dgershman (changes requested):
## Code & Security Review

The core argument holds up under scrutiny, and I verified it rather than taking it on faith:

- **Launch-path coverage is real.** All five `prepareWorktreeForAgentLaunch` call sites confirmed (`pasteDeferredLaunch:814`, `launchAgent:1196`, `handoffAgent:1534`, `createManagerTerminal:2056`, `send` RPC at `EngineRouter.swift:1034`), plus `recreateTerminalSurface` routing transitively through re-armed readiness. The gate is synchronous and MainActor-bound, so the strip completes before the launch command is typed into the pane.
- **The protocol claim checks out.** `rg` confirms no other type implements `launchCommand(…sessionKind:)`, so dropping Cursor's override leaves the extension default correct for every caller. The requirement-as-seam is a defensible call.
- **Tests pass locally:** CrowCursor 67/67, CrowEngine strip suite 30/30, CrowEngine builds clean. `prepareStripsCursorConfigWhenGateFires` is a genuine wiring pin, not just a predicate test.
- **The security tradeoff is stated honestly** and the ADR amendment supersedes #890 properly rather than rewriting it.

One issue that this PR's own change in posture creates.

### Security Review

**Strengths:**
- Promoting the strip to every launch path is the right load-bearing move, and the reasoning for *why* creation-time alone is insufficient (the review skill's `gh pr checkout`, head-advancing re-review, warm `crowd` restart, `crow send`) is correct and matches the Grok/Antigravity precedent.
- Correctly scoped: `shouldStripCursorReviewClone` fires only for `.cursor` + `.review`, so a `.review` on another agent doesn't get `.cursor/` hidden from the reviewing agent — consistent with the `.codex`/`.agents` gates.
- Leaving Codex/Grok's `!= .review` guards alone is right; global TOML trust records are a genuinely different blast radius from a per-workspace marker.

**Concerns:**

**Yellow 1 — the sole-defense strip logs its failure at `info`, not `error`.**
`SessionService.swift:3205` — `stripCursorConfigFromReviewClone` swallows a real removal failure into `CrowLog.info`:

```swift
} catch {
    CrowLog.info("[SessionService] Failed to strip .cursor/ from review clone \(clonePath): …")
}
```

The structurally identical Grok/Antigravity strips route through `removeReviewCloneConfig` (`SessionService.swift:3342`), which uses `CrowLog.error` — with the rationale spelled out at `SessionService.swift:3158`: *"the strip is Antigravity's only defense, so a swallowed failure would leave live attacker config in place with nothing to show for it."*

That was tolerable for Cursor before this PR, because a folder-trust dialog stood behind the strip. **This PR removes that backstop and puts the Cursor strip in exactly the position the `.error` rationale describes.** The failure mode is now: `removeItem` fails (permissions, a `.cursor` that's a busy mount, an immutable flag), the attacker's `hooks.json` survives, and the session launches pre-trusted with `--force --approve-mcps` — while the only trace is an `info` line indistinguishable from routine chatter.

The function's own doc comment at `SessionService.swift:3202-3204` already asserts *"this security control must be audible rather than swallowed"* — the code just doesn't do what the comment says. Raising it to `CrowLog.error` is a one-line change and brings all three review-clone strips to one log level.

### Code Quality

**Green 1 — redundant Cursor handoff strip arm.** `SessionService.swift:1580-1583` still calls `shouldStripCursorReviewCloneOnHandoff` + `stripCursorConfigFromReviewClone` explicitly, but line 1534's `prepareWorktreeForAgentLaunch` (called with `target.kind`) now covers that case. Idempotent, so harmless — but the comment at `SessionService.swift:1530-1532` explicitly credits the shared gate for why *"this PR needs no bespoke Antigravity handoff arm"*, so the Cursor arm is now the odd one out. Dropping it (and its predicate) would leave one strip path per agent.

**Green 2 — no `git checkout -- .cursor` restore-before-pull arm.** `prepareReviewClone` restores stripped paths before `git pull` for Grok (`SessionService.swift:3443`) and Antigravity (`:3460`), because the strip lands as an unstaged deletion of tracked files and a non-fast-forward pull that touches them refuses. Cursor's creation-time strip (`:3536`) has the same property with no matching arm. This is **pre-existing** — the `.cursor` strip predates this PR — so it isn't a blocker here. But this PR multiplies the paths that create that deletion (every launch, including every warm `crowd` restart), so it's worth a parity arm or an explicit note on why Cursor doesn't need one.

### Notes on things I checked and found fine

- The mid-session window where the review skill's `gh pr checkout` restores `.cursor/` into an already-running trusted session is **not** made worse by this PR — the pre-PR dialog only ever gated the initial open, so a reviewer who pressed "trust" landed in the identical state. Genuinely pre-existing and shared with Grok/Antigravity.
- `--trust` ordering (`--trust --force --approve-mcps`, ahead of the prompt argument) is pinned by `autoLaunchCommandReviewSeedsTrustAndRunEverything`.
- Keeping `seedTrust:` as a parameter with no production `false` caller, pinned by `launchCommandWithholdsTrustSeedWhenSeedTrustOff`, is a reasonable way to keep the posture one flip away.

### Summary Table

The Yellow is a one-line log-level change; the design decision at the heart of this PR is sound and well-argued. Fix the `.info` → `.error` and this is good to land.

---

dhilgaertner:
Thanks — the Yellow was a fair catch and is fixed in `ecc6b26`, along with Green 1. Green 2 I'm pushing back on; details below.

## Yellow 1 — strip failure now audible ✅

Fixed, but as a deletion rather than a one-word swap. Rather than changing `CrowLog.info` → `CrowLog.error` in place, `stripCursorConfigFromReviewClone` now delegates to `removeReviewCloneConfig` — the helper the Grok and Antigravity strips already route through:

```swift
nonisolated static func stripCursorConfigFromReviewClone(clonePath: String) {
    removeReviewCloneConfig(
        (clonePath as NSString).appendingPathComponent(".cursor"),
        label: ".cursor/", clonePath: clonePath)
}
```

That drops the duplicated `do`/`catch`/`NSFileNoSuchFileError` block entirely, so all three review-clone strips now share one implementation and one log level instead of three copies that can drift apart on exactly this axis again. Your diagnosis of *why* it matters now is precisely right and I've recorded it on the function: an `info` line was defensible while the folder-trust dialog stood behind the strip, and this PR is what removed that backstop.

## Green 1 — redundant handoff arm retired ✅

Agreed, and confirmed strictly redundant before removing: line 1534's `prepareWorktreeForAgentLaunch` is unconditional on that path and is called with `target.kind`, so `shouldStripCursorReviewClone(.cursor, .review)` already covers every case the explicit arm did. Dropped the arm and the `shouldStripCursorReviewCloneOnHandoff` predicate; `syncCursorMCPBridge()` stays.

The predicate's three unit tests went with it. `shouldStripCursorReviewClone`'s tests already assert the identical agent × session-kind matrix and `prepareStripsCursorConfigWhenGateFires` pins the wiring — and duplicating one gate across two predicates is what allowed #829 round 11's divergence, so re-creating that pairing to preserve test count seemed like the wrong trade. I left a note at the old `MARK` recording where the coverage moved rather than deleting silently.

## Green 2 — respectfully, this one's incorrect ❌

The Cursor restore-before-pull arm you report as missing already exists, at `SessionService.swift:3434-3436`:

```swift
if reviewAgentKind == .cursor {
    _ = try? await runShellAsync(env: env, args: ["git", "-C", clonePath, "checkout", "--", ".cursor"])
}
```

All four agents have one — `rg 'checkout", "--"'` over `SessionService.swift` returns exactly four hits: `.codex:3429`, `.cursor:3435`, `.grok:3445`, `.antigravity:3462`. Cursor's arm predates this PR (#829 review round 9) and sits immediately above the Grok one you cited as the contrast case. So there's no parity gap to close and nothing to document. Happy to be shown otherwise if you were pointing at a different property than the `git checkout --` restore.

## Unrelated

Not from this PR, but the push surfaced a Dependabot advisory on the **default branch** (1 moderate) — worth a separate look.

---

CrowEngine: 682/682 pass locally. Re-requesting review.

dgershman (approved):
## Code & Security Review

Verified the load-bearing claims rather than taking them on trust. Build clean; **CrowCore 544**, **CrowCursor 67**, **CrowEngine 682** all pass on this checkout.

### Critical Issues

None.

### Security Review

**The central tradeoff is real, stated plainly, and I agree with the call.** Trading a folder-trust dialog for a launch-path strip only holds if the strip genuinely covers every path. It does — I checked rather than trusting the comment:

| Path | Routes through `prepareWorktreeForAgentLaunch` |
|---|---|
| `pasteDeferredLaunch` | ✅ `SessionService.swift:814` |
| `launchAgent` (warm `crowd` restart, `crow launch-agent`) | ✅ `:1196` |
| `handoffAgent` (with `target.kind`) | ✅ `:1536` |
| `createManagerTerminal` | ✅ `:2052` |
| `send` RPC | ✅ `EngineRouter.swift:1034` |
| `crow recreate-terminal` | ✅ indirect — re-arms `autoLaunchTerminals` (`:528`) → `launchAgent` |

`recreate-terminal` was the one I expected to be missed, since it isn't in the enumerated list; it's covered because it re-arms rather than sending directly. The `send` arm is correctly gated on `commandLaunchesToken(text, "agent")`, so a Cursor resume strips while a plain `crow send "yes"` doesn't.

**Counterfactual confirmed.** Deleted the `shouldStripCursorReviewClone` arm of `prepareWorktreeForAgentLaunch` and re-ran CrowEngine: exactly one failure, `prepareStripsCursorConfigWhenGateFires` — as claimed, no more and no less.

**Strengths:**
- Strip precedes `writeHookConfig` on every path, so Crow's clean `hooks.json` lands only after the hostile layer is gone.
- Routing the strip through `removeReviewCloneConfig` upgrades a failed removal from `CrowLog.info` to `CrowLog.error` — right call now that it's the sole defense; a silent failure would leave live attacker config with nothing to show for it.
- `shouldStripCursorReviewCloneOnHandoff` is fully retired — no lingering references in source or tests. One strip path per agent is a genuine simplification, and the handoff hazard from #829 r10 stays covered because the shared gate is called with `target.kind`.
- Codex/Grok `!= .review` guards correctly left alone.

**Concerns:** none blocking. Green notes below.

### Code Quality

Green (consider — none block approval):

1. **`docs/agent-harness-matrix.md:469`** — the *Claude* trust-seed bullet still reads "Runs at **four** call sites: `SessionService.launchAgent`, `handoffAgent`, and the two Manager paths." `seedTrustIfNeeded` is now reached solely via `prepareWorktreeForAgentLaunch` (`SessionService.swift:984`), which has five call sites — the enumeration omits `pasteDeferredLaunch` and the `send` RPC. Pre-existing drift, not introduced here, but it's precisely the hand-maintained-count failure `SessionService.swift:962` warns about ("went stale four rounds running"), sitting one bullet above the section this PR rewrites. Cheap to fix while you're in the file.

2. **Residual intra-session window.** The strip fires at *launch*; the `.cursor/` that this very review skill's `gh pr checkout` restores is removed only on the *next* launch. With the dialog gone that window is unattended from t=0. Not a regression against the fleet — Grok and Antigravity have exactly this posture, and Claude has always trusted `.review` — and the comments describe it accurately (the reopen is what's covered). Worth noting that if it's ever closed, closing it in the review skill (re-strip after checkout) fixes all four harnesses at once rather than one.

3. **ADR blast-radius distinction is thinner than stated.** ADR 0015 justifies leaving Codex/Grok alone because they write "durable *global* trust records… a materially larger blast radius than Cursor's per-workspace marker under `~/.cursor/projects/`." Both persist trust for that path beyond the session; the difference is one file per workspace vs. a shared TOML, not scope. The decision to leave Codex/Grok untouched is right regardless (conservative, out of scope), but the stated reason is the kind a future PR might lean on to drop *their* guards.

4. **`existentialDispatchSeedsTrustForReviewHandoff` no longer discriminates.** With zero overrides of `launchCommand(…sessionKind:)`, deleting the protocol *requirement* and leaving only the extension default keeps the suite green — that's the exact property the old `existentialDispatchReachesCursorOverride` pinned. The PR documents keeping the requirement as a seam, which is the right call; just noting nothing enforces it now.

5. **Test-count nit:** PR body says CrowEngine 685; I measured 682 (all passing). CrowCore 544 and CrowCursor 67 match exactly.

### Summary Table

The security reasoning is the part that could have gone wrong, and it holds up: a prompt whose only answer is "trust", in front of a session already running `--force --approve-mcps`, was not a gate, and the empirical `--force`-doesn't-suppress-it finding settles the question #890 left open. The compensating control is real, covers every launch path, and is pinned by a test that fails when and only when the gate is removed.

dhilgaertner:
Merged as `3626e63` (squash). Thanks for the thorough pass — the counterfactual check on `prepareStripsCursorConfigWhenGateFires`, and catching `recreate-terminal`'s indirect routing, were exactly the right things to verify on this one.

The five Greens are filed as follow-ups rather than dropped:

- **#959** — docs accuracy: the stale "four call sites" count in `agent-harness-matrix.md` (Green 1) and ADR 0015's blast-radius rationale (Green 3). Both prose-only; the second matters mainly because the stated reason is the kind a future PR could lean on to drop Codex/Grok's guards.
- **#960** — the residual intra-session window (Green 2), scoped as a review-skill fix so it lands once for all four harnesses rather than four times.
- **#961** — nothing enforces the `launchCommand(…sessionKind:)` protocol requirement now that no agent overrides it (Green 4). Fair point: the seam this PR argues for keeping is currently removable with a green suite.

Green 5 (test count) was a PR-body typo — 682 is correct, and your three numbers match what I measured.

PR #962, opened 2026-08-10
Raise the WebSocket frame ceiling to 1 MiB (CROW-956)

### LINKED ISSUE #956: Settings save fails with "rpc: connection closed" once config exceeds the 16 KB default WebSocket frame size
## Summary

`crowd` builds its Hummingbird server without a WebSocket configuration, so `/rpc` and the terminal socket inherit hummingbird-websocket's default `maxFrameSize` of **16384 bytes** (`1 << 14`). Any single inbound frame above that is rejected by the NIO frame decoder, which closes the connection with close code **1009 (message too big)**.

The web client surfaces this as `Save failed: rpc: connection closed` in Settings — the daemon never crashes, the socket just dies underneath the in-flight call.

`Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift:501`

```swift
let app = Application(
    router: httpRouter,
    server: .http1WebSocketUpgrade(webSocketRouter: wsRouter),
    configuration: .init(
        address: .hostname(options.host, port: options.httpPort),
        serverName: "crowd"))
```

Both handlers already declare a message-level ceiling of `1 << 20` — `RPCWebSocketHandler.swift:89` and `TerminalWebSocket.swift:78` — but neither is ever reached: the frame decoder trips two layers earlier, 64× lower.

## Impact

Settings saves are a whole-blob replace: `set-config` sends the entire `AppConfig` on every save. Once a user's `config.json` passes ~8 KB (it is JSON-string-escaped into the `config` param, so it roughly doubles on the wire), **every** Settings save fails, no matter which field was edited. Scheduled jobs are the usual cause — a handful of jobs with realistic prompts is tens of KB on its own.

Observed on a config of 52.6 KB, 48.9 KB of it `jobs` across 6 jobs. The resulting frame is ~105 KB, ~6× the limit.

The failure is silent and misattributed: the client reconnects a second later, so the UI looks healthy, and the only symptom is a save that will not stick. There is no daemon-side log line.

`TerminalWebSocket` shares the configuration, so pasting >16 KB into a terminal over the web UI drops that socket the same way.

The Unix-socket path is unaffected — the CLI and `crow` verbs bypass the WebSocket entirely, which is why `crow workspace edit` works while the Settings tab does not.

## Reproduction

Against a running daemon on `127.0.0.1:8787`, sending a JSON-RPC frame of increasing size:

| request size | result |
|---|---|
| 16,370 B | normal JSON-RPC reply |
| 16,450 B | `CLOSED code=1009` |

End to end, `get-config` followed by `set-config` echoing the same blob back:

```
open
got config len 51708
CLOSED code= 1009
```

Handler-side there is no defect: the same 51,708-char config round-trips through `set-config` cleanly over the Unix socket and persists correctly.

## Proposed fix

Line the frame ceiling up with the `maxSize` both handlers already declare, so the message-level guard becomes the one that actually enforces:

```swift
server: .http1WebSocketUpgrade(
    webSocketRouter: wsRouter,
    configuration: .init(ws: .init(maxFrameSize: 1 << 20))),
```

## Notes for whoever picks this up

- A regression test should assert an inbound `/rpc` frame well above 16 KB gets a JSON-RPC reply rather than a close — the current tests all sit under the default and pass either way.
- Worth deciding whether a frame that *does* exceed `1 << 20` should still kill the connection, or whether the reader should answer with a JSON-RPC error and stay up. Today's behaviour gives the client no way to tell "too big" from "daemon died", which is what made this hard to diagnose.
- The client could also stop sending the whole blob: `set-config` is a full replace, while the granular RPCs (`workspace-*`, `defaults-set`, `agents-set`) already exist and would keep Settings saves small regardless of config size.

dhilgaertner:
Closes #956

## The bug

`crowd` built its upgrade channel with no WebSocket configuration, so NIO's frame decoder sat at the default `maxFrameSize` of 16 KiB and answered anything larger with close code **1009**.

Both handlers already declared `inbound.messages(maxSize: 1 << 20)`, but that ceiling was **unreachable from a browser**, for a reason worth stating because it explains why the declared limit looked fine: `WSCore` measures a message against `maxSize` only while appending *continuation* frames — the first frame is never compared to it. A browser sends one unfragmented frame per `ws.send()`, so `maxFrameSize` was always the only ceiling that applied.

Result: `set-config` (a whole-blob replace, JSON-string-escaped and so roughly doubled on the wire) failed for **every** field once `config.json` passed ~8 KB, and a >16 KB paste into the web terminal dropped the socket and lost the paste. Both failed silently — the client reconnected a second later, so the UI looked healthy and the only symptom was a save that would not stick.

## The fix

One shared constant, `CrowDaemon.maxWebSocketFrameSize` (1 MiB), feeding the channel *and* both handlers — replacing three separate magic numbers. It matches `SocketServer.maxMessageSize`, so a request's fate no longer depends on whether it arrived over the Unix socket or the WebSocket, which is why `crow workspace edit` worked while the Settings tab did not.

The call site moves to the non-deprecated `HTTP1WebSocketUpgradeChannel.Configuration` overload; leaving `http1:` at its defaults reproduces exactly what the no-configuration form built, so `maxFrameSize` is the only thing that moves.

## Decision: what happens past the new ceiling (ticket note 2)

**The close stays — but it becomes legible.** Answering with a JSON-RPC error and staying up is not achievable at this layer, and I would rather say so than ship something that looks like it:

- **Single frame over `maxFrameSize`** — NIO's decoder fails the channel and writes the 1009 close *below* our handler. The bytes never arrive.
- **Fragmented message over `maxSize`** — catchable, but no `id` was ever decoded, so the only legal reply is `id: null`, which `app.js` correctly treats as a notification and drops.

So instead, both ends now name it:

| | before | after |
|---|---|---|
| Settings | `Save failed: rpc: connection closed` | `Save failed: rpc: set-config was too large for the daemon, which closed the connection (1009)` |
| `crowd.log` | *(nothing)* | `[crowd] WARNING: /rpc read ended abnormally (per-message limit 1048576 bytes) — socket closed, in-flight requests dropped: …` |

The error is tagged `err.rpcTooLarge` / `err.rpcMethod` so callers can branch without matching prose. **Every other close reads exactly as it did** (no code and 1006 both keep the verbatim `rpc: connection closed`), and a 1009 still reconnects — the limit is per-*message*, not per-connection, so refusing would take the whole UI offline over one oversized payload.

One real bug fixed along the way: the reader's error was being **discarded with its child task**, so a fragmented over-limit message closed as if the client had merely gone away and the 1009 never reached the client at all. It is now propagated after teardown, so WSCore can map it to the close code.

### Two corrections to my own initial reasoning, both caught by measuring

1. I expected the `catch` to cover the single-oversized-frame case. **It does not** — instrumenting the reader showed that path takes the *clean* branch: NIO closes below the handler and the inbound stream just ends. The comments now say this explicitly, so nobody "fixes" it by widening the catch; there is nothing to catch. That case is diagnosable only from the client's close code, which is now handled.
2. I assumed the fragmented path already produced a 1009. It did not, on `main` either — hence the propagation fix above.

## Tests

- **`WebSocketFrameSizeTests.swift`** — a real loopback server through the production `RPCWebSocketHandler.mount`, built from the production constant (a test with its own hardcoded ceiling would stay green while `crowd` kept dropping connections). Covers a 64 KiB single text frame, the same as binary, a 64 KiB paste on a second route sharing the upgrade channel, and an over-ceiling fragmented message being refused with 1009. **Verified to fail with the constant back at `1 << 14`** (6 issues across 3 tests) — the ticket's warning that current tests pass either way was the point.
  - The `/terminal` case is an honest stand-in: `TerminalWebSocket.mount` needs a real `TerminalCockpit` (tmux + PTY). What it *does* pin is the whole causal chain minus the PTY, since `maxFrameSize` belongs to the one channel shared by every ws route. The test comment says plainly what it does not cover.
  - Trap worth flagging for review: the test must send with `write(.text/.binary)`, not `writeTextMessage`. The latter fragments at the *client's* `maxFrameSize`, and those fragments sail through a 16 KiB server decoder — the test would pass without the fix.
- **`rpc-close-code.test.js`** — jsdom, and this one **does** run in PR CI. 16 assertions covering the 1009 message and tags, the unchanged behaviour of every other close, settled-entry handling, and that reconnect still happens.

## CI: the step I proposed, and why it isn't here

The plan was a filtered `swift test --package-path Packages/CrowDaemon --filter WebSocketFrameSize` step, mirroring the existing `--filter SmartDetect` precedent. **It is not viable, and I verified that rather than finding out on this PR:** CrowDaemonTests *compiles* on Linux (my new file included) but cannot **link** — `swift:6.1` ships a `libswiftObservation.so` with an undefined reference to `swift::threading::fatal`, a symbol no library in `/usr/lib/swift/linux` defines, on **amd64 and arm64 alike**. Confirmed pre-existing on a clean tree, and CrowIPCTests links fine in the same container, so it is specific to that dependency graph rather than my new test dep.

`ci.yml` gains a comment recording this — the existing note attributes CrowDaemon's exclusion solely to Darwin-only CrowTelemetry, which is incomplete. The Swift suite runs via `make test` and the macOS release lane; the jsdom suite covers the client half in PR CI.

## Deliberately not in scope

**Ticket note 3** — replacing whole-blob `set-config` with the granular `workspace-*` / `defaults-set` / `agents-set` RPCs. It would reduce exposure but does not fix the bug, and it is a much larger client change. Worth a follow-up ticket.

## Side effects worth knowing

- `maxFrameSize` also drives *outbound* fragmentation, so `crowd` now emits frames up to 1 MiB instead of chunking at 16 KiB.
- The decoder may buffer up to 1 MiB per connection while accumulating a frame, against 16 KiB before. `/terminal` is capped at 16 concurrent connections by `TerminalConnectionLimiter`.

## Verification

- `swift test --package-path Packages/CrowDaemon --filter WebSocketFrameSize` — 4/4, and confirmed failing at the old ceiling.
- `npm run test:ci` in `web-tests` — all suites green, including the untouched `rpc-timeout` (whose bare `close()` still yields the verbatim old message).
- `make parity` — notification catalogs, CLI/RPC parity (97 methods), NSLog gate all pass.
- Full `swift test --package-path Packages/CrowDaemon`: 2 failures in `WebTerminalAssetTests` / `WebNotificationCenterTests` — **pre-existing**, verified identical on a clean tree (app.js text-drift guards that went unnoticed because this target doesn't run in PR CI). Not touched here; they look like a worthwhile separate cleanup.


dgershman (approved):
## Code & Security Review

Reviewed the full diff, built it, and ran every suite locally. This is a correct fix for a real, well-diagnosed bug, and the write-up's two self-corrections (the `catch` does *not* cover the single-oversized-frame path; the fragmented path never produced a 1009 on `main` either) are both borne out by the code.

### Critical Issues

None.

### Security Review

**Strengths:**

- **The auth boundary is untouched.** Both `/rpc` and `/terminal` still gate on `WebSocketOriginGuard.isAllowedOrigin` + `WebAuthGuard.authorize` *before* returning `.upgrade()`, so the larger ceiling is only reachable post-authentication. The `localDirect` capture and `localOnlyDenial` call in `RPCWebSocketHandler.swift:110` survive the re-indent unchanged — worth stating explicitly, since a 40-line block moving inside a `do` is exactly where a security gate goes missing.
- **The new constant is a ceiling, not a bypass.** `CrowDaemon.maxWebSocketFrameSize` = `1 << 20` = 1048576, matching `SocketServer.maxMessageSize` (`CrowIPC/SocketServer.swift:26`), so parity with the Unix socket path is real and not just claimed.
- **`ErrorBox` is correctly synchronized** (`NSLock`, `@unchecked Sendable`) — the obvious `var` capture across a task group would have been a data race.

**Concerns:**

- Nothing blocking. Two things I checked that could have been problems and are not:
  - **Outbound fragmentation.** `maxFrameSize` also drives egress, so `crowd` now emits frames up to 1 MiB — a peer with a 16 KiB decoder would answer *its own* 1009. I grepped for native WebSocket clients: the only `WSClient`/`HummingbirdWSClient` use in the tree is the new test file, and it sets `maxFrameSize` from the same constant. Browsers impose no frame ceiling. So terminal scrollback replay (`replayData`, easily >16 KiB) is safe. No action needed — flagging only because it is the non-obvious half of the side effect the description lists.
  - **Memory amplification.** The decoder may buffer 1 MiB/connection against 16 KiB before. `/terminal` is capped at 16 concurrent (`TerminalConnectionLimiter`); `/rpc` has a per-connection in-flight cap (`RPCDispatcher`, 64) but no *connection* cap. That's a pre-existing property and it sits behind auth, so the 64x is bounded by the same trust boundary that already permitted unbounded connections. Green.

### Code Quality

Green — consider, none blocking:

- **`app.js:143-152` — a 1009 close labels *every* pending request "too large", not just the oversized one.** A 1009 has exactly one culprit, but the loop rejects all unsettled waiters with `'rpc: ' + w.method + ' was too large…'`. With `refreshLive` on a 4 s interval and `refreshSessions` on 10 s (`app.js:5787-5788`), an oversized `set-config` will routinely coincide with a poller, which then gets told *it* was too large. Impact is low — both pollers `catch (_) { return; }` so nothing reaches the screen — and `err.rpcTooLarge` on an innocent method is the part that could mislead a future caller that branches on the tag. Wording like `"rpc: connection closed — a message exceeded the daemon's limit (1009); <method> did not complete"` would be true for every waiter at no cost. The new jsdom suite tests only the single-pending case, so this is untested either way.
- **`RPCWebSocketHandler.swift:166-170` — the log line names the per-message limit for *any* abnormal read.** A TCP reset or a sleeping laptop would produce `read ended abnormally (per-message limit 1048576 bytes)`, which reads as an attribution even though the real cause follows in `\(error)`. Same nit on `TerminalWebSocket.swift:127-129`. Cheap fix: keep the raw error, drop the parenthetical, or make it conditional.
- **`app.js:5026-5028` — `alertModal(...)` runs before `resetScrollbackSync()` and the reconnect timer.** `modalDialog` builds DOM; if it ever threw, the terminal would skip its teardown and never reconnect. Moving the call below `setTimeout(connectTerminalWs, …)` is strictly safer at zero cost.
- **`CHANGELOG.md`** puts `### Fixed` above `### Added` under Unreleased; `0.1.0` has the reverse order. Cosmetic.

**On the tests** — these are unusually good and I want to name why: `WebSocketFrameSizeTests` builds its server from `CrowDaemon.maxWebSocketFrameSize` rather than a literal, so it cannot rot into a test that stays green while `crowd` drops connections; and the `write(.text(_))` vs `writeTextMessage` trap called out at line 127 is exactly the mistake that would have produced a test passing without the fix. The `/terminal` stand-in is honestly scoped in its own doc comment. The `ci.yml` comment recording the `libswiftObservation.so` / `swift::threading::fatal` link failure is the right outcome for a step that was investigated and found non-viable.

### Verification

| Check | Result |
|---|---|
| `swift build --package-path Packages/CrowDaemon` | clean |
| `swift test --filter WebSocketFrameSize` | 4/4 pass |
| `npm run test:ci` (web-tests) | 11 suites green, incl. new 16-assertion `rpc-close-code` |
| Full `swift test` (CrowDaemon) | 366 tests, 2 failures — **confirmed unrelated** |

The two failures (`WebTerminalAssetTests.refreshTerminalsRebindsTheActiveTerminalToTheFreshRow`, `WebNotificationCenterTests.bootCatchResetsTheHistory`) assert on `refreshTerminals`'s body and the boot-recovery block respectively. This diff touches neither region and deletes no lines from them — it only edits the two `onclose` handlers — so the "pre-existing" claim holds. Worth the separate cleanup ticket the description suggests, since this target doesn't run in PR CI.

I also confirmed the overload move is behavior-neutral beyond `maxFrameSize`: the `WebSocketServerConfiguration` variant of `http1WebSocketUpgrade(webSocketRouter:configuration:)` is `@_disfavoredOverload`, so the previous no-configuration call already bound to the `HTTP1WebSocketUpgradeChannel.Configuration` overload; leaving `http1:` defaulted reproduces the old channel exactly.

### Summary Table

| Color  | Meaning      | Count |
|--------|--------------|-------|
| Red    | Must fix     | 0     |
| Yellow | Should fix   | 0     |
| Green  | Consider     | 4     |

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 4 Green] findings. The Green items are wording and statement-ordering polish; none affects correctness, and all can land in a follow-up or a quick amend at the author's discretion.

---

Issue #947, opened 2026-08-07
Design discussion: per-session read-only viewer links for the web client


dgershman:
## Context

Crow's web client (`crowd`) exposes three interactive surfaces, all gated the same way once a web password is set (CROW-593):

| Surface | Path | Risk if exposed |
|---------|------|-----------------|
| Full UI + RPC | `/rpc` WebSocket | Runs `add-worktree`, `delete-session`, `send`, etc. — effectively remote control |
| Live terminal | `/terminal` WebSocket | Binary keystrokes → PTY → interactive shell |
| Static assets | `/`, `/app.js`, … | UI shell only |

Auth today is **binary**: local loopback = trusted; remote via Tailscale Serve/ngrok = HTTPS + shared password cookie → **full owner access**. There is no role, scope, or per-session grant.

ADR 0009 says every UI is an equal client on the same RPC surface — so "read-only viewer" is a new concept the daemon does not have yet.

## Goal

Allow the owner to share a **read-only view** of a specific session with another person:

- Viewer gets a link; they can **see** (terminal scrollback / live follow, session metadata, PR status) but **cannot type or take actions**.
- **Per-session** scope — viewer only sees sessions explicitly shared, not the full sidebar.
- **Network boundary stays with the owner** — viewer must already be on the owner's tailnet (Tailscale Serve); Crow handles app-level authorization.
- Share link must **not** be equivalent to the owner's web password.

## Proposed model

```
Network boundary (owner's job):  Tailscale tailnet — viewer can reach the machine
App boundary (Crow's job):       Share link → see only session X, read-only
```

**Per-session viewer token** minted by the owner:

```
https://your-machine.tailnet/#/view/<token>
```

Server-side grant (conceptual):

```swift
struct ShareGrant {
    let token: String           // 256-bit random
    let sessionID: UUID
    let createdAt, expiresAt: Date
    let revoked: Bool
    let capabilities: [.terminalScrollback, .liveFollow, .artifacts]  // tunable
}
```

Viewer auth: token in URL (or exchanged for a short-lived `viewer` cookie) → `ConnectionRole.viewer(sessionID)`.

### Option A — Per-session viewer tokens + RPC allowlist (recommended)
- Daemon enforces `ConnectionRole.viewer` with a tight RPC allowlist (`get-session`, `list-artifacts`, filtered events — **not** `send`, `delete-session`, `set-config`, secrets, etc.).
[...]

PR #966, opened 2026-08-10
Design analysis: per-session read-only viewer links (CROW-947)

dhilgaertner:
Refs #947

Design work only — the issue says **"no coding on this issue"**. No production code, no `ShareGrant` type, no `RPCWebSocketHandler` changes. Two new documents under `docs/spikes/`.

## Three findings that change the ticket's framing

**1. `localOnlyDenial` is a denylist; a viewer needs the opposite polarity.** It is default-*allow* over 97 methods, denying 8 named ones. A viewer needs default-*deny*, allowing 6. Writing the viewer rule as more `case` arms in that same switch would silently grant every method nobody remembered to name.

And `ParityLedger`'s `isWrite` is *not* the predicate, tempting as it looks: 29 methods are reads, but only **6 are session-scoped**. The other 23 are fleet- or config-wide and must be denied — `get-state` (the whole store *and* `AppConfig`), `list-sessions`, `get-config`, `defaults-get` (absolute local binary paths). Allowlisting method *names* is also insufficient: all six safe reads take `session_id` as a trusted parameter, so the gate must pin it to the grant.

**2. `tmux attach -r` does not scope anything.** The reassuring half: each `/terminal` connection already gets its own grouped tmux session, so `-r` would affect only that viewer and leave the owner's attachment untouched. The disqualifying half: `-r` blocks *keystrokes*, and `select-window` is not a keystroke — it is a server-side `tmux` command over a grouped session that shares the cockpit's **entire** window list. A viewer can walk indices `0..N` and watch the Manager and every other session, which is exactly the ticket's "Never shareable" list. Window indices are mutable too, so any allowlist must be re-derived per frame. **Option B is cheap to make read-only and expensive to make per-session.**

**3. EventHub filtering is mostly unnecessary.** `changedFrame` carries zero payload — it is a re-fetch nudge, and the re-fetch already goes through the RPC gate. Only `notify` leaks cross-session. ~15 lines, not the plumbing the ticket budgets.

## Recommendation

**Hybrid A + C**, not the ticket's A + B. Every gate B must add — stdin, `resize`, `select-window` — is a gate C doesn't need, because a one-way route never accepts those frames and picks the window server-side from the grant. C's stated con, "duplicate streaming logic", measures at ~60 lines of PTY boilerplate against three security gates deleted rather than written and maintained.

The five open questions are answered from code, with tradeoffs, in §3 — they're your calls, so they're presented as recommendations.

## Security gaps beyond the checklist

Six, notably:

- **Rate-limiting a hash-fragment route is impossible.** `#/view/<token>` never reaches the server (ADR 0018), so "rate-limit `/view/*`" cannot work as written. The fragment is still the right place for the token; it needs a real `POST /view/redeem` endpoint to have something to limit — with a *separate* limiter, since `LoginRateLimiter` is global-not-per-IP by design and sharing it would let share-link traffic lock you out of `/login`.
- **Scrollback replay vs. the daemon's own stripping.** Crow blanks gateway auth headers from `get-state`, then would ship those same headers verbatim from a shared pane — replay is 50 000 lines of raw agent output.
- **`resize` is unguarded** and `window-size latest` is set, so a viewer reshapes the pane you're working in.
- Plus grant persistence across restart (and the `JSONStore` single-instance trap, #728), `list-artifacts` param pinning, and the local-direct fast path — which returns `Decision(true, "local")` *before* loading config, so a viewer role evaluated inside it silently escalates to owner.

## On the ADR

This warrants one, and **0019** is the next free number. The honest framing is narrower than "it breaks ADR 0009": 0009's load-bearing invariant (crowd is sole writer; all mutation flows through RPCs) survives intact. What it contradicts is "no client is privileged" — and that has *already* been relaxed once, by `localOnlyDenial`, with an 85-line doc comment doing an ADR's job. A viewer role relaxes it in a new way: the first credential class that is mintable, delegable, and held by a third party.

I've drafted the ADR skeleton but **not filed it** — an ADR records a decision, and you haven't made one yet. Same for the implementation ticket: drafted at `docs/spikes/947-implementation-ticket-draft.md`, scoped to Phase 1, deliberately unlabelled (the issue asks for `crow:auto`, but that would dispatch an agent at a scope you haven't signed off).

`Refs`, not `Closes` — #947 stays open until you've answered.

### DIFF
--- a/docs/spikes/947-implementation-ticket-draft.md
+++ b/docs/spikes/947-implementation-ticket-draft.md
@@ -1,0 +1,120 @@
+# DRAFT — implementation ticket for CROW-947 (Phase 1)
+
+> **This file is a draft. It has not been filed.**
+> File with `gh issue create --title "…" --body-file docs/spikes/947-implementation-ticket-draft.md`
+> after the open questions on [#947](https://github.com/corveil/crow/issues/947) are answered.
+>
+> #947 asks for `crow:auto` on the implementation ticket. **Labelling is left to the owner** —
+> this draft deliberately applies none, since `crow:auto` would dispatch an agent at a
+> scope the owner has not yet signed off.
+>
+> **Suggested title:** `Per-session read-only viewer links, Phase 1: grants + scoped RPC gate (CROW-947)`
+>
+> Everything below the rule is the proposed issue body.
+
+---
+
+Implements Phase 1 of the design agreed in [#947](https://github.com/corveil/crow/issues/947). Feasibility analysis, cost breakdown, and the reasoning behind each choice below: `docs/spikes/947-readonly-viewer-links.md`.
+
+The owner mints a per-session, expiring, revocable grant. A holder of the link sees that session's metadata, PR status, and artifacts — and nothing else, enforced at the daemon rather than hidden in the UI.
+
+## Scope
+
+**In:**
+
+1. `ShareGrant` model + store — `{token, sessionID, createdAt, expiresAt, revoked, capabilities}`, persisted in `store.json`.
+2. Token redemption endpoint — `POST /view/redeem`, token in the **body**, exchanged for a scoped cookie.
+3. `ConnectionRole` on `/rpc` — default-deny viewer allowlist, with `session_id` params pinned to the grant.
+4. `notify` event scoping in `EventHub`.
+5. `crow share create | list | revoke` CLI verbs + backing RPCs.
+6. Viewer route `#/view/<token>` in `app.js` — metadata, PR status, artifacts. Read-only affordances only.
+
+**Out (later phases, do not build here):**
+
+- Live terminal follow / any PTY for viewers — Phase 2, and see spike §1.3 before choosing an approach.
+- Terminal scrollback for viewers — the `terminalScrollback` capability may be *defined* in the model but must be unimplemented and rejected in Phase 1.
+- In-app Share button, active viewer count, revoke UI — Phase 3.
+- Any change to owner behaviour. An owner's `/rpc`, `/terminal`, and cookie handling must be byte-identical after this change.
+
+## Requirements
+
+### R1 — The gate is default-deny, and `isWrite` is not the predicate
+
+Add a viewer allowlist **separate from `RPCWebSocketHandler.localOnlyDenial`**. That function is default-*allow* (deny 8 named methods); reusing it or adding viewer cases to its `switch` would grant a viewer every method nobody remembered to name.
+
+Exactly six methods are viewer-reachable, all pinned to the grant's session:
+
+| Method | Param to pin |
+|---|---|
+| `get-session` | `session_id` |
+| `list-links` | `session_id` |
+| `list-worktrees` | `session_id` |
+| `list-terminals` | `session_id` |
+| `list-artifacts` | `session_id` |
+| `get-pr-status` | `session_id` |
+
+Everything else — all 91 remaining methods — is denied. Note in particular that 23 of the 29 `ParityLedger` `.read(…)` methods must be denied, including `get-state`, `list-sessions`, `list-sessions-live`, `get-config`, and `defaults-get`. **Do not derive the allowlist from `isWrite`.**
+
+Allowlisting a method name is insufficient: all six handlers take `session_id` as a parameter and trust it (`EngineRouter.swift:280`, `:695`, `:854`, `:1079`, `:321`; `RPCHandlers.swift:617`). The gate must reject a request whose `session_id` is not the granted one.
+
+Additionally strip `path` and `dir` from `list-artifacts` responses to viewers — absolute filesystem paths are a needless disclosure to a non-owner.
+
+### R2 — The viewer role must not be reachable through the local-direct fast path
+
+`WebAuthGuard.authorize` returns `Decision(isAuthorized: true, reason: "local")` for a loopback peer with no `X-Forwarded-For` *before loading config* (`WebAuth.swift:174-176`), and `RPCWebSocketHandler` skips its gate entirely when `localDirect` (`:114`). Evaluate the role independently of that path. Prefer extending `Decision` with the role over adding a second call at each site, so the check cannot be omitted at a future call site.
+
+### R3 — Cookie must not collide with the owner's
+
+Use a distinct cookie name **and** a distinct path — e.g. `crow_view` at `Path=/view` — not a second `crow_session` at `Path=/`. Same name + path means an owner who opens a link they minted overwrites their own session, and re-logging-in revokes the viewer's. A distinct path also means the browser never attaches the viewer cookie to `/rpc` at all.
+
+Keep the existing flags: `HttpOnly; Secure; SameSite=Lax`.
+
+### R4 — Redemption happens at a real endpoint
+
+`#/view/<token>` is a URL fragment; it never reaches the server (ADR 0018), so the mitigation "rate-limit `/view/*`" in #947 cannot work as written. Keep the token in the fragment — that keeps it out of server logs, `Referer`, and proxy access logs — and add `POST /view/redeem` taking the token **in the body**, returning the R3 cookie. Strip the fragment client-side after redemption.
+
+Rate-limit redemption with a **separate** limiter instance. `LoginRateLimiter` is global rather than per-IP by design (`WebAuth.swift:127-130`, because behind a proxy every attempt shares the loopback peer); sharing one instance means guessing traffic against share links locks the owner out of `/login`.
+
+### R5 — Grants expire and persist
+
+Default TTL 24 h, settable via `crow share create --expires`. Reuse the existing sweeper shape — `sessions.prune()` already runs on a 300 s timer at `CrowDaemon.swift:491`.
+
+Persist in `store.json`, not `config.json`. **Use the single injected `JSONStore` owned by `SessionService`** — a throwaway `JSONStore()` reads its own stale snapshot and its whole-store write can clobber a record another writer just added (#728, and `CLAUDE.md` § Concurrency Safety).
+
+### R6 — Scope `notify`, leave `changed` alone
+
+`EventHub.changedFrame` carries no payload (`EventHub.swift:18`) — it is a re-fetch nudge and needs no filtering, because the re-fetch itself goes through R1. `notify` carries `{event, key, title, body}` (`:66`) with per-session automation text and **does** leak across sessions. Add an optional session scope to `subscribe` (`:25`) and a predicate in `broadcast` (`:38`).
+
+### R7 — Never shareable
+
+Reject grant creation for: the Manager session, any session with no worktree, and global terminals. Enforce server-side at mint time *and* at connection time — a session can become a Manager-adjacent thing after a grant is minted.
+
+## Build gates (these fail CI, not review)
+
+- **`ParityLedger.rpcMethods`** — every new `share-*` RPC needs a row, or `RPCLedgerParityTests` fails. Add a `viewerReachable` classification column in the same change so the six-method allowlist is enumerated where the parity test gates it; a future RPC then cannot become viewer-reachable without an explicit decision.
+- **`RPCLanePolicy.rules`** — every new method needs an entry or `RPCLanePolicyTests` fails.
+- **ADR 0016** (`docs/adr/0016-cli-control-plane-parity.md`) — every user-mutable capability needs a CLI path, enforced by `scripts/check-cli-parity.sh` on the Linux lane. This is why `crow share …` is in scope rather than deferred with the UI.
+- **ADR 0019** (to be written) — the client-authority-tiers ADR should land **before or with** this ticket. The spike's §6 explains why: ADR 0009 states "no client is privileged", and this introduces the first mintable, delegable credential class. Do not merge this without that record.
+
+## Acceptance
+
+- [ ] A viewer connection calling `get-state`, `list-sessions`, `get-config`, or `defaults-get` receives a denial — asserted per method, not by sampling.
+- [ ] A viewer calling `get-session` with a `session_id` other than the grant's receives a denial.
+- [ ] A viewer's `/terminal` upgrade is refused outright (no PTY in Phase 1).
+- [ ] A viewer receives `changed` frames but no `notify` frame keyed to another session.
+- [ ] An expired or revoked grant is refused at redemption **and** on an already-open socket.
+- [ ] Minting against the Manager session is refused.
+- [ ] An owner opening their own share link does not lose their owner session (R3).
+- [ ] Grants survive a `crowd` restart; `crow share revoke` survives one too.
+- [ ] Owner behaviour unchanged — existing `DaemonSecurityTests` and `LocalOnlyRPCGateTests` pass untouched.
+
+## Testing notes
+
+- Extend `DaemonSecurityTests` rather than starting a new file — it already holds the local-only gate's cases and the read/write reasoning they encode.
+- A table-driven test asserting **all 97** methods against the viewer gate is worth more than six positive cases: it is what makes a newly added method fail loudly instead of defaulting open.
+- Client changes need a jsdom suite under `Packages/CrowDaemon/web-tests/`, and it must be added to `package.json`'s `test:ci` script explicitly or CI silently skips it.
+- Per ADR 0012 (`docs/adr/0012-tests-never-touch-live-data.md`), grant-store tests must not touch a live `store.json`.
+
+## Estimate
+
+Plumbing is ~3 Swift files and ~120 lines (spike §2). The audit and its tests are the majority of the work, and the `ParityLedger` column is what keeps that audit from decaying.
--- a/docs/spikes/947-implementation-ticket-draft.md
+++ b/docs/spikes/947-implementation-ticket-draft.md
@@ -53,7 +53,11 @@ Exactly six methods are viewer-reachable, all pinned to the grant's session:
 | `list-artifacts` | `session_id` |
 | `get-pr-status` | `session_id` |
 
-Everything else — all 91 remaining methods — is denied. Note in particular that 23 of the 29 `ParityLedger` `.read(…)` methods must be denied, including `get-state`, `list-sessions`, `list-sessions-live`, `get-config`, and `defaults-get`. **Do not derive the allowlist from `isWrite`.**
+Everything else — all 91 remaining methods — is denied. **Do not derive the allowlist from `isWrite`:** 23 of the 29 `ParityLedger` `.read(…)` methods must be denied. In full, so nothing is inferred:
+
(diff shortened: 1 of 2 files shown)


dgershman (changes requested):
## Code & Security Review

Docs-only (410 lines, two new files under `docs/spikes/`, no code). So the review is an **accuracy audit**: this spike's value is that an implementer will build a security boundary from its line references, counts, and polarity claims. A wrong claim here becomes a wrong gate later.

I verified the load-bearing claims against the tree rather than reading for plausibility. **The substance holds up — including all three headline findings.** Corrections below are small and none of them changes the recommendation.

### Verified correct

| Claim | Result |
|---|---|
| `localOnlyDenial` at `RPCWebSocketHandler.swift:312`, **default-allow**, 8 unconditional + 2 conditional denials | ✅ exact |
| Single call site `:114`, **skipped entirely when `localDirect`**; `localDirect` captured at `:58` | ✅ exact (only production call site in the repo) |
| `ParityLedger` **97 methods / 29 reads / 6 session-scoped / 23 denied** | ✅ counted precisely — 97 entries, no dupes, 29 `.read`, 29−6=23 |
| All six session-scoped reads exist and are read-classified | ✅ `get-session` :93, `list-links` :124, `list-worktrees` :132, `list-terminals` :136, `list-artifacts` :192, `get-pr-status` :161 |
| All six take `session_id` as a **trusted param with no ownership check** | ✅ confirmed at each cited line (`EngineRouter.swift:280` etc.) |
| **`select-window` is unvalidated and server-side** | ✅ `TerminalWebSocket.swift:93-97` → `cockpit.selectWindow(group:index:)` shells out; `if let window = control.window` is the *only* check |
| `list-terminals` hands the client the raw window index | ✅ `EngineRouter.swift:892` |
| EventHub: `changed` zero-payload `:18`, `subscribe` takes only a continuation `:25`, `broadcast` unconditional `:38-42` | ✅ exact, file is 97 lines |
| **Local-direct fast path returns before config is loaded** | ✅ `WebAuth.swift:174-176` returns at :175; `configProvider()` first called at :180. Documented as deliberate at :160-161 and :173 |
| `SessionStore` is `[String: Date]`, no metadata, in-memory, has `prune()` | ✅ :90-123, only stored props are `ttl`/`lock`/`tokens` |
| `LoginRateLimiter` global-not-per-IP, deliberately | ✅ :127-130 |
| Artifacts' four guards + planted-symlink history | ✅ all four; the `shot.png` → `~/.claude/config.json` comment is verbatim at :71-74 |
| `list-artifacts` has **no ownership check**, returns absolute `path` + `dir` | ✅ `RPCHandlers.swift:616-648` |
| `strippedForTransport` blanks Jira token / webAuth hash+salt / gateway header values — **not** `defaults.binaries` | ✅ `SettingsSecrets.swift:30-44` |
| `CommandRouter` is a flat `[String: Handler]` with no caller identity (kills Option D) | ✅ `CommandRouter.swift:13` |
| ADR 0009 quotes | ✅ verbatim at :37-38 and :70 |
| ADRs 0001–0018 exist, **all `Accepted`, none superseded → 0019 free** | ✅ |
| `scrollbackHistoryLimit = 50000` / `TerminalConnectionLimiter(max: 16)` / `CrowDaemon.swift:491` 300 s prune | ✅ all exact |
| ADR 0016 mechanically enforced (`scripts/check-cli-parity.sh`, CI `ci.yml:138`) | ✅ |
| `login.html` re-appends `location.hash` after login | ✅ :54-61 |
| Relative links resolve; `docs/spikes/` precedent exists (spike #822 → ADR 0013) | ✅ |

The three headline findings — denylist-vs-allowlist polarity, `isWrite` being the wrong predicate, and `-r` not scoping `select-window` — are all correct and materially change the ticket. Flagging that `get-state` would have been handed to a share-link holder by an `isWrite`-derived allowlist is the single most valuable line in the document.

---

### Yellow — should fix

**1. `crow-tmux.conf` does not *set* `window-size latest`. §5.2 (L193), and L79.**

> §5.2: "`crow-tmux.conf` sets `window-size latest`."

It does not. That is tmux's **default**; the conf leaves it and sets `aggressive-resize on` (`crow-tmux.conf:205`) *specifically* to narrow its effect — the comment at :193-204 explains it was added for #667 to size each window "from only the clients for which it is the CURRENT window."

The conclusion survives — a viewer sitting on the owner's *current* window can still drag its geometry — but two things need correcting:

- The mechanism attribution is wrong (default, not a Crow setting).
- **A partial mitigation already exists and goes uncredited.** With `aggressive-resize on`, a viewer parked on a *different* window does **not** perturb the owner's. That materially shrinks §5.2's blast radius, and an implementer reading the row as written would not know the closely-related problem was already solved once.

Worth being precise because §5.2 is a severity-rated checklist row someone will action. §1.3's phrasing ("uses") is defensible; §5.2's ("sets") isn't. Recommend restating as: *tmux's default `window-size latest` applies; `aggressive-resize on` (#667) already scopes negotiation to clients viewing the window, so the exposure is a viewer on the **same** window — reachable by composing §5.1.*

**2. Broken intra-doc anchor. L90.**

`[§5.4](#54-grants-and-daemon-restart)` does not resolve. GitHub slugifies `### 5.4 Grants and daemon restart — *design gap, not a vulnerability*` to `54-grants-and-daemon-restart--design-gap-not-a-vulnerability`. The other four anchors (§1.1, §1.2, §1.3, §6) resolve fine — those headings carry no trailing em-dash clause. Trivial, but §5.4 is the one an implementer jumps to for the `JSONStore` #728 trap.

---

### Green — consider

**3. `select-window` also replays scrollback — this strengthens your own argument.** `TerminalWebSocket.swift:98-102`: after `selectWindow`, the handler calls `cockpit.replayData(group:index:)` and yields it. So §5.1's escape is not only "watch the Manager live" — walking indices `0..N` hands the viewer up to **50 000 lines of replayed scrollback per window**. §5.1 and §5.3 compose into something worse than either row states, and it makes the Option C case (never accept a window selection) stronger than the doc currently argues.

**4. "nine `*-get` settings reads" (L40) is exact only under one reading.** Twelve read methods end in `-get`. Subtracting the three you name separately (`defaults-get`, `workspace-get`, `job-get`) leaves exactly nine — so the arithmetic closes. But two of those nine (`gateway-get` :250, `web-password-get` :254) sit under `// Secrets — local-socket only (CROW-815)`, not the Settings MARK; counting only the Settings MARK gives eight. Both are already local-socket-only and can never reach a remote viewer anyway. Since the ticket draft repeats this enumeration in R1, worth tightening.

**5. `list-agents` (`ParityLedger.swift:168`) is missing** from §1.1's list of denied reads. Your "include" hedge covers it and R1's "all 91 remaining methods" catches it regardless — completeness only.

**6. Cookie string omits `Max-Age`.** §1.4/R3 quote `crow_session=…; HttpOnly; Secure; SameSite=Lax; Path=/`; the real string (`WebAuth.swift:231`) also carries `Max-Age=\(Int(ttl))` (default 7 d, :95). R3 says "keep the existing flags" — list `Max-Age` explicitly so the viewer cookie's expiry is deliberately tied to the grant TTL rather than inherited by accident.

**7. §1.1 L40 cites `:274-283`; the paragraph starts at :275** (274 is a bare `///`). Substance correct.

**8. The `viewerReachable` ledger column gates enumeration, not behaviour.** `RPCLedgerParityTests` asserts **only name-set parity** against the live router (both directions, `RPCLedgerParityTests.swift:74-98`); the `isWrite` classification is checked in `CrowCLITests/ParityGateTests.swift`, never cross-checked against runtime. So a `viewerReachable` column does deliver what §1.1 claims — a new method can't be added without someone classifying it — but it can be classified *wrongly* and CI stays green. The ticket's acceptance criterion ("asserted per method, not by sampling") is what actually closes that, which is the right instinct; worth one clause in §1.1 so the guarantee isn't oversold.

---

### Security review

**Strengths:**
- Correctly identifies the polarity inversion as the central hazard, and insists the two gates be **separate functions with separate defaults** rather than more `case` arms. This is the finding that prevents a default-open viewer gate.
- Catches that method-name allowlisting is insufficient because all six reads trust `session_id` — verified true at every cited line. Param pinning is the right control.
- Catches the local-direct escalation path (R2). Verified: the bypass returns *before* config load, by design, so a viewer role evaluated inside it silently becomes owner. This is the kind of thing that gets found in a post-mortem, not a spike.
- §5.5 is correct and invalidates a checklist row from the parent issue: `#/view/<token>` never reaches the server, so "rate-limit `/view/*`" cannot function. The `POST /view/redeem` fix, keeping the token in the fragment (out of logs/`Referer`), plus a **separate** limiter instance, is right — sharing `LoginRateLimiter` really would let share-link traffic lock the owner out of `/login`.
- §3 Q4's asymmetry framing — RPC surface strips secrets, terminal surface structurally cannot — is the correct risk model, and defaulting `terminalScrollback` off is the right call.
- Correctly routes grants to `store.json` and flags the `JSONStore` single-instance trap (#728).

**Concerns:** none blocking. No secrets, credentials, or sensitive data in either document; all references are file paths and line numbers. The one accuracy defect (Yellow 1) understates an existing mitigation rather than overstating a risk — the safe direction to be wrong, but still worth fixing.

### Code quality

No code. Both documents follow the established `docs/spikes/` convention (#822 precedent, which likewise fed an ADR). Not filing the ADR and not labelling the draft ticket `crow:auto` are the correct calls — an ADR records a decision, and `crow:auto` would dispatch an agent at an unapproved scope. `Refs #947` rather than `Closes` is right.

### Summary Table

| Color  | Meaning      | Count | Items |
|--------|--------------|-------|-------|
| Red    | Must fix     | 0     | — |
| Yellow | Should fix   | 2     | `window-size` attribution + uncredited `aggressive-resize` mitigation; broken §5.4 anchor |
| Green  | Consider     | 6     | replay-on-select-window; `*-get` count precision; `list-agents`; `Max-Age`; line range; ledger-column guarantee scope |

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 6 Green] findings.

To be clear about proportion: this is unusually rigorous work and the analysis is correct where it counts. Every claim I could mechanically check held, and the three findings that reframe the ticket are all real. The two Yellows are a factual attribution and a broken link — both a few minutes' work. Fix those and this is a straight approve.

---

dhilgaertner:
Thanks — all eight taken, nothing disputed. I re-verified each against the tree rather than accepting them, and each held. Pushed in bf7976a.

**Yellow 1** — you're right on both halves, and the second half is the one that mattered. `crow-tmux.conf:196` names `window-size latest` only to explain what it's working around, and `:206` sets `aggressive-resize on` (#667) to narrow it. §5.2 now credits that mitigation and annotates the severity as partly mitigated.

I went one step past your suggested restatement, using the conf's own comment at `:202-204`:

- The narrowing is **partial under `latest`** — `aggressive-resize` is fully honored under `smallest`/`largest`, and under `latest` it "still narrows which clients a window tracks" rather than eliminating the interaction.
- The conf pairs it with "app.js's focus-gated resize (only the focused surface sends a resize frame)". **That half is client-side and therefore worth nothing against a viewer**, who controls their own client. So any claim that resize is handled has to rest on `aggressive-resize` alone.

Flagging that explicitly because it cuts against my own correction: the mitigation is real but thinner than the conf comment reads if you take both mechanisms at face value.

**Green 3** is the best catch in the review — I read `TerminalWebSocket.swift:98-106` and still described §5.1 as a live-view escape. It's a scrollback escape too: walking `N = 0…` yields up to 50 000 lines *per window*. §5.1 and §5.3 compose, and it makes the Option C case stronger than I'd argued.

**Green 8** — fair, I oversold it. `RPCLedgerParityTests` asserts name-set parity only; `isWrite` is checked against the method *name* in `ParityGateTests`, never against runtime. So the column gates **enumeration, not correctness** — a row can be classified wrongly and CI stays green. Both documents now say so and point at the per-method acceptance assertion as what actually closes it.

**Green 4 + 5** — "nine `*-get` settings reads" was arithmetically exact but ambiguous, and `list-agents` was missing. Fixed both at once by enumerating **all 23** denied reads explicitly, in the spike and in the ticket's R1. More useful to an implementer than any description of the set, which was the right instinct behind flagging it.

**Yellow 2 / Green 6 / Green 7** — anchor fixed (and the other four audited); `Max-Age` added with the reason it matters (inheriting `SessionStore`'s 7-day default would outlive a 24 h grant, leaving a live cookie on a dead grant); `:274-283` → `:275-283`.

Still docs-only.

dgershman (approved):
## Code & Security Review

Second round. Docs-only still — `git diff main...HEAD` is exactly the two files under `docs/spikes/`, no production code, no `ShareGrant`, no `RPCWebSocketHandler` change. So this is again an **accuracy audit**, scoped to the delta in `bf7976a` plus an independent re-derivation of the claims that changed.

I re-verified all eight of my prior findings against the tree rather than reading the reply for plausibility. **All eight are correctly addressed, and two of the fixes are more precise than my own review was.** One new Green-level nit below; nothing blocking.

### Prior findings — verified fixed

| # | Finding | Fix verified |
|---|---|---|
| **Y1** | `window-size` attribution + uncredited mitigation | ✅ `crow-tmux.conf` contains **no** `set … window-size` line — confirmed it is tmux's default, not a Crow setting. `:196` carries "tmux's default `window-size latest`"; `:206` is `set -gw aggressive-resize on`; `#667` is at `:194`. Quote at `:193` is verbatim. §1.3's looser "uses" phrasing is gone, now deferring to §5.2. |
| **Y2** | Broken §5.4 anchor | ✅ `#54-grants-and-daemon-restart--design-gap-not-a-vulnerability` matches GitHub's slug (double hyphen where the em-dash was). Re-audited **all five** anchors in the doc against all 22 headings — every one resolves. All four relative links resolve to files that exist. |
| **G3** | `select-window` also replays scrollback | ✅ `TerminalWebSocket.swift:98-106` is exact — `:98-103` comment, `:104` `replayData(group:index:)`, `:105` `continuation.yield(replay)`, `:106` close. My `:98-102` was the sloppier citation. |
| **G4+G5** | `*-get` count ambiguity; missing `list-agents` | ✅ Best fix of the eight. I independently derived the denied set (29 `.read` rows − the 6 session-scoped) and compared it to the enumerated list: **set-identical, 23 each, no extras, no omissions**, `list-agents` present. Identical list in both documents. Replacing the description with the enumeration is strictly better for an implementer building a default-deny gate. |
| **G6** | Cookie string omitted `Max-Age` | ✅ `WebAuth.swift:231` is `"\(cookieName)=\(token); HttpOnly; Secure; SameSite=Lax; Path=/; Max-Age=\(Int(ttl))"` — the doc's quoted string now matches exactly. `SessionStore.init(ttl:)` defaults to `7 * 24 * 3600` at `:95` as cited. The added rationale (a 7-day cookie outliving a 24 h grant) is the right reason to state. |
| **G7** | `:274-283` → `:275-283` | ✅ `:274` is a bare `///`; the `defaults-get` paragraph runs `:275-283` and ends on the "gate a *read* only when…" rule. Correct. |
| **G8** | Ledger column oversold | ✅ Every sub-claim holds. `RPCLedgerParityTests.ledgerMatchesRegisteredMethods()` exists at `:75` and asserts **both** directions (`missing` at `:82`, `stale` at `:95`) against `withLiveRouter { $0.methodNames }`. And `ParityGateTests.swift:264` is literally `@Test("isWrite agrees with the method name, or the disagreement is declared")` — so "validated against the method *name*, never against runtime behaviour" is exactly right. "Gates **enumeration, not correctness**" is now the accurate framing. |

Two of these corrected me rather than the document: `aggressive-resize` is at `:206` (I said `:205`), and the replay block runs to `:106` (I said `:102`). Worth recording, since the reply claimed each finding was re-verified rather than accepted — it was.

Also re-confirmed unchanged-and-still-true: 97 methods / 29 reads / 6 session-scoped / 23 denied; `WebAuth.swift` is 236 lines; ADRs 0001–0018 exist with **0019 still free**; `ParityLedger.rpcMethods` at `:87`.

No build or test run: nothing outside `docs/` changed, so there is no compilation surface to regress.

---

### Green — consider

**1. The `:202-204` citation covers only the first of the two qualifications. §5.2 (L205-208).**

> Two further qualifications, both from the conf's own comment (`:202-204`)

The first bullet is exact — "aggressive-resize applies per-window and is honored when window-size is `smallest`/`largest`; under `latest` it still narrows which clients a window tracks" is verbatim `:202-204`.

The second bullet quotes "app.js's focus-gated resize (only the focused surface sends a resize frame)", which lives at **`:200-201`**, outside the cited range. Both quotes do come from the same contiguous comment block (`:193-205`), so "the conf's own comment" is fair — it's the parenthetical narrowing to `:202-204` that doesn't cover the second. Widening it to `:200-204`, or citing the two separately, fixes it.

Flagging it only for symmetry with G7 last round, which was the same class of thing and which you took. Substance is correct, and the point the bullet makes — that a client-side mitigation is worth nothing against a viewer who controls their own client — is the sharper half of the §5.2 rewrite and cuts against your own correction, which is the right instinct to have followed.

---

### Security review

**Strengths:**
- The §5.2 rewrite lands in the right place: it credits `aggressive-resize on` as a real, already-shipped mitigation, then declines to let that close the row — correctly isolating the residual exposure to a viewer on the **same** window, reachable only by composing §5.1. Naming the paired client-side mechanism as worthless against a viewer is the part a less careful revision would have skipped, and it is the part that keeps the severity honest.
- G3's composition point (§5.1 × §5.3) now stated in both §1.3 and §5.1: walking window indices yields not just a live view of the Manager but up to 50 000 lines of replayed scrollback *per window*. Verified against `TerminalWebSocket.swift:98-106`. This is the strongest argument in the document for never accepting a client window selection, and it now appears where an implementer will hit it.
- Enumerating all 23 denied reads removes the last place where an implementer could have inferred the set wrongly. `get-state` and `defaults-get` being on that list — with the reason — remains the single most consequential correction to the parent ticket's Option A.
- Scoping the ledger-column guarantee down to "enumeration, not correctness" is a de-escalation of the document's own claim, with the per-method acceptance assertion correctly identified as what actually closes the gap. A spike that revises its guarantees downward under scrutiny is more trustworthy than one that doesn't.
- Unchanged and still correct: the denylist/allowlist polarity hazard, `session_id` param pinning, the local-direct escalation path (R2), §5.5's invalidation of the "rate-limit `/view/*`" checklist row and the separate-limiter requirement, and routing grants to `store.json` through the single injected `JSONStore` (#728).

**Concerns:** none. No secrets, credentials, tokens, or host-specific data in either document — all references are repo-relative paths, line numbers, and issue numbers. The remaining nit is a citation range, and it under-cites rather than misattributing.

### Code quality

No code. Both documents follow the `docs/spikes/` convention (#822 → ADR 0013 precedent). The judgment calls all still read correctly: ADR 0019 drafted but unfiled (an ADR records a decision; there isn't one yet), the implementation ticket drafted but unlabelled (`crow:auto` would dispatch an agent at unapproved scope), and `Refs #947` rather than `Closes`.

### Summary Table

| Color  | Meaning      | Count | Items |
|--------|--------------|-------|-------|
| Red    | Must fix     | 0     | — |
| Yellow | Should fix   | 0     | — |
| Green  | Consider     | 1     | `:202-204` under-cites the second §5.2 qualification (`:200-201`) |

Both Yellows from the previous round are genuinely fixed, not papered over, and the six Greens were taken further than asked — the 23-read enumeration and the ledger-column de-escalation are both improvements on what I suggested. The one remaining item is a line range in a parenthetical; fix it in passing or leave it.

---

Issue #957, opened 2026-08-10
Review launch breaks when inlined SKILL contains double quotes ($(cat) shell expansion)


dgershman:
## Problem

Cursor (and other agents that inline the full `crow-review-pr` SKILL body) launch review sessions with:

```bash
agent --force --approve-mcps "$(cat .crow-review-prompt.md)"
```

When the prompt file contains literal `"` characters — which the bundled SKILL does (e.g. `"review a PR"`, `--body "YOUR_REVIEW_HERE"`) — **runtime shell expansion breaks**. The inner quotes terminate the outer double-quoted string early, `agent` never starts, and the tail of the prompt is dumped into the terminal scrollback.

Observed this morning on `review-crow-955`: Crow set `reviewPromptDispatched: true` and pasted the launch command into the managed tmux terminal, but the pane ended up showing SKILL markdown and an idle shell prompt — no agent running.

## Reproduction

```bash
PROMPT_PATH='/path/to/.crow-review-prompt.md'
agent --force --approve-mcps "$(cat "$PROMPT_PATH")"
```

Fails with:

```
error: unknown option '---
name: crow-review-pr
...
- All `gh` and `git` commands require `dangerouslyDisableSandbox: true`'
```

(`bash -n` passes — the breakage is at expansion time, not parse time.)

## Affected code

`CursorAgent.autoLaunchCommand` (and the same `$(cat …)` pattern for Codex, Grok, OpenCode, Antigravity review/job launches):

```swift
return "\(agentPath)\(launchArgs) \"$(cat \(CursorLaunchArgs.shellQuote(promptPath)))\"\n"
```

## Expected

Review/job dispatch should survive arbitrary prompt content (quotes, backticks, `$`, newlines) without shell metacharacter breakage.

## Possible fixes

- Pass prompt via file path if the agent CLI supports it
- `base64 -d` / heredoc with a safe delimiter
- Escape the expanded content before passing to the shell (fragile)
- Avoid inlining shell substitution entirely (preferred)

## Related

- Separate from CROW-954 / #955 (workspace `--trust` on review sessions) — that is a second gate even when quoting succeeds.
- CROW-439 guards missing prompt files but not quoting safety.

## Environment

- Crow 0.2.0 (`763211c`)
- Cursor review agent
- Repro: `review-crow-955` clone at `{devRoot}/crow-reviews/crow-pr-955`

[🐦‍⬛ Created with Crow via Cursor](https://github.com/corveil/crow)

PR #967, opened 2026-08-10
Make review verdict severity policy configurable per workspace (CROW-963)

### LINKED ISSUE #963: Make review verdict severity policy configurable per workspace (which of red/yellow/green block)
## Summary

Crow's review verdict rule is fixed for every workspace: **any Red or Yellow finding forces `--request-changes`**, and only an all-Green (or empty) review may approve. That rule lives as prose in the bundled review skill and has no configuration surface at all.

Different codebases want different strictness. A repo where Yellow means "should fix" wants those in the same round trip; a repo where Yellow is closer to "nitpick" wants them reported but not blocking, so PRs aren't bounced on advisory findings. Today the only way to change it is to edit the bundled skill, which changes it for everyone.

Make the blocking set configurable **per workspace**, so e.g. a SecurityScorecard workspace can request changes on Red only, while Yellow findings are still written into the review body but do not gate the verdict.

## Current behaviour

`skills/crow-review-pr/SKILL.md`, Step 5:

> Verdict rule — **only a review whose findings are entirely Green (or empty) may approve**. Any Yellow or Red finding forces `--request-changes`.
>
> - **Approve** (`--approve`): no Red, no Yellow, only Green or no findings.
> - **Request Changes** (`--request-changes`): any Red **or** any Yellow finding.
>
> Yellow findings are "should fix" — the implementing agent will address them as soon as it sees the request-changes verdict, so rejecting on Yellow lands them in the same round trip instead of a follow-up. Comment-only reviews remain forbidden; if uncertain, request changes.

and the summary table the review body must render:

There is **no code-level representation of this policy anywhere** — no enum, no config field, no test. It is entirely instructions to the model.

## Where it would plug in

**The template seam already exists.** The bundled skill is copied into each review clone at `.claude/skills/crow-review-pr/SKILL.md` by `SessionService.swift:3523-3534`, and immediately before writing it runs:

```swift
let skillContent = Scaffolder.bundledReviewSkill()
let resolvedSkillContent = CrowAttribution.expandSkillBody(skillContent, agentKind: reviewAgentKind)
```

`expandSkillBody` already substitutes `{{CROW_AGENT_DISPLAY_NAME}}` (`CrowCore/CrowAttribution.swift:17`). A `{{CROW_REVIEW_VERDICT_POLICY}}` placeholder resolved from the session's workspace is the same mechanism, at the same point.

**Both launch paths need it.** Claude/Grok read the copied `SKILL.md`; Cursor gets the skill body expanded inline into the command via `cursorReviewPrompt(skillBody:prURL:)`. Whatever renders the policy has to run before both, not just the file-copy path.

**Config model:** `WorkspaceInfo` — `Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift:488`, alongside `autoReviewRepos` / `excludeReviewRepos` / `customInstructions` (`:494-497`). New field must be optional so existing `config.json` files keep decoding.

**Surfaces that need to follow**, matching how every other workspace field is done:
- `crow workspace add` / `crow workspace edit` flags, echoed by `workspace get` / `workspace list`
- Settings → Workspaces tab
- The `crow` CLI reference section in the repo's `CLAUDE.md`

## Proposed shape

A per-workspace set of severities that force `--request-changes`:

```jsonc
{
  "name": "SecurityScorecard",
  "provider": "github",
  "reviewBlockingSeverities": ["red"]     // default: ["red", "yellow"]
}
```

```bash
crow workspace edit --workspace SecurityScorecard --review-blocking-severity red
crow workspace edit --workspace SecurityScorecard --clear-review-blocking-severities   # back to default
```

Rendering into the skill would replace both the prose rule and the summary table's "Verdict effect" column, so the review body the agent produces stays self-consistent with the policy it was given.

Absent/unset must mean **today's behaviour** (`["red", "yellow"]`), not "nothing blocks" — an install that never touches this setting should see no change.

## Open questions

1. **Does "configurable colors" mean the blocking set, or the taxonomy itself?** This ticket assumes the former: the three severities stay fixed and you choose which of them gate the verdict. Letting a workspace rename severities or add a fourth is a much larger change (the skill's output format, the summary table, and any future parsing all assume Red/Yellow/Green) and should be its own ticket if wanted.

2. **Workspace-only, or repo-level too?** The ask is per-workspace. But `autoReviewRepos` / `excludeReviewRepos` set a precedent for per-repo lists hanging off a workspace, and "strict on the core service, relaxed on the docs repo" is a plausible want. Decide whether to design the field so repo-level can be added later without a breaking shape change.

3. **What happens when nothing blocks?** `reviewBlockingSeverities: []` means every review approves regardless of findings. Allow it, warn on it, or reject it? Note the skill also forbids comment-only reviews, so there is no "report without a verdict" option — an empty blocking set genuinely means auto-approve.

4. **Non-blocking findings must still be reported.** Whatever renders, a Yellow that no longer gates must still appear in the review body. Losing the finding entirely would be worse than the current over-strictness.

5. **This is advisory, not enforcement.** The review agent runs `gh pr review` itself; Crow does not intercept the call and cannot validate that the verdict matches the policy it handed over. This ticket configures a *prompt*. If actual enforcement is wanted — Crow checking the posted verdict against the findings — that is a separate and considerably harder piece of work. Worth stating plainly in the PR so nobody reads the setting as a guarantee.

## Downstream effects to check

- **Auto-merge.** Relaxing Yellow means more PRs reach an approved state, which feeds `autoMergeWatcherEnabled` and the `crow:merge` gate. A workspace that sets `["red"]` will auto-merge things it previously would not have. Not a blocker — it is the point of the setting — but it should be a deliberate, documented consequence rather than a surprise.
- **`respondToChangesRequested`.** Fewer changes-requested verdicts means fewer auto-respond cycles. Expected, worth confirming nothing assumes the old rate.
- **#957.** `$(cat .crow-review-prompt.md)` shell quoting already breaks when the inlined skill body contains `"` characters. Generated policy prose is new text flowing into that same path — keep it quote-free, or land #957 first.

## Acceptance criteria

- [ ] A workspace can declare which severities force `--request-changes`; unset behaves exactly as today (`red` + `yellow`).
- [ ] The policy reaches the review agent on **all** launch paths — Claude/Grok (copied `SKILL.md`) and Cursor (inlined body).
- [ ] The rendered skill is internally consistent: the prose rule and the summary table's "Verdict effect" column both reflect the configured policy.
- [ ] Non-blocking findings are still reported in the review body; only the verdict changes.
- [ ] Settable via `crow workspace add`/`edit`, readable via `workspace get`/`list`, editable in Settings → Workspaces.
- [ ] Existing `config.json` files without the field decode unchanged.
- [ ] Documented in the repo `CLAUDE.md` CLI reference alongside the other workspace field flags.
- [ ] A test asserts the rendered skill body differs between a default workspace and a `["red"]` workspace, and that the default matches the current bundled text.

dhilgaertner:
Closes #963

## Summary

Crow's review verdict rule — **any Red or Yellow finding forces `--request-changes`** — was prose hardcoded in the bundled `crow-review-pr` skill, with no code-level representation and no configuration surface. The only way to change it was to edit the skill, which changed it for every workspace.

This adds `WorkspaceInfo.reviewBlockingSeverities`, rendered into the skill body by a new `ReviewVerdictPolicy`. A repo where Yellow means "nitpick" can now request changes on Red only; a repo where Yellow means "should fix" keeps blocking on both.

```bash
crow workspace edit --workspace SecurityScorecard --review-blocking-severity red
crow workspace edit --workspace SecurityScorecard --clear-review-blocking-severities  # back to default
```

## The part that matters: both launch paths

Rendering happens **once in `prepareReviewClone`, before the two review launch paths diverge**, so the policy reaches:

- **Claude** — reads the copied `.claude/skills/crow-review-pr/SKILL.md`
- **Cursor / Codex / OpenCode / Grok / Antigravity** — read the whole body inlined into `.crow-review-prompt.md`

Hooking only the file copy would have passed a naive test and done nothing for most installs, since `agentsByKind.review` is commonly Cursor. Folding the expansion into `CrowAttribution.expandSkillBody` behind a defaulted parameter keeps the other five skill-scaffolding call sites rendering the default, so no `{{CROW_REVIEW_VERDICT_*}}` placeholder can reach disk unexpanded. The expansion is idempotent, so the second pass over an already-expanded body is a no-op rather than a conflicting re-render.

Workspace resolution is **by repo slug, not worktree path** — a review clone lives at `{devRoot}/crow-reviews/…`, whose first path component matches no workspace. That is the same trap that silently unset the gateway for every review in CROW-891.

## Decisions on the ticket's open questions

1. **Blocking set, not taxonomy.** Red/Yellow/Green are unchanged.
2. **Workspace-only.** A flat array. Repo-level can be added later as a separate additive key (`reviewBlockingSeveritiesByRepo`) with no breaking shape change — the same way `autoReviewRepos` and `excludeReviewRepos` are separate lists.
3. **An empty blocking set is rejected.** Nothing gating the verdict approves every review, and with `autoMergeWatcherEnabled` on, merges it. `--clear-review-blocking-severities` (which removes the key and restores red + yellow) is what "no custom policy" means. The web form refuses to uncheck the last severity.
4. **Non-blocking findings are still reported.** A severity that stops gating reads `Approve allowed` in the table — never anything suggesting it may be dropped — and the relaxed prose says so explicitly. The rule text and the table's "Verdict effect" column are generated from the same policy, so the brief cannot contradict itself.
5. **Advisory, not enforcement.** The agent invokes `gh pr review` itself. Crow renders the policy into the prompt it hands over, never sees the call, and **cannot validate the verdict actually posted**. No verdict validation or post-hoc checking is included, deliberately — that is a separate and considerably harder piece of work.

## Default-preserving

Unset means today's behaviour (`red` + `yellow`), **not** "nothing blocks". `--clear-*` **removes** the key rather than writing a null — absent, `null` and `[]` all round-trip differently, and one null makes the whole `config.json` undecodable (the `crow agents set --clear` precedent).

Decoding stays deliberately lenient in the other direction: an unknown severity in a hand-edited config is dropped and falls back to the default rather than throwing, because a throwing `WorkspaceInfo` decode makes `ConfigStore.loadConfig` return nil, at which point every writer's `?? AppConfig()` fallback overwrites `config.json` with defaults. Rejection lives on the write path (CLI / RPC / web), never in the decoder.

`ReviewVerdictPolicyTests` pins the default rendering **byte-for-byte** against the pre-change skill text. I also verified the whole chain end to end: substituting the golden blocks back into the templatized skill reproduces the original file exactly.

## Surfaces

`crow workspace add|edit --review-blocking-severity red|yellow|green` (repeatable, parsed into the model enum so a typo is a parse error and `--help` lists the legal values), `workspace get`/`list` echoing the effective list plus `review_blocking_severities_explicit`, Settings → Workspaces → Review verdict, `ParityLedger`, `CLAUDE.md` + template, `docs/configuration.md`, `docs/cli-reference.md`, `docs/cli.md`, `CHANGELOG.md`.

## Notes

- **#957 is not bundled** — it is already fixed on this branch by f44ace6 (`ShellLaunchArgs.evalPromptLaunch`, `printf '%q'` + `eval`), so the quoting hazard for the inlined path is resolved. Generated policy prose is kept quote-free as cheap hygiene regardless.
- **Auto-merge consequence:** a workspace set to `["red"]` reaches an approved state more often, so `autoMergeWatcherEnabled` + `crow:merge` will merge things it previously would not have. That is the point of the setting, but it should be deliberate.
- **New drift gate.** `skills/crow-review-pr/SKILL.md` and `Resources/crow-review-pr-SKILL.md.template` are byte-identical copies (dev builds read the first, release builds the second) with nothing guarding them. `check-workspace-custom-instructions.sh` now diffs them, so a policy edit landing in only one half fails `make parity`. Scoped to `crow-review-pr` on purpose: **the same gate applied to all five skills reveals a pre-existing drift in `crow-batch-workspace`** (`settings.json` vs `settings.local.json` in its Autonomous Execution section), which is a separate fix.
- `handoffAgent` does not rewrite `.crow-review-prompt.md`, so a review handed off to another agent keeps the policy it launched with. Pre-existing and structural, not changed here.

## Test plan

- [x] `make daemon` — clean
- [x] `make test` — all 17 packages pass except CrowDaemon's two known failures (`refreshTerminalsRebindsTheActiveTerminalToTheFreshRow`, `bootCatchResetsTheHistory`), **confirmed failing identically on `origin/main`** in a scratch worktree
- [x] `make parity` — notification catalogs, CLI/RPC parity (97 methods), no-NSLog, and the new skill drift gate all pass
- [x] `make docs` — `docs/cli.md` regenerated; the flag's legal values are picked up automatically from the enum
- [x] `ParityGateTests` — the new field has a ledger row and a populated probe
- [x] Byte-identity of the default rendering verified against the pre-change `SKILL.md`
- [ ] Manual: restart `crowd`, set a workspace to `red`, start a review, confirm `.crow-review-prompt.md` (the inlined path, not the copied skill) carries the red-only rule


dgershman (approved):
## Code & Security Review

### Critical Issues

None.

### Verification performed

| Check | Result |
|---|---|
| `swift build` | clean (92s) |
| `swift test --package-path Packages/CrowCore` | 569 tests, 23 suites — pass |
| `swift test --package-path Packages/CrowEngine` | 691 tests, 73 suites — pass |
| `swift test --package-path Packages/CrowCLI` | 347 tests, 2 suites — pass |
| `scripts/check-workspace-custom-instructions.sh` | `OK` (incl. the new drift gate + placeholder checks) |

I could not reproduce the two `CrowDaemon` failures listed in the test plan because they live outside the three packages the change touches; the PR states they reproduce identically on `origin/main`, and nothing here touches that target.

### Security Review

**Strengths:**

- **No injection surface into the launched prompt.** The rendered policy blocks are derived entirely from a closed three-value enum plus static string literals — `ReviewVerdictPolicy.swift:118-159` interpolates only `displayName`/`meaning`, never config text. A hand-edited `config.json` cannot get arbitrary text into `.crow-review-prompt.md` or the copied `SKILL.md`, which matters because that body is later shell-interpolated at launch. Parsing to the enum *before* rendering is the right boundary, and keeping the prose quote-free is sensible hygiene on top.
- **The lenient-decode / strict-write asymmetry is correct and correctly justified.** `AppConfig.swift:631-644` decodes via `[String]` and drops unknowns rather than throwing; a throwing `WorkspaceInfo` decode would make `ConfigStore.loadConfig` return nil and let every writer's `?? AppConfig()` fallback flatten `config.json`. Rejection lives on the write path (`WorkspaceRPCSupport.swift:174-196`) where a bad value can be reported instead of being catastrophic. `workspaceReviewBlockingSeveritiesDecodeIsLenient` pins all four cases (unknown, whitespace/case, order, empty).
- **Fail-safe defaults throughout.** Absent → `[red, yellow]`, unknown → default, empty → default, no workspace claiming the repo → default. There is no input to the decode path that yields the "nothing blocks" posture.
- **Empty set rejected atomically**, with an error that names the affordance that does what the caller meant. `applyPatchRejectsAnEmptyReviewBlockingSet` asserts the target is unmodified after the throw.
- **Auto-merge blast radius is called out** in the PR body, `CLAUDE.md`, `docs/cli-reference.md` and `docs/configuration.md` rather than left implicit — relaxing to `["red"]` widens what `autoMergeWatcherEnabled` + `crow:merge` will merge. Documenting that at every surface is the right call for a setting whose downside is invisible at the point of use.

**Concerns:** none rising to Red or Yellow.

### Code Quality

The load-bearing design decision — rendering in `prepareReviewClone` *before* the launch paths diverge (`SessionService.swift:3555-3559`) — is correct and is the part a naive implementation gets wrong. Hooking only the file copy would have left every inlining agent (Cursor, Codex, OpenCode, Grok, Antigravity) on the default rule, and `agentsByKind.review` is commonly Cursor, so that would have been the *normal* case rather than an edge one. I traced both consumers: the inlined path reaches `cursorReviewPrompt` → `expandSkillBody` (`SessionService.swift:3916-3920`) and the Claude path reaches `expandSkillBody` at `:3583`; both re-expand with the default, and idempotence makes that a no-op. `expandIsIdempotentSoThePreExpandedReviewBodySurvivesASecondPass` pins exactly that, including the negative (`!twice.contains("any Red **or** any Yellow finding")`).

Workspace resolution by repo slug rather than worktree path (`:2893-2895`) is the right fix for the CROW-891 trap, and uses the same `config.workspace(forRepoSlug:)` helper as the auto-create path at `:1959` — consistent, not a one-off.

Storing `defaultRuleBlock` verbatim instead of regenerating it is the correct trade. The historical text is internally inconsistent (`Yellow or Red` in the opening sentence, `Red **or** Yellow` in the bullet), so a generator matching it would have been contorted to reproduce a quirk — and the byte-for-byte golden tests are what actually protect the no-op-for-existing-installs guarantee.

The new drift gate is a genuine improvement beyond the ticket: the two halves of `crow-review-pr` were byte-identical copies with nothing guarding them, and this change is precisely the kind that could have shipped to only one. Scoping it to `crow-review-pr` because extending it would surface a pre-existing `crow-batch-workspace` drift is the right instinct — the gate lands green rather than being pre-broken.

Three minor observations, none blocking:

**Green — the memberwise init doesn't collapse `[]` to nil, unlike the decoder.** `AppConfig.swift:609` is `reviewBlockingSeverities.map(ReviewSeverity.canonicalize)`, so `WorkspaceInfo(…, reviewBlockingSeverities: [])` stores `[]`. `effectiveReviewBlockingSeverities` then returns `[]` (the `??` never fires), and `ReviewVerdictPolicy.ruleBlock(blocking: [])` renders a degenerate brief — `only a review with no  findings may approve. Any  finding forces --request-changes` — i.e. the exact "nothing blocks" posture rejected everywhere else, with a double space where the severity list should be. Unreachable today: I grepped every non-test reference and no production call site passes a non-nil value, the decoder maps empty→nil, `decodeReviewBlockingSeverities` rejects empty, and the web form refuses to uncheck the last box. Worth closing anyway since the field's own doc comment promises rejection "at every write boundary" and the init is a public one — either `.flatMap { $0.isEmpty ? nil : ReviewSeverity.canonicalize($0) }` in the init (mirroring `init(from:)`), or a `guard !canonical.isEmpty else { return defaultRuleBlock }` in `ruleBlock`/`notesBullets` so the renderer cannot emit a degenerate brief regardless of caller.

**Green — a settings.js comment overstates the server-side guard for that path.** The new comment says "The server rejects an empty set too", but the web saves the whole config via `set-config` (`settings.js:230`), which goes through the deliberately lenient `WorkspaceInfo.init(from:)` and maps `[]` → nil → default. The rejection lives on the `workspace edit` RPC path, which the web form doesn't use. The behaviour is safe either way (default, never "nothing blocks"), so this is comment accuracy only — but the comment is the reason a future reader would trust the client-side guard is redundant, and on this path it isn't.

**Green — `tableRows` doesn't canonicalize its input** while its two sibling renderers do (`ReviewVerdictPolicy.swift:145` uses the raw `blocking` for `contains`). Order-independent, so behaviour is identical and `policyRenderingIsIndependentOfInputOrder` covers it; canonicalizing there too would just make the three public renderers uniform.

### Test Coverage

Unusually strong for a change of this shape. Beyond the golden default rendering, the suite covers the column-width invariant (53 chars, matching the static separator), the all-blocking branch and its absent advisory tier, input-order independence, idempotence with its negative assertion, both halves of the bundled skill carrying every placeholder, end-to-end expansion of the *real* skill file, lenient decode across four malformed shapes, empty/unknown rejection with a no-partial-write assertion, `hasAnyField` for both new keys, and the `explicit` flag distinguishing inherited from pinned. `ParityLedger` row and `ParityGateTests` probe are both present.

### Summary Table

The three Green items are unreachable-today hardening and comment accuracy; none changes behaviour on any shipped path. The default-preservation guarantee is verified byte-for-byte, both review launch paths carry the policy, and the write/decode asymmetry is deliberate and well-reasoned.

---

PR #958, opened 2026-08-10
Fix review launch quoting when SKILL body contains double quotes

dgershman:
Closes #957

## Summary

Review and job sessions inlined the prompt via `"$(cat .crow-*-prompt.md)"`, which breaks when the bundled `crow-review-pr` SKILL body contains literal `"` characters (e.g. `--body "YOUR_REVIEW_HERE"`). The shell terminates the outer double-quoted string early and `agent` never starts.

This adds `ShellLaunchArgs.evalPromptLaunch` in CrowCore: read the prompt file into `_CROW_P`, then `eval` the agent command with `printf '%q'` so arbitrary prompt content (quotes, backticks, `$`, newlines) survives paste into zsh/bash. All agents that used `$(cat …)` (Cursor, Claude, Codex, OpenCode, Antigravity) and the `/crow-workspace` setup script now share this helper. Grok already used `--prompt-file` and is unchanged.

## Test plan

- [x] `ShellLaunchArgsTests.evalPromptLaunchSurvivesQuotesAndDollars` executes the launch snippet in bash and verifies the full prompt is passed through
- [x] Agent package tests updated to assert `eval` form and no `$(cat`
- [ ] Rebuild/restart `crowd`, start a review on a PR whose inlined SKILL has `"` in the body, confirm `agent` launches instead of dumping markdown into scrollback

Made with [Cursor](https://cursor.com)

PR #970, opened 2026-08-11
Strip SKILL frontmatter from inlined review prompts (CROW-968)

### LINKED ISSUE #968: Cursor review sessions fail to launch: SKILL frontmatter parsed as a CLI option
## Summary

Cursor review sessions fail to launch. The `agent` CLI rejects the inlined review
prompt as a command-line option:

```
error: unknown option '---
name: crow-review-pr
description: >-
  Perform a comprehensive code and security review on a GitHub pull request,
…
```

Both currently-active review sessions are dead on arrival:

- `review-corveil-2170` (`53D18B11-0F12-466B-BDB2-A575A1FAC9C4`)
- `review-corveil-cloud-terraform-378` (`B7E9B52E-AB55-4A06-9CA9-D91AFEF8B4E3`)

## Root cause

`SessionService.cursorReviewPrompt` inlines the **entire** `crow-review-pr` SKILL.md
body into `.crow-review-prompt.md`, including its YAML frontmatter. The file therefore
begins with `---`:

```
$ head -c 4 .crow-review-prompt.md | xxd
00000000: 2d2d 2d0a                                ---.
```

`CursorAgent.autoLaunchCommand` passes that file's contents as a **bare positional
argument** via `ShellLaunchArgs.evalPromptLaunch`:

```bash
_CROW_P=$(< '…/.crow-review-prompt.md'); eval "'…/agent' --trust --force --approve-mcps $(printf '%q' "$_CROW_P")"
```

`printf %q` protects the string from the *shell*, but leading hyphens are not shell-special,
so they survive verbatim:

```
$ printf 'quoted: %q\n' "$(printf -- '---\nname: x\n')"
quoted: ---$'\n'name:\ x
```

`agent` is commander-based (`Usage: agent [options] [command] [prompt...]`) and parses any
argv element starting with `-` as an option. There is no `--` end-of-options separator in
`evalPromptLaunch`, so the prompt is parsed as a flag and the process exits before starting.

## On the "trust fix broke it" theory

Worth correcting: **CROW-954 is not the cause.** The frontmatter has been in SKILL.md for a
long time (it predates #778), and `evalPromptLaunch` has never emitted a `--` separator.
The `--trust` flag has no bearing on positional parsing.

What changed is that two fixes landed ten minutes apart on 2026-08-10 and between them
removed the *earlier* failure modes that were masking this one:

- `f44ace6` (#958, 13:25) replaced `"$(cat …)"` with the `eval` + `printf %q` form, so the
  launch now assembles correctly and actually reaches `agent`'s argument parser.
- `3626e63` (CROW-954, 13:35) seeded `--trust` on review launches, so sessions no longer
  stop at the workspace-trust dialog first.

Net effect: the launch now gets far enough for the argument parser to reject it. The defect
itself is older than both.

## Scope

Affects every agent that receives the **inlined** SKILL body — `cursor`, `openCode`,
`codex`, `grok`, `antigravity` (the `buildReviewPrompt` non-default branch).

**Claude Code is unaffected** — it gets the terse `/crow-review-pr <URL>` form, which has no
leading hyphen.

Job sessions (`.crow-job-prompt.md`) are only affected if a user's job prompt happens to
begin with `-`.

## Proposed fix

**1. Strip YAML frontmatter when inlining the SKILL body** *(primary)*

In `cursorReviewPrompt`, drop a leading `---\n … \n---\n` block before substitution. This is
correct independent of the parsing bug: the frontmatter is skill-engine metadata (`name`,
`description`) that means nothing to an inlined prompt, and it burns tokens on every review.

**2. Emit a `--` end-of-options separator** *(defense in depth)*

Add an opt-in `endOfOptions` parameter to `ShellLaunchArgs.evalPromptLaunch` so any prompt
beginning with `-` survives. Do **not** flip this on globally without checking each CLI —
`evalPromptLaunch` is shared by Claude, Codex, OpenCode, and Antigravity, and not every
parser treats a bare `--` the same way. Verify per-agent before enabling.

## Test plan

- [ ] Unit test: `cursorReviewPrompt` output does not begin with `---` and retains the body.
- [ ] Unit test: frontmatter stripping is a no-op on a body that has none.
- [ ] Unit test: `evalPromptLaunch(endOfOptions: true)` emits `--` immediately before the prompt.
- [ ] Verify empirically that `agent … -- '<prompt>'` parses (commander should honor `--`, but
      this was not confirmed against the installed binary while diagnosing).
- [ ] Rebuild/restart `crowd`, start a review, confirm `agent` launches and posts a verdict.
- [ ] Relaunch the two stuck sessions above and confirm both complete.

## Cleanup

The two active review sessions will need relaunching (or deleting and re-kicking from the
Reviews board) once the fix lands — their worktrees still hold the bad prompt file.

dhilgaertner:
Closes #968

## The bug

Cursor review sessions never started — `agent` exited during argument parsing:

```
error: unknown option '---
name: crow-review-pr
…
```

`SessionService.cursorReviewPrompt` inlined the **entire** `crow-review-pr` SKILL body — YAML frontmatter included — into `.crow-review-prompt.md`, so the file's first bytes were `---\n`. `CursorAgent.autoLaunchCommand` hands those contents to `agent` as a bare positional argument, and `agent` is commander-based (`Usage: agent [options] [command] [prompt...]`), so it claimed the leading hyphens as a flag.

Shell quoting never covered this: `printf %q` protects the string from the *shell*, but a leading `-` is not shell-special and survives into argv.

Affected every harness that receives the inlined body — `cursor`, `openCode`, `codex`, `grok`, `antigravity`. Claude Code was unaffected (terse `/crow-review-pr <URL>` form, no leading hyphen).

As the ticket notes, the defect predates CROW-954 and #958 — `evalPromptLaunch` has never emitted a `--`, and the frontmatter has been in SKILL.md far longer. Those two commits only removed the earlier failures that were masking this one.

## The fix

**1. Strip the frontmatter when inlining** — new `MarkdownFrontmatter.stripped` in CrowCore. Conservative: strips only when line 1 is exactly `---` and a closing `---` exists, returns its input untouched otherwise. The no-op path is the common one (`Scaffolder.bundledReviewSkill()`'s built-in fallback and the test fixtures carry no frontmatter), so it must never damage its input.

Two placement decisions worth reviewing:

- **In `cursorReviewPrompt`, below the fork in `prepareReviewClone`.** That function renders the workspace verdict policy into one body and forks it two ways; the other consumer is the clone's `.claude/skills/crow-review-pr/SKILL.md`, which **must keep** its frontmatter or Claude Code won't load the skill. Hoisting the strip up to `policySkillBody` would fix Cursor and break Claude. There's a comment at that call site saying so.
- **In CrowCore, not CrowEngine** — CrowEngine isn't in the Linux PR-CI package list (`.github/workflows/ci.yml`), so engine tests don't run on PRs. CrowCore's do.

The strip is right independent of the parsing bug: the metadata has no reader in an inlined brief and costs tokens on every review.

**2. Opt-in `--` end-of-options separator** — `ShellLaunchArgs.evalPromptLaunch` gains an `endOfOptions` flag, enabled at Cursor's two call sites only (`CursorAgent.autoLaunchCommand`, `CursorLauncher.launchCommand`).

Opt-in per harness on purpose, per the ticket: `--` is a convention, not a guarantee, and this helper is shared by Claude, Codex, OpenCode and Antigravity. Only Cursor's parser was verifiable here, so only Cursor opts in; everyone else keeps the `false` default and is source-compatible. Grok is unaffected either way — it passes `--prompt-file`, so its prompt never becomes argv.

## The second commit is worth a look

The first push went green on macOS and **failed on Linux CI**, on a test in this PR. It was a real bug, not a flaky test.

Swift treats `"\r\n"` as a single grapheme cluster, and Foundation's grapheme-aware `components(separatedBy: "\n")` differs by platform: Darwin splits on the embedded `\n`, swift-corelibs-foundation does not. On Linux the whole file came back as one "line", `lines.first` was never `---`, and the guard returned the body untouched — meaning **every Cursor review on a CRLF checkout would have kept its frontmatter and reproduced this exact bug**, on the platform the daemon isn't developed on.

`5d8859f` replaces the line split with a scan over `body.utf8`: walk to each `\n`, drop a trailing `\r` from the line's content but not from the offsets, slice the original bytes at a line boundary. Byte comparisons behave identically everywhere. Behaviour is otherwise unchanged.

## Verification

The ticket's one unverified checkbox, settled against the installed `cursor-agent 2026.08.04`:

| Invocation | Result |
|---|---|
| `agent --list-models --bogus-flag-xyz` | `error: unknown option '--bogus-flag-xyz'` |
| `agent --list-models -- --bogus-flag-xyz` | parses clean |

Against the real `skills/crow-review-pr/SKILL.md`: first bytes go from `2d 2d 2d 0a` (`---\n`, the exact signature in the ticket) to `23 20 43 72` (`# Cr`), 162 → 154 lines, with `$ARGUMENTS`, the attribution footer and the body tail all intact.

Run in the `swift:6.1` container CI uses, not just on macOS: **CrowCore 582** and **CrowCursor 69** green, CrowEngine builds clean.

New tests: 10 in `MarkdownFrontmatterTests` (strip, five no-op shapes, CRLF, LF/CRLF equivalence, delimiter strictness, later-horizontal-rules-survive), 3 in `ShellLaunchArgsTests` (separator placement, default output byte-identical to before, and a real `/bin/bash` round-trip proving `--` lands as its own argv element ahead of a `---`-leading body), 3 in `SessionServiceReviewPromptTests` (including a loop asserting no inlining harness emits a leading `-`), 1 in `CursorAgentTests` (the separator rides prompt dispatch only, not resume or `.work`) and 1 in `CursorLauncherTests`.

## Pre-existing failures, not from this PR

`make test` is red on three tests that reproduce identically on clean `main` with these changes stashed: `bootCatchResetsTheHistory` and `refreshTerminalsRebindsTheActiveTerminalToTheFreshRow` (CrowDaemon), and `retryReadinessEmitsTimedOutWhenSentinelMissing` (CrowTerminal, a real-tmux integration test). None are touched here. Worth noting separately that `make test`'s `for` loop doesn't abort on failure, so its exit code masks mid-loop failures — these have to be found by grepping the log.

## Not in this PR

- **End-to-end confirmation.** Steps 5–6 of the ticket's test plan (rebuild/restart `crowd`, run a review to a posted verdict) need a daemon restart, so they're left to the maintainer.
- **The two stuck sessions.** `review-corveil-2170` (`53D18B11-…`) and `review-corveil-cloud-terraform-378` (`B7E9B52E-…`) still hold the bad prompt file and need relaunching — or deleting and re-kicking from the Reviews board — once this is built and `crowd` restarted. Left alone deliberately, since `crow delete-session` is destructive.

## Incidental

The `docs/agent-harness-matrix.md` bullets in the section I edited still showed the pre-#957 `$(cat …)` launch notation, which `evalPromptLaunch` stopped emitting when it moved to `$(< path)` + `printf %q` — and which `CursorAgentTests` explicitly asserts is absent. Refreshed so the section doesn't contradict the paragraph this PR adds.

### DIFF
--- a/Packages/CrowCore/Sources/CrowCore/MarkdownFrontmatter.swift
+++ b/Packages/CrowCore/Sources/CrowCore/MarkdownFrontmatter.swift
@@ -23,8 +23,6 @@ import Foundation
 /// checked-in `skills/crow-review-pr/SKILL.md` and its bundled `.template`
 /// (`scripts/check-workspace-custom-instructions.sh` enforces that).
 public enum MarkdownFrontmatter {
-    private static let delimiter = "---"
-
     /// Drop a leading YAML frontmatter block, returning the body that follows.
     ///
     /// Deliberately conservative — the inputs include bodies that legitimately
@@ -43,25 +41,70 @@ public enum MarkdownFrontmatter {
     /// document that opens with a markdown horizontal rule and contains a second
     /// one — indistinguishable from frontmatter by any line-based parser, and not
     /// a shape a SKILL body takes.
+    ///
+    /// Scans the UTF-8 view rather than `components(separatedBy: "\n")`. Swift
+    /// treats `"\r\n"` as a **single** grapheme cluster, and Foundation's
+    /// grapheme-aware search differs by platform: Darwin splits a CRLF document
+    /// on the embedded `\n`, swift-corelibs-foundation does not, so on Linux the
+    /// whole file came back as one "line" and nothing was ever stripped. Bytes
+    /// behave identically everywhere.
     public static func stripped(_ body: String) -> String {
-        let lines = body.components(separatedBy: "\n")
-        guard let first = lines.first, isDelimiter(first) else { return body }
+        let bytes = Array(body.utf8)
+
+        guard let opening = line(in: bytes, at: 0), isDelimiter(bytes, opening.content) else {
+            return body
+        }
+
         // Search from line 2: the opening delimiter must not close itself.
-        // `dropFirst()` yields a slice sharing `lines`' indices, so the returned
-        // index addresses `lines` directly.
-        guard let closing = lines.dropFirst().firstIndex(where: isDelimiter) else { return body }
+        var cursor = opening.next
+        while let current = line(in: bytes, at: cursor) {
+            guard isDelimiter(bytes, current.content) else {
+                cursor = current.next
+                continue
+            }
+            // Skip blank lines after the closing delimiter so the result starts
+            // at the body's first real character.
+            var start = current.next
+            while let following = line(in: bytes, at: start), isBlank(bytes, following.content) {
+                start = following.next
+            }
+            // `start` is always a line boundary, so the slice is valid UTF-8.
+            return String(decoding: bytes[start...], as: UTF8.self)
+        }
 
-        return lines[lines.index(after: closing)...]
-            // `.whitespacesAndNewlines`, not `.whitespaces`: splitting a CRLF
-            // document on `\n` makes a blank line the single character `\r`,
-            // which `.whitespaces` (space + tab) does not match.
-            .drop { $0.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty }
-            .joined(separator: "\n")
+        return body
     }
 
-    /// A line that is exactly `---`, tolerating a CRLF carriage return. Splitting
-    /// on `\n` leaves the `\r` attached to every line of a CRLF document.
-    private static func isDelimiter(_ line: String) -> Bool {
-        (line.hasSuffix("\r") ? String(line.dropLast()) : line) == delimiter
+    /// The line beginning at `offset`: its content without the line terminator,
+    /// and the offset the next line starts at. `nil` once `offset` is past the end.
+    private static func line(
+        in bytes: [UInt8], at offset: Int
+    ) -> (content: Range<Int>, next: Int)? {
+        guard offset < bytes.count else { return nil }
+
+        var end = offset
+        while end < bytes.count, bytes[end] != Self.lineFeed { end += 1 }
+
+        // Drop a CRLF's carriage return from the content, not from the offsets.
+        var contentEnd = end
+        if contentEnd > offset, bytes[contentEnd - 1] == Self.carriageReturn { contentEnd -= 1 }
+
+        return (offset..<contentEnd, end < bytes.count ? end + 1 : bytes.count)
     }
+
+    /// A line whose content is exactly `---`.
+    private static func isDelimiter(_ bytes: [UInt8], _ content: Range<Int>) -> Bool {
+        content.count == 3 && bytes[content].allSatisfy { $0 == Self.hyphen }
+    }
+
+    /// A line with nothing but horizontal whitespace on it.
+    private static func isBlank(_ bytes: [UInt8], _ content: Range<Int>) -> Bool {
+        bytes[content].allSatisfy { $0 == Self.space || $0 == Self.tab }
+    }
+
+    private static let lineFeed: UInt8 = 0x0A
+    private static let carriageReturn: UInt8 = 0x0D
+    private static let hyphen: UInt8 = 0x2D
+    private static let space: UInt8 = 0x20
+    private static let tab: UInt8 = 0x09
 }
--- a/Packages/CrowCore/Sources/CrowCore/MarkdownFrontmatter.swift
+++ b/Packages/CrowCore/Sources/CrowCore/MarkdownFrontmatter.swift
@@ -1,0 +1,67 @@
+import Foundation
+
+/// Leading YAML frontmatter (`---` … `---`) handling for markdown bodies Crow
+/// inlines into agent prompts.
+///
+/// Crow ships the `crow-review-pr` instructions as a SKILL file, which carries
+/// `name`/`description` frontmatter because Claude Code's skill engine requires
+/// it. Agents without a slash-command engine (Cursor, OpenCode, Codex, Grok,
+/// Antigravity) get that body **inlined** into `.crow-review-prompt.md` instead,
+/// where the frontmatter is worse than useless:
+///
+/// - It is skill-engine metadata the inlined brief has no reader for, and it
+///   costs tokens on every review.
+/// - It makes the prompt's first byte a `-`, so a CLI that parses its positional
+///   prompt with an option parser reads it as a flag and exits. Cursor's
+///   commander-based `agent` did exactly that — `error: unknown option '---`
+///   (CROW-968). Shell quoting does not save you here: `printf %q` protects the
+///   string from the *shell*, but a leading hyphen is not shell-special, so it
+///   reaches the binary's argv intact.
+///
+/// Only the **inlined** copy is stripped. The `.claude/skills/crow-review-pr/
+/// SKILL.md` Crow writes into a review clone keeps its frontmatter, as do the
+/// checked-in `skills/crow-review-pr/SKILL.md` and its bundled `.template`
+/// (`scripts/check-workspace-custom-instructions.sh` enforces that).
+public enum MarkdownFrontmatter {
+    private static let delimiter = "---"
+
+    /// Drop a leading YAML frontmatter block, returning the body that follows.
+    ///
+    /// Deliberately conservative — the inputs include bodies that legitimately
+    /// have no frontmatter (`Scaffolder.bundledReviewSkill()`'s built-in fallback,
+    /// every test fixture), so the no-op path is the common one and must never
+    /// damage its input:
+    ///
+    /// - The **first** line must be exactly `---`. Frontmatter is column-1 only,
+    ///   so an indented `---` is left alone.
+    /// - A closing `---` line must exist. Without one this is not a frontmatter
+    ///   block, and the input is returned unchanged rather than consumed whole.
+    /// - Leading blank lines after the closing delimiter are dropped, so the
+    ///   result starts at the body's first real character.
+    ///
+    /// CRLF-terminated lines are recognized. The one accepted false positive is a
+    /// document that opens with a markdown horizontal rule and contains a second
+    /// one — indistinguishable from frontmatter by any line-based parser, and not
+    /// a shape a SKILL body takes.
+    public static func stripped(_ body: String) -> String {
+        let lines = body.components(separatedBy: "\n")
+        guard let first = lines.first, isDelimiter(first) else { return body }
+        // Search from line 2: the opening delimiter must not close itself.
+        // `dropFirst()` yields a slice sharing `lines`' indices, so the returned
+        // index addresses `lines` directly.
+        guard let closing = lines.dropFirst().firstIndex(where: isDelimiter) else { return body }
+
+        return lines[lines.index(after: closing)...]
+            // `.whitespacesAndNewlines`, not `.whitespaces`: splitting a CRLF
+            // document on `\n` makes a blank line the single character `\r`,
+            // which `.whitespaces` (space + tab) does not match.
+            .drop { $0.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty }
+            .joined(separator: "\n")
+    }
+
+    /// A line that is exactly `---`, tolerating a CRLF carriage return. Splitting
+    /// on `\n` leaves the `\r` attached to every line of a CRLF document.
+    private static func isDelimiter(_ line: String) -> Bool {
+        (line.hasSuffix("\r") ? String(line.dropLast()) : line) == delimiter
+    }
+}
--- a/Packages/CrowCore/Sources/CrowCore/ShellLaunchArgs.swift
+++ b/Packages/CrowCore/Sources/CrowCore/ShellLaunchArgs.swift
@@ -16,8 +16,29 @@ public enum ShellLaunchArgs {
     /// macOS bash's `printf %q` emits backslash escapes that must be re-parsed
     /// via `eval` (unlike GNU bash's `$'…'` form). Crow pastes this into the
     /// user's interactive shell (zsh/bash via `crow-shell-wrapper`).
-    public static func evalPromptLaunch(prefix: String, promptPath: String) -> String {
+    ///
+    /// `endOfOptions` emits a literal `--` before the prompt, so a prompt whose
+    /// first character is `-` reaches the binary as an operand instead of being
+    /// parsed as a flag. Quoting does **not** cover this: `printf %q` protects the
+    /// string from the *shell*, but a leading hyphen is not shell-special and
+    /// survives into argv, where an option parser claims it (CROW-968).
+    ///
+    /// It is **opt-in per agent, not a global default**, because `--` is a
+    /// convention rather than a guarantee — this helper is shared by Claude,
+    /// Codex, OpenCode and Antigravity, and a parser that treats a bare `--` as a
+    /// literal prompt word would silently corrupt every launch. Enable it only for
+    /// a CLI whose behaviour has been checked. Verified so far:
+    ///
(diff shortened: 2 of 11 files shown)


dgershman (approved):
## Code & Security Review

Two-layer fix for a launch-killing defect: strip the SKILL's YAML frontmatter from the **inlined** review prompt, plus an opt-in `--` end-of-options separator for Cursor. I verified the diagnosis, the placement of both layers, and ran the affected suites.

### Critical Issues

None.

### Correctness — verified independently

**The strip is in the right place.** `prepareReviewClone` renders `policySkillBody` once (`SessionService.swift:3558`) and forks it: the inlined prompt (`:3566` → `buildReviewPrompt` → `cursorReviewPrompt`) and the clone's `.claude/skills/crow-review-pr/SKILL.md` (`:3588`). Only the first is stripped. Hoisting it to `policySkillBody` would strip the Claude-facing copy and silently stop the skill loading — the new comment at `SessionService.swift:3582-3585` says exactly that, and `scripts/check-workspace-custom-instructions.sh:29` (`head -n 1` must equal `---`) is a PR-lane gate that keeps the checked-in file honest. Invariant is enforced, not just documented.

**`MarkdownFrontmatter.stripped` index handling is sound.** `lines.dropFirst().firstIndex(where:)` returns an index in the base array's space, so `lines[lines.index(after: closing)...]` addresses `lines` correctly; a closing delimiter on the final line yields `endIndex` and an empty slice rather than a trap. The no-op path returns `body` itself, so frontmatter-free bodies (`Scaffolder.bundledReviewSkill()`'s fallback, every fixture) are byte-identical — which is the property that matters most, since that path is the common one.

**`--` can't be swallowed by a preceding flag.** `CursorLaunchArgs.launchSuffix` only ever emits value-less booleans (`--trust`, `--force`, `--approve-mcps`), so the separator always lands as its own argv element. The `/bin/bash` round-trip in `ShellLaunchArgsTests.swift:90` proves that at the shell layer against a `---`-leading body.

**Blast radius is contained.** `endOfOptions` defaults to `false`, so the six non-Cursor `evalPromptLaunch` call sites (Claude, Codex ×3, OpenCode, Antigravity) are source- and output-identical — pinned by `evalPromptLaunchWithoutSeparatorIsUnchanged`. Grok is genuinely unaffected (`--prompt-file`, never argv), matching the docs claim.

**CI placement checks out.** `CrowCore` and `CrowCursor` are both in `LINUX_PACKAGES` (`.github/workflows/ci.yml:68`); `CrowEngine` is not, exactly as the PR description states. Putting the reusable logic in CrowCore is the right call — the regression guard runs on every PR.

### Test results

| Suite | Result |
|---|---|
| `CrowCore` — `MarkdownFrontmatter` + `ShellLaunchArgs` | 15/15 pass |
| `CrowCursor` (full) | 69/69 pass |
| `CrowEngine` — `SessionServiceReviewPrompt` | 17/17 pass |
| `CrowEngine` — `AgentLaunch`, `AgentHandoff`, `ReviewCloneStrip` | 43/43 pass |

Ran the launch/handoff/clone-strip suites specifically to check nothing else asserted the pre-`--` Cursor command string. Nothing did.

Confirmed the terse-form claim empirically: this Claude review session's own `.crow-review-prompt.md` begins `2f 63 72 6f` (`/cro…`), not `2d 2d 2d`, so Claude was never in the blast radius.

### Security Review

**Strengths:**
- The `--` separator is a small but real argument-injection hardening on the one prompt that is genuinely user-authored (`.crow-job-prompt.md` via `crow job add --prompt-file`): a job prompt beginning with `-` can no longer reach Cursor's parser as flags. Combined with `printf %q` + `eval`, the whole body stays a single argv word.
- The strip removes only `name`/`description` metadata — no security-relevant content, and nothing PR-controlled feeds the SKILL body (`ReviewVerdictPolicy.expand` reads workspace config only).
- Fail-safe direction on ambiguity: unclosed block → return input unchanged, rather than consuming to EOF and handing the agent an empty prompt.

**Concerns:** none found. No credential, network, filesystem-permission, or authz surface is touched.

### Code Quality — Green (consider, non-blocking)

1. **Exact-match delimiter, closing side.** `isDelimiter` (`MarkdownFrontmatter.swift:64`) requires the line to be exactly `---`. YAML permits trailing whitespace on the delimiter line. The *opening* line is CI-guarded by `check-workspace-custom-instructions.sh`, but the *closing* one is not — a stray trailing space there would silently no-op the strip and restore the frontmatter to Codex/OpenCode/Antigravity prompts, which have no `--` belt. A `.trimmingCharacters(in: .whitespaces)` in `isDelimiter` closes it for free without weakening the column-1 rule.

2. **Metadata-only body returns `""`.** `"---\nname: x\n---"` strips to empty. Not a shape a SKILL takes, and unreachable today — but CROW-439's preflight only checks the file *exists*, not that it's non-empty, so the failure mode would be a silent idle. Worth one test pinning whichever behaviour you intend.

3. **Resumed-`.job` case.** `endOfOptionsSeparatorOnlyRidesThePromptDispatch` covers `.work` and a resumed `.review`; the resumed `.job` shares the same `--continue` branch, so coverage is real, but adding the third loop element is free and makes the assertion self-evident.

4. **`--` verification is indirect.** `agent --list-models -- --bogus-flag-xyz` proves commander doesn't *reject* the separator; it doesn't prove commander *strips* it before the prompt reaches the model (`--list-models` exits first). Commander's documented behaviour says it does, and the downside if not is cosmetic (a brief opening with `-- `), so this isn't blocking — but it's one more reason to run the deferred step 5–6 Cursor review before leaning on the belt. Your "Not in this PR" section already flags this.

5. **Docs notation.** The matrix now writes `-p "$prompt"` where the emitted command is really `_CROW_P=$(< …); eval "… $(printf '%q' "$_CROW_P")"`. Strictly better than the stale `$(cat …)` it replaces (which `CursorAgentTests` asserts is absent), and the new `evalPromptLaunch` paragraph immediately below gives the real form, so this reads fine as placeholder notation.

The doc comments here are unusually good — the `MarkdownFrontmatter` header, the `endOfOptions` opt-in rationale, and the "don't hoist this strip" note at the fork are all the kind of thing that stops the next person re-breaking it.

### Summary Table

---

dgershman (approved):
## Code & Security Review

Reviewed at `5d8859f` against `origin/main`. Both halves of the fix are correct, the root-cause analysis in the description matches what the code does, and the test coverage is unusually good for a bug of this shape.

### Critical Issues

None.

### Security Review

**Strengths:**

- **No new injection surface.** `endOfOptions` interpolates a *static* literal `--`; nothing user-controlled reaches the shell construction that wasn't already there. The existing `shellQuote(promptPath)` + `printf '%q'` protections are untouched, and `ShellLaunchArgsTests.evalPromptLaunchWithoutSeparatorIsUnchanged` pins the default output byte-for-byte so the four opted-out harnesses are provably unaffected.
- **The `--` closes a small argv-confusion hole, it doesn't open one.** Verified the real `/bin/bash` round-trip test (`evalPromptLaunchSeparatorSurvivesAHyphenLeadingPrompt`) actually asserts `$2 == "--"` and `$3 == <hyphen-leading body>` — i.e. the separator lands as a distinct argv element rather than being glued onto the prompt. That's the right assertion; a `contains("--")` check would have passed vacuously against `--trust`/`--force`.
- **`MarkdownFrontmatter.stripped` only ever removes bytes**, never rewrites or re-encodes them. The slice boundary is a line boundary and `\n` is ASCII, so `String(decoding: bytes[start...], as: UTF8.self)` can't split a multi-byte sequence — the inline comment says exactly this and it holds.
- **The trust/strip posture is unchanged.** `stripCursorConfigFromReviewClone`, the `0o600` on the handoff prompt file, and the `--force --approve-mcps` review policy are all left alone; this PR touches only argv shaping and prompt-body content.
- **The Claude path is preserved where it matters.** Verified the fork: `prepareReviewClone` writes the clone's `.claude/skills/crow-review-pr/SKILL.md` from `policySkillBody` (`SessionService.swift:3584-3590`) — *unstripped* — while only `cursorReviewPrompt` strips. Hoisting the strip would have silently disabled Claude reviews; the comment at the call site says so explicitly. Good placement, correctly justified.

**Concerns:**

None found.

### Code Quality

**Verified against the real inputs, not just fixtures:**

- `skills/crow-review-pr/SKILL.md` and `Resources/crow-review-pr-SKILL.md.template` both have `^---$` at lines **1, 7, and 133**. The parser closes at line 7 and leaves 133 (the horizontal rule above the attribution footer) intact — matching the description's 162 → 154 claim, and covered by `leavesLaterHorizontalRulesAlone`.
- `scripts/check-workspace-custom-instructions.sh:81-86` does enforce frontmatter on both files, so the `MarkdownFrontmatter` doc comment's claim about the guard is accurate (not doc drift).
- `.crow-review-prompt.md` has exactly one writer (`SessionService.swift:3565-3566`), so there's no second path that could bypass the strip.

**Parser review** — `line(in:at:)` terminates in every branch (`next` strictly exceeds `offset` whenever a line is returned, and returns `nil` past the end), `isBlank` on an empty range correctly returns `true` via `allSatisfy`, and the unclosed-block case returns the input rather than consuming to EOF. The byte-level scan is the right call, and the comment explaining *why* (`"\r\n"` is one grapheme cluster; Darwin and swift-corelibs-foundation disagree on `components(separatedBy:)`) is the kind of note that saves the next person a Linux-only CI mystery. `crlfAndLFStripEquivalently` is the test that actually guards it.

**Test results:** `CrowCore` 582/582, `CrowCursor` 23/23, `SessionServiceReviewPromptTests` 17/17, `AgentLaunchTests` 12/12 — all green locally. The three pre-existing failures called out in the description are in `CrowDaemon`/`CrowTerminal` and unrelated to this diff.

**Findings (both Green — consider, non-blocking):**

1. **`ShellLaunchArgs.swift` / `CursorAgent.swift` — `endOfOptions` covers Cursor only.** The frontmatter strip removes the *known* trigger everywhere, but a user-authored **job** prompt beginning with `-` would still be claimed as a flag on OpenCode, Codex and Antigravity, whose parsers keep the `false` default. The PR argues this correctly — `--` is a convention, not a guarantee, and a harness that treats a bare `--` as a prompt word would be corrupted on *every* launch, so opting in without a verified binary would be worse than the gap. Agreed with the call; noting it only so the residual gap is on the record for whoever next has those CLIs installed. The docs already say "Check the CLI before enabling it anywhere else."

2. **`scripts/check-workspace-custom-instructions.sh:29` — the guard checks the opening delimiter but not the closing one.** `require_frontmatter` asserts `head -n 1 == "---"` plus the presence of `name:`/`description:`, so a SKILL.md that *lost* its line-7 close would still pass CI — and `stripped` would then latch onto the line-133 horizontal rule and silently discard the entire review brief, leaving the agent with just the footer. That's a hypothetical, not a live defect (today's files are well-formed, and the malformed-frontmatter case is separately caught by Claude Code failing to load the skill), and it's arguably out of scope for this PR. Worth a one-line `sed -n '2,20p' | grep -qx -- ---` in the guard, or a bounded search in `stripped`, if you want to close the "silently gutted prompt" mode rather than rely on the input staying well-formed.

**Docs:** `docs/agent-harness-matrix.md` is updated in the same PR (per the file's own house rule), and the incidental `$(cat …)` → `$prompt` refresh removes a genuine contradiction — `CursorAgentTests` asserts `$(cat` is *absent*, so the old notation was actively wrong.

### Summary Table

The fix is well-scoped, the two placement decisions the description flags for review are both correct and defended in code comments where the next reader will actually see them, and the tests pin the failure mode rather than the implementation. The two Green items are hardening suggestions, not defects in this diff.

Note the description's own carve-outs still stand: end-to-end confirmation needs a `crowd` rebuild/restart, and the two stuck sessions (`review-corveil-2170`, `review-corveil-cloud-terraform-378`) still hold the bad prompt file and need relaunching afterward.

---

PR #971, opened 2026-08-11
Reject quote-wrapped gateway headers, report gateway resolution (CROW-969)

### LINKED ISSUE #969: Gateway header values are stored verbatim: stray quotes silently break Claude sessions, and the failure is invisible while the role runs on Cursor
## Symptom

With `agentsByKind.review` set to `cursor`, reviews work. Clear that override so reviews run on `claude-code`, and every review session fails at launch with a bare **"API error"**. Work sessions on the same machine, against the same repo, are unaffected.

The error names nothing actionable — nothing points at the gateway, the workspace, or the header.

## Root cause

A gateway header value is stored and transmitted **verbatim**, including stray quote characters. One of my workspaces had its `x-citadel-api-key` value saved as

```
"Bearer sk-citadel-…"      ← literal " at both ends
```

instead of

```
Bearer sk-citadel-…
```

almost certainly a shell-quoting slip when the gateway was first set. Structural comparison of the two workspaces in my `config.json` (values redacted):

| workspace | header len | starts with `"` | ends with `"` |
|---|---|---|---|
| RadiusMethod | 66 | false | false |
| Corveil | 68 | **true** | **true** |

Exactly two extra characters — the two quotes.

Nothing anywhere in the chain notices:

1. `validateHeaderLine` (`Packages/CrowCLI/Sources/CrowCLILib/Validation.swift:305-318`) checks only for an embedded newline and a colon with a non-empty name. Surrounding quotes on the *value* pass.
2. `GatewayResolver.serializeHeaders` (`Packages/CrowCore/Sources/CrowCore/GatewayResolver.swift:30-35`) interpolates `"\(key): \(value)"` verbatim.
3. `ClaudeLaunchArgs.gatewayEnvPrefix` (`Packages/CrowClaude/Sources/CrowClaude/ClaudeLaunchArgs.swift:66-75`) shell-quotes the whole value, so the stray quotes survive intact into `ANTHROPIC_CUSTOM_HEADERS`.
4. Claude Code sends `x-citadel-api-key: "Bearer sk-…"`. The gateway rejects it. → **"API error"**.

## Why switching agents surfaced it

`Packages/CrowCursor/Sources/CrowCursor/` contains **zero** references to `GatewayResolver` or `ANTHROPIC` — the Cursor adapter never applies the workspace gateway at all. So a malformed gateway is entirely invisible while a role is pointed at Cursor, and becomes a hard failure the moment that role moves to Claude Code.

The config error can therefore be arbitrarily old. Changing the review agent didn't cause it; it revealed it.

## Secondary finding: work and review resolve to *different workspaces* for the same repo

This is why work sessions kept working and only reviews broke, and it is worth treating as its own defect.

`workspaceGatewayResolved(for:)` (`Packages/CrowEngine/Sources/CrowEngine/SessionService.swift:1929-1971`) resolves in two steps:

1. **By path** — `{devRoot}/{workspace}/…`. Used by work and job sessions.
2. **By repo slug** — the CROW-891 fallback, because a review clone lives at `{devRoot}/crow-reviews/…` and path math yields the literal `crow-reviews`.

My `corveil` repo is cloned at `{devRoot}/RadiusMethod/corveil`, so:

- a **work** session resolves by path → **RadiusMethod** workspace → clean header → works;
- a **review** of a `corveil/corveil` PR resolves by slug → `alwaysInclude: ["corveil/*"]` → **Corveil** workspace → malformed header → fails.

Every PR currently on my reviews board is `corveil/*`, so 100% of reviews hit the broken gateway while 0% of work sessions do. Same repo, same machine, two different gateways depending on session kind.

That divergence is defensible on its own (the slug fallback exists for a real reason), but nothing surfaces it, so "reviews fail and work sessions don't" reads as an agent bug rather than a workspace-binding one.

## Repro

1. Set a workspace gateway whose header value has surrounding quotes:
   `crow gateway set --workspace X --base-url https://gw.example.com --header 'X-Api-Key: "Bearer sk-test"'`
   Note it is accepted.
2. Point `review` at `cursor` → reviews launch fine (gateway ignored).
3. `crow agents set --clear review` (or pick Claude Code) → reviews fail with "API error".

## Proposed fixes

Roughly in order of value:

1. **Normalize or reject quoted header values.** In `validateHeaderLine`, reject a value that is wrapped in matching `"` or `'` — a header value legitimately containing quotes at both ends is vanishingly rare, and the failure mode it prevents is silent and total. Rejecting at the flag is better than stripping, since stripping would silently alter a value someone did mean.
2. **Make the failure diagnosable.** "API error" should say which gateway and workspace the session was launched with. `GatewayResolver` already logs a redacted warning when an `op://` reference fails to resolve (`GatewayResolver.swift:58`) — the same treatment for an auth rejection at launch would have made this a 30-second diagnosis.
3. **Surface gateway applicability per agent.** Since Cursor/Codex/OpenCode ignore `ANTHROPIC_*` entirely, the Settings agent picker (or `crow agents set`) should warn when moving a role onto an agent that ignores a configured gateway, and vice versa. Today the gateway silently becomes a no-op.
4. **Show which workspace a session resolved to.** Even a line in `crow get-session` would make the work-vs-review divergence above visible instead of inferred.
5. Consider a `crow gateway test --workspace X` that makes one authenticated request and reports the status — cheap, and it would catch this class of error at configuration time rather than at launch time.

## Notes

- No credential is included in this issue; the comparison above is structural only.
- Diagnosed against `crow` at the currently checked-out `main`.

dhilgaertner:
Closes #969

## The bug

A gateway header value saved as `"Bearer sk-…"` — a shell-quoting slip, exactly two characters longer than it should be — was stored and transmitted **verbatim**. Nothing in the chain noticed:

1. `validateHeaderLine` checked only for an embedded newline and a colon.
2. `GatewayResolver.serializeHeaders` interpolated `"\(key): \(value)"` as-is.
3. `ClaudeLaunchArgs.gatewayEnvPrefix` shell-quoted the *whole* value, so the stray quotes survived intact into `ANTHROPIC_CUSTOM_HEADERS`.

Claude Code sent a malformed credential, the gateway rejected it, and the agent surfaced a bare **"API error"** naming nothing — not the gateway, not the workspace, not the header.

It was also **invisible for as long as the role ran on Cursor**: `CrowCursor` (like `CrowOpenCode` and `CrowAntigravity`) never applies the workspace gateway at all. So a config broken months ago becomes a hard failure the moment `agentsByKind.review` is cleared to `claude-code` — which reads as an agent bug rather than a config one.

## The rule

Two predicates on `WorkspaceGateway` in CrowCore, shared by the CLI's `validateHeaderLine` and by `SecretRoutes.buildGateway` — the one server-side chokepoint that sees the final header map from **both** the `gateway-set` RPC and the web editor's `/config/*-gateway` POSTs, so the CLI and the browser can't drift on what they accept or what they say.

- **Value** — rejected when wrapped in a matching `"` or `'`. Two characters minimum, so a lone `"` isn't read as matching itself.
- **Name** — rejected for a `"` anywhere (RFC 9110's `tchar` excludes it outright, so such a name can never be valid), and for a **leading** `'` only, since `'` *is* a legal token character.

The name rule isn't redundant: quoting the whole pair (`--header '"X-Api-Key: sk-…"'`) splits on the first colon into a name and value where **neither half is individually wrapped**, so nothing else catches it.

Still accepted, deliberately: blank values (the load-bearing "keep the stored secret" signal), interior quotes, JSON-ish values, and `op://` references.

## Three places the check deliberately does *not* live

| Location | Why not |
|---|---|
| `SecretsRPC.decodeHeaderLines` | Runs strictly upstream of `buildGateway`; its message would shadow it, so the CLI and browser would report different text for the same mistake. |
| `WorkspaceGateway.init(from:)` | A throwing decode makes `ConfigStore.loadConfig` return nil, and the next write replaces every workspace, job and credential with defaults (the review-Red on #623). |
| `mergingPreservedHeaders` | It substitutes **stored** values for blank ones — so checking there would make the documented base-URL-only edit fail on any pre-existing bad secret, and permanently break the web editor for that workspace, locking the user out of fixing the very config the rule is about. |

Values **already** stored this way still launch and are *warned* about rather than repaired: stripping would mean guessing the quotes aren't part of the secret, and a loud rejection at the gateway beats authenticating with a credential nobody verified. The warning is checked **before** the `op://` branch on purpose — quotes around a reference defeat `hasPrefix("op://")`, so it silently falls into the plaintext arm and is never resolved at all.

## Making the failure diagnosable

`workspaceGatewayResolved` matched a workspace two ways and then **discarded it**, so nothing reported where a session's gateway came from. The match is now a pure `SessionService.gatewayMatch(worktreePath:prLinks:devRoot:config:)` shared by the launch path and `crow get-session`.

Every Claude launch and handoff now logs one line:

```
[SessionService] Gateway for session <uuid>: workspace 'Corveil' (matched by repo_slug)
    -> https://gw.example.com, headers: x-citadel-api-key
```

Header **names** only — never values. `get-session` gains `workspace_id`, `workspace_name`, `workspace_match`, `gateway_set` and `gateway_base_url`, following `WorkspaceRPC.workspaceJSON`'s redaction exactly because `get-session` is **not** in `localOnlyDenial` and a remote `/rpc` peer reads it; header names and values stay with the local-only `crow gateway get`. Manager sessions report their own gateway under `workspace_match: "manager"` rather than an empty match — otherwise the new diagnostic would report "no gateway" for a Manager that has one, recreating this ticket's blind spot inside the tool built to close it.

This makes visible the divergence behind the original report: work and job sessions resolve **by worktree path** while review clones resolve **by repo slug** (the CROW-891 fallback), so a repo cloned under one workspace but claimed by another workspace's `alwaysInclude` gets a *different gateway* depending on session kind — which is why reviews failed while work sessions on the same repo, on the same machine, kept working. Carrying the matched workspace's gateway even when nil also separates "nothing claims this session" from "a workspace claims it and has no gateway": identical behaviour, different fixes, previously indistinguishable.

## Testing

`make test` passes. Added coverage across 8 files: both predicates against a full edge-case table; the warn-don't-strip contract pinned in `GatewayResolverTests` (including that a quoted `op://` ref never invokes `resolveSecret`); CLI acceptance/rejection including an assertion that **the error names the header and not the secret**; `buildGateway` rejection with blank values still passing and both-or-neither still winning; an end-to-end `gateway-set` rejection asserting the stored config is unchanged; and `get-session` asserting no header material in the payload.

Verified by hand against the built CLI — all three malformed shapes are rejected with `sk-SECRET` absent from every message.

`SessionServiceGatewayResolutionTests` previously *documented* that the two-lookup rule couldn't be tested (`ConfigStore` is all-static; ADR 0012 forbids touching the live devroot pointer), so only the individual helpers were covered and their composition never was. The pure matcher is that missing seam — the suite header is rewritten and 8 composition cases added, including path-wins-over-slug and the `crow-reviews`-shadowing guard.

**Two pre-existing failures on this branch are unrelated and untouched**: `bootCatchResetsTheHistory` and `refreshTerminalsRebindsTheActiveTerminalToTheFreshRow`, both drift guards asserting on `Resources/web/app.js` source shape. Confirmed by stashing this branch's changes and reproducing both against the clean tree.

## Notes for review

- The two **pre-existing** `validateHeaderLine` messages still interpolate the whole raw line. Left alone deliberately: the no-colon case has no name to print instead, and the newline case needs the raw line to be diagnosable at all. The two new messages quote the name only.
- Out of scope, per the ticket's own ordering: a `honorsGateway` capability on `CodingAgent` (proposal 3) and `crow gateway test` (proposal 5).
- `docs/cli-reference.md`'s `get-session` entry claimed it returns "worktrees, terminals, and links", which it never has. Corrected while rewriting that sentence rather than propagating it.


dgershman (approved):
## Code & Security Review

Verified locally: `swift build` clean; `CrowCore`, `CrowCLI`, `CrowEngine` test suites all green. `CrowDaemon` fails exactly the two tests the PR body names — `bootCatchResetsTheHistory` and `refreshTerminalsRebindsTheActiveTerminalToTheFreshRow` — both `Resources/web/*` source-shape drift guards. This diff touches no file under `Resources/`, so the attribution to pre-existing drift holds.

### Critical Issues

None.

### Security Review

**Strengths:**

- **Redaction is correct and precedent-consistent.** The new `get-session` fields (`EngineRouter.swift:337-341`) expose only `gateway_set` + `gateway_base_url`, matching `WorkspaceRPCSupport.swift:501` byte for byte. Confirmed `get-session` is genuinely absent from `RPCWebSocketHandler.localOnlyDenial` (only `gateway-get`/`gateway-set` are gated there), so the caution is warranted and correctly applied — a remote `/rpc` peer already reads the same two facts via `workspace-list`, so the marginal disclosure is nil.
- **No secret reaches an error string.** Both new CLI messages (`Validation.swift:338-347`) and both new daemon messages (`SecretRoutes.swift:187-195`) interpolate the header *name* only. `headerLineErrorNamesTheHeaderNotTheSecret` and `rejectionMessageNamesTheHeaderNotTheValue` pin this with explicit `!contains("sk-SECRET")` assertions — the right shape for a regression floor. The launch log (`SessionService.logGatewayLaunchDecision`) likewise emits `gateway.customHeaders.keys` only.
- **The `op://` ordering is the subtle part and it's right.** Checking `isQuoteWrapped` *before* the `hasPrefix("op://")` branch (`GatewayResolver.swift:53-71`) is what catches the worst case: a quoted reference fails the prefix test, silently falls into the plaintext arm, and ships the literal `"op://…"` string as the credential. `gatewayResolverDoesNotResolveQuoteWrappedOpReference` pins it by making `resolveSecret` record an issue if called.
- **Warn-don't-strip is the correct call.** Mutating a stored credential on the guess that its quotes are extraneous would be worse than a loud gateway rejection. `gatewayResolverPassesQuoteWrappedValueThroughUnchanged` pins the contract so a future "helpful" strip fails at the test.
- **The three deliberate non-locations are each justified.** Confirmed `WorkspaceGateway.init(from:)` still doesn't throw on this (`AppConfig.swift:319-341`) — a throwing decode there really would make `loadConfig` return nil and let the next write default-replace the config. And `mergingPreservedHeaders` staying exempt is what keeps `--header "X-Api-Key:"` (base-URL-only edit) and the web editor working against an already-bad stored value, instead of locking the user out of the config the rule exists to fix.
- **No new write path.** Verified `SettingsSecrets.preservingSecrets` still overwrites `managerGateway` and per-id workspace gateways from stored config, so `set-config` can't author header values around `buildGateway`.

**Concerns:**

None blocking.

### Code Quality

The `workspaceGatewayResolved` → pure `gatewayMatch(worktreePath:prLinks:devRoot:config:)` extraction is behavior-preserving — I walked both branches:

- Path lookup finding a workspace *name* that no config entry matches still falls through to the slug fallback (old: `matched` stays nil; new: the `if let workspace = …first(where:)` fails and control reaches the slug arm).
- Path lookup landing on a workspace with a nil gateway still short-circuits without trying the slug fallback, in both versions.

Carrying `gateway` even when nil is the actual diagnostic win — separating "nothing claims this session" from "a workspace claims it and has no gateway", which behave identically (`ANTHROPIC_*` unset) but need different fixes. `matchedWorkspaceWithNoGatewayStillMatches` covers exactly that.

`gatewayMatch` and its two helpers are `nonisolated static` on a `@MainActor` class, so the pure rule is callable off the actor and directly unit-testable — that's the seam the old suite header said didn't exist, and the 8 new composition cases (`pathMatchWinsOverSlug`, `reviewCloneIsNotCapturedByAWorkspaceNamedCrowReviews`) cover the composition that was previously untested.

The `⚠️ workspaceGatewayMatch, NOT workspaceGatewayResolved` note at `EngineRouter.swift:293-298` is load-bearing and worth keeping: `resolve` shells out to `op read` at 15s per header, and this closure runs on the MainActor, so the wrong call there would wedge the daemon on a biometric prompt. The match path is config-read only, and `get-session` is CLI-only (no `get-session` hits in `Resources/web/`), so the two added file reads per call don't land in a hot path.

**Green — consider, none blocking:**

1. **No escape hatch for a genuinely quote-delimited credential.** `isQuoteWrapped` is unconditional on both write paths, so a secret that really does start and end with a matching `"` or `'` can only be stored by hand-editing `config.json`. Vanishingly unlikely for an API key, and the failure is a clear message rather than silent corruption — but there is no `--force`.

2. **`SecretRoutes.swift:165-166` slightly overstates the outcome.** "Enforcing them at this one chokepoint is what keeps the CLI's and the browser's error text identical for the same mistake" — but `GatewaySet.validate()` runs `validateHeaderLine` client-side, so a CLI user is rejected before the RPC leaves and never sees `buildGateway`'s wording. The shared *predicates* genuinely can't drift (which is the property that matters, and `Validation.swift:314-316` states it correctly); the *text* deliberately differs, and the CLI's is the more actionable of the two. Comment-only.

3. **The "one server-side chokepoint" framing isn't strictly exhaustive.** `SettingsSecrets.preservingSecrets` (`SettingsSecrets.swift:70`) restores stored gateways by workspace `id`; a workspace present in `incoming` but *not* in `current` keeps the gateway it arrived with. So a `set-config` blob carrying a brand-new workspace could seat a gateway without passing `buildGateway`. Pre-existing, out of scope here, and the launch-time `GatewayResolver` warning still covers the resulting config — but the CHANGELOG's absolute phrasing has this one edge.

4. **Cosmetic:** a stored-but-empty `WorkspaceGateway` (`baseURL: ""`, no headers — which `init(from:)` permits) reports `gateway_set: false` alongside `gateway_base_url: ""` rather than `null`. Identical to `workspaceJSON`'s existing behavior, so consistency argues for leaving it.

The docs are unusually good here — `configuration.md`'s "reviews fail and work sessions don't can be a workspace-binding symptom rather than an agent one" is the sentence that would have short-circuited the original investigation, and it's now written down next to the log line that proves it.

### Summary Table

---

Issue #680, opened 2026-07-13
Web UI: hide the terminal tab bar for manager sessions (managers have no tabs)


dgershman:
Managers shouldn't have terminal tabs — a manager session is just "the manager" (a single terminal). Hide the tab feature for manager sessions. This also fixes an observed naming bug (below).

## Observed bug (motivating this)
When you rename a **manager** session, the name up top (header/session name) updates, but the terminal **tab** still shows the old name (e.g. "Manager"). Cause: tabs are labeled from the **terminal** name — `t.name` in `renderTabs()` (`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js:1683`) — and renaming a *session* doesn't change the terminal's name, so the tab goes stale. Hiding tabs for managers makes this moot.

## Decision / fix
Managers have no tabs. In `renderTabs()` (`web/app.js:1678`), when the currently selected session is a manager, clear `#tabbar` and return early — no tabs, no "+" new-terminal button, no close "×".

- Manager check: look up the current session by `selectedId` and test `kind === 'manager'` (the same predicate already used throughout app.js — e.g. :616, :869, :1201, :1411, :1456).
- Collapse the tab bar's layout so there's no empty strip above the terminal for managers (hide `#tabbar` / don't reserve its height), so the manager terminal fills the area cleanly.
- Leave `#tabbar` clearing on session switch (~:713/:1651/:1742) intact.

## Non-goals / notes
- Work (non-manager) sessions keep tabs exactly as-is — tabs, "+", and close still work.
- If the desktop (SwiftUI) tab UI also shows tabs for managers, mirror this there as a **separate** parallel follow-up; this ticket is the web UI.

## Acceptance criteria
- Selecting a **manager** session shows **no** terminal tab bar (no tabs, no "+", no "×"); the manager terminal fills the area with no empty bar left behind.
- Renaming a manager session no longer leaves a stale "Manager" tab (there is no tab to go stale).
- Work sessions are unchanged: tabs render, "+" adds a terminal, "×" closes one.

## Target branch
Web UI — base on `feature/crow-593-web-ui-parity`, not `main` (`crow:auto` defaults to `main`, where the web UI doesn't exist).

[🐦‍⬛ Created with Crow via Cursor](https://github.com/radiusmethod/crow)

PR #981, opened 2026-08-11
Add a ↻ Reload button to the session top bar (CROW-979)

### LINKED ISSUE #979: Add a terminal refresh (↻) control to the session top bar — reload on click, recreate behind a secondary gesture
## Ask

Put a **refresh control for the terminal in the session's top bar**, so reloading the xterm surface is a visible one-click action instead of something you have to know to right-click for. Use the same ↻ glyph + in-flight spinner idiom the sidebar already uses.

## Today

Two different recoveries exist, and neither is discoverable from the top bar:

| Action | What it does | How you reach it today |
|---|---|---|
| `reloadTerminal()` (`app.js:5241`) | Non-destructive: `term.reset()`, drop the socket, reconnect — the fresh attach re-fits, re-selects the window, and `crowd` replays the pane's tmux scrollback. Same recovery as a page reload, scoped to the surface. | **Terminal right-click menu only** (`app.js:4925`) |
| `recreateTerminal()` (`app.js:3117`) | Destructive: rebuilds the tmux window and restarts the agent (resumed where the harness supports it). Already `confirmModal`-gated with `danger: true`. | **Only when Crow flags `scrollback_degraded`** — the `⚠` badge on a tab (`app.js:3060`) or the Manager degraded strip (`app.js:3043`) |

So the cheap fix is hidden behind a right-click, and the heavy fix can't be asked for on demand at all — it only appears once Crow has already decided the window is degraded.

## Proposal

Add a ↻ button to the session top bar (`#detail-header`, rendered by `renderHeader()` at `app.js:2547`), reusing the existing `.action-button` + `.action-spinner` pattern already used there for the In Review action (CROW-749) and on the sidebar tickets card (`app.js:1363-1367`).

- **Click ⇒ `reloadTerminal()`** — the safe one, no confirmation. It acts on the currently attached window (`activeTerminal`), which is what the user is looking at.
- **Right-click / overflow on the same button ⇒ "Recreate terminal…"** — routes to the existing `recreateTerminal(activeTerminal)`, keeping its danger-confirm modal untouched. This is the part that makes recreate reachable without waiting for a `⚠`.
- Show the spinner while the socket is reconnecting, then settle — matching the tickets-refresh treatment.
- **Leave the right-click terminal menu item in place.** This adds a discoverable path; it doesn't move the existing one.

### Why the header and not the tab bar

The tab bar is the more obvious home for a per-terminal action, but `renderTabs()` early-returns for manager sessions (#680 — managers have no tabs) and `#tabbar:empty { display: none }` (`app.css:374`) collapses the strip entirely. A control placed there would be missing on exactly the surface that needs it most: the Manager window is the common CROW-804 stuck-alt-screen case, which is why the degraded strip had to be special-cased for it in the first place. The header renders for every session kind, and it is literally "the top bar area of each session" the ask names.

If the implementer prefers the tab bar, it must additionally handle the manager case rather than inheriting the collapse.

## Ambiguity worth resolving in the PR

"Refresh" maps to either recovery above. This ticket defaults to **reload on click, recreate behind a secondary gesture**, on the grounds that reload is cheap, reversible, and fixes the common corrupted-grid case, while recreate restarts the agent and should stay deliberate. If the intent was the opposite, say so on the PR rather than silently swapping them.

Also note: the ask says "like we can on the sidebar," but no terminal-reload control exists on the sidebar today — the session context menu (`sessionMenuItems`, `app.js:2170`) has no terminal actions, and the sidebar ↻ belongs to the tickets card. Read the reference as *the sidebar's refresh idiom*, not an existing sidebar terminal action.

## Test plan

- The ↻ appears in the top bar for work, review, job, and **manager** sessions.
- Click reloads the attached terminal: buffer resets, socket reconnects, tmux scrollback replays, and scroll-up history is intact afterward.
- Spinner shows while reconnecting and clears when the socket is open; a failed reconnect doesn't leave it spinning forever.
- The secondary gesture opens the existing recreate confirmation; cancelling changes nothing; confirming rebuilds the window and re-attaches (the `attachedWindow = null` reattach path in `recreateTerminal` still applies).
- Switching tabs/sessions then clicking ↻ reloads the *currently attached* window, not a stale one.
- The right-click "Reload terminal" menu item still works.
- With no terminal attached (session with no live pane), the button is disabled or absent rather than throwing.

dgershman:
Closes #979.

## The problem

`reloadTerminal()` (`app.js`) — reset the xterm buffer, drop the socket, reconnect so `crowd` replays the pane's tmux scrollback — was reachable from exactly one place: the terminal's **right-click menu**. A touch device has no way to open that menu, so on a phone the cheap recovery for a corrupted surface didn't exist at all. The only way out was leaving the session and coming back.

Meanwhile the session top bar already carries *In Review*, *Mark as Completed* and *Delete*, and the `max-width: 720px` block leaves `#detail-header` and `.actions-cluster` visible — there was simply no reload button among them.

## The change

A `↻ Reload` button in that cluster, leading position so it's never adjacent to `Delete`.

| | before | after |
|---|---|---|
| work session | `Mark as Completed` · `Delete` | **`↻ Reload`** · `Mark as Completed` · `Delete` |
| manager session | *(no action cluster at all)* | **`↻ Reload`** |

It sits **outside** `renderHeader`'s `kind !== 'manager'` guard on purpose. That guard is why a Manager renders no action cluster, and the Manager window is the common CROW-804 stuck-surface case — it also has no tabs (#680, `#tabbar:empty { display: none }`) to hang a control off, which is why the tab bar was the wrong home for this.

**In-flight state is module-scoped, not on the node.** `refreshLive` repaints the whole header every 4s and would wipe a DOM-held flag mid-reload — the same reason `ticketRefreshPending` lives outside `.tickets-refresh`. While reloading, the `↻` swaps for the shared `.action-spinner`, both boxed to 12×12 so the chrome stays put (the CROW-797 tickets-refresh fix).

**Settling is on the socket opening** — the honest "attached again" signal. But `onclose` retries with backoff *forever*, so a daemon that never comes back would spin indefinitely; a 10s settle timer is the floor under that, plus a clear from `showEmptyDetail` when the session is dropped out from under it.

`refreshTerminals` is what binds `activeTerminal` (the button is disabled until there is one), and `selectSession` paints the header *before* awaiting it — so it nudges the button. That nudge is guarded to repaint only when the enablement actually changed: an unconditional `renderHeader` there would also throw away an in-flight `markInReviewAction`, which does keep its spinner on the node.

## Ambiguity resolved (per the ticket)

The ticket asked for this to be called out rather than silently swapped. **This ships reload only.** The proposed secondary gesture routing to `recreateTerminal()` is not here — the ask is reload, and recreate restarts the agent, so it stays deliberate and stays exactly where it is (the `⚠` tab badge and the Manager degraded strip, both untouched). The existing right-click *Reload terminal* item is untouched too: this adds a discoverable path, it doesn't move the old one.

## Corrections to the ticket's premises

- The class is `.action-btn`, not `.action-button`.
- There's no refresh glyph in the `ICONS` map — this uses the literal `↻`, as `.tickets-refresh` does.
- `reloadTerminal()` takes **no argument** and is synchronous; it acts on the shared surface and `connectTerminalWs`'s `onopen` re-selects `activeTerminal.window`. So there's no promise to await, and the handler reads `activeTerminal` at click time rather than capturing it at paint time.
- It must **not** route through `attachWindow(activeTerminal.window)` — the `win === attachedWindow` guard would make an explicit reload a no-op.

## Tests

New `web-tests/terminal-reload.test.js` — 51 assertions, registered in **both** `test` and `test:ci`, documented in `web-tests/README.md`. Covers the ticket's test plan: renders for work/review/job **and manager**; disabled with nothing attached; click resets the buffer, detaches the old socket's handlers *before* closing it, opens a fresh one; the spinner swap and its 12×12 no-layout-shift box; settling on `onopen`; settling via the timer when `onopen` never comes; timer disarmed when open wins the race; double-tap coalescing; no-op (not a throw) with no terminal; `showEmptyDetail` dropping a pending reload; and the right-click menu item still present.

Mutation-checked rather than assumed — dropping the `onopen` clear (5 fail), the disabled gate (5 fail), the busy read, the repaint guard (2 fail), or the `.reload-glyph` CSS rule (1 fail) each turns it red.

`npm run test:ci` is green. `row.test.js` is 57 passed / 10 failed both with and without this change — pre-existing CROW-802 drift, which is why `test:ci` omits it.

## Verify

```bash
cd Packages/CrowDaemon/web-tests && npm ci && npm test
```

Then in the browser with `crowd` running:
- `↻ Reload` shows in the top bar on a **work** session and on the **Manager**.
- Click: buffer resets, socket reconnects, tmux scrollback replays, scroll-up history intact.
- Spinner turns while reconnecting and settles on open, with no width jump.
- Stop `crowd` and click: settles within ~10s instead of spinning forever.
- Switch tabs, then click ↻ — the *currently attached* window reloads.
- Under 720px (or on a phone): present and tappable — the case this is for.
- Right-click *Reload terminal* still works.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -1443,6 +1443,19 @@ body.update-banner-visible #back-to-sidebar {
   animation: wizard-spin 0.7s linear infinite;
 }
 
+/* The ↻ on the header's Reload button (CROW-979). Boxed to the same 12×12 as
+   `.action-spinner` so swapping one for the other while reloading doesn't shift
+   the button's width. */
+.reload-glyph {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  width: 12px;
+  height: 12px;
+  font-size: 12px;
+  line-height: 1;
+}
+
 /* ---- Scorecard (ADR 0008 web parity, #721) ---- */
 .score-weeklabel { color: var(--text-secondary); font-size: 12px; }
 .score-banner {
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -2427,6 +2427,11 @@ function showEmptyDetail(msg, opts) {
   selectedBoard = null;
   terminals = [];
   activeTerminal = null;
+  // CROW-979: the ↻ that was spinning belonged to the session we just dropped, and
+  // its `onopen` may never come (the socket is being torn down with it). Clearing
+  // after `selectedId` is already null makes the repaint a no-op — the header is
+  // emptied below regardless — but leaves no stale flag for the next selection.
+  clearTerminalReloadPending();
   const app = document.getElementById('app');
   app.classList.remove('has-selection', 'board-active', 'mobile-show-sidebar');
   // On narrow screens `#app:not(.has-selection) #detail` is display:none, which
@@ -3159,8 +3164,30 @@ function renderHeader(s) {
                                          liveFor(s.id).auto_rebase_state));
   }
 
-  // Right-aligned action cluster: PR quick-actions, then status transitions + delete.
+  // Right-aligned action cluster: terminal reload, PR quick-actions, then status
+  // transitions + delete.
   const actions = el('div', 'actions-cluster');
+  // Terminal reload (CROW-979). `reloadTerminal()` was reachable only from the
+  // terminal's right-click menu, which a touch device has no way to open — so on a
+  // phone the cheap recovery for a corrupted surface didn't exist and the only way
+  // out was leaving the session and coming back. Deliberately OUTSIDE the
+  // `kind !== 'manager'` guard below: that guard is why a Manager session shows no
+  // buttons at all, and the Manager window is the common CROW-804 stuck-surface
+  // case (it has no tabs to hang a control off either, #680).
+  const reloadBusy = terminalReloadPending;
+  const reload = el('button', 'action-btn action-btn-reload', '');
+  // Swap the ↻ glyph for the shared spinner ring rather than spinning the button,
+  // so the chrome stays put while it turns (the CROW-797 tickets-refresh fix).
+  reload.appendChild(reloadBusy ? el('span', 'action-spinner') : el('span', 'reload-glyph', '↻'));
+  reload.appendChild(el('span', null, 'Reload'));
+  reload.disabled = reloadBusy || !activeTerminal;
+  reload.title = reloadBusy
+    ? 'Reloading terminal…'
+    : (activeTerminal
+        ? 'Reload the terminal — reset the view and reconnect'
+        : 'No terminal attached');
+  reload.onclick = () => reloadTerminalAction();
+  actions.appendChild(reload);
   if (pr && pr.has_pr && !pr.is_merged) {
     // Quick-actions dispatch a prompt into the session's managed Claude Code
     // terminal — disable them when there is none (native `canDispatchQuickAction`;
@@ -3535,6 +3562,11 @@ async function refreshTerminals() {
     }, { replace: true });
   }
   renderTabs();
+  // CROW-979: this call is what binds `activeTerminal`, and the header's ↻ Reload
+  // is disabled until there is one. selectSession renders the header *before*
+  // awaiting this, so without a nudge the button would sit disabled until the next
+  // 4s refreshLive tick.
+  syncTerminalReloadEnabled();
   // #673: session switches funnel through here too (selectSession → refreshTerminals),
   // changing which window this shared socket shows — same corruption as a terminal-tab
   // switch. attachWindow reloads only when the window actually changes, so a same-session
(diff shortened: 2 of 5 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- UI-only change that routes through the existing `reloadTerminal()` path — no new server endpoints, RPC methods, or auth surface.
- Button is disabled when `!activeTerminal`; the action handler double-checks `!term || !activeTerminal` before tearing down sockets.
- Label text is static (`↻`, "Reload"); no user-controlled strings are injected into the DOM or WebSocket URLs.
- Old socket handlers are detached before `close()`, preventing stale `onclose` handlers from interfering with the new connection — same safe pattern as the pre-existing right-click reload.

**Concerns:**
- None identified.

### Code Quality
- Module-scoped `terminalReloadPending` correctly survives `refreshLive`'s 4s header repaints (mirrors `ticketRefreshPending`).
- `syncTerminalReloadEnabled()` only repaints when enablement actually changed — avoids wiping an in-flight `markInReviewAction` spinner parked on the node (`app.js:4497–4500`).
- Settle semantics are sound: `onopen` clears pending (honest "attached again" signal) with a 10s floor timer for the daemon-never-returns case, and `showEmptyDetail` clears stranded state on session deselect (`app.js:2407`, `5089`).
- Reload deliberately sits outside the `kind !== 'manager'` guard so Manager sessions (CROW-804 stuck-surface case, no tabs) get a discoverable recovery path.
- CSS boxes `.reload-glyph` at 12×12 to match `.action-spinner` — no layout shift during the swap (`app.css:1352–1360`).
- Comprehensive regression suite: `terminal-reload.test.js` — 51 assertions, mutation-checked against dropping the `onopen` clear, disabled gate, busy flag, repaint guard, and CSS rule.

### Summary Table

| # | Finding | Location |
|---|---------|----------|
| 🟢 | Follows established async-UI patterns (`ticketRefreshPending`, CROW-797 spinner swap) with clear inline rationale | `app.js:4476–4531` |
| 🟢 | Guarded header repaint avoids disrupting other in-flight action spinners | `app.js:4492–4500` |
| 🟢 | Manager + mobile discoverability without changing recreate/degraded-scrollback paths | `app.js:2679–2699` |
| 🟢 | 51-assertion test suite with controllable timers and handler-retaining fake sockets | `terminal-reload.test.js` |

`npm run test:ci` green (including `terminal-reload.test.js`: 51 passed).

---

Issue #953, opened 2026-08-08
Review sessions auto-completed and cleaned up while reviewer still has open work — no visibility into what's left


dgershman:
## Problem

After upgrading to the CROW-945 fix (#949), **all open review sessions disappeared from the left sidebar** even though the human reviewer was not finished reviewing those PRs personally.

The user can no longer tell:
- which PRs they already looked at vs still need to revisit
- what is still waiting on them on GitHub vs what Crow already acted on (automated review / changes-requested)
- what review work is "left out there"

This is worse than the pre-945 stuck-session bug: before, stale `review-*` sessions piled up; now they vanish entirely.

## What we observed (dgershman, 2026-08-08)

- `store.json` has **0** sessions with `kind: review` (not even completed ones remain).
- `crow list-reviews` returns **[]** — GitHub queue also empty in Crow after label/repo filters (`dependencies`, `zarf-dev/zarf`, etc.).
- Cleanup is enabled with **4h retention** — completed sessions are deleted quickly.
- Logs show review sessions being auto-cleaned, e.g. `review-crow-920`, `review-corveil-1899`.
- CROW-945 enabled `decideReviewCompletions` rule 1 ("viewer submitted review") by fixing `viewerLastReviewedAt` parsing.

## Likely root cause (hypothesis)

1. **CROW-945 rule 1 is too aggressive for historical sessions:** Any review session where the viewer had *ever* submitted a review after `session.createdAt` gets auto-completed — including sessions the user still considers in-progress for their own manual review, and including sessions where Crow/agent posted an automated review on the user's behalf.
2. **4h cleanup removes the audit trail:** Once completed, review sessions are deleted from the sidebar/store, so there is no "reviewed / pending / revisit" history.
3. **Reviews board filters hide the GitHub queue:** Pending `review-requested:@me` PRs that remain on GitHub may all be filtered out (dependabot `dependencies`, excluded repos), so neither the sidebar nor the board shows what's left.

## Impact

- Reviewer loses situational awareness: can't distinguish "I reviewed this" vs "Crow auto-reviewed" vs "still need my eyes."
- No way to resume a review session for a PR that was auto-completed.
- Empty sidebar + empty Reviews board feels like data loss / regression.

### A. Narrow rule 1 completion criteria
- Only auto-complete when **this session's round** is done, e.g.:
  - viewer submitted review **and** `viewerLastReviewedAt` is after session's `lastReviewedHeadSha` / current round anchor
[...]

PR #984, opened 2026-08-12
Reviews board: split into In review / Not approved yet / Approved recently (24h)

### LINKED ISSUE #982: Reviews board: split into In review / Not approved yet / Approved recently (24h) so approvals stop vanishing (#953 follow-through)
## Ask

Split the Reviews board into **three groups** instead of one flat list:

1. **In review** — what the board shows today: a review is open/in progress on it.
2. **Not approved yet** — requested of me, no approval from me yet (never reviewed, or I requested changes and the ball is back in my court).
3. **Approved recently (last 24 h)** — I approved it in the last day, so it's still visible instead of vanishing.

## Why

This is the follow-through on **#953** ("Review sessions auto-completed and cleaned up while reviewer still has open work — **no visibility into what's left**"). #953 closed with the rule-1 completion fix, but its directions **B** (preserve review history instead of deleting) and **C** (surface "what's left" explicitly) were left as "for discussion" and never got their own ticket. This is that ticket, scoped to the grouping.

The underlying complaint, from the 2026-08-07 call: work keeps slipping out of view — *"it gets lost and there's like — I look in GitHub and I'm like holy s***"* — and reviews are the sharpest case because **an approved PR disappears from the board the instant you approve it**. Group 3 exists specifically to give approvals a 24-hour tail so "did I already do that one?" is answerable without going to GitHub.

## The blocker to know about up front

**Group 3 cannot be built from the current query.** The board is fed by `review-requested:@me state:open type:pr` (`GitHubCodeBackend`), and GitHub *clears the pending review request* the moment you submit a review — the code says so in several places ("invisible to the reviewer's queue and to `review-requested:@me` alike"). So an approved PR is not merely un-flagged, it is **absent from the only source the board reads**.

Group 3 therefore needs one of:

- **a second query** — e.g. `reviewed-by:@me state:open type:pr`, filtered client-side to viewer approvals within 24 h (costs one more search per poll); or
- **a local ledger** — persist `{pr_url, reviewed_at, state}` when Crow observes the viewer's review, and render group 3 from that (this is #953 direction B, and it survives the PR being closed/merged).

Recommend the second query for v1: it needs no new persistence and stays correct across machines. Note it in the PR if you pick differently.

## The other gap: the payload doesn't carry the verdict

`ReviewRequest` already has `viewerLastReviewedAt` (`ReviewRequest.swift`), **but `ReviewsPayload.build` never serializes it** — the emitted object stops at `head_ref_oid` / `kickoff_action`. And nothing in the model records *which* verdict it was; the field's own comment says it's the timestamp of the most recent `APPROVED` / `CHANGES_REQUESTED` / `DISMISSED`, collapsed together.

So groups 2 and 3 need:

- `viewer_last_reviewed_at` added to the `ReviewsPayload` object (model side already exists), and
- the viewer's last review **state** captured in `ReviewRequest` and emitted — without it, "not approved yet" cannot be distinguished from "I requested changes" or "I approved."

## Grouping rules (proposed)

| Group | Rule |
|---|---|
| **In review** | `review_session_id` is set (or `kickoff_action` indicates a round in flight) |
| **Not approved yet** | requested of the viewer, and viewer's last review state is not `APPROVED` — includes never-reviewed and `CHANGES_REQUESTED` |
| **Approved recently** | viewer's last review state is `APPROVED` and `viewer_last_reviewed_at` is within 24 h |

A PR belongs to exactly one group; when it qualifies for two, the earlier group in that order wins (a PR with an open review session stays under **In review** even if you approved it a minute ago). Groups render collapsed-with-count when empty rather than disappearing, so an empty board still says *what* is empty — that was half the shock in #953.

`done_last_24h` on the tickets payload is the existing precedent for the 24-hour window; keep the wording and the boundary consistent with it.

## Also worth folding in (from #953 direction C, cheap here)

While the groups are being built, surface the **filtered-out count** — `ignoreReviewLabels` + `excludeReviewRepos` silently hid real `review-requested:@me` PRs during the #953 incident, and the board looked empty when GitHub's queue was not. A "12 hidden by filters" line under the groups closes that hole. Skip if it balloons the diff; say so on the PR.

## CLI parity

`crow list-reviews` should return the same three buckets (grouped, or each review carrying its group) so the CLI and the web board agree — the board payload is the one source, per `ReviewsPayload`.

## Test plan

- A PR requested of me with no review from me lands in **Not approved yet**.
- I request changes on it → it stays in **Not approved yet** (does not vanish, does not move to approved).
- A review session is started on it → it moves to **In review** and stays there.
- I approve it → it leaves the `review-requested:@me` queue but appears under **Approved recently**, and drops off after 24 h.
- A PR approved 25 h ago is in no group.
- Empty groups render with a zero count instead of disappearing.
- With `ignoreReviewLabels` / `excludeReviewRepos` hiding PRs, the hidden count is shown and matches what GitHub actually returns.
- `crow list-reviews` groups match the web board exactly.

dgershman:
Closes #982

The Reviews board was one flat list fed by `review-requested:@me`, so **an approved PR vanished the instant you approved it** — GitHub clears the pending review request on submit, and that search was the board's only source. This is the #953 follow-through: directions **B** and **C**, scoped to the grouping.

## Three groups, one payload

| Group | Rule |
|---|---|
| **In review** | an active review session covers the PR |
| **Not approved yet** | still requested of you — never reviewed, or you requested changes |
| **Approved recently · 24h** | your last verdict is `APPROVED`, within 24 h |

Precedence runs in that order, so a PR you approved a minute ago that has a live review session stays under **In review**. `ReviewGroup.classify` is the single rule; `ReviewsPayload` stamps `group` on every review and publishes `group_order` + `group_counts`, so **`crow list-reviews` and the web board are grouped identically by construction** rather than by two clients agreeing on a rule.

## The blocker: group 3 needed a second query

As the ticket called out, group 3 cannot be built from the current query. I took the **second search** option:

```
reviewed-by:@me state:open type:pr sort:updated-desc
```

It rides in the **same GraphQL document** as the existing `review-requested:@me` search — so it is one more search node on a request already being made, **not an extra HTTP round trip** (cheaper than the ticket's "one more search per poll" estimate; ~1 extra GraphQL point per ~60s poll). Chosen over the local ledger because it needs no new persistence and stays correct across machines.

`sort:updated-desc` is load-bearing at `first: 50`: an approval bumps the PR's `updatedAt`, so the rows this group wants sort to the front instead of relying on GitHub's relevance ordering.

`reviewed-by:` matches *any* submitted review, so the result is narrowed to approvals on the parsed verdict — otherwise a PR you requested changes on would be duplicated into the approved group.

## The other gap: the verdict now reaches the wire

`viewerLastReviewedAt` collapsed `APPROVED`/`CHANGES_REQUESTED`/`DISMISSED` into one timestamp and **was never serialized at all**. New `ReviewVerdict` captures which verdict it was; both `viewer_last_reviewed_at` and `viewer_last_review_state` are now emitted. Without the state, "not approved yet" cannot be distinguished from "I requested changes" or "I approved" — which is the whole split.

## Folded in: `hidden_by_filters` (#953 direction C)

`ignoreReviewLabels` / `excludeReviewRepos` silently hid real `review-requested:@me` PRs during the #953 incident and the board looked empty while GitHub's queue was not. The board now states `N hidden by filters` under the groups. Small diff, so it's in.

## Design notes

- **The approved tail is excluded from `detectReviewSounds`.** It enters the list precisely when you *finish* a review; chiming `reviewRequested` for your own approval would announce work arriving when work just left.
- **The 24 h cutoff is applied at serialization time, not poll time**, so the tail expires on the clock rather than up to a poll late. Boundary and wording match the existing `done_last_24h` precedent.
- **Empty groups render with a zero count** instead of disappearing, and say *what* is empty — a board showing nothing at all was half the shock in #953.
- **A PR in both searches is owned by the requested queue** (it is asking for something), so nothing renders in two groups.
- **GitLab** exposes no per-reviewer verdict, so its "Approved recently" group stays empty rather than guessing — same reasoning as the existing `viewerLastReviewedAt` nil there.
- **Backwards compatible:** `reviews` stays a flat array; an older daemon sending no `group` renders every review under "Not approved yet" with a row-derived count.
- On the approved tail, cards show *when you approved* rather than the PR's `updatedAt` — that's the question the group exists to answer.

## Test plan (from the ticket)

All covered by automated tests, not just by hand:

| Ticket case | Test |
|---|---|
| requested, no review from me → Not approved yet | `neverReviewedIsNotApprovedYet` |
| I request changes → stays put, doesn't vanish | `changesRequestedStaysNotApprovedYet` |
| review session started → In review, stays there | `activeSessionWinsOverFreshApproval` |
| I approve → leaves the queue, appears under Approved recently | `recentApprovalSurvivesInItsOwnGroup` |
| approved 25 h ago → no group | `approvalOlderThan24hIsInNoGroup` |
| window tracks the clock | `windowBoundaryIsEvaluatedAtSerializationTime` |
| empty groups render with a zero count | `emptyGroupsStillCarryAZeroCount` + web `three headers still present` |
| hidden count shown and matches | `hiddenByFiltersIsReported` + web `the hidden count is stated` |
| `crow list-reviews` matches the web board | one builder + `group_order`/`group_counts`; parity gate green |

Plus provider-level coverage that the two searches split correctly, that only the *viewer's* verdict counts (a later `COMMENTED` or another reviewer's approval must not win), that a missing `approvedPRs` alias or unknown viewer login degrades to an empty tail rather than a parse failure, and that the query declares both variables.

**Results:** CrowCore 592 ✅ · CrowEngine 717 ✅ · CrowProvider 168 ✅ · CrowCLI 354 ✅ · web-tests 100 ✅ (18 new).

CrowDaemon has 2 failures in `WebNotificationCenterTests` / `WebTerminalAssetTests` — **pre-existing on main**, verified by diffing the exact asserted strings between `HEAD` and this branch (identical). Both scan `app.js` regions this PR never touches (`refreshTerminals`, the boot recovery block).


dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Second GitHub search uses server-controlled query strings passed via GraphQL variables (`-F reviewQuery` / `-F approvedQuery`), not user input — no command or injection surface.
- Viewer verdict parsing filters to round-closing states and the authenticated viewer login; `COMMENTED`/`PENDING` cannot close a round.
- SAML partial-recovery paths degrade to empty approved tail rather than throwing, consistent with existing monitored-PR handling.
- Web board group titles come from a fixed map; unknown group ids fall back to the raw id rather than being dropped silently.

**Concerns:**
- No new security issues identified. Approved-tail data is the same PR metadata already shown on the board.

### Code Quality
- **Yellow — `ReviewGroup.classify` mis-groups re-requested PRs within the 24h window** (`Packages/CrowCore/Sources/CrowCore/Models/ReviewGroup.swift:70-76`, `Packages/CrowEngine/Sources/CrowEngine/ReviewsPayload.swift:66-71`): If you approved a PR and the author re-requests you within 24 hours, the PR returns to `review-requested:@me` with `viewerLastReviewState` still `.approved`. Precedence classifies it as `approved_recently` instead of `not_approved_yet`. The >24h re-request case is handled (falls through to `not_approved_yet`), but the within-24h case can hide actionable work under the wrong section. Fix: only apply `approvedRecently` to the approved-tail list (e.g. pass `fromApprovedTail` into classify), or check “still in requested queue” before the approval branch.
- **Green** — Single `ReviewGroup.classify` + `ReviewsPayload.build` keeps CLI and web in sync; `group_order` / `group_counts` / per-review `group` is a clean contract.
- **Green** — `viewer_last_review_state` serialization closes the gap that collapsed approvals vs changes-requested; provider tests cover verdict parsing, COMMENTED masking, and missing `approvedPRs` alias degradation.
- **Green** — `hidden_by_filters`, empty-group rendering, sound suppression for `approved_recently`, and backwards-compatible defaults for older daemons are thoughtful #953 follow-through.
- **Green** — `swift build` succeeds; CrowEngine `ReviewsPayload` suite (19 tests) passes.

### Summary Table

---

dgershman:
Thanks — the Yellow was a real bug and it's fixed in f84457b2.

**You were right, and it was worse than one mis-grouped PR.** The >24h re-request case already fell through to `not_approved_yet`, so the same PR read `approved_recently` at hour 1 and `not_approved_yet` at hour 25 — one PR, two groups, decided by the clock alone and nothing else. My own doc comment on `classify` argued the re-requested PR belongs in `not_approved_yet` ("the ball is genuinely back in your court"); I then applied the ticket's literal table, which contradicted it inside the window. Queue membership is the signal, not the last verdict.

**Fix — your first suggestion.** `ReviewGroup.classify` now takes `isRequestedOfViewer` and checks it *before* the approval branch:

```swift
if hasActiveReviewSession { return .inReview }
if isRequestedOfViewer { return .notApprovedYet }   // ← new
if viewerLastReviewState == .approved, …within 24h { return .approvedRecently }
```

`ReviewsPayload` passes it per source list — `true` for the `review-requested:@me` queue, `false` for the `reviewed-by:@me` tail — rather than re-deriving "still in the queue" from the data, since the two lists *are* exactly that distinction. (`IssueTracker` already dedupes a PR in both searches into the requested queue, so the re-requested PR arrives tagged correctly.)

The approved tail keeps its group: the gate is on queue membership, not on ignoring approvals.

**Regression coverage** (3 new tests in `ReviewsPayloadTests`):

- `reRequestedAfterApprovalIsNotApprovedYet` — the exact case you described: requested queue, verdict `APPROVED` 1 h ago → `not_approved_yet`, and `approved_recently` count is 0.
- `reRequestedAfterApprovalIsStableAcrossTheWindowBoundary` — same PR at hour 1 and hour 25 now returns the same group, pinning the self-contradiction shut.
- `approvedTailStillGroupsAsApprovedRecently` — the tail didn't regress into the fallback.

Also updated the `ReviewGroup` doc table and the `crow list-reviews` notes in `CLAUDE.md` to state the rule: *anything still in `review-requested:@me` is Not approved yet, whatever you decided last round.*

**Suites after the fix:** CrowCore 592 ✅ · CrowEngine 720 ✅ · CrowProvider 168 ✅ · CrowCLI 354 ✅ · web-tests 100 ✅. No web change was needed — grouping is server-side and the board renders whatever `group` it's handed.

The 2 CrowDaemon failures remain pre-existing on `main` (`WebNotificationCenterTests`, `WebTerminalAssetTests`), unrelated to this PR — verified earlier by diffing the asserted strings against `HEAD`.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- GraphQL search queries are fixed strings passed via `gh api` flags — no user-controlled query injection surface.
- New board UI uses `textContent` via `el()` for group titles/counts/notes, so server-provided `group` ids cannot become XSS even on the unknown-id fallback path (`app.js:1258-1262`, `app.js:4720-4725`).
- `viewer_last_review_state` / `viewer_last_reviewed_at` are emitted only for the authenticated viewer's own dashboard payload; no new cross-user data exposure.
- Degraded paths are safe: missing `approvedPRs` alias, empty viewer login, and SAML-partial responses all degrade to an empty approved tail rather than misclassifying or crashing (`GitHubCodeBackend.swift:872-882`, `BackendsTests.swift:2005-2050`).
- Re-requested-after-approval is gated on queue membership (`isRequestedOfViewer`) rather than verdict alone, preventing actionable work from hiding under "Approved recently" (`ReviewGroup.swift:79-80`, `ReviewsPayloadTests.swift:347-392`).

**Concerns:**
- None blocking. The approved-tail search still relies on `reviews(last: 20)` client-side verdict scanning (`GitHubCodeBackend.swift:819-837`) — same pre-existing pattern as the requested queue; a PR with >20 round-closing viewer reviews could theoretically mis-pick the latest verdict. Extremely unlikely in practice.

### Code Quality
- **Architecture:** `ReviewGroup.classify` is the single grouping rule; `ReviewsPayload.build` is the sole wire encoder — CLI and web cannot drift. Good separation.
- **Correctness:** 24h window evaluated at serialization time (`now` injection) so the tail expires on the clock, not the poll interval. Deduping approved tail against the requested queue prevents double-rendering (`IssueTracker.swift:967-972`).
- **UX polish:** `detectReviewSounds` excludes `approved_recently` so self-approvals don't false-trigger notifications (`app.js:977-982`). Empty groups render with zero counts and explanatory copy. `hidden_by_filters` addresses the #953 silent-empty-board failure mode.
- **Backwards compatibility:** Ungrouped reviews default to `not_approved_yet` in the web client; `MonitoredPRListing.recentlyApprovedPRs` defaults to `[]`.
- **Tests:** Excellent coverage — grouping precedence, window boundary, re-request stability, filter parity, provider parsing, and 18 new web tests all pass locally (CrowCore 592, ReviewsPayload 22/22, new BackendsTests 5/5, web-tests 100/100). `swift build` clean.

### Summary Table

**Green (non-blocking):**
- `first: 50` cap on the `reviewed-by:@me` search means power reviewers with >50 recent open approvals may not see the oldest tail entries — acceptable given poll cadence and 24h window.
- GitLab intentionally leaves "Approved recently" empty (no per-reviewer verdict API) — documented and consistent with prior `viewerLastReviewedAt` behavior.

---

PR #985, opened 2026-08-12
Default the session switcher to esc+tab, with esc as a prefix key (CROW-980)

### LINKED ISSUE #980: Change session switcher default binding from shift+tab to esc+tab
## Summary

Change the session switcher default key chord from **`shift+tab`** to **`esc+tab`**.

Follow-up to CROW-976 / PR #978.

## Why

**Shift+Tab already means something** in Crow's web terminal — Claude Code uses it for permission-mode cycling when the switcher is not capturing the binding inside the terminal. Even with `captureInTerminal` off by default, Shift+Tab is overloaded and confusing as the switcher chord.

**Esc+Tab** avoids that collision and keeps Esc available for cancel while the overlay is open.

## Scope

Update the default everywhere it is defined or documented:

- `AppConfig.SwitcherSettings.binding` default (`AppConfig.swift`)
- Web UI client defaults (`app.js` `uiConfig.switcherBinding` + fallbacks)
- CLI help text (`SettingsCommands.swift`, `docs/cli.md`)
- Tests that assert the default (`AppConfigTests`, `switcher.test.js`, etc.)

No change to the binding parser or overlay behavior beyond the new default string — existing installs keep their stored config unless they never set `switcher.binding`.

## Acceptance

- Fresh install / missing `switcher` config decodes to `esc+tab`
- `crow ui get` and Settings UI show `esc+tab` as the default
- Docs and CLI `--help` say `esc+tab`
- Tests updated accordingly

dgershman:
## Summary

Moves the session switcher's default binding from **`shift+tab`** to **`esc+tab`** — Shift+Tab is already Claude Code's permission-mode cycle, and `switcher.captureInTerminal` defaults to **true**, so the old default ate that keystroke inside every focused terminal.

**This could not be a string swap, and the ticket's "no change to the binding parser" is where it breaks.** Esc is not a modifier — no later keyboard event reports it the way `shiftKey` reports Shift. Under the existing parser, `esc+tab` splits to `['esc','tab']`, `'esc'` matches none of shift/ctrl/alt/meta, and the chord collapses to bare **`tab`**: every unmodified Tab anywhere in the UI would open the overlay, commit on release, and (with `captureInTerminal: true`) be swallowed before reaching the PTY — Tab-completion dead in every terminal. So `esc` is now parsed as a **prefix key**, which is also the only reading under which the ticket's "keeps Esc available for cancel while the overlay is open" makes sense.

### How the prefix works

- **Tap Esc, then Tab.** The prefix arms for 1.5s and is broken by any intervening keystroke — a bare modifier excepted, since a resting Shift shouldn't cancel the sequence.
- **Esc is never swallowed.** It has to stay instant for interrupting an agent, and holding it back on the chance a Tab follows would add latency to every press — so it reaches the terminal on its way through and the overlay opens on the Tab that follows.
- **A prefix chord holds nothing down**, so it can't commit on release the way `shift+tab` does — releasing Tab would close the overlay on the very keystroke that opened it. It stays open instead: Tab / ←→ cycle, **Enter** (or a click) switches, **Esc** cancels. The hint line under the strip states whichever gesture the configured chord actually uses (`release Shift` / `Enter` / `release Tab`).
- Cycling once open deliberately skips the armed check — re-tapping Esc for every step would be unusable.
- Modifier chords (`shift+tab`, `ctrl+space`, …) are untouched: same parse, same hold-and-release commit.

### Migration

Existing installs are unaffected — the new default reaches only a config with **no** `switcher.binding` key, pinned by a new decode test. `crow ui set --switcher-binding shift+tab` restores the old chord.

## Changes

| File | What |
| --- | --- |
| `AppConfig.swift` | `SwitcherSettings.binding` default → `esc+tab` (init + decode), with the why in the doc comment |
| `app.js` | `uiConfig` default + `loadUIConfig` fallback; `parseSwitcherBinding` gains `prefix`; arm/disarm state + `switcherBindingOpens`; Enter-to-commit; prefix-aware keyup and hint; terminal key handler gated on the armed prefix |
| `SettingsCommands.swift` | `--switcher-binding` help + a discussion paragraph on the sequence form |
| `docs/cli.md` | regenerated via `make docs` |
| `CHANGELOG.md` | `### Changed` entry under `[Unreleased]` |
| tests | `AppConfigTests` (defaults + stored-binding-wins), `switcher.test.js` (+21 checks), `key-handler.test.js` (+2) |

Note `parseSwitcherBinding` now reads modifiers from the *lead* segments rather than the whole split, so a trailing key part can no longer double as its own modifier.

## Test plan

- [ ] `Esc` then `Tab` from anywhere in the web UI opens the overlay; `Tab` / `←→` cycle; `Enter` switches; `Esc` returns to the origin session.
- [ ] A bare `Tab` (no preceding Esc) does **not** open the overlay and still reaches the focused terminal — completion works.
- [ ] `Esc` alone still interrupts Claude Code in a focused terminal (it is passed through, not swallowed).
- [ ] `Esc`, then any other key, then `Tab` → no overlay (the sequence is broken); waiting >1.5s after `Esc` → no overlay.
- [ ] `crow ui set --switcher-binding shift+tab` restores hold-Shift/release-Shift behavior with no reload; the hint line reads `release Shift to switch`.
- [ ] A config with no `switcher` block reports `esc+tab` from `crow ui get`.

Closes #980


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Client-side keyboard handling only; no new attack surface for auth, injection, or credential exposure.
- Esc is intentionally never swallowed, preserving the agent-interrupt path — a deliberate availability/safety choice, not a bypass.
- Prefix arming uses a bounded 1.5s window (`SWITCHER_PREFIX_WINDOW_MS`) so a stray Esc cannot leave the UI in a permanently armed state.
- Intervening keystrokes disarm the prefix (with a sensible exception for resting modifier keys), reducing accidental activation.

**Concerns:**
- None identified. The Esc→terminal passthrough is by design and documented.

### Code Quality
- **Green** — The prefix-key model is the right fix for CROW-980: without treating `esc` separately from `shift/ctrl/alt/meta`, `esc+tab` would collapse to bare `tab` and break Tab-completion in terminals. The `parseSwitcherBinding` change (modifiers read from lead segments only) is a correct parser hardening.
- **Green** — `switcherBindingOpens()` cleanly gates both the document-level handler and `handleTerminalKey`, so bare Tab still reaches the PTY while Esc+Tab is captured when `captureInTerminal` is true.
- **Green** — Commit semantics are well split: modifier chords commit on release; prefix chords stay open and commit on Enter; `switcherCommitHint()` reflects the active gesture.
- **Green** — Migration is safe: new default applies only when `switcher.binding` is absent; stored bindings (e.g. `shift+tab`) are preserved, with decode test coverage.
- **Green** — Defaults are synchronized across `AppConfig.swift`, `app.js` `uiConfig`, CLI help, and regenerated `docs/cli.md`.
- **Green** — Test coverage is strong: +21 switcher checks, +2 key-handler checks, and a new `switcherSettingsKeepsStoredBindingOverDefault` Swift test. All targeted tests pass locally; `swift build` succeeds.
- **Green** — Consider adding an automated test for prefix expiry (>1.5s after Esc → Tab does not open). The logic is straightforward (`switcherPrefixArmed()`), but a `Date.now` mock would lock in the documented behavior from the manual test plan.

### Summary Table

**Recommendation:** Approve — driven by [0 Red, 0 Yellow, 7 Green] findings.

---

Issue #767, opened 2026-07-20
Restore web parity: Scorecard 'Manager usage' (ungraded) section


dhilgaertner:
## Summary

Restore the **Manager-usage section** on the Scorecard, lost in the native→web move
(ADR-0010, CROW-593). The native `ScorecardView` rendered an **ungraded "Manager usage"**
section alongside the graded session cards; the web scorecard (`renderScorecard`) doesn't.

## Background

- Web scorecard: `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` —
  `renderScorecard(root)` (~L2057), "ADR 0008 web parity, #721". It renders grade, sessions
  shipped, combined-score (v2), baseline, this-week stats, previous weeks, and this-week's
  sessions — computed on the daemon by `ScorecardModel` and shipped as a flat `ScorecardDTO`
  over `get-scorecard`.
- Native `ScorecardView` (deleted at `eb7a489^`, PR #594) additionally rendered a **Manager
  usage** section: manager sessions' usage shown **ungraded** (managers aren't scored the way
  work sessions are), so you can still see manager cost/activity without it distorting the grade.

## Scope

- [ ] Confirm whether `ScorecardModel` / `ScorecardDTO` already carries manager-usage data; if
      not, add it (ungraded manager rollup: cost, active time, tokens, etc.) on the daemon side.
- [ ] Render a **Manager usage** section in `renderScorecard`, visually distinct and clearly
      ungraded, matching the native layout.
- [ ] Empty/loading states consistent with the rest of the scorecard.

## Acceptance criteria

- [ ] The web scorecard shows a Manager-usage section (ungraded) reproducing the native
      `ScorecardView` behavior.
- [ ] Manager usage does not affect the graded score/combined-score math.

## References

- Web: `renderScorecard` in `.../Resources/web/app.js` (~L2057); `get-scorecard` RPC + `ScorecardModel` / `ScorecardDTO`.
- Native (history): `ScorecardView` at `eb7a489^`.
- ADR-0010 (`docs/adr/0010-retire-the-macos-app.md`) — app retirement / web-parity context. Related: ADR-0008 (scorecard), #721.

PR #987, opened 2026-08-12
Meter Manager sessions like work sessions (CROW-983)

### LINKED ISSUE #983: Scorecard: meter Manager sessions like work sessions (widen the 3-field rollup, add the outcome-independent efficiency grade)
## Ask

Meter Manager sessions the same way work sessions are metered — the scorecard figures, per-week and per-session, not just the three numbers they get today.

## Today

Manager sessions are **already partially metered** (#745, #767), but as a deliberately thin, ungraded bucket:

- `ManagerWeeklyUsage` aggregates the Manager's telemetry per ISO-8601 week — same bucketing `ScorecardModel.build` uses — and persists it (`JSONStore.managerUsageWeekly`) so weeks survive telemetry retention. Built by `SessionService.refreshManagerUsage`.
- But `ManagerUsageWeekDTO` (`ScorecardDTO.swift:263`) projects only **three figures**: `promptCount`, `totalTokens`, `totalCost`. Everything else in the `SessionAnalytics` it already holds is dropped at the DTO boundary.

Work sessions, by contrast, get the full `SessionGradeRow` / `WeeklyScorecard` treatment: efficiency grade + its input raws, cost, active time, commits, churn hint, wall-clock, cost-per-shipped, the v2 combined score, and trailing-4-week baseline medians.

**Why the gap exists:** the pipeline is snapshot-driven, and a `SessionAnalyticsSnapshot` is produced when a session reaches a terminal status. The Manager is permanent and never terminates, so it never emits one — #745's weekly rollup was the workaround, and it was scoped to visibility only.

## What to build

1. **Widen the DTO to the data already collected.** `ManagerWeeklyUsage.analytics` is a full `SessionAnalytics`; the week DTO should carry the rest of it — active time, cache hit ratio, compactions, input-tokens-per-prompt, API error rate — instead of three fields. This is mostly a projection change, not new collection.
2. **Render the Manager row with the same chips and a weekly trend**, so a Manager week reads like a work week rather than a footnote.
3. **Give Managers the efficiency grade.** This is the part worth doing carefully: the efficiency half of ADR 0008's grade (compactions per active hour, input tokens per prompt, cache hit ratio, API error rate) is **outcome-independent** — none of it needs a shipped session. It can be computed for a Manager from data already in hand, and it is the half that actually answers "is this session being run efficiently."

## The decision to confirm — outcomes stay out

Do **not** fold Managers into the graded aggregate, the throughput count, cost-per-shipped, the combined score, or the baseline.

ADR 0008's outcome flag is `status == .completed`, and a Manager never completes: it *spawns* the sessions that ship. Folding its cost into cost-per-shipped would charge work it enabled but did not perform, and grading a Manager standalone on throughput would read `insufficientOutcomes` forever. That's why #745 made it a separate ungraded bucket, and this ticket keeps that posture — it widens **metering**, not grading, with the one exception of the outcome-independent efficiency grade above.

If the intent was actually "count Manager cost as overhead inside cost-per-shipped," that is an **ADR 0008 amendment**, not a code tweak — say so on the PR and it becomes a separate ticket with the ADR change. Do not quietly change the denominator.

## Second gap: multiple Managers collapse into one bucket

`managerUsageWeekly` is `[String: ManagerWeeklyUsage]` keyed by **week alone** (`SessionService.refreshManagerUsage` writes `weekKeyFormatter.string(from: interval.start)`). `crow create-manager` can make N Managers, and today they all aggregate into a single row — there is no way to see which Manager burned what.

Per-Manager breakdown needs a composite key (week + session id) and a DTO that carries the session identity. Include it if the diff stays reasonable; if not, file the follow-up and say so rather than leaving the single-bucket behavior undocumented.

## CLI parity

`crow get-scorecard` returns the payload the web tab renders, so any field added to the DTO shows up there for free — confirm the Manager rows carry the widened figures in the CLI output too.

## ADR note

This extends the ADR 0008 addendum that introduced the Manager bucket (#745). Update that addendum in the same PR to describe the widened metering and the explicit "efficiency graded, outcomes not" split, so the posture stays recorded rather than inferred from code.

## Test plan

- A Manager week renders every widened figure (active time, cache hit ratio, compactions, input tokens/prompt, API error rate) alongside prompts/tokens/cost.
- The Manager's efficiency grade computes from a week of telemetry and matches a hand-computed `EfficiencyGrading` result over the same raws.
- The graded aggregate is unchanged by Manager data: `sessionsShipped`, `costPerShipped`, the combined score, and the baseline medians are byte-identical with and without Manager weeks present.
- Weeks that aged out of telemetry retention still render from the persisted rollup.
- With two Managers running, each is attributed separately (or, if deferred, the single-bucket behavior is documented and the follow-up filed).
- `crow get-scorecard` shows the same Manager figures as the web tab.
- Existing `ScorecardDTOTests` / `EfficiencyScorecardTests` stay green; a store written before this change still decodes.

dgershman:
Closes #983

Manager sessions were metered since #745, but as a thin ungraded bucket: `ManagerWeeklyUsage` persisted a full 13-field `SessionAnalytics` and `ManagerUsageWeekDTO` projected three fields of it. This widens the metering and grades the outcome-independent half.

## Efficiency graded, outcomes not

ADR 0008's outcome flag is `status == .completed`, which a Manager can never reach — so the shipped count, cost-per-shipped, the combined score, and the baseline stay closed to it. The other four metrics need no outcome at all, and they are the half that answers "is this being run efficiently".

**No new grading code.** `EfficiencyGrading.efficiencyInput(for:)` leaves `costContext` nil, and `costPerShippedDeduction` — the only outcome-touching deduction — returns nil without one. There is still exactly one `grade(_:)`.

`ScorecardModel` stays pure over snapshots and Manager rows still join only at the DTO boundary, so the graded aggregate is byte-identical with and without Manager weeks (pinned by the existing #767 invariant test).

> **On the denominator:** folding Manager cost into cost-per-shipped as overhead was considered and **rejected** here. A Manager *spawns* the sessions that ship, so charging its cost against their outcomes bills work it enabled but did not perform. If that was the real intent, it is an **ADR 0008 amendment** and a separate ticket — say so and I'll file it rather than quietly move a graded denominator.

## Two ticket premises the code contradicted

Both were verified in the source, and both changed the work:

**1. Compactions were inert in production.** `noteCompactionEvent` and its persist path shipped with #704, but nothing ever called it — `PostCompact` fell through to the state signal source's `default`. ADR 0008's heaviest penalty (up to −30) had been dead since v1, **for work sessions too**. Now wired in the `hook-event` handler, completing follow-up 3.

⚠️ **Consequence worth a reviewer's attention:** real work-session grades begin taking compaction deductions, so some will drop a letter. That is the metric working for the first time, and it lands inside the ADR's tunable-priors/calibration framing — but it is a live behavior change, not just new metering.

**2. Secondary Managers were invisible, not merged into one bucket.** The rollup provider hardcoded `AppState.managerSessionID`, so `crow create-manager` sessions contributed nothing despite emitting telemetry under their own ids. Rollups are now keyed by week *and* session id and computed for every `.manager` session — a data-completeness fix, not a display nicety.

## Notes for review

- **Store-wiping hazard avoided.** `StoreData` decoding is all-or-nothing (`JSONStore.init` copies to `store.json.bak` and resets on any throw), so a non-optional field on `ManagerWeeklyUsage` would have wiped the user's **sessions**, not just the scorecard. Both new fields are `Optional`, with a test pinning a pre-CROW-983 `store.json`.
- **A race caught in self-review.** `refreshManagerUsage` suspends at every provider `await`, and `noteManagerCompaction` runs on the same actor — reading the compaction carry-forward while building the update set silently dropped any compaction that landed mid-refresh. The read moved inside the `store.mutate`. The regression test fails without the fix (verified by reverting it).
- **Duplicate rows.** A legacy bare-keyed week and a recomputed composite-keyed week both persist after upgrade; the recomputed one wins, while a legacy week nothing superseded (aged out of retention) still renders.
- **Unbounded growth.** `managerWeeks` was uncapped while `priorWeeks` was bounded to 4 — ~52 rows after a year, each now carrying a full chip set. Bounded per Manager so a second one can't evict the first's history.
- **Latent timezone bug fixed.** The week key came from a `static DateFormatter` capturing `.current` at first access, while the refresh re-read `.current` per call.

`crow get-scorecard` is a verbatim DTO passthrough, so the widened figures reach the CLI with no RPC or CLI change.

## Test plan

Mapped to the ticket's plan:

| Ticket item | Covered by |
|---|---|
| Manager week renders every widened figure | `scorecard.test.js` chip assertions; `managerWeekProjectsTheWidenedFigures` |
| Efficiency grade matches hand-computed `EfficiencyGrading` | `managerEfficiencyGradeMatchesHandComputedGrading` |
| Graded aggregate unchanged | `managerUsageNeverTouchesTheGradedMath` (existing, still green) + `managerGradeIsOutcomeIndependent` |
| Aged-out weeks still render | `agedOutWeeksKeepTheirPersistedRollup` |
| Two Managers attributed separately | `eachManagerGetsItsOwnRollup`, `eachManagerIsAttributedSeparately` |
| Pre-change store still decodes | `preCROW983ManagerRollupStillDecodes` |
| `crow get-scorecard` parity | DTO passthrough; **not yet run against a live daemon** — see below |

New `web-tests/scorecard.test.js` (32 assertions), registered in both `package.json` script lists and the README. #767 shipped the Manager card with no render coverage at all.

**Results:** `make test` passes across all packages; CLI/RPC parity clean. `npm run test:ci` green.

Two known-red suites are **pre-existing on `main`**, confirmed by stashing and re-running on the clean tree:
- 2 CrowDaemon `app.js` asset tests (`WebTerminalAssetTests`, `WebNotificationCenterTests`)
- `row.test.js`'s 10 CROW-802 failures, which is why CI runs `test:ci`

**Not verified live:** the `crow get-scorecard` output and the rendered card need `telemetry.enabled` plus a `crowd` restart, which would disrupt the running instance. The DTO is a verbatim passthrough and the render path is covered headlessly, but a live check is worth doing before merge.

## ADR

Extends ADR 0008 with a **Manager metering addendum** documenting the "efficiency graded, outcomes not" split, per-Manager attribution, and the compaction wiring fix; marks follow-up 3 implemented and narrows the #745 addendum's blanket "never graded".


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- No new attack surface: scorecard data is local/private (ADR 0008), and the web UI renders Manager figures via `textContent` (`app.js:1253-1257`), so session names and deduction labels cannot inject HTML.
- Store migration is handled safely: new `ManagerWeeklyUsage` fields are `Optional` to avoid all-or-nothing `StoreData` decode failures that would wipe sessions (`ManagerWeeklyUsage.swift:20-35`, pinned by `preCROW983ManagerRollupStillDecodes`).
- The refresh/compaction race is fixed correctly by reading `compactionCount` inside `store.mutate` rather than capturing it before the async provider loop (`SessionService.swift:4271-4277`).

**Concerns:**
- None identified. Hook-event handling compares against a shared constant (`SessionHookState.compactionEventName`) rather than a scattered string literal, reducing mismatch risk (`AppState.swift:786`, `EngineRouter.swift:1396`).

### Code Quality
- The "efficiency graded, outcomes not" split is clean: `EfficiencyGrading.efficiencyInput(for:)` leaves `costContext` nil, so `costPerShippedDeduction` never applies to Managers while reusing the single `grade(_:)` path (`EfficiencyGrading.swift:240-251`).
- Per-Manager attribution is end-to-end: provider signature takes `UUID`, rollups use composite keys, DTO projection dedupes legacy vs recomputed weeks and caps per Manager (`SessionService.swift:4248-4261`, `ScorecardDTO.swift:101-131`).
- Compaction wiring for work sessions is the intended ADR follow-up 3 fix; it will change some existing grades now that `PostCompact` is counted — well documented in the PR and ADR addendum.
- Test coverage is thorough: 18 CrowCore DTO tests (including graded-aggregate isolation), 6 CrowEngine rollup/race tests, store back-compat, and 32 headless web assertions in `scorecard.test.js`. All passed locally; `swift build` clean.

### Summary Table

---

dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Backward-compatible store migration is handled correctly: new `ManagerWeeklyUsage` fields are optional, with `JSONStoreTests.preCROW983ManagerRollupStillDecodes` pinning that a decode failure would wipe sessions, not just scorecard data.
- The `refreshManagerUsage` / `noteManagerCompaction` race is fixed by reading compaction carry-forward inside `store.mutate` (not while building the update map), with a regression test that interleaves compaction during provider `await`s.
- `PostCompact` compaction wiring in the `hook-event` handler uses the existing session-resolution path; no new injection surface or credential exposure.
- `get-scorecard` remains read-only; Manager names are display metadata only. Web rendering uses `textContent` via `el()`, so DTO strings are not interpreted as HTML.

**Concerns:**
- No security blockers identified.

### Code Quality
- **Yellow — stale empty-state copy (`app.js:3988`):** `scorecardEmpty()` still says *"The Manager session is never graded — its usage appears in its own ungraded section once captured."* CROW-983 now grades the outcome-independent efficiency half and renders an efficiency badge. Users who hit the empty state (no snapshots and no Manager weeks yet) get messaging that contradicts the shipped behavior. The Manager card itself is correct (`efficiency only` pill, updated explainer); only this empty-state paragraph needs updating.
- **Green — architecture is sound:** `EfficiencyGrading.efficiencyInput(for:)` reuses the single `grade(_:)` with `costContext == nil`, keeping Manager grading outcome-independent without touching `ScorecardModel`. `#expect(with.currentWeek == without.currentWeek)` and related DTO tests lock the graded aggregate invariant.
- **Green — data-completeness fixes are well tested:** per-Manager rollups (`eachManagerGetsItsOwnRollup`), legacy-key supersession (`recomputedManagerWeekSupersedesTheLegacyRow`), DTO per-Manager caps, and compaction accumulation/survival across refresh all have targeted coverage.
- **Green — intentional behavior change is documented:** wiring `PostCompact` → `noteCompactionEvent` will apply compaction deductions to work-session grades for the first time; ADR 0008 addendum and PR description call this out clearly.
- **Green — build/tests:** `swift build` succeeds; CrowCore ScorecardDTO tests (18), CrowEngine Manager usage tests (6), CrowPersistence back-compat test (1), and `scorecard.test.js` (32 assertions) all pass locally.

### Summary Table

The implementation is solid and well-tested; updating the empty-state copy to match the new efficiency-graded Manager metering should be a one-line fix.

---

dgershman:
Fixed in d09a607e — thanks, that was a real contradiction.

**Yellow (stale empty-state copy):** `scorecardEmpty()` now reads *"A Manager session never completes, so it is metered in its own section instead — graded on efficiency, but never on outcomes."* I also found two more instances of the same drift you'd have hit later: the Rebuild button's "recomputes the ungraded Manager rollups" comment (`app.js`) and a "for the ungraded bucket" comment in `SessionService` hydration.

The copy drifted because **nothing covered that sentence** — it took a human read to catch. So rather than just editing the string, I added four assertions to `scorecard.test.js`: two negative (no "never graded", no "ungraded" in the empty-state message) and one positive (it must state the efficiency/outcomes split). The next change to the Manager's posture now has to update this copy instead of silently contradicting the card below it.

### Verification
- `npm run test:ci` — green; the scorecard suite is now 36 assertions
- `swift test --package-path Packages/CrowEngine` — 726 pass

⚠️ **CrowDaemon's suite was not run.** It stands up real HTTP/WebSocket servers, and on this host that takes down the live `crowd` serving the environment. Its two `app.js` asset tests (`WebTerminalAssetTests`, `WebNotificationCenterTests`) are already red on `main` independently of this branch, but I want to be explicit that I did not verify this `app.js` edit against them locally rather than imply otherwise — CI will.

dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Store migration is load-bearing and handled correctly: new `ManagerWeeklyUsage` fields are `Optional` to avoid all-or-nothing `JSONStore` decode failures that would wipe sessions (`ManagerWeeklyUsage.swift:20-36`, `JSONStoreTests.swift:383-428`).
- Compaction wiring uses a shared constant (`SessionHookState.compactionEventName`) rather than scattering magic strings, reducing mismatch risk between counter and handler (`AppState.swift:816-822`, `EngineRouter.swift:1396-1404`).
- No new attack surface: `hook-event` compaction handling only increments counters on an exact event-name match; `get-scorecard` remains read-only with no credential exposure.
- Manager identity on the wire is resolved server-side from internal session IDs/names — no user-controlled HTML injection path (web uses `textContent` throughout `scorecard.test.js`).

**Concerns:**
- None identified.

### Code Quality
- **Green** — The efficiency-vs-outcomes split is structurally enforced: `EfficiencyGrading.efficiencyInput(for:)` leaves `costContext` nil, so `costPerShippedDeduction` never applies to Managers; graded aggregate invariants are pinned by `managerUsageNeverTouchesTheGradedMath` and `managerGradeIsOutcomeIndependent`.
- **Green** — Race between `refreshManagerUsage` and `noteManagerCompaction` is correctly fixed by reading `compactionCount` inside `store.mutate`, with a regression test that interleaves the two (`SessionService.swift:4270-4277`, `SessionServiceManagerUsageTests.swift:127-158`).
- **Green** — Per-Manager attribution closes a real data gap: `managerUsageProvider` now takes `UUID`, and rollups use composite keys (`SessionService.swift:4248-4261`, `CrowDaemon.swift:260-265`).
- **Green** — DTO projection handles legacy bare keys, deduplication, and per-Manager week capping cleanly (`ScorecardDTO.swift:101-131`).
- **Green** — Test coverage is thorough: 18 `ScorecardDTOTests`, 6 `SessionServiceManagerUsageTests`, store back-compat, daemon `get-scorecard` reseed path, and 32 web assertions in `scorecard.test.js`. All ran green locally (`swift build`, targeted Swift test suites, `npm run test` in web-tests).
- **Green** — Intentional behavior change (compaction deductions now active for work sessions) is well-documented in the PR body and ADR addendum; this is the metric working for the first time, not a regression.
- **Green** — Consider a quick live `crow get-scorecard` smoke test with telemetry enabled before merge; DTO passthrough and render paths are covered headlessly, but the author noted this wasn't run against a live daemon.

### Summary Table

---

PR #996, opened 2026-08-12
Refit the web terminal to the visual viewport (CROW-988)

### LINKED ISSUE #988: Mobile web: software keyboard hides the prompt line — terminal never refits to the visual viewport
## Problem

On the Crow web UI on a phone, opening the software keyboard hides the agent's prompt line. The terminal keeps rendering at its full pre-keyboard height, so the bottom rows — where the prompt sits — end up underneath the keyboard, with no way to scroll them into view.

## Root cause

The web terminal surface is keyboard-blind. Three specific gaps, all in `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/`:

**1. No `visualViewport` handling anywhere.** `rg visualViewport` over `*.html *.js *.css` in the web resources directory returns zero hits.

On iOS Safari (and Chrome/Android by default), opening the keyboard shrinks the **visual** viewport but leaves the **layout** viewport unchanged. `window.innerHeight` and `height: 100%` therefore still report the full pre-keyboard height. The only signal that anything changed is `window.visualViewport`, which nothing subscribes to.

**2. The resize listener fires on the wrong event.** `app.js:4600`:

```js
window.addEventListener('resize', fitTerminal);
```

`window.resize` does not fire for the iOS keyboard (layout viewport is unchanged), so `fitTerminal()` never re-runs and the xterm grid is never refit to the reduced visible area.

**3. `terminal.html` sizes to a keyboard-blind 100%.** `terminal.html:9-21`:

```css
html, body { height: 100%; overflow: hidden; }
#terminal  { height: 100%; }
```

With `overflow: hidden` on `body`, the occluded bottom rows are not reachable by scrolling either.

Also note `terminal.html:5` is the only viewport meta in the web resources missing `viewport-fit=cover` — `index.html:5` has it. Neither sets `interactive-widget`.

`app.css:36` already uses `100dvh` for the shell (*"exclude mobile browser chrome so the last row is reachable"*), so the mobile-height problem was recognized for the outer chrome but never carried into the terminal frame. `100dvh` alone would not fix this anyway — dynamic viewport units track browser chrome, not the keyboard, on iOS.

## Proposed fix

1. **Subscribe to `visualViewport`** in the web terminal. On `resize` / `scroll`, size the terminal host to `visualViewport.height` (offset by `visualViewport.offsetTop`) and call `fitAddon.fit()`. Guard for absence — `visualViewport` is undefined in some embedded webviews; fall back to current behavior.
2. **Keep the prompt pinned after refit.** Once refit, scroll the terminal to the bottom so the prompt line lands just above the keyboard. The existing `xterm-addon-crow-jumpbottom` addon is the natural place for this.
3. **Add `interactive-widget=resizes-content`** to the viewport meta so Chrome/Android resizes the layout viewport instead of overlaying, which makes the Android path mostly self-correcting.
4. **Add `viewport-fit=cover` to `terminal.html:5`** for parity with `index.html`, and use `env(safe-area-inset-bottom)` so the prompt clears the home indicator once the keyboard closes.

Per the three-front-end note in the harness docs, the desktop `CrowTerminal/.../xterm/terminal.html` does not need this (no software keyboard) — but if the behavior is implemented as an xterm addon rather than inline script, both surfaces stay in sync and the web/desktop drift doesn't widen.

## Acceptance

- On iOS Safari and Chrome/Android, opening the keyboard leaves the agent's prompt line fully visible directly above it.
- Closing the keyboard restores the full-height grid with no leftover gap or clipped rows.
- Rotating with the keyboard open re-fits correctly.
- Desktop and non-`visualViewport` webviews are unchanged.

## Notes

Reported against the mobile web app in a phone browser. The exact device/OS was not captured; the fix above covers both the iOS overlay behavior and the Android resize behavior, so it should not need narrowing to one.

dgershman:
Closes #988

## Problem

On a phone, opening the software keyboard hid the agent's prompt line. iOS Safari (and Chrome/Android by default) shrink the **visual** viewport and leave the **layout** viewport alone, so `window.innerHeight`, `height: 100%` and even `100dvh` all still reported the full pre-keyboard height and `window.resize` never fired. The grid kept rendering at its old size, its bottom rows ended up under the keyboard, and `overflow: hidden` meant they couldn't be scrolled back into view either.

Nothing in `Resources/web/` subscribed to `window.visualViewport` — the only signal that anything had changed.

## Fix

New shared addon `xterm-addon-crow-viewport.js`. On `visualViewport` resize/scroll it sizes the terminal's host element so its bottom edge lands on the bottom of the *visible* area, asks the page to refit, and pins the prompt.

Written as an addon rather than inline script so the surfaces that load it stay in sync instead of hand-mirroring the logic per front-end — the parity drift that bit reflow-debounce #661/#662 and the mouse-mode swallow #776.

A few decisions worth calling out:

- **Measures from `getBoundingClientRect().top`**, not from `visualViewport.height` alone. `visibleBottom = offsetTop + height` is in layout-viewport coordinates — the same space `getBoundingClientRect()` reports in — so the same maths works for the app's grid (which sits below a header and a tab bar) and the full-page debug terminal, and a shifted visual viewport is handled rather than assumed away.
- **Pins one frame after the refit, not before.** Both pages coalesce their fit to the next frame, so a `scrollToBottom()` issued immediately would land on the pre-fit geometry and the fit would drop the prompt straight back out of view.
- **Never calls `fitAddon.fit()` itself.** Each page wraps its fit in its own dedup/ownership rules — `app.js` gates on focus so a background tab can't steal tmux's shared window size (#667) — so the addon hands back through an `onResize` callback instead of reaching past them.
- **A scrolled-up viewport is left alone.** Opening the keyboard pins (revealing the prompt is the point), and so does a refit while already at the live edge (a shrinking grid can push a pinned viewport off the bottom); someone who scrolled up with the keyboard open keeps their place.
- **A host with no layout box is never sized.** The app hides the whole terminal while a board is open, and a board has its own focusable inputs, so a zero-rect measurement would strand a full-viewport height on it. Releasing a height stays unconditional, so hiding the terminal mid-keyboard can't strand the override for good.

### Staying out of the way

The acceptance criterion "desktop and non-`visualViewport` webviews are unchanged" is enforced structurally, not by hoping:

- absent `visualViewport` → `activate()` returns before registering anything;
- occlusion ≤ 120px → no-op. That threshold sits above what mobile browser chrome (iOS Safari's collapsing toolbars, ~40-60px) and desktop pinch-zoom produce, and below the shortest phone keyboard (landscape, small device, comfortably >150px);
- nothing is written to `host.style` until a keyboard-sized inset actually appears, and the inline height found there is restored verbatim when it goes.

### Also

- `interactive-widget=resizes-content` on both viewport metas — Chrome/Android then shrinks the *layout* viewport, which puts it back on the ordinary `100dvh` + window-`resize` path the pages already handle (and makes the addon correctly no-op there, since occlusion reads ~0). iOS ignores the hint, which is what the addon is for.
- `viewport-fit=cover` on `terminal.html` for parity with `index.html`, paired with `env(safe-area-inset-bottom)` so its prompt clears the home indicator once the keyboard closes. The two go together: `cover` is what creates the inset that the padding then clears.

**`app.css` is deliberately untouched.** The ticket's safe-area point is scoped to `terminal.html` here: the app shell already handles its own safe area (`100dvh` plus the existing `env()` insets on the statusbar and hamburger), and adding padding under `#terminal-wrap` would put a permanent black band above the home indicator — a visible regression on the surface this is being fixed for, not a fix.

`CrowTerminal/Resources/xterm/terminal.html` is also untouched: that page is the **retired** WKWebView surface (its own header says fixes belong in the daemon's web resources, and it's never served). The addon file lives in that directory only because `/xterm/:file` serves the library assets out of it.

## Tests

- `web-tests/visual-viewport.test.js` — 31 checks driving the real addon against a fake `visualViewport`, fake xterm and a hand-driven rAF: geometry (including `offsetTop`), the no-`visualViewport` and sub-threshold no-ops, keyboard open/close/rotate, same-geometry dedup, scrolled-up users, hidden hosts, and `dispose` unwinding listeners + inline height. Wired into both `test` and `test:ci`.
- `BundledResourcesTests.viewportAddonIsBundled` — the file ships (it only does so because `Resources/xterm` is `.copy`d wholesale; a missing file would be a silent 404 and a keyboard-blind terminal, not a build error).
- `WebTerminalAssetTests` — three new drift guards: both live pages construct the addon, both hand it an `onResize` (an omitted one would resize the host and leave the grid at its old row count — the clipped-rows half of the bug, silently), both ship the `<script src>`, and both viewport metas carry the keyboard-aware keywords.

Full `npm run test:ci` green (15 suites), `BundledResourcesTests` green (9/9), `WebTerminalAssetTests` 15/16 — the one failure, `refreshTerminalsRebindsTheActiveTerminalToTheFreshRow`, is pre-existing on `main` (it asserts a string `|| terminals[0] || null` that doesn't appear in `origin/main`'s `app.js` either) and is untouched by this diff.

## Test plan

On a phone (iOS Safari and Chrome/Android):

1. Open a session, tap the terminal to raise the keyboard → the agent's prompt line sits fully visible directly above it.
2. Type and submit → output flows, prompt stays above the keyboard.
3. Dismiss the keyboard → full-height grid returns, no leftover gap, no clipped rows.
4. Rotate with the keyboard open → refits to the new geometry, prompt still visible.
5. Scroll up through history with the keyboard open → the viewport stays where you put it.

On desktop: resize the window, drag the splitter, open the board and come back — all unchanged.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -5416,6 +5416,16 @@ function ensureTerminal() {
   // Jump-to-bottom pill (#668), shared with the desktop surface. Must load after
   // open() so the addon can anchor its button to the terminal's container.
   term.loadAddon(new CrowJumpBottomAddon.CrowJumpBottomAddon());
+  // CROW-988: keep the grid inside the *visual* viewport so a phone's software
+  // keyboard can't sit on top of the prompt line. Inert without
+  // window.visualViewport and while nothing keyboard-sized occludes the page, so
+  // desktop is untouched. Feeds the same coalesced fitTerminal every other
+  // resize route uses. Guarded like the WebGL load below — a failed asset fetch
+  // must not take ensureTerminal (and with it the whole terminal) down.
+  if (typeof CrowViewportAddon === 'object' && CrowViewportAddon
+      && typeof CrowViewportAddon.CrowViewportAddon === 'function') {
+    term.loadAddon(new CrowViewportAddon.CrowViewportAddon({ onResize: fitTerminal }));
+  }
   // WebGL renderer for throughput; must load after open(). Falls back to the
   // default renderer if the GL context is unavailable or gets lost.
   try {
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/index.html
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/index.html
@@ -2,7 +2,12 @@
 <html lang="en">
 <head>
   <meta charset="utf-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover">
+  <!-- CROW-988: `interactive-widget=resizes-content` makes Chrome/Android shrink
+       the LAYOUT viewport for the software keyboard, so `100dvh` below and the
+       existing window-`resize` fit path self-correct there. iOS ignores it and
+       only shrinks the visual viewport, which is what xterm-addon-crow-viewport
+       handles. -->
+  <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover, interactive-widget=resizes-content">
   <title>Crow</title>
   <link rel="icon" type="image/svg+xml" href="/brand.svg">
   <link rel="stylesheet" href="/xterm/xterm.css">
@@ -76,6 +81,7 @@
   <script src="/xterm/xterm-addon-web-links.js"></script>
   <script src="/xterm/xterm-addon-webgl.js"></script>
   <script src="/xterm/xterm-addon-crow-jumpbottom.js"></script>
+  <script src="/xterm/xterm-addon-crow-viewport.js"></script>
   <script src="/app.js"></script>
   <script src="/settings.js"></script>
 </body>
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/terminal.html
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/terminal.html
@@ -2,7 +2,12 @@
 <html lang="en">
 <head>
   <meta charset="utf-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <!-- CROW-988: parity with index.html. `viewport-fit=cover` lets the #1e1e1e
+       background go full-bleed (the safe-area inset on body below keeps the last
+       row off the home indicator), and `interactive-widget=resizes-content`
+       makes Chrome/Android shrink the layout viewport for the software keyboard
+       so the plain `resize` fit below self-corrects there. -->
+  <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover, interactive-widget=resizes-content">
   <title>Crow Terminal</title>
   <link rel="stylesheet" href="/xterm/xterm.css">
   <style>
@@ -13,6 +18,13 @@
       height: 100%;
       overflow: hidden;
       background: #1e1e1e;
+    }
+    /* Under viewport-fit=cover the page runs under the home indicator, so inset
+       the grid — otherwise the prompt line sits behind it once the keyboard
+       closes (CROW-988). border-box keeps the page exactly one viewport tall. */
+    body {
+      box-sizing: border-box;
+      padding-bottom: env(safe-area-inset-bottom);
     }
     #terminal {
       width: 100%;
@@ -40,6 +52,7 @@
   <script src="/xterm/xterm-addon-fit.js"></script>
   <script src="/xterm/xterm-addon-image.js"></script>
   <script src="/xterm/xterm-addon-crow-jumpbottom.js"></script>
+  <script src="/xterm/xterm-addon-crow-viewport.js"></script>
   <script>
     // WebSocket transport variant of CrowTerminal's terminal.html. The xterm.js
     // config block is copied verbatim so font/scrollback/inline-image behavior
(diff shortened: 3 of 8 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Client-side DOM manipulation only — sets `host.style.height` from `Math.round()`-computed pixel values; no user-controlled strings reach the DOM or network layer.
- Graceful degradation: absent `visualViewport`, failed script load (guarded `typeof` check in both surfaces), and sub-threshold occlusion all no-op without touching the host.
- `dispose()` removes `visualViewport` listeners and restores the saved inline height, preventing listener leaks and stranded styles on recycled hosts.
- `onResize` and `scrollToBottom` are wrapped in try/catch so a page-level fit failure cannot take down the addon.
- No secrets, credentials, or new attack surface introduced.

**Concerns:**
- None identified.

### Code Quality
- **Well-factored addon** (`xterm-addon-crow-viewport.js`): clear separation of concerns — measures geometry, sizes the host, delegates fit/PTY resize via `onResize`, and defers `scrollToBottom` one frame so it lands after the page's coalesced refit. Matches the existing Crow addon UMD pattern.
- **Integration is consistent**: both `app.js` and `terminal.html` load the addon after `term.open()`, wire `onResize` to each surface's own fit path (`fitTerminal` / `sendResize`), and guard the load like WebGL so a 404 cannot break terminal init.
- **Desktop safety is structural**, not aspirational: no `visualViewport` → early return; occlusion ≤ 120px → restore/no-op; inline height is never written until a keyboard-sized inset appears.
- **Edge cases handled**: hidden host (`clientHeight < 1`), scrolled-up user preservation, `offsetTop` shift, rotation, same-geometry dedup, mid-keyboard hide/show, and dispose during an in-flight rAF.
- **Test coverage is strong**: 31 jsdom checks exercise the real addon file; Swift drift guards pin script tags, construction, `onResize` wiring, and viewport meta keywords on both live pages; `BundledResourcesTests.viewportAddonIsBundled` catches a silent 404.
- **Static analysis**: `npm run test:ci` green (15 suites); `visual-viewport.test.js` 31/31; `BundledResourcesTests.viewportAddonIsBundled` passed; `swift build` succeeded. `WebTerminalAssetTests.refreshTerminalsRebindsTheActiveTerminalToTheFreshRow` still fails — pre-existing on `main` (asserts `activeTerminal = terminals.find(` but the function now uses a `pendingTerminalId`-aware expression); untouched by this diff.

### Summary Table

| # | Finding | Color |
|---|---------|-------|
| 1 | Addon design cleanly isolates visual-viewport logic with proper lifecycle, guards, and callback delegation | Green |
| 2 | Comprehensive unit + drift tests give high confidence in geometry math and cross-surface parity | Green |
| 3 | `terminal.html`'s `sendResize` is not frame-coalesced (unlike `app.js`'s `fitTerminal`), so the debug surface may emit more PTY resize frames during keyboard animation — acceptable for that page, but worth noting if parity ever matters | Green |

---

PR #998, opened 2026-08-12
Stop offering Start Review on Waiting-on-author rows (CROW-997)

### LINKED ISSUE #997: Reviews board: Waiting on author rows still offer Start Review (CROW-990 follow-up)
Follow-up to #990 / #994, which added the **Waiting on author** bucket.

## Symptom

A card under **Waiting on author** renders a **Start Review** button and is tickable in the board's batch "Start Review (N)" mode. The whole point of that heading is that the ball is with the author — there is nothing new to look at until they push — so the primary action on those cards invites redundant work, and the batch bar makes it easy to kick off eight of them at once by accident.

Verified against the CROW-990 implementation:

```
group=waiting_on_author  kickoff_action=create
```

## Cause

`ReviewsPayload.build` short-circuits `kickoff_action` to `skip` only for PRs that are no longer open:

```swift
let action = r.isCompleted ? .skip : IssueTracker.reviewKickoffAction(...)
```

A Waiting-on-author row *is* open, so `reviewKickoffAction` runs — and with no live review session (the previous round was auto-completed by `decideReviewCompletions` rule 1 the moment the verdict was submitted) it falls through to:

```swift
if reviewSessionID == nil && existingByPRSessionID == nil { return .create }
```

The web's `reviewIsActionable` treats `create` as actionable, which is what renders the button and makes the row selectable.

## The case a blanket suppression would break

**The author can push without re-requesting.** Then there genuinely is something new to review, the PR is still not in `review-requested:@me`, and it stays under Waiting on author. Suppressing the button unconditionally would strand exactly that PR with no way forward from the board.

Two things bear on how much that matters:

- `headRefOid` vs the session's `lastReviewedHeadSha` already distinguishes "author pushed" from "author hasn't touched it" — that comparison is `reviewKickoffAction`'s existing `shaAdvanced` branch, but it needs a `linkedSession` to compare against and there isn't one once the round auto-completed. Making this work probably means recording the head SHA the verdict was submitted against on the `ReviewRequest`, not on the session.
- With `--auto-re-request-review` on (CROW-921), a pushed fix gets the reviewer re-added automatically, which moves the PR back into the requested queue and out of this group on its own. So for anyone running that watcher the stranding case largely self-resolves — but the toggle is not mandatory and the board shouldn't depend on it.

## Options

1. **Suppress unconditionally** — `kickoff_action: skip` for `waiting_on_author`, same as the completed groups. Simplest; strands the author-pushed-without-re-requesting case for anyone with the CROW-921 watcher off.
2. **Suppress unless the head moved** — offer `re_review` only when the PR's current head differs from the head the viewer's last verdict was submitted against. Correct, and needs a new field carrying that SHA through `parseReviewRequests` → `ReviewRequest`.
3. **Demote rather than suppress** — keep the action available but render it as a secondary/ghost button rather than `action-primary`, and exclude the row from batch selection. Cheapest way to stop the accidental-batch problem without deciding the SHA question.

Option 2 is the one that matches what the group means, and 1 is a reasonable interim if 2 looks expensive. Worth deciding whether the batch-selection change should land regardless of which is chosen — that is the sharper of the two symptoms.

## Acceptance

- [ ] A Waiting-on-author card does not present Start Review as its primary action
- [ ] Waiting-on-author rows are not selectable in the board's batch Start Review mode
- [ ] A PR whose author pushed new commits without re-requesting is still reachable from the board (whichever option lands, this case is explicitly decided rather than overlooked)
- [ ] `crow list-reviews` and the web board agree, from one rule, on which rows are actionable

## Related

- #990 / #994 — added the Waiting on author bucket
- #982 / #984 — the three-group split it extends
- CROW-921 — the auto-re-request watcher that mitigates the stranding case

dgershman:
Closes #997

A card under **Waiting on author** rendered **Start Review** as its primary action and was tickable in the board's batch "Start Review (N)" mode. The heading exists to say the ball is with the author — there is nothing new to look at until they push — so the card invited redundant work, and the batch bar put eight redundant rounds one click away.

## Cause

`ReviewsPayload.build` short-circuited `kickoff_action` to `.skip` only for PRs that are no longer open. A Waiting-on-author row *is* open, so `reviewKickoffAction` ran — and the round was auto-completed by `decideReviewCompletions` rule 1 the instant the verdict landed, so there was no `linkedSession` for `shaAdvanced` to compare against and every such row fell through to `.create`.

## Approach — ticket option 2

Suppress it, but only while there is genuinely nothing new. The author can push **without** re-requesting: GitHub never puts the PR back in `review-requested:@me`, it stays under this heading, and it does have something to review. Telling the two cases apart needs the head the verdict was submitted against, which `Session.lastReviewedHeadSha` cannot supply once the round is gone.

- `ReviewRequest.viewerLastReviewedHeadSha` carries it, parsed from GitHub's `PullRequestReview.commit.oid` in the same review node as the timestamp and verdict. Verified against the live API before building on it — it populates, including the discriminating shape (#966 here: `CHANGES_REQUESTED` at `645fab6…`, head now `41388d2…`).
- `viewerReviewCoversCurrentHead` compares it to `headRefOid` and answers `false` unless it can **prove** the heads match. The asymmetry is deliberate: suppressing claims "there is nothing here to look at" and takes the row's only way forward off the card, so an unfetched SHA must not be allowed to make that claim. Same "absence is not evidence" rule the `state` and verdict fields already follow.
- The decision moves into `ReviewsPayload.kickoffAction(for:group:linked:existing:)`, which now takes the group — two of the four groups mean "nothing to review" for reasons the session logic cannot see, and they now sit together instead of one being an inline ternary.

CROW-921's `--auto-re-request-review` watcher mitigates the stranding case by re-adding the reviewer on a pushed fix, but it is optional, so the board does not depend on it.

## One rule, both clients

`kickoff_action` is still the single field, and `.skip` is what makes a row unselectable as well as button-less — so the button and the batch checkbox disappear together rather than off two rules that could drift. The browser only re-labels the field; it does not re-derive it. `viewer_last_reviewed_head_sha` also rides the payload beside `head_ref_oid`, so `crow list-reviews` shows the comparison rather than an unexplained `skip`.

`crow start-review --url` is unchanged and still works on anything, including a quiet Waiting-on-author PR — the board declines to *suggest* a round, it does not forbid one. The payload's doc comment now says so, since `kickoff_action` reads as "what this row offers", not "what `createReviewSession` would do if called anyway".

## Acceptance

- [x] A Waiting-on-author card does not present Start Review as its primary action
- [x] Waiting-on-author rows are not selectable in batch Start Review mode
- [x] A PR whose author pushed without re-requesting keeps both, by an explicit head comparison rather than by oversight
- [x] `crow list-reviews` and the web board agree from one rule

## Test plan

- `CrowEngine` — 749 tests pass. New in `ReviewsPayloadTests`: matched heads skip; a moved head offers `create`; three unknown-SHA permutations keep the button; matched heads do **not** suppress a row still in the requested queue; the reviewed head is serialized.
- `CrowProvider` — 58 tests pass. New in `BackendsTests`: the reviewed tail carries the verdict's head; a null `commit` leaves it unknown (and `viewerReviewCoversCurrentHead` false); a later verdict without a commit **clears** the earlier round's SHA rather than leaving a two-rounds-old head to compare against; all three searches select `commit { oid }`.
- `web-tests/board.test.js` — 109 pass, including new checks that a Waiting-on-author card has no Start Review button and no batch checkbox, and that a pushed-since-review one has both again.
- `CrowDaemon` builds; its Swift tests were not run here (they boot `crowd` on this host).


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Conservative fail-open design: `viewerReviewCoversCurrentHead` returns `false` when either SHA is missing, so suppression never hides actionable work on partial or ambiguous data (`ReviewRequest.swift:110-112`).
- No new attack surface — SHAs are parsed from GitHub GraphQL responses and compared as opaque strings; no shell interpolation or user-controlled input paths.
- `crow start-review --url` remains an explicit escape hatch; the board only declines to *suggest* redundant rounds, it does not block server-side session creation (`ReviewsPayload.swift:34-40`).
- Clearing `viewerLastReviewedHeadSha` when a later verdict lacks `commit` prevents stale head comparisons that could incorrectly suppress Start Review (`GitHubCodeBackend.swift:902-908`).

**Concerns:**
- None identified.

### Code Quality
- **Green:** Single rule in `ReviewsPayload.kickoffAction(for:group:linked:existing:)` drives both the Start Review button and batch selectability via existing `reviewIsActionable` — no client-side drift (`ReviewsPayload.swift:182-196`, `app.js:4960-4962`).
- **Green:** GraphQL query updated consistently across all three monitored-PR searches with a regression guard (`GitHubCodeBackend.swift:542-566`, `BackendsTests.swift:2137-2147`).
- **Green:** Thorough test matrix — matched heads skip, moved head offers `create`, three unknown-SHA permutations keep the button, queued reviews unaffected, serialization verified (`ReviewsPayloadTests.swift:411-507`; `BackendsTests.swift:2010-2134`; `board.test.js:435-477`).
- **Green:** Two minor comment-line merge artifacts in test files (`BackendsTests.swift:2149`, `ReviewsPayloadTests.swift:509`) — cosmetic only, no behavioral impact.

### Summary Table

**Static analysis:** `swift test` — CrowEngine 749 tests pass, CrowProvider 177 tests pass; `swift build` — CrowDaemon succeeds. Web tests (`board.test.js`) not run here (missing `jsdom` in this environment); PR author reports 109 pass.

---

PR #1003, opened 2026-08-12
Move the session switcher to cmd+/ and stop hijacking Shift+Tab (CROW-1002)

### LINKED ISSUE #1002: Move task manager shortcut off Shift+Tab; investigate keystroke hijacking
A task-manager feature was added that used **Shift+Tab**, but it kept overriding the permissions-swap shortcut on all coding agents.

We moved it to **Tab+Escape**, but that sequence also isn't great. **Cmd+/** is probably a better binding.

Separately, **Shift+Tab no longer works** — investigate whether something is still hijacking that keystroke (even after the move away from it).

## Acceptance criteria
- Task manager opens on a non-conflicting shortcut (proposed: **Cmd+/**).
- Agent permission-swap shortcuts (Shift+Tab where applicable) work again without Crow intercepting them.
- Confirm whether any remaining global/local key handler still captures Shift+Tab and remove or gate it.

dgershman:
Closes #1002

## The Shift+Tab hijack

CROW-980 moved the switcher off `shift+tab` by changing the **default** — and a default reaches only a config with no `switcher.binding` key at all. `AppConfig` encodes the whole struct, so every install whose config had ever been saved kept the old value and went on swallowing the agents' permission-mode cycle in every focused terminal (`captureInTerminal` defaults to true).

Confirmed on the live config for this devroot before touching anything:

```json
"switcher": { "binding": "shift+tab", "captureInTerminal": true, ... }
```

That is the answer to the ticket's "investigate whether something is still hijacking that keystroke": **the persisted config was**, not a leftover handler. `app.js` had no `shift+tab` branch of its own — the generic binding path was doing exactly what the stored binding told it to. It reached the terminal through two paths, both gated on `switcherCaptureInTerminal`: the document-level capture listener (`onSwitcherKeyDown`) and xterm's `attachCustomKeyEventHandler` (`handleTerminalKey`).

## What changed

**Default is now `cmd+/`.** An ordinary modifier chord: opens on press, commits on release of Cmd like a macOS app switcher — no 1.5s prefix timer to arm, no intervening keystroke to break the sequence, no Enter needed to commit. macOS produces no key data for a Cmd chord in xterm, so capturing it leaks nothing to the PTY. `esc+tab` remains a legal binding and the prefix machinery is untouched; it merely stopped being the default, so an install that has it keeps it. No Cmd key? `crow ui set --switcher-binding 'ctrl+/'`.

**`shift+tab` is reserved three ways**, so a fourth round of this isn't possible:

| | Covers |
|---|---|
| Migrated to the default on decode | Legacy configs, and the whole-blob `set-config` path (it decodes through the same initializer) |
| Rejected by `ui-set` with an explanation | New writes, incl. `crow ui set` |
| Never captured in a focused terminal, regardless of `captureInTerminal` | A hand-edited `config.json`, and a browser still holding pre-migration config |

Migration and validation share one set (`SwitcherSettings.reservedBindings`) deliberately: a binding the decoder silently rewrites but the setter still accepts would revert on the next daemon start with nothing to explain why — a worse bug than the one being fixed. Matching is trimmed and case-folded, since the setter stores the string verbatim while the client parses it case-insensitively, so `Shift+Tab ` would otherwise slip through as the same swallowed keystroke.

The terminal gate is **chord-scoped, not a blanket opt-out** — it compares parsed chords including the prefix field, so bare `tab` and `esc+tab` are correctly left alone and every non-reserved binding is still captured on the same path.

## Scope notes

- **`esc+tab` is not migrated.** It takes nothing from the terminal, so there is no correctness reason to reset it, and rewriting a binding the setter still accepts is the silent-revert trap above. Only the *default* moved.
- **Only the binding is rewritten** by the migration; the rest of a user's switcher settings survive.
- **The old chord can't be restored, by design.** It buys one keystroke for a control with three other ways in (sidebar, click, the configured chord outside the terminal) and costs the agent a control with no alternative binding at all.
- Two cosmetics that fell out of the new default: `/` gets an explicit key alias (the length-1 branch yielded `['/','/']`, which printed as a duplicate in the hint line), and the hint separator moved from `/` to `or` now that the cycle key is itself `/`.
- `make docs` also regenerates `CLIVersion.swift` (0.1.0 → 0.2.0, stale since #97). **Left out of this PR** as unrelated drift — worth a follow-up.

## Acceptance criteria

- [x] Task manager opens on a non-conflicting shortcut — `cmd+/`.
- [x] Agent permission-swap shortcuts work again without Crow intercepting them.
- [x] Confirmed what was still capturing Shift+Tab (the persisted binding, not a stray handler) and gated it — the terminal keeps the chord whatever the config says.

## Test plan

All run per-package (`make test` boots `crowd` on this host):

- `swift test --package-path Packages/CrowCore` — **622 passed**. 7 switcher config tests: new default, `shift+tab` migration (config-shaped and via the helper), casing/whitespace variants, blank-is-unset, `esc+tab` preserved, unrelated binding preserved.
- `swift test --package-path Packages/CrowCLI` — **354 passed**, including `generatedDocCoversEveryCommandAndFlag` (docs regenerated via `make docs`).
- `npm run test:ci` in `web-tests` — **all suites green**. 12 new switcher.test.js checks (`cmd+/` parsing/matching/hint, the reserved-chord predicate, the capture gate) and 5 in key-handler.test.js. The two pre-existing key-handler assertions that asserted Shift+Tab *was* captured now assert the opposite — that inversion is the fix.
- `make parity` — notification catalogs, CLI/RPC parity (98 methods), no-NSLog: all pass.
- `swift build --package-path Packages/CrowDaemon --build-tests` — compiles clean, including the 2 new `SettingsHandlerTests` (reject the reserved chord, still accept `esc+tab`). Not executed here: CrowDaemon's suite boots `crowd` and would take down the live environment.

**Manual check after merge:** restart `crowd`, confirm `crow ui get` reports `cmd+/` (the migration rewriting the stored `shift+tab`), then Shift+Tab in a session terminal should cycle the agent's permission mode and Cmd+/ should open the switcher.


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Reserved-binding validation at the `ui-set` RPC boundary prevents persisting a chord that would hijack agent permission-mode cycling (`RPCHandlers.swift:1692-1697`).
- Whole-blob `set-config` still decodes through `AppConfig`, so `shift+tab` is migrated on load rather than silently stored (`RPCHandlers.swift:1137-1141` + `SwitcherSettings.migratedBinding`).
- Defense-in-depth terminal gate (`switcherCapturesInTerminal`) ensures Shift+Tab reaches the PTY even with stale in-memory config or hand-edited `config.json` (`app.js:2655-2657`, `app.js:3040`, `app.js:5783`).
- No secrets, injection surfaces, or auth-boundary changes introduced.

**Concerns:**
- None.

### Code Quality
- Root cause correctly identified: CROW-980 changed only the default, so persisted `shift+tab` bindings kept swallowing the keystroke. Migration + rejection + terminal gate closes all three paths described in the PR.
- `SwitcherSettings.reservedBindings` / `isReservedBinding` / `migratedBinding` in CrowCore is clean, test-covered, and shared by decode and write paths.
- `cmd+/` default is well-motivated (modifier chord, no PTY leakage on macOS, commits on release). Hint-line `or` separator and `/` key alias are sensible cosmetics.
- Tests are thorough: 7 new CrowCore config tests, 2 CrowDaemon handler tests, 17 new web-tests assertions; key-handler tests correctly invert the prior Shift+Tab capture expectation.
- `swift build` succeeds; targeted `AppConfigTests` and web `switcher.test.js` / `key-handler.test.js` suites pass locally (56 + 51 assertions).

### Summary Table

**Green (non-blocking):**
1. `SWITCHER_TERMINAL_RESERVED` in `app.js` duplicates `SwitcherSettings.reservedBindings` in Swift — intentional belt-and-suspenders for stale browser state, but worth a brief cross-reference comment if more reserved chords are ever added.
2. Non-macOS users without a Cmd key will need `crow ui set --switcher-binding 'ctrl+/'`; the PR documents this, and the CLI help text covers it.

---

PR #1005, opened 2026-08-13
Add a read-only MCP server with scoped tokens (CROW-1004)

### LINKED ISSUE #1004: Add read-only crowd MCP server with scoped tokens
Expose Crow's control plane to Cowork and Grok bot via MCP, without requiring a Crow-launched agent session. `crowd` already has JSON-RPC (Unix socket + `/rpc`); MCP is the protocol those clients speak.

**This ticket is read-only.** Do not add `send_prompt` / `crow send` (or any other write that injects into an agent TUI). Prompt insert is deferred.

## Why not 1:1 every RPC

Do not wrap all ~80 RPCs as MCP tools. That blows the model context, invites the wrong tool, and would leak internals remote `/rpc` already refuses (`hook-event`, gateway/web-password, `run-setup`, host-app open, `defaults.binaries`). ADR 0016 keeps the CLI as the full control plane; MCP is a third, task-shaped client.

## v1 tools (read-only)

| Tool | Backed by | Answers |
|------|-----------|---------|
| `get_board_summary` | `list-sessions` / `list-sessions-live` | Counts: active, inReview, paused (plus names) |
| `list_sessions` | `list-sessions` | Filter by status |
| `get_session` | `get-session` | One session + ticket/PR |
| `list_stuck_sessions` | `list-sessions-live` | Waiting on input, checks failing, rebase/auto-merge stuck, idle-too-long (`activity` + `attention` + auto-merge/rebase verdicts) |
| `list_tickets` / `list_reviews` | existing board RPCs | Optional; keep if they stay small |

Out of scope: `send`, `quick-action`, `new-session`, `set-status`, `delete-session`, jobs writes, config/gateway/admin.

## Capability layer + minted tokens

- Generate the export set from `ParityLedger` (add an MCP flag / scope per row) so a new RPC cannot silently become a Grok tool.
- Mint/revoke tokens in Settings (next to web auth). Store hashes, not the secret.
- Scopes for v1: `sessions:read`, `board:read`. No `prompt:send`, no `sessions:write`, no `admin`.
- `tools/list` must be filtered by the token — a deny-after-call still teaches the model that write tools exist.
- Reuse `RPCWebSocketHandler.localOnlyDenial`: those methods never appear on a remote token.
- Local stdio against the Unix socket can be broader (machine-local trust). A Grok-bot / off-box token is narrow and should expire.
- Remote transport needs HTTP/SSE (or equivalent) plus the same class of auth as the web UI. Local Cowork can use stdio.

## Acceptance criteria

- [ ] MCP server serves the read-only tools above against a running `crowd`.
- [ ] Token mint + scope (`sessions:read` / `board:read`); tool list filtered to the token.
- [ ] No write/prompt-inject tools in this PR.
- [ ] Local-only RPCs remain unreachable over the remote MCP path.
- [ ] Ledger/parity updated so MCP export cannot drift silently.
- [ ] Brief docs: how to point Cowork / Grok bot at it.

dgershman:
Closes #1004

Exposes Crow's board to MCP clients — Cowork locally, a Grok bot off-box — without requiring a Crow-launched agent session. **Six tools over five read RPCs. No prompt-send, no session creation, no config access.**

## Not a 1:1 wrapper

Projecting all ~101 RPC methods into tools would blow the model's context, invite the wrong tool by sheer surface area, re-expose what remote `/rpc` already refuses, and — worst — make every future handler a public API by default, with an invisible failure mode.

So the surface is a **closed, task-shaped allowlist** that the `ParityLedger` gates. `RPCEntry` gained `mcp: MCPExposure`, defaulting to *not exported*: a newly registered RPC stays invisible to MCP until someone types `mcp: .read(scope:)` on its ledger row, which is the diff a reviewer reads (ADR 0016's argument, applied to a second axis).

`MCPLedgerExportTests` fails the build when:

- a tool reads a method with no export flag,
- a flagged method has no tool (an unused standing permission),
- an exported method is a **write**,
- an exported method is **local-only** — the acceptance criterion "local-only RPCs remain unreachable over the remote MCP path", asserted rather than inspected.

There is no passthrough tool and no way to name an RPC method from the wire — all five call sites are string literals in the catalog — so `gateway-*`, `web-password-*`, `mcp-token-*`, `run-setup`, `hook-event` and the host-app launches are unreachable **structurally**, on both transports, including the stdio bridge that bypasses `localOnlyDenial` entirely.

## Two transports, two trust models

| Transport | Credential | For |
|---|---|---|
| `crow mcp serve` (stdio) | none — the 0600 Unix socket is the boundary | a client on this machine |
| `POST /mcp` | `Authorization: Bearer crow_mcp_…` | an off-box client |

A caller who can run `crow mcp serve` could already run every other `crow` verb, so a token there would gate nothing.

**`/mcp` is exempt from `WebAuthMiddleware` and authenticates itself, which is *stricter*, not looser.** That middleware gates on the web-session cookie and is opt-in — on a daemon with no web password it returns authorized for everything — while `/mcp` has no tokenless path and no loopback bypass. `MCPRoutesTests` pins that a tokenless request is refused *even with no web password set*.

## Tokens

Scopes are exactly `sessions:read` and `board:read`. No `sessions:write`, no `prompt:send`, no `admin` — not deferred but **undefined**, so a token cannot name a capability the server might later grow into. `tools/list` is filtered **by the presented token**, since a deny-after-call still teaches the model that hidden tools exist (the spec explicitly permits varying the set by authorization).

Expiry defaults to 90 days; `--no-expiry` must be typed, and `--expires-in 90` is rejected as ambiguous between seconds and days.

**Deliberately SHA-256, not PBKDF2 like the web password.** That protects a low-entropy human secret and is verified once per login; a token is 32 CSPRNG bytes verified on *every request*, where 210,000 iterations would be a self-inflicted DoS. Guessing is 2²⁵⁶ regardless of KDF. Same reasoning as a GitHub PAT; constant-time compared either way. This will read as an inconsistency next to `WebAuthConfig` — it isn't, and both files say so.

`SettingsSecrets` strips token hashes outbound and **restores them from disk inbound**, so a `set-config` round-trip can neither mint a token nor revoke one. Without that line an authenticated remote peer could POST back a record whose hash it chose; `preserveIgnoresBrowserMintedMCPTokens` covers it.

## Two tools that earn their shape

`list_stuck_sessions` joins `list-sessions` (activity/attention) and `list-sessions-live` (PR checks, auto-merge/auto-rebase verdicts) — disjoint payloads that neither answer "what needs a human" alone. It reports `waiting_on_input` / `idle_too_long` / `checks_failing` / `auto_merge_stuck` / `auto_rebase_stuck` with each verdict's own message.

That join surfaced a real gap: `HookNotification.timestamp` is persisted but was dropped by the `list-sessions` handler, so **`idle_too_long` could never fire**. Now emitted as `attention_since` (additive; the web client ignores unknown keys).

`list_tickets` drops the issue `body`, the one unbounded field in that payload — one long issue would otherwise crowd the board out of the model's context.

## Dual-era protocol

Revision `2026-07-28` replaced the `initialize` handshake with per-request `_meta` and a mandatory `server/discover`. It is weeks old and shipping clients still open with the handshake, so both are answered on one endpoint with the era decided **per request** (the spec blesses this). Header/body mirroring is validated for modern requests only — rejecting a legacy client for omitting headers it has never heard of would break the compatibility this exists for.

## Placement

The MCP layer lives in **`CrowIPC`**, which now depends on `CrowCore` (for `MCPScope` / `MCPTokenRecord`, which `ParityLedger` and `AppConfig` need). Acyclic — `CrowCore` has no dependencies at all — but worth flagging: the wire package is no longer a leaf. The alternative was moving `JSONValue` into `CrowCore`, a much larger refactor.

That placement is what makes the gate *run*: `CrowIPC` and `CrowCore` are both in `LINUX_PACKAGES`, while `CrowDaemon` is Darwin-only and its tests do not run on PRs (ADR 0007). Only the Hummingbird route and the SHA-256 stayed in the daemon.

`ParityLedger.localOnlyRPCMethods` mirrors `localOnlyDenial`'s unconditional cases so the export gate can assert disjointness from the Linux lane; `LocalOnlyRPCGateTests` pins the two in **both** directions, because a mirror that can drift is worse than none.

## Verification

```
make parity                                  # 101 methods, catalogs agree, no NSLog
swift test --package-path Packages/CrowCore  # 639 passed
swift test --package-path Packages/CrowIPC   # 120 passed
swift test --package-path Packages/CrowCLI   # 369 passed (incl. ledger + doc gates)
npm run test:ci  (web-tests)                 # 24 passed
make daemon                                  # clean
```

`CrowDaemonTests` (`MCPRoutesTests`, the `LocalOnlyRPCGateTests` additions) are **written and type-check** but were not executed here — this host has a live Crow install and those tests boot real servers (ADR 0012 / prior data-loss incident). They run in CI.

End-to-end, against a running `crowd`:

```bash
crow mcp token mint --name grok-bot --scope board:read
curl -s localhost:8787/mcp -H 'Authorization: Bearer crow_mcp_…' \
  -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/list","params":{}}'
# -> only list_tickets + list_reviews
curl -si localhost:8787/mcp          # 405
curl -si localhost:8787/mcp -d '{}'  # 401, no token
printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"server/discover","params":{}}' | crow mcp serve
```

## Docs

[ADR 0019](docs/adr/0019-read-only-mcp-server.md) records why the surface is task-shaped and why this does **not** reverse CROW-528 — that decision is about Crow *provisioning* MCP into harnesses, the opposite direction on the same wire. [`docs/mcp.md`](docs/mcp.md) covers client setup for both transports.


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- **Closed allowlist by construction.** `MCPToolCatalog` exposes exactly six tools over five read RPCs with no passthrough; `MCPLedgerExportTests` gates export flags against the ledger, write methods, and `localOnlyRPCMethods` at build time in the Linux CI lane.
- **Two transports, two trust models, correctly separated.** `crow mcp serve` (stdio → Unix socket) needs no token; `POST /mcp` requires a scoped bearer token on every request. `/mcp` is exempt from `WebAuthMiddleware` because that gate is opt-in and weaker on a passwordless daemon — `MCPRoutes` is stricter, and `MCPRoutesTests` pins tokenless refusal even with no web password.
- **Token handling is sound.** 32-byte CSPRNG secrets, SHA-256 with constant-time comparison (appropriate for high-entropy per-request verification), unified error messages that don't distinguish unknown/expired/revoked, and comparison over all records to avoid timing leaks.
- **Scope enforcement is tight.** `tools/list` is filtered by token scopes; out-of-scope `tools/call` is indistinguishable from unknown tools — no enumeration leak.
- **Secrets cannot be minted via web config round-trip.** `SettingsSecrets.preservingSecrets` restores `mcpTokens` from disk (`SettingsSecretsTests.preserveIgnoresBrowserMintedMCPTokens`); token mint/revoke is local-only via `SecretRoutes.gateOK` and `localOnlyRPCMethods`.
- **Request hardening.** Bearer auth and Origin check precede body parsing; 1 MiB body cap; modern-era `Mcp-*` header/body mirroring prevents intermediary routing attacks.
- **Data minimization.** `list_tickets` drops unbounded issue `body`; `get_session` documents that gateway header values stay behind local-only `gateway-get`.

**Concerns:**
- None blocking. The SHA-256-vs-PBKDF2 split and stdio's implicit full-scope trust are explicitly documented and appropriate for their threat models.

### Code Quality
- Layering is clean: protocol core in `CrowIPC`, crypto in `CrowDaemon`, shared validation in `MCPTokenWire`/`MCPTokenRPC`.
- `attention_since` fix in `RPCHandlers.swift` correctly unblocks `list_stuck_sessions` idle detection — a real bug surfaced by the join tool.
- Dual-era MCP support (`initialize` + `server/discover`) with per-request era detection is well-tested.
- **Static analysis:** `swift test --package-path Packages/CrowCore` (639 passed), `Packages/CrowIPC` (120 passed), `Packages/CrowCLI` (369 passed); `swift build` clean. `CrowDaemonTests` (`MCPRoutesTests`) not executed locally per ADR 0012 — CI lane.

### Summary Table

**Green (consider, not blocking):**
- `MCPRoutes.swift` reloads `config.json` on every authenticated request for token lookup — fine for v1 scale; worth watching if MCP traffic grows.
- Off-box MCP inherits whatever network exposure the daemon already has; operators should treat scoped tokens like PATs and prefer narrow scopes + expiry (defaults are sensible: 90d, `--no-expiry` explicit).

---

PR #1007, opened 2026-08-14
Open the terminal context menu on long-press (CROW-1006)

### LINKED ISSUE #1006: Mobile: long-press should show the context menu so copy/paste works
On a mobile device, copy and paste of text does not work. There is no way to select text or get a context menu.

Long-press should show the normal context menu (the mobile equivalent of right-click) so you can copy, paste, and select text in the Crow web UI — especially the terminal / agent transcript.

## Acceptance criteria
- Long-press on text in the Crow web UI (terminal and other selectable surfaces) opens the native context menu.
- Copy and paste work on iOS/Android browsers (or the Crow mobile web view).
- Existing desktop right-click / selection behavior is unchanged.

dgershman:
Closes #1006

## The problem

On a phone there was no way to copy or paste out of the agent transcript. The terminal is an xterm canvas that `xterm.css` marks `user-select: none`, so the browser has **no native selection or copy callout** to offer over it — Crow's own menu (Copy / Paste / Select all / Clear / Reload) *is* the context menu for that surface. Its only entry point was a right-click, and iOS Safari dispatches no `contextmenu` for a long-press on a plain element. So on the surface that needed it most, the menu was simply unreachable.

Non-terminal surfaces (sidebar, board cards, detail header) are ordinary DOM text and already long-press-select natively — nothing there was blocking them, so nothing there changed.

## The fix

Bridge long-press to the same menu on both terminal surfaces:

- **`app.js`** reuses the shared `attachLongPress` (500 ms hold, 10 px movement threshold, single-finger) that sidebar rows already use, pointed at `#terminal-wrap` → `showTerminalMenu`. A drag past the threshold cancels, so the #777 touch-scroll shim keeps the gesture; the trailing emulated click is swallowed so the menu isn't dismissed by the very tap that opened it.
- **`terminal.html`** (the standalone debug page) had the same right-click-only menu, so it gets an equivalent inline bridge. Both of its entry points now build the menu through one `openMenu`, so they can't drift *within* the page the way the two pages once drifted from each other.

`-webkit-touch-callout: none` over each grid keeps iOS from racing the menu with its own long-press sheet. Nothing is given up: there is no native selection there for a callout to act on. Menu rows get a finger-sized hit area under `@media (pointer: coarse)`.

**Android Chrome** fires `contextmenu` for the same press, so both paths run ~500 ms in. That's harmless — `showTerminalMenu` closes any open menu before rebuilding it at the same point, so it reads as one menu.

## Desktop is unchanged

Right-click still goes through the identical `contextmenu` → `showTerminalMenu` path, with the same items at the same cursor position (pinned by a test). The coarse-pointer sizing is scoped to `pointer: coarse`, so a mouse — including on a hybrid touchscreen laptop, which reports `pointer: fine` — sees the old menu exactly.

## Test plan

`Packages/CrowDaemon/web-tests/long-press.test.js` (new, wired into `test` + `test:ci`, so it runs on the PR lane) drives the **real** boot-time registration against a fake xterm, synthesised touch events and a controllable clock — 27 assertions:

- a still 500 ms press opens the menu at the touch point
- the trailing emulated click is swallowed (menu survives the finger lifting)
- a drag past 10 px cancels instead, leaving `touchend` uncancelled
- a sub-threshold wobble still counts; multi-touch (pinch) arms nothing
- `Copy` appears only with a selection
- **the full touch-only `Select all` → `Copy` route** — the thing the ticket is actually asking for, since xterm has no touch drag-select
- the agent-surface force-select hint, the pre-attach `term === null` guard
- the desktop `contextmenu` path producing the same menu at the cursor

Plus three drift guards in `WebTerminalAssetTests` pinning the bridge, the gesture contract (held / single-fingered / still / click-swallowing) and the callout suppression into **both** surfaces — that suite exists precisely to stop these two files diverging.

Verified locally: `npm run test:ci` in `web-tests` is green (16 files), and `swift build --build-tests --package-path Packages/CrowDaemon` compiles and links. I did **not** run CrowDaemon's `swift test` — it boots `crowd` and would kill the live environment on this host; that suite runs on `release.yml`'s macOS lane. The three new assertions are pure string checks, and I verified each one byte-exactly against the shipped assets before committing.

Manual check worth doing on a phone: long-press the terminal → menu appears → **Select all** → long-press again → **Copy**, then long-press → **Paste**. And confirm a one-finger swipe still scrolls the buffer rather than opening the menu.

## Unrelated finding, not fixed here

`WebTerminalAssetTests.stripComments` pairs the `/*` inside `app.js`'s `// … #/settings/*` line comment with a later `*after*`, swallowing ~105 lines (6556–6661). No current assertion breaks — the only whole-file use is negative, so it just passes vacuously over that region — but it means a future negative guard could silently stop guarding. Flagging rather than fixing: a correct JS comment stripper (strings, regexes) is its own change. My new assertions are all positive and deliberately read the raw source, so they're immune.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -408,8 +408,13 @@ body {
 /* #777: `touch-action: none` hands the one-finger gesture to enableTouchScroll
    instead of iOS Safari's overscroll — without it Safari rubber-bands the fixed
    xterm canvas and the swipe never reaches the buffer. Costs pinch-zoom over the
-   terminal (change the font size instead); long-press selection is unaffected. */
-#terminal { width: 100%; height: 100%; touch-action: none; }
+   terminal (change the font size instead).
+   CROW-1006: `-webkit-touch-callout: none` keeps iOS from popping its own
+   long-press sheet over the canvas, so Crow's menu (attached in app.js's boot
+   block) is the only thing a long-press here produces. Nothing is lost — the
+   grid is already `user-select: none` via xterm.css, so there is no native
+   selection to call out; the menu's Select all / Copy is the way out. */
+#terminal { width: 100%; height: 100%; touch-action: none; -webkit-touch-callout: none; }
 
 /* ---- Terminal skeleton loading overlay (#677) ----
    Masks the xterm attach/reload window (blank/garbled grid, reflow flash,
@@ -948,11 +953,16 @@ body.update-banner-visible #back-to-sidebar {
    now a tinted chip inside the PR pill, so its color can distinguish "armed"
    from "Crow gave up". See PR_AUTOMERGE_GLYPH in app.js. */
 
-/* ---- Custom right-click context menu ---- */
+/* ---- Custom right-click / long-press context menu ---- */
+/* CROW-1006: the menu now also opens on touch, where a second long-press landing
+   on it would otherwise raise iOS's text-selection callout over the item labels.
+   These are actions, not copyable text — same treatment terminal.html's #ctxmenu
+   has always had. */
 .ctx-menu {
   position: fixed; z-index: 1000; min-width: 190px; padding: 4px;
   background: var(--bg-card); border: 1px solid var(--border-strong);
   border-radius: 8px; box-shadow: 0 8px 28px rgba(0, 0, 0, 0.5);
+  user-select: none; -webkit-user-select: none; -webkit-touch-callout: none;
 }
 .ctx-item { padding: 6px 10px; border-radius: 5px; font-size: 12px; color: var(--text-primary); cursor: pointer; }
 .ctx-item:hover { background: rgba(221, 196, 130, 0.14); }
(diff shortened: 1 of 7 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Long-press bridges to the existing `showTerminalMenu` / `openMenu` builders — menu labels and actions remain hardcoded; no new injection surface from touch coordinates.
- iOS callout suppression (`-webkit-touch-callout: none`) and `user-select: none` on `.ctx-menu` prevent the platform from surfacing copyable text over action rows; paste still flows through the established clipboard-read path invoked only from explicit menu taps.
- `attachLongPress` reuses the same gesture contract already proven on sidebar rows: single-finger, 10px movement cancel, trailing-click swallow via `preventDefault` on `touchend` (non-passive), keeping accidental xterm click-through from dismissing the menu.
- `showTerminalMenu` still bails when `term` is null, so the pre-attach window cannot surface a menu over an empty grid.

**Concerns:**
- No new security concerns identified. Android’s duplicate `contextmenu` dispatch is harmless because `showTerminalMenu` closes any open menu before rebuilding at the same point.

### Code Quality
- **Green** — `app.js` correctly reuses shared `attachLongPress` on `#terminal-wrap`, pairing with the existing `contextmenu` listener and synthesizing `{ preventDefault, clientX, clientY }` for `showTerminalMenu`. The 10px cancel threshold coexists cleanly with `#777` touch-scroll on the child `#terminal` (verified: scroll drags clear the timer and leave `touchend` uncancelled).
- **Green** — `terminal.html` consolidates right-click and long-press through `openMenu(x, y)` and inlines an equivalent `attachLongPress`; duplication is intentional and pinned by `WebTerminalAssetTests` drift guards.
- **Green** — CSS coarse-pointer sizing scoped to `@media (pointer: coarse)` leaves desktop hit targets unchanged while giving phone users finger-sized rows for the only copy/paste route on touch.
- **Green** — `long-press.test.js` is thorough (27 assertions): timer contract, click swallow, scroll cancel, multi-touch, Copy gating, full Select all → Copy touch path, agent hint, `term === null` guard, and desktop `contextmenu` parity. `npm run test:ci` in `web-tests` passes (16 files). `swift build --package-path Packages/CrowDaemon` succeeds; new `WebTerminalAssetTests` cases pass.
- **Green** — Pre-existing `refreshTerminalsRebindsTheActiveTerminalToTheFreshRow` failure in `WebTerminalAssetTests` is unrelated to this diff (expects legacy `activeTerminal = terminals.find(` shape; this PR does not touch `refreshTerminals`). Not blocking here.
- **Green** — Author-documented `stripComments` hazard over `app.js` is a follow-up for negative guards, not a defect in this change; new assertions are positive and read raw source.

### Summary Table

---

PR #1009, opened 2026-08-14
Kill Cursor duplicate-frame scroll sediment without alt screen

### LINKED ISSUE #1008: Cursor terminal shows duplicate-frame scroll sediment — cursor-agent never enters the alt screen (ADR-0013 gap)
## Summary

Cursor sessions have the **duplicate-frame / scroll-walks-sediment** symptom that ADR-0013 (#822/#824) fixed for Claude Code. Scrolling up in a Cursor terminal shows stacked copies of the agent's TUI instead of letting Cursor own its own scroll.

The per-surface hybrid scroll model is agent-agnostic and **is** being applied to Cursor — the daemon correctly sets `alternate-screen on` for Cursor windows. The problem is that **`cursor-agent` never switches into the alternate screen**, so the setting is inert and every repaint lands in the main-buffer scrollback as sediment. Claude Code's fix relies on the agent opting into the alt buffer; Cursor doesn't, and (unlike Claude Code) has no flag to make it.

## Diagnostic evidence (live tmux, `crow-tmux.sock`, 2026-08-13)

Read straight off the daemon's tmux server — `alternate_on` is the *live* buffer, `alternate-screen` is the *option Crow set*:

```
win  name        alt_on  altopt  mouse_any  history_size   cmd
1    Manager     1       1       1          0              claude   ← Claude Code: correct
2    Cursor      0       1       0          1169           node     ← Cursor: sediment
3    Cursor      0       1       0          187            node
4    Manager 3   0       1       0          1995           node     ← Cursor Manager: sediment
5    Cursor      0       1       0          2339           node
7    Cursor      0       1       0          265            node
```

- **Claude Code** (`win1`): `alt_on=1`, `history_size=0`, mouse-tracking on → living in the alt buffer, no scrollback to fossilize. ADR-0013 works.
- **Cursor** (`win2/3/5/7`, `win4`): `altopt=1` (Crow *did* set `alternate-screen on`) but `alt_on=0` — Cursor renders **inline in the main buffer** and its full-frame repaints accumulate (`history_size` 1169 → 2339 and climbing). That growing history is the #822 frame-sediment.

`cursor-agent` = the `agent` node CLI (`~/.local/bin/agent`, v`2026.08.11-e8db854`).

## Root cause

ADR-0013 defeats frame-sediment by putting agent TUIs in the **alternate screen** (`alternate-screen on` per window), which has no scrollback — repaints can't fossilize. But tmux can only *honor* an app's alt-screen request (`smcup`); it **cannot force** an app that never requests it. Claude Code requests it; **`cursor-agent`'s interactive TUI does not** (it's an inline renderer). So for Cursor:

1. `alternate-screen on` is set but inert → repaints pile into the main buffer's 50k scrollback (the sediment above).
2. Client-side, the daemon derives `agent_surface` from the `alternate-screen` **option** (`#{alternate-screen}` = 1), not the live `alternate_on`, so it reports Cursor as an agent surface — correct intent, wrong reality.
3. Cursor doesn't mouse-track at its prompt (`mouse_any=0`), so `appOwnsScroll()` = `agent_surface && mouseTracking` = `true && false` = **false** → the wheel scrolls xterm's local viewport.
4. That local viewport is the main buffer, now full of repaint sediment → scrolling up walks stacked duplicate frames. **The reported symptom.**

The classifier and `alternate-screen on` are **not** the bug — they're agent-agnostic and correct (`SessionTerminal.isAgentSurface` returns true for Cursor's managed work terminals). The bug is that the model assumes every agent surface *enters* the alt buffer, and Cursor is a counterexample.

## Why Claude Code's fix can't be copied verbatim

Claude Code's fix = "the agent owns its own viewport in the alt buffer." Cursor can't: `agent --help` exposes no fullscreen/alt-screen/TUI-mode flag — only `--print` (headless, non-interactive) and `--output-format`. There is no switch that makes the interactive TUI use `smcup`. So Crow needs a **second mechanism for inline-rendering agent surfaces**, not a copy of the alt-screen path.

## Suggested direction (needs a decision — likely an ADR-0013 amendment / small spike)

Recommend a combination:

1. **Per-agent capability flag on the `CodingAgent` adapter** — e.g. `usesAlternateScreen` (Claude Code = true, Cursor = false). This is the clean home for the alt-screen-vs-inline distinction and fits the ADR-0014 adapter seam. Files: `CodingAgent` protocol (`CrowCore`), `CursorAgent.swift`.
2. **For inline agent surfaces, kill the sediment instead of relying on the alt buffer** — clamp the window's `history-limit` to 0 (or tiny) at creation so xterm-local scroll has no fossil frames to walk (emulates the scrollback-less alt buffer). Evidence supports Cursor being a full-frame repainter (history grows to thousands of lines while the visible frame is ~1 screen), but the spike should confirm Cursor doesn't stream a legit transcript we'd be truncating. Files: `TmuxBackend.registerTerminal` / a sibling of `enableAlternateScreen`.
3. **Close the option-vs-reality gap** — the daemon already reads `#{alternate-screen}` for `agent_surface` in `windowScrollbackClassification` / `listWindowScrollback`; have it also consider live `#{alternate_on}` so "option says agent surface, buffer says main" is detectable and the scroll policy can adapt. Files: `TmuxBackend` classification, `EngineRouter` `list-terminals` (`agent_surface`), `web/app.js` (`appOwnsScroll`).

### Generalizes beyond Cursor
The evidence shows the Cursor **Manager** (`win4`) has the same `alt_on=0` + sediment. Any inline-rendering harness (Cursor, and likely Codex / OpenCode / Grok / Antigravity — unverified) shares this. Whatever mechanism lands should key off the per-agent capability, not hardcode Cursor.

## Repro
1. Start a Cursor work (or Manager) session; let it emit a few frames.
2. In the web terminal, mouse-wheel up.
3. Expected: clean transcript / Cursor owns scroll. Actual: stacked duplicate TUI frames.
4. Confirm underneath: `tmux -S <crow-sock> list-windows -a -F '#{window_name} alt_on=#{alternate_on} altopt=#{alternate-screen} hist=#{history_size}'` → Cursor window shows `alt_on=0 altopt=1 hist>0`.

## References
- ADR-0013 `docs/adr/0013-terminal-scroll-model.md` (this is the amendment target)
- Prior work: spike #822, impl #824, wheel/touch #850, sensitivity #835
- Code: `SessionTerminal.isAgentSurface` (`CrowCore/Models/Terminal.swift`); `TmuxBackend.enableAlternateScreen` / `registerTerminal` / `windowScrollbackClassification`; `EngineRouter` `list-terminals`; `web/app.js` `appOwnsScroll` / `activeSurfaceIsAgent`; `CrowCursor/CursorAgent.swift`

dhilgaertner:
Closes #1008

## Summary
- ADR-0013 assumed every agent TUI would enter the alt buffer once Crow set `alternate-screen on`. `cursor-agent` never issues `smcup`, so that option was inert and full-frame repaints fossilized in the 50k main-buffer history — scrolling up walked stacked duplicate TUI frames.
- Add `CodingAgent.usesAlternateScreen` (Claude Code `true`; protocol default `false` for inline renderers). Inline agent windows are still classified as agent surfaces (`alternate-screen on`) but are born with `history-limit 0`, and the web client caps xterm scrollback to 0 while those tabs are attached.
- Plain shells keep the unified 50k scrollback and CROW-606 replay. Claude Code stays on the alt-buffer path (`alt_on=1`, `hist=0`). Existing Cursor windows cannot have `history-limit` retrofitted (frozen at birth); Recreate rebuilds them, and the client cap is the immediate wheel fix.

## Test plan
- [ ] Start a Cursor work (or Manager) session, let it emit a few frames, mouse-wheel up in the web terminal — no stacked duplicate TUI frames.
- [ ] Confirm live tmux: `tmux -S "$TMPDIR/crow-tmux.sock" list-windows -a -F '#{window_name} alt_on=#{alternate_on} altopt=#{alternate-screen} hist=#{history_size}/#{history_limit}'` — a new Cursor window shows `hist=0/0` (or not climbing) with `altopt=1`.
- [ ] Claude Code Manager/work window still `alt_on=1 hist=0` and scrolls as before.
- [ ] A plain shell still has 50k scrollback and CROW-606 replay after reload.
- [ ] Recreate an *existing* pre-fix Cursor window (50k sediment) and confirm the new one is clamped.


dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- The new tmux option write is argv-based (`TmuxController.run`), not a shell string, and the sandwich only interpolates the constants `0` and `50000`.
- Scroll-model selection is a `CodingAgent` capability (ADR-0014) rather than a Cursor name check, and `AgentRegistry.usesAlternateScreen` reads **known** agents (including off-PATH) so an unavailable harness still gets its declared model.
- No new auth surface, secrets, or user-controlled tmux option names.

**Concerns:**
- None. `setSessionOption` is as trusted as the existing `setWindowOption` path.

### Code Quality
- **Green** — `registerTerminal`'s default `usesAlternateScreen: Bool = true` (`Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:242`) is the opposite of the protocol default `false` (`Packages/CrowCore/Sources/CrowCore/Agent/CodingAgent.swift:253`). Production call sites pass the registry result explicitly, and unknown/`nil` kinds intentionally keep the Claude path so an unattributed agent surface is not clamped. A future caller that omits the flag would silently skip the inline clamp. Worth a one-line comment on the parameter default so it does not look like a copy-paste error.
- **Green** — `applySurfaceScrollback()` is load-bearing on tab switch (`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js:3754`) because `refreshTerminals` is not polled. `WebTerminalAssetTests` pins the call on `refreshTerminals` (`Packages/CrowDaemon/Tests/CrowDaemonTests/WebTerminalAssetTests.swift`) but not on `switchTerminal`. The call is present; a matching shape assertion would lock the path that actually fires when the user clicks a tab.
- **Green** — `applySessionHistoryLimit` is best-effort: if the restore-to-50k `set-option` fails after a successful clamp, later `new-window`s inherit `history-limit 0` until something else resets the session option. MainActor serializes the sandwich (verified by the real-tmux sibling tests), and tmux dying is the realistic failure mode, so this is an accepted availability tradeoff rather than a race.

The rest of the design matches the ADR-0013 amendment: history-limit is frozen at birth (adopt documents that Recreate is the tmux-side heal; the client cap is the immediate wheel fix), `isScrollbackDegraded` treats an agent window at 0 as healthy, Claude stays on the alt-buffer path, and a tab switch still `reloadTerminal`s so shrinking the shared xterm buffer does not steal a shell's CROW-606 replay.

### Summary Table

---

PR #1012, opened 2026-08-14
Flip Codex supportsRemoteControl — the old reason was wrong (CROW-1001)

### LINKED ISSUE #1001: Codex: re-evaluate remote-control for supportsRemoteControl (ADR 0015 re-check)
## What changed upstream

Codex **remote-control** has matured since the gap audit (2026-07-23):

- **Stable 0.147.0** (2026-08-07) ships documented `codex remote-control {start,stop,pair}` + `codex --remote <ws://…>` in the official CLI reference: https://developers.openai.com/codex/cli/reference.md
- Ecosystem reports Codex Remote reached **GA** (2026-06-25) with Noise-encrypted relay, pairing codes, and app-server JSON-RPC — still requires Codex Desktop as bridge for mobile pairing in common setups.
- Crow still sets `supportsRemoteControl = false` (#830 deferred pending E2E validation).

Baseline: experimental/unwired; matrix notes stdin fake unavailable for Codex TUI.

## Matrix / ADR 0015 gap

- Matrix row: **Remote control** (Codex ❌)
- Gap audit §2 #3 / §3b (partial); #830 closure deferred RC flip

**needs-eval.** Native RC is a genuine capability (not `--dangerously-bypass-approvals-and-sandbox`), but wiring `supportsRemoteControl=true` requires validating Crow can drive sessions end-to-end (app-server lifecycle, pairing, tmux integration) without breaking completion contracts. Heavier than `crow send` for Cursor/OpenCode; for Codex it is the only native drive path. Do **not** wire until manual E2E pass confirms badge + `crow send`/RPC story.

1. E2E validation plan: `codex remote-control start` + session attach from a Crow-managed terminal.
2. If viable: flip `OpenAICodexAgent.supportsRemoteControl` and document operational requirements (Desktop bridge, auth).
3. If not viable: refresh matrix pin with concrete blocker.

dgershman:
Closes #1001

## TL;DR

The ticket asked whether native `codex remote-control` had matured enough to flip `supportsRemoteControl`. It hasn't. But while validating that, the **premise for `false` turned out to be wrong**, so the flag flips anyway — for a different, verified reason.

## The recorded reason was wrong

`OpenAICodexAgent.supportsRemoteControl` was `false` because (matrix, ADR 0015, code comment):

> Codex's TUI isn't stdin-drivable the way `crow send` fakes RC for the others.

That is a claim about **Crow's** plumbing, not upstream Codex, and it fails twice.

1. **`crow send` does not write stdin.** `TerminalRouter.send` → `TmuxBackend.sendText` is tmux `load-buffer` → `paste-buffer` → `send-keys Enter` — a paste into the pane. A TUI that ignores stdin is not thereby undrivable.
2. **Codex accepts the paste.** Verified end-to-end in a live tmux pane on `codex-cli 0.141.0`, using Crow's exact sequence including the 0.2 s `pasteEnterSettleDelay`:

```
$ printf 'hello from crow send: …' | tmux load-buffer -b crow-probe -
$ tmux paste-buffer -b crow-probe -t probe:0
› hello from crow send: does the codex composer accept a tmux paste-buffer?     ← composer, verbatim

$ printf '/status' | tmux load-buffer …; tmux paste-buffer …; sleep 0.2; tmux send-keys Enter
│  Model:     gpt-5.5 (reasoning medium, summaries auto)                         ← submitted, executed
│  Session:   019ffe27-5dee-7f61-9274-d6cb03ca942a
```

And nothing gated on the flag: `TerminalRouter.send` is agent-agnostic; `supportsRemoteControl` feeds only `remoteControlActiveTerminals` and the `remote_control_available` payload. **Codex sessions were already remotely drivable — only the badge disagreed.** This fixes a false negative, not a missing capability.

## Native `codex remote-control`: evaluated, declined

Three blockers, any one disqualifying:

1. **Wrong install channel.** `codex remote-control start` on the npm build Crow resolves:
   > Error: managed standalone Codex install not found at `/Users/…/.codex/packages/standalone/current/codex` … requires the standalone install managed by the Codex installer

   Crow resolves `codex` by PATH walk + homebrew / `/usr/local/bin` / `~/.local/bin` fallbacks — all npm-shaped.
2. **Machine-global singleton.** `codex app-server daemon {start,restart,stop}` expose no port, socket-path, or instance flag, and bind one fixed socket `~/.codex/app-server-control/app-server-control.sock`. Crow runs N concurrent per-worktree sessions; none could be addressed individually — the hooks `cwd`-collision defect, relocated to the drive path.
3. **Opposite direction.** `--remote` attaches a *local* TUI to a *remote* app server. Crow already **is** the remote surface (browser → `crowd` → tmux), so this would stand up a second parallel one outside Crow's session model.

**Probed on the installed `0.141.0` while stable was `0.147.0`** — the pin says which blockers age. (3) is architectural; (1) and (2) are version-specific and are what `0.147.0`'s added `pair` subcommand might move. None of it gates the badge, which rests on the paste path.

## Behavioural side effect (a fix)

`/rename` passthrough now reaches Codex **worker** sessions. `agentRenameTargets` routes non-Manager renames through `remoteControlRenameTargets` — which Codex terminals never entered while the flag was `false`. So the matrix's ✅ rename row was true only for its Manager. It only ever targets agent-launched terminals, so no Shell tab receives a `/rename`.

Otherwise: **no launch text changes.** Codex has no `--rc`; a new test pins that `remoteControlEnabled` leaves every `SessionKind` byte-identical, plus the Manager launcher. The Manager's own gate (`" --rc"` in the built command) is untouched — same as Cursor's.

## Also verified

`alternate_on=0` on the live pane, so `usesAlternateScreen = false` is now **confirmed** rather than inherited from the inline default (the matrix said "unverified"). Noted but not acted on: 0.141.0 gates a fresh directory behind *two* prompts — folder trust (`CodexTrustSeeder` handles it) and a newer hook-trust prompt listing Crow's global `~/.codex/hooks.json` entries. Untested against unattended auto-launch; recorded in the gap audit as the next probe if a Codex `.job` ever stalls at startup.

## Docs

Corrected at all three places the wrong claim lived — matrix, gap audit (§2 #3, §3b), ADR 0015 item 3 — plus an ADR amendment on **why the re-check discipline missed this**: it targets claims with an expiry date, and re-probing *upstream* could never have caught a claim about Crow's own send path. The grid now reads "paste" instead of "stdin" for every harness, since that word is what made the cell wrong.

## Test plan

- `swift test` per package: CrowCodex **71**, CrowEngine **749**, CrowCore **641**, CrowCursor **76** — all pass.
- `scripts/check-workspace-custom-instructions.sh` → OK.
- Manual E2E as shown above (isolated tmux socket; probe torn down).
- Not run: `make test` / CrowDaemon `swift test` — they boot `crowd` and would kill the live environment on this host.


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- The change correctly separates **badge semantics** from **drive capability**: `TerminalRouter.send` is agent-agnostic and never consulted `supportsRemoteControl`, so flipping the flag does not expand the remote-send attack surface — it only aligns UI state with behavior that was already reachable from authenticated web clients.
- The new `remoteControlEnabledNeverEmitsAnRCFlag` test in `OpenAICodexAgentTests.swift` guards against a real failure mode: emitting a nonexistent `--rc` flag would make Codex reject its own launch line.
- The `/rename` side effect routes through `Validation.isValidSessionName` (no control characters) and only targets agent-launched terminals via `remoteControlActiveTerminals` / `agentRenameTargets` — same gate Cursor/OpenCode already use.
- Native `codex remote-control` is explicitly documented as declined with three concrete blockers (wrong install channel, singleton daemon socket, reversed direction), so the badge cannot be misread as wiring an unvetted protocol.

**Concerns:**
- None blocking. The hook-trust prompt on fresh Codex directories (noted in the gap audit) is an accepted, documented operational risk outside this PR's scope.

### Code Quality
- **Green:** The investigation is thorough and falsifiable — the PR corrects a premise about Crow's own `paste-buffer` path rather than re-probing upstream Codex, and documents that lesson in ADR 0015.
- **Green:** `supportsRemoteControl = true` is paired with explicit comments in `OpenAICodexAgent.swift` that `remoteControlEnabled` remains inert on every launch path, including Manager (`" --rc"` gate unchanged).
- **Green:** Documentation updates (`docs/agent-harness-matrix.md`, `docs/agent-harness-gap-audit.md`, `docs/adr/0015-harness-capability-tiers.md`) are consistent — matrix cells now say "paste" not "stdin" across harnesses, eliminating the framing that caused the false negative.
- **Green:** The `/rename` passthrough fix for Codex worker sessions is the right behavioral correction; the matrix already claimed ✅ rename support.
- **Green:** `CursorAgent.swift` comment correction (stdin → tmux paste-buffer) keeps cross-harness terminology aligned.
- Tests: `swift test` on CrowCodex (71 tests) and CrowCursor (76 tests) — all pass.

### Summary Table

---

PR #1013, opened 2026-08-14
Migrate OpenCode done-detection to session.status (CROW-1000)

### LINKED ISSUE #1000: OpenCode: migrate crow-hooks from session.idle to session.status (ADR 0015 re-check)
## What changed upstream

OpenCode upstream now treats **`session.idle` as deprecated** and documents **`session.status`** as the canonical idle/busy signal. `session.idle` remains emitted for backward compatibility but plugin handlers are **fire-and-forget** (not awaited).

Evidence:
- Events reference: `session.status` `{ type: "idle" | "busy" | "retry" }`; `session.idle` marked deprecated — https://github.com/uplift-labs/opencode-dev-docs/blob/main/03-events-reference.md
- Plugin manual: "Use `session.status` to detect when the agent finishes work" — https://github.com/joshuadavidthomas/opencode-plugins-manual/blob/main/docs/07-events.md
- OpenCode ≥ **v1.18.16** (2026-08-10); baseline audited **1.17.10+**

Crow still maps `session.idle` → `Stop` in `~/.config/opencode/plugins/crow-hooks.js` (`OpenCodeHookConfigWriter`).

## Matrix / ADR 0015 gap

- Matrix row: **Hook async delivery / timing** — OpenCode ⚠️ (`session.idle` "done" semantics unconfirmed, CROW-545)
- Version-pinned empirical target in `agent-harness-matrix.md`

**needs-eval.** Migrating to `session.status` with `type: "idle"` is wire-worthy if it improves done-detection reliability for interactive TUI sessions. Must empirically compare timing vs current `session.idle` mapping and confirm `busy`/`retry` do not false-trigger `.done`. No brittle flags involved.

1. Update `crow-hooks.js` template in `OpenCodeHookConfigWriter` to handle `session.status` idle transitions (keep `session.idle` fallback for older OpenCode if needed).
2. Adjust `OpenCodeSignalSource` / tests accordingly.
3. Empirical validation on OpenCode ≥ 1.18.x TUI job sessions.
4. Human updates matrix "Last verified" + CROW-545 pin after verification.

dgershman:
Closes #1000

OpenCode deprecated `session.idle` in favor of `session.status`; Crow's `crow-hooks.js` still keyed done-detection off the compat event. This migrates it, and turns CROW-545's open empirical question into a measured, version-pinned answer.

## What upstream actually does

From `packages/opencode/src/session/status.ts` (and `packages/schema/src/session-status-event.ts`, where `Idle` is literally marked `// deprecated`):

```ts
const set = (sessionID, status) => {
  publish(Event.Status, { sessionID, status })
  if (status.type === "idle") publish(Event.Idle, { sessionID })   // ← both fire
}
```

Two consequences drove the implementation — both **measured on opencode 1.18.5**, not assumed:

1. **Both events fire at turn end, status first — 246 µs apart.** Handling both naively emits `Stop` twice per turn. The plugin **latches** on the first `session.status` it sees and ignores `session.idle` from then on. One body serves both eras with no version probe; a build too old to emit `session.status` keeps the fallback.
2. **`session.status` is published on every status write, not only on changes** — the probe recorded *three consecutive* `busy` events in one turn. The plugin tracks the last state per `sessionID` and acts on transitions only, so a turn costs one `UserPromptSubmit` + one `Stop` instead of a subprocess per internal write.

## Why `busy` is mapped too

This is where the user-visible win is. OpenCode gave Crow no "turn started" edge, so a session left `.done` only when its **first tool ran** — a turn that thought for a while, or answered without calling a tool, showed a stale `.done` card. `busy` → `UserPromptSubmit` closes that.

`retry` is folded into `busy`: it's a provider retry *mid-turn*, so it must never read as done, and a `busy→retry→busy` round trip stays one working stretch.

## Empirical validation

**Live end-to-end** — the actual generated plugin, real `opencode run` on 1.18.5, with a logging stand-in for `crow`:

```
SessionStart      t+0.00s
UserPromptSubmit  t+2.74s   ← new edge, 6.65s BEFORE the first tool call
PreToolUse        t+9.39s
PostToolUse       t+9.54s
Stop              t+11.42s
```

Exactly one of each — no double `Stop`, i.e. the latch suppressed the deprecated event. That 6.65 s is precisely the window the card previously spent on a stale `.done`.

The generated plugin was also driven through the **exact captured 1.18.5 event sequence** plus legacy-build, retry-folding, two-concurrent-sessions and multi-turn cases; all produce one `UserPromptSubmit` and one `Stop` per turn.

## Honest limits (recorded in the matrix, not just here)

- The probe was **headless `opencode run`, not the interactive TUI**. The events come off the same server bus, so the mapping carries — but the TUI claim CROW-545 originally asked about is *inferred*, not measured. A TUI job session is the remaining human re-check (ticket item 4).
- Verified on **1.18.5**, not the ≥ 1.18.16 the ticket cites. `session.status` is already present and fully functional at 1.18.5; re-pin after a TUI pass on a newer build.

## Gap found while probing — pre-existing, deliberately not fixed here

A subagent's **child session** publishes its own status, and `session.status` carries no parent link, so it maps onto the one Crow session UUID:

```
PreToolUse        ← parent's task tool starts
SessionStart      ← child session.created
UserPromptSubmit  ← child busy
Stop              ← CHILD idle — parent still working, card reads .done 1.9s early
PostToolUse / Stop ← parent actually finishes
```

The deprecated `session.idle` had the **identical** failure, so this is a standing gap, not a regression from this PR. Closing it means correlating `session.created`'s `info.parentID` and dropping non-root sessions — a distinct behavior change (it re-opens what `SessionStart`/`PreToolUse` should mean for subagents), so it's out of scope here and documented in the matrix. Happy to file a follow-up.

## Files

- `OpenCodeHookConfigWriter` — plugin template: `session.status` handling, latch, per-session transition dedup; `session.idle` demoted to fallback.
- `OpenCodeSignalSource` — no logic change (it already handled `UserPromptSubmit`/`Stop`); doc corrected — it no longer claims the plugin never emits `UserPromptSubmit`, and the `permission.asked` typo is now `permission.ask`.
- `docs/agent-harness-matrix.md` — Hook async delivery bullet rewritten with measured timings; the CROW-545 pin row replaced with a pinned, dated one; child-session gap documented.
- `docs/adr/0015` — OpenCode moved out of "unpinned empirical timings" (Cursor async is now the only one left).

## Test plan

- `swift test --package-path Packages/CrowOpenCode` — 58 tests pass (4 new/updated).
- Behavioral: generated plugin syntax-checked (`node --check`, `bun build`) and driven through the 5 scenarios above.
- Live: real opencode 1.18.5 run, trace above.

### DIFF
--- a/Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeHookConfigWriter.swift
+++ b/Packages/CrowOpenCode/Sources/CrowOpenCode/OpenCodeHookConfigWriter.swift
@@ -42,12 +42,56 @@ import CrowCore
 /// the `@opencode-ai/sdk` `Event` union and the `Hooks` interface in
 /// `@opencode-ai/plugin` (CROW-545 review):
 ///
-///   session.created           → SessionStart     (event bus)
+///   session.created           → SessionStart      (event bus)
 ///   tool.execute.before       → PreToolUse        (hook)
 ///   tool.execute.after        → PostToolUse       (hook)
-///   session.idle              → Stop              (event bus; "agent finished")
+///   session.status {idle}     → Stop              (event bus; "agent finished")
+///   session.status {busy}     → UserPromptSubmit  (event bus; "turn started")
+///   session.status {retry}    → (none — still busy; see below)
+///   session.idle              → Stop              (deprecated fallback only)
 ///   permission.ask            → PermissionRequest (hook — see below)
 ///   session.error             → Notification      (event bus)
+///
+/// **`session.status`, not `session.idle` (CROW-1000).** Upstream deprecated
+/// `session.idle` in favor of `session.status`, whose payload is
+/// `{ sessionID, status: { type: "idle" | "busy" | "retry" } }`
+/// (`packages/schema/src/session-status-event.ts` — `Idle` is literally marked
+/// `// deprecated`). Two properties of the upstream publisher
+/// (`packages/opencode/src/session/status.ts`) drive the plugin's shape, both
+/// confirmed empirically against **opencode 1.18.5** (headless `opencode run`;
+/// the events come off the same server bus the TUI uses, but the *interactive
+/// TUI* pass is still the human re-check in CROW-1000):
+///
+///  1. **Both events fire, status first.** `SessionStatus.set` publishes
+///     `session.status` and then, only when the type is `idle`, `session.idle`
+///     — 246 µs apart in the probe. So a plugin that handles both naively emits
+///     `Stop` twice per turn. The plugin therefore latches: once *any*
+///     `session.status` arrives, this build speaks the modern event and the
+///     deprecated `session.idle` is ignored. Older builds that never emit
+///     `session.status` keep the `session.idle` path, so one plugin body works
+///     across both eras with no version probe.
+///  2. **`session.status` is published on every set, not only on changes** —
+///     the probe recorded three consecutive `busy` events in a single turn. The
+///     plugin tracks the last status per `sessionID` and acts only on
+///     transitions, so one turn costs one `UserPromptSubmit` and one `Stop`
+///     rather than a subprocess per internal status write.
+///
+/// `busy` earns its own mapping: it is the "turn started" edge OpenCode
+/// otherwise never gave us. Before this, a session left `.done` only when its
+/// first tool ran, so a turn that answered without calling a tool (or thought
+/// for a while first) showed a stale `.done` card. `retry` is a *busy*
+/// sub-state (a provider retry mid-turn), so it is normalized to `busy` for
+/// transition bookkeeping and never mapped to `Stop` — treating it as "done"
+/// would park the card on a turn that is still running.
+///
+/// **Known gap (pre-existing, not CROW-1000).** OpenCode's bus is per-*server*
+/// and `session.status` carries no parent link, so a **subagent's** child
+/// session is mapped onto this plugin's one Crow session UUID: the child going
+/// idle emits a `Stop` while the parent is still working (measured 1.9 s early
+/// on 1.18.5). The deprecated `session.idle` behaved identically, so nothing
+/// here regresses it; closing it means correlating `session.created`'s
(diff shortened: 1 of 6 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- `crowPath` and session UUID are embedded via `jsStringLiteral` in `OpenCodeHookConfigWriter.swift`, preventing injection into the generated plugin.
- Hook emissions remain fire-and-forget with swallowed errors, so a failed `crow hook-event` cannot disrupt OpenCode execution.
- Event names and status mappings are hardcoded; no user-controlled strings reach the shell beyond JSON-serialized payloads.
- The pre-existing child-session status leak is explicitly documented as a standing gap (not a regression) in both the plugin comments and `docs/agent-harness-matrix.md`.

**Concerns:**
- None blocking. The subagent child-session `Stop` leak is acknowledged, measured, and unchanged from the deprecated `session.idle` behavior — an accepted follow-up, not a defect in this diff.

### Code Quality
- **Green:** The `session.status` latch (`sawSessionStatus`) correctly prevents double `Stop` when upstream fires both `session.status {idle}` and `session.idle` (~246 µs apart on 1.18.5).
- **Green:** Per-`sessionID` transition dedup (`lastStatus` Map) handles repeated `busy` writes without spamming `crow hook-event`.
- **Green:** `retry` folded into `busy` prevents premature `.done` during provider retries.
- **Green:** `busy` → `UserPromptSubmit` closes the stale-card window before the first tool call; `OpenCodeSignalSource` already handles this edge and gains a targeted test.
- **Green:** Four new/updated Swift tests lock in the generated-plugin contract; `swift test --package-path Packages/CrowOpenCode` — 58/58 pass; build clean.
- **Green:** Docs (`docs/agent-harness-matrix.md`, `docs/adr/0015-harness-capability-tiers.md`) are updated with measured timings and honest limits (headless probe, TUI re-check still open).

### Summary Table

---

Issue #850, opened 2026-07-26
Terminal: mouse wheel & touch should scroll viewport up/down, not navigate prompt history


dgershman:
## Problem

In the Crow terminal / prompts box for agent sessions (Claude, Cursor, etc.), scrolling the **mouse wheel** navigates *prompt/command history* (as if pressing Up/Down arrows) instead of scrolling the terminal viewport. The user expects the wheel to scroll the visible buffer up **and** down like a normal terminal. Same expectation for **touch-drag** on a mobile device / touchscreen.

## Steps to reproduce

1. Open an agent terminal (Claude or Cursor) in Crow.
2. Focus the prompt/terminal.
3. Scroll the mouse wheel up (or down).

**Actual:** the wheel steps back/forward through prompt history (wheel is being translated into arrow-key / history navigation).
**Expected:** the wheel scrolls the terminal viewport (scrollback) — up scrolls up, down scrolls down.

On a mobile device, touch-drag should likewise scroll the viewport up/down rather than triggering history.

## Likely area

- Wheel→key translation config: `terminal.agentWheelNotches` and `terminal.wheelScrollLines` in `.claude/config.json` suggest Crow maps wheel notches to key events for the agent TUI. When that translation reaches the agent's interactive prompt, wheel becomes Up/Down history nav instead of viewport scroll.
- The terminal front-end is hand-mirrored across three surfaces — desktop `terminal.html`, web `terminal.html`, and web `app.js` — so the fix likely needs to land in all three (ideally shared via a custom xterm.js addon, with host-specific transport).
- Mobile: xterm.js touch-scroll handling in the web front-end.

## Acceptance criteria

- Mouse wheel over an agent terminal scrolls the viewport/scrollback **both up and down**, not prompt/command history.
- Touch-drag on a touchscreen scrolls the viewport up/down.
- Consistent across Claude and Cursor terminals and across desktop + web front-ends.
- If wheel-as-arrow-key translation is intentional for some full-screen interactive apps, keep viewport scroll as the default in the normal prompt without breaking that case.

## Open questions / ambiguity

- Confirm whether the current wheel→history behavior is driven by `terminal.agentWheelNotches` (Crow translating wheel to arrows) or by the agent TUI requesting mouse mode — this decides whether the fix is Crow-side or xterm config (`scrollOnUserInput`, app-cursor-key wheel mapping).
[...]

PR #1014, opened 2026-08-14
Restore Cursor wheel scroll by keeping unified history on inline agents

### LINKED ISSUE #1010: Regression (#1009): Cursor mouse wheel doesn't scroll at all — scrollback capped to 0 on a non-mouse-tracking inline agent
## Summary

**Regression from #1009 (closed #1008).** Cursor terminals no longer scroll at all with the mouse wheel. #1009 removed scrollback from "agent surfaces" — server-side (`history-limit 0`) and client-side (xterm `scrollback` capped to 0) — on the premise that an inline agent's main-buffer history is worthless duplicate-frame *sediment*. For Cursor that premise is false: its history is a **clean, legitimate transcript**, and it relies on the terminal's native scrollback to scroll (it neither mouse-tracks nor enters the alt buffer). Removing the scrollback removed Cursor's only scroll path.

This is the mirror image of #1008: we went from "scroll works but (allegedly) shows duplicate frames" to "no scroll at all."

## Symptom

Mouse wheel (and touch drag) does nothing in a Cursor session's terminal. Claude Code sessions still scroll fine.

## Root cause

Two changes in #1009, both gated on `agent_surface` (the `alternate-screen` option read-back) rather than on whether the agent actually owns the wheel / lacks terminal scrollback:

1. **Client cap** — `web/app.js` `applySurfaceScrollback()` (app.js:5351):
   ```js
   const wanted = activeSurfaceIsAgent() ? 0 : UNIFIED_SCROLLBACK;
   if (term.options.scrollback !== wanted) term.options.scrollback = wanted;
   ```
   For any agent surface (Cursor included) the shared xterm buffer keeps **0** scrollback lines — even on existing windows whose tmux server still holds 50k of history, and even though CROW-606 replay streams that history on attach (it's dropped past the viewport).

2. **Server clamp** — `TmuxBackend.registerTerminal` (TmuxBackend.swift:305): `clampHistory = agentSurface && !usesAlternateScreen` → new inline-agent windows are born with `history-limit 0`, so the server keeps no transcript either (and CROW-606 replay has nothing to restore).

Why this kills Cursor's wheel specifically (`appOwnsScroll`, app.js:5420):
- `appOwnsScroll()` = `agent_surface && mouseTracking`. Cursor does **not** mouse-track at its prompt (`mouse_any=0`), so this is `true && false` = **false** → the wheel is **not** forwarded to Cursor; it falls to a **local** xterm-viewport scroll (the #850 path).
- The client cap just set that local viewport's scrollback to **0** → nothing to scroll. Net: the wheel is inert.

Claude Code is unaffected: it mouse-tracks (`mouse_any=1`) and lives in the alt buffer, so `appOwnsScroll()` is `true`, its wheel is forwarded as SGR wheel events, and it scrolls its own transcript. The scrollback cap never matters for it.

## Decisive evidence (live tmux, `crow-tmux.sock`)

A Cursor window's full scrollback is a **single clean transcript**, not stacked frames. Counting distinctive footer/banner elements across the whole history:

```
window 2 (Cursor), 909 non-empty scrollback lines:
  "Run Everything"    → 1
  "Cursor Grok"       → 1
  top banner ▄▄▄▄     → 1
  "→ Add a follow-up" → 1
```

Every live-frame element appears **exactly once** — there is no frame duplication. The 900+ lines are the real transcript (compiles, edits, test runs, git/gh commands). Cursor redraws only its footer, in place. So the "sediment" #1008 set out to delete does not exist for Cursor at its streaming prompt; #1009 discarded real content.

Window state confirming the asymmetry:

```
name       alt_on  altopt  mouse_any  hist
Manager    1       1       1          0/50000    ← Claude Code: owns wheel, scrolls fine
Cursor     0       1       0          1065/50000 ← legit transcript, client-capped → dead wheel
```

(Existing Cursor windows still show `history_limit=50000` because tmux freezes it at window birth — the server clamp only affects windows *created* after the rebuild. The **client cap** is what breaks the wheel on every Cursor tab regardless.)

## The real model gap

ADR-0013 has two categories: unified-scrollback shells, and alt-buffer agent TUIs that own their own viewport (Claude Code). Cursor is a **third**: an inline-streaming agent that, for scroll purposes, behaves like a **plain shell** — its transcript lives in native scrollback and the wheel scrolls that scrollback. It should never have had its scrollback removed. `agent_surface` (does it get `alternate-screen on`) is the wrong axis to gate scrollback removal on; the right axis is `usesAlternateScreen` / "app owns the wheel."

## Suggested fix

1. **Gate the scrollback removal on `usesAlternateScreen`, not `agent_surface`.** Only true alt-buffer agents (Claude Code, `usesAlternateScreen == true`) should get `history-limit 0` (server) and the xterm `scrollback = 0` cap (client). Inline agents (`usesAlternateScreen == false`, i.e. Cursor and every current default) keep the unified 50k scrollback + CROW-606 replay and scroll like a plain shell via the local-viewport path. This is exactly the pre-#1008 behavior, which the evidence shows was correct and artifact-free for Cursor.
   - Server: `TmuxBackend.registerTerminal` clamp (TmuxBackend.swift:305) already tests `!usesAlternateScreen` — the bug is the **client** cap has no equivalent gate.
   - Client: `applySurfaceScrollback()` (app.js:5351) keys off `activeSurfaceIsAgent()` (= `agent_surface`) only. It needs to know `usesAlternateScreen`.
2. **Plumb `usesAlternateScreen` to the client.** Today only `agent_surface` crosses on the `list-terminals` payload (`EngineRouter`). Add a sibling per-terminal flag (e.g. `alt_screen` / `app_owns_scroll`) so the client caps scrollback only for alt-buffer agents. `applySurfaceScrollback` then gates on that, not on `agent_surface`.
3. **Revisit the #1008 premise.** Cursor does not produce duplicate-frame sediment (footer appears once in 909 lines). Consider whether the inline-agent path of #1009 should be largely reverted rather than re-gated — the alt-screen capability abstraction is worth keeping, but the scrollback-removal it drives was solving a non-problem for Cursor.
4. **Heal already-clamped windows.** Any Cursor window *born* after the #1009 rebuild has `history-limit 0` server-side and has permanently lost its transcript (frozen at birth) — only Recreate rebuilds it with the corrected limit. Fixing the client cap restores the wheel immediately on the pre-existing 50k windows without a restart.

## Repro
1. Open a Cursor session terminal in the web UI; let it emit a few frames.
2. Mouse-wheel up. Actual: nothing moves. Expected: scroll back through the transcript.
3. Underneath: `tmux -S "$TMPDIR/crow-tmux.sock" capture-pane -t <cursor-win> -p -S -1000 | rg -c "Run Everything"` → `1` (one clean transcript, no sediment), while the client shows 0 scrollback.

## References
- Regressed in #1009 (closed #1008); amends ADR-0013 `docs/adr/0013-terminal-scroll-model.md`.
- Code: `web/app.js` `applySurfaceScrollback` (app.js:5351) + `appOwnsScroll` (app.js:5420); `TmuxBackend.registerTerminal` clamp (TmuxBackend.swift:305); `CodingAgent.usesAlternateScreen`; `SessionTerminal.isAgentSurface` (`CrowCore/Models/Terminal.swift:132`); `EngineRouter` `list-terminals` (`agent_surface`).
- Prior art: ADR-0013, #822/#824, #850 (wheel at non-mouse-tracking prompt scrolls local viewport).

dhilgaertner:
Closes #1010

## Summary
- Retract the #1009 inline-agent `history-limit 0` clamp. Cursor's transcript is a clean native scrollback, not duplicate-frame sediment, and the wheel at a non-mouse-tracking prompt scrolls that local viewport (#850). Capping it to 0 made the wheel a no-op.
- Plumb `uses_alternate_screen` on `list-terminals`. `applySurfaceScrollback` now caps xterm scrollback to 0 only for true alt-buffer agents (`agent_surface && uses_alternate_screen`); inline agents keep the unified 50k.
- Leftover windows born under the clamp fail the history floor and are badged ⚠ Recreate. ADR-0013 records a third surface class (inline-streaming agents).

## Test plan
- [ ] Open a Cursor session terminal, let it emit a few frames, mouse-wheel up — expected: scroll back through the transcript (not a dead wheel).
- [ ] Claude Code session still scrolls (alt buffer, wheel forwarded while mouse-tracking).
- [ ] A plain shell tab still has the unified 50k + CROW-606 replay.
- [ ] A Cursor window born after #1009 at `history-limit 0` shows ⚠ Recreate; Recreate rebuilds it at 50k.
- [ ] `swift test --package-path Packages/CrowTerminal --filter ScrollbackHealthTests --filter inlineAgentSurfaceKeepsUnifiedHistory --filter sessionHistoryLimitSandwich`
- [ ] `node Packages/CrowDaemon/web-tests/wheel-scroll.test.js` — inline agent keeps `UNIFIED_SCROLLBACK`


dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- `uses_alternate_screen` is a daemon-derived boolean from `AgentRegistry`, not client-supplied input. `app.js` only uses it to choose an xterm scrollback cap.
- A missing flag (older daemon) fails open to `UNIFIED_SCROLLBACK` rather than stripping history — the failure mode that caused #1010.
- No new auth, command construction, or secret handling.

**Concerns:**
- None. Scrollback limits are a UX/correctness axis, not an access-control one.

### Code Quality
- The three-class split is consistent: tmux no longer clamps inline agents to `history-limit 0`; `list-terminals` emits `uses_alternate_screen`; `applySurfaceScrollback` in `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` caps xterm only when `agent_surface && uses_alternate_screen`. A plain shell in a Claude session stays at 50k because `agent_surface` is false.
- Leftover CROW-1008 windows at `history-limit 0` now fail the floor and badge ⚠ Recreate (`Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift` `isScrollbackDegraded`; `Packages/CrowTerminal/Tests/CrowTerminalTests/ScrollbackHealthTests.swift`). That is the right heal — the limit is frozen at birth.
- Targeted tests passed: `AgentRegistryScrollCapabilityTests`, `ScrollbackHealthTests`, `inlineAgentSurfaceKeepsUnifiedHistoryWithoutAffectingSiblings`, `sessionHistoryLimitSandwichClampsOnlyTheNewWindow`, `WebTerminalAssetTests`, and `wheel-scroll.test.js` (56/56, including the new alt-buffer vs inline vs older-daemon truth table).

**Green**
- `Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:245` — `usesAlternateScreen` is now discarded (`_`) with a leftover default of `true`, while `CodingAgent.usesAlternateScreen` defaults to `false`. Harmless today because `list-terminals` reads `AgentRegistry` independently (`Packages/CrowEngine/Sources/CrowEngine/EngineRouter.swift:953-954`). Consider dropping the parameter so a future clamp cannot silently ignore a caller-supplied value.
- No RPC-level assertion that `list-terminals` actually emits `uses_alternate_screen`. Client tests inject the field; a dropped key would be silent on the daemon side.
- Unverified harnesses inherit the inline 50k path (documented in `docs/adr/0013-terminal-scroll-model.md` and `docs/agent-harness-matrix.md`). If one later proves to be a full-frame TUI, frame sediment returns until it opts into `usesAlternateScreen = true`. Accepted risk.

### Summary Table

---

PR #1015, opened 2026-08-14
Emit Codex async hooks behind a version gate (CROW-999)

### LINKED ISSUE #999: Codex: wire async hooks when 0.148+ stable (ADR 0015 re-check)
## What changed upstream

Codex **async command hooks** for state-machine events (`PostToolUse`, `PreToolUse`, etc.) land in upstream **0.148.0-alpha** / `main`; they remain **skipped on latest stable 0.147.0**.

- **0.147.0 (stable, 2026-08-07):** `discovery.rs` still skips non-`SessionEnd` hooks with `async: true` (`"async hooks are not supported yet"`). Pin from matrix/gap audit **still holds on stable**.
- **0.148.0-alpha.9+ / `main`:** `discovery.rs` now sets `runs_async = async && event != SessionEnd` and passes it to the handler — async hooks run in the background for all events except `SessionEnd` (forced sync with warning).
- **Official docs** now document `"async": true` for command hooks: https://learn.chatgpt.com/docs/hooks (SessionEnd exception documented).

Evidence:
- Stable skip: https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/hooks/src/engine/discovery.rs#L476-L482
- Alpha/main async: https://github.com/openai/codex/blob/main/codex-rs/hooks/src/engine/discovery.rs#L500-L544
- Hooks docs: https://learn.chatgpt.com/docs/hooks

**Upstream min version:** ≥ **0.148.0** (currently alpha; re-probe when stable ships).

## Matrix / ADR 0015 gap

- Matrix row: **Hook async delivery** (Codex ❌ sync-only)
- ADR 0015 gap **#4** (Codex hooks sync-only; `CodexHookConfigWriter.asyncEvents` empty)
- Version-pinned re-check target in `agent-harness-matrix.md`

**needs-eval → wire-worthy once 0.148 is stable.** Async hooks are an honest capability (upstream documents them; no full-bypass). Crow should gate on min Codex version and populate `asyncEvents` for the same events Claude uses (`PostToolUse*`), but must re-verify against daemon apply-order caveat **#903** (fire-and-forget `hook-event` + async delivery may widen the PreToolUse/PermissionRequest inversion window).

**Not brittle-reject:** does not require `--dangerously-bypass-approvals-and-sandbox` or global config mutation.

1. Version-probe Codex ≥ 0.148 before emitting `async: true` in hook config.
2. Populate `CodexHookConfigWriter.asyncEvents` for appropriate events.
3. Empirical pass: state card correctness for `.job`/`.work` sessions; no double-count with notify bridge.
4. Update matrix pin + ADR 0015 when verified (separate doc PR).

dgershman:
Closes #999

Codex honors `async: true` on command hooks from **0.148.0**. Through `0.147.0` its hook engine parsed the key and then **skipped the entry** — so an async hook wasn't slower, it was *absent*, and since Crow's session-state detection runs on those hooks, emitting one would have silently frozen the state card. That is why `CodexHookConfigWriter.asyncEvents` had been empty since v0.139.0, and why this is a **gate**, not a flag flip.

## What landed

**`CodexVersionProbe` (new, CrowCodex).** Runs `codex --version` once at boot — hard-bounded by the same resume-once race `BinaryIdentityProbe` uses, so a `codex` that hangs (or forks a child holding the stdout pipe) can't stall startup — parses the banner, and clears the gate only at `>= 0.148.0`.

- **Fail-closed.** No Codex registered, no binary resolved, spawn failure, timeout, unreadable banner → sync. Sync is correct on every Codex build and costs only latency; a wrong `async` costs the hook.
- **Pre-releases are rejected.** Async landed at `0.148.0-alpha.9`, so `0.148.0-alpha.1` is a build that would still drop the hooks. Standard semver precedence (a pre-release sorts below its release) rejects the whole alpha series with no special case, and starts passing on its own when the stable ships.

**`CodexHookConfigWriter`.** `asyncEvents` is now `["PostToolUse"]`, emitted only when the probe says so. Membership mirrors `ClaudeHookConfigWriter` minus the events Codex doesn't emit: `PreToolUse` stays sync so it is *accepted* ahead of the `PermissionRequest` that follows it, and making it async would widen precisely the inversion that does **not** self-heal (#903 — a `PreToolUse` applied after `PermissionRequest` clears the permission badge while the agent is parked at a prompt, with no further event until the user answers).

**Downgrade is handled.** Each event's entry is rebuilt whole on every boot and overwritten in `installGlobalConfig`, so a machine that stops clearing the gate has the `async` key **removed**, not stranded on a build that would skip the hook. Covered by a test.

**`AgentSemVer` (new, CrowCore).** Semver parsing hoisted out of `OpenCodeLaunchArgs` so one comparison serves both probes. Adds pre-release precedence (which OpenCode's window check never needed) and **whole-token** matching, so a `--version` preamble naming a path like `.../node/v22.22.2/bin/codex` can't be misread as the agent's version — a real shape, since the local `codex` prints a PATH-aliasing warning before its banner. OpenCode keeps its lenient regex scan and feeds the extracted triple to the shared parser; its behavior is byte-identical and its tests are untouched.

**Boot wiring.** The probe runs in `CrowDaemon.run` between `registerAgents` and `LaunchScaffold.run`, gated on the same `AgentRegistry.shared.agent(for: .codex)` expression that already gates Codex scaffolding — so if the config gets written, the probe ran. `LaunchScaffold.run` takes the verdict as a **required** parameter (no default: a silent `false` would quietly drop async on a capable Codex). One log line names the detected version and the pin, so a stale gate is diagnosable without a rebuild.

## Test plan

Per-package (not `make test` — CrowDaemon's suite boots `crowd` and would kill the live environment):

- `swift test --package-path Packages/CrowCodex` → **83 passed** (was 70; +13)
- `swift test --package-path Packages/CrowCore` → **655 passed**
- `swift test --package-path Packages/CrowOpenCode` → **54 passed** (unchanged — migration is behavior-preserving)
- `swift build` → clean
- `make parity` → notification catalogs, CLI/RPC parity (101 methods), no-NSLog all pass

New coverage: gate boundary (`0.147.0` sync / `0.148.0` async / `0.148.0-alpha.9` sync), fail-closed paths (empty output, unparseable, runner throws, non-zero exit output *kept*), whole-token rejection of path-shaped tokens, pre-release ordering, `PostToolUse`-only async, and async stripped on downgrade.

## Not verified — the one open item

**The async *timing* has no empirical pass** (ticket scope item 3: state-card correctness for `.job`/`.work`, no double-count against the `notify` bridge). It can't be run yet: `0.148.0` is still pre-release. npm `dist-tags` on 2026-08-13:

```
latest: 0.147.0
alpha:  0.148.0-alpha.14
```

So this ships **inert on every stable install** — the gate rejects `0.148.0-alpha.14`, and the enabled branch has unit coverage but no observation. It begins emitting `async` on its own when the release lands, and that first stable build is where the timing pass belongs. The docs say exactly this rather than claiming the pass.

## Docs

Ticket scope item 4 defers the matrix/ADR update to a separate PR "when verified". Leaving the docs reading *"❌ sync-only (v0.141.0)"* while the code emits `async` would be exactly the drift this repo flags, so they're updated to describe **what the code does** and marked ⚠️ (gated, timing unverified) rather than ✅ — no drift, no overclaiming. `agent-harness-matrix.md` (row, per-dimension note, re-check pin), `agent-harness-gap-audit.md` (§2 row 4, §3b, §4, plus a dated re-check banner — the 2026-07-23 version snapshot is left as recorded rather than rewritten), `adr/0015` (gap 4, re-check list, alternatives), and a CHANGELOG entry.


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Fail-closed version gating is the right posture: spawn failure, timeout, empty output, or an unparseable banner all land on sync-only hook registration, which every Codex build handles. Emitting `async: true` on a pre-0.148 build would silently drop hooks and break session-state detection.
- `AgentSemVer.parse(token:)` uses whole-token matching via `firstToken(in:)`, preventing a path-shaped preamble (e.g. `v22.22.2` inside an NVM path) from being misread as the agent version and incorrectly clearing the gate.
- Pre-releases are rejected below their release (`0.148.0-alpha.*` < `0.148.0`), conservatively handling the fact that async landed mid-alpha.
- Hook commands are built only from the static `allEvents` list; no user-controlled input reaches command construction.
- `LaunchScaffold.run` requires an explicit `codexAsyncHooksSupported` parameter (no silent default), and `installGlobalConfig` defaults to `false` for callers that skip the probe.
- Downgrade path rebuilds each hook entry whole, stripping a stale `async` key rather than leaving it on a build that would skip the hook.

**Concerns:**
- None blocking. The subprocess probe reuses `BinaryIdentityProbe`'s bounded timeout, which is appropriate for the boot path.

### Code Quality
- `AgentSemVer` cleanly centralizes semver comparison for both Codex and OpenCode; OpenCode's migration is behavior-preserving and its suite still passes.
- Boot wiring order is correct: probe runs after `registerAgents` and before `LaunchScaffold.run`, with a diagnostic log line naming the detected version and pin.
- `PostToolUse`-only async (keeping `PreToolUse` sync) is well-motivated by the #903 permission-badge ordering hazard.
- Test coverage is thorough: gate boundaries (`0.147.0`/`0.148.0`/alpha), fail-closed paths, downgrade stripping, whole-token rejection, and subprocess integration via `StubShellRunner`.
- Static analysis: `swift build` clean; `swift test --package-path Packages/CrowCodex` (83 passed), `Packages/CrowCore --filter AgentSemVer` (14 passed), `Packages/CrowOpenCode` (54 passed), `Packages/CrowDaemon --filter LaunchScaffold` (4 passed).

**Green (documented, not defects in this diff):**
- Async hook *timing* correctness (state card, notify bridge) is explicitly deferred until stable `0.148.0` ships; the gate keeps this inert on current stable installs — an accepted scope boundary, not a code bug.
- Pre-release identifier comparison uses opaque string ordering rather than full semver dot rules; the PR documents why this is sufficient for a minimum-version gate.
- Probe verdict is cached for the daemon lifetime; upgrading `codex` under a running `crowd` requires restart (same as agent availability).

### Summary Table

---

PR #1019, opened 2026-08-14
Restore the visible scrollbar CROW-593 hid (CROW-1016)

### LINKED ISSUE #1016: Restore visible scrollbars hidden in the web/crowd UI
The old Crow desktop app had a visible scrollbar. After the rewrite / move to `crowd` + the web UI, it is gone.

This is not “scroll is broken” — wheel/trackpad still work. The thumb was deliberately removed for web-UI parity (CROW-593):

```css
/* Scroll, but no visible scrollbar. */
#board, #sidebar { scrollbar-width: none; -ms-overflow-style: none; }
#board::-webkit-scrollbar, #sidebar::-webkit-scrollbar { width: 0; height: 0; display: none; }
```

(`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css`)

That hide is what dropped the gutter users used to grab. Restore a visible scrollbar on the surfaces that used to have one (sidebar session list, ticket/reviews board, and any other pane that scrolled with a thumb in the old app). Desktop overlay-on-hover is fine; a permanently invisible bar is not.

## Acceptance criteria
- Sidebar and board show a usable vertical scrollbar when content overflows.
- Existing wheel / trackpad / touch scroll still works.
- Settings (already `overflow-y: auto` with no hide rule) stays visible if it overflows.
- Do not hide the bar again for “parity” with overlay-only chrome.

dgershman:
Closes #1016

## What

CROW-593 hid the scrollbar on `#board` and `#sidebar` for "web-UI parity" with overlay chrome:

```css
/* Scroll, but no visible scrollbar. */
#board, #sidebar { scrollbar-width: none; -ms-overflow-style: none; }
#board::-webkit-scrollbar, #sidebar::-webkit-scrollbar { width: 0; height: 0; display: none; }
```

Scroll itself kept working, so nothing ever failed — but the gutter people grabbed in the desktop app was gone, along with any hint that a pane has more content below. This deletes that rule and styles the bar instead.

## Why it takes two mechanisms

Neither reaches the other engine, so both are stated:

| Engine | Reads | Ignores |
|---|---|---|
| Firefox | `scrollbar-width` / `scrollbar-color` | `::-webkit-*` |
| WebKit / Blink | `::-webkit-scrollbar*` | `scrollbar-color` |

The explicit `::-webkit-scrollbar` **width** is the load-bearing part on WebKit: it opts the element out of macOS overlay scrollbars, so the thumb is drawn *before* the wheel moves. Dropping it silently returns the bar to "visible only once you already scrolled" — the thing the ticket rejects.

## Scoping and contrast

- **Gated to `(hover: hover) and (pointer: fine)`.** A permanent 10px gutter is real width off a 350px sidebar, and this UI is driven from a phone as much as a desktop. Touch platforms already draw an overlay thumb during a flick and — with the hide rule gone — are now free to. Desktop gets a persistent grabbable bar, touch keeps the native one, neither gets nothing.
- **Thumb is gold at 0.50 alpha**, the lowest that clears WCAG 2.2 §1.4.11's 3:1 non-text contrast on every surface it can sit on (3.3:1 on `--bg-card`, 3.6:1 on `--bg-deep`). The 0.30 I first tried measured 2.0:1 — the same invisibility by another name. Painted thumb is inset with a transparent border + `background-clip: content-box`, so the *hit* area stays the full 10px track.
- Applied to the sidebar and board (the panes the ticket names) plus the settings body and the two popover lists, so the app has one scrollbar policy rather than a per-pane accident.

The selector list is spelled out per rule rather than factored into one `:is(…)`. Deliberate: custom-scrollbar pseudo-elements resolve on a narrow path in WebKit/Blink, an unsupported selector form drops the whole rule silently, and that failure mode is exactly the bug being fixed. Repetition is the cheaper risk.

## Acceptance criteria

| Criterion | Status |
|---|---|
| Sidebar + board show a usable vertical scrollbar on overflow | ✅ hide rule deleted, both panes styled |
| Wheel / trackpad / touch scroll still works | ✅ no `overflow` values changed; only bar *appearance* |
| Settings stays visible if it overflows | ✅ `.settings-body` never had a hide rule, and now matches the theme |
| Not hidden again for "parity" | ✅ pinned by test (below) |

## Tests

New `WebTerminalAssetTests.sidebarAndBoardKeepAVisibleScrollbar` pins both halves — the hide must stay gone from **both** stylesheets, and each engine's mechanism, the two tokens, and the two named panes must stay present. Declaration pins are scoped to their own rule via a new `ruleBody` helper, so they catch the width migrating off `::-webkit-scrollbar` without also freezing the selector's indentation. Pane coverage is checked by scanning the scrollbar selectors rather than pinning the whole list, so adding a sixth pane later doesn't turn this red.

- `swift build --build-tests` for CrowDaemon: **Build complete** (compiles clean; not run here — `swift test` boots `crowd` on this host)
- `npm run test:ci` in `web-tests`: **24 passed, 0 failed**
- app.css re-parsed through jsdom's CSSOM: all five rules land with their declarations intact, and no `scrollbar-width: none` remains
- Contrast ratios computed against all three surface tokens (see table above)

**Not verified here:** the rendered bar itself. Headless Chrome would not start in this sandbox, so the visual check — thumb visible in sidebar/board/settings on desktop, native overlay preserved on a phone — is the one step left for review.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -18,6 +18,16 @@
   --blue: #0A84FF;
   --purple: #BF5AF2;
   --font-ui: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif;
+  /* Scrollbar (CROW-1016). Gold so the thumb reads as chrome, not content, at
+     the lowest alpha that still clears WCAG 2.2 §1.4.11's 3:1 non-text contrast
+     against every surface it can sit on — 0.50 gives 3.3:1 (--bg-card) to
+     3.6:1 (--bg-deep), where 0.30 gave a barely-there 2.0:1, which is the
+     invisibility this ticket exists to fix. Dimming this below ~0.45 re-hides
+     the bar by another name. --scroll-gutter is the track width AND the grab
+     target. */
+  --scroll-thumb: rgba(221, 196, 130, 0.50);
+  --scroll-thumb-hover: rgba(221, 196, 130, 0.72);
+  --scroll-gutter: 10px;
 }
 
 * { box-sizing: border-box; }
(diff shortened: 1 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Pure presentation-layer CSS with no user-controlled input, credential handling, or server-side surface area — no injection or auth concerns introduced.
- Scrollbar tokens (`--scroll-thumb`, `--scroll-gutter`) are static design values, not derived from runtime data.

**Concerns:**
- None identified.

### Code Quality
- The CROW-593 hide rule is cleanly removed and replaced with a deliberate dual-engine strategy: `scrollbar-width`/`scrollbar-color` for Firefox and `::-webkit-scrollbar*` pseudo-elements for WebKit/Blink, each documented with the rationale for why both are required (`app.css`).
- Gating under `@media (hover: hover) and (pointer: fine)` is a sound trade-off — persistent gutter on desktop, native overlay behavior preserved on touch — and is well explained in the PR body.
- Explicit per-selector repetition instead of `:is(…)` is justified given WebKit's silent-failure mode for unsupported selector forms on scrollbar pseudo-elements.
- `WebTerminalAssetTests.sidebarAndBoardKeepAVisibleScrollbar` is a strong regression guard: `stripComments` avoids false passes from prose, `ruleBody` scopes width/thumb assertions to the correct rule, and negative checks block reintroduction of `scrollbar-width: none` in either `app.css` or `settings.css`.
- `swift build --build-tests` compiles clean; `sidebarAndBoardKeepAVisibleScrollbar` passes locally.

- Visual verification of the rendered thumb on desktop sidebar/board/settings remains manual (noted in the PR) — acceptable for a CSS-only change where static analysis and token contrast math are the practical checks in CI.
- The test pins `#sidebar` and `#board` coverage but intentionally does not freeze the full selector list, so `.settings-body`/`.notif-list`/`.wizard-list` join coverage is documentation-driven rather than asserted — consistent with the forward-compat goal stated in the PR.

### Summary Table

---

PR #1021, opened 2026-08-14
Move the restored scrollbar off the sidebar and onto the xterm (CROW-1020)

### LINKED ISSUE #1020: Scrollbar landed on the sidebar; restore it on the xterm, not the left rail
#1016 / PR #1019 restored a visible scrollbar on the **left sidebar** (and board). That was the wrong surface.

What actually disappeared — and what people used to grab — is the **xterm** scrollbar: scroll back terminal history. The sidebar was never supposed to grow a permanent gutter.

#1016’s acceptance criteria named `#sidebar` / `#board` because those were the panes with the CROW-593 hide rule. The report was “the scrollbar is gone”; the intended one was the terminal history thumb, not the session list.

## Do this

1. **Revert the sidebar/board visible-bar policy** from #1019. Put back the hide on `#sidebar` (and `#board` unless a board thumb is independently useful). Update `sidebarAndBoardKeepAVisibleScrollbar` so it no longer pins the wrong chrome.
2. **Restore a visible, grabbable scrollbar on the xterm** so you can scroll back history (shell surfaces with xterm scrollback; the thumb that used to sit on the terminal).
3. Do not trade one for the other — sidebar stays overlay/hidden; terminal gets the bar.

## Acceptance criteria
- Left sidebar has **no** persistent scrollbar gutter.
- xterm shows a usable vertical scrollbar when there is history to scroll.
- Wheel / trackpad / existing hydrate-on-top (CROW-934) still work.
- #1019’s test is rewritten to match this (sidebar hidden, terminal visible), not deleted without a pin.

Related: #1016 (wrong pane), #1019 (implemented that), #934 (shell scrollback depth).

dgershman:
Closes #1020

## What went wrong in #1019

#1016 named `#sidebar` / `#board` in its acceptance criteria because those were the panes carrying CROW-593's hide rule. But the report was "the scrollbar is gone", and the bar people actually reach for is the **terminal's history thumb**. #1019 gave the sidebar a permanent styled gutter and left the terminal exactly as invisible as it was.

This puts the hide back and fixes the pane that was meant.

## Sidebar / board: reverted

The CROW-593 rule goes back verbatim:

```css
#board, #sidebar { scrollbar-width: none; -ms-overflow-style: none; }
#board::-webkit-scrollbar, #sidebar::-webkit-scrollbar { width: 0; height: 0; display: none; }
```

#1019's block also reached `.settings-body`, `.notif-list` and `.wizard-list` — panes that never had a hide rule and weren't in scope for either ticket. It comes out whole rather than leaving half a policy behind. `--scroll-gutter` went with it (no consumer left).

## Terminal: two independent bugs, both fixed

The bar was missing for two reasons, and fixing either alone leaves you with an invisible bar or a disappearing one.

**1. Colour — app.js.** xterm paints its slider from the JS `theme` object, not from CSS. Left alone it derives the slider from the *foreground* at 0.20 alpha:

```js
scrollbarSliderBackground: color.opacity(foreground, .2)   // xterm.js 6.0.0
```

`#d4d4d4` at 0.20 over `#1e1e1e` is **1.67:1** — under WCAG 2.2 §1.4.11's 3:1 floor. That is why it read as "gone" while technically being drawn. `scrollbarTheme()` reads app.css's `--scroll-*` tokens back out and hands them over, so the tokens keep one definition and gain a second consumer rather than the gold being written down twice. A missing token yields an *omitted key* (xterm's own default), never an empty string — `css.toColor` throws on that, at Terminal construction.

| Slider state | Token | Contrast on `#1e1e1e` |
|---|---|---|
| xterm default | fg @ 0.20 | 1.67:1 ❌ |
| resting | `--scroll-thumb` (gold @ 0.50) | **3.46:1** ✅ |
| hover | `--scroll-thumb-hover` (0.72) | 5.70:1 ✅ |
| drag | `--scroll-thumb-active` (0.90) | 8.14:1 ✅ |

0.50 is the floor, not a preference: 0.45 measures 3.06:1 and 0.44 measures 2.98:1.

**2. Persistence — app.css.** xterm 6 scrolls through a VS Code scrollable element constructed with `vertical: ScrollbarVisibility.Auto`, so the thumb fades out whenever the pointer isn't over the grid. The override pins **both** declarations xterm.css's `.invisible` sets:

```css
#terminal-wrap.has-scrollback .xterm-scrollable-element > .scrollbar.vertical {
  opacity: 1;
  pointer-events: auto;
}
```

`pointer-events` is not incidental — a thumb you can see and cannot grab is the same complaint one step later.

## Why `.has-scrollback` is load-bearing

When the bar isn't needed, xterm's `ScrollbarState` sizes the slider to the **whole** track. Pinning it unconditionally would draw a permanent gold stripe down the right of every alt-buffer agent session. The class is set from `buffer.active.baseY` — the same quantity as xterm's own "is the bar needed" test. That covers alt-buffer agents (Claude Code) for free, though **not** via the buffer type: crow-tmux.conf strips `smcup`/`rmcup` toward the client, so the web xterm is permanently in its main buffer and `buffer.active.type` is never `'alternate'` there (the trap ADR-0013 calls out for the wheel routing). What zeroes `baseY` on those surfaces is `applySurfaceScrollback` pinning `options.scrollback = 0`, capping the main buffer at `rows`. Updated on `onScroll` + `onRender`, the same pair the #668 jump-to-bottom pill uses, for the same reason.

## Why only the persistence half is gated to a fine pointer

Not for width. FitAddon already subtracts the 14px scrollbar from the column count on *every* device whenever `scrollback !== 0`, so the gutter is paid for regardless — unlike the sidebar, showing the bar here costs no layout.

It's for hit-testing: restoring `pointer-events` on touch would let a flick starting in that gutter drive the slider **and** `enableTouchScroll`'s one-finger shim from the same gesture. Touch keeps xterm's own reveal-on-scroll, which is now actually visible thanks to the colour fix.

## Acceptance criteria

| Criterion | Status |
|---|---|
| Left sidebar has no persistent scrollbar gutter | ✅ CROW-593 hide restored verbatim |
| xterm shows a usable vertical scrollbar when there is history | ✅ themed + pinned, gated on `baseY > 0` |
| Wheel / trackpad / hydrate-on-top (CROW-934) still work | ✅ no scroll handler or `overflow` value changed — appearance and hit-testing only |
| #1019's test rewritten, not deleted without a pin | ✅ see below |

## Tests

`sidebarAndBoardKeepAVisibleScrollbar` → **`sidebarStaysOverlayAndTerminalGetsAVisibleScrollbar`**, same file. Both halves in one test on purpose: they're one decision and must not drift apart. It pins the hide (scoped to its own rule body, so it fails if the hide migrates to another pane), both terminal declarations, the `.has-scrollback` gate still being asked for, the token→theme-key handoff, and that the three tokens exist.

New **`web-tests/terminal-scrollbar.test.js`** drives the real `updateTerminalScrollbar` and `scrollbarTheme` against jsdom — the gate in both directions, teardown paths, and the token mapping with three distinct values so a transposition can't pass.

Both were mutation-checked:
- forcing the class on → **4 failures**
- dropping the empty-token guard → **2 failures**
- swapping `--scroll-thumb-hover` for `--scroll-thumb-active` → **1 failure**

- `swift build --build-tests` (CrowDaemon): **Build complete** — not run here, `swift test` boots `crowd` on this host
- `npm run test:ci` (web-tests): **exit 0, 0 failures** across all 16 files, including the new one (14 assertions)
- The Swift pins were additionally replayed against the real assets with the test's own `stripComments`/`ruleBody` logic: all 11 hold
- Contrast figures above computed from the WCAG relative-luminance formula

**Not verified here:** the rendered bar. Headless Chrome won't start in this sandbox, so the visual check is the one step left for review — thumb visible and draggable in a shell tab with history, absent in an empty tab and in a Claude Code session, sidebar showing no gutter.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -18,16 +18,21 @@
   --blue: #0A84FF;
   --purple: #BF5AF2;
   --font-ui: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif;
-  /* Scrollbar (CROW-1016). Gold so the thumb reads as chrome, not content, at
-     the lowest alpha that still clears WCAG 2.2 §1.4.11's 3:1 non-text contrast
-     against every surface it can sit on — 0.50 gives 3.3:1 (--bg-card) to
-     3.6:1 (--bg-deep), where 0.30 gave a barely-there 2.0:1, which is the
-     invisibility this ticket exists to fix. Dimming this below ~0.45 re-hides
-     the bar by another name. --scroll-gutter is the track width AND the grab
-     target. */
+  /* Terminal scrollbar thumb (CROW-1016 → CROW-1020). Gold so it reads as
+     chrome, not content, at the lowest alpha that still clears WCAG 2.2
+     §1.4.11's 3:1 non-text contrast against the surface it sits on — 0.50 is
+     3.46:1 over the xterm background (#1e1e1e), where the 0.20 xterm derives by
+     default from the foreground measured 1.67:1, which is the invisibility
+     CROW-1020 exists to fix. Dimming this below ~0.45 re-hides the bar by
+     another name (0.45 is 3.06:1; 0.44 is 2.98:1).
+
+     Declared here rather than in app.js because these are theme tokens and this
+     is where the palette lives — but xterm paints its slider from JS `theme`
+     options, not from CSS, so app.js's scrollbarTheme() reads them back out.
+     One definition, two consumers; see the CROW-1020 note by #terminal. */
   --scroll-thumb: rgba(221, 196, 130, 0.50);
   --scroll-thumb-hover: rgba(221, 196, 130, 0.72);
-  --scroll-gutter: 10px;
+  --scroll-thumb-active: rgba(221, 196, 130, 0.90);
 }
 
 * { box-sizing: border-box; }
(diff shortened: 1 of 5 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- No new attack surface: changes are confined to static CSS tokens, DOM class toggling, and xterm theme options — no user-controlled input flows into these paths.
- `scrollbarTheme()` defensively omits keys when tokens are missing or empty, avoiding xterm's `css.toColor` throw on malformed values at Terminal construction (`app.js:5463-5471`).
- `updateTerminalScrollbar()` null-guards `term`, `term.buffer`, and `terminal-wrap` before reading state (`app.js:5486-5490`).

**Concerns:**
- None identified. Colour values originate from hardcoded `:root` tokens in `app.css`, not from runtime or remote data.

### Code Quality
- **Green** — The two-bug diagnosis (colour via JS theme + persistence via CSS override) is correctly split across `scrollbarTheme()` and the `.has-scrollback` gate. Fixing only one half would leave an invisible or unusable bar, and the implementation addresses both.
- **Green** — Gating persistence on `@media (hover: hover) and (pointer: fine)` while applying the colour fix everywhere is a sound tradeoff: avoids conflicting with `enableTouchScroll`'s one-finger shim while still making the thumb visible on touch during reveal-on-scroll (`app.css:463-467`).
- **Green** — Using `baseY > 0` as the scrollback gate matches xterm's own "bar needed" semantics and correctly suppresses the pin on alt-buffer sessions where the slider would otherwise fill the full track (`app.js:5490`).
- **Green** — Test coverage is strong: Swift structural pins in `WebTerminalAssetTests.swift` (hide rule, persistence declarations, theme handoff, token existence) plus 14 jsdom assertions in `terminal-scrollbar.test.js` covering gate toggling, teardown paths, and token mapping with distinct values to catch transposition bugs.
- **Green** — Reverting `#sidebar`/`#board` to CROW-593's overlay hide and removing the #1019 gutter from out-of-scope panes (`.settings-body`, `.notif-list`, `.wizard-list`) is intentional and well-documented in the PR body.

### Summary Table

---

PR #1025, opened 2026-08-14
Revalidate the daemon's own web assets with ETags (CROW-1024)

### LINKED ISSUE #1024: Web UI serves app.js with no cache headers → browsers run stale JS (Cursor scroll fix #1014 dead on :8787 until hard-refresh)
## Summary

The Cursor wheel-scroll fix (#1014, "Restore Cursor wheel scroll by keeping unified history on inline agents") works in the **desktop (Tauri) app** but Cursor terminals still have a **dead wheel in the browser at `http://127.0.0.1:8787`**. Root cause is **not** a second broken frontend and **not** stale daemon state: the daemon serves its web JS/CSS/HTML with **no cache-control headers**, and `index.html` references `/app.js` **unversioned**, so a browser keeps executing a **stale cached pre-#1014 `app.js`** whose `applySurfaceScrollback` caps xterm scrollback to `0` on `agent_surface` alone — killing Cursor's wheel. The desktop app runs a freshly-launched WebView with its own cache, so it fetched the fixed `app.js`; that is the entire difference.

This is a **durable web-asset caching defect** (browsers can silently run stale UI after *any* update), surfaced by the Cursor scroll fix.

## What was ruled out (with evidence)

- **Not stale served assets / unfixed frontend.** The `app.js` actually served on `:8787` *contains* #1014 — `curl http://127.0.0.1:8787/app.js` has `activeSurfaceUsesAltScreen()` and `applySurfaceScrollback()` capping to `0` only when `agent_surface && uses_alternate_screen`. Daemon is `crow 0.2.0` (current).
- **Not a daemon flag problem.** Live Cursor Manager terminals report the correct inline-agent flags: `agent_surface=true`, `uses_alternate_screen=false`, `scrollback_degraded=false`. `activeSurfaceUsesAltScreen()` is also missing-flag-safe (`!!(activeTerminal && undefined) === false` → keeps 50k).
- **Not a frozen `history-limit 0` window.** `scrollback_degraded=false` on the live terminals.
- **Not `terminal.html`.** That is a debug-only page served at `/terminal.html` (`StaticAssets.swift:35-36`), not loaded by the normal UI, and it wouldn't dead-wheel Cursor anyway (unconditional mouse-swallow + fixed `scrollback: 50000`).
- **Desktop and web are the same frontend.** `crow-desktop` (Tauri) navigates its webview to `http://127.0.0.1:8787` (`crow-desktop/src-tauri/src/lib.rs`), loading the *same* `app.js` from the *same* daemon. So the only thing that can differ between them is **which `app.js` bytes each client has cached**.

## Root cause

The daemon attaches no caching/validation headers to its web assets:

- `Packages/CrowDaemon/Sources/CrowDaemon/StaticAssets.swift:82-101` — `fileResponse` sets only `Content-Type` + `X-Content-Type-Options: nosniff` (and CSP on `index.html`). **No `Cache-Control`, no `ETag`, no `Last-Modified`.**
- `Resources/web/index.html:85` — `<script src="/app.js"></script>` (and `:14` `app.css`, `:86` `settings.js`) are referenced **unversioned** — no `?v=` cache-bust.
- `Resources/web/version.json` carries a usable `gitSha` (`f364aa9b`) but it is only About-tab build info; it is **not** used to fingerprint or bust asset URLs.

With no freshness/validation headers and no versioned URL, a browser heuristically caches `/app.js` and reuses the old copy. The pre-#1014 `applySurfaceScrollback` body was `wanted = activeSurfaceIsAgent() ? 0 : UNIFIED_SCROLLBACK` — it caps to `0` on `agent_surface` alone, and Cursor is an agent surface → xterm scrollback `0` → wheel is a no-op. That stale copy is what the browser is still executing.

## Impact

- Cursor terminals can't wheel-scroll in any browser tab that cached `app.js` before the #1014 build — the reported symptom.
- Broader: **every** future web-UI change can silently fail to reach an already-open browser until a manual hard-refresh. The test suite can't catch it — `wheel-scroll.test.js` / `WebTerminalAssetTests` assert on the on-disk source, not on what a browser has cached.

## Immediate workaround

Hard-refresh the `:8787` tab (Cmd-Shift-R / empty-cache-and-reload). Cursor scroll then works with no code change — proving the source fix is correct and the issue is client cache.

## Proposed durable fix

Add cache-busting to `StaticAssets` (pick one; option A is simplest and correct):

- **A — revalidation headers.** Serve the HTML/JS/CSS with `Cache-Control: no-cache` plus a strong `ETag` (hash of the bytes) or `Last-Modified`, and answer conditional GETs with `304`. Browser always revalidates; a changed build ships immediately, unchanged builds stay cheap.
- **B — fingerprinted/versioned URLs.** Reference assets as `/app.js?v=<gitSha>` (the `gitSha` already in `version.json`) or content-hashed names, served `Cache-Control: immutable`. A new build changes the URL and busts the cache.

Keep xterm/static vendor assets long-cached; only the app's own HTML/JS/CSS need this.

## Acceptance criteria

- [ ] After deploying a new `crow` build, an already-open `:8787` browser tab picks up the new `app.js` on normal reload (no manual empty-cache required).
- [ ] Cursor terminals wheel-scroll in the browser at `:8787` (matching the desktop app).
- [ ] `app.js` / `app.css` / `settings.js` / `index.html` are served with `Cache-Control` + validation (or versioned URLs); a regression test asserts the header/URL policy.

## References

- Fix that this cache defect masks: #1014 (Cursor inline-agent scroll); ADR-0013 (`docs/adr/0013-terminal-scroll-model.md`), three-surface model.
- Related but distinct: **#1023** (Claude Code alt-buffer surface has no client scrollback / dead wheel when not mouse-tracking) — a different surface class, not this caching issue.
- Code: `Packages/CrowDaemon/Sources/CrowDaemon/StaticAssets.swift:82-101`; `Resources/web/index.html:14,85-86`; `Resources/web/version.json`.

---
*Filed from a live investigation on 2026-08-14: served `:8787` asset confirmed fixed, daemon flags confirmed correct, root cause isolated to missing asset cache-control + unversioned asset URLs.*

dhilgaertner:
Closes #1024

## Problem

The daemon served its web UI (`index.html`, `app.js`, `app.css`, `settings.js`, …) with **no `Cache-Control` or validation headers**, and `index.html` references `/app.js` unversioned. Browsers therefore heuristically cache `app.js` and keep executing the **stale** copy after a new build — which is why the Cursor wheel-scroll fix (#1014) worked in the desktop Tauri app (fresh WebView cache) but stayed **dead in a `:8787` browser tab** until a manual hard-refresh.

## Fix (Option A — revalidation headers)

Serve the app's own HTML/JS/CSS with `Cache-Control: no-cache` plus a **strong content-hash (SHA-256) `ETag`**, and answer a matching conditional GET with an empty **`304`**. The browser now revalidates on every reload, so a new build lands immediately while unchanged assets stay cheap (empty 304, no body).

- Localized to `StaticAssets.fileResponse` — the single choke point every app asset flows through. The request is threaded into `webResponse`/`loginPage` so `If-None-Match` can be read.
- The ETag is a **content hash** (not mtime/path), so identical bytes validate identically across installs and between the `webDir` and bundle sources.
- `If-None-Match` handling covers a comma-separated list, an optional `W/` weak prefix, and `*`.

## `/xterm/*` left as-is (deliberate)

The xterm 6.0.0 vendor bundle keeps its long-lived heuristic cache (`revalidate: false`). Its URL is **unversioned**, so marking it `immutable` would re-introduce the same staleness class on a future xterm upgrade — leaving it untouched is the safest minimal choice, and it matches the ticket's scoping.

## Tests

New `StaticAssetsCacheTests` drives the real router via `HummingbirdTesting` (`.test(.router)`):
- `app.js` → `200` with `Cache-Control: no-cache` + a quoted strong `ETag`.
- Conditional GET with the current ETag → empty `304` (validating headers retained).
- A stale `If-None-Match` → full `200` body.
- `index.html` keeps its CSP **and** the new cache headers.
- `/xterm/xterm.js` carries **neither** `Cache-Control` nor `ETag` (scoping guard).

`swift build` + the new suite pass. (Two unrelated suites — `RPCLanePolicyTests` and `WebNotificationCenterTests` — fail identically on the branch **before** this change; they're pre-existing and untouched here.)

## Acceptance criteria

- [x] An already-open `:8787` tab picks up the new `app.js` on a normal reload (no empty-cache required).
- [x] Cursor terminals can wheel-scroll in the browser once the new `app.js` is fetched (the #1014 fix is no longer masked by a stale cache).
- [x] `app.js` / `app.css` / `settings.js` / `index.html` are served with `Cache-Control` + validation; a regression test asserts the header/304 policy.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/StaticAssets.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/StaticAssets.swift
@@ -1,4 +1,5 @@
 import CrowTerminal
+import Crypto
 import Foundation
 import Hummingbird
 import NIOCore
@@ -13,27 +14,27 @@ import NIOCore
 /// refresh, no rebuild.
 enum StaticAssets {
     static func mount(on router: Router<CrowHTTPContext>, webDir: String? = nil) {
-        router.get("/") { _, _ in webResponse("index.html", webDir: webDir) }
-        router.get("/index.html") { _, _ in webResponse("index.html", webDir: webDir) }
+        router.get("/") { req, _ in webResponse("index.html", webDir: webDir, request: req) }
+        router.get("/index.html") { req, _ in webResponse("index.html", webDir: webDir, request: req) }
         // Login page (CROW-593) — reachable without auth; the auth middleware
         // also serves it as the fallback for unauthenticated navigational GETs.
-        router.get("/login") { _, _ in webResponse("login.html", webDir: webDir) }
-        router.get("/app.css") { _, _ in webResponse("app.css", webDir: webDir) }
-        router.get("/app.js") { _, _ in webResponse("app.js", webDir: webDir) }
+        router.get("/login") { req, _ in webResponse("login.html", webDir: webDir, request: req) }
+        router.get("/app.css") { req, _ in webResponse("app.css", webDir: webDir, request: req) }
+        router.get("/app.js") { req, _ in webResponse("app.js", webDir: webDir, request: req) }
         // Web Settings modal assets (CROW-581) — split out of app.css/app.js.
-        router.get("/settings.css") { _, _ in webResponse("settings.css", webDir: webDir) }
-        router.get("/settings.js") { _, _ in webResponse("settings.js", webDir: webDir) }
-        router.get("/brand.svg") { _, _ in webResponse("brand.svg", webDir: webDir) }
+        router.get("/settings.css") { req, _ in webResponse("settings.css", webDir: webDir, request: req) }
+        router.get("/settings.js") { req, _ in webResponse("settings.js", webDir: webDir, request: req) }
+        router.get("/brand.svg") { req, _ in webResponse("brand.svg", webDir: webDir, request: req) }
         // Build info for the Settings → About tab, written by
         // scripts/generate-build-info.sh. 404s gracefully when absent (the daemon
         // stays buildable without it — CROW-581).
-        router.get("/version.json") { _, _ in webResponse("version.json", webDir: webDir) }
+        router.get("/version.json") { req, _ in webResponse("version.json", webDir: webDir, request: req) }
         // Session-validity probe, gated by WebAuthMiddleware: 204 when the session
         // cookie is valid (or loopback), 401 when it isn't. The web UI polls this on
         // disconnect to tell "session expired" from "crowd is down" (CROW-593).
         router.get("/auth/check") { _, _ in Response(status: .noContent) }
         // The standalone single-terminal page from M1, kept for debugging.
-        router.get("/terminal.html") { _, _ in webResponse("terminal.html", webDir: webDir) }
+        router.get("/terminal.html") { req, _ in webResponse("terminal.html", webDir: webDir, request: req) }
 
         router.get("/xterm/:file") { _, context -> Response in
             // Basename-only guard against path traversal.
@@ -56,18 +57,19 @@ enum StaticAssets {
     }
 
     /// The login page as a 200 response — used by the auth middleware as the
-    /// fallback for unauthenticated navigational GETs (CROW-593).
-    static func loginPage(webDir: String?) -> Response {
-        webResponse("login.html", webDir: webDir)
+    /// fallback for unauthenticated navigational GETs (CROW-593). The `request`,
+    /// when the caller has one, lets the response answer a conditional GET.
+    static func loginPage(webDir: String?, request: Request? = nil) -> Response {
+        webResponse("login.html", webDir: webDir, request: request)
     }
 
     /// Load a web UI file — from `webDir` on disk when set (live/hot-reload),
     /// otherwise from the daemon bundle's `web/` resource directory.
-    private static func webResponse(_ name: String, webDir: String?) -> Response {
+    private static func webResponse(_ name: String, webDir: String?, request: Request? = nil) -> Response {
         if let webDir {
             let url = URL(fileURLWithPath: webDir).appendingPathComponent(name)
             if let data = try? Data(contentsOf: url) {
-                return fileResponse(data, name: name)
+                return fileResponse(data, name: name, request: request, revalidate: true)
             }
         }
         let base = (name as NSString).deletingPathExtension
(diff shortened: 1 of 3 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Revalidation is localized to `StaticAssets.fileResponse` (`Packages/CrowDaemon/Sources/CrowDaemon/StaticAssets.swift:107`), the single choke point for the app’s HTML/JS/CSS. `/xterm/*` stays on `revalidate: false`, matching the ticket’s scope and avoiding an unversioned `immutable` trap on a future xterm bump.
- ETags are strong, quoted SHA-256 content hashes — not path/mtime — so identical bytes validate across installs and between `webDir` and the bundle. A 304 is therefore tied to the bytes that would have been served, with no TOCTOU between hash and body.
- Auth still runs in `WebAuthMiddleware` before the handler. `Packages/CrowDaemon/Sources/CrowDaemon/WebAuthRoutes.swift:75` threads the request into `loginPage` only so a conditional GET can 304; it cannot skip the cookie/loopback check on gated assets.
- Existing hardening is preserved: `X-Content-Type-Options: nosniff`, and `index.html` still carries its CSP on both 200 and 304.

**Concerns:**
- Green: `Cache-Control: no-cache` without `private` lets a shared cache store these responses (they must still revalidate). The bytes are identical for every user, so this is not a leak; `private, no-cache` would only be slightly stricter for reverse-proxied daemons.

### Code Quality
- Green: `ifNoneMatch` (`Packages/CrowDaemon/Sources/CrowDaemon/StaticAssets.swift:132`) implements comma-separated lists, an optional `W/` prefix, and `*` per RFC 9110 §13.1.2, but `Packages/CrowDaemon/Tests/CrowDaemonTests/StaticAssetsCacheTests.swift` only round-trips a single current ETag and a stale one. A couple of extra cases (`W/"…"`, a two-tag list, `*`) would lock those branches.
- The five new tests pass locally (`swift test --package-path Packages/CrowDaemon --filter StaticAssetsCacheTests`). Linux CI does not run CrowDaemon tests (existing link limitation), so this suite is macOS-only by current policy.

### Summary Table

---

Issue #1026, opened 2026-08-14
Terminal scrollback develops holes in the middle (tmux thinning); Cursor/agent surfaces never auto-heal


dhilgaertner:
## Summary

In a long-running terminal, the **middle of the scrollback develops holes** as output accumulates: the beginning stays intact, the current screen (bottom) stays intact, but a growing band in the middle loses lines. It reproduces on **both** the Tauri desktop app and the web UI on `:8787` (so it's in the shared daemon/JS layer, not a browser-cache issue like #1024). The only recovery today is manually pressing **↻ Reload** (or switching tabs) — and on agent/Cursor surfaces there is **no automatic recovery at all**.

## Root cause

The local xterm buffer is **not** an authoritative mirror of the tmux pane; the only thing that reconciles them is the CROW-606 `select-window` → `capture-pane -pe -S -50000` replay. Between replays, **tmux thins pane redraws for a slow-draining client, and the browser/webview is that slow client.** The code's own measurement (`app.js:6010-6023`): `seq 1 20000` leaves ~19,989 lines in the pane, but a browser-paced xterm receives only **~337** of them. So:

- **Beginning** = complete (the connect-time replay was authoritative at that moment).
- **End** = complete (the live current frame is always fully painted).
- **Middle** = post-connect live output, thinned by tmux → **holes that grow as output accumulates.**

The automatic healer, `maybeHydrateScrollback` (CROW-934, `app.js:6127-6155`), cannot fix this for the affected surfaces:

1. **It only fires as a transition to the very top of the buffer** — `arrivedAtTop = buf.viewportY === 0 && lastViewportY !== 0` (`:6130`), bails when `viewportY !== 0` (`:6134`) and unless it's the arrival transition (`:6137`). A hole at viewportY 500 of a 5000-line buffer is never a trigger → **never healed**. There is no mid-buffer or idle re-sync (the 4s `setInterval(refreshLive, 4000)` at `app.js:6705` does not touch the terminal; `refreshTerminals` is only called on user actions).
[...]

PR #1028, opened 2026-08-14
Self-heal a reload that rebuilds a one-screen terminal buffer (CROW-1027)

### LINKED ISSUE #1027: ↻ Reload intermittently rebuilds a one-screen terminal buffer → dead wheel (resize races the capture-pane)
## Summary

Pressing the **↻ Reload** button on a terminal — the main way to recover a holey scrollback (see sibling bug) — **sometimes leaves the terminal completely unable to scroll** (dead wheel, no scrollbar). Another Reload usually brings it back. It reproduces on **both** the Tauri desktop app and the web UI on `:8787`. This is a **different mechanism** from #1024: there the wheel died because `scrollback` was capped to `0` (stale cached JS); here the `scrollback` capacity is correct (50000) but the rebuilt buffer comes back **only one screen tall** (`baseY === 0`), so there is nothing to scroll.

## Root cause

`reloadTerminal` (`app.js:6181-6197`) empties the buffer — `term.reset()` (`:6183`) plus the replay's `ESC[3J` clear-scrollback — and then refills it **only** from the CROW-606 replay. Whether that replay rebuilds a tall (scrollable) buffer or a one-screen (dead) one is hostage to timing, because the reload **resizes the pane at the exact moment it re-captures it**:

- On (re)connect, `onopen` fires `applyTermFit()` (a resize) **immediately before** `selectWindow()` (the replay) — deliberately: "fit synchronously so the resize reaches the PTY before the scrollback replay" (`app.js:5891-5898`).
- The daemon's `select-window` handler runs `capture-pane` **right after** switching the window (`TerminalWebSocket.swift` → `cockpit.selectWindow(...)` then `cockpit.replayData(...)` → `capturePane(...)`, `TerminalCockpit.swift:134-138`).
- So the `capture-pane` can run **mid-reflow / mid-repaint** from the resize's SIGWINCH. A main-buffer app (Cursor, a shell) that clears+redraws on SIGWINCH can be captured with only its current screen present → `replayFrame` prepends `ESC[3J` (clears scrollback) then writes that one screen → `baseY = 0` → `term.scrollLines` is a silent no-op (`app.js:5669`) → **dead wheel**.

The client never verifies or re-establishes local scrollback after a reload — it trusts the single replay.

### Why "sometimes"

The outcome depends on interleavings that vary per reload:

- **Focus gate on the resize.** `applyTermFit` early-returns when `document.hidden || !document.hasFocus()` (`app.js:5967`), and only sends a resize if the grid actually changed (`:5973`). So whether a reload perturbs the pane at capture time depends on focus/visibility — which also explains why Tauri desktop vs browser behave differently from the *same* code.
- **Reflow vs capture timing** between the two sequential tmux calls (`select-window`, then `capture-pane`).
- **Multi-client size flap.** Each reload opens a fresh grouped view/PTY (`TerminalWebSocket.swift`); with `window-size latest` + `aggressive-resize on` (`crow-tmux.conf`) a transient overlap with the tearing-down old client (or the desktop app) resizes the shared window during the capture.
- **Double-connect hazard.** If the user hits ↻ while an `onclose` auto-reconnect is already pending (`setTimeout(connectTerminalWs, …)`, `app.js:5940`), `reloadTerminal` starts one socket while the pending timer starts another → two attaches, two `select-window` replays, two resizes racing on the shared window and the shared `term`. `reloadTerminal` does not cancel a pending reconnect timer.

## Impact

- The primary manual recovery for the holey-scrollback bug (sibling ticket) is itself unreliable: a user hits a holey middle, reloads to fix it, and can land in a **totally unscrollable** terminal — compounding the two bugs.
- Purely a client/attach-sequencing issue; the tmux pane still holds the full history (another reload, or a tab switch, re-captures it).

## Proposed fixes (ranked)

1. **Capture against a settled size — don't let the resize race the capture.** Either issue `select-window` *before* the resize on the reload attach, or (daemon) capture the pane *before* applying a just-received resize, or add a short settle so `capture-pane` runs after the reflow. (`app.js:5896-5898`; `TerminalWebSocket.swift` select-window handler.) Highest leverage — removes the intermittency at its source.
2. **Verify/repair local scrollback after reload.** Once the reload's replay settles, if `buf.baseY === 0` for a non-alt surface that should have history, re-issue one `selectWindow` capture. Client-only guard that turns "sometimes dead" into "self-heals." (`app.js` onopen, ~`:5897-5902`.)
3. **Close the double-connect window.** In `reloadTerminal` (`:6181-6197`) and the `onclose` reconnect (`:5940`), cancel any pending reconnect `setTimeout` before creating a new socket.
4. **Make the reload's resize deterministic** — don't gate it on focus, or skip the resize entirely on a reload attach (`app.js:5967`), so desktop-vs-browser focus differences stop making the result non-deterministic.

## Acceptance criteria

- [ ] Pressing ↻ Reload on a Cursor/Manager (or shell) terminal reliably rebuilds a **scrollable** buffer — the wheel/scrollbar work every time, not sometimes.
- [ ] Rapidly pressing Reload (or Reload during an auto-reconnect) does not spawn two racing attaches.
- [ ] Alt-buffer agents (Claude Code) unaffected (their local wheel is by-design forwarded, not local).

## References

- Distinct from #1024 (dead wheel via `scrollback = 0` from stale cached JS). This is an **empty-buffer / `baseY = 0`** dead wheel with correct scrollback capacity.
- Shares an underlying cause with the sibling bug **"middle of scrollback develops holes"** (filed alongside) — both are failure modes of the single `select-window`/`capture-pane` reconciliation. This bug makes that bug's manual heal unreliable.
- Related: ADR-0013 (`docs/adr/0013-terminal-scroll-model.md`), CROW-979/#981 (the ↻ Reload button), #1023.
- Code: `app.js:6181-6197` (`reloadTerminal`), `:5880-5903` (`onopen`: resize then select-window), `:5958-5978` (`applyTermFit` focus gate), `:5940` (uncancelled reconnect timer); `TerminalWebSocket.swift` (`select-window` → `replayData`), `TerminalCockpit.swift:178-184` (`replayFrame` `ESC[3J`).

---
*Filed from a live investigation on 2026-08-14; mechanism verified against source (onopen resize-before-replay, daemon capture-after-select, replayFrame clear-scrollback).*

dhilgaertner:
Closes #1027

## Problem

Pressing **↻ Reload** on a shell/Manager terminal *sometimes* leaves it unable to scroll — dead wheel, no scrollbar — until another Reload. Reproduces on both the Tauri desktop app and the web UI.

**Mechanism.** `reloadTerminal()` blanks the buffer and refills it only from the CROW-606 select-window replay. On the attach, `onopen` runs `applyTermFit()` (a resize) immediately before `selectWindow()` (the replay) — deliberately, so the real pane width reaches the PTY before the capture. That resize fires a SIGWINCH; the daemon's select-window handler runs `capture-pane` right after switching the window, so the capture can land **mid-reflow** and come back one screen tall. The replay's `ESC[3J` then wipes the little history the buffer had → `baseY === 0` → `term.scrollLines` is a silent no-op → dead wheel. It's "sometimes" because the reflow-vs-capture timing, the `applyTermFit` focus gate, and multi-client `window-size latest` flap all vary.

## Fix

Kept **client-only** — the daemon `/terminal` handler needs a real PTY + tmux and can't be unit-tested, whereas the jsdom harness fully drives sockets, timers, and `baseY`.

1. **Post-attach settle-and-verify self-heal.** Keep the deliberate resize-before-replay ordering (a pure reorder would just trade a dead wheel for history captured at the fresh attach's 80×24 default, and the grouped attach perturbs window size on its own). After each attach a ~250 ms settle checks the rebuilt buffer: if a non-alt **shell** surface came back one screen tall (`baseY === 0`), re-issue **one** `select-window` so crowd re-captures the now-settled pane and the wheel returns. A healthy buffer (`baseY > 0`) sends nothing; the retry is bounded (`SCROLLBACK_HEAL_MAX`) so a genuinely one-screen pane is accepted rather than re-captured forever. Alt-buffer agents (Claude Code) are skipped — their wheel is forwarded by design.
2. **Close the double-connect race.** The `onclose` auto-reconnect `setTimeout` stored no id, so a manual Reload (or `attachWindow`) while a reconnect was pending could open a second socket racing the first on the shared window/term. Track the id and cancel it in `connectTerminalWs` and `reloadTerminal`.

Not touching the `applyTermFit` focus gate (load-bearing for #667 shared-size ownership) or ADR-0013 scroll behavior. Scoped to the reload race — #1026's holey-middle hydrate gating is separate.

## Acceptance criteria

- [x] ↻ Reload reliably rebuilds a **scrollable** buffer (self-heal re-captures when the first replay was one-screen).
- [x] Rapid Reload / Reload-during-reconnect does not spawn two racing attaches.
- [x] Alt-buffer agents (Claude Code) unaffected.

## Tests

- `Packages/CrowDaemon/web-tests/terminal-reload.test.js` (jsdom): new scenarios for self-heal, healthy-no-recapture, bounded/never-loops, alt-buffer + agent-surface skip, and no-double-connect. **65 passed, 0 failed.**
- Full web-tests CI suite: all 16 files, 0 failed.
- `swift test --package-path Packages/CrowDaemon` builds clean; `TerminalReplayTests` green (daemon unchanged).

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -5227,6 +5227,11 @@ let searchAddon = null;
 let termWs = null;
 let termSkelTimer = null; // #677: safety timeout that clears the skeleton overlay
 let termReconnectDelay = 1000; // #687: backoff between reconnect attempts (ms), reset on open
+// CROW-1027: id of the pending onclose auto-reconnect. Tracked so a manual
+// ↻ Reload (or attachWindow) can cancel it before opening its own socket —
+// otherwise the stray timer fires a SECOND attach that races the reload's on the
+// shared window/term (dead-wheel double-connect hazard).
+let termReconnectTimer = null;
 
 // In-flight state for the header's ↻ Reload button (CROW-979). Module state, not
 // DOM state: `refreshLive` repaints the whole detail header every 4s, so a busy
@@ -5923,8 +5928,19 @@ function showTerminalMenu(e) {
   armContextMenuClose();
 }
 
+// CROW-1027: cancel a queued onclose auto-reconnect. Idempotent — safe to call
+// when nothing is pending, and a no-op when the timer's own callback (which nulls
+// the id first) is what's running.
+function clearTermReconnectTimer() {
+  clearTimeout(termReconnectTimer);
+  termReconnectTimer = null;
+}
+
 function connectTerminalWs() {
   if (sessionDead) { hideTerminalSkeleton(); return; } // don't loop after an expired remote cookie (review Yellow); clear any overlay so the #679 scrim owns the screen
+  // CROW-1027: this attach supersedes any queued auto-reconnect — drop it so two
+  // sockets can't race on the shared window/term.
+  clearTermReconnectTimer();
   // #677: cover this (re)attach — initial connect, auto-reconnect, and (via
   // reloadTerminal) the #675 right-click Reload + tab/session switch — with the
   // skeleton. `painted` gates the hide to the FIRST PTY byte of THIS socket.
@@ -5954,6 +5970,9 @@ function connectTerminalWs() {
       // Keep the attach bookkeeping truthful after any (re)connect — initial
       // connect, reloadTerminal, or a dropped-socket auto-reconnect (#673).
       attachedWindow = activeTerminal.window;
+      // CROW-1027: the fit above races the replay's capture, which can rebuild a
+      // one-screen (dead-wheel) buffer. Verify shortly after and re-capture if so.
+      armScrollbackHeal();
     }
   };
   termWs.onmessage = (event) => {
@@ -5992,7 +6011,9 @@ function connectTerminalWs() {
     if (sessionDead) { hideTerminalSkeleton(); return; }
     setTerminalReconnecting(true);
     showTerminalSkeleton();
-    setTimeout(connectTerminalWs, termReconnectDelay);
+    // CROW-1027: track the id so a manual reload can cancel this before it fires
+    // (null it first inside the callback so a later clear is a harmless no-op).
+    termReconnectTimer = setTimeout(() => { termReconnectTimer = null; connectTerminalWs(); }, termReconnectDelay);
     termReconnectDelay = Math.min(termReconnectDelay * 2, 10000);
   };
   termWs.onerror = () => termWs.close(); // funnels to onclose (keeps overlay up)
@@ -6121,6 +6142,18 @@ let scrollbackFullySynced = false; // last capture returned nothing new
 let lastHydrateAt = 0;
 let lastViewportY = 0; // previous scroll position, to detect arrival at the top
 
+// CROW-1027: post-attach scrollback self-heal. The reload/attach fits (resizes)
+// the pane immediately before select-window replays it, so `capture-pane` can run
+// mid-reflow and come back ONE screen tall — the replay's ESC[3J then wipes what
+// little history the buffer had, leaving `baseY === 0` and a dead wheel. A brief
+// settle after attach, then a re-capture when the buffer really is one screen
+// tall, rebuilds it against a now-settled pane. Gated on `baseY === 0` so a
+// healthy attach (the common case) sends nothing, and bounded so it never loops.
+const SCROLLBACK_HEAL_MS = 250; // distinct from 1500 (skeleton) / 10000 (reload settle)
+const SCROLLBACK_HEAL_MAX = 2; // bounded re-captures — accept a genuine one-screen pane after this
+let scrollbackHealTimer = null;
+let scrollbackHealLeft = 0;
+
 // The flight ends this long after the last frame, but never later than the
 // arm-time deadline. There is deliberately NO viewport restore here: the replay
 // rebuilds the buffer while `isUserScrolling` is true (which the trigger's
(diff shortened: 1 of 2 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- No new endpoints, auth, or control-frame types. The self-heal reuses the existing authenticated `/terminal` `select-window` path with `activeTerminal.window` — the same index `onopen` already sends.
- Recapture is bounded (`SCROLLBACK_HEAL_MAX = 2`), so a flapping attach cannot turn into an unbounded `capture-pane` loop.
- Tracking `termReconnectTimer` and cancelling it in `connectTerminalWs` / `reloadTerminal` closes a real double-attach race: two sockets writing the same xterm instance.

**Concerns:**
- None. Extra `select-window` traffic is same-origin, same-privilege, and capped.

### Code Quality
- The reconnect-timer cancel in `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` is correct: the callback nulls the id before re-entering `connectTerminalWs`, `clearTermReconnectTimer` is idempotent, and `reloadTerminal` still detaches the old socket's `onclose` before opening the next one.
- Self-heal teardown is wired through `resetScrollbackSync`, so a drop or a subsequent reload cannot fire `select-window` on a dead socket. Gates for missing term/window, non-OPEN `termWs`, and `buffer.type === 'alternate'` are sound.
- Tests in `Packages/CrowDaemon/web-tests/terminal-reload.test.js` cover the three ACs (heal, healthy no-op, bound, agent skip, no double-connect). `node terminal-reload.test.js` → 65 passed, 0 failed. Unrelated `row.test.js` glyph assertions failed in this environment (10 failures, emoji/glyph matching); this PR does not touch row rendering.

**Green — consider:**
- **Inline agents are skipped with the alt-buffer predicate.** `verifyScrollbackAfterAttach` returns on `activeTerminal.agent_surface` (`app.js:6239`) and the comment calls this “alt-buffer agents (Claude Code)”. After CROW-1010 / ADR-0013, Cursor is `agent_surface` with *unified* 50k scrollback (`activeSurfaceUsesAltScreen()` is the cap that actually means “no local history”). Copying `maybeHydrateScrollback`’s gate is consistent and not a regression, but a Cursor Reload can still land `baseY === 0` and will not self-heal. `activeSurfaceUsesAltScreen()` (already in this file) would match the three-class model without changing Claude Code’s skip.
- **Settle clock starts at `onopen`, not after the first replay.** `armScrollbackHeal` runs as soon as `select-window` is *sent* (`app.js:5975`). `baseY === 0` at 250 ms is therefore ambiguous: genuinely one-screen (the bug) vs. a large CROW-606 replay still in flight / still inside async `term.write`. The dead-wheel case is a *small* capture (arrives fast); the healthy case is a *large* one (arrives slower), so the heuristic can spend its two recaptures on a pane that was about to be fine. Bounded, so it will not loop — starting the 250 ms after the first binary frame (or `term.write`’s callback) would distinguish the two. Author-accepted timing; not a blocker.
- `terminal-reload.test.js` is missing a trailing newline.

### Summary Table

---

PR #1029, opened 2026-08-14
Auto-heal scrollback holes on inline agents and mid-buffer (CROW-1026)

### LINKED ISSUE #1026: Terminal scrollback develops holes in the middle (tmux thinning); Cursor/agent surfaces never auto-heal

dhilgaertner:
Closes #1026

## Problem

In a long-running terminal, the **middle of the scrollback develops holes** as output accumulates (beginning and current screen stay intact). Root cause (verified against source): tmux thins pane redraws for a slow-draining client — the browser is that client — so live output between CROW-606 replays arrives incomplete. `app.js`'s own measurement: `seq 1 20000` leaves ~19,989 lines in the pane but a browser-paced xterm receives only ~337.

The CROW-934 healer `maybeHydrateScrollback` couldn't fix this:

1. **It excluded every `agent_surface` terminal.** That guard was written for **alt-buffer** agents (Claude Code), which legitimately have no scrollback — but post-ADR-0013 it also silenced **inline** agents (Cursor, and every **Manager** tab), which keep a real unified 50k buffer that *does* thin. They never auto-healed; the only recovery was ↻ Reload.
2. **It only fired on arrival at the very top.** A hole the user was looking at mid-buffer never healed unless they scrolled all the way to line 1.

## Fix

**Fix 1 — narrow the exclusion.** From `agent_surface` to `agent_surface && uses_alternate_screen`, reusing the existing `activeSurfaceIsAgent()` / `activeSurfaceUsesAltScreen()` helpers so it stays in lockstep with `applySurfaceScrollback` (the same axis ADR-0013 #1010 uses to cap xterm scrollback). Inline agents and plain shells — including a plain shell inside a Claude session, which is `uses_alternate_screen` but **not** `agent_surface` — become eligible; only true alt-buffer agents stay excluded.

**Fix 2 — scroll-idle mid-buffer heal.** A debounce armed off `onScroll` (which also fires per bottom-pushed line) re-captures **once**, only after scrolling *and* output have both gone quiet — never during a streaming build. It reuses every existing brake (in-flight guard, `scrollbackFullySynced` latch, dirty flag, `HYDRATE_MIN_INTERVAL_MS` cooldown, unchanged settle), so the #935 extend-forever bug cannot return. As on the arrival path there is deliberately no viewport restore, so the `isUserScrolling` rebuild leaves the user at the top of the now-complete history.

The arm tail is factored into a shared `fireScrollbackCapture(buf)`; the arrival-at-top path is otherwise unchanged. **JS-only** — no daemon behavior change, and the ADR-0013 scroll model / reload path (sibling #1027) are untouched.

## Acceptance criteria

- [x] A Cursor/Manager terminal shows a complete, hole-free scrollback without a manual Reload (inline agents now eligible).
- [x] A hole in the middle of the buffer heals automatically, not only at the very top (scroll-idle trigger).
- [x] Alt-buffer agents (Claude Code) are unaffected — still no spurious captures.

## Tests

- `web-tests/scrollback-hydrate.test.js`: split the agent-surface case into **alt-buffer (still excluded)** + **inline agent (now hydrates)**, added a **plain-shell-in-Claude-session** case, and a **mid-buffer scroll-idle** group (heals once; a streaming build never fires it; alt-buffer excluded; dirty/cooldown/latch respected). **44 pass, 0 fail.**
- `WebTerminalAssetTests` (CrowDaemon, 20) and `ScrollbackHealthTests` (CrowTerminal, 9) stay green.
- Full CI-gated web suite: 0 failures.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -6083,14 +6083,15 @@ function selectWindow(win) {
   }
 }
 
-// CROW-934: re-sync a plain-shell surface's scrollback from tmux's history.
+// CROW-934: re-sync a scrollback surface's history from tmux's pane.
 //
 // tmux collapses pane redraws when the attached client can't drain them fast
 // enough, and the browser IS that slow client. Measured against the bundled
 // config: `seq 1 20000` leaves all 19 989 lines in the pane's own history, but
 // a browser-paced reader receives only ~337 of them (a raw reader gets 19 768).
 // So the local xterm buffer is NOT a faithful copy of the pane — scroll-up hits
-// its top after a few screens while tmux still holds the rest.
+// its top after a few screens while tmux still holds the rest, and the middle
+// grows holes as thinned live output accumulates (CROW-1026).
 //
 // crowd's `select-window` reply carries the authoritative copy
 // (`capture-pane -pe -S -50000`, CROW-606) and REBUILDS the buffer in place, so
@@ -6098,11 +6099,13 @@ function selectWindow(win) {
 // (attachWindow → reloadTerminal), which is why switching away and back
 // "fixed" the history — but the tab you sit on never re-synced.
 //
-// Trigger on reaching the top of the local buffer: that is exactly when the
-// user is asking for older lines. Shell surfaces only — an agent keeps its
-// transcript in the scrollback-less alt buffer and repaints it itself
-// (ADR-0013), so a capture there returns just the viewport and would buy
-// nothing.
+// Two triggers: reaching the top of the local buffer (asking for older lines)
+// and, since CROW-1026, going idle parked on a hole mid-buffer. Any surface that
+// keeps the unified 50k scrollback is eligible — plain shells AND inline agents
+// (Cursor, and the Manager, which runs one). Only a true alt-buffer agent (Claude
+// Code) is excluded: it keeps its transcript in the scrollback-less alt buffer
+// and repaints it itself (ADR-0013 #1010), so a capture there returns just the
+// viewport and would buy nothing. Same axis as applySurfaceScrollback.
 //
 // Two independent brakes, because `onScroll` is NOT a user-scroll event: xterm's
 // BufferService.scroll ends in an unconditional `this._onScroll.fire(ydisp)`, so
@@ -6154,6 +6157,16 @@ const SCROLLBACK_HEAL_MAX = 2; // bounded re-captures — accept a genuine one-s
 let scrollbackHealTimer = null;
 let scrollbackHealLeft = 0;
 
+// CROW-1026: arrival-at-top is not the only place a hole appears — the thinned
+// live stream drops lines anywhere in the middle, and the user may be parked on
+// one. Debounce off `onScroll` (which also fires for every line pushed at the
+// bottom): the timer re-arms on each scroll AND each output frame, so it fires
+// exactly once, only after scrolling AND output have both gone quiet. During a
+// streaming build it never fires; when the user stops on a mid-buffer hole it
+// re-captures once, subject to every brake the arrival path uses.
+const HYDRATE_SCROLL_IDLE_MS = 400;
+let hydrateIdleTimer = null;
+
 // The flight ends this long after the last frame, but never later than the
 // arm-time deadline. There is deliberately NO viewport restore here: the replay
 // rebuilds the buffer while `isUserScrolling` is true (which the trigger's
@@ -6204,6 +6217,8 @@ function noteTerminalFrame() {
 function resetScrollbackSync() {
   clearTimeout(hydrateSettleTimer);
   hydrateSettleTimer = null;
+  clearTimeout(hydrateIdleTimer); // CROW-1026: drop a pending mid-buffer re-sync too
+  hydrateIdleTimer = null;
   hydratingScrollback = false;
   hydrateSawReplay = false;
   scrollbackFullySynced = false;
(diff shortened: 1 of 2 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- No new endpoints, credentials, or auth changes. Hydrate still sends the existing `select-window` control frame on the already origin-checked, cookie-gated `/terminal` socket.
- Eligibility uses the daemon-supplied `agent_surface && uses_alternate_screen` AND — the same axis as `applySurfaceScrollback` (ADR-0013 #1010). That AND is load-bearing: `uses_alternate_screen` is session-scoped, so a plain shell in a Claude session is `uses_alternate_screen: true` but `agent_surface: false` and correctly stays eligible. The issue's "exclude on `uses_alternate_screen` alone" would have silenced those shells.
- Every #935 brake (in-flight, dirty, latch, 5s cooldown, absolute settle deadline) is shared through `fireScrollbackCapture`. `resetScrollbackSync` also drops the new idle timer, so a reload/reconnect cannot fire a stale capture into a fresh socket.

**Concerns:**
- None. More frequent `capture-pane` stays per authenticated connection and is rate-limited by the existing cooldown + dirty/latch brakes.

### Code Quality
- **Green** — Idle heal leaves the user at line 1. `hydrateWhenIdle` in `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` documents that the `isUserScrolling` rebuild pins `ydisp` at 0, matching the PR body. That is a conscious trade against #935 round 3 (a viewport restore was harmful on the jump-to-bottom path). A later in-place restore scoped to the idle path only would match issue #1026's "heal without scrolling to line 1" more closely; it is not a defect in this diff.
- **Green** — Teardown coverage for the new timer is indirect. `resetScrollbackSync` clears `hydrateIdleTimer`, but the teardown fixtures in `Packages/CrowDaemon/web-tests/scrollback-hydrate.test.js` arm from `viewportY === 0`, where `hydrateWhenIdle` would no-op anyway. A mid-buffer setup + reset + `advance(500)` asserting no `select-window` would lock the new clear.
- Local `scrollback-hydrate.test.js`: **44 passed, 0 failed**. The alt-buffer / inline-agent / plain-shell-in-Claude-session matrix and the idle-heal group (stream debounce, dirty, exclusion) match the change. Arrival-path regressions stay intact.

### Summary Table

---

dhilgaertner:
Rebased onto latest `main` to resolve the conflict with the now-merged sibling PR #1027 (post-attach scrollback self-heal). The conflict was a both-added state block at the same spot in `app.js` — I kept both #1027's `SCROLLBACK_HEAL_*` declarations and this PR's `HYDRATE_SCROLL_IDLE_MS` / `hydrateIdleTimer`. The auto-merged `resetScrollbackSync` correctly clears both timers.

Re-verified after the rebase: `scrollback-hydrate.test.js` 44/0, full CI web suite 0 failures, `WebTerminalAssetTests` 20/20, `ScrollbackHealthTests` 9/9.

**Heads-up on a gate drift for the #1027 owner (not changed here, on purpose):** #1027's `verifyScrollbackAfterAttach` excludes surfaces with `if (activeTerminal.agent_surface) return;` and its comment says *"Same gate maybeHydrateScrollback uses."* This PR narrows `maybeHydrateScrollback`'s gate from `agent_surface` to `agent_surface && uses_alternate_screen` (so inline agents like Cursor/Manager, which keep a real 50k buffer, stay eligible). After this merges, the two gates no longer match, and #1027's post-attach one-screen heal still skips inline agents — which by the same reasoning could also be left one-screen-tall by a mid-reflow capture. I kept this PR scoped to the hydrate path per the ticket and did **not** touch the reload path; flagging it so #1027 can decide whether to align its gate (to `activeSurfaceIsAgent() && activeSurfaceUsesAltScreen()`) in a follow-up.

PR #1031, opened 2026-08-14
Make the About build SHA open its commit on GitHub (CROW-1030)

### LINKED ISSUE #1030: About menu: clicking the git SHA should open that commit on GitHub
In the About menu, the build’s git SHA is shown but is not clickable. Clicking it should open the Crow GitHub repo at that SHA (the commit page).

Example: SHA `abc1234` → `https://github.com/corveil/crow/commit/abc1234` (full hash if that’s what About displays).

## Acceptance criteria
- The SHA in About is a control (link or button), not inert text.
- Click opens the browser to that commit on `corveil/crow`.
- Dirty / unknown / missing SHA stays unlinked (or disabled) rather than opening a 404.
- Works from the desktop (Tauri) About surface and the web About if both show the SHA.

dgershman:
Closes #1030

## What

The build SHA is stamped and displayed on both About surfaces but wasn't reachable — answering "what is actually running here?" meant retyping seven characters into a URL.

**Web (Settings → About).** The SHA renders as an anchor to `github.com/corveil/crow/commit/<sha>`. The version line is built as DOM nodes instead of one `textContent` so the SHA can carry the `<a>`; it **displays the short SHA and hrefs `gitShaFull`** when `version.json` stamps it, so the visible line is byte-identical to before while the URL is unambiguous.

**Desktop (Tauri).** A **Crow → Build on GitHub** menu item beside About, opening the same commit through `tauri_plugin_opener`'s Rust-side `open_url` (not ACL-gated the way the frontend IPC command is). `build.rs` now also embeds `CROW_GIT_SHA_FULL`.

## Why the desktop control sits *beside* About, not inside it

This is a platform limit, not a preference. macOS draws the About panel itself (`NSAboutPanel`), and muda hands it plain strings — `credits` reaches it as an `NSAttributedString` built by `from_nsstring`, with no link attribute — so **no part of that panel's text can be made clickable from Tauri**. A menu item one row down is the closest reachable control. Happy to swap it for a different label or placement if you'd rather.

## The unlinkable case

Whether to link is one predicate per side — `crowCommitURL` (app.js) and `crow_commit_url` (Tauri shell) — returning nothing for any stamp that would 404: `dev`, absent/blank, non-hex, or outside git's 7–40 character range. Those bounds match `VersionUpdateClient.githubCompareURL`, which already guards the same stamp Swift-side for the compare URL, so all three agree on what counts as a commit. An unlinkable stamp is **plain text on the web and a disabled menu item on the desktop** — which is also why the check returns `nil`/`null` rather than a boolean beside a separately-built string.

One case this deliberately does **not** catch: a well-formed SHA that exists only locally (unpushed commit, fork build) still links and GitHub will 404 it. Detecting that needs a network round-trip, which the version-update check already does on its own schedule.

## Test plan

- `Packages/CrowDaemon/web-tests/about-sha-link.test.js` — 27 checks, registered in both `test` and `test:ci`. Covers the predicate (short/full/uppercase/whitespace; `dev`, empty, null, 6 chars, 41 chars, non-hex, path traversal, interior space) **and** the rendered About tab driven through the real `settings.js` modal, because a correct helper the About tab never calls looks identical to the bug from the user's side.
- Two Rust unit tests on the mirrored predicate (`cargo test` in `crow-desktop/src-tauri`).
- Mutation-checked: stubbing `crowCommitURL` to `null` in `renderVersionLine` fails exactly the 6 link assertions, so the suite isn't vacuous.
- `npm run test:ci` in `Packages/CrowDaemon/web-tests` — exit 0, no pre-existing suites disturbed.

**Not verified by me:** the native menu item was not clicked in a running desktop build — `crow-desktop` compiles and its unit tests pass, but I didn't launch the app. Worth a click before merge.


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- `crowCommitURL` in `app.js` and `crow_commit_url` in `lib.rs` both enforce a strict 7–40 lowercase-hex predicate before constructing URLs, blocking path traversal (`2a24aeb3/../../evil`), non-hex stamps, and out-of-range lengths — consistent with `VersionUpdateClient.githubCompareURL` in Swift.
- The upstream repo is hardcoded (`corveil/crow`); only validated hex enters the commit path, so there is no open-redirect or injection surface from `version.json` stamps.
- Web links use `target="_blank"` with `rel="noopener"`, and SHA display text is set via `textContent` (not `innerHTML`).
- Desktop menu item is disabled at build time for unlinkable stamps, and the click handler re-derives the URL from `CROW_GIT_SHA_FULL` rather than caching a stale string.

**Concerns:**
- None blocking. A well-formed SHA from a fork or unpushed local build still links to `corveil/crow` and may 404 on GitHub — the PR documents this as an accepted trade-off (network check deferred to the version-update path).

### Code Quality
- Web About tab correctly builds the version line as DOM nodes: short SHA displayed, full SHA in `href` when `gitShaFull` is stamped, plain text when the predicate returns null.
- `build.rs` refactor to `git_rev()` cleanly embeds both short and full SHAs with the same `dev` fallback.
- `about-sha-link.test.js` covers both the predicate and the end-to-end Settings → About render path (27 assertions); Rust unit tests mirror the refusal cases.
- `swift build` and `cargo test` in `crow-desktop/src-tauri` pass locally; `node about-sha-link.test.js` passes (27/27).

### Summary Table

| Finding | Color | Notes |
|---------|-------|-------|
| Fork/unpushed SHA may 404 on GitHub | Green | Documented accepted risk |
| Predicate duplicated JS/Rust/Swift | Green | Matches existing `githubCompareURL` pattern; tests keep JS/Rust aligned |
| Desktop menu item not manually clicked | Green | Author noted; compile + unit tests pass |

---

PR #1032, opened 2026-08-14
Detect Claude Code alt-buffer per-window, not per-kind (CROW-1023)

### LINKED ISSUE #1023: Claude Code sessions have no client-side scrollback — no scrollbar, and a dead wheel whenever it isn't mouse-tracking
## Symptom

Scrolling back works on Cursor sessions and does not on Claude Code sessions. On a Claude Code tab there is no scrollbar at all — not a faint one, not a fading one — and the mouse wheel is a no-op unless Claude Code happens to have mouse reporting on at that instant.

This is the unrestored half of #1009. #1014 fixed it for inline agents only.

## What actually happens

`applySurfaceScrollback()` pins `term.options.scrollback = 0` for `agent_surface && uses_alternate_screen`, which today is Claude Code alone (`ClaudeCodeAgent.usesAlternateScreen = true`; Cursor, Codex, OpenCode, Grok, Antigravity are all `false`).

Three things follow, and the third is the one nobody planned:

1. **`baseY` is permanently 0.** With `scrollback: 0` the main buffer is capped at `rows`, so nothing ever scrolls off the top.
2. **The wheel depends on Claude Code's mode.** `appOwnsScroll()` returns `agent_surface && mouseTracking`. While Claude Code is mouse-tracking the tick is forwarded as SGR `\x1b[<64/65` and Claude Code scrolls its own transcript — that's the ADR-0013 design and it works. When it *isn't* mouse-tracking, the else branch is `term.scrollLines()` against a zero-length buffer: a silent no-op. So the wheel works or doesn't depending on what the TUI is doing at the time, which reads as "scrolling is broken."
3. **The gutter disappears.** FitAddon computes the reserved scrollbar width as `scrollback === 0 ? 0 : 14`. A zeroed tab reserves nothing, so there is physically nowhere to draw a bar. CROW-1020 restored a visible, grabbable xterm scrollbar (#1021), and it cannot appear here for this reason.

## How we got here

| When | | |
|---|---|---|
| 2026-07-02 | #556 | xterm.js 6.0.0 lands; its slider is a VS Code scrollable element, `visibility: Auto`, colour = foreground @ 0.20 (1.67:1) |
| 2026-07-23 17:20 | #826 | ADR-0013: agent windows get tmux `alternate-screen on`. Client still keeps 50k, so the gutter and thumb survive |
| 2026-08-13 21:27 | #1009 | `const wanted = activeSurfaceIsAgent() ? 0 : UNIFIED_SCROLLBACK` — caps **every** agent surface client-side. Terminal scrollbar disappears on all agent tabs |
| 2026-08-13 22:22 | #1014 | Narrowed to `agent_surface && uses_alternate_screen`. Cursor restored; Claude Code deliberately not |
| 2026-08-14 09:36 | #1021 | CROW-1020 restores the terminal bar for surfaces that have scrollback — which excludes Claude Code |

Worth recording because it caused a misdiagnosis: #1016 was filed inside that 55-minute window, correctly reporting a terminal regression, and attributed it to CROW-593 (2026-07-19) because that commit contains a literal `scrollbar-width: none`. It does — on `#sidebar`/`#board`, a different pane. CROW-593 never touched `xterm.css`. #1019 then fixed the sidebar, and CROW-1020 had to move it back.

## Why this isn't just "revert #1009 for Claude Code too"

The alt buffer is load-bearing for Claude Code specifically. #822 / #1008 are the history: Claude Code issues `smcup` and repaints full frames, and without a scrollback-less buffer each repaint fossilizes its predecessor — scrolling up walks stacked duplicate TUI frames rather than a transcript. That is exactly what ADR-0013's `alternate-screen on` exists to prevent, and it is why #1014 restored inline agents (whose transcript *is* clean native scrollback) and left Claude Code alone.

So the ask is not "give Claude Code 50k too." It is: **decide whether a Claude Code tab should have any client-reachable history, and if so, where it comes from.**

## Options to weigh

1. **Do nothing; make the design legible instead.** Claude Code owns its own scrollback and the wheel is forwarded to it. Fix the *edge*: when the surface is an alt-buffer agent and it is not mouse-tracking, the wheel currently no-ops silently. Either forward anyway or surface why nothing happened. Cheapest, keeps ADR-0013 intact, but there is still no grabbable bar.
2. **Hydrate from tmux on demand**, the way CROW-934 does for shells (`select-window` → `capture-pane -pe -S -50000`). Blocked as written: `maybeHydrateScrollback` explicitly skips agent surfaces, and a pane in the alt buffer has no tmux history to capture (see `ScrollbackHealthTests`). Would need history captured somewhere before the alt-screen switch, or a second non-alt pane.
3. **Give Claude Code windows a main-buffer transcript** and reintroduce dedup/sediment control by another means. Reopens #1008 directly and needs a real answer for full-frame repaints.
4. **Bounded local scrollback** — e.g. a few thousand lines instead of 0 — accepting some sediment in exchange for a working wheel and a visible bar. Needs measurement: how bad is the sediment at N lines, and is the trade legible to a user scrolling up?

## What "done" looks like

A decision recorded against ADR-0013 (it already carries three surface classes after #1014) plus whichever of these applies:

- Scrolling back on a Claude Code tab either works with a grabbable scrollbar, or predictably does nothing with the reason visible — not "works sometimes, depending on the TUI's mouse mode."
- Cursor and plain shells are unchanged: unified 50k, CROW-606 replay, CROW-934 hydrate-on-top, CROW-1020's bar.
- No return of #1008's duplicate-frame sediment.
- If the answer is option 1, that is a legitimate close — but say so in the ADR, because this is the second ticket to rediscover the same gap.

## Pointers

- `applySurfaceScrollback`, `appOwnsScroll`, `sendScrollToPTY`, `updateTerminalScrollbar` — `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js`
- `CodingAgent.usesAlternateScreen` — `Packages/CrowClaude/Sources/CrowClaude/ClaudeCodeAgent.swift` (`true`), `Packages/CrowCursor/Sources/CrowCursor/CursorAgent.swift` (`false`)
- `registerTerminal(agentSurface:usesAlternateScreen:)` — `Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift`
- The client-buffer trap — `Packages/CrowTerminal/Sources/CrowTerminal/Resources/crow-tmux.conf`, the `terminal-overrides` block
- ADR-0013; related: #822, #824, #850, #934, #1008, #1010, #1016, #1020

🐦‍⬛ Filed from the CROW-1020 follow-up

dhilgaertner:
> ⚠️ **needs-decision-review — do not merge before @dhilgaertner signs off.**
> This is the design-decision ticket #1023. The *direction* (Option 5 + sticky latch) was signed off; this PR is the implementation for final human review. One data item is still open (see Open items).

Closes #1023

## The bug

Two Claude Code builds behave differently in a Crow tmux pane:

| Build | `alternate_on` | `mouse_any` | Renders | Scroll before this PR |
|---|:---:|:---:|---|---|
| `2.1.233` | 1 | 1 | alt buffer, full-frame repaint | **works** — SGR-forwarded wheel |
| the reported build | 0 | 0 | **inline** (main buffer), streams | **dead** — no scrollbar, wheel no-op |

`uses_alternate_screen` was a **static per-kind capability** (`ClaudeCodeAgent = true`), so an inline Claude was declared an alt-buffer agent: the client capped its xterm scrollback to 0 (nothing to wheel, no reserved gutter → no scrollbar) and never forwarded its non-mouse wheel. This is the unrestored half of #1009 — #1014 fixed it for inline agents *by kind*, but the kind is `claudeCode`, so an inline Claude was excluded.

## The decision (ADR-0013 Amendment #1023)

Weighed the ticket's four options + a fifth (the investigator's per-window-detect). **Option 5 wins** and subsumes 1/2:

- **#1 (make edge legible)** — "scrolling does nothing" is the *wrong* answer for an inline build with a real transcript.
- **#2 (hydrate)** — blocked for a true alt buffer; subsumed by #5 (an inline Claude inherits CROW-934 hydrate for free).
- **#3 (main-buffer transcript)** — reopens #1008: the alt-buffer build full-frame-repaints → 42 stacked frames (measured).
- **#4 (bounded scrollback)** — bounded #1008 sediment on the working build via the client-cap fossilization path.
- **#5 (detect per-window)** — `uses_alternate_screen` becomes a per-*window* fact from the runtime `#{alternate_on}` the daemon already reads, latched sticky. Alt-buffer build unchanged (capped-0, forwarded wheel, no sediment); inline build treated like Cursor (unified 50k + CROW-606 replay + CROW-934 hydrate + CROW-1020 grabbable bar).

## Implementation

- **TmuxBackend** — `updatedAltBufferLatch` (pure, unit-tested) + a per-index `observedAltBufferWindows` latch (union-then-prune each read; cleared at (re)register/kill so a recycled index can't inherit a stale observation). `windowScrollbackClassification` returns a third `altBuffer` set. No extra tmux round-trip.
- **EngineRouter** — `uses_alternate_screen` = detected `altBuffer` membership, with the `AgentRegistry` capability as the pre-window / read-failure prior (mirrors the `agent_surface` fallback beside it).
- **Client** — no rewiring; `applySurfaceScrollback` / `fireScrollbackCapture` already key on `agent_surface && uses_alternate_screen`. One consistency fix: `verifyScrollbackAfterAttach` moves onto the same axis so inline agents also get the CROW-1027 post-attach heal.
- **CodingAgent / ClaudeCodeAgent** — capability demoted to the pre-window prior (docs only).

**Sticky latch** keeps the `2.1.233` build byte-for-byte: it latches to capped-0 on its first frame and stays there through transient main-buffer drops (shell-out, exit) — no flip-flop, no new sediment path. Only builds that *never* reach `alt_on=1` get 50k. Accepted transient: a starting alt-buffer tab shows a gutter for ≤1 poll before it latches (history empty then → cosmetic).

## Residual risk

The inline **Claude** build could not be reproduced on the dev machine (it runs `2.1.233` = alt-buffer). The clean-transcript evidence is **Cursor** (`alt_on=0`, clean 628-line transcript, live). If the inline Claude full-frame-repaints instead of streaming inline, 50k could show some #1008 duplication — but even then a working-but-slightly-duplicated bar strictly beats a dead wheel, and detection still routes correctly. **Verify on the inline-build machine before merge** (`capture-pane -pe -S -2000`, eyeball for stacked frames).

## Tests

- `ScrollbackHealthTests` — 6 new pure latch tests (records / **sticky** / inline-never-latches / prune-killed / retain-still-living).
- `web-tests/terminal-reload.test.js` — CROW-1027/1023: alt-buffer agent skips the post-attach heal, inline agent heals like a shell.
- Green: CrowTerminal (120), CrowEngine (764), full web-test suite. (`row.test.js`'s 10 glyph-rendering failures are pre-existing — unrelated to this change, `row` is excluded from `test:ci`.)

## Open items for @dhilgaertner

1. **Record the inline build's `claude --version`** in ADR-0013 (dev machine is `2.1.233`; the ADR has a `version TODO` placeholder). If a *newer* Claude went inline, this hits every user on update.
2. Confirm the inline Claude transcript is clean (verification command above) before merge.


dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- No new auth surface, no user-controlled strings in tmux formats, no secrets.
- Latch prune plus the register-time index clear keep a recycled tmux window index from inheriting capped-0 (which would disable scrollback).
- On a successful read, an unobserved window is treated as inline (wheel works) rather than capped to 0.

**Concerns:**
- None of security severity. This change is scroll-model routing only.

### Code Quality

**Yellow — `uses_alternate_screen` never reaches the client after the first `list-terminals` fetch.**

ADR-0013 Amendment #1023 (and the PR body) say the alt-buffer Claude build is unchanged except a ≤1-poll cosmetic gutter, because `refreshTerminals` rebinds `activeTerminal` and `applySurfaceScrollback` re-applies the cap.

`refreshTerminals` is not polled. `onServerChanged` in `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js:227` refreshes sessions/live/boards/artifacts — not terminals. `list-terminals` runs only on session select, add/close/recreate, and handoff.

Managed agent launch is deferred until `.shellReady` (`Packages/CrowEngine/Sources/CrowEngine/EngineRouter.swift`). The `refreshTerminals` that follows `new-terminal` / `recreate-terminal` therefore snapshots `uses_alternate_screen: false` (window exists, `alternate_on` still 0). Switching tabs in the same session reuses that stale `terminals[]` row (`switchTerminal` at `app.js:3745`). The server latch in `Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift:801` can flip on a later `list-terminals`, but the client never asks again.

For the common `2.1.233` alt-buffer build that is a regression of #822: xterm stays at `UNIFIED_SCROLLBACK` (50000) and live attach bytes fossilize in the local viewport — the reason `applySurfaceScrollback` (`app.js:5367`) caps to 0. The inline build (this ticket) happens to work, because `false` is the correct steady state.

The client must re-read the flag until it latches (or a short timeout) for `agent_surface && !uses_alternate_screen`. Cheapest fit: call `refreshTerminals()` from `onServerChanged` / `refreshLive` when the selected tab is an unlatched agent surface. Do not wait for a tab switch.

**Green — `destroyTerminal` does not clear `observedAltBufferWindows`.** `TmuxBackend.killWindow` does (`TmuxBackend.swift:923`); close-tab goes through `controller?.killWindow` (`destroyTerminal` at `TmuxBackend.swift:681`). Prune + `registerTerminal`'s `remove` (`TmuxBackend.swift:343`) cover recycled indices. Consider routing destroy through `killWindow` so the ADR's "cleared at kill" holds on every path.

**Green — accepted residual risk** on inline Claude transcript cleanliness (`docs/adr/0013-terminal-scroll-model.md`) is documented and is the author's call; not a blocker.

**Green —** `classificationReportsFailureDistinctlyFromAnEmptyResult` in `Packages/CrowTerminal/Tests/CrowTerminalTests/ScrollbackHealthTests.swift` still does not assert `altBufferWindowIndices()` fail-open, and its message still says `([], [])` after the return became a triple.

Tests run here: `ScrollbackHealthTests` 14/14 pass; `terminal-reload.test.js` 66/66 pass, including the new CROW-1027/1023 alt-vs-inline heal cases.

### Summary Table

---

dhilgaertner:
Thanks @dgershman — addressed in `3c5f1e54`.

**Yellow (the #822 regression) — fixed.** You're right: `list-terminals` isn't polled, so the stale `uses_alternate_screen: false` snapshot never flipped and an alt-buffer `2.1.233` build would have sat at `UNIFIED_SCROLLBACK` and fossilized live frames. Added `maybePollAltScreenLatch` (`app.js`): while the **attached** surface is an agent that hasn't latched, it re-reads `list-terminals` on a bounded schedule (~1s × 15) until it latches (`applySurfaceScrollback` caps to 0, and xterm discards any frames that fossilized in the meantime) or the budget is spent — at which point it's a genuinely inline build that correctly keeps the 50k. Per-surface budget, self-arms from every `refreshTerminals`; not a tab-switch dependency. Called right after `applySurfaceScrollback` in `refreshTerminals`.

**Green — `destroyTerminal`:** routed the close-tab path's `controller?.killWindow` through the latch clear too, so the ADR's "cleared at kill" holds on every teardown path (prune + register-remove already covered recycled indices; this is the consistency you flagged).

**Green — failure test:** now asserts `altBufferWindowIndices().isEmpty` on the failed-read path, and the message is corrected to `([], [], [])` (the return became a triple).

**Green — residual inline-transcript risk:** left as the documented author's-call in the ADR; still needs verification on the inline-build machine before merge (noted in the PR body).

New coverage: `web-tests/alt-latch-poll.test.js` (12 cases — arms only for an unlatched agent surface, latches then stops, bounded budget with no runaway, budget refresh on tab switch). Also corrected the ADR-0013 amendment, which previously implied a poll that didn't exist — that was the root of your finding.

Re-requesting review.

dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- No new auth surface, secrets, or user-controlled strings in tmux formats.
- Latch prune plus the register/kill/`destroyTerminal` clears still keep a recycled tmux index from inheriting capped-0.
- The client re-poll is bounded (~1s × 15), attached-surface-only, and stops once latched — no unbounded RPC loop.

**Concerns:**
- None of security severity. This remains scroll-model routing.

### Code Quality

**Yellow — `switchTerminal` never arms or disarms `maybePollAltScreenLatch`.**

The follow-up correctly starts a bounded `list-terminals` re-read from `refreshTerminals` (`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js`), which fixes the stay-on-the-tab new-terminal / session-select path that the last review flagged.

`switchTerminal` still only rebinds `activeTerminal` and calls `applySurfaceScrollback` — it does not call `maybePollAltScreenLatch`. Two holes, both the original #822 fossilization on the common `2.1.233` alt-buffer build:

1. Switch away from an unlatched agent to a shell (or already-latched) tab. The pending tick fires, sees a non-unlatched surface, and clears `altLatchPollTimer` / `altLatchPollTermId`. Switch back: `switchTerminal` reapplies the stale `uses_alternate_screen: false` row and never re-arms. Adding a shell while Claude is starting is the everyday version of this.
2. `refreshTerminals` lands on a non-agent `terminals[0]`, then the user clicks the Claude tab. No poll is running, and `switchTerminal` does not start one.

The new `Packages/CrowDaemon/web-tests/alt-latch-poll.test.js` case "switching to a different unlatched agent tab refreshes the budget" already models the intended contract — it calls `maybePollAltScreenLatch()` after changing `activeTerminal`. `switchTerminal` does not. One call after `applySurfaceScrollback()` closes both holes: leave disarms immediately, and return to the same unlatched id gets a fresh budget because disarm nulls `altLatchPollTermId`.

**Green — previous round, addressed.** `destroyTerminal` now clears the latch; `classificationReportsFailureDistinctlyFromAnEmptyResult` asserts `altBufferWindowIndices()` fail-open and the message is `([], [], [])`.

**Green — accepted residual risk** on inline Claude transcript cleanliness (`docs/adr/0013-terminal-scroll-model.md`) remains documented; still the author's call. Verify on the inline-build machine before merge.

Tests run here: `ScrollbackHealthTests` 14/14; `alt-latch-poll.test.js` 12/12; `terminal-reload.test.js` 66/66.

### Summary Table

---

dhilgaertner:
Thanks @dgershman — fixed in `fb974bab`.

**Yellow (`switchTerminal` never (dis)armed the poll) — fixed.** You're right, and both holes are the same #822 fossilization on the `2.1.233` build. Added `maybePollAltScreenLatch()` to `switchTerminal` right after `applySurfaceScrollback()` — the tab-switch path is now the same "bind → apply → (dis)arm" as `refreshTerminals`. Leaving an unlatched agent disarms immediately; returning to a still-unlatched tab gets a fresh budget because disarm nulls `altLatchPollTermId`. That closes both:
- switch-away-then-back (opening a shell while Claude launches), and
- `refreshTerminals` landing on a non-agent `terminals[0]` then clicking the Claude tab.

**Test:** promoted the contract from "call maybePoll after changing activeTerminal" to driving the **real** `switchTerminal` — 4 new `alt-latch-poll.test.js` cases: arms on switch-to-unlatched-agent, disarms on switch-to-shell, re-arms on switch-back-to-same-unlatched, no-arm on switch-to-latched (16/16). Full web-tests green (switcher/terminal-reload/router unaffected).

The two prior-round Greens remain resolved; the inline-transcript residual risk is still the documented pre-merge verification item. Re-requesting review.

dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- No new auth surface, secrets, or user-controlled strings in tmux formats.
- Latch prune plus the register / `killWindow` / `destroyTerminal` index clears in `Packages/CrowTerminal/Sources/CrowTerminal/TmuxBackend.swift` still keep a recycled tmux index from inheriting capped-0 (which would disable scrollback).
- The client re-poll is bounded (~1s × 15), attached-surface-only, and stops once latched — no unbounded RPC loop.

**Concerns:**
- None of security severity. This remains scroll-model routing.

### Code Quality
- **Previous Yellows, addressed.** `maybePollAltScreenLatch` now self-arms from `refreshTerminals` (`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js`) and `(dis)arm`s from `switchTerminal` on the same bind → apply → poll sequence. Leave-then-return to an unlatched agent gets a fresh budget; switching to a shell or already-latched tab disarms. Covered by `Packages/CrowDaemon/web-tests/alt-latch-poll.test.js` (16/16), including the real `switchTerminal` path.
- **Green — accepted residual risk** on inline Claude transcript cleanliness (`docs/adr/0013-terminal-scroll-model.md`) remains documented; still the author's call. Verify on the inline-build machine before merge.
- **Green — 15s poll vs 30s `.shellReady`.** `ALT_LATCH_POLL_MAX = 15` can expire before a slow shell reports ready (the 30s readiness budget in `TmuxBackend`). A Manager has no tab switch to re-arm. Typical launch is well under 15s, and leaving/re-entering the session re-reads `list-terminals`, so this is not a defect in the written latch — consider tying the budget to readiness or bumping it if cold-start hydrate proves slow in the field.

Tests run here: `ScrollbackHealthTests` 14/14; `alt-latch-poll.test.js` 16/16; `terminal-reload.test.js` 66/66 (including the CROW-1027/1023 alt-vs-inline heal cases).

### Summary Table

---

PR #1036, opened 2026-08-15
Keep agent session switches off the full terminal reload

### LINKED ISSUE #1035: Switching to a Claude session jumps the cursor below the input box
When switching to a **Claude Code** session in Crow, the terminal display jumps and the typing cursor ends up at the **bottom of the screen, outside the agent input box**.

**Cursor** is similar but milder: the view jumps down about one line and text **below the input box doubles** (stacked/repeated TUI chrome).

Does not feel like a tmux.conf drift (stock Crow conf matches another machine). Looks like a client attach / `select-window` / scrollback-replay / alt-screen layout glitch on tab switch.

## Repro
1. Have a Claude Code session with the TUI at the idle prompt.
2. Switch away, then switch back to that session (sidebar or session switcher).
3. The surface jumps; the caret is below the input box. Typing does not land in the prompt field.
4. Repeat with a Cursor session: smaller jump (~1 line) and doubled text under the input box.

## Acceptance criteria
- Switching to a Claude session leaves the cursor in the input box; no jump below the TUI.
- Cursor session switch does not double the chrome below the input box or shift the viewport a line.
- Plain shell tabs are unchanged.
- If the cause is attach replay / alt-screen / jump-to-bottom, fix that path rather than papering over with a full reload.

Related: ADR-0013 (per-surface scroll / alt-screen), CROW-677 (terminal skeleton on attach), jump-to-bottom (#668).

dgershman:
Closes #1035

Switching to a Claude Code session parked the caret below the input box; Cursor jumped about a line and stacked the footer chrome. Both were the #673 full reload on tab/session switch: a new PTY at tmux's default 24×80, a SIGWINCH to the real size, then a `capture-pane` replay injected into the same stream as the live attach redraw.

**Fix the attach/replay path rather than adding another reload:**

- Agent surfaces (Claude, Cursor, Manager) now switch in place: `term.reset()` + `select-window` on the live socket. No new PTY, no same-size SIGWINCH.
- The daemon skips the CROW-606 replay when `#{alternate_on}=1`. Capture-pane of an alt-buffer pane is only the current frame, and a second paint of that frame is what left the caret below the TUI.
- Plain shell tabs still take the #673 reload. Explicit Reload is unchanged.

## Test plan
1. Claude Code session at the idle prompt: switch away (sidebar or session switcher), switch back. Cursor stays in the input box; no jump below the TUI. Typing lands in the prompt.
2. Cursor session: switch away and back. No doubled chrome under the input box, no one-line viewport shift.
3. Plain shell tab: switch away and back. Unchanged (still the full reload recovery).
4. Re-click the already-active tab / a same-session background refresh: no reload or flash.
5. Header / right-click Reload still recovers a garbled surface.


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- No new attack surface: `/terminal` WebSocket auth, origin guard, and connection limiter are unchanged.
- Server-side replay gating uses the server-issued tmux group name plus the existing `select-window` index — no new client-controlled shell/tmux injection paths.
- `paneIsOnAlternateScreen` fails open to replay (not silent history drop), which is the safer default for non-alt-buffer panes.

**Concerns:**
- None blocking. The intentional tradeoff — agent tabs no longer take the #673 full reload and therefore won't self-heal a grid another client reshaped — is documented in `docs/adr/0013-terminal-scroll-model.md` and scoped correctly to agent surfaces while shells keep `reloadTerminal()`.

### Code Quality
- Root cause is addressed at both layers: client routes agent tab/session switches through `switchAgentWindow` (`term.reset` + live `select-window`, no PTY teardown), and daemon skips CROW-606 replay when `#{alternate_on}=1` via `TerminalCockpit.shouldReplayScrollback`.
- `attachWindow` correctly rebinds only when the window changes and preserves the shell reload path for #673 grid-mismatch recovery.
- `switchAgentWindow` mirrors the reload path's hygiene (`applySurfaceScrollback`, `resetScrollbackSync`, skeleton overlay) and keeps inline agents eligible for CROW-1027 heal while alt-buffer agents skip it — consistent with existing scroll model.
- Test coverage is strong across three layers: `TerminalReplayTests.altBufferPaneSkipsScrollbackReplay`, `WebTerminalAssetTests.attachWindowSwitchesAgentSurfacesInPlace` (drift guard), and `web-tests/session-switch-attach.test.js` (17 assertions). All passed locally; `swift build` succeeded.
- ADR-0013 and CHANGELOG accurately reflect the behavior change.

### Summary Table

| Finding | Severity | Location |
|---------|----------|----------|
| Agent surfaces trade #673 reload self-heal for in-place switch (documented, intentional) | Green | `app.js` `attachWindow` / ADR-0013 |
| Failed `alternate_on` probe still replays scrollback (conservative for shells/inline agents) | Green | `TerminalCockpit.paneIsOnAlternateScreen` |

---

PR #1040, opened 2026-08-15
Install all embedded corveil skills, not just query-corveil (CROW-1039)

### LINKED ISSUE #1039: Generalize Scaffolder.installCorveilSkill to install all embedded corveil skills, not just query-corveil
## Problem

Crow's launch-time corveil skill installer only ever installs **`query-corveil`**. Every other skill embedded in the `corveil` binary — currently `create-corveil-workflow` (shipped via corveil/corveil#2226, ticket corveil/corveil#2217), `crow-runner`, and `worker-runner` — is never installed, so it never becomes an available `/slash-command` in Manager/session Claude Code, even though it's sitting right there in the binary. `corveil skill list` shows all four; only one reaches `.claude/commands/`.

## Root cause

`Scaffolder.installCorveilSkill(_:)` hardcodes a single skill in two places:

- `Packages/CrowEngine/Sources/CrowEngine/Scaffolder.swift:326` builds `["skill", "install", "--path", target]` with **no `--skill` flag**, so `corveil skill install` falls back to its own default — which is `query-corveil`.
- `target` comes from `CorveilCLI.skillPath(devRoot:)` (`Packages/CrowEngine/Sources/CrowEngine/CorveilCLI.swift:142`), which is literally `{devRoot}/.claude/commands/query-corveil.md`.

So both the skill name and the destination filename are pinned to `query-corveil`. `crow corveil reinstall-skill` (Settings → Corveil CLI button + the CLI verb) calls the same function, so it inherits the same limitation — clicking Reinstall just rewrites `query-corveil.md`.

This was correct when `query-corveil` was the only embedded skill; it was never generalized as more skills were added to the binary.

## Proposed change

Generalize `installCorveilSkill` to install the full set of embedded skills (or a configured subset):

1. **Enumerate, don't hardcode** — `corveil skill list --format json` returns the embedded set, so Crow doesn't need to know skill names ahead of time.
2. For each skill, run `corveil skill install --skill <name> --path {devRoot}/.claude/commands/<name>.md`.
3. **Warning aggregation** — the function returns a single `String?` today (→ `AppState.corveilSkillInstallWarning`). Decide whether one failing skill fails the whole install or is reported per-skill.
4. **Timeout budget** — today there's one `corveilInstallTimeout`-bounded process on the synchronous launch path (`CrowDaemon.run` → `LaunchScaffold.run`, before the Manager comes up). N installs means either a per-skill timeout or a combined budget; keep the guarantee that a hung binary can't stall startup.
5. **Path surface** — `CorveilCLI.skillPath(devRoot:)` returns a single path today and is named to the user by the Reinstall button (`skill_path`). Either return a directory + derive per-skill filenames, or return the set, and update that reporting surface.

Optional: a config field for *which* skills to install (an allowlist), if we don't want all of them by default. Default-install-all is simplest and matches "the binary ships what it wants installed."

## Acceptance

- On daemon launch (and on `crow corveil reinstall-skill`), all embedded corveil skills land in `{devRoot}/.claude/commands/` — verified by `create-corveil-workflow`, `crow-runner`, and `worker-runner` appearing alongside `query-corveil`.
- A missing / hung / non-executable binary still can't stall startup.
- The Reinstall button's report reflects the multi-skill result.

## Context

Surfaced 2026-08-15: the `create-corveil-workflow` skill (corveil/corveil#2217, shipped in the binary via corveil/corveil#2226) is embedded and listed by `corveil skill list` but never became available in Claude Code because of this gap. Manual workaround for a single skill today: `corveil skill install --skill create-corveil-workflow`.

dhilgaertner:
Closes #1039

## Problem

`Scaffolder.installCorveilSkill` hardcoded a single skill (`query-corveil`) in two places — it built `["skill", "install", "--path", target]` with **no `--skill` flag** (so `corveil skill install` fell back to its own `query-corveil` default), and `target` came from `CorveilCLI.skillPath(devRoot:)`, literally `.../query-corveil.md`. So the other embedded skills the binary ships — `create-corveil-workflow`, `crow-runner`, `worker-runner` — never reached `.claude/commands/` and never became `/slash-commands`, even though `corveil skill list` shows all four. `crow corveil reinstall-skill` shared the same function and the same limitation.

## Change

Generalize the installer to **enumerate** the binary's embedded skills (`corveil skill list`) and install each to `{devRoot}/.claude/commands/<name>.md` independently. This covers the ticket's five concerns:

- **Enumerate, don't hardcode** — `parseSkillNames` accepts one-name-per-line (what current corveil builds print, even under `--format json`) and a JSON array of strings/objects (a future build), keeping only well-formed slugs so a banner line or a path-traversal name can't become an install target or filename.
- **Warning aggregation** — per-skill best-effort: one skill failing doesn't abort the rest; failures aggregate into the single `String?` warning (→ `AppState.corveilSkillInstallWarning`), naming each failed skill with corveil's own diagnostic.
- **Timeout budget** — keeps the per-process 5s cap **and** adds a 10s overall `corveilInstallBudget` shared across enumerate + all installs, so N installs can't stall the synchronous launch path (`CrowDaemon.run` → `LaunchScaffold.run`). A hang on `list` (the common wedged case) still costs just one 5s cap.
- **Fallback** — if the binary can't enumerate (older build, no `list` subcommand, empty output), install the historical `query-corveil` default with **no `--skill`**, so a launch never regresses to installing fewer skills than before.
- **Path surface** — `CorveilCLI.skillPath(devRoot:)` (a single `query-corveil` file) is replaced by `commandsDir(devRoot:)` + `skillPath(devRoot:skill:)`. The Reinstall button / `corveil-reinstall-skill` RPC now report the commands **directory** as `skill_path`.

Default-install-all (no config allowlist) — simplest, and matches "the binary ships what it wants installed."

## Acceptance

- On daemon launch and on `crow corveil reinstall-skill`, every embedded skill lands in `{devRoot}/.claude/commands/` — verified by a test installing `create-corveil-workflow`, `crow-runner`, `worker-runner` alongside `query-corveil`.
- A missing / hung / non-executable / non-enumerable binary still can't stall startup (per-process cap + overall budget + `list`-timeout bail + fallback).
- The Reinstall report reflects the multi-skill result (directory `skill_path`, per-skill failure summary).

## Docs

Updated the `reinstall-skill` CLI abstract/discussion (regenerated `docs/cli.md`), the hand-written `docs/cli-reference.md`, and `CLAUDE.md`.

## Tests

- New `ScaffolderCorveilSkillInstallTests`: enumerate-and-install-each, fallback (list fails / empty output), per-skill failure aggregation, unconfigured/non-executable no-op, and `parseSkillNames` (plain lines, JSON array of strings, JSON array of objects, banner/blank filtering, path-traversal rejection).
- Updated `CorveilCLITests` and `CorveilRouteGatingTests` for the new multi-skill shape and `commandsDir`/`skillPath(devRoot:skill:)` surface.
- CrowEngine (49), CrowDaemon corveil/launch (11), and CrowCLI corveil/docs (11) suites pass; CLI parity + docs-freshness gates pass.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -136,7 +136,7 @@ crow corveil reinstall-skill [--path PATH]     → {"ok":bool,"message":"Skill r
 ```
 
 - **Branch on `ok`, not the exit code.** A corveil that is missing, non-executable, exits non-zero, or hangs past 5s is a successful *report* of a broken binary. A non-zero `crow` exit means the request never ran (no daemon, or no path configured anywhere).
-- `reinstall-skill` re-runs the launch-time `corveil skill install`, writing `{devRoot}/.claude/commands/query-corveil.md` — the "I just rebuilt corveil, pick up its new embedded skill" loop, no `crowd` restart. It is idempotent.
+- `reinstall-skill` re-runs the launch-time `corveil skill install` for **every** embedded skill (enumerated via `corveil skill list`), writing each into `{devRoot}/.claude/commands/<name>.md` — the "I just rebuilt corveil, pick up its new embedded skills" loop, no `crowd` restart. `skill_path` names that directory. It is idempotent; one skill failing doesn't abort the rest (the warning names any that didn't install).
 - A reinstall also updates the launch-time corveil warning: succeeding clears it, failing replaces it.
 - Both are **local-only** on `/rpc` (they execute a path on the daemon host), like `gateway`/`web-password`/`mcp token`. The CLI is unaffected — it goes over the Unix socket.
 
--- a/Packages/CrowCLI/Sources/CrowCLILib/Commands/CorveilCommands.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Commands/CorveilCommands.swift
@@ -67,16 +67,17 @@ public struct CorveilVerify: ParsableCommand {
 public struct CorveilReinstallSkill: ParsableCommand {
     public static let configuration = CommandConfiguration(
         commandName: "reinstall-skill",
-        abstract: "Reinstall the /query-corveil slash command from the corveil binary",
+        abstract: "Reinstall every embedded slash command from the corveil binary",
         discussion: """
         Re-runs the `corveil skill install` that `crowd` runs at launch, writing \
-        `{devRoot}/.claude/commands/query-corveil.md`. Use it after rebuilding \
-        corveil locally to pick up its new embedded skill without restarting \
-        the daemon.
+        every embedded skill the binary ships (`corveil skill list`) into \
+        `{devRoot}/.claude/commands/` — one `<skill>.md` per skill. Use it after \
+        rebuilding corveil locally to pick up its new embedded skills without \
+        restarting the daemon. `skill_path` in the response is that directory.
 
         A run also updates the launch-time corveil warning: succeeding clears \
-        it, failing replaces it, so there is one answer to "is corveil broken?" \
-        rather than a startup one and a button one.
+        it, a per-skill failure replaces it, so there is one answer to "is \
+        corveil broken?" rather than a startup one and a button one.
         """
     )
 
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CorveilRoutes.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CorveilRoutes.swift
@@ -80,7 +80,7 @@ enum CorveilRoutes {
                     appState.corveilSkillInstallWarning = outcome.ok ? nil : outcome.message
                 }
                 var payload = outcomeJSON(outcome)
-                payload["skill_path"] = CorveilCLI.skillPath(devRoot: devRoot)
+                payload["skill_path"] = CorveilCLI.commandsDir(devRoot: devRoot)
                 return json(payload)
 
             default:
--- a/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/RPCHandlers.swift
@@ -2181,8 +2181,8 @@ private func makeMCPTokenHandlers(devRoot: String) -> [String: CommandRouter.Han
 /// `open-in-vscode` / `open-terminal` are gated on the same argument.
 ///
 /// Neither writes config: `corveil-verify` runs `--version`, and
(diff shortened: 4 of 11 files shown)


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- Skill names are sanitized to `[a-z0-9_-]` before they become `--skill` values or `<name>.md` filenames (`Packages/CrowEngine/Sources/CrowEngine/Scaffolder.swift:489`), with tests covering `../`, `/etc/passwd`, banners, and blanks.
- Installs run via `Process.arguments`, not a shell. The binary path still requires `isExecutableFile`, and the HTTP/RPC doors remain local-only.
- Launch stall is bounded: 5s per subprocess plus a 10s overall budget; a hung `skill list` bails instead of spending the rest of the budget on installs that would also hang.
- Enumeration failure falls back to `query-corveil` with no `--skill`, so an older binary cannot regress below the pre-CROW-1039 one-skill install.

**Concerns:**
- **Green:** `runCorveil` now captures stdout with a `Pipe()` and drains it after `waitUntilExit` (`Packages/CrowEngine/Sources/CrowEngine/Scaffolder.swift:432`). The previous install path used `FileHandle.nullDevice` specifically to avoid a pipe-buffer deadlock if the child wrote >64KB. The new comment accepts that list/install output stays small, and the 5s watchdog still bounds a hang. Consider keeping `nullDevice` on the install invocations, which do not need stdout.

### Code Quality
- **Green:** `CLAUDE.md:135` still shows `"message":"Skill reinstalled"` while `CorveilCLI.reinstallSkill` now returns `"Skills reinstalled"`.
- **Green:** `Packages/CrowCLI/Sources/CrowCLILib/Commands/CorveilCommands.swift:20` parent abstract is still singular ("reinstall its skill") after the subcommand was generalized.
- **Green:** `timeoutWarning` is documented as reused for per-skill timeouts (`Packages/CrowEngine/Sources/CrowEngine/Scaffolder.swift:518`), but a per-skill timeout goes through the aggregated "failed for N skill(s): name (timed out)" string instead. List-timeout is the only caller.
- Tests are strong for the new behavior (enumerate-and-install-each, fallback, per-skill aggregation, `parseSkillNames` including path-traversal). `ScaffolderCorveilSkillInstallTests` (11) and `CorveilCLITests` (15) passed locally. Timeout/budget paths are untested, which is understandable given the 5s/10s waits.

### Summary Table

---

PR #1044, opened 2026-08-15
Re-arm the pane's mouse mode on in-place session switch (CROW-1043)

### LINKED ISSUE #1043: Claude Code session unscrollable after in-place session switch (alt-buffer replay skipped) — regression from #1036
## Symptom

After the recent terminal-scroll work, switching **between sessions** sometimes leaves a **Claude Code** session unscrollable (mouse wheel / scrollbar do nothing). Explicit **Reload** on that session restores scrolling — but switching away and back returns it to the unscrollable state. Non-agent shells and everything else scroll fine.

## Root cause

Regression from **#1036 ("Keep agent session switches off the full terminal reload")**. That change made agent surfaces (Claude / Cursor / Manager) switch **in place** instead of doing the #673 full reload:

1. Client (`web/app.js`) does `term.reset()` on the in-place switch — which **clears xterm's scrollback buffer**.
2. It then relies on the daemon's `select-window` reply to rebuild the buffer via the CROW-606 replay (`capture-pane -pe -S -50000`).
3. **But the daemon skips that replay when the pane is on the alternate screen.** The decision is kind-blind:

   ```swift
   // TerminalCockpit.swift:161-162
   static func shouldReplayScrollback(alternateOn: Bool) -> Bool { !alternateOn }
   ```
   called at `TerminalCockpit.swift:138`.

The skip was correct for a genuine alt-screen shell app (capture-pane of an alt buffer is only the current frame, and re-painting it parked the caret below the TUI — #1035). **The bug is the assumption that inline agents are always `alternate_on=0`.** They aren't: **Claude Code renders its TUI in the alternate-screen buffer**, per-window (which is exactly why CROW-1023 detects Claude's alt-buffer per-window rather than per-kind). Crow deliberately gives inline agents a *unified* scrollback via the CROW-606 replay (CROW-1016/1026).

So when you switch to a Claude session **while it is on the alt screen** (`alternate_on=1`):
`term.reset()` wipes xterm's scrollback → `shouldReplayScrollback` returns `false` → replay skipped → the pane has only the live current frame → **nothing to scroll.** Reload's recovery (or catching Claude on the normal buffer) restores it; the next switch reproduces it. The "sometimes" is Claude's buffer state at the instant you switch.

The test even encodes the wrong assumption:
```swift
// TerminalReplayTests.swift:48-51 — "Shells and inline agents (alternate_on=0) still replay."
#expect(TerminalCockpit.shouldReplayScrollback(alternateOn: true) == false)
```

## Why Reload works but switch-back doesn't

Explicit Reload / the self-heal (CROW-1027/1026) takes the full recovery path and rebuilds history. The in-place agent switch (#1036) deliberately does not — it resets and depends on a replay that gets skipped for alt-buffer panes. So Reload fixes it once; the next in-place switch wipes it again.

## Fix direction (don't re-introduce #1035)

The double-paint that #1035/#1036 fixed was the **current frame**, not the **history**. So the fix is to restore the scrollback *history* for inline agents without re-painting the visible frame:

1. **Thread the existing `agent_surface` signal into the replay decision.** `web/app.js` already gets an `agent_surface` flag from `list-terminals` (see `web/terminal.html:105-109`, ADR-0013). Make the replay decision `shouldReplayScrollback(alternateOn:isAgentSurface:)` return `true` for agent surfaces even when `alternate_on=1` — their unified scrollback is the whole point. Update `TerminalReplayTests.swift`.
2. **Avoid the caret/double-paint regression:** for an alt-buffer *agent* pane, replay the history region **above** the visible screen but let the live attach paint the current frame — e.g. capture-pane the scrollback excluding the last `pane_height` rows — so scrollback is restored with no second paint of the frame that parked the caret below the TUI.
   - *Alternative:* on the in-place agent switch, stop doing a full `term.reset()` (which drops scrollback); reset modes/cells only and preserve xterm's existing scrollback across the same-window switch. Simpler, but verify it doesn't leak stale content when the pane was reshaped while hidden (the original reason `term.reset()` was added).

Pick one; option 1+2 is the more faithful fix. Update **ADR 0013 (terminal-scroll-model)** to record that inline/agent surfaces keep their unified scrollback across in-place switches regardless of `alternate_on`.

## Repro
1. Open a Claude Code session with enough conversation to scroll.
2. While Claude's TUI is up, switch to another session (sidebar or `cmd+/` switcher), then switch back.
3. Try to scroll the Claude session → wheel/scrollbar dead. Press Reload → scroll works. Switch away and back → dead again.

## Key files
- `Packages/CrowDaemon/Sources/CrowDaemon/TerminalCockpit.swift` (`shouldReplayScrollback` :161-162, callsite :138, `paneIsOnAlternateScreen` :147-152)
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js` (in-place agent switch: `term.reset()` + `select-window`; #1036)
- `Packages/CrowDaemon/Sources/CrowDaemon/TerminalWebSocket.swift` (select-window reply / replay wiring)
- `Packages/CrowDaemon/Tests/CrowDaemonTests/TerminalReplayTests.swift` (:48-51 — assertion to revise)
- `Packages/CrowDaemon/web-tests/session-switch-attach.test.js` (add an alt-buffer agent case)
- `docs/adr/0013-terminal-scroll-model.md`

## Context
Regression introduced by #1036 (closes #1035), interacting with the inline-agent unified-scrollback model (CROW-1016/1023/1026). Diagnosed on `origin/main` 2026-08-15.

dhilgaertner:
Closes #1043

## What was happening

A Claude Code (or Manager) session became **unscrollable after switching away and back**: the mouse wheel did nothing, an explicit **Reload** restored it, and the next switch killed it again.

## Root cause — the mouse-tracking mode, not the CROW-606 replay

The ticket diagnosed this as the daemon skipping the CROW-606 scrollback replay for alt-buffer panes. I reproduced it in the running app and the actual cause is different — and the ticket's prescribed fix would not have helped this build (and its `-E -1` variant risked blanking the frame). Details:

- #1035 made an agent tab/session switch happen **in place** (`term.reset()` then `select-window` on the live socket) instead of the #673 full reload.
- `term.reset()` clears xterm's `mouseTrackingMode`. For an **alt-buffer** agent that mode is the *only* scroll path: its xterm `scrollback` is capped to 0 (no local buffer, no bar), so the wheel works solely because `appOwnsScroll()` is `agent_surface && mouseTracking` and the tick is forwarded to Claude as an SGR wheel button (#850).
- The `select-window` redraw does **not** put the mode back. tmux emits mouse-mode changes to a client as **deltas**, so switching between two agent windows that share the same mouse state (Claude → Claude, or the Manager; both `mouse_any=1 mouse_sgr=1`) repaints the frame but **no mouse DECSET** — from tmux's view the client already has that mode, while the client just cleared it.
- **Reload** recovers because it is a *fresh* grouped attach where tmux re-emits the full initial state (`?1000h`/`?1006h`); the next in-place switch clears it again.

### Reproduced live (build 2.1.233)

Inspecting the running web client's state across one round-trip switch between two Claude sessions:

| | `mouseTrackingMode` | `appOwnsScroll()` |
|---|---|---|
| fresh attach | `any` | `true` |
| after switch away & back | **`none`** | **`false`** |

Writing the pane's mouse DECSETs back to xterm restored `mouseTrackingMode: any` / `appOwnsScroll: true` — confirming the fix mechanism end-to-end (through the real `swallowMouseMode` parser path).

Every live Claude window here is `alternate_on=1, history_size=0`, so the CROW-606 replay had **nothing to restore** — the dead wheel was the mouse mode.

## The fix

The daemon re-arms the pane's **actual** mouse mode on every `select-window`. `TerminalWebSocket` reads `#{mouse_standard_flag}#{mouse_button_flag}#{mouse_any_flag}#{mouse_utf8_flag}#{mouse_sgr_flag}` and yields the matching DECSET enables (`TerminalCockpit.mouseModeReArmSequence` → `?1003h?1006h` for a Claude pane) into the same stream as the replay, so the client's mode matches the pane deterministically, independent of tmux's per-client delta.

It is:
- **inert where it isn't needed** — a shell (or an agent idling without tracking) reports no mouse flag, so nothing is emitted; on a non-agent surface `swallowMouseMode` drops DECSET 1000–1016 anyway;
- **idempotent** — re-emitted on the connect select and the CROW-934/1027 hydrate/heal re-selects with no effect beyond re-asserting the same mode;
- **visually inert** — it sets an input mode and paints nothing, so unlike a scrollback replay it can't race the live frame or blank the pane.

This is deliberately **orthogonal to the CROW-606 replay**, which stays skipped for an alt-buffer pane (`shouldReplayScrollback` unchanged and still correct — `history_size=0`).

## Changes

- `TerminalCockpit`: `mouseModeReArmData` (tmux read) + pure `mouseModeReArmSequence` (DECSET builder).
- `TerminalWebSocket`: emit the re-arm before the replay on `select-window`.
- `TerminalReplayTests`: DECSET-builder truth table (the CROW-1035 `shouldReplayScrollback` skip is unchanged).
- ADR 0013: `#1043` amendment + footer.
- CHANGELOG entry.

## Testing

- `swift test --package-path Packages/CrowDaemon --filter TerminalReplayTests` — all pass, including the new `mouseModeReArm*` cases.
- Full `CrowDaemon` suite: 430/432 pass. The **2 failures are pre-existing on the base branch** and unrelated to this change (verified by stashing): `RPCLanePolicyTests` (`corveil-verify`/`corveil-reinstall-skill` lane, CROW-1011) and `WebNotificationCenterTests.bootCatchResetsTheHistory` (an `app.js` catch-pattern check). This branch is 1 commit behind `origin/main`.
- Live reproduction + fix-mechanism confirmation in the running app (table above).


dgershman (approved):
## Code & Security Review

### Critical Issues
None.

### Security Review
**Strengths:**
- `mouseModeReArmSequence` never interpolates tmux output into CSI. It emits a fixed table of DECSET enables (`1000`/`1002`/`1003`/`1005`/`1006`) only when a flag character is exactly `1`. A failed or wrong-length `display-message` returns `nil` rather than a truncated escape.
- The tmux read goes through `TmuxController.run` argv (no shell) with a constant format string. The window index is the existing `Int` from the `select-window` control frame.
- Re-arm bytes ride the already-authenticated `/terminal` PTY→WebSocket stream (same path as CROW-606 replay). On a non-agent surface `swallowMouseMode` still drops DECSET 1000–1016.

**Concerns:**
- None. Re-enabling mouse tracking on a pane that already requested it is the intended #850 SGR-wheel path, not a new input surface.

### Code Quality
The diagnosis is stronger than the ticket's. #1043 blamed the CROW-606 alt-buffer replay skip, but the live round-trip (`mouseTrackingMode: any → none`, `appOwnsScroll(): true → false`) plus restoring the pane's DECSETs through the real parser path shows the dead wheel is `term.reset()` versus tmux's mouse-mode deltas. Leaving `shouldReplayScrollback` unchanged is the right call — replaying an alt-buffer capture is the CROW-1035 caret jump.

Wiring matches that diagnosis: after `select-window`, `TerminalWebSocket.swift:109` yields the re-arm *before* the replay, so an alt-buffer Claude (replay skipped) still gets `?1003h?1006h`. Enables-only is sufficient because `switchAgentWindow` already resets xterm. The builder truth table covers the Claude shape (`00101`), empty-mode `nil`, and length rejects. `swift test --package-path Packages/CrowDaemon --filter TerminalReplayTests` — 10/10 passed, including the three new `mouseModeReArm*` cases.

- **Green** `Packages/CrowDaemon/Sources/CrowDaemon/TerminalCockpit.swift:196` — `mouseModeReArmSequence` accepts any 5-character string and treats non-`1` as off, so `"xx1xx"` would emit `?1003h`. The comment says any other shape yields `nil`; requiring each character ∈ `{0,1}` would match that contract. Realistic tmux output is already `0`/`1`, so this is defense-in-depth.

### Summary Table

**Recommendation:** Approve — driven by 0 Red, 0 Yellow, 1 Green findings.

---

Issue #1048, opened 2026-08-16
Cursor still paints duplicate TUI chrome (sediment back after #1014)


dgershman:
Cursor sessions still show **stacked duplicate lines** — the same TUI chrome appearing twice (or more) in the live view / just below the input box. Seen in this session after ordinary tool output, not only when scrolling history.

#1008 / #1009 tried to kill this without alt-screen. #1014 restored the wheel by keeping unified 50k history on inline agents (`cursor-agent` never issues smcup). That trade-off put the sediment back.

## Live example (this Crow session, cwd `~/Projects/devroot`)

After `gh pr close 1022`, the Cursor footer painted twice:

```
    Tip: Use /config to customize Cursor settings and behavior.
    Tip: Use /config to customize Cursor settings and behavior.
  → Add a follow-up
  Cursor Grok 4.6 High Fast · 78.8%
  ~/Projects/devroot
  ~/Projects/devroot
```

Same class as spike #822 / ADR-0013: full-frame repaint into a scrolling main buffer deposits the previous frame. Also matches the milder “doubling below the input box” noted on #1035.

## Acceptance criteria
- Cursor idle/prompt chrome (tips, model line, cwd) appears **once**, not stacked.
- Wheel still scrolls Cursor history (do not regress #1010/#1014 by capping scrollback to 0 again).
- Claude alt-buffer path unchanged.

Related: #1008, #1009, #1010, #1014, #1035, #1047, ADR-0013.

[🐦‍⬛ Created with Crow via Cursor](https://github.com/corveil/crow)

PR #1051, opened 2026-08-16
Fix agent session switch cursor jump regression (CROW-1035)

### LINKED ISSUE #1035: Switching to a Claude session jumps the cursor below the input box

dgershman:
Closes #1035

## Summary

The #1036 in-place agent switch fixed the full-reload SIGWINCH/replay race, but two gaps remained and caused the reopened regression:

- **`term.reset()` before `select-window`** cleared parser state (including mouse mode) that tmux re-sends only as deltas — same class of bug as CROW-1043. For alt-buffer Claude, the live redraw after reset parked the caret below the TUI input box.
- **CROW-606 replay still ran on inline agents** (Cursor) during in-place switches. #1048 stopped mid-session hydrate from stacking chrome, but switch-time replay still raced the live attach.

This change:
1. Clears the shared xterm buffer with `ESC[H ESC[2J ESC[3J` (`clearTermBuffer`) instead of `term.reset()` on in-place agent switches.
2. Sends `select-window` with `replay: false` from `switchAgentWindow`; connect/reload/heal paths keep the default replay behavior.

Plain shell tabs still take the #673 full reload. Explicit Reload is unchanged.

## Test plan

- [ ] Claude Code at idle prompt: switch away (sidebar or session switcher), switch back. Cursor stays in the input box; typing lands in the prompt.
- [ ] Cursor session: switch away and back. No doubled chrome under the input box, no one-line viewport shift.
- [ ] Plain shell tab: switch away and back. Still uses full reload (unchanged).
- [ ] Re-click active tab / same-session background refresh: no reload or flash.
- [ ] Header / right-click Reload still recovers a garbled surface.
- [ ] Claude alt-buffer wheel still forwards while mouse-tracking after a round-trip switch (CROW-1043).


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- The new `replay` flag is optional and defaults to the existing connect/reload behavior (`control.replay != false`); in-place agent switches are the only caller that sets `false`, so there is no broad change to scrollback restoration for shells or first attach.
- `clearTermBuffer()` writes a fixed ANSI sequence (`ESC[H ESC[2J ESC[3J`) locally — no user-controlled input on that path.
- Mouse re-arm (CROW-1043) still runs on every `select-window`, including `replay: false`, preserving the wheel path for alt-buffer agents without reintroducing capture-pane races.
- `/terminal` WebSocket auth, origin guard, and connection limits are unchanged.

**Concerns:**
- None blocking. Client-controlled `replay: false` only affects what scrollback the browser paints for that socket; it does not bypass daemon-side tmux access controls.

### Code Quality
- **Green** — `app.js:5497` comment says `clearTermBuffer()` clears stale `mouseTrackingMode` on in-place agent switches, but `clearTermBuffer()` only emits the buffer-clear sequence and deliberately preserves parser modes (the reason `term.reset()` was removed). Wheel correctness relies on `appOwnsScroll()` consulting `agent_surface` before mouse mode and on daemon mouse re-arm — both still in place. Worth correcting the comment so future readers do not assume buffer clear resets DECSET state.
- **Green** — No Swift unit test exercises `TerminalWebSocket`'s `replay: false` branch directly; coverage is via `WebTerminalAssetTests.attachWindowSwitchesAgentSurfacesInPlace()` (static pins on `app.js`) and the updated `session-switch-attach.test.js` jsdom suite. Acceptable for this small daemon gate, but a focused decode/handler test would make regressions easier to spot.
- `swift build` succeeds; `swift test --package-path Packages/CrowDaemon --filter WebTerminalAssetTests` passes (22 tests). jsdom web-tests were not run here (missing local `jsdom` install).

### Summary Table

---

PR #1059, opened 2026-08-19
Launch the Cursor binary discovery verified, not a fresh PATH walk (CROW-1058)

### LINKED ISSUE #1058: Cursor sessions still launch Grok when the agent alias is grok-build
## Problem

When Crow launches a session whose agent is **Cursor**, it still starts **Grok Build**. This happens even though the `cursor-agent` alias exists on the machine. The session is configured as Cursor (`agentsByKind.work` / `--agent cursor`); the process that comes up is grok-build.

This is the CROW-989 collision firing again, past the defenses that were supposed to close it.

## Why the existing defenses are not enough

CROW-989 / the harness matrix already document two mitigations:

1. **Preferred-name-first PATH walk** — `resolveBinary()` searches every PATH entry for `cursor-agent` before any `agent` (`CursorAgent.launchCommandToken` + `alternateLaunchCommandTokens: ["agent"]`).
2. **Identity probe at registration** — a bare `agent` match is probed (`--approve-mcps`, `--trust`, …). A grok-build `agent` fails the probe and Cursor is shown disabled.

Observed residual (matrix): if only the legacy `agent` name exists and grok-build is earlier on PATH, the walk stops at the first `agent` and Cursor is marked unavailable.

What is happening now is **worse than that residual**: Cursor is treated as the selected/launchable agent, `cursor-agent` exists, and the launched command is still grok-build. So discovery and launch disagree.

Likely holes to close (any one is enough):

- **`findBinary()` re-resolves at launch** (`CursorAgent.autoLaunchCommand` → `findBinary() ?? launchCommandToken`) and does **not** re-run the identity probe. Registration may have accepted a good `cursor-agent`; a later walk (different PATH, fallback list) can return `~/.grok/bin/agent` or `~/.local/bin/agent` (grok-build mirrors there) and that path is interpolated into the tmux command.
- **`crowd`'s PATH is not the user's login PATH.** `cursor-agent` is visible in an interactive shell but missing from the daemon's `ShellEnvironment`, so the token-major walk never sees it and falls through to the `agent` alias / `fallbackCandidates` (`~/.local/bin/agent` is on that list).
- **The probed path is not cached.** `findBinary()` is a fresh walk every launch. The path that passed `verifyBinaryIdentity` at boot is not the path that gets launched.
- **A failed probe greys Cursor out at boot, but does not block a later launch** if something else still emits an `agent` command.

## Required outcome

A session whose `agentKind` is Cursor must never exec grok-build (or any other `agent`). If the only `agent` on PATH is grok-build and `cursor-agent` cannot be resolved, Cursor must stay **unavailable** and Crow must not launch. Do not "succeed" by running the colliding binary.

Grok sessions (`agentKind == grok`) are unchanged — they launch `grok`, not `agent`.

## Suggested direction

- Cache the absolute path that passed the identity probe and use **that** in `autoLaunchCommand` / `launchCommand` / Manager launch. Do not re-walk to a different `agent` at launch time.
- If launch must re-resolve, **re-probe** before exec; refuse to launch on probe failure (same log as boot: name the path, suggest `defaults.binaries.cursor`).
- Walk **every** `agent` on PATH until one verifies, instead of stopping at the first (closes the documented residual gap).
- Never emit a bare `agent` token into the pane command. Prefer the resolved absolute `cursor-agent` path; if `findBinary()` is nil, do not fall back to the colliding alias.
- Drop or probe `…/bin/agent` entries in `fallbackCandidates` the same way as PATH matches (`~/.local/bin/agent` is grok-build's mirror).
- Add a regression that plants grok-build's `agent` earlier on PATH than a real `cursor-agent` (and the reverse: only grok's `agent`) and asserts Cursor launch text contains the Cursor binary / does not launch, never grok-build.

## Verify

1. `cursor-agent` on PATH, grok-build's `agent` earlier on PATH → new Cursor session runs Cursor.
2. Only grok-build's `agent` on PATH → Cursor stays disabled; creating/starting a Cursor session does not spawn grok.
3. `defaults.binaries.cursor` pin still wins and skips the probe.
4. A Grok session still launches `grok`.

dgershman:
Closes #1058

A session configured as Cursor came up running xAI's grok-build, on a machine where `cursor-agent` was installed and the boot identity probe had passed. Both CROW-989 defenses were in place — they just didn't cover the launch.

## The two holes

**Discovery and launch resolved independently.** Registration probed whatever `resolveBinary()` returned; every later launch ran a *fresh* PATH walk via `findBinary()` and interpolated the result into the tmux command unprobed. For a colliding token those two walks can disagree — grok-build mirrors its own `agent` into `~/.grok/bin` and `~/.local/bin` — so "the binary we identified" and "the binary we launch" were two strings that happened to agree, not one.

**The walk stopped at the first sample.** The residual CROW-989 documented: with grok-build's `agent` earlier on PATH, the single sample *is* the impostor, and Cursor was written off as unavailable even when a genuine install sat further down PATH.

## The fix

| | |
|---|---|
| `VerifiedBinaries` (new) | `AgentDiscovery.evaluate` pins the path that passed the probe; cleared when an agent goes unavailable, so a stale pin can't survive an uninstall. Not persisted — a pin must not outlive the process that verified it. |
| `resolveBinaryCandidates()` | Every PATH hit across `binaryTokens` (still token-major) plus every executable fallback. `resolveBinary()` is its head, so **which binary Crow picks is unchanged** — the runners-up just survive for discovery to probe. |
| `AgentDiscovery.evaluate` | Probes down that list until one verifies, instead of judging the token by its first sample. A healthy install still costs one spawn: the preferred name is sampled first. |
| `launchBinary()` (new) | pin → verified path → *(for an agent that declares an alias)* the unambiguous preferred name only. Every launch path reads this instead of `findBinary()`. |

So **no Cursor launch text can name a `.../bin/agent` under any resolution outcome.** The last-resort fallback is the bare `cursor-agent` token, which cannot be grok-build — worst case is an honest `command not found`, never the wrong agent. (Bare rather than a hard refusal on purpose: the pane runs a login shell whose PATH is richer than `crowd`'s, one of the ticket's own hypothesized holes.)

`defaults.binaries.cursor` still outranks everything and still skips the probe.

## Why every adapter changed, not just Cursor

Widening the walk means discovery may now probe *past* a failing candidate for **any** agent that overrides `verifyBinaryIdentity` — Grok and Muse do — so a launch-time re-walk would return the impostor at the head of the list for them too. Leaving them on `findBinary()` would have introduced the same bug I was fixing. For the alias-free agents it's a no-op: their verified path *is* the head of the walk, and the preferred-name narrowing only applies to an agent with `alternateLaunchCommandTokens` (Cursor alone today). **Grok launch behavior is unchanged on a single-`grok` box**, per the ticket.

`handoffAgent`'s binary gate moves to `launchBinary()` as well, so it asks the same question the launch will rather than "some binary by that name exists".

## Ticket verify list

| | Covered by |
|---|---|
| 1. `cursor-agent` on PATH, grok's `agent` earlier → runs Cursor | `launchUsesTheVerifiedCursorBinary`, `managerAndHandoffLaunchUseTheVerifiedCursorBinary` |
| 2. Only grok's `agent` → Cursor disabled, never spawns grok | `launchNeverNamesAnUnverifiedAgentAliasBinary` + `failedProbeReportsTheMostPreferredCandidate` |
| 3. `defaults.binaries.cursor` pin wins, skips the probe | `explicitOverrideStillOutranksTheVerifiedPath`, `overridePinIsAlsoRecordedForLaunch` |
| 4. Grok still launches `grok` | CrowGrok suite unchanged and green |

Manager, auto-launch, and handoff are asserted **separately** — each builds its command text on its own, so a fix covering only auto-launch would leave two live routes to the wrong binary.

Both new suites are `.serialized`: they write the process-wide `VerifiedBinaries` / `BinaryOverrides` singletons. The CrowCore override case is deliberately *not* duplicated there — `BinaryOverridesTests` is a parallel sibling suite that resets the same map, and the CrowCursor version is the stronger assertion anyway (real launch text, not a stub's resolution).

## Testing

Per-package (never `make test` on this host). All green, run repeatedly for flakes:

`CrowCore` 672 · `CrowCursor` 80 · `CrowEngine` 784 · `CrowClaude` 93 · `CrowCodex` 84 · `CrowGrok` 58 · `CrowMuse` 58 · `CrowOpenCode` 58 · `CrowAntigravity` 41 · `swift build --product crowd` links.

Docs: `docs/agent-harness-matrix.md` — the ⚠️ residual-gap bullet is replaced by the two defenses that close it, and the risk-register row updated.

### DIFF
--- a/Packages/CrowAntigravity/Sources/CrowAntigravity/AntigravityAgent.swift
+++ b/Packages/CrowAntigravity/Sources/CrowAntigravity/AntigravityAgent.swift
@@ -73,7 +73,7 @@ public struct AntigravityAgent: CodingAgent {
         autoPermissionMode: Bool,
         telemetryPort: UInt16?
     ) -> String? {
-        let agentPath = AntigravityLaunchArgs.shellQuote(findBinary() ?? "agy")
+        let agentPath = AntigravityLaunchArgs.shellQuote(launchBinary() ?? "agy")
         // Bounded auto-permission — a no-op suffix on the pinned version (see
         // `AntigravityLaunchArgs.autoPermissionSuffix`); kept for call-site
         // uniformity so a future bounded flag lands once for every path.
@@ -166,7 +166,7 @@ public struct AntigravityAgent: CodingAgent {
             sessionID: sessionID,
             worktreePath: worktreePath,
             prompt: prompt,
-            binary: findBinary() ?? "agy"
+            binary: launchBinary() ?? "agy"
         )
     }
 
@@ -181,7 +181,7 @@ public struct AntigravityAgent: CodingAgent {
         // control doesn't apply. Terminal backend appends the submitting Enter,
         // so return the command without a trailing newline (cross-agent
         // convention).
-        let agentPath = AntigravityLaunchArgs.shellQuote(findBinary() ?? "agy")
+        let agentPath = AntigravityLaunchArgs.shellQuote(launchBinary() ?? "agy")
         return agentPath + AntigravityLaunchArgs.autoPermissionSuffix(autoPermissionMode)
     }
 
--- a/Packages/CrowClaude/Sources/CrowClaude/ClaudeCodeAgent.swift
+++ b/Packages/CrowClaude/Sources/CrowClaude/ClaudeCodeAgent.swift
@@ -51,7 +51,7 @@ public struct ClaudeCodeAgent: CodingAgent {
         autoPermissionMode: Bool,
         telemetryPort: UInt16?
     ) -> String? {
-        let claudePath = findBinary() ?? "claude"
+        let claudePath = launchBinary() ?? "claude"
         let rcArgs = ClaudeLaunchArgs.argsSuffix(
             remoteControl: remoteControlEnabled,
             sessionName: session.name,
@@ -130,7 +130,7 @@ public struct ClaudeCodeAgent: CodingAgent {
         autoPermissionMode: Bool,
         telemetryPort: UInt16?
     ) -> String {
-        let claudePath = findBinary() ?? "claude"
+        let claudePath = launchBinary() ?? "claude"
         return claudePath + ClaudeLaunchArgs.argsSuffix(
             remoteControl: remoteControlEnabled,
             sessionName: sessionName,
--- a/Packages/CrowCodex/Sources/CrowCodex/OpenAICodexAgent.swift
+++ b/Packages/CrowCodex/Sources/CrowCodex/OpenAICodexAgent.swift
@@ -79,7 +79,7 @@ public struct OpenAICodexAgent: CodingAgent {
         autoPermissionMode: Bool,
         telemetryPort: UInt16?
     ) -> String? {
-        let codexPath = findBinary() ?? "codex"
+        let codexPath = launchBinary() ?? "codex"
 
         switch session.kind {
         case .work:
@@ -225,7 +225,7 @@ public struct OpenAICodexAgent: CodingAgent {
         // command rather than on `supportsRemoteControl`
         // (`SessionService.ensureManagerSession`), so the CROW-1001 flip
         // leaves this path exactly as it was — same as Cursor's Manager.
-        return findBinary() ?? "codex"
+        return launchBinary() ?? "codex"
     }
 
     /// Codex TUI exposes `/rename` for the current thread (CROW-629).
--- a/Packages/CrowCore/Sources/CrowCore/Agent/AgentDiscovery.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Agent/AgentDiscovery.swift
@@ -22,19 +22,61 @@ public enum AgentAvailability: Sendable, Equatable {
 /// decision lived inside a `private static` local function).
 public enum AgentDiscovery {
     /// Resolve `agent`'s binary and decide availability. An explicit
-    /// `.override` pin is trusted without probing; a PATH/`.fallback` match on a
-    /// collision-prone token is identity-probed (`verifyBinaryIdentity`) so a
+    /// `.override` pin is trusted without probing; PATH/`.fallback` matches on a
+    /// collision-prone token are identity-probed (`verifyBinaryIdentity`) so a
     /// foreign same-named binary is reported unavailable rather than launchable.
+    ///
+    /// Also **pins the verified path** in `VerifiedBinaries` so launch execs the
+    /// binary discovery identified, rather than re-resolving and possibly
+    /// landing elsewhere (CROW-1058). That side effect is why this is the one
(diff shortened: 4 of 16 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- **Green** — `VerifiedBinaries` pins the identity-probed path at registration and `launchBinary()` re-checks executability on every read, closing the discovery-vs-launch TOCTOU that let a Cursor session exec grok-build's `agent` (`Packages/CrowCore/Sources/CrowCore/Agent/VerifiedBinaries.swift`, `CodingAgent.swift`).
- **Green** — For alias-bearing agents (Cursor), unverified `…/bin/agent` paths resolve to `nil` rather than launching the colliding binary; the bare `cursor-agent` fallback is unambiguous (`Packages/CrowCursor/Sources/CrowCursor/CursorAgent.swift`).
- **Green** — Pins are in-memory only and cleared on failed re-evaluation, so stale paths cannot survive uninstall or a failed re-registration (`AgentDiscovery.swift`).
- **Green** — `handoffAgent` gates on `launchBinary()` instead of `findBinary()`, aligning the permission check with what will actually exec (`SessionService.swift`).
- **Green** — All adapters route launch text through `launchBinary()`, preventing identity-probed agents (Grok, Muse) from regressing to a fresh PATH walk that could return an impostor at the list head.

**Concerns:**
- **Green** — Launch does not re-run `verifyBinaryIdentity` at exec time; only file executability is re-checked. A same-path binary swap between boot and launch is theoretically possible but is the documented tradeoff for performance, and explicit `defaults.binaries.*` overrides remain the escape hatch.
- **Green** — Falling back to the bare `cursor-agent` token when no verified path is available relies on the login shell's richer PATH. The PR documents this as an intentional, bounded risk (honest `command not found` vs. wrong agent).

### Code Quality
- **Green** — `resolveBinaryCandidates()` / `BinaryTokenResolver.allOnPath` preserve token-major preference while surfacing every PATH hit for multi-probe discovery; `resolveBinary()` remains the head, so selection semantics are unchanged (`CodingAgent.swift`, `ShellEnvironment.swift`).
- **Green** — Test coverage is thorough and correctly serialized around process-wide singletons: core resolution/launch selection (`BinaryResolutionTests.swift`), discovery pinning (`AgentDiscoveryTests.swift`), and all three Cursor launch surfaces — auto-launch, manager, handoff (`CursorAgentTests.swift`).
- **Green** — `swift build` succeeds; `CrowCore` (672), `CrowCursor` (80), and `CrowGrok` (58) test suites all pass locally.

### Summary Table

**Recommendation:** Approve — driven by 0 Red, 0 Yellow, 8 Green findings.

---

PR #1071, opened 2026-08-19
crow-review-pr skill: parse PR ref, detect Go module, keep test tail, add --no-post + verify-red step

### LINKED ISSUE #1069: review-pr skill: $ARGUMENTS mangles flags, hardcoded core/ path, truncated test output, no review-only mode
The `review-pr` skill (`~/.claude/commands/review-pr.md`) has three bugs that surfaced running it against corveil#2437, plus one gap worth closing while it is open.

## 1. `$ARGUMENTS` is substituted whole, so any flag corrupts every command

The skill interpolates the raw argument string into each command as if it were only a PR number. Invoked as `/review-pr 2437 --no-post`, step 1 becomes:

```bash
gh pr checkout 2437 --no-post          # fails: unknown flag
```

and step 5 would become `gh pr review 2437 --no-post --approve …`. Any invocation carrying a flag is broken, and the failure is at the *first* command, so the reviewer has to work around it before doing any review.

Needs the PR number parsed out of the argument string. There is also no documented flag set today — see item 4.

## 2. Hardcoded `cd core &&` in the static-analysis step

Step 4 runs:

```bash
cd core && go vet ./... 2>&1
cd core && go test ./... -v 2>&1 | head -50
```

`core/` does not exist in corveil — the Go module is `go/`. Both commands fail outright. Looks like a path carried over from another project.

Either detect the module directory (`git ls-files '*/go.mod'`) or make it configurable rather than assumed.

## 3. `go test ./... -v | head -50` truncates before the summary

Unqualified `-v` on corveil's 97-package suite emits thousands of lines of PASS output; `head -50` then cuts inside the first package or two. The reviewer sees neither the failures nor the final tally — the one line that actually matters.

Dropping `-v` and keeping the tail of the output would give strictly more signal for less noise.

## 4. No review-only mode

Step 5 always posts, via `--approve` / `--request-changes` / `--comment`. There is no path for "produce the review, don't post it," which is exactly what you want when:

- the reviewer is the PR author (GitHub does not accept a self-review anyway),
- you are testing or iterating on the skill itself,
- you want to read the findings before deciding whether they are worth posting.

A `--no-post` (or `--dry-run`) flag that renders the review to stdout and skips step 5 would cover all three. This is the flag whose absence triggered bug 1 in the first place.

## 5. Suggestion: require findings to be verified, not just asserted

Not a bug, but it changed the outcome in practice. The skill asks for findings and severities but never asks the reviewer to *establish* that a finding is real. On #2437 two red findings looked like plausible-but-arguable readings of the code until they were actually checked — one by running a throwaway test that reproduced the bad write and was then deleted, one by reading the second mount's constructor and confirming no filter was applied. Written up unverified, both would have been hedged "this might…" comments that a reader could reasonably dismiss.

A step along the lines of *"for each red finding, either reproduce it (a temporary test, a probe, a log line) or cite the code that refutes it; state which you did"* would raise the floor of the output considerably, and it composes with the existing `--review-blocking-severity` gating in `crow workspace` (a red finding that blocks a merge should have to earn it).

---

Found while running the skill against [corveil#2437](https://github.com/corveil/corveil/pull/2437).

dgershman:
Closes #1069

## Which skill

The ticket reports these bugs against `~/.claude/commands/review-pr.md` — a **corveil-binary** embedded skill (installed by `corveil skill install`), not in this repo. The crow-repo artifact that carries the identical defects is the **`crow-review-pr`** skill, which Crow ships and scaffolds into review worktrees. This PR fixes that skill; if the intent was the corveil binary, redirect me.

The skill lives in two byte-identical halves (dev reads `skills/crow-review-pr/SKILL.md`, release scaffolds from `Resources/crow-review-pr-SKILL.md.template`); both are updated and stay identical (enforced by `scripts/check-workspace-custom-instructions.sh` + `ReviewVerdictPolicyTests`).

## Changes (mapping to the ticket)

1. **`$ARGUMENTS` mangles flags** → new **Step 0** parses the PR reference into `$PR` (used in every `gh`/`git` command) and detects `--no-post`/`--dry-run` into `$POST`. Crow review sessions substitute `$ARGUMENTS`→PR URL and pass no flags, so they are unchanged; the bug only bit the manual `/crow-review-pr 2437 --no-post` path, and only once a flag existed (item 4).
2. **Hardcoded `core/`** → detect the Go module with `git ls-files '*go.mod'` and run `go -C <module-dir>` against its parent (corveil's module is `go/`, not `core/`).
3. **`go test -v | head -50` truncates** → drop `-v` and switch to `tail -50`, so the failures and the final tally survive a large suite instead of being cut off inside the first package.
4. **No review-only mode** → `--no-post` / `--dry-run` renders the review to stdout and skips `gh pr review` (Step 5). For self-review (GitHub rejects it), iterating on the skill, or previewing findings. The Step 5a guardrails still run on the drafted body.
5. **Verify findings (suggestion)** → new **Step 4b**: each Red finding must be *reproduced* (throwaway test/probe) or *refuted* by cited code, stating which; an unverified Red downgrades to Yellow or is dropped. Composes with the workspace `--review-blocking-severity` gate.

## Preserved / untouched

- CROW-963 verdict placeholders (`{{CROW_REVIEW_VERDICT_*}}`, `{{CROW_REVIEW_GRADING_GUIDANCE}}`) and their golden expansion.
- CROW-974 Step 5a pre-submit guardrails (target + verdict-consistency checks).
- Attribution footer (Step 5b).

## Verification

- `bash scripts/check-workspace-custom-instructions.sh` → OK (byte-identity, placeholders, guardrails).
- `swift test --package-path Packages/CrowCore --filter ReviewVerdictPolicy` → 21/21.
- `swift test --package-path Packages/CrowEngine --filter SessionServiceReviewPrompt` → 20/20.

### DIFF
--- a/Resources/crow-review-pr-SKILL.md.template
+++ b/Resources/crow-review-pr-SKILL.md.template
@@ -16,7 +16,10 @@ All `gh` and `git` commands require `dangerouslyDisableSandbox: true` because th
 
 ## Arguments
 
-- `$ARGUMENTS` - The PR URL or number to review (required)
+`$ARGUMENTS` is the raw argument string. **Parse it — never paste it whole into a command.** A Crow review session passes just the PR reference, but a manual `/crow-review-pr` invocation may append flags (`/crow-review-pr 2437 --no-post`), and pasting the whole string makes step 1 `gh pr checkout 2437 --no-post` fail on an unknown flag.
+
+- **PR reference** (required) — the first URL or number token in `$ARGUMENTS` (e.g. `2437` or `https://github.com/org/repo/pull/2437`). Every command below writes it as `$PR`; set `PR` to it in Step 0 and pass `"$PR"` to `gh`/`git`, never the raw `$ARGUMENTS`.
+- **`--no-post`** / **`--dry-run`** (optional) — render the review to stdout and **skip Step 5** (do not post). Use it when you are the PR author (GitHub rejects a self-review), when iterating on this skill, or when you want to read the findings before deciding to post. Sets `POST=false`; otherwise `POST=true`.
 
 ## Activation
 
@@ -29,10 +32,17 @@ This skill activates when:
 
 You are performing a code and security review on PR $ARGUMENTS. Follow these steps:
 
+### Step 0: Parse the arguments
+
+Split `$ARGUMENTS` (see **Arguments** above) before running anything:
+
+- `PR` — the PR URL or number. Use `"$PR"` in every `gh`/`git` command below; a stray `--no-post` reaching `gh pr checkout` fails at the first command.
+- `POST` — `false` if `$ARGUMENTS` contains `--no-post` or `--dry-run`, otherwise `true`. Gates Step 5.
+
 ### Step 1: Checkout the PR
 
 ```bash
-gh pr checkout $ARGUMENTS
+gh pr checkout "$PR"
 ```
 
 ### Step 2: Gather PR Information
@@ -40,7 +50,7 @@ gh pr checkout $ARGUMENTS
 Get the PR details including title, description, and changed files:
 
 ```bash
-gh pr view $ARGUMENTS --json title,body,headRefName,baseRefName,additions,deletions,changedFiles,files
+gh pr view "$PR" --json title,body,headRefName,baseRefName,additions,deletions,changedFiles,files
 ```
 
 ### Step 3: Review the Code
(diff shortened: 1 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

- **Existing pathway:** The `crow-review-pr` skill’s dual-half layout (`skills/crow-review-pr/SKILL.md` + `Resources/crow-review-pr-SKILL.md.template`) with byte-identity enforced by `scripts/check-workspace-custom-instructions.sh` and golden tests in `ReviewVerdictPolicyTests`.
- The change extends that pathway in place — Step 0 argument parsing, Go module discovery, dry-run gating, and Red verification — without introducing a parallel review mechanism. Confirmed: `diff` between the two halves is empty and `check-workspace-custom-instructions.sh` passes.

### Security Review
**Strengths:**
- All `gh`/`git` invocations now use quoted `"$PR"` instead of raw `$ARGUMENTS`, reducing shell-metacharacter risk if flags or odd tokens appear in the argument string.
- `--no-post` / `--dry-run` is opt-in documentation only; it does not weaken posting guardrails (Step 5a still applies to the drafted body).
- No secrets, credentials, or auth-surface changes — documentation-only diff.

**Concerns:**
- None identified in this diff.

### Code Quality
- **Step 0 (`skills/crow-review-pr/SKILL.md:35-40`):** Clean split of `PR` vs `POST`; fixes the reported `gh pr checkout … --no-post` failure mode without changing Crow review-session behavior (sessions pass only the URL).
- **Go static analysis (`skills/crow-review-pr/SKILL.md:95-100`):** Replaces hardcoded `core/` with `git ls-files '*go.mod'` + `go -C <module-dir>`; drops `-v` and switches `head` → `tail` so failure summaries survive large suites — a sensible tradeoff for agent-readable output.
- **Step 4b (`skills/crow-review-pr/SKILL.md:118-125`):** Requires reproducible or refute-checked Red findings before blocking; aligns review rigor with the workspace severity gate.
- **Green:** Multi-module repos may list several `go.mod` paths; the skill leaves module choice to agent judgment. A future one-liner (“vet/test each listed module”) would help, but that is optional polish, not a defect in this change.
- **Verification:** `ReviewVerdictPolicy` 21/21 and `SessionServiceReviewPrompt` 20/20 pass locally; both skill halves remain byte-identical.

### Summary Table

---

PR #1081, opened 2026-08-20
Realign the xterm cursor when the iOS keyboard opens (CROW-1078)

### LINKED ISSUE #1078: Mobile web: iOS/iPad keyboard leaves the xterm cursor misaligned
## Problem

On the Crow web terminal on iOS / iPad, opening the software keyboard does not keep the xterm caret aligned with the cell it belongs to. The blinking cursor (and the iOS IME/caret accessory) land at the wrong place on the xterm surface — often shifted, or still sitting in the pre-keyboard layout — so typed input no longer matches where the cursor is drawn.

This is related to, but not fixed by, CROW-988 (#996). That work refits the **host height** so the prompt is not hidden under the keyboard. It does not resync the **cursor**.

## Root cause

Two coordinate systems diverge the moment the keyboard opens, and Crow never reconciles them.

**1. iOS shrinks only the visual viewport.** Safari leaves the layout viewport (`window.innerHeight`, `100dvh`, `height: 100%`) unchanged and does not fire `window.resize`. Crow's fit path is wired to that event (`app.js` `window.addEventListener('resize', fitTerminal)`), so FitAddon keeps a full-height grid unless `CrowViewportAddon` (CROW-988) happens to shrink the host. `html, body { overflow: hidden }` also means the occluded rows cannot be scrolled into view.

**2. The drawn cursor is not the focused input.** The block the user sees is painted on the WebGL canvas. Focus lives on `.xterm-helper-textarea`, which xterm.css parks off-screen:

```css
.xterm .xterm-helper-textarea {
  position: absolute;
  opacity: 0;
  left: -9999em;
  top: 0;
}
```

Tapping the terminal focuses that textarea. iOS then scrolls the *visual* viewport so the focused field sits above the keyboard. Because the field starts at `-9999em` (and is later moved to `cursorX * cellWidth` / `cursorY * cellHeight` in **layout** pixels), Safari shifts `visualViewport.offsetLeft` / `offsetTop` instead of resizing the page.

**3. xterm positions the textarea in layout space only.** `CompositionHelper` (and the same cell-math elsewhere) sets `textarea.style.left/top` from buffer `x`/`y` × cell size, with no `visualViewport.offset*` compensation. The canvas cursor stays in layout/WebGL space; the IME caret follows the textarea in visual-viewport space. Those no longer line up.

**4. CROW-988 does not touch this path.** `xterm-addon-crow-viewport.js` only writes `host.style.height` and calls `fitTerminal()`. It never repositions the helper textarea, never subtracts `visualViewport.offset*`, and never forces a cursor/WebGL refresh. If iOS has already scrolled the visual viewport to chase the textarea, `offsetTop + height ≈ layoutHeight`, occlusion computes to ~0, and the addon treats that as "no keyboard" and does nothing.

**5. WebGL makes the mismatch visible.** `ensureTerminal()` loads `WebglAddon` after `open()`. On iOS a WebGL canvas is a compositor layer that often does not re-composite when the visual viewport animates with the keyboard, so the drawn cursor can stay at the old screen position even after a host refit.

## Proposed fix

1. On touch / coarse-pointer surfaces, keep the helper textarea inside the visual viewport (at the cursor cell, or at `0,0` if off-screen parking is required) so iOS does not scroll to `-9999em`.
2. When placing the textarea / composition view, subtract `visualViewport.offsetLeft` / `offsetTop` so IME and caret stay on the cell.
3. After `CrowViewportAddon` applies or restores a keyboard inset, force a cursor resync (`term.refresh` / textarea restyle) and, if needed, a WebGL context/canvas resize so the compositor layer catches up.
4. Revisit the addon's occlusion math so a visual-viewport *scroll* that iOS used to chase the textarea is not treated as "no keyboard."
5. Consider skipping WebGL on iOS/iPad (DOM renderer) if the canvas layer still drifts after (3).

Do this as a shared addon or a single helper used by both `app.js` and `terminal.html` so the surfaces do not drift the way they did before CROW-988.

## Acceptance

- On iPhone and iPad Safari, opening the keyboard leaves the blinking xterm cursor on the same cell as the prompt / typed text, fully visible above the keyboard.
- The iOS IME / caret accessory (when shown) sits on that same cell, not offset horizontally or vertically.
- Closing the keyboard restores the full-height grid with the cursor still on the correct cell — no leftover gap, clip, or drift.
- Rotating or changing keyboard height with the keyboard open keeps the cursor aligned.
- Desktop and non-`visualViewport` webviews are unchanged (CROW-1045 must stay inert on desktop).
- CROW-988's "prompt stays above the keyboard" behavior still holds.

## Notes

Reproduced conceptually against the live web UI (`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/`) and the vendored xterm helper-textarea CSS. CROW-988 / CROW-1045 are the prior art for the host-height path; this ticket is the cursor/textarea/WebGL follow-up.

dgershman:
Closes #1078

## Problem

On iOS/iPad Safari, opening the software keyboard left the xterm caret off the cell it belonged to: the **drawn** cursor (canvas/WebGL) and the **iOS IME/caret accessory** landed on different cells, so typed input no longer matched where the cursor was painted. CROW-988 (#996) refit the host *height* so the prompt clears the keyboard, but never resynced the *cursor*.

Two coordinate systems diverge the moment the keyboard opens:
1. **iOS chases the off-screen textarea.** xterm parks `.xterm-helper-textarea` at `left:-9999em`; the instant it's focused, Safari scrolls the *visual* viewport to drag that off-screen field into view — shifting `offsetLeft/offsetTop`, stranding the drawn cursor, and (as a side effect) collapsing the keyboard-occlusion math to ~0 so the CROW-988 fit thinks there's no keyboard.
2. **WebGL doesn't re-composite.** On iOS a WebGL canvas is a compositor layer that often doesn't re-composite while the visual viewport animates with the keyboard, so the drawn cursor stays at its old screen position.

## Fix

All the cursor/viewport reconciliation lives in the **shared `CrowViewportAddon`** — served from one source (`CrowTerminal/…/xterm-addon-crow-viewport.js`) to *both* web front-ends (`app.js` and web `terminal.html`), so they can't hand-mirror-drift. Built on the existing CROW-988 / CROW-1045 host-height path, not replacing it.

- **Re-home the parked textarea** (touch surfaces only): a stylesheet rule moves `.xterm-helper-textarea`'s *parked* state from `left:-9999em` to on-screen `left:0`. It is deliberately **not** `!important` and is appended after `xterm.css`, so it wins only the parked state and yields to xterm's own inline cursor-cell sync (written on cursor-move/render). iOS no longer flies off to `-9999em`.
- **Keyboard detection is now offset-independent:** the presence test is `layoutHeight - vv.height` (the keyboard's height) instead of `layoutHeight - (offsetTop + vv.height)` — a large `offsetTop` (from a chased scroll) used to collapse that to ~0 and read as "no keyboard." Host *sizing* still uses `offsetTop`, so CROW-988's "prompt stays above the keyboard" is unchanged.
- **Repaint on inset apply/restore:** forces the drawn cursor to follow the moved host and re-runs xterm's textarea sync onto the cell.
- **Reconcile on textarea `focus`:** makes the first keyboard frame correct instead of waiting for the trailing vv resize/scroll.
- **`app.js` drops WebGL on touch surfaces** (`visualViewport` present **and** `keyboardCapable()` — the *same* test exported once from the addon, so the two can't drift). The DOM/canvas renderer follows the viewport and honors the repaint; WebGL's compositor layer ignores it. Desktop and non-`visualViewport` webviews keep WebGL.

### One intentional deviation from the ticket's proposed steps

The ticket suggests *subtracting* `visualViewport.offsetLeft/offsetTop` when placing the textarea. I did **not** do that, on purpose: xterm's helper textarea and the drawn cursor are siblings in the **same absolute coordinate space**, so a visual-viewport shift moves them together — subtracting `offset*` from only the textarea would *de*-align it from the canvas cursor, not align it. The real defects are the off-screen chase (fixed by the on-screen park) and WebGL not re-compositing (fixed by dropping WebGL on touch); the acceptance criteria — caret and IME accessory on the same cell — are met by those. Rationale is documented in `_homeTextarea`.

## Tests

`Packages/CrowDaemon/web-tests/visual-viewport.test.js` gains 21 assertions (textarea home, focus reconcile, chased-scroll detection, apply/restore repaint, shared `keyboardCapable` export). The whole `test:ci` suite is green:

```
cd Packages/CrowDaemon/web-tests && npm ci && npm run test:ci   # exit 0
```

Desktop / non-touch stays fully inert (existing CROW-1045 cases + new case 14 confirm no park style, no focus reconcile, WebGL retained).

## Manual verification (human-side — iOS/iPad Safari, cannot be done from CI)

On **iPhone and iPad Safari**, against the web UI:
1. Tap the terminal to open the keyboard → the blinking cursor stays on the same cell as the prompt/typed text, fully visible above the keyboard; the IME/caret accessory (when shown) sits on that same cell, not offset.
2. Type a few characters → what you type lands where the cursor is drawn.
3. Rotate the device / change keyboard height with the keyboard open → cursor stays aligned.
4. Close the keyboard → full-height grid restored, cursor still on the correct cell, no leftover gap/clip/drift.
5. Confirm a **desktop** browser and the macOS app are unchanged (WebGL still used on desktop; no layout change).

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -5633,11 +5633,33 @@ function ensureTerminal() {
   }
   // WebGL renderer for throughput; must load after open(). Falls back to the
   // default renderer if the GL context is unavailable or gets lost.
+  //
+  // CROW-1078: skipped on touch surfaces (iOS/iPad). A WebGL canvas is its own
+  // compositor layer that often fails to re-composite while Safari animates the
+  // *visual* viewport with the software keyboard, so the drawn cursor strands at
+  // its old screen position while the IME caret (a DOM textarea) follows the
+  // keyboard — the two land on different cells. The DOM/canvas renderer follows
+  // the viewport correctly and honors the viewport addon's post-inset repaint.
+  // `keyboardCapable()` is the SAME touch test the viewport addon gates on,
+  // shared from its namespace so the two can't drift; requiring `visualViewport`
+  // too makes this condition identical to "the viewport addon will engage" — so
+  // WebGL is dropped on exactly the surfaces where the keyboard can animate the
+  // viewport, and non-visualViewport webviews keep WebGL unchanged. If the addon
+  // failed to load, `skipWebgl` stays false and WebGL loads exactly as before.
+  let skipWebgl = false;
   try {
-    const webglAddon = new WebglAddon.WebglAddon();
-    webglAddon.onContextLoss(() => webglAddon.dispose());
-    term.loadAddon(webglAddon);
-  } catch (_) { /* WebGL unavailable → canvas/DOM renderer */ }
+    skipWebgl = !!window.visualViewport
+      && typeof CrowViewportAddon === 'object' && CrowViewportAddon
+      && typeof CrowViewportAddon.keyboardCapable === 'function'
+      && CrowViewportAddon.keyboardCapable();
+  } catch (_) { skipWebgl = false; }
+  if (!skipWebgl) {
+    try {
+      const webglAddon = new WebglAddon.WebglAddon();
+      webglAddon.onContextLoss(() => webglAddon.dispose());
+      term.loadAddon(webglAddon);
+    } catch (_) { /* WebGL unavailable → canvas/DOM renderer */ }
+  }
   term.onData(sendToPTY);
   // CROW-934: hydrate the shell scrollback from tmux when the user reaches the
   // top of what the live stream actually delivered.
--- a/Packages/CrowDaemon/web-tests/visual-viewport.test.js
+++ b/Packages/CrowDaemon/web-tests/visual-viewport.test.js
@@ -8,6 +8,11 @@ const { JSDOM } = require('jsdom');
 // the "leave every non-phone surface alone" guards are pinned without a device.
 // CROW-1045 adds the desktop-WKWebView guard: `visualViewport` presence is not a
 // keyboard, so a non-touch surface must stay fully inert (case 1b).
+// CROW-1078 adds the cursor reconciliation: re-home the off-screen helper
+// textarea so iOS stops chasing it (cases 10, 11), keep the keyboard-presence
+// test independent of a chased visual-viewport scroll (case 12), and repaint the
+// cursor when the inset changes (case 13). keyboardCapable is exported and shared
+// with app.js's WebGL gate (case 14).
 const ADDON_JS =
   __dirname + '/../../CrowTerminal/Sources/CrowTerminal/Resources/xterm/xterm-addon-crow-viewport.js';
 
@@ -70,10 +75,20 @@ function makeWorld({ hasVisualViewport = true, hostTop = HOST_TOP, keyboardCapab
   // term.open(), which is what the addon defaults its host to.
   const element = window.document.createElement('div');
   host.appendChild(element);
+  // The helper textarea xterm parks off-screen; the addon re-homes it and hooks
+  // its focus to reconcile the cursor (CROW-1078). A real element so a dispatched
+  // focus event reaches the addon's listener.
+  const textarea = window.document.createElement('textarea');
+  textarea.className = 'xterm-helper-textarea';
+  element.appendChild(textarea);
   const scrolls = [];
   const resizes = [];
+  const refreshes = [];
   const term = {
     element,
+    textarea,
+    rows: 24,
+    refresh: (a, b) => refreshes.push([a, b]),
     buffer: { active: { viewportY: 10, baseY: 10 } }, // at the live edge
     scrollToBottom: () => scrolls.push(1),
   };
@@ -89,7 +104,7 @@ function makeWorld({ hasVisualViewport = true, hostTop = HOST_TOP, keyboardCapab
   };
 
   return {
-    window, vv, host, term, addon, scrolls, resizes, listeners, flush, frames, emit,
+    window, vv, host, term, addon, scrolls, resizes, refreshes, listeners, flush, frames, emit,
     activate: () => addon.activate(term),
     height: () => host.style.height,
     // A keyboard of `px`, iOS-style: visual viewport shrinks, layout does not.
@@ -97,6 +112,12 @@ function makeWorld({ hasVisualViewport = true, hostTop = HOST_TOP, keyboardCapab
     // `display: none` — the web app hides the terminal while a board is open.
     hide: () => { clientHeight = 0; },
     show: () => { clientHeight = LAYOUT_HEIGHT - hostTop; },
+    // iOS raises the keyboard on focus; dispatch it, then settle the frame.
+    focus: () => { textarea.dispatchEvent(new window.Event('focus')); flush(); },
+    // The on-screen park <style> the addon injects (touch surfaces only).
+    homeStyle: () => window.document.getElementById('crow-vv-textarea-home'),
+    // The addon's exported namespace (keyboardCapable is shared with app.js).
+    ns: () => ctx.CrowViewportAddon,
   };
 }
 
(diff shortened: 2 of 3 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- All DOM mutations use hardcoded, non-user-controlled values (`_homeTextarea` injects a fixed CSS rule; no dynamic string assembly from untrusted input).
- Touch/desktop gating reuses the existing CROW-1045 `keyboardCapable` guard, so desktop WKWebViews stay fully inert — no new attack surface on non-touch surfaces.
- Event listeners (`visualViewport` resize/scroll, textarea focus) are removed in `dispose`, and height overrides are restored — no obvious listener or style leaks on terminal teardown.
- The WebGL skip is defensive (`try/catch`, falls back to prior behavior if the addon fails to load).

**Concerns:**
- None identified. This is client-side viewport/IME layout reconciliation with no auth, network, or credential handling.

### Code Quality
- **Green:** The keyboard-presence fix in `xterm-addon-crow-viewport.js` (`layoutHeight - vv.height` instead of `layoutHeight - visibleBottom`) correctly decouples detection from iOS's chased-scroll `offsetTop`, while sizing still honors `visibleBottom` — case 12 pins this well.
- **Green:** Re-homing the parked textarea via a non-`!important` stylesheet rule is the right tradeoff: it stops the `-9999em` chase without overriding xterm's inline cursor-cell sync.
- **Green:** Exporting `keyboardCapable` from the addon namespace and consuming it in `app.js` for the WebGL gate prevents the parity drift this addon exists to avoid.
- **Green:** 21 new assertions in `visual-viewport.test.js` (cases 10–14) cover the new paths; full `npm run test:ci` passes locally (all suites green, 55 assertions in the viewport test alone).
- **Green:** The PR documents the intentional deviation from the ticket's offset-subtraction approach with clear rationale in `_homeTextarea` — grading against the diff, that accepted design choice is not a defect.
- **Green:** Manual iOS/iPad Safari verification remains human-side (called out in the PR); CI cannot substitute for it, but the automated coverage is solid for everything jsdom can model.

### Summary Table

**Recommendation:** Approve — driven by 0 Red, 0 Yellow, 5 Green findings.

---

dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- All DOM changes use static, hard-coded CSS (`left:0` park rule) — no user-controlled strings reach the injected `<style>` block.
- Touch-surface gating reuses the existing `keyboardCapable` guard (CROW-1045), so desktop WKWebViews stay inert and never receive the park style or focus hooks.
- Event listeners (`focus`, `visualViewport` resize/scroll) are removed in `dispose`; inline height overrides are restored on dispose when applied.
- `term.refresh` and `onResize` callbacks are wrapped in try/catch so a renderer failure cannot take down the terminal host.

**Concerns:**
- None identified. This is client-side layout/IME reconciliation with no auth, network, or credential surface.

### Code Quality
- **Green** — `Packages/CrowTerminal/Sources/CrowTerminal/Resources/xterm/xterm-addon-crow-viewport.js`: Keyboard presence now uses `layoutHeight - vv.height` instead of `layoutHeight - visibleBottom`, correctly decoupling detection from iOS's chased `offsetTop` scroll while still sizing via `visibleBottom`. The chased-scroll case (test case 12) pins this behavior.
- **Green** — `xterm-addon-crow-viewport.js`: `_homeTextarea` only overrides the parked `left:-9999em` state (xterm.css already parks `top:0`), deliberately without `!important`, so inline cursor-cell sync still wins. Rationale for omitting `offset*` compensation is documented and sound.
- **Green** — `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js`: WebGL is skipped only when `visualViewport` exists and `CrowViewportAddon.keyboardCapable()` is true — the same gate the addon uses. Fallback to WebGL when the addon fails to load preserves prior behavior. `terminal.html` does not load WebGL, so no parity gap there.
- **Green** — `Packages/CrowDaemon/web-tests/visual-viewport.test.js`: 21 new assertions cover park injection, focus reconcile, chased-scroll detection, apply/restore repaint, and the exported `keyboardCapable` namespace. Full `npm run test:ci` suite is green locally.

### Summary Table

---

Issue #637, opened 2026-07-09
Terminal display clobbers / double lines / misaligned rows after resize or layout


dgershman:
## Summary
The terminal display sometimes gets clobbered: double lines, staggered/overlapping rows, and UI frames that don't line up. This is intermittent and is likely a cols/rows desync across Crow's terminal stack.

## Likely cause
Crow paints through three layers that must agree on terminal size:

**xterm.js (`FitAddon`) → PTY (`TIOCSWINSZ`) → tmux pane → agent TUI**

When those disagree even briefly, fullscreen agent TUIs (Claude/Cursor/etc.) keep drawing with absolute cursor moves for the *old* width while xterm shows a different grid. Shell soft-wrap can look fine; TUI apps need a clean redraw after resize and leave leftovers until that completes.

## Crow-specific triggers
1. **Aggressive fit on layout** — `XTermSurfaceView.layout()` calls `scheduleFit()` on layout (sidebar resize, tab switch, window chrome). Fit clears/resizes xterm and posts a resize; the PTY SIGWINCHes tmux/the agent. If another redraw is already in flight, the screen partially clobbers until the next full paint.
2. **Fit race at first paint** — first `fit()` can run before fonts fully settle (`MesloLGS NF` → Menlo fallback). Cell width changes → col count changes after the agent already drew.
3. **No-op / churny resize** — repeated fits that re-fire the same size (or near-identical sizes) still SIGWINCH the agent and can corrupt an in-progress frame.
4. **Scrolling / tmux copy-mode** (related) — wheel scroll can put tmux in copy-mode while the live pane keeps updating underneath; jumping back can also look mangled.

## Repro hints
- More common right after resizing the window/sidebar or switching tabs.
- Often clears after a tiny resize or refocus (forces agent redraw).

## Proposed direction
- Debounce fit longer / ignore no-op same-size fits before posting `crowResize`.
- Avoid fit storms from `layout()` (only fit when container size or content scale actually changes).
- Delay or re-fit after font metrics settle on first paint.
- Consider forcing a clean redraw path after resize (tmux client-size sync / agent redraw) so leftover TUI cells don't stick.
- Ensure xterm cols/rows, PTY winsize, and tmux pane size stay in lockstep.

## Acceptance criteria
- Resizing the Crow window/sidebar or switching tabs does not leave permanently clobbered / double / misaligned terminal content.
- Same-size fit cycles do not spam SIGWINCH.
[...]

PR #1093, opened 2026-08-22
Fit the pane on in-place agent session switch (CROW-1091)

### LINKED ISSUE #1091: Scrollbar missing when switching coding sessions until manual refresh
## Symptom

When switching between coding sessions, the terminal has **no scroll bar** — you can't scroll back until you manually hit refresh, after which the scroll bar appears and works normally.

## Observation / hypothesis

Switching sessions already triggers a refresh internally, so the working theory is a **race condition**: the terminal/viewport is rendered (or the xterm.js instance attached) for the newly selected session **before** the pane has its final layout size, so the scrollbar geometry is computed against a stale/zero viewport and never recalculated. The manual refresh re-runs the sizing after layout has settled, which is why it fixes it.

## Repro

1. Open a session with enough scrollback to need a scroll bar.
2. Switch to another coding session, then back (or just switch in).
3. Observe: no scroll bar on the newly shown terminal.
4. Hit refresh → scroll bar appears and scrolling works.

## Where to look

The terminal scroll/viewport behavior is mirrored across the three hand-written front-ends (desktop `terminal.html`, web `terminal.html`, web `app.js`), so a fix likely needs to land in whichever path the session-switch render goes through — and be checked in the others for the same gap. Likely suspects:

- session-switch render attaching/showing the xterm terminal before the pane's final dimensions are known (a `fit()` / resize that runs too early or not at all on switch).
- the on-switch refresh doing a `fit`/resize the initial switch render skips — i.e. the switch path and the refresh path aren't running the same sizing step.
- a missing resize-after-layout (e.g. needs to defer the fit to the next frame / after the container is visible, or observe the container resize) so switching relies on the manual refresh to trigger it.

## Acceptance

- Switching into a session shows a correctly-sized, working scroll bar immediately, with no manual refresh needed.
- Fix verified in all three terminal front-ends that share this behavior.

dgershman:
Closes #1091

## Problem

Switching between coding sessions left the terminal with **no working scroll bar** until a manual reload — after which it appeared and scrolled normally.

## Root cause

CROW-1035 (#1036/#1051) moved agent session/tab switches off the full terminal reload onto an **in-place** `switchAgentWindow` (select-window with `replay:false`, no new PTY) to stop the caret jump (Claude) and footer-chrome stacking (Cursor).

But the reload/connect attach path also runs `applyTermFit()` on attach — and `fitAddon.fit()` re-syncs xterm's scroll area, which is exactly what gives the history scrollbar its geometry. The in-place switch ran **no fit at all**, so a switched-in agent surface kept the *previous* window's grid geometry and stale scrollbar sizing until a manual reload re-ran the fit. That matches the ticket's own hypothesis: the switch path and the refresh path weren't running the same sizing step.

## Fix

`switchAgentWindow` now runs the same sizing step the reload path runs, after the in-place `select-window`:

- **`fitTerminal()`** — deferred a frame (rAF), so it measures the pane at its *settled* post-switch layout (the tab-bar / header height differs between session kinds).
- **`updateTerminalScrollbar()`** — immediate, so the pinned-visible class reflects the switched-in buffer instead of the previous window's (clears a stale bar right away).

No replay and no new PTY, so this stays clear of the CROW-1035 caret jump and CROW-1048 chrome stacking a full reload would reintroduce. A same-size switch fits to identical cols/rows, so `applyTermFit` sends no SIGWINCH (#637); a genuine size change sends the one SIGWINCH it always needed.

## "Verified in all three front-ends"

The session-switch path **and** the pinned history scrollbar (CROW-1020) live only in `app.js`. The other two front-ends are single-terminal surfaces with neither, so the gap is structurally unreachable there — both confirmed:

- `web/terminal.html` — standalone debug page, no session switch, no `.has-scrollback`; already fits on `resize` and on WS open (`sendResize`).
- `CrowTerminal/…/terminal.html` — the retired macOS WKWebView surface (`crowInput`/`crowWrite` bridge, no WebSocket, no `select-window`); per app.js's own note it is "no longer a front-end to stay in sync with."

## Test plan

- Extends `session-switch-attach.test.js`: an in-place agent switch now (a) schedules a fit — parity with the reload attach, (b) evaluates the history scrollbar on switch rather than leaving it to a manual refresh, (c) still skips the replay (no CROW-1035/1048 regression), and (d) clears a stale bar when the switched-in surface has no history.
- `npm run test:ci` in `Packages/CrowDaemon/web-tests` — all suites green (terminal-scrollbar, terminal-reload, session-switch-attach included).
- Manual (ticket test plan): switch into a session with scrollback → the correctly-sized, working scroll bar shows immediately, no manual refresh.

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js
@@ -6519,6 +6519,22 @@ function switchAgentWindow(win) {
   applySurfaceScrollback();
   resetScrollbackSync();
   selectWindow(win, { replay: false });
+  // CROW-1091: run the SAME sizing step the reload/connect attach runs, which
+  // this in-place path had skipped. `reloadTerminal`/onopen call `applyTermFit()`
+  // on attach — and `fitAddon.fit()` re-syncs xterm's scroll area, which is what
+  // gives the history scrollbar its geometry — so a shell switch already resized
+  // itself. The agent in-place switch (CROW-1035) did neither, so a switched-in
+  // surface kept the previous window's grid geometry and its scrollbar sizing
+  // until a manual reload re-ran the fit — the reported "no scroll bar until
+  // refresh." Fit here too (deferred a frame via fitTerminal so it measures the
+  // pane at its settled post-switch layout — the tab bar / header height differs
+  // between session kinds), and re-check the pinned-visible class right away so it
+  // reflects the switched-in buffer instead of the previous window's. No replay
+  // and no new PTY, so this stays clear of the CROW-1035 caret jump and CROW-1048
+  // chrome stacking a full reload would reintroduce; a same-size switch fits to
+  // identical cols/rows, so applyTermFit sends no SIGWINCH (#637).
+  fitTerminal();
+  updateTerminalScrollbar();
   // Do not arm CROW-1027 here. In-place switch has no new PTY and no
   // SIGWINCH, so the mid-reflow one-screen capture that heal exists for
   // cannot happen. A second select-window 250ms later (the fake-xterm
--- a/Packages/CrowDaemon/web-tests/session-switch-attach.test.js
+++ b/Packages/CrowDaemon/web-tests/session-switch-attach.test.js
@@ -22,6 +22,9 @@ const epilogue = `
   set term(v){ term = v; },
   set activeTerminal(v){ activeTerminal = v; },
   get activeTerminal(){ return activeTerminal; },
+  updateTerminalScrollbar(){ return updateTerminalScrollbar(); },
+  get fitScheduled(){ return fitScheduled; },
+  set fitScheduled(v){ fitScheduled = v; },
   SCROLLBACK_HEAL_MS,
   SCROLLBACK_HEAL_MAX,
   TERM_BUFFER_CLEAR,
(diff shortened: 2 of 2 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- Change is confined to client-side terminal layout and DOM class toggling; no new network surface, credential handling, or server interaction.
- `applyTermFit()` retains existing guards (hidden/unfocused document, zero-size container, same-size dedup before SIGWINCH) so the added `fitTerminal()` call does not widen the resize storm surface.

**Concerns:**
- None identified.

### Code Quality
- **Green** — Root-cause analysis is sound: CROW-1035's in-place `select-window` path skipped the `applyTermFit()` step that reload/connect attach runs, leaving stale xterm scroll geometry until a manual reload. Adding `fitTerminal()` + `updateTerminalScrollbar()` in `switchAgentWindow` restores parity without reintroducing replay or a new PTY (`Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.js:6536-6537`).
- **Green** — Deferred fit via `requestAnimationFrame` matches established layout-settling practice elsewhere in `app.js`; immediate scrollbar class refresh correctly clears stale `has-scrollback` from the prior window while `term.onScroll`/`onRender` continue to drive updates as live output arrives.
- **Green** — Tests extend `session-switch-attach.test.js` with focused CROW-1091 cases: fit scheduling, scrollbar evaluation, replay still `false`, and stale-class clearing — all passing locally (`npm run test:ci` in `Packages/CrowDaemon/web-tests`, 0 failures after `npm install`).

### Summary Table

---

Issue #1125, opened 2026-08-27
Crow: Corveil token health/refresh + reconnect UX


dhilgaertner:
Part of epic: https://github.com/corveil/crow/issues/1117

**Token health/refresh + reconnect UX.**

**Acceptance**
- Background refresh of the access token before expiry.
- Revoked/expired connection surfaces a clear "Reconnect" state in the Integrations tab and `crow corveil status`.

**Pointers:** K1 client + K4 UI.

Issue #1122, opened 2026-08-27
Crow: Settings → Integrations tab + Corveil card


dhilgaertner:
Part of epic: https://github.com/corveil/crow/issues/1117

**Settings → Integrations tab + Corveil card.**

**Acceptance**
- `['integrations','Integrations']` added to `TABS` + `renderIntegrations(body)` dispatch in `renderTab`.
- Corveil card: Connect / Disconnect, connected identity + base URL, org list; editable only when `isLocal` (read-only remote), matching existing secret controls.

**Pointers:** `Packages/CrowDaemon/.../Resources/web/settings.js` (`TABS` L34, `renderTab` L313, `postConfig`/`gatewayEditor` patterns).

Issue #1121, opened 2026-08-27
Crow: org listing + one-key-per-org provisioning


dhilgaertner:
Part of epic: https://github.com/corveil/crow/issues/1117

**Org listing + one-key-per-org provisioning.**

Using the connection token: list orgs and, on org selection, mint/reuse exactly one gateway key per org.

**Acceptance**
- `GET /api/me/organizations` populates the org list (cached).
- Selecting an org mints one `sk-citadel-…` key via `POST /api/keys` (reused across workspaces bound to that org); key value stored as a secret (per-org ref).
- Deselect/rotate cleanly re-provisions or revokes; no duplicate keys per org.

**Pointers:** depends on C2 + C4; store per-org key in `corveilConnection`.

Issue #1120, opened 2026-08-27
Crow: local-only write path for the Corveil connection + CLI verbs


dhilgaertner:
Part of epic: https://github.com/corveil/crow/issues/1117

**Local-only write path for the connection + CLI verbs.**

OAuth/refresh tokens are credentials → written only through the local-only doors, mirroring `gateway-*` / `web-password-*` / `mcp-token-*`.

**Acceptance**
- `POST /config/corveil-connection` (SecretRoutes, `gateOK` = Origin + `isLocalDirect`) and a gated RPC listed in `localOnlyDenial`.
- `RPCLanePolicy` lane + `ParityLedger` entries added (build fails otherwise).
- CLI: `crow corveil connect | status | disconnect | orgs`.

**Pointers:** `RPCWebSocketHandler.swift` (`localOnlyDenial` L319), `RPCLanePolicy.swift` (L33), `Parity/ParityLedger.swift`, `Packages/CrowCLI/.../Commands/`.

PR #1128, opened 2026-08-27
Add Corveil OAuth client: DCR + PKCE + loopback callback + refresh (CROW-1119)

### LINKED ISSUE #1119: Crow: Corveil OAuth client (DCR + PKCE + loopback callback + refresh)

**Corveil OAuth client: DCR + PKCE authorize + loopback callback + refresh.**

Net-new (Crow has zero OAuth code today). Implement against Corveil's DCR AS: `POST /mcp/oauth/register` (self-register with the loopback redirect + `token_endpoint_auth_method=none`), PKCE S256 authorize (open browser to Corveil), a loopback callback route, `POST /mcp/oauth/token` exchange, and refresh.

**Acceptance**
- New callback route (e.g. `GET /integrations/corveil/callback`) in a `CorveilIntegrationRoutes.swift`, mounted in `CrowDaemon.swift`, **gated local-direct** (loopback + no XFF).
- End-to-end: Connect opens browser → consent → callback → tokens stored (via the local-only door from K2).
- Access-token refresh works; PKCE `code_verifier`/`state` validated.

**Pointers:** `Packages/CrowDaemon/.../CrowDaemon.swift` (route mounting L534-569), `SecretRoutes.swift`/`CorveilRoutes.swift` as templates, Hummingbird `router.get`.

dhilgaertner:
Closes #1119

Net-new OAuth 2.0 client for the Corveil **Connect** flow (epic #1117). Crow had zero OAuth code before this; it now self-registers a public PKCE client against Corveil's own DCR authorization server (`/mcp/oauth/*`) and stores the resulting user-scoped tokens locally.

## What's here
- **`CorveilOAuthClient`** — RFC 7591 Dynamic Client Registration (public client, `token_endpoint_auth_method: none`), RFC 7636 **S256 PKCE**, RFC 6749 authorization-code exchange + refresh-token rotation, against `{base}/mcp/oauth/{register,authorize,token}`. Request/response shapes were verified against the corveil backend (`go/internal/handler/oauth_{register,authorize,token}.go`). Transport is injected, so the whole flow is exercised without a live Corveil.
- **`CorveilConnectionStore`** — a single-use, TTL-bounded pending-authorization store (`state → code_verifier`) that also *is* the `state` validation; the **local-only** `ConfigStore` write door for `corveilConnection`; and a load→refresh→store refresh helper.
- **`CorveilIntegrationRoutes`** (mounted in `CrowDaemon`):
  - `POST /integrations/corveil/connect` — gated `gateOK` (local-direct **+** same-origin CSRF), like the other secret writes. Registers, opens the browser to Corveil, returns the authorize URL.
  - `GET /integrations/corveil/callback` — gated **local-direct** (loopback + no `X-Forwarded-For`), per the ticket. Validates `state`, exchanges the `code` with the PKCE `code_verifier`, stores the tokens, renders a result page.

## Acceptance
- ✅ New `GET /integrations/corveil/callback` in `CorveilIntegrationRoutes.swift`, mounted in `CrowDaemon.swift`, gated local-direct (loopback + no XFF).
- ✅ End-to-end: Connect opens the browser → consent → callback → tokens stored via the local-only config door (`SettingsSecrets` names the OAuth flow as the author of `corveilConnection`, CROW-1118).
- ✅ Access-token refresh works; PKCE `code_verifier`/`state` validated (state via the single-use pending store; verifier presented at exchange; challenge pinned against the RFC 7636 test vector).

## Tests
Client unit coverage (PKCE incl. RFC 7636 vector, DCR, exchange, refresh, OAuth/HTTP/transport error mapping), pending-store / persistence / refresher, and **live-server** route gating + full `connect → callback → stored` end-to-end. All 31 new tests pass.

## Scope
Stays strictly within #1119. Parallel siblings: CLI write verbs (#1120), org listing / one-key-per-org provisioning (#1121), the Integrations UI (#1122), and refresh scheduling / reconnect UX (#1125). The requested scopes (`orgs.read keys.provision`) are what the connected app needs; the backend may downscope until its sibling tickets land.

## Note
Two **pre-existing** `CrowDaemon` test failures — `RPCLanePolicyTests` (`terminal-set` / `corveil-verify` / `corveil-reinstall-skill` declared as ledger writes without a lane) and `WebNotificationCenterTests.bootCatchResetsTheHistory` — are unrelated to this change and reproduce on a clean checkout of the branch base (verified by stashing all of this PR's changes and re-running).

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/CorveilConnectionStore.swift
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/CorveilConnectionStore.swift
@@ -1,0 +1,184 @@
+import CrowCore
+import CrowPersistence
+import Foundation
+
+/// The three moving parts that sit *around* ``CorveilOAuthClient`` for the
+/// Connect flow (CROW-1119): the in-flight-authorization store that ties a
+/// browser callback back to its PKCE verifier, the local-only config write that
+/// persists the resulting tokens, and the load→refresh→store helper that renews
+/// an access token.
+///
+/// The client itself is pure HTTP; everything here is Crow-side state.
+
+// MARK: - In-flight authorization
+
+/// One pending authorization, created when Connect opens the browser and consumed
+/// (once) when the loopback callback arrives. Carries the PKCE `codeVerifier` and
+/// the `state` that the callback is validated against, plus the context the token
+/// exchange and the config write need.
+struct CorveilPendingAuthorization: Sendable, Equatable {
+    let state: String
+    let codeVerifier: String
+    let clientID: String
+    let redirectURI: String
+    let baseURL: String
+    let registrationAccessToken: String
+    let scope: String
+    let createdAt: Date
+}
+
+/// Holds pending authorizations keyed by `state` between the Connect POST and the
+/// loopback callback. Single-use (``consume(state:now:)`` removes the entry) and
+/// TTL-bounded, so a state can't be replayed and abandoned flows don't accumulate.
+///
+/// `state` is the OAuth anti-forgery value **and** the lookup key: a callback whose
+/// `state` isn't present (never issued, already consumed, or expired) is rejected,
+/// which is exactly the PKCE/`state` validation the ticket calls for. Lock-guarded
+/// so the concurrent HTTP handlers can share one instance.
+final class CorveilPendingAuthStore: @unchecked Sendable {
+    /// How long a browser has to complete consent before the pending state expires.
+    static let defaultTTL: TimeInterval = 10 * 60
+
+    private let lock = NSLock()
+    private var byState: [String: CorveilPendingAuthorization] = [:]
+    private let ttl: TimeInterval
+
+    init(ttl: TimeInterval = defaultTTL) { self.ttl = ttl }
+
+    /// Record a freshly-started authorization.
+    func put(_ pending: CorveilPendingAuthorization) {
+        lock.lock(); defer { lock.unlock() }
+        byState[pending.state] = pending
+    }
+
+    /// Remove and return the authorization for `state`, or nil if it was never
+    /// issued, already consumed, or has expired. Single-use by construction.
+    func consume(state: String, now: Date = Date()) -> CorveilPendingAuthorization? {
+        lock.lock(); defer { lock.unlock() }
+        guard let pending = byState.removeValue(forKey: state) else { return nil }
+        guard now.timeIntervalSince(pending.createdAt) <= ttl else { return nil }
+        return pending
+    }
+
+    /// Drop expired entries (called periodically by the daemon so long-abandoned
+    /// flows don't leak memory).
+    func prune(now: Date = Date()) {
+        lock.lock(); defer { lock.unlock() }
+        byState = byState.filter { now.timeIntervalSince($0.value.createdAt) <= ttl }
+    }
+
+    /// Current count — for tests.
+    var count: Int {
+        lock.lock(); defer { lock.unlock() }
+        return byState.count
+    }
+}
+
+// MARK: - Persistence (the local-only write door)
+
+/// Writes the Corveil connection into `config.json`.
+///
+/// This IS the local-only door the callback stores tokens through: it is reachable
+/// only from ``CorveilIntegrationRoutes`` handlers that have already passed the
+/// local-direct gate, and it writes the OAuth token strings that
+/// `SettingsSecrets.strippedForTransport` blanks — so a browser can neither read
+/// nor set them (corveil/crow#1118). `SettingsSecrets` names the OAuth flow as a
+/// legitimate author of `corveilConnection`; the CLI verbs are a second author
+/// (corveil/crow#1120), which is why the read-modify-write goes through
+/// `ConfigStore.withConfigLock`, shared with every other config writer.
+enum CorveilConnectionPersistence {
+    /// Store the tokens from a fresh connect, merging into any existing connection
+    /// (a reconnect) so the connected user and per-org key metadata a later step
+    /// populated (corveil/crow#1121) survive a token refresh-by-reconnect. Base URL,
+    /// client id, and the whole OAuth block are overwritten.
+    static func store(
+        devRoot: String,
+        baseURL: String,
+        clientID: String,
+        tokens: CorveilOAuthTokens
+    ) throws {
+        try ConfigStore.withConfigLock {
+            var config = ConfigStore.loadConfig(devRoot: devRoot) ?? AppConfig()
+            var connection = config.corveilConnection ?? CorveilConnection()
+            connection.baseURL = baseURL
+            connection.clientID = clientID
+            connection.oauth = tokens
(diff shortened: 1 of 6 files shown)


dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

- **Existing pathway:** local-only HTTP writes via `SecretRoutes` / `CorveilRoutes` / `AutostartRoutes` (`gateOK` = local-direct + same-origin; callback uses `isLocalDirect` only, as a top-level redirect has no `Origin`); persistence through `ConfigStore.withConfigLock`; secrets already blanked/restored by `SettingsSecrets` (CROW-1118). OAuth itself is net-new — Crow had no client — and this PR extends those doors rather than adding a parallel RPC/config-write path.
- Mount in `CrowDaemon.swift` matches the surrounding local-only routes, with a periodic prune of the in-memory pending-auth store. No architecture reject.

### Security Review
**Strengths:**
- Public-client DCR (`token_endpoint_auth_method: none`) + PKCE S256 with 32-byte verifiers, `state` as both anti-forgery and single-use lookup key, TTL-bounded pending store, RFC 7636 Appendix B vector pinned in tests.
- Connect is `SecretRoutes.gateOK`; callback is `isLocalDirect` (loopback + no `X-Forwarded-For`) before any token exchange — live-server tests assert the backend is not reached when the gate fails.
- Callback HTML runs `escape()` on title and message (including provider `error_description`), so that string cannot inject markup.
- Tokens are written only through the local-only config door (`0o600`); `SettingsSecrets` already blanks the three OAuth strings for transport.
- Redirect URI is `http://127.0.0.1:{port}/…` (RFC 8252 §7.3), matching the loopback-peer gate rather than `localhost`.

**Concerns:**
- **Yellow — `intValue` traps the process on an out-of-range `expires_in`.** Confirmed against `Packages/CrowDaemon/Sources/CrowDaemon/CorveilOAuthClient.swift:311-314` (called from `postToken` at `:274` on every successful exchange and refresh): `JSONSerialization` yields an `NSNumber` that fails `as? Int` for `1e20` / `1e308` and matches `as? Double`; `Int(double)` then fatal-errors. Reproduced with a throwaway that ran this same `switch` over `JSONSerialization.jsonObject` of `{"access_token":"x","expires_in":1e20}` — process aborted with `Fatal error: Double value cannot be converted to Int because the result would be greater than Int.max`. A 2xx token body from a buggy or hostile AS (user-supplied `baseURL`) would kill `crowd` after the pending `state` is already consumed and before tokens are stored. Return `nil` unless the value is finite and inside `Int.min...Int.max` (same for a numeric string).

### Code Quality
- Client unit tests (PKCE, DCR, exchange, refresh, error mapping, pending store, persistence, refresher) plus live-server gating and connect→callback e2e are the right shape. `swift build` succeeded; `swift test --package-path Packages/CrowDaemon --filter Corveil` — 38 tests in 3 suites passed.
- **Green:** `Endpoints` accepts any scheme with a host (`file://localhost`, `javascript://alert.com`, `slack://open` all parse). DCR over those schemes should fail before `openBrowser`, so this is hardening — restrict to `http`/`https`.
- **Green:** PKCE/`state` entropy uses per-byte `UInt8.random` (`CorveilOAuthClient.swift:347-350`). That is still `SystemRandomNumberGenerator`, but `PasswordHash.randomBytes` is the existing CSPRNG helper with the explicit comment; reuse it so this path cannot drift.
- **Green:** The callback result page has no CSP / `X-Frame-Options`, unlike `index.html`. Escaping is the real control; headers would make a future interpolation slip non-exploitable.

### Summary Table

---

dgershman (approved):
## Code & Security Review

Re-review of `5b0c62c` after the prior request-changes on `2220016`.

### Critical Issues (if any)
None. The previous Yellow (`intValue` trapping on an out-of-range `expires_in`) is fixed: `CorveilOAuthClient.intValue` now returns nil unless the value is finite and inside the `Int` range (`CorveilOAuthClient.swift:316-325`). Confirmed by `expiresInOutOfRangeIsIgnoredNotFatal`, which passed in this run.

- **Existing pathway:** local-only HTTP writes via `SecretRoutes` / `CorveilRoutes` / `AutostartRoutes` (`gateOK` = local-direct + same-origin; callback uses `isLocalDirect` only, because a top-level redirect has no `Origin`); persistence through `ConfigStore.withConfigLock`; secrets already blanked/restored by `SettingsSecrets` (CROW-1118). OAuth itself is net-new and extends those doors rather than adding a parallel RPC/config-write path.
- Mount in `CrowDaemon.swift` matches the surrounding local-only routes, with a periodic prune of the in-memory pending-auth store. No architecture reject.

### Security Review
**Strengths:**
- Public-client DCR (`token_endpoint_auth_method: none`) + PKCE S256 with 32-byte verifiers via `PasswordHash.randomBytes`, `state` as both anti-forgery and single-use lookup key, TTL-bounded pending store, RFC 7636 Appendix B vector pinned in tests.
- Connect is `SecretRoutes.gateOK`; callback is `isLocalDirect` (loopback + no `X-Forwarded-For`) before any token exchange — live-server tests assert the backend is not reached when the gate fails.
- Callback HTML runs `escape()` on title and message (including provider `error_description`) and sends `Content-Security-Policy: default-src 'none'; …; frame-ancestors 'none'` plus `X-Frame-Options: DENY` (`CorveilIntegrationRoutes.swift:256-264`).
- Tokens are written only through the local-only config door (`0o600`); `SettingsSecrets` already blanks the three OAuth strings for transport.
- Redirect URI is `http://127.0.0.1:{port}/…` (RFC 8252 §7.3), matching the loopback-peer gate rather than `localhost`.
- `Endpoints` now rejects non-`http`/`https` schemes (`CorveilOAuthClient.swift:96-101`).

**Concerns:**
- **Green — `http` is allowed for any host, not only loopback.** `Endpoints` accepts `http://corveil.test` (`CorveilOAuthClientTests.swift:104`); the comment says "loopback dest" but the check is scheme-only. A mistyped `http://` production URL would exchange tokens in the clear. Restricting `http` to loopback hosts (`127.0.0.1` / `localhost` / `::1`) would match the comment; `https` stays unrestricted. Not blocking — the user supplies the base URL.
- **Green — callback response has no `Cache-Control: no-store`.** The authorization code lives in the GET query string (`CorveilIntegrationRoutes.swift:133-135`). The HTML body does not echo it, and the request is loopback, so this is hardening: `Artifacts.swift` already uses `no-store` for secret-adjacent responses.

### Code Quality
- Prior Greens (scheme allowlist, `PasswordHash.randomBytes`, CSP / `X-Frame-Options`) are implemented and tested. Do not re-raise them.
- Client unit tests (PKCE, DCR, exchange, refresh, error mapping, pending store, persistence, refresher) plus live-server gating and connect→callback e2e remain the right shape. `swift build --package-path Packages/CrowDaemon` succeeded; `swift test --package-path Packages/CrowDaemon --filter Corveil` — 40 tests in 3 suites passed, including the new overflow/scheme/CSP assertions.

### Summary Table

---

Issue #1124, opened 2026-08-27
Crow: wire logsync + gateway from the Corveil connection


dhilgaertner:
Part of epic: https://github.com/corveil/crow/issues/1117

**Wire logsync + gateway from the connection.**

**Acceptance**
- Enabling an org for a workspace can auto-set `uploadSessionLogs` using that org's key; `GatewayResolver` + `LogSyncCollector` consume the derived gateway with no changes.
- Key rotation propagates to both the gateway header and the log-upload credential.

**Pointers:** `Packages/CrowEngine/.../SessionService.swift` (gateway resolution), `Packages/CrowDaemon/.../LogSyncCollector.swift` (`resolvedUpload`/`corveilAPIKey`).

Issue #1117, opened 2026-08-27
Epic: First-class Corveil integration — Connect (OAuth) + org-scoped gateway & log wiring


dhilgaertner:
# First-class Corveil integration — Connect (OAuth) + org-scoped gateway & log wiring

Replace the manual Corveil wiring in Crow (hand-entered AI-gateway base URL + `x-citadel-api-key` header, per-workspace log-shipping toggle) with a first-class **Integrations → Corveil** experience: a **Connect** button runs an OAuth flow against Corveil, after which Crow knows the user's Corveil orgs and exposes **org dropdowns** in the Manager and per-workspace gateway settings, auto-provisioning the per-org gateway key and wiring log shipping. This also becomes the base for future Corveil features in Crow.

## Decisions (from design review)
- **Connected App now** — build the real OAuth grant against Corveil's own DCR authorization server (not the phased/thin path). Crow self-registers via Dynamic Client Registration.
- **One auto-provisioned key per org** — Crow mints one `sk-citadel-…` gateway key per selected org via `POST /api/keys` and reuses it across workspaces bound to that org.
- **All three repos** in scope — `corveil` (backend), `crow` (app), `corveil-cloud-terraform` (WAF/env).
- **Local-only Connect** — OAuth uses a `127.0.0.1` loopback callback; Connect runs on the crowd host (same constraint as today's gateway editor, which already goes read-only for remote sessions).

## Architecture summary
- Corveil already runs its own OAuth Authorization Server (the DCR flow at `/mcp/oauth/*`), so Crow talks to Corveil (not WorkOS directly) and **self-registers** — no WorkOS redirect-URI change, no manual client provisioning. WorkOS is only the human login *inside* the authorize step.
- Net-new backend: a **user-scoped, cross-org** token (today's `crv_at_` is single-org-bound and rejected on `/api/*`) that can reach `GET /api/me/organizations` (the dropdown) and a constrained `POST /api/keys` (auto-mint the per-org key).
- Crow-side: a new **`corveilConnection`** config object is the source of truth; the existing `WorkspaceGateway` + logsync configs are **generated** from it (org picker → writes the gateway header + enables uploads), so gateway resolution and the log collector are unchanged.

## Critical path
`C1 → C3 → C4` (backend) gates `K3`. `K0/K1/K2/K4` (Crow client + UI) can proceed in parallel against a mocked backend. `T1/T2` are small and land alongside the backend.

## Security note
[...]

PR #1131, opened 2026-08-27
Corveil token health/refresh + reconnect state (CROW-1125)

### LINKED ISSUE #1125: Crow: Corveil token health/refresh + reconnect UX

dhilgaertner:
Closes #1125

Part of the first-class Corveil integration epic (#1117). Builds on the OAuth client (#1119) and the local-only connection write path (#1120): wires the background refresh loop and the health/reconnect state those left to this ticket.

## What

**Background refresh before expiry.** `CorveilTokenRefreshWatcher` runs for the daemon's lifetime (5-min cadence, like the pending-auth prune). Each tick loads the stored connection and, when the access token is within 10 min of expiry (margin > cadence, so a due token is always caught first), refreshes it through the existing `CorveilConnectionRefresher` primitive and persists the rotated tokens.

**A clear Reconnect state.** The watcher records the outcome as connection health:
- a definitively-rejected grant (`invalid_grant` / `invalid_client`) **latches** `needsReconnect` → state `revoked`;
- a transient network / 5xx error only records a diagnostic and retries next tick — it must not push a working connection into "Reconnect";
- success clears health and stamps `lastRefreshAt`.

`CorveilConnection.healthState(now:)` derives one of `connected` / `expired` (past expiry, refresh not keeping up) / `revoked` / `disconnected` (revoked wins over a clock expiry, since it's the sooner, definitive signal). `crow corveil status` now returns `state`, `needs_reconnect`, `last_refresh_at`, `last_refresh_error` alongside the existing non-secret fields.

**Available to the Integrations tab.** `CorveilConnectionHealth` carries no token, so it passes through `SettingsSecrets` untouched — the read-only Integrations view (#1122) reads `health.needsReconnect` + the token expiry from the stripped config and keys its Reconnect affordance off the same signal. (This PR does not build that tab — #1122 owns the UI.)

## Notes / scope
- Health is server-computed and owned by the refresher: reset to healthy on any (re)connect (replacing tokens invalidates a prior observation), updated in place by the watcher. Parity ledger rows for the three fields are read-via-`corveil status`, write-exempt.
- Revocation surfaces at the next scheduled refresh (near expiry) — refresh rotates the refresh token, so probing every tick regardless of expiry would churn tokens; margin-based refresh is the correct mechanism. Gateway-failure-driven detection is out of scope (#1124).

## Tests
- Watcher: refreshes a due token + stamps health; skips a fresh token (no network); `invalid_grant` → `revoked` with the stale token left in place; 5xx → retry, not revoked; no-connection / no-refresh-token no-ops; `isDue` / `isGrantRejection` units.
- Persistence: `recordRefreshSuccess`/`recordRefreshFailure` (latch-only-when-asked, never clears on transient), `store` resets health, disconnect-wins reload.
- Model: `healthState` across all four states, tolerant decode of a pre-health connection, round-trip; `SettingsSecrets` keeps health through a strip.
- Status payload state fields; parity gate (ledger ⇔ AppConfig) and `make parity` green.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -165,7 +165,7 @@ The local-only write path for the **Corveil OAuth connection** (CROW-1120) — t
 ```
 crow corveil connect [--base-url URL] [--client-id ID] [--user-id ID] [--user-email E] [--user-name N] [--access-token T] [--refresh-token T] [--registration-access-token T] [--access-token-expires-at ISO8601]
                                                 → status payload + {"saved":true}
-crow corveil status                             → {connected, base_url, client_id, connected_user, org_count, has_*_token, access_token_expires_at}
+crow corveil status                             → {connected, state, needs_reconnect, base_url, client_id, connected_user, org_count, has_*_token, access_token_expires_at, last_refresh_at, last_refresh_error}
 crow corveil disconnect                         → {"saved":true,"was_connected":bool}
 crow corveil orgs                               → {"orgs":[{org_id,org_name,key_id,key_prefix,created_at}],"count":N}
 ```
@@ -173,6 +173,7 @@ crow corveil orgs                               → {"orgs":[{org_id,org_name,ke
 - All four are **local-only** on `/rpc`, like `gateway`/`web-password`/`mcp token`: `connect` stores OAuth tokens and `disconnect` clears them, and the two reads are gated alongside the writes so the whole connection is one local-only surface. The CLI is unaffected — it goes over the Unix socket.
 - `connect` is a **merge**: every field is optional, and a blank/omitted one keeps the stored value, so a token refresh can restate only `--access-token` + `--access-token-expires-at`. The merged result needs at least a client id and an access token; `orgKeys` (provisioned by corveil/crow#1121) are preserved.
 - `status`/`orgs` never return a token value — only presence booleans and non-secret metadata.
+- **Token health/refresh (CROW-1125):** a daemon-lifetime watcher renews the access token before it expires. `status.state` is the derived health — `connected` · `expired` (past expiry, refresh not keeping up) · `revoked` (a refresh was rejected `invalid_grant`/`invalid_client` — the grant is dead) · `disconnected` — and `needs_reconnect` is true for `expired`/`revoked`. `last_refresh_at`/`last_refresh_error` expose the watcher's most recent outcome. `revoked`/`disconnected` mean rerun Connect; `expired` self-heals once the daemon can reach Corveil again. The Integrations tab (corveil/crow#1122) keys its Reconnect affordance off the same `needs_reconnect`.
 - `disconnect` clears the local record; revoking the per-org gateway keys on the Corveil side is a separate step (corveil/crow#1121).
 
 ### Job Commands
--- a/Packages/CrowCLI/Tests/CrowCLITests/ParityGateTests.swift
+++ b/Packages/CrowCLI/Tests/CrowCLITests/ParityGateTests.swift
@@ -213,7 +213,9 @@ struct ParityGateTests {
                 ],
                 oauth: CorveilOAuthTokens(
                     accessToken: "at", refreshToken: "rt",
-                    registrationAccessToken: "rat", accessTokenExpiresAt: Date()))
+                    registrationAccessToken: "rat", accessTokenExpiresAt: Date()),
+                health: CorveilConnectionHealth(
+                    lastRefreshAt: Date(), lastRefreshError: "probe", needsReconnect: true))
         )
     }
 
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -547,6 +547,11 @@ public struct CorveilConnection: Codable, Sendable, Equatable {
     /// OAuth token material — the secrets. Stripped for transport by
     /// `SettingsSecrets`.
     public var oauth: CorveilOAuthTokens
+    /// Token-refresh health, owned by the background refresher (CROW-1125). Not a
+    /// secret — it carries no token, only the *outcome* of the last refresh — so it
+    /// passes through `SettingsSecrets` untouched and the read-only Integrations
+    /// view can render a "Reconnect" state from it.
+    public var health: CorveilConnectionHealth
 
     /// Whether the connection has enough to be usable — a client id and an access
     /// token. An all-empty block (e.g. a partially-written record) reads as unset.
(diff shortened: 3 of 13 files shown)


dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

- **Existing pathway:** `CorveilConnectionRefresher` + `CorveilConnectionPersistence` (`ConfigStore.withConfigLock`) for token I/O, and the pending-auth prune `Task` loop in `CrowDaemon` for the scheduler. This PR extends those rather than inventing a second connection store.
- Watcher ticks are self-contained (load → maybe refresh → persist) and unit-tested against a stub transport — that matches how `CorveilOAuthClient` was built.
- **Green:** `CorveilTokenRefreshWatcher.tick` inlines the load/refresh/merge path instead of calling `CorveilConnectionRefresher.refresh`, and the primitive still persists via `updateTokens` (no health stamp). A small extension of `refresh()` to use `recordRefreshSuccess` would keep one writer. Not wrong, just a fork.

### Security Review
**Strengths:**
- `statusJSON` still never returns token values; health fields are non-secret by construction.
- `SettingsSecrets` keeps passing health through while blanking the three OAuth strings; `preservingSecrets` still restores the whole connection from disk so a browser cannot latch or clear `needsReconnect`.
- `isGrantRejection` is tight (only `invalid_grant` / `invalid_client` on `CorveilOAuthClient.Failure.oauth`) — a 5xx or transport error cannot revoke the grant.
- Local-only write path is unchanged; logs do not include token material.

**Concerns:**
- None that leak credentials. `lastRefreshError` can carry up to ~200 characters of a token-endpoint body into `config.json` and the web config (via `Failure.description`). Diagnostic only; not a blocker.

### Code Quality
- **Yellow — in-flight refresh vs reconnect clobbers the new grant.** `recordRefreshSuccess` / `recordRefreshFailure` in `Packages/CrowDaemon/Sources/CrowDaemon/CorveilConnectionStore.swift` reload under the config lock and no-op when the connection is **nil** (disconnect-wins, as documented). They do **not** check that the stored grant is still the one this tick presented. `CorveilTokenRefreshWatcher.tick` (`Packages/CrowDaemon/Sources/CrowDaemon/CorveilTokenRefreshWatcher.swift`) loads the connection, awaits the token HTTP call unlocked, then persists.

  That is the Connect path this ticket exists to serve: the watcher is retrying a due/expired grant while the user completes Reconnect.

  **Reproduced** with a temporary test (deleted after): parked the stub transport, called `CorveilConnectionPersistence.store` with `fresh-at` / `fresh-rt`, then released the transport with `{"error":"invalid_grant"}`. After tick: `accessToken == "fresh-at"` **and** `health.needsReconnect == true` (error `"invalid_grant: token revoked"`). The new grant is immediately shown as revoked. The success path has the same missing identity check and would overwrite the new oauth block with tokens minted from the old refresh token.

  **Fix:** pass the presented refresh token (or the pre-refresh access+refresh pair) into both record methods and return `false` without writing if the stored connection no longer matches — reconnect-wins, same as disconnect-wins. Cover it with the race test above.

- **Green:** `CorveilConnectionRPC.merge` resets health only because it omits `health:` and hits `CorveilConnection`'s default. `store()` is explicit. Make the merge reset explicit so a later `health: base.health` does not re-break CLI reconnect.
- **Green:** the daemon loop sleeps a full `tickInterval` before the first tick (`Packages/CrowDaemon/Sources/CrowDaemon/CrowDaemon.swift`), same as the prune loop. Catch-up of an already-expired token therefore waits up to 5 minutes after boot. Fine if intentional; an immediate first tick would match the "catch up on boot" comment on `refreshMargin`.

### Tests
`swift build` succeeded. `CorveilOAuthClientTests` (34), `CorveilConnectionRPCTests` (19), and `corveilHealth*` (2) all passed. No Go module in this repo.

### Summary Table

---

dhilgaertner:
Thanks @dgershman — addressed in e7fd84b2.

**Yellow — in-flight refresh vs reconnect clobber (fixed).** `recordRefreshSuccess` / `recordRefreshFailure` now take `expectedRefreshToken` and write only if the stored refresh token still matches the grant this tick presented — reconnect-wins, the sibling of disconnect-wins. `tick` passes the grant it loaded, and returns a new `Outcome.superseded` when the write is discarded. I reproduced your exact scenario as two tests: a reconnect landing mid-flight leaves the fresh grant intact on the success path, and does **not** mark it revoked on the `invalid_grant` path (`accessToken == "fresh-at"`, `needsReconnect == false`). Also added a direct grant-mismatch no-op test.

**Green — explicit health reset in `merge` (done).** `CorveilConnectionRPC.merge` now passes `health: CorveilConnectionHealth()` with a comment, so a later `health: base.health` can't silently re-break CLI reconnect.

**Green — catch up on boot (done).** The refresh loop now ticks immediately, then sleeps, so an already-expired token is caught at startup rather than up to 5 min later — matches the `refreshMargin` comment.

**Green — one writer (`refresh()` vs `tick`) (left as-is, intentional).** I kept `CorveilConnectionRefresher.refresh` as the low-level force-refresh primitive and `tick` as the health-aware scheduled path. The reconnect-identity guard and health-stamp are specific to the scheduled watcher racing a user reconnect; folding them into the primitive would change its contract for its existing callers/tests without removing a real second store (both still go through `CorveilConnectionPersistence` + `withConfigLock`). Happy to unify if you'd prefer it in a follow-up.

**Note — `lastRefreshError` size.** Left as-is per your "not a blocker": it's already bounded (`Failure.http` caps the body at 200 chars) and carries no token material.

All corveil suites + `make parity` green.

dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

- **Existing pathway:** `CorveilConnectionRefresher` + `CorveilConnectionPersistence` (`ConfigStore.withConfigLock`) for token I/O, and the pending-auth prune `Task` loop in `CrowDaemon` for the scheduler. The watcher is the 1125 scheduler those primitives were built for; persist goes through new `recordRefreshSuccess` / `recordRefreshFailure` rather than inventing a second store.
- Follow-up `e7fd84b` addresses the prior review's reconnect-wins gap: both record methods compare `expectedRefreshToken` under the config lock, and `watcherSuccessSupersededByAReconnectMidFlight` / `watcherRevocationSupersededByAReconnectMidFlight` cover the in-flight Connect race. Tick-first-then-sleep in `CrowDaemon.swift` and the explicit `health: CorveilConnectionHealth()` reset in `CorveilConnectionRPC.merge` also land.
- **Green:** `CorveilTokenRefreshWatcher.tick` still inlines load/refresh/merge instead of calling `CorveilConnectionRefresher.refresh`, and that primitive still persists via `updateTokens` in `CorveilConnectionStore.swift` (no CAS, no health stamp). Production only uses the watcher; fold `refresh()` onto `recordRefreshSuccess` before anything else (e.g. #1124) calls it.

### Security Review
**Strengths:**
- `statusJSON` still never returns token values; health fields are non-secret by construction.
- `SettingsSecrets` keeps health through a strip while blanking the three OAuth strings; `preservingSecrets` still restores the whole connection from disk, so a browser cannot latch or clear `needsReconnect`.
- `isGrantRejection` is tight (only `invalid_grant` / `invalid_client` on `CorveilOAuthClient.Failure.oauth`) — a 5xx or transport error cannot revoke the grant.
- Local-only write path is unchanged; refresh logs do not include token material.

**Concerns:**
- **Green:** `lastRefreshError` is persisted from `Failure.description`, which for a non-OAuth HTTP failure includes `body.prefix(200)`. That string then rides `config.json` and the stripped web config. Diagnostic only; not a credential leak.

### Code Quality
- **Green:** In `CorveilTokenRefreshWatcher.tick`, `recordRefreshSuccess` sits in the same `do/catch` as `client.refresh`. A persist throw after a successful, rotating refresh is classified as a refresh failure; the next due tick then presents the old refresh token, gets `invalid_grant`, and latches revoked. Catch persist separately (retry the write, not the HTTP) if that path matters — disk failure after `saveConfig` is rare.
- **Green:** `CorveilConnectionHealth.needsReconnect` (stored latch, revoked only) and `CorveilConnectionState.needsReconnect` (derived, expired **or** revoked) share a name but not a meaning. `statusJSON`'s `needs_reconnect` is the derived flag. #1122 should key off `healthState()`, not the stored field alone.
- **Green:** `merge` now resets health explicitly, but `mergeBlankFieldsKeepStoredValues` does not assert that a latched `needsReconnect` is cleared. An assertion there would lock the CLI-reconnect fix.

### Tests
`swift build` succeeded. `CorveilOAuthClientTests` + `CorveilConnectionRPCTests` (56), `corveilHealth*` + `SettingsSecretsTests` (20), and `ParityGateTests` (7) all passed. No Go module in this repo.

### Summary Table

---

PR #1129, opened 2026-08-27
Add local-only write path + CLI verbs for the Corveil connection (CROW-1120)

dhilgaertner:
Closes #1120

Part of epic #1117 (first-class Corveil integration). The Corveil connection (CROW-1118) holds OAuth tokens, so — like the AI gateways, the web password and MCP tokens — it must be authored only through **local-only doors**, never `set-config`. This ticket adds that write path plus its CLI surface: the "door" the OAuth client (#1119) and org provisioning (#1121) persist a `corveilConnection` through.

## What's here

**Two doors, one implementation**
- `POST /config/corveil-connection` in `SecretRoutes` (`gateOK` = Origin + `isLocalDirect`) — the browser's door.
- Four RPCs — `corveil-connect` / `corveil-status` / `corveil-disconnect` / `corveil-orgs` — the CLI's door over the Unix socket.
- Decoding, the secret-safe **merge** (a blank field keeps the stored secret, so a token refresh needn't restate the refresh token; `orgKeys` are preserved), and the non-secret response shapes live in `CorveilConnectionRPC`, so the two doors can't drift.

**Local-only gating**
- All four RPCs are on `localOnlyDenial` (+ the `ParityLedger.localOnlyRPCMethods` mirror, pinned in both directions by `LocalOnlyRPCGateTests`). The two writes author a credential; the two reads are gated alongside them so the whole connection is one local-only surface — the `mcp-token-list` precedent. `status`/`orgs` never return a token value.

**Parity plumbing** (the build fails otherwise)
- `RPCLanePolicy`: `corveil-connect`/`corveil-disconnect` share the config lane; the reads take none. Also classifies three writes the Darwin-only lane gate never caught in the Linux PR lane — `terminal-set` (CROW-1085) and `corveil-verify` / `corveil-reinstall-skill` (CROW-1011).
- `ParityLedger`: 4 `rpcMethods` rows + 4 `localOnlyRPCMethods` entries; the `corveilConnection.*` config-field rows move from `noCLI` to their new verbs (token values stay read-exempt; `orgKeys[].*` stay write-exempt, provisioned by #1121).

**CLI** — `crow corveil connect | status | disconnect | orgs`, alongside the existing `verify`/`reinstall-skill`.

**Docs & tests**
- `docs/cli.md` regenerated; `docs/cli-reference.md` + `CLAUDE.md` written up.
[...]

PR #1127, opened 2026-08-27
Add corveilConnection to AppConfig (secret-safe transport) (CROW-1118)

dhilgaertner:
Closes #1118

Part of epic #1117 (first-class Corveil integration). Adds the config object that later tickets read from; stays strictly within this ticket's scope (model + secret-safe transport — no CLI verbs, no OAuth client, no gateway/logsync generation).

## What

Introduces `AppConfig.corveilConnection` as the source of truth for the Corveil **Connect** (OAuth) integration:

- `baseURL`, `clientID` (DCR-registered) — non-secret
- `connectedUser` (`CorveilConnectedUser`) — read-only signed-in identity
- `orgKeys` (`[CorveilOrgKey]`) — per-org gateway-key **metadata** only. The `sk-citadel-…` key value itself flows into the generated `WorkspaceGateway` header (sibling #1124), never stored here.
- `oauth` (`CorveilOAuthTokens`) — the secrets: `accessToken` / `refreshToken` / `registrationAccessToken` + `accessTokenExpiresAt`

## Acceptance criteria

- **Lenient decode** — every field is `decodeIfPresent … ?? default`; a missing key never traps, a partial/empty object fills from defaults, and `CodingKeys` are complete so `encode(to:)` doesn't silently drop a field (CROW-809 / CROW-814 lesson). Covered by `corveilConnectionDecodesLeniently`, `…DecodesEmptyObject`, `…DefaultsNilWhenKeyMissing`, `…SurvivesAConfigRewrite`.
- **Secret-safe transport** — `SettingsSecrets.strippedForTransport` blanks the three OAuth token strings (keeping the rest for a read-only Integrations view); `preservingSecrets` restores the whole connection from the stored config, exactly like `managerGateway`. The block is authored only by the local-only Connect flow / CLI verbs (#1120), never `set-config`, so a browser can neither read the tokens nor change/clear the connection.
- **Round-trip** — `corveilConnectionSurvivesAStripPreserveRoundTripUntouched` + `stripThenPreserveIsIdentity` prove a config with a connection survives a `set-config` untouched; `strippedCorveilTokensNeverAppearInTheSerializedConfig` proves the tokens never serialize to a browser.

## Parity

Adds `ParityLedger.configFields` rows for all 14 new leaves (all `noCLI` — write verbs land in #1120) and inhabits the reflection probe, so `ParityGateTests` (\"Ledger covers exactly AppConfig's fields\", \"reflection probe reaches every subtree\") passes.

## Testing

[...]

Issue #1123, opened 2026-08-27
Crow: org-dropdown gateway editors (Manager + per-workspace)


dhilgaertner:
Part of epic: https://github.com/corveil/crow/issues/1117

**Org-dropdown gateway editors (Manager + per-workspace).**

When a connection exists, replace the raw base-URL+headers `gatewayEditor` with an **org picker**; selecting an org derives the `WorkspaceGateway` (`baseURL` + `x-citadel-api-key`) from the connection. Keep the manual editor as an advanced fallback.

**Acceptance**
- Manager gateway (Automation tab) and per-workspace gateway (Workspaces tab) both offer the org dropdown when connected.
- Picking an org writes the derived gateway via the existing local-only POST; manual entry still available.

**Pointers:** `settings.js` `renderAutomation` (Manager gateway L818-821), `renderWorkspaceForm` (workspace gateway L1643).

PR #1132, opened 2026-08-27
Add Corveil org listing + one-key-per-org provisioning (CROW-1121)

### LINKED ISSUE #1121: Crow: org listing + one-key-per-org provisioning

dhilgaertner:
Closes #1121

Part of the first-class Corveil integration epic (#1117). Builds on the OAuth connection (#1128), the local-only write path (#1129), and the `corveilConnection` model (#1127).

## What this adds

Using the connection's OAuth bearer, list the user's Corveil orgs and, on org selection, mint or reuse **exactly one** gateway key per org.

- **`CorveilAPIClient`** — pure-HTTP client (injected transport, mirrors `CorveilOAuthClient`) for the three constrained endpoints the backend opened for Crow (`RadiusMethod/corveil#2704`, `#2706`):
  - `GET /api/me/organizations` → the memberships list
  - `POST /api/keys` → mint one `sk-citadel-…` key in a named org
  - `DELETE /api/keys/{id}` → revoke (404 tolerated as "already gone")
- **`CorveilOrgProvisioner` + `CorveilOrgListCache`** — the coordination on top:
  - **reuse, don't re-mint** — a select for an org that already has a stored key returns it with no network call. This is load-bearing: the backend deactivates the prior key on every `POST /api/keys`, so re-minting would silently invalidate every workspace bound to the old value.
  - **rotate** (`--rotate`) — force a fresh key; the backend revokes the old one as part of the mint.
  - **deselect** — revoke server-side, then drop the local metadata + secret.
  - a short TTL cache for the org list (acceptance: "populates the org list (cached)"), and proactive access-token refresh when the token is at/near expiry.
- **Secret storage** — the `sk-citadel-…` value is stored per-org in `corveilConnection.orgKeySecrets`, stripped for transport by `SettingsSecrets` (like the OAuth tokens) and preserved across a token-refresh merge. `orgKeys[]` stays pure display metadata.
- **Surface** — RPCs `corveil-list-orgs` / `corveil-select-org` / `corveil-deselect-org` and the `crow corveil list-orgs` / `select-org` / `deselect-org` verbs, all local-only (they act with a credential), with lane rules, parity-ledger rows, and reference + CLAUDE.md docs.

## Scope

Strictly the provisioning mechanism. The org-dropdown UI (#1123), the `WorkspaceGateway` generated from the stored secret (#1124), and the token health/reconnect UX (#1125) are sibling tickets and are not touched here.

## Acceptance

- ✅ `GET /api/me/organizations` populates the org list (cached) — `corveil list-orgs`, `CorveilOrgListCache`.
- ✅ Selecting an org mints one `sk-citadel-…` key via `POST /api/keys`, reused across workspaces bound to that org; value stored as a per-org secret.
- ✅ Deselect/rotate cleanly re-provisions or revokes; no duplicate keys per org.

## Testing

- `CorveilAPIClientTests` — parse, bearer auth, structured/401 error mapping, revoke-404 tolerance.
- `CorveilOrgProvisionerTests` — reuse (no re-mint), exactly-one-key-per-org, rotate, revoke + remove, no-op deselect, not-connected, cache TTL/miss, expired-token refresh.
- `SettingsSecretsTests`, `CorveilConnectionRPCTests`, `CorveilCommandParsingTests`, and the parity/lane/local-only/docs gates.

All package suites green (`CrowCore`, `CrowCLI`, `CrowDaemon`) except one pre-existing flake unrelated to this change (`WebNotificationCenterTests.bootCatchResetsTheHistory`, a web-asset test that also fails on the base commit in a multi-suite run).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -167,14 +167,18 @@ crow corveil connect [--base-url URL] [--client-id ID] [--user-id ID] [--user-em
                                                 → status payload + {"saved":true}
 crow corveil status                             → {connected, state, needs_reconnect, base_url, client_id, connected_user, org_count, has_*_token, access_token_expires_at, last_refresh_at, last_refresh_error}
 crow corveil disconnect                         → {"saved":true,"was_connected":bool}
-crow corveil orgs                               → {"orgs":[{org_id,org_name,key_id,key_prefix,created_at}],"count":N}
+crow corveil orgs                               → {"orgs":[{org_id,org_name,key_id,key_prefix,created_at}],"count":N}   the LOCAL provisioned-key metadata
+crow corveil list-orgs [--refresh]              → {"orgs":[{org_id,org_name,role,is_active,provisioned}],"count":N}   the orgs you belong to, cached
+crow corveil select-org --org ID [--name N] [--rotate]  → {"saved":true,"reused":bool,"org":{org_id,org_name,key_id,key_prefix,created_at}}
+crow corveil deselect-org --org ID              → {"saved":true,"removed":bool}
 ```
 
-- All four are **local-only** on `/rpc`, like `gateway`/`web-password`/`mcp token`: `connect` stores OAuth tokens and `disconnect` clears them, and the two reads are gated alongside the writes so the whole connection is one local-only surface. The CLI is unaffected — it goes over the Unix socket.
-- `connect` is a **merge**: every field is optional, and a blank/omitted one keeps the stored value, so a token refresh can restate only `--access-token` + `--access-token-expires-at`. The merged result needs at least a client id and an access token; `orgKeys` (provisioned by corveil/crow#1121) are preserved.
-- `status`/`orgs` never return a token value — only presence booleans and non-secret metadata.
+- All seven are **local-only** on `/rpc`, like `gateway`/`web-password`/`mcp token`: `connect` stores OAuth tokens and `disconnect` clears them, and the reads + the provisioning verbs are gated alongside the writes so the whole connection is one local-only surface. The CLI is unaffected — it goes over the Unix socket.
+- `connect` is a **merge**: every field is optional, and a blank/omitted one keeps the stored value, so a token refresh can restate only `--access-token` + `--access-token-expires-at`. The merged result needs at least a client id and an access token; `orgKeys` + their per-org secrets (provisioned by corveil/crow#1121) are preserved.
+- `status`/`orgs`/`list-orgs`/`select-org`/`deselect-org` never return a token or a key value — only presence booleans and non-secret metadata.
 - **Token health/refresh (CROW-1125):** a daemon-lifetime watcher renews the access token before it expires. `status.state` is the derived health — `connected` · `expired` (past expiry, refresh not keeping up) · `revoked` (a refresh was rejected `invalid_grant`/`invalid_client` — the grant is dead) · `disconnected` — and `needs_reconnect` is true for `expired`/`revoked`. `last_refresh_at`/`last_refresh_error` expose the watcher's most recent outcome. `revoked`/`disconnected` mean rerun Connect; `expired` self-heals once the daemon can reach Corveil again. The Integrations tab (corveil/crow#1122) keys its Reconnect affordance off the same `needs_reconnect`.
-- `disconnect` clears the local record; revoking the per-org gateway keys on the Corveil side is a separate step (corveil/crow#1121).
+- **Org provisioning (CROW-1121):** `list-orgs` lists the orgs you belong to (`GET /api/me/organizations`, cached; `--refresh` re-fetches) and flags which already have a key. `select-org` mints **one** `sk-citadel-…` gateway key per org via `POST /api/keys`, **reusing** the stored one if the org already has a key (`reused:true`, no re-mint — the backend rotates the key on each POST, so re-minting would break bound gateways); `--rotate` forces a fresh key. The key value is stored as a per-org secret in `corveilConnection`; only its metadata is ever printed. `deselect-org` revokes the key server-side and drops the local record (idempotent).
+- `disconnect` clears the local record; revoking the per-org gateway keys on the Corveil side is `deselect-org` (or a dashboard cascade on revoke).
 
 ### Job Commands
 
--- a/Packages/CrowCLI/Sources/CrowCLILib/Commands/CorveilCommands.swift
+++ b/Packages/CrowCLI/Sources/CrowCLILib/Commands/CorveilCommands.swift
@@ -32,6 +32,9 @@ public struct Corveil: ParsableCommand {
             CorveilStatus.self,
             CorveilDisconnect.self,
             CorveilOrgs.self,
+            CorveilListOrgs.self,
+            CorveilSelectOrg.self,
+            CorveilDeselectOrg.self,
         ]
     )
 
@@ -240,7 +243,9 @@ public struct CorveilOrgs: ParsableCommand {
         discussion: """
         Prints the metadata for each auto-provisioned per-org gateway key — org id \
         and name, key id, display prefix, and mint time — never the key material \
-        itself. Empty until an org is provisioned (corveil/crow#1121).
+        itself. This is the LOCAL record of what has been provisioned; use `corveil \
+        list-orgs` to see every org you could select. Empty until an org is \
+        provisioned with `corveil select-org` (corveil/crow#1121).
         """
     )
 
@@ -248,5 +253,96 @@ public struct CorveilOrgs: ParsableCommand {
 
     public func run() throws {
         printJSON(try rpc("corveil-orgs"))
+    }
+}
+
+// MARK: - Org provisioning (CROW-1121)
+
+/// `crow corveil list-orgs` — list the user's Corveil orgs from the API (cached).
+public struct CorveilListOrgs: ParsableCommand {
+    public static let configuration = CommandConfiguration(
+        commandName: "list-orgs",
+        abstract: "List the Corveil orgs you belong to (local-only)",
+        discussion: """
+        Fetches your Corveil memberships with the connection's OAuth token and \
+        prints each org's id, name, role, active flag, and whether it already has a \
+        provisioned gateway key. The result is cached briefly; pass `--refresh` to \
+        force a re-fetch. This is the list you pick from — `corveil orgs` shows only \
+        the orgs already provisioned locally. Local-only: it acts with a credential, \
+        so it runs over the Unix socket and is refused for remote web clients.
+        """
+    )
+
+    @Flag(name: .long, help: "Bypass the cache and re-fetch the org list from Corveil.")
+    public var refresh: Bool = false
+
+    public init() {}
+
+    public func run() throws {
+        var params: [String: JSONValue] = [:]
+        if refresh { params["refresh"] = .bool(true) }
+        printJSON(try rpc("corveil-list-orgs", params: params))
+    }
+}
+
+/// `crow corveil select-org` — provision (or reuse) the one gateway key for an org.
+public struct CorveilSelectOrg: ParsableCommand {
+    public static let configuration = CommandConfiguration(
+        commandName: "select-org",
+        abstract: "Provision or reuse the gateway key for a Corveil org (local-only)",
+        discussion: """
+        Mints exactly one `sk-citadel-…` gateway key for the org, reusing the stored \
+        one if the org already has a key — so every workspace bound to the org shares \
+        it. The key value is stored as a secret; only its metadata (id, prefix, mint \
+        time) is ever printed. Pass `--rotate` to force a fresh key (the old one is \
+        revoked). The org name is taken from `--name` if given, else looked up from \
+        your memberships. Local-only: it mints a credential over the Unix socket and \
+        is refused for remote web clients.
+
+          crow corveil select-org --org <org-id>
+          crow corveil select-org --org <org-id> --rotate
+        """
+    )
+
+    @Option(name: .customLong("org"), help: "Corveil organization id to provision a key for")
+    public var org: String
+
+    @Option(name: .customLong("name"), help: "Org display name to store (looked up if omitted)")
+    public var name: String?
+
+    @Flag(name: .long, help: "Revoke the existing key and mint a fresh one.")
+    public var rotate: Bool = false
+
+    public init() {}
+
+    public func run() throws {
+        var params: [String: JSONValue] = ["org_id": .string(org)]
+        let trimmedName = name?.trimmingCharacters(in: .whitespacesAndNewlines) ?? ""
+        if !trimmedName.isEmpty { params["org_name"] = .string(trimmedName) }
+        if rotate { params["rotate"] = .bool(true) }
+        printJSON(try rpc("corveil-select-org", params: params))
+    }
+}
+
+/// `crow corveil deselect-org` — revoke an org's gateway key and drop the record.
+public struct CorveilDeselectOrg: ParsableCommand {
+    public static let configuration = CommandConfiguration(
+        commandName: "deselect-org",
+        abstract: "Revoke a Corveil org's gateway key (local-only)",
+        discussion: """
+        Revokes the org's provisioned gateway key on the Corveil side and removes its \
+        local metadata and stored secret. Idempotent — deselecting an org with no key \
+        is a no-op. Local-only: it acts with a credential over the Unix socket and is \
+        refused for remote web clients.
+        """
+    )
+
+    @Option(name: .customLong("org"), help: "Corveil organization id whose key to revoke")
+    public var org: String
+
+    public init() {}
+
+    public func run() throws {
+        printJSON(try rpc("corveil-deselect-org", params: ["org_id": .string(org)]))
     }
 }
--- a/Packages/CrowCLI/Tests/CrowCLITests/CorveilCommandParsingTests.swift
+++ b/Packages/CrowCLI/Tests/CrowCLITests/CorveilCommandParsingTests.swift
@@ -23,6 +23,7 @@ struct CorveilCommandParsingTests {
         let names = Corveil.configuration.subcommands.map { $0.configuration.commandName }
         #expect(Set(names) == [
             "verify", "reinstall-skill", "connect", "status", "disconnect", "orgs",
+            "list-orgs", "select-org", "deselect-org",
         ])
     }
 
@@ -61,6 +62,47 @@ struct CorveilCommandParsingTests {
         #expect((try parse(["corveil", "status"]) as? CorveilStatus) != nil)
         #expect((try parse(["corveil", "disconnect"]) as? CorveilDisconnect) != nil)
         #expect((try parse(["corveil", "orgs"]) as? CorveilOrgs) != nil)
+    }
+
+    // MARK: - Org provisioning (CROW-1121)
+
+    @Test("list-orgs sends refresh only when the flag is set")
+    func listOrgsRefreshFlag() throws {
+        let plain = try #require(try parse(["corveil", "list-orgs"]) as? CorveilListOrgs)
+        #expect(plain.refresh == false)
+
+        let refreshed = try #require(
+            try parse(["corveil", "list-orgs", "--refresh"]) as? CorveilListOrgs)
+        #expect(refreshed.refresh == true)
+    }
+
+    @Test("select-org requires --org and maps optional name/rotate")
+    func selectOrgFlags() throws {
+        let minimal = try #require(
+            try parse(["corveil", "select-org", "--org", "org-123"]) as? CorveilSelectOrg)
+        #expect(minimal.org == "org-123")
+        #expect(minimal.name == nil)
+        #expect(minimal.rotate == false)
+
+        let full = try #require(try parse([
+            "corveil", "select-org", "--org", "org-123", "--name", "Acme Inc", "--rotate",
+        ]) as? CorveilSelectOrg)
+        #expect(full.org == "org-123")
+        #expect(full.name == "Acme Inc")
+        #expect(full.rotate == true)
+    }
+
+    @Test("select-org without --org is rejected")
+    func selectOrgRequiresOrg() {
+        #expect(throws: (any Error).self) { _ = try self.parse(["corveil", "select-org"]) }
+    }
+
+    @Test("deselect-org requires --org")
+    func deselectOrgFlags() throws {
+        let deselect = try #require(
+            try parse(["corveil", "deselect-org", "--org", "org-9"]) as? CorveilDeselectOrg)
+        #expect(deselect.org == "org-9")
+        #expect(throws: (any Error).self) { _ = try self.parse(["corveil", "deselect-org"]) }
     }
 
     // MARK: - --path
--- a/Packages/CrowCLI/Tests/CrowCLITests/ParityGateTests.swift
+++ b/Packages/CrowCLI/Tests/CrowCLITests/ParityGateTests.swift
@@ -211,6 +211,7 @@ struct ParityGateTests {
                         orgID: "org1", orgName: "Acme", keyID: "key1",
                         keyPrefix: "sk-citadel-AbC", createdAt: Date())
                 ],
+                orgKeySecrets: ["org1": "sk-citadel-AbCdEfGhIjKl"],
                 oauth: CorveilOAuthTokens(
                     accessToken: "at", refreshToken: "rt",
                     registrationAccessToken: "rat", accessTokenExpiresAt: Date()),
@@ -290,6 +291,10 @@ struct ParityGateTests {
         // presumes writes (CROW-1120).
         "corveil-status",
         "corveil-orgs",
+        // `corveil-list-orgs` lists the user's Corveil orgs from the API (cached) —
+        // a read — but `corveil-list-orgs` is neither `list-`-prefixed nor
+        // `-list`-suffixed, so the heuristic presumes a write (CROW-1121).
+        "corveil-list-orgs",
     ]
 
     /// `isWrite` is hand-set per row, deliberately, because a `get-`/`list-`
--- a/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
+++ b/Packages/CrowCore/Sources/CrowCore/Models/AppConfig.swift
@@ -541,9 +541,21 @@ public struct CorveilConnection: Codable, Sendable, Equatable {
     /// The signed-in Corveil user this connection belongs to.
     public var connectedUser: CorveilConnectedUser
     /// Metadata (never key material) for the one auto-provisioned gateway key per
-    /// Corveil org. The `sk-citadel-…` value itself lives in the generated
-    /// ``WorkspaceGateway`` header, not here.
+    /// Corveil org — org id/name, key id, display prefix, mint time. The
+    /// `sk-citadel-…` value itself is **not** here; it lives in the sibling
+    /// ``orgKeySecrets`` (a secret), from which the generated ``WorkspaceGateway``
+    /// header is populated (corveil/crow#1124). Keeping the metadata free of key
+    /// material means this array is safe to serialize to the read-only web view.
     public var orgKeys: [CorveilOrgKey]
+    /// The `sk-citadel-…` gateway-key value for each provisioned org, keyed by
+    /// Corveil org id — the **secret** half of ``orgKeys`` (CROW-1121). This is the
+    /// source of truth `corveilConnection` holds so a key is minted once per org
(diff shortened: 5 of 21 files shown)


dgershman (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

- **Existing pathway:** `CorveilOAuthClient` (injected-transport HTTP) + `CorveilConnectionPersistence` (`ConfigStore.withConfigLock`) + local-only RPCs + `SettingsSecrets` strip/restore. This PR extends that path correctly: a sibling `CorveilAPIClient`, provisioner coordination, `orgKeySecrets` as the secret half of `orgKeys`, and three new local-only verbs. It does not invent a parallel credential door or go through `set-config`.
- **Yellow — org-list cache is keyed on a field the Connect flow never sets, and `invalidate()` is dead.** `CorveilOrgListCache` is documented as keyed on `(base URL, user)` so a reconnect as a different user is a miss rather than a leak (`CorveilOrgProvisioner.swift:192-196`, `cacheKey` at `CorveilOrgProvisioner.swift:187-189`). The OAuth persist path that actually writes a fresh connection — `CorveilConnectionPersistence.store` (`CorveilConnectionStore.swift:94-108`) — overwrites `baseURL` / `clientID` / `oauth` and never touches `connectedUser`. After a browser Connect, `connectedUser.id` is empty, so every connection to the same host shares the cache key `"<baseURL>\u{1}"`. `invalidate()` (`CorveilOrgProvisioner.swift:225-232`) is documented as the disconnect hook but is never called: `corveil-disconnect` (`RPCHandlers.swift:2450-2456`) has no reference to the cache, which is constructed only inside `makeCorveilProvisioningHandlers` (`RPCHandlers.swift:2099-2101`). Confirmed against those sites — no filter/invalidation is applied on disconnect, and the OAuth store path does not populate the user id the cache depends on. A disconnect + reconnect as a different Corveil account on the same daemon can therefore be served the previous user's memberships for up to 5 minutes. Fix: hoist the cache so disconnect (RPC + HTTP clear) can `invalidate()`, and/or key on something the Connect path actually writes (`clientID` is fresh per DCR).

### Security Review
**Strengths:**
- The three new RPCs are gated local-only (`RPCWebSocketHandler.swift:372-380`) and ledgered in `ParityLedger.localOnlyRPCMethods`; they are not MCP-exported.
- `sk-citadel-…` values live in `orgKeySecrets`, are blanked by `SettingsSecrets.strippedForTransport`, and the whole `corveilConnection` block is restored verbatim on `set-config` (`SettingsSecrets.swift:72-78`, `119`). Token-refresh merge preserves `orgKeySecrets` (`CorveilConnectionRPCSupport.swift:104-105`).
- Reuse-before-remint is load-bearing and tested: a second select does not `POST /api/keys`, matching the backend's rotate-on-mint behavior.
- `Endpoints` rejects non-http(s) base URLs (same guard as `CorveilOAuthClient`); key ids are path-appended rather than interpolated.
- Revoke treats 404 as success and does not drop the local record until the server call returns, so a 5xx revoke fails closed.

**Concerns:**
- **Yellow** — see the cache identity leak above. It is membership metadata (org id/name/role), not a spendable key, but it is the leak the cache comments claim to prevent.
- **Green** — `resolveCorveilOrgName` (`RPCHandlers.swift:2550-2561`) returns `--name` without checking membership. First mint then relies entirely on the backend's `keys.provision` org check. Fine if that check is real; worth still validating `org_id` against the list even when a display name is supplied.

### Code Quality
- Provisioner, API client, secrets, CLI parsing, and parity/lane/local-only gates are covered and the Corveil suites are green (84 tests across 8 suites, plus `SettingsSecretsTests` and `CorveilCommandParsingTests`).
- **Green** — no router-level tests for `corveil-list-orgs` / `corveil-select-org` / `corveil-deselect-org` (unlike `CorveilConnectionRPCTests` for the connection verbs). The cache/user-id miss was not exercised because provisioner tests always seed `connectedUser.id = "u1"`.
- **Green** — `corveil-select-org` always calls `resolveCorveilOrgName` before `provision` (`RPCHandlers.swift:2517-2521`), so a CLI re-select without `--name` still hits `GET /api/me/organizations` even when reuse would not remint. A 401/outage then fails a no-op reuse. Skip the lookup when a stored key already exists, or pass `--name` from the UI.
- **Green** — disconnect still does not cascade-revoke per-org keys. The PR documents this as accepted (`deselect-org` or a dashboard cascade); leaving it as stated.
- **Green** — if `POST /api/keys` succeeds and `upsertOrgKey` then throws, a later non-`--rotate` select reuses the stale local record. Rare (config write), but rotate-retry should remint rather than reuse after a failed persist.

### Summary Table

**Recommendation:** Request Changes — driven by 0 Red, 1 Yellow, 4 Green findings.

---

dgershman (changes requested):
## Code & Security Review

### Critical Issues

**Red — `upsertOrgKey` resurrects a disconnected connection and leaves a spendable gateway key on disk.** Verified by reproduction (throwaway tests, then deleted).

`CorveilOrgProvisioner.provision` mints over the network (`CorveilOrgProvisioner.swift:115-124`) and only then calls `CorveilConnectionPersistence.upsertOrgKey`. That write does:

```swift
var connection = config.corveilConnection ?? CorveilConnection()
```

(`CorveilConnectionStore.swift:192-200`). If `corveil-disconnect` cleared the block while the mint was in flight, this materializes a **new** empty `CorveilConnection` and stores the `sk-citadel-…` value in `orgKeySecrets`.

This is reachable in production, not only in a unit probe: `corveil-select-org` is laned on `org_id` while `corveil-disconnect` is `.fixed(.config)` (`RPCLanePolicy.swift:139-148`), so the two are allowed to overlap by design. `withConfigLock` serializes the *writes*, it does not make the mint-then-write atomic against disconnect.

Reproduced:

1. Direct `upsertOrgKey` after `corveilConnection = nil` → connection came back as `isEmpty == true` with `orgKeySecrets["org1"]` set.
2. Delayed the mint HTTP response, nilled the connection mid-flight, then let `provision` finish → same zombie connection holding `sk-citadel-LEAKED`.

`CorveilConnection.isEmpty` (`AppConfig.swift:570-573`) only looks at client id + access token, so `corveil status` reports disconnected while `config.json` still holds a spendable key. The next `corveil-connect` merge copies `orgKeySecrets` through untouched (`CorveilConnectionRPCSupport.swift:104-105`), so whoever connects next inherits it.

This is the opposite of the disconnect-wins pattern already in the same file: `updateTokens` / `recordRefreshSuccess` no-op when no connection is stored (`CorveilConnectionStore.swift:120-128`). `upsertOrgKey` should do the same (`guard` a live connection, do not `?? CorveilConnection()`). If the mint already succeeded and the connection is gone, drop (or revoke) the key rather than writing it onto a zombie block. Please add a test for the mid-flight disconnect window — the current provisioner suite never overlaps a mint with a clear.

- **Existing pathway:** `CorveilConnectionPersistence` (local-only door from CROW-1120) + `SettingsSecrets` stripping + `RPCWebSocketHandler.localOnlyDenial`. The new API client / provisioner / CLI verbs extend that door correctly in shape. The miss is that `upsertOrgKey` did not reuse the disconnect-wins / reconnect-wins guards `updateTokens` and `recordRefreshSuccess` already implement in the same type.
- The HTTP client mirrors `CorveilOAuthClient` (injected transport, http(s)+host guard, no I/O beyond HTTP). Reuse-not-remint is the right coordination given the backend's rotate-on-POST contract. Lanes for select/deselect are the right *intent* (don't park Settings saves behind a mint) — they just make the missing guard load-bearing.

### Security Review
**Strengths:**
- `corveil-list-orgs` / `corveil-select-org` / `corveil-deselect-org` are local-only on `/rpc` and listed in `ParityLedger.localOnlyRPCMethods`.
- Select/orgs/list-orgs responses never include the key value; `SettingsSecrets` blanks `orgKeySecrets` values for transport and the preserve path still restores the whole connection from disk (browser cannot mint or clear it via `set-config`).
- Reuse-not-remint avoids silently rotating a live gateway key.
- Base URL is restricted to `http`/`https` with a host; key ids are path-escaped via `appendingPathComponent`.
- Token refresh before expiry is best-effort (`try?`) and a dead token surfaces as `.unauthorized`, matching the CROW-1125 reconnect signal.

**Concerns:**
- The Red finding above — a spendable `sk-citadel-…` key surviving disconnect and riding into the next connect.
- **Green:** `resolveCorveilOrgName` (`RPCHandlers.swift:2550-2561`) skips the membership lookup whenever `--name` is supplied, so `unknownOrg` is never thrown on that path. The backend is still the authorization gate on `POST /api/keys`; treat `--name` as display-only and always resolve/validate the id against memberships.

### Code Quality
- Provisioner / API-client unit coverage is strong on the happy path (reuse, rotate, revoke-404, cache TTL, expiry refresh, secret stripping). It does not cover disconnect-during-mint or the RPC handler (`resolveCorveilOrgName`, payload shape).
- **Green:** `CorveilOrgListCache.invalidate()` is documented as the disconnect hook and is never called. Cache keying on `(baseURL, user id)` plus `notConnected` before a read mostly covers it; wiring `invalidate` from disconnect would still match the comment.
- New Corveil daemon/CLI/Core tests passed locally. `swift build` succeeded. No new failures in the suites this PR adds.

### Summary Table

**Recommendation:** Request Changes — driven by 1 Red, 0 Yellow, 2 Green findings.

---

dhilgaertner:
Thanks for the thorough review @dgershman. Addressed in 09c8ab0b:

**Yellow — org-list cache identity leak (fixed).** You're right on both counts: `CorveilConnectionPersistence.store` never populates `connectedUser`, so keying on `connectedUser.id` collapsed every account on a host to `"<baseURL>\u{1}"`, and `invalidate()` was dead. Fixed by:
- **Keying the cache on the OAuth `clientID`** — DCR issues a fresh client per Connect ("Fresh per connect" in `CorveilIntegrationRoutes`), so any disconnect+reconnect is a guaranteed miss. This is the cross-path guarantee: it also covers the HTTP `POST /config/corveil-connection` clear path in `SecretRoutes`, which the cache can't reach. Regression test: `orgCacheIsKeyedOnClientIdSoAReconnectMisses` seeds an **empty** `connectedUser.id` with two different client ids and asserts the second is a miss.
- **Hoisting the shared cache into `makeCommandRouter`** so `corveil-disconnect` calls `invalidate()` (now live, not dead). Router-level test: `disconnectInvalidatesTheOrgListCache`.

**Green — reuse hit the network (fixed).** `corveil-select-org` no longer resolves the org name before provisioning. Name resolution moved into the provisioner and runs **only on the mint path**; a reuse now returns with zero network calls (`provisionReusesExistingKeyWithoutMinting` asserts `api.lists == 0 && api.posts == 0`), so a no-op reuse can't fail on a transient outage.

**Green — validate `org_id` even when `--name` is supplied.** Left to the backend by design, and I think this is the right call: a supplied `--name` deliberately skips `GET /api/me/organizations` (the UI optimization you suggested), and the `keys.provision` mint enforces membership server-side (403 on a non-member, which surfaces as a clear error). Forcing a local list on every named select would undo the very optimization. A select *without* `--name` still lists and validates locally (`mintForANonMembershipThrowsWithoutMinting`).

**Green — router-level tests for the 3 verbs.** The provisioning handlers call `.live` clients, so they can't be driven end-to-end through the router with stubs (unlike the config-only connection verbs). Coverage instead lives at the provisioner layer where the logic is, now including the empty-user-id cache-isolation case you flagged as unexercised, plus the router-level disconnect→invalidate test above. 85 tests across 7 suites green.

**Green — failed persist → stale reuse.** Accepted as a rare edge (config-write failure after a successful mint). The recovery is exactly what you suggested: `--rotate` skips the reuse short-circuit and reminits, so a stuck stale record self-heals on the next rotate.

**Green — disconnect doesn't cascade-revoke.** Left as documented (`deselect-org` or the dashboard cascade).

dgershman (changes requested):
## Code & Security Review

### Critical Issues
- **Yellow — `upsertOrgKey` resurrects a connection (and the freshly minted `sk-citadel-…` secret) after disconnect.** `CorveilConnectionPersistence.upsertOrgKey` does `config.corveilConnection ?? CorveilConnection()` and always writes (`CorveilConnectionStore.swift:192-201`). The sibling `removeOrg` (`CorveilConnectionStore.swift:209-211`) already refuses to create a connection when none is stored — upsert should do the same. `provision` mints over the network *then* upserts (`CorveilOrgProvisioner.swift:146-156`), and `corveil-select-org` is lane `.on("org_id")` while `corveil-disconnect` is `.fixed(.config)` (`RPCLanePolicy.swift:140-148`), so the two run concurrently. A disconnect that lands in that window is overwritten by a zombie `CorveilConnection` with empty OAuth (`isEmpty == true`, status looks disconnected) but a live secret in `orgKeySecrets`. The next `corveil-connect` merge copies `orgKeys`/`orgKeySecrets` from whatever is stored (`CorveilConnectionRPCSupport.swift:104-105`), so a later Connect — including as a different account — inherits that key. Reproduced with a temporary test that disconnected, called `upsertOrgKey`, and observed the secret written back; scratch deleted. Fix: require an existing connection (and refuse an undecodable `config.json`, same CROW-814 guard as `mutateConfig`); if the connection vanished mid-mint, skip the write (and ideally revoke the key just minted).

- **Existing pathway:** this correctly extends the CROW-1120 door — `CorveilOAuthClient` (injected-transport HTTP) + `CorveilConnectionPersistence` (`ConfigStore.withConfigLock`) + `CorveilConnectionRPC` / `SettingsSecrets` strip-and-restore + `RPCWebSocketHandler.localOnlyDenial` + `ParityLedger` — rather than inventing a parallel credential store. `CorveilAPIClient` mirrors the OAuth client; reuse-don't-remint is the right shape given the backend's one-active-key contract.
- **Yellow** — `upsertOrgKey` is the one persistence helper that *creates* a connection; `updateTokens` / `recordRefreshSuccess` / `removeOrg` all no-op when none is stored. See Critical Issues.

### Security Review
**Strengths:**
- New RPCs are local-only (`RPCWebSocketHandler.swift:372-380`) and listed in `ParityLedger.localOnlyRPCMethods`; responses never echo the key value, only metadata (`key_prefix`, `key_id`).
- `SettingsSecrets` blanks `orgKeySecrets` values for transport and restores the whole `corveilConnection` block on `set-config`, so a browser cannot read or plant a gateway key.
- Reuse of a stored key skips the network entirely, which is load-bearing: a re-`POST /api/keys` would rotate the backend key and orphan bound gateways.
- HTTP scheme/host guard on `Endpoints` matches `CorveilOAuthClient`; 401/403 is split out as `.unauthorized`.

**Concerns:**
- **Yellow** — secret surviving disconnect, then riding into the next Connect via merge (Critical Issues). Confirmed against `CorveilConnectionStore.swift:195` — `?? CorveilConnection()` is applied; no "connection still present" filter.
- **Green** — `Endpoints.key(id:)` (`CorveilAPIClient.swift:96-100`) claims the id is path-escaped, but `URL.appendingPathComponent("../me")` yields `…/api/keys/../me` (`URL.standardized` → `/api/me`). Spaces are encoded; `/` and `..` are not. Key ids come from Corveil, so this is defense-in-depth against a hostile `baseURL` the user already sent a bearer to — encode the path component (or reject `id` containing `/` or `..`) so the comment holds.

### Code Quality
- Provisioner tests cover reuse, rotate, revoke-404, cache keying on client id, and expired-token refresh; CLI parsing and SettingsSecrets round-trips are in place. `swift build` succeeded; CrowDaemon `--filter Corveil` (107), CrowCore `SettingsSecrets` (18), and CrowCLI Corveil parsing (12) all passed.
- **Green** — `resolvedOrgName` skips the membership check when `--name` / `org_name` is supplied (`CorveilOrgProvisioner.swift:171-173`). Backend `POST /api/keys` is the real gate; Crow will mint for any id the caller names. Fine if that is intentional for the UI, but the CLI can skip the lookup the no-`--name` path uses as validation.
- **Green** — no router-level tests for `corveil-list-orgs` / `select-org` / `deselect-org` (connect/status/disconnect have `CorveilConnectionRPCTests`). Handlers are thin, but a JSON-shape or `corveilMapErrors` regression would not be caught. `DaemonSecurityTests.corveilConnectionMethodsAreLocalOnly` also still lists only the four CROW-1120 methods; the ledger-mirror test still covers the new three.

### Summary Table

---

dhilgaertner:
Thanks for catching the mint/disconnect race @dgershman — that's a real credential leak. Fixed in 97f26c2f.

**Red/Yellow — `upsertOrgKey` resurrecting a disconnected connection (fixed).** You're exactly right: `select-org` (`.on("org_id")`) and `disconnect` (`.fixed(.config)`) sit in different lanes, so a disconnect can land during the mint's HTTP call, and `?? CorveilConnection()` then wrote a zombie block holding the `sk-citadel-…` that the next connect-merge inherited. `upsertOrgKey` now follows the disconnect-wins pattern its siblings (`updateTokens` / `recordRefreshSuccess` / `removeOrg`) already use — it guards a live connection and returns whether it wrote. When the write is refused, `provision` **revokes the orphaned key** best-effort and throws rather than reporting success. Regression test `disconnectDuringMintDoesNotResurrectTheConnectionOrLeakTheKey` nils the connection from inside the mint transport and asserts (a) it throws, (b) `corveilConnection` stays `nil` — no zombie, and (c) the orphaned key was revoked (one DELETE).

**Green — `--name` bypassing membership validation (fixed).** Agreed, and simplest to just close it: name resolution now **always** validates `org_id` against the memberships list on the mint path (cheap, cached); `--name` is a display-only override. `mintForANonMembershipThrowsWithoutMinting` now asserts a non-membership id is rejected *both* with and without `--name`. Reuse still returns before any network call.

**Green — `Endpoints.key(id:)` path-escaping overclaim (fixed).** You're right that `appendingPathComponent` neither encodes `/` nor collapses `..`. `revokeKey` now rejects an id that isn't a safe single segment (`isSafeKeyID`) before issuing the request, and the comment is corrected. Tests cover `../me`, `a/b`, `..`, backslash, and empty.

**Green — `DaemonSecurityTests` explicit list (fixed).** `corveilConnectionMethodsAreLocalOnly` now also asserts `corveil-list-orgs` / `corveil-select-org` / `corveil-deselect-org` are gated local-only (the bidirectional ledger-mirror already covered them; this makes it explicit).

**Green — router-level tests for the 3 verbs.** Still constrained here: the provisioning handlers call `.live` clients, so they can't be exercised end-to-end through the router with stubs the way the config-only connection verbs can. Coverage lives at the provisioner layer (now including the mint/disconnect race and membership validation), plus the router-level disconnect→invalidate test. Happy to add a client-injection seam for the handlers if you'd prefer full router coverage, but it'd mean threading test clients through `makeCommandRouter`.

Full CrowDaemon suite green except the pre-existing `WebNotificationCenterTests.bootCatchResetsTheHistory` flake (a web-asset test that also fails on `main` in a multi-suite run — unrelated to this PR).

dgershman (changes requested):
## Code & Security Review

Prior round's requested fixes are in and hold: org-list cache is keyed on the DCR `clientID` (`CorveilOrgProvisioner.swift:260-262`) with `invalidate()` on `corveil-disconnect`; `upsertOrgKey` no longer materializes a connection when none is stored; `--name` is display-only and membership is always validated on the mint path; `revokeKey` rejects unsafe key ids; the three new RPCs are explicitly asserted local-only. Those are not re-raised.

### Critical Issues
- **Yellow — `upsertOrgKey` still plants a freshly minted `sk-citadel-…` onto whatever connection is stored when the write lands, including a different account that connected during the mint.** The nil-connection guard (`CorveilConnectionStore.swift:207`) stops the zombie case from the last round. It does not check that the stored grant is still the one that minted. `provision` mints over the network then upserts (`CorveilOrgProvisioner.swift:148-162`). `corveil-select-org` is lane `.on("org_id")` while `corveil-connect` / `corveil-disconnect` are `.fixed(.config)` (`RPCLanePolicy.swift:138-148`), so they overlap by design; browser Connect is HTTP and is not lane-gated at all.

  Reproduced with a temporary test (deleted after): delayed the mint HTTP response, replaced the stored connection mid-flight with a different DCR `clientID` (`client-2`), then let `provision` finish. Outcome: select returned success (`reused: false`), `orgKeySecrets["org1"]` was `sk-citadel-LEAKED` on the new connection, and zero `DELETE`s ran — the orphaned key was not revoked. Confirmed against `CorveilConnectionStore.swift:207` — the only filter is `corveilConnection != nil`; there is no client-id / grant check.

  The existing pattern in the same type is reconnect-wins: `recordRefreshSuccess` refuses the write unless `expectedRefreshToken` still matches (`CorveilConnectionStore.swift:147-149`). `upsertOrgKey` should do the equivalent with the minting connection's **client id** (same identity the org-list cache already uses — survives a concurrent token refresh, misses a DCR reconnect). On mismatch: return false, and `provision` already revokes best-effort. Please add a regression next to `disconnectDuringMintDoesNotResurrectTheConnectionOrLeakTheKey`.

- **Existing pathway:** CROW-1120's local-only door — `CorveilOAuthClient` (injected-transport HTTP) + `CorveilConnectionPersistence` (`ConfigStore.withConfigLock`) + `CorveilConnectionRPC` / `SettingsSecrets` strip-and-restore + `RPCWebSocketHandler.localOnlyDenial` + `ParityLedger`. This PR extends that door correctly: sibling `CorveilAPIClient`, provisioner coordination, `orgKeySecrets` as the secret half of `orgKeys`, three local-only verbs. It does not invent a parallel credential store or go through `set-config`. Reuse-don't-remint is the right shape given the backend's rotate-on-POST contract. Lanes for select/deselect are the right intent (don't park Settings behind a mint).
- **Yellow** — `upsertOrgKey` took only the disconnect-wins half (`nil` → refuse) from `updateTokens` / `removeOrg`. The overlapping-HTTP-then-write helper in the same file is `recordRefreshSuccess`, which also checks grant identity. See Critical Issues.

### Security Review
**Strengths:**
- `corveil-list-orgs` / `corveil-select-org` / `corveil-deselect-org` are local-only (`RPCWebSocketHandler.swift:372-380`) and listed in `ParityLedger.localOnlyRPCMethods`; they are not MCP-exported. Responses never echo the key value.
- `SettingsSecrets` blanks `orgKeySecrets` values for transport and restores the whole `corveilConnection` block on `set-config`, so a browser cannot read or plant a gateway key.
- Reuse of a stored key skips the network entirely (load-bearing: a re-`POST /api/keys` would rotate the backend key). Membership is always validated on the mint path, including when `--name` is supplied.
- `Endpoints` rejects non-http(s) base URLs; `revokeKey` rejects ids that `appendingPathComponent` would mishandle (`/`, `..`, `\`).
- Revoke treats 404 as success and does not drop the local record until the server call returns, so a 5xx revoke fails closed. Disconnect-during-mint now refuses the write and revokes the orphan.

**Concerns:**
- **Yellow** — reconnect-during-mint attaching the minted secret to a different connection (Critical Issues).

### Code Quality
- Provisioner / API-client coverage is strong on reuse, rotate, revoke-404, cache keying on client id, expired-token refresh, and the nil-connection mint race. CLI parsing, SettingsSecrets round-trips, and the parity/lane/local-only gates are in place.
- `swift build` succeeded. CrowDaemon `--filter Corveil`: 110 tests in 8 suites passed. No new failures in the suites this PR adds.
- **Green** — `CorveilOrgListCache`'s class comment still says it is "keyed on (base URL, user)" (`CorveilOrgProvisioner.swift:266-267`) but `cacheKey` is `(baseURL, clientID)`. Docs-only; the implementation is the one the last round asked for.

### Summary Table

**Recommendation:** Request Changes — driven by 0 Red, 1 Yellow, 1 Green findings.

---

dhilgaertner:
Good catch on the reconnect variant @dgershman — the nil-guard only covered half of it. Fixed in 151997b0.

**Yellow — reconnect-during-mint attaching the key to a different connection (fixed).** You're right: a non-nil check isn't enough, because a *different account* connecting mid-mint (fresh DCR `clientID`) leaves a non-nil-but-wrong connection. `upsertOrgKey` now takes **reconnect-wins** semantics exactly like `recordRefreshSuccess`: it accepts the minting connection's `expectedClientID` and writes only if the stored `clientID` still matches. A concurrent token refresh keeps the client id (the write lands); a disconnect (nil) or a reconnect / different-account Connect (fresh DCR client id) refuses. On refusal `provision` revokes the orphan best-effort and throws `ProvisionError.connectionChanged`.

Regression `reconnectAsADifferentAccountDuringMintDoesNotInheritTheKey` reproduces exactly your scenario — replaces the stored connection with `clientID: "client-2"` from inside the mint transport — and asserts the reconnected account keeps **no** `orgKeys`/`orgKeySecrets["org1"]` and the orphan is revoked (one DELETE). The prior disconnect test now asserts `connectionChanged` too, and both share one `provisionRacingAMint` helper.

**Green (docs) — fixed.** The `CorveilOrgListCache` class comment now says "(base URL, DCR client id)" to match `cacheKey`.

CrowDaemon `--filter Corveil` green (89 tests); full suite green except the pre-existing `WebNotificationCenterTests.bootCatchResetsTheHistory` flake (unrelated web-asset test that also fails on `main`).

dgershman (approved):
## Code & Security Review

Prior round's requested fix is in and holds: `upsertOrgKey` now takes reconnect-wins semantics on the minting connection's DCR `clientID` (`CorveilConnectionStore.swift:207-214`), so a nil connection *or* a different live grant refuses the write. `provision` revokes the orphan best-effort and throws `.connectionChanged` (`CorveilOrgProvisioner.swift:171-177`). Regression `reconnectAsADifferentAccountDuringMintDoesNotInheritTheKey` covers the exact prior scenario. The `CorveilOrgListCache` class comment now matches `cacheKey`. Those are not re-raised.

### Critical Issues
None.

- **Existing pathway:** CROW-1120's local-only door — `CorveilOAuthClient` (injected-transport HTTP) + `CorveilConnectionPersistence` (`ConfigStore.withConfigLock`) + `CorveilConnectionRPC` / `SettingsSecrets` strip-and-restore + `RPCWebSocketHandler.localOnlyDenial` + `ParityLedger`. This PR extends that door correctly: sibling `CorveilAPIClient`, provisioner coordination, `orgKeySecrets` as the secret half of `orgKeys`, three local-only verbs. It does not invent a parallel credential store or go through `set-config`. Reuse-don't-remint is the right shape given the backend's rotate-on-POST contract. Lanes for select/deselect (`.on("org_id")`, not `.config`) are the right intent — don't park Settings behind a mint — and the persist helpers now share one reconnect-wins identity check (`clientID` here, refresh token on `recordRefreshSuccess`).

### Security Review
**Strengths:**
- `corveil-list-orgs` / `corveil-select-org` / `corveil-deselect-org` are local-only (`RPCWebSocketHandler.swift:372-380`) and listed in `ParityLedger.localOnlyRPCMethods`; they are not MCP-exported. Responses never echo the key value.
- `SettingsSecrets` blanks `orgKeySecrets` values for transport and restores the whole `corveilConnection` block on `set-config`, so a browser cannot read or plant a gateway key.
- Reuse of a stored key skips the network entirely (load-bearing: a re-`POST /api/keys` would rotate the backend key). Membership is always validated on the mint path, including when `--name` is supplied.
- Org-list cache is keyed on `(baseURL, clientID)` so a DCR reconnect misses; `corveil-disconnect` also `invalidate()`s the shared cache.
- `Endpoints` rejects non-http(s) base URLs; `revokeKey` rejects ids that `appendingPathComponent` would mishandle (`/`, `..`, `\`).
- Disconnect-during-mint and reconnect-during-mint both refuse the write and revoke the orphan. Confirmed against `CorveilConnectionStore.swift:212-214` — the filter is `corveilConnection != nil` **and** `clientID == expectedClientID`.

**Concerns:**
- None remaining at blocking severity.

### Code Quality
- Provisioner / API-client coverage is strong on reuse, rotate, revoke-404, cache keying on client id, expired-token refresh, the nil-connection mint race, and the different-account reconnect mint race. CLI parsing, SettingsSecrets round-trips, and the parity/lane/local-only gates are in place.
- `swift build` succeeded. CrowDaemon `--filter Corveil`: 111 tests in 8 suites passed. CrowCore `SettingsSecrets`: 18 passed. CrowCLI Corveil parsing: 12 passed.

### Summary Table

**Recommendation:** Approve — driven by 0 Red, 0 Yellow, 0 Green findings.

---

PR #1154, opened 2026-08-28
Add a session grid to watch up to 16 sessions

### LINKED ISSUE #1153: Single-pane-of-glass session grid: pin and watch up to 16 active sessions in one screen
## Summary

Add a new "single pane of glass" screen to Crow: a grid that shows several **pinned / active** sessions at once — up to **16** in one view — so you can watch a whole fleet of running sessions without tabbing between them.

## The view
- A new top-level screen (peer of the existing board views), reachable from the sidebar/nav.
- **Grid of live session cells**, responsive up to 16 (2×2 → 3×3 → 4×4); overflow paginates rather than shrinking cells into uselessness.
- Each cell renders that session's **primary terminal** live, with a header: session name, status, and the activity badge (working / waiting / done / idle) already surfaced by the state machine.
- **Pin to include:** pin a session to add it to the grid; the wall shows pinned sessions (and/or all active). Pin state persists per user.
- Click/focus a cell to expand to the full session; unpin / reorder from the cell header.

## Cross-frontend note
Crow's terminal UI is hand-mirrored across desktop `terminal.html`, web `terminal.html`, and web `app.js` (they drift — see the shared-xterm-addon pattern). A 16-pane grid multiplies that surface, so build the grid as a layout **around** the existing terminal component rather than a fourth bespoke renderer, and decide desktop-first vs both up front.

## Open questions (decide during impl)
- **Interactive vs read-only** cells: are the 16 panes live/typeable, or read-only mirrors with click-to-focus for input? (Perf + input-routing tradeoff.)
- Performance budget for up to 16 concurrent xterm.js instances — throttle/suspend rendering for off-screen or idle panes?
- Which terminal per session when a session has several (primary/managed only?).
- Pin model: a per-session flag vs a per-user pinned-set in `AppConfig`.
- Selection source: only pinned, or "active sessions" auto-filled up to 16 with pinning to override ordering.
- Desktop app, web, or both for v1.

dgershman:
Closes #1153

## Summary
- New top-level **Grid** board (`#/grid`) shows up to 16 sessions at once, with 2×2 / 3×3 / 4×4 layout and pagination past 16.
- Each cell is a **read-only capture-pane snapshot** of the session's managed terminal (not a live PTY), painted into the existing xterm.js component. Click a cell to open the full session. [ADR 0022](docs/adr/0022-session-grid-snapshots.md) records why: 16 attaches would SIGWINCH the shared tmux cockpit under `window-size latest`.
- Pin from the session menu or the cell header. Pins persist per browser and lead the roster; active / in-review sessions auto-fill the rest.

## Decisions (ticket open questions)
- **Read-only cells**, click-to-expand for input.
- **Web only** (Tauri loads the same UI; ADR 0010).
- **Managed terminal** per session (Managers fall through to the switcher's window picker).
- **Per-browser pin list** in `localStorage`, not AppConfig.
- **Pinned first + auto-fill** of active sessions, overflow paginates.

## Test plan
- [ ] Open `#/grid` from the sidebar Grid pill; empty workspace shows the pin CTA.
- [ ] With several active sessions, cells appear without pinning; activity badges and names match the sidebar.
- [ ] Pin / unpin from the session ⋮ menu and from the cell header; order is preserved across reload.
- [ ] Click a cell (or its name) and land on that session's full terminal.
- [ ] Unpin / complete sessions and confirm the wall updates; page 2 appears only with more than 16.
- [ ] Focused session TUI size is unchanged while the grid is open (the reason snapshots exist).


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- ADR 0022 documents why read-only `capture-pane` snapshots are used instead of live PTY attaches — the right mitigation for tmux `window-size latest` and the 16-connection limiter.
- `list-session-terminal-snapshots` is best-effort, caps input at 16 UUIDs, and bounds reported dimensions (`cols` ≤ 400, `rows` ≤ 200) in `TerminalCockpit.snapshotPane`.
- Grid cells use `disableStdin: true`, CSS `pointer-events: none` on xterm, and click-to-expand — no accidental input into watched panes.
- Terminal content exposure follows the same trust model as `get-session-terminal-preview` (authenticated daemon client); ParityLedger is updated.
- Session names in cell headers go through `textContent` via `el()`, avoiding HTML injection.

**Concerns:**
- None blocking. Batch snapshot RPC can surface terminal ANSI for any session id the caller already knows — identical sensitivity to the existing preview endpoint and expected for this feature.

### Code Quality
- `terminalWatchWindowIndex` cleanly prefers the managed agent pane with a tested fallback for Managers.
- Client lifecycle is handled well: `disposeUnusedGridTerms`, ResizeObserver teardown, `gridKey` short-circuit to preserve xterm instances across header-only updates, and poll skip when `document.hidden`.
- `replayFrame` reuse keeps grid snapshots consistent with scrollback replay elsewhere.
- Solid test coverage: 5 new Swift handler tests + 2 watch-index unit tests; `grid.test.js` covers roster/pin/layout/routing/DOM (31 assertions). `swift build` and targeted Swift/web tests pass locally.

**Green (consider, not blocking):**
- Orphaned pin UUIDs in `localStorage` are skipped by `gridRoster` but not pruned when sessions disappear — harmless, could tidy in a follow-up.
- The handler runs up to 16 sequential tmux captures per poll; acceptable at ~800ms cadence but could batch MainActor lookups or parallelize later if latency shows up.

### Summary Table

---

PR #1165, opened 2026-08-30
Skip same-size SIGWINCH on session switch and tab refocus

### LINKED ISSUE #1162: Session/window switch lag: resize→SIGWINCH→TUI reflow, amplified by the grid
## Description

Switching between sessions (and between terminal windows) has a visible lag: the agent TUI briefly reflows/repaints. It is noticeably worse with the new session grid (CROW-1153) open. The root cause is that a switch — and even a plain tab refocus — forces the shared tmux window to resize, which SIGWINCHes the agent and makes its whole TUI re-lay-out.

## Where it comes from

All surfaces (Tauri desktop, browser tabs, phone) attach to **one** grouped tmux session window sized `window-size latest`, so only the focused/visible surface "owns" the size (#667). Two paths re-assert that size and trigger a full agent reflow:

1. **Session/window switch** — `attachWindow()` → `switchAgentWindow(win)` (`web/terminal.js`) clears the buffer, `select-window {replay:false}`, then defers `fitTerminal()`. `applyTermFit()` runs `fitAddon.fit()` and, when cols/rows differ from the last, sends a `resize` frame → SIGWINCH → the agent repaints. Session kinds render **different header/tab-bar heights**, so the pane geometry genuinely differs between two sessions → the fit is *not* a no-op → real SIGWINCH on every switch.

2. **Tab refocus** — `takeTerminalOwnership()` (wired to `window` `focus` + `visibilitychange`) resets the dedup (`lastTermCols = 0`) and calls `applyTermFit()`, which then sends a `resize` frame **even when the size is unchanged**, purely to reclaim "latest" ownership from a background surface. So merely clicking back into the tab forces a redundant SIGWINCH → reflow.

## Why the grid amplifies it

`web/grid.js` (CROW-1153) keeps up to **16** xterm instances mounted (`mountGridTerm`), each with its own `ResizeObserver` → `scaleGridTerm` (CSS `transform: scale`). `startGridPolling` polls `list-session-terminal-snapshots` for every visible cell **every 800 ms**, and `paintGridSnapshot` does `term.resize()` + `term.write()` + `scaleGridTerm()` per changed cell. Sixteen xterm writes plus CSS-transform recomputes on an 800 ms cadence is heavy main-thread work; interleaving it with a live session's fit/reflow is what makes swaps feel laggy. Clicking a cell → `selectSession` → a full live attach + fit (another SIGWINCH), and the 16 grid xterms are retained in `gridTerms` (detached, not disposed) unless `leaveGridView()` runs.

## Directions to investigate (not prescriptive)

- **Reclaim ownership without a resize.** `takeTerminalOwnership` sends a same-size `resize` frame just to become tmux's latest client. If ownership can be re-asserted without changing dimensions (e.g. a `refresh-client`/`resize-window` that no-ops when the size already matches), the refocus reflow disappears while `window-size latest` behavior is preserved.
- **Normalize pane chrome height across session kinds** so switching between two equally-sized sessions yields an identical grid — then `applyTermFit`'s `lastTermCols/lastTermRows` dedup makes the switch fit a true no-op (no SIGWINCH, no reflow).
- **Grid throttling/teardown.** Confirm `document.hidden` pausing is enough; consider an adaptive/longer poll interval, debouncing per-cell `scaleGridTerm`/write, and disposing grid terms via `leaveGridView()` when `selectSession` opens a cell (verify the wall is torn down, not just detached).
- Check whether `switchAgentWindow`'s deferred `fitTerminal()` double-fires with the container `ResizeObserver` on the same post-switch layout settle.

## Files

- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/terminal.js` — `applyTermFit`, `fitTerminal`, `takeTerminalOwnership`, `attachWindow`, `switchAgentWindow`, `connectTerminalWs` (onopen fit).
- `Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/grid.js` — `startGridPolling`, `refreshGridSnapshots`, `paintGridSnapshot`, `scaleGridTerm`, `observeGridTerm`.

## Related

CROW-1091 (#1093) fit-on-in-place-switch, #661/#637/#667 resize/SIGWINCH dedup history, ADR 0022 (grid uses read-only capture-pane, not live attaches).

dgershman:
Closes #1162

## Summary

Switching sessions or refocusing a tab was SIGWINCHing the agent TUI because a same-size `resize` frame always became `TIOCSWINSZ`, and the session grid made that worse by disposing 16 xterms on the live-attach turn.

- Tab refocus still reclaims `window-size latest`, but sends `if_needed` so the daemon skips the ioctl when the window is already that size.
- The empty Manager tab bar keeps its 34px slot so Manager ↔ work no longer changes pane geometry by collapsing the strip.
- Grid polling is 1500ms, CSS scale is skipped when unchanged, and `leaveGridView` defers xterm dispose off `selectSession`.

## Test plan

- [x] `node Packages/CrowDaemon/web-tests/session-switch-attach.test.js` (incl. CROW-1162 refocus cases)
- [x] `node Packages/CrowDaemon/web-tests/grid.test.js` (deferred dispose + scale skip)
- [x] CrowDaemon Swift tests for `parseColsRows`, `takeTerminalOwnership`, and `#tabbar:empty`
- [ ] Click between two work sessions of similar chrome — agent TUI should not reflow
- [ ] Click away from the Crow tab and back — same, no reflow
- [ ] Phone/desktop both attached at different sizes — focusing the desktop still takes the window size
- [ ] Open the session grid, click a cell — attach should feel snappy; wall tears down after idle

### DIFF
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/app.css
@@ -414,7 +414,9 @@ body {
   padding: 0 8px;
   min-height: 34px;
 }
-#tabbar:empty { display: none; }
+/* CROW-1162: keep the 34px slot even with no tabs (Manager). Collapsing it
+   on Manager ↔ work switches is a real row of TUI geometry → SIGWINCH. */
+#tabbar:empty { display: flex; }
 .tab {
   display: flex;
   align-items: center;
--- a/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/grid.js
+++ b/Packages/CrowDaemon/Sources/CrowDaemon/Resources/web/grid.js
@@ -109,9 +109,39 @@ function gridVisibleSessions() {
 
 function leaveGridView() {
   stopGridPolling();
-  disposeUnusedGridTerms([]);
   const root = document.getElementById('board');
   if (root) delete root.dataset.gridKey;
+  // CROW-1162: disposing up to 16 xterm instances is heavy. Do it after the
+  // live session attach paints, not on the same turn as `selectSession`.
+  scheduleGridTeardown();
+}
+
+function cancelGridTeardown() {
+  gridTeardownGen += 1;
+  if (gridTeardownTimer != null) {
+    clearTimeout(gridTeardownTimer);
+    gridTeardownTimer = null;
+  }
+}
+
+function scheduleGridTeardown() {
+  const gen = ++gridTeardownGen;
+  if (gridTeardownTimer != null) {
+    clearTimeout(gridTeardownTimer);
+    gridTeardownTimer = null;
+  }
+  const ids = [...gridTerms.keys()];
+  const tearDown = () => {
+    gridTeardownTimer = null;
+    if (gen !== gridTeardownGen) return;
+    if (selectedBoard === 'grid') return;
+    for (const id of ids) disposeGridTerm(id);
+  };
+  if (typeof requestIdleCallback === 'function') {
+    requestIdleCallback(tearDown, { timeout: 400 });
+  } else {
+    gridTeardownTimer = setTimeout(tearDown, 0);
+  }
 }
 
 function stopGridPolling() {
@@ -124,7 +154,7 @@ function startGridPolling() {
     if (selectedBoard !== 'grid') { stopGridPolling(); return; }
     if (typeof document !== 'undefined' && document.hidden) return;
     refreshGridSnapshots();
-  }, 800);
+  }, GRID_POLL_MS);
 }
 
 function disposeGridTerm(id) {
@@ -143,6 +173,7 @@ function disposeUnusedGridTerms(keepIds) {
 }
 
 function renderSessionGrid(root) {
+  cancelGridTeardown();
   root.classList.add('session-grid-board');
   const { visible, pages, roster } = gridVisibleSessions();
   const visIds = visible.map((s) => s.id);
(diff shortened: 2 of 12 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- The `if_needed` resize path is gated server-side in `TerminalWebSocket.swift`: skip only when tmux reports the window is already the requested size; a failed `windowSize` query falls through to `pty.resize`, so uncertainty fails open to a real resize rather than silently ignoring a mismatch.
- `parseColsRows` rejects zero, negative, and non-numeric input before the skip-ioctl gate runs.
- Existing `/terminal` auth (origin guard + WebAuth) is unchanged; the new flag only affects ioctl behavior on an already-authorized connection.

**Concerns:**
- None blocking. Client-supplied `if_needed` cannot bypass a size mismatch because the daemon compares against live tmux dimensions, not the client claim alone.

### Code Quality
- **Green:** `takeTerminalOwnership` / `sendTermResize(ifNeeded)` in `terminal.js` cleanly separates tab-refocus reclaim (conditional ioctl) from ordinary `applyTermFit` dedup (silent when unchanged). The server-side tmux check correctly handles the cross-surface case where local fit dimensions match cached dedup but the shared window was reshaped elsewhere.
- **Green:** Grid perf work in `grid.js` — deferred xterm dispose via `requestIdleCallback`, poll interval bump to 1500 ms, rAF-coalesced `ResizeObserver`, and `lastScale` skip — targets main-thread contention without changing grid semantics.
- **Green:** `#tabbar:empty { display: flex }` in `app.css` is a deliberate geometry-stability trade (34 px reserved on Manager); comments in `session.js` document the rationale.
- **Green:** Test coverage is solid at three layers: Swift unit tests for `parseColsRows` and source-structure assertions in `WebTerminalAssetTests.swift`; jsdom tests for refocus/resize behavior and grid teardown/scale skip. All 613 CrowDaemon Swift tests pass; CROW-1162 web tests pass after `npm ci`.

### Summary Table

---

PR #1188, opened 2026-09-02
Fix Claude Code caret offset in Crow terminal

### LINKED ISSUE #1157: Claude Code cursor position offset in Crow terminal
## Description

The text cursor (caret) in Claude Code's input prompt is visually offset from where characters are actually inserted when running inside a Crow terminal (tmux).

## Screenshot

The cursor block appears several characters to the right of the actual insertion point — after `ctx: 0%` there's a visible gap before the cursor, but typed text appears at the prompt position, not at the cursor.

cursor-offset

## Environment

- Claude Code running inside Crow terminal (tmux-backed)
- macOS
- `tmux detected · scroll with` banner visible in top-right

## Steps to Reproduce

1. Launch a Crow workspace session (Claude Code agent)
2. Observe the input prompt area
3. The white cursor block is offset to the right of where text is actually entered

## Expected Behavior

Cursor position should match the actual text insertion point in the prompt.

## Likely Cause

Terminal width / column reporting mismatch between Claude Code's TUI and the tmux pane dimensions, possibly related to how Crow sizes or resizes the tmux pane, or Unicode/emoji width calculation in the status line (`ctx: 0%`, mode indicators).

dgershman:
Closes #1157

## Summary
- xterm.js was measuring emoji and status-line symbols with Unicode 6 (1 cell) while tmux and Claude Code use Unicode 11 / Node `string-width` (2 cells), so the hardware cursor sat several columns past `ctx: 0%` while typed text inserted at the prompt.
- Both live terminal surfaces now load `@xterm/addon-unicode11` (matching the vendored xterm 6.0.0) and set `term.unicode.activeVersion = '11'` before any PTY bytes. `allowProposedApi` is required to switch the table; a missing `/xterm` fetch is skipped rather than taking the terminal down.

## Test plan
- [ ] Open a Claude Code session in Crow and confirm the blinking caret sits on the prompt insertion point, not after `ctx: 0%` / mode glyphs.
- [ ] Type in the prompt: characters appear under the caret, including after status-line emoji (tmux banner, mode indicators).
- [ ] Resize the pane and switch away/back: caret stays aligned (no regression of CROW-1035 / CROW-1162).
- [ ] Plain shell tab still renders normally (Nerd Font / CJK / emoji prompts).
- [ ] Reload the UI so the new `/xterm/xterm-addon-unicode11.js` asset is fetched.


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

### Security Review
**Strengths:**
- The new asset is covered by `StaticAssets.isSafeAssetName` in `DaemonSecurityTests`, consistent with other `/xterm` addons.
- The Unicode 11 addon is vendored from the official xterm.js 6.0.0 addon bundle (same pattern as fit/image/viewport addons) and served from the existing same-origin static path — no new injection surface or credential handling.
- `allowProposedApi` is scoped to client-side terminal configuration only; it does not widen server-side attack surface.

**Concerns:**
- None.

### Code Quality
- Root-cause analysis is correct: Unicode 6 vs Unicode 11 wcwidth mismatch between xterm.js defaults and tmux/Node `string-width` explains the hardware-cursor offset while typed text stays at the prompt.
- Fix is applied consistently on both live web surfaces (`terminal.js` / `index.html` and `Packages/CrowDaemon/.../terminal.html`), with the guarded load pattern matching other addons (viewport, WebGL) so a failed `/xterm` fetch does not break `ensureTerminal`.
- Addon is loaded and `term.unicode.activeVersion = '11'` is set before `open()` / first PTY write — correct ordering for width tables.
- Drift-guard tests in `WebTerminalAssetTests` pin both the `<script>` tag and the activation code on each surface; `BundledResourcesTests` pins the vendored file and UMD global. Related tests pass locally.
- Retired `CrowTerminal/Resources/xterm/terminal.html` is intentionally left unchanged (ADR 0010); the two updated pages are the actual live surfaces.

### Summary Table

| Finding | Color | Location |
|---------|-------|----------|
| Silent fallback when the addon script 404s leaves the caret bug in place (documented tradeoff: availability over hard failure) | Green | `terminal.js`, `terminal.html` |
| `xterm-addon-unicode11.js` references a source map that is not bundled (harmless at runtime) | Green | `xterm-addon-unicode11.js` |

---

PR #1225, opened 2026-09-10
Infer a missing SessionWorktree from the managed terminal cwd

### LINKED ISSUE #1219: Infer and register a worktree from the managed terminal cwd when the session has none
## Use Case

A work session can already be running with a managed terminal whose cwd **is** a git worktree, while Crow's store has no `SessionWorktree` row (#1218). PR auto-link, merge labels, and `primaryWorktree(for:)` all no-op until someone runs `crow add-worktree` by hand.

Recovery should not require the operator to notice. The terminal cwd is enough to reconstruct repo, path, and branch.

## Proposed Solution

On poll (and/or on terminal create), if a non-manager work session has no worktrees:

1. Take the managed terminal's `cwd`.
2. If it is a git worktree under `devRoot`, read `git rev-parse --abbrev-ref HEAD` and `git remote get-url origin`.
3. `add-worktree` with `--primary`, `repo` from the slug, `repo-path` from the main clone, `path` = cwd, `branch` = HEAD.

Then the existing `applySessionPRLinks` / `reconcileMissingPRLinks` paths can match.

Do not invent a second worktree if one already exists. Do not clobber a session that is deliberately worktree-less (explore/job) — gate on `kind == .work`.

## Alternatives Considered

- Only fix bootstrap (#1218) and leave already-running sessions broken until recreate.
- Parse the worktree directory name (`{repo}-{number}-{slug}`) without asking git. Fragile if the coder renamed the branch.

## Additional Context

Observed: session `F3E0A179-2422-4FBF-96F1-8B07B46ACDC2`, terminal cwd `/Users/danny/Projects/devroot/corveil/shell-crm-473-open-geo-enrichment`, branch `feature/shell-crm-473-open-geo-enrichment`, PR https://github.com/corveil/shell-crm/pull/478. Registering that worktree via `/rpc` immediately made the session matchable.

`add-worktree` RPC does not create the git worktree; it only records the row. The git checkout already exists in this failure mode.

dgershman:
Closes #1219

## Summary
- If a work session has no `SessionWorktree` row, recover one from the managed terminal's cwd by reading `HEAD` and `origin` — metadata only, the checkout already exists.
- Runs on every board poll (even when GitHub is rate-limited) so the same tick's PR auto-link can match, and again on managed `new-terminal` so a bootstrap miss is repaired at create time.
- Gated on `kind == .work` and an empty worktree list, so jobs, reviews, and Managers stay worktree-less on purpose.

## Test plan
- [ ] `swift test --package-path Packages/CrowGit --filter WorktreeInference`
- [ ] `swift test --package-path Packages/CrowEngine --filter MissingWorktreeRecovery`
- [ ] On a work session with a managed terminal cwd that is a git worktree and no store row, wait one poll (or create a managed terminal) and confirm `crow list-worktrees` shows a primary row; PR auto-link / merge label then match.

### DIFF
--- a/Packages/CrowEngine/Sources/CrowEngine/EngineTerminalRPCHandlers.swift
+++ b/Packages/CrowEngine/Sources/CrowEngine/EngineTerminalRPCHandlers.swift
@@ -162,6 +162,19 @@ func makeEngineTerminalHandlers(
                 }
                 capturedAppState.terminals[sessionID, default: []].append(terminal)
                 capturedStore.mutate { $0.terminals.append(terminal) }
+                // CROW-1219: if this is a managed agent terminal and the
+                // session still has no worktree row, the cwd we just
+                // persisted is enough to reconstruct one. Poll recovery
+                // covers already-running sessions; this catches the
+                // bootstrap race at create time.
+                if isManaged {
+                    _ = MissingWorktreeRecovery.recoverIfNeeded(
+                        sessionID: sessionID,
+                        appState: capturedAppState,
+                        store: capturedStore,
+                        devRoot: devRoot
+                    )
+                }
                 if trackReadiness {
                     TerminalRouter.trackReadiness(for: terminal)
                 }
--- a/Packages/CrowEngine/Sources/CrowEngine/IssueTracker.swift
+++ b/Packages/CrowEngine/Sources/CrowEngine/IssueTracker.swift
@@ -445,6 +445,17 @@ public final class IssueTracker {
 
     public func refresh() async {
         guard !isRefreshing else { return }
+
+        // CROW-1219: local git + store only — does not create a checkout, and
+        // does not need GitHub. Runs even when the poll is rate-limited so
+        // `primaryWorktree(for:)` / merge labels recover without waiting on
+        // `gh`. Sync on MainActor (no await) so a second refresh cannot
+        // double-append.
+        if let devRoot = ConfigStore.loadDevRoot() {
+            _ = MissingWorktreeRecovery.recoverAll(
+                appState: appState, store: store, devRoot: devRoot)
+        }
+
         guard shouldPoll() else {
             if let suspendedUntil {
                 print("[IssueTracker] skipping refresh — rate-limited until \(suspendedUntil)")
--- a/Packages/CrowEngine/Sources/CrowEngine/MissingWorktreeRecovery.swift
+++ b/Packages/CrowEngine/Sources/CrowEngine/MissingWorktreeRecovery.swift
@@ -1,0 +1,95 @@
+import Foundation
+import CrowCore
+import CrowGit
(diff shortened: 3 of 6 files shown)


dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

- **Existing pathway:** `add-worktree` metadata-only registration when `{path}/.git` already exists, plus `IssueTracker.refresh()` / `new-terminal` as the two natural recovery hooks for store state that lags a live checkout.
- **Green:** The change extends those paths cleanly rather than inventing a parallel mechanism. `WorktreeInference` belongs in `CrowGit` (shared git layer, same package as `GitManager`), and `MissingWorktreeRecovery` in `CrowEngine` correctly gates on `kind == .work`, empty worktree list, and a managed terminal cwd — matching how jobs, reviews, and Managers are intentionally worktree-less.
- **Green:** Running recovery before the rate-limit guard in `IssueTracker.refresh()` is the right shape so `primaryWorktree(for:)` and PR auto-link can succeed on the same poll tick without waiting on `gh`.
- **Green:** `WorktreeInference.resolveMainClone` uses `git rev-parse --git-common-dir` while `ClaudeHookRepair.resolveMainClone` parses the gitdir chain without a subprocess. The poll path can afford git, and tests assert agreement with git; a later consolidation of the file-based resolver into `CrowGit` (for callers that cannot shell out) would be a nice follow-up, not a defect here.

### Security Review
**Strengths:**
- `Validation.isPathWithinRoot` guards both the terminal cwd and inferred `repoPath`, mirroring `add-worktree` and `new-terminal` traversal defenses.
- Branch inference rejects detached `HEAD` and leading-dash names (same option-injection guard as `add-worktree`).
- Git is invoked via `Process` argument arrays with `-C <path>` — no shell interpolation.
- Recovery only considers managed agent terminals, not arbitrary shell tabs; non-work session kinds are excluded.

**Concerns:**
- None identified. Inference registers metadata for an on-disk checkout Crow already launched; it does not execute untrusted input or create new checkouts.

### Code Quality
- **Green:** Idempotency is sound — `worktrees(for:).isEmpty` prevents duplicate rows; `recoverIfNeeded` after `recoverAll` is a no-op once persisted.
- **Green:** `@MainActor` recovery called synchronously from `IssueTracker.refresh()` (also `@MainActor`) avoids double-append races without needing `isRefreshing`.
- **Green:** If a session ever has multiple managed terminals with different cwds, `first(where:)` picks arbitrarily; unlikely in practice (one agent pane per work session).
- **Green:** `IssueTracker.refresh()` hook is not integration-tested, but `MissingWorktreeRecovery` and `WorktreeInference` have focused unit coverage (14 tests total).
- All targeted tests pass (`WorktreeInference` ×8, `Missing worktree recovery` ×6); full `swift build` succeeds.

### Summary Table

---

PR #1228, opened 2026-09-10
Auto-download the corveil CLI from corveil-releases

### LINKED ISSUE #1210: Auto-download and link the corveil CLI from corveil/corveil-releases instead of a manual source build
## Problem

Today the `corveil` CLI is a **manually managed artifact**. An operator must clone `corveil/corveil`, build the binary with a Go toolchain into `out/corveil-<os>-<arch>`, and point Crow at it via `defaults.binaries["corveil"]`. Crow never fetches, builds, or updates it — at startup it only (re)creates the `.claude/bin/corveil` symlink to whatever path is configured.

The result is silent, unbounded drift. Observed on a live machine:

- Linked binary: **v0.3.13** (built 2026-08-13)
- Latest source release: **v0.4.37** (2026-09-05)
- Gap: **429 commits / ~54 releases behind `main`**

A daemon restart repoints the symlink but to the *same stale artifact*, so nothing self-heals. This caused real confusion: a control-plane task was blocked wondering whether `worker.tools.*` verbs were "missing," when part of the ambiguity was just an old CLI.

## Goal

When enabled, Crow keeps a correct, current `corveil` CLI linked for the host platform with **no manual build and no manual path config** — mirroring how `crow version` already checks Crow against its own repo.

## ✅ Prebuilt binaries already exist (public, no auth)

**Download source: [`corveil/corveil-releases`](https://github.com/corveil/corveil-releases) — a public repo** that already ships exactly the matrix we need. Latest release **v0.4.32** carries:

```
checksums.txt
corveil-darwin-amd64
corveil-darwin-arm64
corveil-linux-amd64
corveil-linux-arm64
```

So there is **no upstream artifact prerequisite and no private-repo auth** to solve — this is purely Crow-side download/verify/link/update work. (Earlier revision of this issue wrongly claimed corveil published no binaries; that was checking `corveil/corveil` (source, tag-only, private) rather than the `-releases` mirror.)

**One caveat — mirror lag:** `corveil-releases` trails source by a few patch versions (v0.4.32 vs source v0.4.37, ~1 day). So "latest downloadable" may momentarily lag "latest tagged." Fine for the use case (getting operators off stale builds and roughly current); worth a note, not a blocker. If tighter currency is wanted later, that's a `corveil-releases` publish-cadence ask, not part of this.

## Reusable machinery that already exists

- `crow version --check` / `crow version set [--interval-hours N]` — the self-update-check watcher pattern to copy.
- `.claude/bin` symlink setup at daemon startup (already owns the `corveil` link).
- `crow corveil verify` — runs the binary, returns `{ok, message: "corveil <ver>", path}`; use as a post-download gate.
- `crow corveil reinstall-skill` — re-runs `corveil skill install` for every embedded skill; run after a successful swap and clear the launch-time corveil warning.
- A managed app-support dir already used for other Crow state.

## Proposed design

- **Config:** new `defaults.corveilAutoUpdate` (bool, default **off** initially) + optional `defaults.corveilVersion` (`"latest"` or a pinned tag). Expose via `crow defaults get/set` and Settings → General.
- **Precedence — do not clobber developers:** if an operator explicitly set `binaries["corveil"]` to a real source build, that path **wins** and auto-manage is skipped. Crow manages the binary only when the path is unset or already points into Crow's managed dir. Keeps the local `out/` dev-build workflow intact.
- **Flow:** resolve host `os/arch` → pick `corveil-<os>-<arch>` from the target `corveil/corveil-releases` release → download over HTTPS (public, unauthenticated) → **verify against `checksums.txt`** and confirm `corveil verify` reports the expected version → store under a versioned managed dir (e.g. `~/Library/Application Support/crow/bin/corveil/<version>/corveil`) → **atomically** repoint `.claude/bin/corveil` → `corveil reinstall-skill` → clear the launch warning. Keep the previous version for rollback.
- **Cadence:** check at daemon startup + on the existing version-check interval.
- **Restart semantics:** binary identity is resolved at startup (agent discovery + symlinks), so mirror the existing `restart_required` / `manager_restart_required` signaling, or hot-swap the symlink and flag a Manager restart.
- **macOS note:** downloaded darwin binaries may need the quarantine bit cleared / a notarization check; confirm the `-releases` artifacts are notarized or strip `com.apple.quarantine` on link.

## Acceptance criteria

- With auto-update **on** and no manual override, a fresh `crowd` on a clean machine downloads + links a working `corveil` of the target version; `crow corveil verify` reports it.
- A newer `-releases` version moves the linked binary within one check interval; checksum verified; skills reinstalled; launch warning cleared.
- An operator-set **source-build path is never overwritten**.
- Offline / asset-missing / checksum-mismatch → keep the last-good binary, warn, never break startup.

## Out of scope / open questions

- Signing / notarization / quarantine policy for downloaded darwin binaries.
- Whether other agent CLIs should be managed the same way later.
- Default posture: pin-to-tested vs. float-to-latest.
- `corveil-releases` publish cadence (currently trails source ~1 day) — only relevant if strict currency is required.

---
_Filed from the Manager. Reasonable defaults chosen per repo convention; no external blocker — prebuilt public binaries already exist in `corveil/corveil-releases`._

dgershman:
Closes #1210

## Summary
- Opt-in `defaults.corveilAutoUpdate` (default off) plus `defaults.corveilVersion` (`latest` or a pin), exposed via `crow defaults get/set` and Settings → General.
- When enabled and `binaries["corveil"]` is unset or already in Crow's managed dir, Crow downloads the host-platform asset from public `corveil/corveil-releases`, verifies `checksums.txt`, stores it under Application Support, hot-swaps `.claude/bin/corveil`, and reinstalls skills. An operator-set source-build path is never overwritten.
- Failures (offline, missing asset, checksum mismatch) keep the last-good binary and never break startup. Cadence is daemon start plus the existing `crow version` interval.

## Test plan
- [ ] `crow defaults set --corveil-auto-update true` on a machine with no `binaries["corveil"]`; after the next check, `crow corveil verify` reports the `-releases` version.
- [ ] A newer `-releases` tag is picked up within one version-check interval; checksum verified; skills reinstalled.
- [ ] `crow defaults set --binary corveil=/path/to/out/corveil-*` is left alone while auto-update is on.
- [ ] Offline / checksum mismatch leaves the last-good binary linked; `crowd` still starts.
- [ ] Settings → General shows the Auto-download toggle and version field.


dhilgaertner (changes requested):
## Code & Security Review

### Critical Issues (if any)
None.

- **Existing pathway:** `VersionUpdateService` / `VersionUpdateClient` — periodic GitHub release checks with injectable transport, interval gating, and a background poll wired from `CrowDaemon.run`. This PR correctly extends that pattern rather than inventing a parallel mechanism: `CorveilReleaseClient` mirrors the transport shape, `CorveilAutoUpdate` holds testable policy (ADR 0012), and `startCorveilAutoUpdatePoll` reuses `versionUpdate.intervalHours` cadence.
- **Green:** `Scaffolder.replaceBinarySymlink` is the right surgical hook for hot-swap without running the full `installBinarySymlinks` reap loop — matches how binary overrides are materialized today.
- **Green:** Operator override semantics (`shouldAutoManage` / `isManagedPath`) align with the existing `defaults.binaries["corveil"]` + scaffold symlink model documented in ADR 0025.

### Security Review
**Strengths:**
- Defense-in-depth on the download path: HTTPS to GitHub, `maxAssetBytes` cap (pre- and post-download), SHA-256 verification against `checksums.txt`, and a post-download `corveil --version` check with prefix-safe version matching (`CorveilAutoUpdate.verifyMessage`).
- Operator source-build paths are never overwritten; auto-manage only runs when `binaries["corveil"]` is unset or already under the managed Application Support tree.
- Version pins are validated at every write boundary (`ConfigDefaults.normalizedCorveilVersion`, CLI `validateCorveilVersion`, `DefaultsRPC.patchCorveilVersion`) — path-like values cannot reach the managed-dir layout.
- Opt-in default (`corveilAutoUpdate: false`); failures are non-fatal and keep the last-good binary.
- Atomic partial write + executable-bit set + `0700` version directories; Gatekeeper quarantine stripped on macOS before verify/exec.

**Concerns:**
- **Green:** `release.tag` from the GitHub API is used directly as a filesystem directory name without re-applying `normalizedCorveilVersion`. Acceptable given the trust model (content from `corveil/corveil-releases` is already fully trusted; a malicious tag would be indistinguishable from a compromised release), but applying the same tag validator before `versionDirectory` would be cheap defense-in-depth.

### Code Quality
- **Yellow:** `CorveilAutoUpdateService.performCheck` returns `state: .updated` even when `replaceBinarySymlink` fails (`CorveilAutoUpdateService.swift:223-237`). In that case the managed binary and `binaries["corveil"]` config are updated, but `{devRoot}/.claude/bin/corveil` may still point at the previous target — the hot-swap path agents actually resolve via PATH prepend. The log line is easy to miss; the status misreports success. Consider returning `.failed` (or a distinct partial state) when the symlink swap fails, or retrying scaffold's symlink path before reporting `.updated`.
- **Green:** Unit coverage is solid across policy (`CorveilAutoUpdateTests`), client parsing (`CorveilReleaseClientTests`), service flows (override skip, checksum mismatch, successful install), config round-trips, and CLI/RPC wiring.
- **Green:** `pruneOldVersions` comment mentions only pruning dirs that contain `corveil`/`.partial`, but the implementation prunes by mtime alone. Low risk (managed root is Crow-owned), but the comment oversells the guard.

### Static Analysis
- `swift build` — pass.
- `swift test --package-path Packages/CrowDaemon` — 625 tests pass (includes all `CorveilAutoUpdateServiceTests`).
- `swift test --package-path Packages/CrowEngine` — 818/819 pass; the single failure (`BoardPollOffMainActorTests.fetchOffActorDoesNotRunOnMainThread`) is a pre-existing timing flake unrelated to this diff (CROW-1179).
- `swift test --package-path Packages/CrowCore --filter Corveil` — pass.

### Summary Table

---

dhilgaertner (approved):
## Code & Security Review

### Critical Issues (if any)
None.

- **Existing pathway:** `VersionUpdateService` / `VersionUpdateClient` (CROW-938) — periodic GitHub check with injectable transport, cached status, and poll loop tied to `versionUpdate.intervalHours`.
- **Green:** The PR extends that pattern rather than inventing a parallel updater. Policy lives in `CorveilAutoUpdate` + `CorveilReleaseClient` (CrowEngine, unit-testable per ADR 0012); network, config writes, and skill reinstall live in `CorveilAutoUpdateService` (CrowDaemon). `Scaffolder.replaceBinarySymlink` is the right hot-swap hook — it avoids the full `installBinarySymlinks` reap loop that would drop unrelated bin links.
- **Green:** ADR 0025 documents the opt-in default, operator-override precedence, failure semantics, and quarantine workaround. Config/CLI/RPC validation is centralized in `ConfigDefaults.normalizedCorveilVersion`, keeping the three writers in sync.

### Security Review
**Strengths:**
- Defense in depth on the download path: GitHub asset metadata size cap, in-memory `maxAssetBytes` guard in `CorveilReleaseClient.download`, SHA-256 verification against `checksums.txt`, post-download `corveil verify`, and `verifyMessage` token-boundary matching before accepting a binary (`CorveilAutoUpdateService.swift`, `CorveilAutoUpdate.swift`).
- Operator source-build paths are never overwritten (`shouldAutoManage` / `isManagedPath`); user-controlled pins reject path-like values before they become directory names (`AppConfig.swift`).
- Managed install dirs are created `0700`; binaries written `0755` via atomic `.partial` → rename. Failures never throw out of the poll loop — last-good binary stays linked.
- Opt-in default (`corveilAutoUpdate: false`) preserves existing workflows; supply-chain trust is explicitly scoped to the public `corveil/corveil-releases` repo (ADR 0025).

**Concerns:**
- **Green:** When `corveilVersion` is `latest`, `release.tag` from the GitHub API is used directly as a filesystem path component without re-running `normalizedCorveilVersion`. This is acceptable given the trusted publisher model documented in ADR 0025; a defense-in-depth check could reject unexpected tag shapes before `versionDirectory`.
- **Green:** Quarantine stripping on macOS is a documented pragmatic workaround (ADR 0025) until artifacts are notarized.

### Code Quality
- **Green:** Solid test coverage across layers — parsing (`DefaultsCommandParsingTests`), normalization (`AppConfigTests`, `DefaultsRPCSupportTests`), pure helpers (`CorveilAutoUpdateTests`), release client transport mocks (`CorveilReleaseClientTests`), and end-to-end service flows including checksum mismatch and operator override (`CorveilAutoUpdateServiceTests`).
- **Green:** `defaults-set` kicks an immediate `runCheck()` when auto-update is enabled or the version pin changes (`SettingsRPCHandlers.swift`); the poll loop also short-circuits its sleep when auto-update is newly enabled before the first check completes (`CorveilAutoUpdateService.swift`).
- **Green:** If `replaceBinarySymlink` fails after a successful download, the service still returns `.updated` while logging the symlink failure (`CorveilAutoUpdateService.swift:223-237`). Config and on-disk binary are correct and a daemon restart would re-materialize the link; consider a distinct outcome if auto-update status is surfaced to operators later.
- **Green:** `swift build` succeeds; targeted test suites for `CorveilAutoUpdate` pass in CrowEngine, CrowDaemon, CrowCLI, and CrowCore.

### Summary Table

---

PR #1269, opened 2026-09-17
Canonicalize aliased PR link owners for PR-status matching (CROW-1268)

### LINKED ISSUE #1268: PR-status icons stay empty when a session's .pr link uses a redirected repo-owner alias (e.g. post-rename org name)
## Summary

A session's PR-status icons (CI checks, review state, auto-merge, mergeability) never populate when the registered `.pr` link's URL uses a **repo-owner alias that GitHub redirects** (e.g. an old org name after a rename) instead of the canonical `owner/repo`. GitHub resolves the alias via redirect, but Crow's PR-status poller matches the link's owner/repo as a **literal string** against the workspace's canonical repo, so it never correlates the link to a PR to fetch status for — the row stays iconless even though the PR is healthy.

## Repro (observed)

- Session `corveil-3453-admin-403-error-shape` had a `.pr` link `https://github.com/**RadiusMethod**/corveil/pull/3466`.
- The repo was renamed org `RadiusMethod` → `corveil`; `gh api repos/RadiusMethod/corveil` → `full_name: corveil/corveil` (redirect). The PR is fine: #3466 OPEN, **APPROVED**, `mergeStateStatus: BLOCKED`, head `feature/corveil-3453-admin-403-error-shape` (matches the registered worktree).
- Every healthy session in the same workspace had `.pr` links on the canonical `https://github.com/**corveil**/corveil/pull/...` and showed status icons normally. Only the alias-owner link was iconless.
- Re-pointing the link to the canonical `corveil/corveil` URL (via `crow edit-link --new-url …`) + a `refresh-tickets` restored the icons — confirming the owner string mismatch is the sole cause.

## Root cause

The PR-status correlation compares the `.pr` link's parsed `owner/repo` to the workspace's configured provider repo (canonical `corveil/corveil`) as an exact string. An alias owner (`RadiusMethod`) that GitHub 301-redirects is treated as a different repo, so the poller either skips the link or can't join it to the PR it's already polling. (Likely in the issue-tracker poller / the review board's PR-status join — wherever a session's `.pr` link is matched to fetched PR data.)

## How links end up with an alias owner

The coder self-registers the PR link (skill step 7). If the URL comes from a **git-remote-derived source** — `gh pr create`'s echoed URL, or a URL constructed from `origin` whose remote still uses the old org name (`git@github.com:RadiusMethod/corveil.git`) — the owner is the alias. The healthy sessions used `gh pr view --json url` (which returns GitHub's **canonical** URL, `corveil/corveil/...`). So the failure is remote-name-dependent and coder-path-dependent, not deterministic.

Note: repointing the affected clone's remote to `git@github.com:corveil/corveil.git` fixes it going forward for that workspace, but Crow should be robust to alias URLs regardless, since org renames + stale remotes are common.

## Proposed fix (Crow side)

- **Normalize / canonicalize the `.pr` link's owner-repo before matching** — resolve the GitHub repo redirect once (e.g. `GET /repos/{owner}/{repo}` → `full_name`, cached) and/or compare against the workspace repo by canonical id/full_name rather than the raw URL string. An alias-owner PR URL that redirects to the workspace's canonical repo should match.
- Alternatively (or additionally), **at link-registration time** normalize the stored PR URL to canonical `owner/repo` (follow the redirect once when `add-link --type pr` / the issue-tracker poller first sees it), so the stored link and all downstream matching use the canonical form.
- Belt-and-suspenders: if the poller can't correlate a `.pr` link to any polled PR, surface a subtle "unlinked / owner-mismatch" hint rather than a silently blank status, so this is diagnosable without hand-inspecting link URLs.

## Impact

Low-frequency but silent and confusing: after any repo/org rename (or when a clone's remote lags a rename), a coder-registered PR link can render permanently statusless, and the session looks stuck in review with no CI/merge signal despite a green, approved PR. Cheap to make robust; hard to notice otherwise.

## References

- Affected session: `corveil-3453-admin-403-error-shape` (PR corveil/corveil#3466).
- Skill contract: `/crow-workspace` step 7 registers the PR link via `gh pr view --json url` (canonical) — but the poller should not depend on every coder taking that exact path.

dhilgaertner:
Closes #1268

## Problem

A session's `.pr` link registered on a stale GitHub **owner alias** — the old org name after a rename, which GitHub 301-redirects — never populated its PR-status chips (CI / review / auto-merge / mergeability). The PR is healthy and approved; the row just stays iconless.

Repro from the ticket: session `corveil-3453-admin-403-error-shape` had a `.pr` link on `https://github.com/**RadiusMethod**/corveil/pull/3466`. The repo was renamed `RadiusMethod` → `corveil`; `gh api repos/RadiusMethod/corveil` → `full_name: corveil/corveil` (redirect). PR `corveil/corveil#3466` was OPEN + APPROVED, yet iconless — while every sibling session on the canonical `corveil/corveil` URL showed status normally.

## The matcher that failed

Every downstream matcher keys on the **stored link URL** derived from `prLink.url`, and both places that consume it break on an alias owner:

1. **`PRStatusController.applyPRStatuses`** (and `AutoMergeController` / `AutoRebaseController` / `AutoReReviewController`): `let byURL = Dictionary(viewerPRs.map { ($0.url, $0) })` then `guard let pr = byURL[prLink.url]`. `byURL` is keyed on GitHub's **canonical** `pr.url`; an alias-owner `prLink.url` never matches → the chip is never set.
2. **`BoardPoller.fetchStalePRStates`**: builds `PRRef(owner: p.org…)` from the alias URL and calls GitHub **GraphQL** `repository(owner:name:)`, which — unlike the REST `/repos/{owner}/{repo}` endpoint — does **not** follow renames → null → merged/closed state never surfaces.

## Fix — canonicalize the one stored URL (self-healing on the poll)

Because both failure sites derive from `prLink.url`, canonicalizing that single stored value repairs all of them with **no per-matcher change**:

- **`CodeBackend.resolveCanonicalRepoSlug(_:)`** — GitHub resolves `owner/repo` → canonical `full_name` via the REST `/repos/{owner}/{repo}` endpoint, which **301-follows renames** (the GraphQL path used elsewhere does not — that's the root cause). Default `nil` for backends that can't canonicalize (GitLab/Corveil), scoping the change to GitHub.
- **`PRLinkReconciler.canonicalizeAliasedPRLinks(knownPRURLs:)`** — a poll pass that, for each `.pr` link the payload couldn't match, resolves its canonical slug and rewrites the stored link URL to canonical (persisted like `edit-link`). All matchers then correlate on the next poll.

### Approach chosen: canonicalize-at-match vs normalize-at-registration

I chose **canonicalize-at-match (self-healing on the poll)** over normalizing at `add-link` time. The ticket asks that "the poller should not depend on every coder taking that exact path," and the impact note wants robustness to alias URLs "regardless, since org renames + stale remotes are common." Registration-time normalization wouldn't self-heal a rename that lands *after* a link was registered, nor a link registered from a stale-remote source; the poll pass does. (The org's documented "solve the current case before generalizing" preference is respected: the change is scoped to the observed GitHub-alias case, not a general URL-rewrite framework.)

### Cost: at most once per repo, never per poll

- **Payload-miss gate**: a healthy canonical link is already in the poll payload (`viewer PRs ∪ stale follow-up`), so it is never a candidate and **never triggers a lookup** — the steady-state path adds zero latency.
- **Per-repo cache** (`canonicalRepoSlugCache`): each distinct repo resolves at most once for the daemon's lifetime; canonical and definitively-missing (404) slugs cache to themselves, so they don't re-query. Transient failures aren't cached, so they retry on a later poll.

The rewrite is persisted, so chips populate on the next poll's status pass — a one-poll, self-healing lag for a rare correction. A `CrowLog.automation` line records each rewrite, so an owner mismatch is diagnosable without hand-inspecting link URLs (the ticket's belt-and-suspenders ask).

## Tests

`PRLinkCanonicalizeTests` (swift-testing, the `CrowEngineTests` convention):

- **Pure matcher, redirect mocked**: an alias owner rewrites to canonical (number + trailing path preserved); a canonical owner / unknown slug / case-only difference is **not** rewritten; non-`/pull/`, GitLab, and GitHub-Enterprise URLs are rejected.
- **End-to-end**: an alias-owner link is canonicalized, then `applyPRStatuses` correlates the canonical payload URL and the chip populates (`.approved`).
- **No redirect on the steady-state path**: a canonical link already in the payload triggers **zero** `gh api repos` calls (call-counting mock `ShellRunner`).
- **At most once per repo**: a payload-miss candidate resolves its slug once and caches it; a second poll makes no further API call. A 404 slug caches and doesn't loop.

`swift test` green for both touched targets: **CrowProvider** (155 XCTest + 52 swift-testing) and **CrowEngine** (new suite + existing PR-status/reconcile/auto-* suites). The one flake in the full parallel CrowEngine run (`fetchOffActorDoesNotRunOnMainThread`, a `< 0.1s` timing assertion in `BoardPoller` I don't touch) passes in isolation.


dgershman (approved):
## Code & Security Review

### Critical Issues (if any)
None.

- **Existing pathway:** `PRLinkReconciler` (session ↔ `.pr` link writes through the injected `JSONStore`) plus `CodeBackend` (ADR 0005). Downstream chips and watchers already join on the stored link URL (`PRStatusController.applyPRStatuses`, auto-merge/rebase/re-review, `BoardPoller.fetchStalePRStates`); GitHub GraphQL `repository(owner:name:)` does not follow renames, while REST `/repos/{owner}/{repo}` does.
- This extends that path in the right place: one persisted canonical URL, no per-matcher alias logic, GitHub-only via a defaulted protocol method. Payload-miss gating plus the per-repo cache match how the rest of the poller avoids extra `gh` work.

### Security Review
**Strengths:**
- `resolveCanonicalRepoSlug` shells `gh api` as direct argv (`GitHubCodeBackend.swift:86-88`), same pattern as other backend methods — no `sh -c` around the slug.
- `parseGitHubPRURL` rejects non-`github.com` hosts (GitLab, GHE), so the REST redirect-follow cannot be pointed at an arbitrary host.
- Steady-state canonical links in the poll payload never hit the network; 404s cache as identity; transients are not cached and go through `handleGitHubBackendError`.
- Rewrites use the shared injected `JSONStore` (not a throwaway instance — #728) and skip a same-session duplicate canonical URL.

**Concerns:**
- **Green:** `canonicalizedPRURL` skips a rewrite when the resolved `full_name` differs only in case (`PRLinkReconciler.swift:850-855`, locked in by `caseOnlyDifferenceIsNotRewritten`). `byURL[prLink.url]` is a case-sensitive `Dictionary` key, so a stored `Corveil/crow` link still will not join a payload URL that uses GitHub's `corveil/crow` casing. Out of ticket scope (this bug is an org *rename*), but persisting GitHub's exact `full_name` would close that remaining exact-URL miss without extra API cost.

### Code Quality
- New helpers are pure and unit-tested; the live pass has end-to-end correlation, payload-miss, once-per-repo cache, and 404-loop tests in `PRLinkCanonicalizeTests.swift`.
- Local `swift test --package-path Packages/CrowEngine --filter PRLinkCanonicalize`: 11/11 passed. Existing `GitHubCodeBackend` BackendsTests still pass (16/16).
- **Green:** canonicalize runs *after* `applyFetchedBoard` (`IssueTracker.swift:418` then `:436-438`), so chips wait until the next poll even when the canonical PR is already in `viewerPRs`. That lag is documented and is the safe order — `applyPRStatuses` has side effects (auto-merge / needs-refine), so a second status pass in the same cycle is not free. Fine as accepted.
- **Green:** `resolveCanonicalRepoSlug`'s 404-vs-throw split is exercised through the engine mock, but `BackendsTests` has no direct cases for malformed slug / 404 / non-404 throw, unlike every other `GitHubCodeBackend` method in that file.

### Summary Table

---

